PR #41, opened 2014-03-18
Refactor zen_href_link (catalog) and zen_catalog_href_link (admin)

lhungil:
- Fixed the wrong separator in zen_href_link when sid must be passed with static URLs and no other parameters.
- Updated the code handling sessions (zenid) to use "constant('SID')" instead of only "SID".
- Consolidated the link "cleanup" code in zen_href_link into one rtrim, one preg_replace, and one str_replace.
- Refactored connection check in zen_href_link to default to HTTP and issue a WARNING in the debug log if an unknown connection type is specified.
- Added a notifier to zen_href_link. This should allow observers to "listen" and "act" upon the function being called.
- Updated the admin function zen_catalog_href_link to more closely match the catalog zen_href_link.

<sub>**Note:** Should the new language constant (used when issuing a WARNING about unknown connection types) go into the main language file or into a new file under extra_definitions? My initial thoughts were in the main language file as they affect every page (and requires one less file load by the web server / PHP). For the purposes of this patch I've made the edits directly to the main language file(s). These may need to be moved and / or changed.</sub>

### DIFF
--- a/includes/functions/html_output.php
+++ b/includes/functions/html_output.php
@@ -183,7 +183,7 @@
       }
       else if ( ( ($request_type == 'NONSSL') && ($connection == 'SSL') && (ENABLE_SSL == 'true') ) || ( ($request_type == 'SSL') && ($connection == 'NONSSL') ) ) {
         if ($http_domain != $https_domain) {
-          $link .= zen_output_string(zen_session_name() . '=' . zen_session_id());
+          $link .= $separator . zen_output_string(zen_session_name() . '=' . zen_session_id());
         }
       }
     }


admin/includes/functions/html_output.php
```diff
@@ -45,27 +45,167 @@ function zen_href_link($page = '', $parameters = '', $connection = 'NONSSL', $ad
...
+   * @param string $static true when $page is a file / web resource,
+   *   false otherwise. Default is false.
+   *
+   * @param string $use_dir_ws_catalog true to include the website catalog
+   *   directory when creating the link, false otherwise. Default is true.
+   *
+   * @return string
+   */
+  function zen_catalog_href_link($page = '', $parameters = '', $connection = 'NONSSL', $search_engine_safe = true, $static = false, $use_dir_ws_catalog = true) {
+    global $zco_notifier;
```
torvista, line 106: What would be an example of the use of the $static parameter?

PR #251, opened 2014-12-23
MySQL GROUP BY clauses should list all SELECTed columns

drbyte:
Not including all SELECTed fields can result in ambiguous GROUP BY results, and is non-standard SQL

Ref: https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/251)


zcwilt:
:+1:

ajeh:
:+1:

PR #252, opened 2014-12-23
Remove 0001-01-01 dates for MySQL 5.7 compatibility

drbyte:
MySQL 5.7.4 is retiring support for "zero dates" such as 0001-01-01 and 0000-00-00
Acceptable "old" dates are 1000-01-01 or epoch 1970-01-01

NOTE: This commit is missing an "upgrade script" to convert existing data and fields.

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/252)


drbyte:
NOTE: I'm working on some regression testing, and experimenting with various MySQL modes. 

It's possible some of this may be moot. More to follow.

PR #256, opened 2014-12-27
Add function to return customers_email_address

ajeh:
[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/256)


drbyte:
I'm torn whether it's best to put this in functions_email or elsewhere.

I think this is fine for now, but when we refactor the lookups stuff to consolidate customer data into a class this should move to there.

So, for now: :+1:

drbyte:
:+1:

zcwilt:
:+1:

PR #262, opened 2015-01-05
Template markup for Order attribs list

drbyte:
ref: http://www.zen-cart.com/showthread.php?215429-tpl_account_history_info_default-php-id-instead-of-class-in-attributes-list

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/262)


ajeh:
:+1:

zcwilt:
:+1:

PR #269, opened 2015-01-09
Customers-Approval auth setting loops

drbyte:
With the `Admin-Configuration->Customer Details->Customer Approval Status - Authorization Pending` setting set to `1 = Must be Authorized to Browse`, a couple problems can occur:

- the in-page ajax handler code may cause the login form to be displayed multiple times on the page.
- if Down-For-Maintenance is enabled, the login page gets stuck in a "too many redirects" loop.

### DIFF
--- a/includes/init_includes/init_customer_auth.php
+++ b/includes/init_includes/init_customer_auth.php
@@ -82,8 +82,8 @@ switch (true) {
   /**
    * bypass redirects for these scripts, to processing regardless of store mode or cust auth mode
    */
-  case (preg_match('/ipn_main_handler\.php$/', $_SERVER['SCRIPT_NAME'])):
-  case (preg_match('/ajax\.php$/', $_SERVER['SCRIPT_NAME'])):
+  case (preg_match('|_handler\.php$|', $_SERVER['SCRIPT_NAME'])):
+  case (preg_match('|ajax\.php$|', $_SERVER['SCRIPT_NAME'])):
   break;
 
   // if DFM is in strict mode, then block access to all pages:
@@ -138,8 +138,8 @@ switch (true) {
   /**
    * bypass redirects for these scripts, to processing regardless of store mode or cust auth mode
    */
-  case (preg_match('/ipn_main_handler\.php$/', $_SERVER['SCRIPT_NAME'])):
-  case (preg_match('/ajax\.php$/', $_SERVER['SCRIPT_NAME'])):
+  case (preg_match('|_handler\.php$|', $_SERVER['SCRIPT_NAME'])):
+  case (preg_match('|ajax\.php$|', $_SERVER['SCRIPT_NAME'])):
   break;
 
 /**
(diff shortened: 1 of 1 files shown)


texdc:
:+1:

includes/init_includes/init_customer_auth.php
```diff
@@ -79,27 +79,34 @@
  * 2 = customer authorization pending can browse but no prices
  */
 switch (true) {
+  /**
+   * bypass redirects for these scripts, to processing regardless of store mode or cust auth mode
+   */
+  case (preg_match('/ipn_main_handler\.php$/', $_SERVER['SCRIPT_NAME'])):
```
lhungil: Do we want to match '/some-random-folder-inside-zencart-installation-folder/ipn_main_handler.php'? Or should it be a more specific?

EX: '~^' . ($request_type == 'SSL' ? DIR_WS_HTTPS_CATALOG : DIR_WS_CATALOG) . 'ipn_main_handler\.php$~'?

drbyte: The intent was that it would be directory-agnostic, thus leaving off the ^ in the regex pattern.

Is there a use-case where this fails to work as intended?

drbyte: I suppose changing the delimiter to `|` could be good, and then also make the beginning `/` become a literal character, thus attempting to ensure that it's a match on the complete filename and not just a suffix to a filename.

Or should it be even more generic, as in `|_handler\.php$|` so that other scripts can be easily supported?

I had also considered that it might be ideal to add notifier hook support so that additional plugins could register themselves for inclusion here. Haven't figured out how best to do that yet, so was leaving that for a separate PR.

lhungil: Just thinking about the case where someone adds a file with the same name in a subdirectory. More of an edge case, but thinking about 3rd party code adding a page, module, or template. Would it be desirable for such a modification to be always accessible irregardless of the customer approval setting?

lhungil: Agreed, a separate PR sounds like a good idea for adding hooks.

drbyte: I hadn't considered forcing it to test which directory it was in, or jailing/sandboxing it.

I guess the answer to your question about "desirable" is: it depends on what that "modification" was for in the first place.

lhungil: When it all comes down to it, I'm good with this change either way. Was just curious if restricting the checks to the files shipped with Zen Cart might help avoid mistakes in 3rd party code...

But as you point out, some 3rd party code need to allow access to a PHP script when customer approval is not yet known (or pending). I'm all for adding some sort of hook in the future.

:+1:

texdc: `|_handler\.php$|` sounds good to me

zcwilt:
While I would like to see the customer approvals stuff be re-written, can't see any problem with the code as posted.

:+1:

ajeh:
:+1:

PR #273, opened 2015-01-14
Port the ajax framework code from v1.5.4 forward, including PCI empowered pmt mods

drbyte:
Port the ajax framework code from v1.5.4 forward, 
including PCI empowered payment modules


zcwilt:
:+1: 
It should be noted, that these code changes may have an effect on COWOA type code. we will look to mitigate this later.

ajeh:
:+1:

PR #275, opened 2015-01-14
Fix authnet AIM field type for MySQL strict mode

drbyte:
Fix authnet AIM field type in bindVar query preparation

Can't send a `string` to an `integer` field when MySQL is in `strict` mode.


zcwilt:
:+1:

ajeh:
:+1:

PR #277, opened 2015-01-26
Update sessions.php to allow session_id to contain comma and dash

ksteslicki:
Update preg_replace regex to include commas and minus signs for session_id

Ref: http://www.zen-cart.com/showthread.php?215961-Timeout-during-final-step-of-checkout-Solution


drbyte:
History:

The regex pattern was added in v1.3.8 and later applied in all 3 functions in v1.3.9e along with a few logic changes to mitigate against false-positives triggered during testing for potential Format String Attacks.

Since then PHP has added additional options for session id generation, via the setting for `session.hash_bits_per_character` (PHP >= 5.0), which allow for additional characters: `,` and `-`.  Documented further at http://php.net/manual/en/session.configuration.php#ini.session.hash-bits-per-character

So, I agree with the proposed regex pattern change.

drbyte:
Will wait for additional peer review throughout the week.

zcwilt:
:+1:

ajeh:
:+1:

lhungil:
:+1:

PR #279, opened 2015-01-27
remove extraneous call to $pInfo

drbyte:
This segment of code is not needed, and throws an error on PHP 5.4+ due to the object not being defined. Removal is the best option.


zcwilt:
:+1:

ajeh:
:+1:

PR #282, opened 2015-02-07
Fix detection of whether "cheapest shipping selector" should be triggered

drbyte:
Fix detection of whether "cheapest shipping selector" should be triggered

Detection should be based on the 'id' element, not the 'shipping' element alone.

Without this fix the cheapest shipping calculation will never run. Was broken in v1.5.3 and v1.5.4.


ajeh:
:+1: 
This has been tested and is working correctly.

zcwilt:
:+1:

PR #283, opened 2015-02-07
Let plugin version checker handle null responses

### LINKED ISSUE #281: Plugin Support Not working
For certain plugins:
Example: http://www.zen-cart.com/downloads.php?do=file&id=1781
Are not returning the json
https://www.zen-cart.com/downloads.php?do=versioncheck&id=1781

Which is causing:
PHP Warning:  in_array() expects parameter 2 to be array, null given in /home/andeza200/public_html/includes/functions/plugin_support.php on line 42

See:
https://github.com/zencart/zc-v1-series/blob/v160/includes/functions/plugin_support.php#L42

drbyte:
Let plugin version checker handle null responses
Fixes #281

### DIFF
--- a/includes/functions/plugin_support.php
+++ b/includes/functions/plugin_support.php
@@ -34,7 +34,7 @@
     if ($plugin_file_id == 0) return FALSE;
     $new_version_available = FALSE;
     $lookup_index = 0;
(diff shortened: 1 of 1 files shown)


zcwilt:
:+1:

ajeh:
:+1: 
This eliminates errors when unable to complete the call. Good catch.

PR #285, opened 2015-02-10
Add doc comment about compatibility

drbyte:


ajeh:
:+1:

texdc:
:+1: Note: `5.4` will be EOL by the end of the year.  It's best to move to `5.5` min asap (and update the code-base!).

drbyte:
I'd love to force a 5.5 minimum right away. But most of the real benefits of PHP improvements were in 5.4 anyway. Lots of code updates could be done to embrace those, and would be a welcome improvement.

But hosts barely have 5.3 as a minimum yet. We have to consider 250,000 existing users, not only new installs.

We are indeed pushing that forward for ZC by requiring 5.4 minimum. And if for some horrible reason v1.6.0 can't be released by year end then we'll update PHP minimum requirements again at that time.

zcwilt:
:+1:

texdc:
@drbyte http://phpversions.info/

PR #286, opened 2015-02-10
Update pwd compatibility layer to v1.0.4

drbyte:
Update pwd compatibility layer (from ircmaxell) to v1.0.4

(Will consider relocating package with other packages like aura, but as a separate commit.)


zcwilt:
:+1:

ajeh:
:+1:

PR #288, opened 2015-02-11
Add "hostname" override for email generator.

drbyte:
Add "hostname" override for email generator.

Although rare, some heavily "shared" servers are configured to handle sub-sub-domains, and the auto-detection of the email hostname isn't always accurate, especially if SERVER_NAME isn't passed correctly to PHP.

This change allows the ability to specify the hostname to use in Message-Id and Received headers and as default HELO string.

To use: define EMAIL_HOSTNAME in the extra_configures or extra_datafiles folder on both catalog and admin sides.

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -334,6 +334,10 @@
       @ini_set('mail.add_x_header', 0);
 
       $ErrorInfo = '';
+
+      // set Hostname, since it can aid in delivery of emails. If emails are being rejected, comment out the following line and try again:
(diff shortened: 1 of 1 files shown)


includes/functions/functions_email.php
```diff
@@ -334,6 +334,10 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
       @ini_set('mail.add_x_header', 0);
 
       $ErrorInfo = '';
+
+      // set Hostname, since it can aid in delivery of emails. If emails are being rejected, comment out the following line and try again:
+      $mail->Hostname = defined('EMAIL_HOSTNAME') ? EMAIL_HOSTNAME : preg_replace('~(^https?://|\/.*$)~', '', defined('HTTP_CATALOG_SERVER') ? HTTP_CATALOG_SERVER : HTTP_SERVER);
```
texdc: Please keep line lengths to 80 chars or less.

ajeh:
:+1:

PR #291, opened 2015-02-11
Fix call to nonexistent ErrorMsg() method

drbyte:
zc_install has some legacy SQL execution statements that make a call to $this->db->ErrorMsg() which hasn't existed for several years.

This is refactoring to remove those constructs, and to trigger regular logging and error reporting instead.

### DIFF
--- a/zc_install/includes/classes/class.zcDatabaseInstaller.php
+++ b/zc_install/includes/classes/class.zcDatabaseInstaller.php
@@ -34,7 +34,7 @@ class zcDatabaseInstaller
     $this->dbPrefix = $options['db_prefix'];
     $this->dbCharset = $options['db_charset'];
     $this->dbType = in_array($options['db_type'], $dbtypes) ? $options['db_type'] : 'mysql';
-    $this->debugLevel = isset($options['debug']) ? $options['debug'] : FALSE;
+    $this->dieOnErrors = isset($options['dieOnErrors']) ? (bool)$options['dieOnErrors'] : FALSE;
     $this->errors = array();
     $this->basicParseStrings = array(
     'DROP TABLE IF EXISTS ',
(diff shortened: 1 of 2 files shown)


texdc:
[Boolean arguments](http://www.togsblom.com/2009/01/boolean-arguments-are-trouble.html) are code smells.  Use [SRP](http://en.wikipedia.org/wiki/Single_responsibility_principle) instead.

Regardless, there's no need to pass in a flag value.  The original had a `die` hard-coded, so the value passed to `$db->Connect()` should be `true`.

drbyte:
> Regardless, there's no need to pass in a flag value. The original had a `die` hard-coded, so the value passed to `$db->Connect()` should be `true`.

This second paragraph doesn't make sense.  The code being changed replaced `false` with the flag. So why are you saying the value passed should be `true`?

texdc:
> The original had a die hard-coded

Self-explanatory.  Unless, you're trying to change functionality, the code should `die` on error.

drbyte:
Oy. Apparently we were talking about 2 different things. The boolean named $dieOnErrors was in the getConnection method, and when set to true, triggers a die according to respective rules. When you were saying "the original had a die hardcoded", I was still reading in the context of that same method, not the `on die` you had jumped to down in the doCompletion method. 

So, yes, indeed, it should `die` on error ... which is the whole reason why $dieOnError was adapted here. 

You can't set die-on-error to true globally, as there are legitimate times during zc_install when you don't want it to halt execution.

Will move it elsewhere so that a class property is set instead.

zcwilt:
:+1: given the problems with legacy queryFactory. have no problems with the changes here.

ajeh:
:+1:

texdc:
While I'd still prefer to avoid `bool` flags in arguments, this is a better solution.  Thanks!

PR #292, opened 2015-02-12
Send 503 header on db error message

drbyte:
When a database error occurs, we output a generic message, but it still sends a normal 200 response header, which doesn't tell search engines to skip it.

This sends the 503-service-unavailable header to tell SE's to come back later instead of indexing the error message.

(We were already sending the 503 header, but only when also redirecting to nddbc.html when a "not connected" condition was found ... which is too narrow criteria.)

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -142,8 +142,10 @@ class queryFactory extends base {
   }
 
   function show_error() {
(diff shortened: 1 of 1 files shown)


texdc:
:+1: (kinda) the `show_error()` method doesn't belong in a db connection class.  It's better to throw an exception and decide how to communicate that in an exception handler.

Can we remove all the error handling from `queryFactory` (in another PR)?  I don't mind working on it.

zcwilt:
:+1:

ajeh:
:+1:

PR #302, opened 2015-02-22
Ensure IP address is not blank

drbyte:
If all proxy-detections fail to determine the correct IP address, set it to a single dot (.) to denote this fact.

Ref: http://www.zen-cart.com/showthread.php?214518&p=1257981#post1257981

Also moved the sanitization logic into the same function to keep it in context.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -1,10 +1,10 @@
 <?php
 /**
  * @package admin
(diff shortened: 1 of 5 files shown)


zcwilt:
:+1: Looks OK to me.

ajeh:
:+1:

PR #303, opened 2015-02-22
Update proxy/load-balancer detection to accommodate X-Forwarded-Port headers

drbyte:
Update proxy/load-balancer detection to accommodate X-Forwarded-Port headers

While all the existing proxy/load-balancer detections will probably "catch" any proxy forwarded SSL headers, `X-Forwarded-Port` is one additional low-priority case, now added to the list.


zcwilt:
:+1:

ajeh:
:+1:

PR #307, opened 2015-02-25
Fix issue 144 Admin Title Tags.

scottcwilson:
Adds admin page name to title for admin pages.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -2668,12 +2668,11 @@ function zen_copy_products_attributes($products_id_from, $products_id_to) {
 
   function zen_get_configuration_group_value($lookup) {
     global $db;
-    $configuration_query= $db->Execute("select configuration_group_title from " . TABLE_CONFIGURATION_GROUP . " where configuration_group_id ='" . zen_db_input($lookup) . "'");
-    $lookup_value= $configuration_query->fields['configuration_group_title'];
+    $configuration_query= $db->Execute("select configuration_group_title from " . TABLE_CONFIGURATION_GROUP . " where configuration_group_id ='" . (int)$lookup . "'");
     if ( $configuration_query->RecordCount() == 0 ) {
-      $lookup_value='<span class="lookupAttention">' . $lookup . '</span>';
+      return (int)$lookup; 
     }
-    return $lookup_value;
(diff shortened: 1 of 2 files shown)


admin/includes/functions/general.php
```diff
@@ -2666,6 +2666,17 @@ function zen_get_configuration_key_value($lookup) {
     return $lookup_value;
   }
 
+  function zen_get_configuration_group_value($lookup) {
+    global $db;
+    $configuration_query= $db->Execute("select configuration_group_title from " . TABLE_CONFIGURATION_GROUP . " where configuration_group_id ='" . zen_db_input($lookup) . "'");
+    $lookup_value= $configuration_query->fields['configuration_group_title'];
+    if ( $configuration_query->RecordCount() == 0 ) {
+      $lookup_value='<span class="lookupAttention">' . $lookup . '</span>';
```
drbyte: Why the insertion of HTML here, when the only use of this function (presently) is to return a string that gets output as part of a page `<title>` and thus wouldn't want nested `<span>`s in it?

scottcwilson: It's an error case.  It's taken from the routine just above it.  It can be deleted if you wish; it's not critical to this change.

drbyte: I think it's better to just return `$lookup` in this case.

But I'd rewrite the function like this, using an early-return:

```php
  function zen_get_configuration_group_value($lookup) {
    global $db;
    $configuration_query= $db->Execute("select configuration_group_title from " . TABLE_CONFIGURATION_GROUP . " where configuration_group_id ='" . (int)$lookup . "'");
    if ($configuration_query->RecordCount() == 0) {
      return (int)$lookup;
    }
    return $configuration_query->fields['configuration_group_title'];
  }
```

scottcwilson: No problem - will do.

scottcwilson: All set.

drbyte:
:+1: 

I'm thinking of splitting out this calculated TITLE tag for the admin side into another spot (ie: not introducing more "logic" into a language file). But merging this will be a good springboard for that.

ajeh:
:+1:

scottcwilson:
Great!

Issue #144, opened 2014-09-01
Admin TITLE tags


bislewl:
I think it would be incredibly helpful to store owners to have the name of the of the page in the title tag, so they can go back quickly.

Examples might be:
Configuration - Stock
Shipping - USPS
Define Pages - Main Page
EZ Pages - Shipping

scottcwilson:
PR submitted.

PR #308, opened 2015-02-27
Refinements to #144 Admin Title Tags which build upon #307

drbyte:
Removes the Admin `TITLE` logic from the catalog language file, into admin init_templates script, so the catalog-side file is now catalog-only. (Note: the admin still reads the file to allow for some display elements to offer "preview" capability from within the admin)

Uses new constant `ADMIN_TITLE` as distinct from the catalog-side `TITLE`

Builds the name from the `cmd` value, as per submission by @scottcwilson in #307

Catches missing names to allow fallback detection for plugins that might not get upgraded properly

Minor whitespace cleanup to the catalog meta_tags.php language file.

### DIFF
--- a/admin/admin_activity.php
+++ b/admin/admin_activity.php
@@ -3,10 +3,10 @@
  * Admin Activity Log Viewer/Archiver
  *
  * @package admin
- * @copyright Copyright 2003-2014 Zen Cart Development Team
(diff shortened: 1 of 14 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #316, opened 2015-03-03
$products_id is set twice

Zen4All:
$products_id is set on line 34 and 58

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/316)

### DIFF
--- a/admin/specials.php
+++ b/admin/specials.php
@@ -55,7 +55,6 @@
         $specials_date_available = ((zen_db_prepare_input($_POST['start']) == '') ? '0001-01-01' : zen_format_date_raw($_POST['start']));
         $expires_date = ((zen_db_prepare_input($_POST['end']) == '') ? '0001-01-01' : zen_format_date_raw($_POST['end']));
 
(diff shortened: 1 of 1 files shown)


admin/specials.php
```diff
@@ -55,7 +55,6 @@
         $specials_date_available = ((zen_db_prepare_input($_POST['start']) == '') ? '0001-01-01' : zen_format_date_raw($_POST['start']));
         $expires_date = ((zen_db_prepare_input($_POST['end']) == '') ? '0001-01-01' : zen_format_date_raw($_POST['end']));
 
-        $products_id = zen_db_prepare_input($_POST['products_id']);
```
drbyte: Agreed. This was a duplicate of line 34.

drbyte:
:+1:

ajeh:
:+1:

ajeh:
As a heads up, when changing files it is also a good idea to fix the header copyright to the current year.

And if the headers are the old style, they should also be brought up to date.

Zen4All:
:+1:

PR #317, opened 2015-03-03
Fixes issue with mismatched signatures for calculate_deductions.

scottcwilson:
See http://www.zen-cart.com/showthread.php?216420-Error-in-ot_coupon-php

### DIFF
--- a/includes/modules/order_total/ot_coupon.php
+++ b/includes/modules/order_total/ot_coupon.php
@@ -110,11 +110,10 @@ class ot_coupon {
    */
   function pre_confirmation_check($order_total) {
     global $order;
+    $od_amount = array('tax'=>0, 'total'=>0);
     if ($order_total > 0) { 
        $od_amount = $this->calculate_deductions();
-    } else { 
-       $od_amount = array('tax'=>0, 'total'=>0);
-    }
+    } 
 //    print_r($od_amount);
     $order->info['total'] = $order->info['total'] - $od_amount['total'];
     if (DISPLAY_PRICE_WITH_TAX != 'true') {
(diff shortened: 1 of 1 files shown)


includes/modules/order_total/ot_coupon.php
```diff
@@ -57,7 +57,11 @@ function ot_coupon() {
   function process() {
     global $order, $currencies;
     $order_total = $this->get_order_total(isset($_SESSION['cc_id']) ? $_SESSION['cc_id'] : '');
-    $od_amount = $this->calculate_deductions();
+    if ($order_total > 0) { 
+       $od_amount = $this->calculate_deductions();
+    } else { 
+       $od_amount = array('tax'=>0, 'total'=>0);
+    }
```
drbyte: Rather than use an `else` at all, what do you think of first initializing `$od_amount` to the zero-array before the `if`?

ajeh:
As a heads up, when changing files it is also a good idea to fix the header copyright to the current year.

And if the headers are the old style, they should also be brought up to date.

scottcwilson:
Both changes have been made.  Thanks!

includes/modules/order_total/ot_coupon.php
drbyte: Can we do the same here too?

:smirk:

scottcwilson:
Chris requested an additional change so that the pre-creation of the $od_amount array was done in both places before the calculate_deductions call.  This has been done.

zcwilt:
:+1:

ajeh:
:+1:

PR #319, opened 2015-03-03
Fixing html errors

Zen4All:
The html in this file is just plain wrong.

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/319)


zcwilt:
:+1:

drbyte:
:+1: 
Normally it would be preferable to not use inline styling, but we have other styling to do in the Admin, and that will be taken care of at that stage.

ajeh:
:+1:

PR #326, opened 2015-03-04
Skip blank keyword/description metatags

drbyte:
Empty description or keywords serves no purpose. Skipping if blank.

Similar with title, but I'll argue that title should never be blank, and will rely on core code to always ensure it is set.


ajeh:
:+1:

zcwilt:
:+1:

PR #329, opened 2015-03-04
Continuation of #275 - int vs string for MySQL strict mode

drbyte:
Applying same changes as #275 to sibling modules

### DIFF
--- a/includes/modules/payment/authorizenet.php
+++ b/includes/modules/payment/authorizenet.php
@@ -618,7 +618,7 @@ class authorizenet extends base {
       $sql = $db->bindVars($sql, ':respCode', $response['x_response_code'], 'integer');
       $sql = $db->bindVars($sql, ':respText', $db_response_text, 'string');
       $sql = $db->bindVars($sql, ':authType', $response['x_type'], 'string');
-      $sql = $db->bindVars($sql, ':transID', $response['x_trans_id'], 'string');
+      $sql = $db->bindVars($sql, ':transID', $response['x_trans_id'], 'integer');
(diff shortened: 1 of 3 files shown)


lhungil:
Does this break being able to designate the Transaction ID as NULL? For example when logging an error condition?

drbyte:
It's actually fixing the situation where Transaction ID is not returned. It's setting it to 0 in that case.

drbyte:
It could be rewritten differently, perhaps like:

```php
      if (trim($response['x_trans_id']) != '') {
        $sql = $db->bindVars($sql, ':transID', $response['x_trans_id'], 'string');
      } else {
        $sql = $db->bindVars($sql, ':transID', 'NULL', 'passthru');
      }
```

lhungil:
I'm not sure which way would be better in the long term. I think NULL is more obvious at the database level, but using 0 would work just as well as long as we know the third party will never return 0 as a transaction id (suspect this is a safe bet).

If sticking with the code forcing to an integer, I would suggest altering the table to no longer allow NULL (save some database storage space and make it obvious from the schema the field will always contain an integer).

Otherwise I'd go with a check similiar to what you mentioned instead of forcing to an integer.

drbyte:
@lhungil  said
> using 0 would work just as well as long as we know the third party will never return 0 as a transaction id (suspect this is a safe bet).

Correct. 0 is safe.

But I'm inclined to use NULL, the more I think about it.

lhungil:
:+1:

ajeh:
:+1:

PR #336, opened 2015-03-07
Correctly check admin module 'key' counts

scottcwilson:
Fixes admin/modules.php so that module key counts are correctly checked.  This check was skipped because the if statement was incorrect.

### DIFF
--- a/admin/modules.php
+++ b/admin/modules.php
@@ -181,32 +181,31 @@ require('includes/admin_html_head.php');
       include($module_directory . $file);
       $class = substr($file, 0, strrpos($file, '.'));
       if (class_exists($class)) {
-          $module = new $class;
-          $check = $module->check();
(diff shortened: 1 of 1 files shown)


scottcwilson:
Testing procedure: correctly configure payment module Money Order; observe no errors.  In mysql, run

delete from configuration where configuration_key = 'MODULE_PAYMENT_MONEYORDER_SORT_ORDER';

Refresh admin->modules->payment; still no error is observed.  Fix file admin/modules.php; see error message.  Fix configuration; observe no errors.

scottcwilson:
Updated at Linda's recommendation.  Additional tests:
- test when Money Order installed by set to false.
- test after adding an additional key to money order (upgrade scenario).

drbyte:
:+1:

ajeh:
:+1:

PR #337, opened 2015-03-08
Update Zen Cart logo

DivaVocals:
Minor change as follows:
~ Replace outdated Zen Cart logo with the current logo.
~ Remove logo.gif from template_default and classic folders
~ Update header.php to reflect the correct logo file


DivaVocals:
Sorry.. pushed the wrong thing.. Will try again in a second..

PR #345, opened 2015-03-11
Adds error checking to Min Values->Admin Usernames

scottcwilson:
Currently changing this value to something less than 4 is permitted but ignored.  This fix rejects the change.

If people like this I can extend it to do other things such as verify that min values/max values are numeric, etc.


drbyte:
Testing based on gID or cID is unreliable. 

I think in the long term a better way to handle validation for this sort of thing would be to declare a validation type in a column for that record in the db, and then apply the corresponding test.  Akin to the set_function and use_function stuff, but newer and improved. (The set_function and use_function stuff is dated and needs an overhaul.)

scottcwilson:
How could it be unreliable? Those are the parms that are used in the SQL
update statement.

drbyte:
```php
if ($gID == 2 && $cID == 59)
```
You've hard-coded specific integers for gID and cID. Those are assigned by auto-increment in the db, and are not the canonical master.  `configuration_key` would be the canonical master for cID. And for groups it's the non-multilingual-friendly `configuration_group_title`.

scottcwilson:
Would you accept the PR if I change gID/CID to configuration_group_title /
configuration_key or should I abandon it?

scottcwilson:
Alternately I could change the Admin User Creation facility to respect the
value the user set and permit the creation of admin usernames like "foo" if
the min length value is 3.

drbyte:
Does it currently "silently reject" creation of short usernames?
I don't have the code in front of me right now, but I thought we at least threw a messageStack alert that the name is too short, in the admin user-create dialog page.

Ultra-short admin login usernames is just a bad idea, regardless of PCI needs or not. Hence the internal enforced 4-char minimum. 4 is hardly enough, but for those who use ZC as a CMS and not for ecommerce, it feels like a relatively happy medium.

scottcwilson:
It does reject (not silently) a 3 character username, but this is confusing
since you are allowed to set the min length to 3.

scottcwilson:
To test on an existing database, do: 

ALTER TABLE configuration ADD val_function text default NULL AFTER set_function;
UPDATE configuration SET val_function='FILTER_VALIDATE_INT, array("options" => array("min_range" => 4))' WHERE configuration_key = 'ADMIN_NAME_MINIMUM_LENGTH';

scottcwilson:
If you like this but don't like the default error msg, I could add a text string as the first parm; just LMK.

admin/includes/configuration_checks.php
```diff
@@ -0,0 +1,17 @@
...
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version 
+ */
+
+  // Function used for configuration checks only
+  function check_configuration($variable, $check_string) { 
+     $parts = explode(",", $check_string, 2); 
+     eval('$id = ' . $parts[0] . ';'); 
+     eval('$options = ' . $parts[1] . ';');
```
drbyte: Is it at all possible to avoid using `eval()` here?

I know there are some web apps that use eval() generously, but it's also a potential sec urity hole target, especially when its inputs are coming directly from the db.  I know ZC uses it in a couple places, but we try hard to limit those where possible.

I wonder if the `$id` one can be read as 
```php
$id=constant($parts[0]);
```
And then check whether its value is a sane value before passing it along to `filter_var`.

Not sure about `$parts[1]` though. Running it through a sanitizing regex first might be doable, but there are so many characters that are potentially acceptable that it might be hard to come up with a blacklist of mal icious characters to exclude.

admin/configuration.php
```diff
@@ -21,7 +21,17 @@
...
         $configuration_value = zen_db_prepare_input($_POST['configuration_value']);
+        // See if there are any configuration checks 
+        $checks = $db->Execute("SELECT val_function FROM " . TABLE_CONFIGURATION . " WHERE configuration_id = '" . (int)$cID . "'");
+       if (!$checks->EOF) { 
+           require_once("includes/configuration_checks.php"); 
+           if (check_configuration($configuration_value, $checks->fields['val_function']) === false) {
+                $messageStack->add_session(TEXT_ADMIN_VALUE_OUT_OF_RANGE, 'caution');
+                zen_redirect(zen_href_link(FILENAME_CONFIGURATION, 'gID=' . $_GET['gID'] . '&cID=' . (int)$cID));
+              } 
+       }
```
drbyte: This is gonna sound nit-picky, but your indentation is all off, and doesn't match the section of code you're inserting this into.

scottcwilson:
OK please note change in config value: 

UPDATE configuration SET val_function='TEXT_MIN_ADMIN_USER_LENGTH, FILTER_VALIDATE_INT, array("options" => array("min_range" => 4))' WHERE configuration_key = 'ADMIN_NAME_MINIMUM_LENGTH';

scottcwilson:
Config value for eval-free code: 
UPDATE configuration SET val_function='{"error":"TEXT_MIN_ADMIN_USER_LENGTH","id":257,"options":{"options":{"min_range":4}}}' WHERE configuration_key = 'ADMIN_NAME_MINIMUM_LENGTH';

admin/includes/configuration_checks.php
```diff
@@ -0,0 +1,28 @@
...
+  function check_configuration($variable, $check_string) { 
+     global $messageStack; 
+     $data = json_decode($check_string, true); 
+     if (!empty($data['error']) && defined($data['error'])) { 
+        $error_msg = constant($data['error']); 
+     } else {
+        $error_msg = 'Validation error'; 
+     }
+     $id = $data['id']; 
+     $options = $data['options'];
```
drbyte: Can we do a little more input validation here? 

Currently you're expecting `$id` to be an integer, but people who try to extend this further are going to be a bit confused by having an integer value. I might suggest testing whether `$data['id']` is an integer. If so, carry on. If not, test whether the passed string is defined as a constant (like you do for `$data['error']`) and use that. If not defined and not integer, may as well `return` because filter_var is going to do nothing useful if an invalid `$id` is passed to it. 

And then what about perhaps doing some very basic validation on `$data['options']`? Admittedly that's quite a lot more involved, and I'm not sure it's worth a lot of effort beyond testing whether it contains an `options` element which has an array in it.  Beyond that it's probably best to let error-trapping or exception-handling deal with it.

drbyte:
This is definitely looking better so far. :)

scottcwilson:
To clarify: The id parameter in filter_var is an integer - these are not defined strings.   See 
http://php.net/manual/en/function.filter-var.php
But I can certainly add more documentation and error checking, since admittedly this is hairy. :)

drbyte:
I think you missed my point.

The id parameter in filter_var certainly needs to be passed an integer value. But the same filter_var page you reference mentions the defined constants that constitute acceptable integer values to pass.

So, this could be included too:

```php
if (defined($data['id'])) {
  $id = constant($data['id']);
}
```

That would allow the JSON in your json array to provide `FILTER_VALIDATE_INT` instead of having to go figure out the integer value of that constant to put into the array instead.

You could toy with whether to require it to be stored in the json as `FILTER_VALIDATE_INT` or `'FILTER_VALIDATE_INT'`(quoted), to see what's easiest.

admin/includes/configuration_checks.php
```diff
@@ -0,0 +1,45 @@
...
+ * @version 
+ */
+
+  // Function used for configuration checks only
+  function check_configuration($variable, $check_string) { 
+     // Expected format of $check string is a json encoded array 
+     // with the following paramters: 
+     // error: defined constant with error message
+     // id: id of the filter to apply.   
+     // options: per http://php.net/manual/en/function.filter-var.php
```
drbyte: The preferred modern-day approach to putting documentational comments in code is to use DocBlock syntax, above the function, instead of inside it.

scottcwilson:
Apparently I am extra thick today.  I am looking at http://php.net/manual/en/function.filter-var.php and not seeing a mention of defined constants, and in the function definition, it shows $filter as an int.  Can you show me a line of code that works with a defined constant so I can test with it?

admin/includes/configuration_checks.php
scottcwilson: fixed - will be in next checkin.

zcwilt:
I think what Chris is getting at is this,

In your sql to define the check, you do something like

"id":257

within the json definition. 

While perfectly valid, it would easier for other devs if they could use the defined constants here, 
so 
"id":257
would become 
"id":'FILTER_VALIDATE_INT'

zcwilt:
BTW, like the idea behind this. so don't get disheartened by the (hopefully helpful) critiques. :)

scottcwilson:
I had *completely* forgotten I used 257 - I thought I used the mnemonic name.  You are right - this is hard to read.  OK will fix later today.

scottcwilson:
Test with mnemonic value: (PR coming shortly)

UPDATE configuration SET val_function='{"error":"TEXT_MIN_ADMIN_USER_LENGTH","id":'FILTER_VALIDATE_INT',"options":{"options":{"min_range":4}}}' WHERE configuration_key = 'ADMIN_NAME_MINIMUM_LENGTH';

scottcwilson:
LOL reading comprehension FTW!

admin/includes/configuration_checks.php
```diff
@@ -0,0 +1,50 @@
...
+        $error_msg = 'Validation error - bad options field'; 
+        return; 
+     }
+     $result = filter_var($variable, $id, $options); 
+     if ($result === false) { 
+        $messageStack->add_session($error_msg, 'error');
+        zen_redirect(zen_href_link(FILENAME_CONFIGURATION, 'gID=' . $_GET['gID'] . '&cID=' . (int)$cID));
+     }
+     return; 
+  }
```
drbyte: To collapse the `if/else` nesting, I propose the following, which uses "guard statements" to do early-return using the `if` and then simply carrying on with the var assignments when the individual tests pass. 

This also drops the setting of $error_msg where it's never actually used.

```php
function check_configuration($variable, $check_string) { 
     global $messageStack; 
     $data = json_decode($check_string, true); 

     // check inputs - error message should be a defined constant in the language files
     if (empty($data['error']) || !defined($data['error'])) return;
     $error_msg = constant($data['error']); 

     if (defined($data['id'])) { 
        $id = constant($data['id']); 
     } else if (is_integer($data['id'])) { 
        $id = $data['id']; 
     } else { 
        return; 
     }

     // example: $options = array('options' => array('min_range' => 4));
     if (!is_array($data['options'])) return;
     $options = $data['options']; 

     $result = filter_var($variable, $id, $options); 
     if ($result === false) { 
        $messageStack->add_session($error_msg, 'error');
        zen_redirect(zen_href_link(FILENAME_CONFIGURATION, 'gID=' . $_GET['gID'] . '&cID=' . (int)$cID));
     }
  }
```

drbyte:
> @zcwilt said:
>BTW, like the idea behind this. so don't get disheartened by the (hopefully helpful) critiques. :)

Agreed.

admin/configuration.php
```diff
@@ -21,7 +21,14 @@
           $messageStack->add_session(ERROR_ADMIN_DEMO, 'caution');
           zen_redirect(zen_href_link(FILENAME_CONFIGURATION, 'gID=' . $_GET['gID'] . '&cID=' . (int)$cID));
         }
+        
         $configuration_value = zen_db_prepare_input($_POST['configuration_value']);
+        // See if there are any configuration checks 
+        $checks = $db->Execute("SELECT val_function FROM " . TABLE_CONFIGURATION . " WHERE configuration_id = '" . (int)$cID . "'");
+       if (!$checks->EOF && $checks->fields['val_function'] != NULL) { 
+           require_once("includes/configuration_checks.php");
```
drbyte: Can we move the file into the `admin/includes/functions/` folder instead of just `admin/includes`?

And while in this section, touch the indenting of your `if` statement and the closing `}` to add one more space so they align with the lines above and below this section.

zcwilt:
:+1:

drbyte:
:+1:

PR #327, opened 2015-03-04
Refactor file-download code to allow AWS etc

drbyte:
- cleaned up the My Account list of downloads, for simpler styling
- changed both admin and catalog to allow observers to provide details of file availability
- added support for AWS-hosted downloads (files on AWS S3) ... simply give `aws:bucketname/filename.ext` as the filename in attributes controller, with expiring links to prevent theft
- added support for HTTP-hosted downloads, such as Dropbox or any other http/https URL which requires no authentication by the customer (offers no theft prevention)
- moved download-by-redirect and download-by-streaming to observer class
- removed COWOA pages for downloads, since tidied template no longer needs them

PR #347, opened 2015-03-12
Fix size/unknown indicators on downloads page

drbyte:
Ajeh spotted that the display of MB/bytes on the downloads page had some problems after the commits in #327


ajeh:
:+1:

ajeh:
:+1:

zcwilt:
:+1:

Issue #340, opened 2015-03-08
Question on work item "Template - multi-tier ... parent, child, grandchild ?"


scottcwilson:
Would it be satisfactory to add *one* more level to the template lookup: 
- check template defined in admin->tools->template selection
- (new) check backup template (perhaps defined in same place, or perhaps statically named).
- use default file (template default or base directory for a languages file).

Spike00:
I used exactly this solution to manage mobile template

scottcwilson:
Thank you @Spike00 - did you use a variable name (like the current template) or did you use a fixed name?  I am tending towards a fixed name (perhaps "common"). I think this would work nicely for the Defines Page Editor (for example), where the order of search could be: 

common -> current template -> base directory (includes/languages/CURRENT_LANGUAGE/html_includes.) 

Anyone else?

Spike00:
I wrote this as if I would release it as a plugin some day, so I thought of possible conflicts with template names and decided to use a define (so a variable name) defined in a file in extra_configures, to easy fix possible conflicts.
Introducing this feature at core distribution, would not require to have the name variable, but I don't really see the reason not to.

define('ZW_MOBILE_DIR', 'mobile');

Then 2 examples of functions regarding module and template (I didn't check if the code is the same for 1.5.4,  I made this thing on a super/ultra/hyper modified version of (an ancient) Zen)

  function zen_get_module_directory($check_file, $dir_only = 'false') {
	global $template_dir;
	$zv_filename = $check_file;
	if (!strstr($zv_filename, '.php')) $zv_filename .= '.php';
// ZW MOBILE - Start
	if (file_exists(DIR_WS_MODULES . $template_dir . '/' . ZW_MOBILE_DIR . '/' . $zv_filename)) {
	  $template_dir_select = $template_dir . '/' . ZW_MOBILE_DIR . '/';
	} else {
// ZW MOBILE - End
// Would be better to use an elseif, but this way I may simply add code without touching the existing core code
		if (file_exists(DIR_WS_MODULES . $template_dir . '/' . $zv_filename)) {
		  $template_dir_select = $template_dir . '/';
		} else {
		  $template_dir_select = '';
		}
// ZW MOBILE - Start
	}
// ZW MOBILE - End
	if ($dir_only == 'true') {
	  return $template_dir_select;
	} else {
	  return $template_dir_select . $zv_filename;
	}
  }

[...]

PR #350, opened 2015-03-18
Implement intermediate "shared" template tier

scottcwilson:
Provides for "shared" directory which is searched after your_template but before template_default.

Tested with the following files: 
includes/languages/english/html_includes/shared/define_main_page.php
includes/languages/english/shared/index.php
includes/languages/shared/english.php
includes/languages/english/shared/header.php
includes/templates/shared/common/tpl_header.php
includes/templates/shared/templates/tpl_login_default.php
- also tested with editing the first file in the admin defines pages editor.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -3135,6 +3135,8 @@ function zen_copy_products_attributes($products_id_from, $products_id_to) {
 
     if (file_exists($check_directory . $template_dir . '/' . $zv_filename)) {
       $zv_directory = $check_directory . $template_dir . '/';
+    } else if (file_exists($check_directory . 'shared' . '/' . $zv_filename)) {
+      $zv_directory = $check_directory . 'shared' . '/';
     } else {
       $zv_directory = $check_directory;
     }
--- a/includes/classes/template_func.php
+++ b/includes/classes/template_func.php
@@ -1,11 +1,12 @@
 <?php
 /**
- * template_func Class.
+ * template_func class.
  *
- * @package classes
- * @copyright Copyright 2003-2014 Zen Cart Development Team
+ * @package initSystem
+ * @copyright Copyright 2003-2015 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: template_func.php 14141 2009-08-10 19:34:47Z wilt $
+ * @version GIT: $Id:
  */
 if (!defined('IS_ADMIN_FLAG')) {
   die('Illegal Access');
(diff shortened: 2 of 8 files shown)


scottcwilson:
The order of inclusion in includes/languages/english.php is shared first, then template or default.  I just realized this.  I want to get some feedback before I make more changes though since people might prefer this order.

drbyte:
Okay :+1:

zcwilt:
:+1:

ajeh:
:+1:

drbyte:
Closes #340

Zen4All:
When testing this brilliant new feature, I stumbled across a possible bug.
I created a new css folder in the shared template folder and a new (empty) css file called stylesheet_test.css. After this I reloaded my test zc 1.6 site and all the formatting is gone, because zc now only looks in the shared/css folder for all the style sheets. The same is happening when putting a single image in the shared/images/icons directory

scottcwilson:
Do people agree that the comment above by @Zen4All is in fact an issue?  Putting a blank stylesheet in your current template would have the same effect.

drbyte:
The behavior until v1.5.x has always been that any stylesheet or javascript file being present in the override folder would cause the override folder to be the only folder used for css or js.  For images it would still cascade up to template_default.

Zen4All:
May be I misunderstood the function of the "shared" folder, but let me explain what I thought I could do with it.
I have build a new multi store module, but I think it would also work on the current multi site module.
And have ported the shared folder functionality to Zen Cart 1.5.4
Each store has its own unique template folder so the structure would look like this:
```
includes
 |- templates
   |- classic
   |- shared
   |- store1
   |- store2
   |- store3
   |- store4
   |- template_default
```
To reduce the number of files I wanted to put all "shared" files between the different store templates in the shared folder. Like said in the other post this works for the template folder, but not the css, jscript, and images.

drbyte:
Understood. 

What you described should work fine for images (er, well, it always has before the `shared` folder was introduced).

But ... 

Question: do you want it to automatically load *ALL* the stylesheets for `store1` AND `shared` AND `template_default`? Wouldn't that potentially result in some undesired results?  And wouldn't that therefore cause extreme temptation to alter the `template_default` files?

Same with javascript files.

Zen4All:
I have been rethinking this problem, and it came to the conclusion the **functionality should stay as it is** right now. Sharing Image/css/jscript folders will create a whole bunch of new problems, which is not worth the effort.

scottcwilson:
Should this issue be closed?

drbyte:
@Zen4All thanks for posting back after having researched further. 

I'm glad that you didn't just "accept" the offered response, but instead took the time to challenge it, even though you decided it should stay as-is.  I don't mind being wrong ... in fact I'd rather know that there's a problem or "use case" that hasn't been considered yet, so that we can collectively come up with an even better solution. That's the value of community and open source!

Thanks again for contributing!

PR #358, opened 2015-03-23
Fix typo in method call

drbyte:
Fix typo in method call, as identified by @lat9 in http://www.zen-cart.com/showthread.php?208310-paypaldp-php-Invalid-class-variable


zcwilt:
:+1:

ajeh:
:+1:

PR #360, opened 2015-03-24
Replace phpBB with hooks for external "forum" systems - Closes #272

### LINKED ISSUE #272: Add hooks for external "forum" systems
- add hooks to make @lat9's BB plugin to only use notifier hooks (instead of touching any core files)
- drop built-in phpBB integration, in favor of community-managed BB integrations (phpBB, Discourse, etc)

drbyte:
Notifier hooks have been added so that all former phpBB integration can be handled with custom observer class methods.

Integration note: there are a couple places where global variables like `$external_bb_text` are referenced, but no specific notifier hook is called to set them. This is because the code that 'uses' those variables is already in global scope, and therefore from within your observer class you can simply declare those variables as `global`s, and assign them a value there. Since this is still procedural code, in an observer class it is probably appropriate to do this in `__construct`.

Most other variables are set by simply letting observer class functions set the desired values and pass-by-reference (ie: params 2 - 9 for observer `update` triggers)

ALSO NOTE:

> An advanced observer class feature is to create several `update` functions, named according to the specific notifier hook you want them to respond to. So, responding to NOTIFY_NICK_SET_TEMPLATE_FLAG could be done with its own function:

```php
function updateNotifyNickSetTemplateFlag(&$class, $eventID, $none, $display_nick_field) {
  if ($this->myBbSystemIsEnabled) $display_nick_field = true;
}
```

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -37,7 +37,6 @@ filter:
         - 'includes/classes/class.phpmailer.php'
         - 'includes/classes/class.smtp.php'
         - 'includes/classes/http_client.php'
-        - 'includes/classes/class.phpbb.php'
         - 'includes/library/aura/*'
         - 'includes/modules/debug_blocks/*'
         - 'includes/modules/payment/linkpoint_api/*'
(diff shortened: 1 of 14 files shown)


zcwilt:
:+1:

ajeh:
:+1:

DivaVocals:
:+1: AWESOME!!!!

PR #363, opened 2015-03-24
Add TOPMOST_CATEGORY_PARENT_ID constant to denote which top-most category-id is parent to everyone

drbyte:
Add TOPMOST_CATEGORY_PARENT_ID constant to denote which top-most category-id is parent to everyone

Knowing what the top-most parent category is can be helpful in many areas.
Presently we assume all top-level cats are children of "0" (inherited from legacy code).
If that were to change (such as MPTT or custom uses of ZC code), the constant is a better reference point than having to change the "0" in all of these places all throughout the code.

This can aid in making multi-site plugins simpler to integrate.


ajeh:
:+1:

zcwilt:
:+1:

PR #365, opened 2015-03-24
Allow single-attribute products to do add-to-cart from prod-listing without going to product page

drbyte:
This implements code I posted awhile back for allowing products with a single attribute to be directly added-to-cart from the product listing. This is especially useful for Download products.

Ref: http://www.zen-cart.com/showthread.php?142161-HOW-CAN-I-AUTO-SELECT-SINGLE-ATTRIBUTE-CHOICES-DURING-ADD-TO-CART-process&p=809142#post809142


ajeh:
This does not take into account Products with 1 Attribute that is an upload file. This cannot be just added to the cart. 
Sometimes it is a required upload file and sometimes optional. There isn't a setting to flag this.

ajeh:
This does not take into account Products with 1 Attribute that is Text. This cannot be just added to the cart. 
Sometimes it is a required Text and sometimes optional. There is a setting to flag this. But it is a choice the user needs to make.

zcwilt:
:+1:

ajeh:
:+1: 
After a bazillion tests and glasses of wine, I have not been able to break this. I will probably break it 5 minutes after it is merged.

zcwilt:
:+1:

PR #367, opened 2015-03-24
Strip out "number_of_uploads" param when done with it

drbyte:
In places such as add-to-cart, an extra GET param `number_of_uploads` gets left on the URI, but serves no purpose since it's used *before* getting to the shopping-cart page.


ajeh:
:+1:

zcwilt:
:+1:

PR #368, opened 2015-03-24
Fix SQL errors from #345

drbyte:
In #345 some SQL errors were introduced mistakenly.


ajeh:
:+1: 
This appears to fix the issues correctly.

zcwilt:
:+1:

PR #369, opened 2015-03-24
Ampersands were being double-encoded in mfg pulldowns

drbyte:
ref: http://www.zen-cart.com/showthread.php?204962-Manufacters-names-in-dropdown-menus-being-html-converted and @lat9


ajeh:
:+1: 
This appears to correct the errors with &amp; or & in the names.

zcwilt:
:+1:

PR #370, opened 2015-03-24
Disable code-coverage reporting in TravisCI (in favor of Scrutinizer)

drbyte:
This will speed up tests considerably, and Scrutinizer can run the coverage reports separately.


ajeh:
:+1:

zcwilt:
:+1:

PR #382, opened 2015-04-07
Refactor language class and make admin share with catalog

drbyte:
Refactor language class and make admin share with catalog

- Remove admin language class, since it's identical to catalog class
- Refactor language class to use new QueryFactory iterator
- Refactor language class to declare visibility of vars and methods


zcwilt:
:+1:

ajeh:
:+1:

PR #388, opened 2015-04-14
Fix bug introduced in #382

drbyte:


ajeh:
:+1: 
This appears to fix the issues.

zcwilt:
:+1:

PR #389, opened 2015-04-15
Why query all fields when only products_quantity is needed

Zen4All:
Is it necessary to call for all fields when only 'products_quantity' is used?

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/389)

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -770,7 +770,7 @@ class order extends base {
           }
           $stock_values = $db->Execute($stock_query_raw, false, false, 0, true);
         } else {
-          $stock_values = $db->Execute("select * from " . TABLE_PRODUCTS . " where products_id = '" . zen_get_prid($this->products[$i]['id']) . "'", false, false, 0, true);
+          $stock_values = $db->Execute("select products_quantity from " . TABLE_PRODUCTS . " where products_id = '" . zen_get_prid($this->products[$i]['id']) . "'", false, false, 0, true);
         }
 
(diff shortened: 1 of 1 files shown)


drbyte:
:+1: 

NOTE: There are some places where we do use `select *` or additionally name "apparently unused" fields in seemingly unnecessary places, but that's because those queries are similar enough that internal caching provides benefits that outweigh the seeming excess of asking for extra fields back.  The case you mention doesn't appear to be relevant to this situation. I'm simply mentioning it so you can keep that dynamic in mind if your eye is spotting other similar cases where you're questioning a particular query. 

Thanks for the catch!

includes/classes/order.php
```diff
@@ -770,7 +770,7 @@ function create_add_products($zf_insert_id, $zf_mode = FALSE) {
           }
           $stock_values = $db->Execute($stock_query_raw, false, false, 0, true);
         } else {
-          $stock_values = $db->Execute("select * from " . TABLE_PRODUCTS . " where products_id = '" . zen_get_prid($this->products[$i]['id']) . "'", false, false, 0, true);
+          $stock_values = $db->Execute("select products_quantity, product_is_always_free_shipping from " . TABLE_PRODUCTS . " where products_id = '" . zen_get_prid($this->products[$i]['id']) . "'", false, false, 0, true);
```
drbyte: Given that `'products_attributes_filename'` is also used on line 780, I think the query should also include `'' as products_attributes_filename`. But I suppose 780 could be rewritten too.

zcwilt:
:+1:

PR #390, opened 2015-04-15
Why query all fields when only products_attributes_id is needed

Zen4All:
Here only a simple check is done for the existence of a record. I do not think all fields are needed to do that

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/390)


ajeh:
:+1:

PR #391, opened 2015-04-15
Fix bug from v154 which incremented language_id incorrectly

drbyte:
Fix bug from v154 (CHANGE-709) which incremented language_id incorrectly

Ref: http://www.zen-cart.com/showthread.php?215842-Insertion-of-an-additional-language-does-not-copy-existing-category-product-texts&p=1269562#post1269562

### DIFF
--- a/admin/languages.php
+++ b/admin/languages.php
@@ -1,10 +1,10 @@
 <?php
 /**
  * @package admin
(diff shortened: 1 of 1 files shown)


ajeh:
:+1: 
This appears to fix the issues.

zcwilt:
:+1:

PR #394, opened 2015-04-16
Banners miss fire and set wrong date

ajeh:
Banners were not firing correctly for enable/disable and incorrect dates were being stored.

This affected the results in the Catalog and Admin.

### DIFF
--- a/admin/banner_manager.php
+++ b/admin/banner_manager.php
@@ -4,7 +4,7 @@
  * @copyright Copyright 2003-2015 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: banner_manager.php 19330 2011-08-07 06:32:56Z drbyte $
+ * @version GIT: $Id: banner_manager.php Author: ajeh  Modified in v1.6.0 $
  */
 
   require('includes/application_top.php');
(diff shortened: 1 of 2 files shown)


admin/banner_manager.php
```diff
@@ -130,7 +130,7 @@
 
           $sql = "UPDATE " . TABLE_BANNERS . "
                   SET
-                    date_scheduled = :scheduledDate,
+                    date_scheduled = DATE_ADD(:scheduledDate, INTERVAL '00:00:00' HOUR_SECOND),
```
drbyte: Doesn't this need something more, for when `$date_scheduled` is blank?

Something similar to line 140? `if ($date_scheduled != '') {`

zcwilt:
@drbyte good catch on the update. Will defer to @ajeh on that one :)

admin/banner_manager.php
ajeh: I do not believe so as around line 69 is:
        $expires_date = zen_db_prepare_input($_POST['expires_date']) == '' ? 'null' : zen_format_date_raw($_POST['expires_date']);
        $expires_impressions = zen_db_prepare_input($_POST['expires_impressions']);
        $date_scheduled = zen_db_prepare_input($_POST['date_scheduled']) == '' ? 'null' : zen_format_date_raw($_POST['date_scheduled']);

So far I cannot reproduce an issue. Are you seeing one?

drbyte: Then the `if` statements around those already-initialized vars on lines 137-145 are not needed, and should be removed, leaving just the `$sql=$db->bindVars()` calls because the `if`s have no purpose, because they'll always be true.

drbyte:
:+1: 
Appears to look good.

zcwilt:
:+1:

PR #402, opened 2015-04-30
Additional update to query from PR #389

drbyte:
This adds the products_attributes_filename field to the resultset for consistency, so that there's no chance the array element will not be defined and throw an error.


ajeh:
:+1:

PR #403, opened 2015-05-01
Add regex search to DTK (searches only individual lines, not across multiple lines)

drbyte:
Add regex search to DTK (searches only individual lines, not across multiple lines)
Also fixed context-highlight to show number of context lines in input box.
Also added "search all file types" to dropdown.


ajeh:
:+1:

zcwilt:
:+1:

PR #412, opened 2015-05-07
multiple-add-to-cart button was missing from "bottom"

drbyte:
The multiple-add-to-cart button was only showing on the "top" of the list, but not also the bottom when the option was set to `3=both`


ajeh:
:+1: 
Good catch!

zcwilt:
:+1:

PR #413, opened 2015-05-10
Handle multiple-language situations with single-attrib-add

drbyte:
PR #365 wasn't handling multi-lang scenarios


ajeh:
:+1: 
This appears to correctly fix issues with multiple languages and single attributes

zcwilt:
:+1:

PR #422, opened 2015-05-20
Fix paypal issue with store is set to allow a quantity of 1 or less

diego-vieira:
Just had an issue with a customer where the store can sell fabric which can have a quantity of 1 or less sometimes.

ie. 0.25, 0.75, etc

### DIFF
--- a/includes/modules/payment/paypaldp.php
+++ b/includes/modules/payment/paypaldp.php
@@ -1605,7 +1605,7 @@ class paypaldp extends base {
       } // endif attribute-info
 
       // PayPal can't handle fractional-quantity values, so convert it to qty 1 here
-      if ($order->products[$i]['qty'] > 1 && ($order->products[$i]['qty'] != (int)$order->products[$i]['qty'] || $flag_treat_as_partial)) {
+      if (is_float($order->products[$i]['qty'])) {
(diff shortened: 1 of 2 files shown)


drbyte:
Hi Diego,
Thanks for this!

There's a bit more to consider here. Using `is_float()` as the only test might be too limiting.  

For example, consider the following code:

```php
$foo = 3.3 - 1.3;
echo $foo;  // result = 2
echo is_float($foo) ? 'yes':'no'; // gives 'yes' even though the value is 2
```

In cases like this we don't want to do a translation of `2` to `1` of quantities in preparation for sending to paypal. Hence the check against `(int)`.

Also, the `$flag_treat_as_partial` is there for the sole purpose of letting storeowners set it to `true` (95 lines earlier) for _all_ transactions, in the event they always want the translation. Granted, I don't think anybody actually does that now. And your suggestion to use is_float might make the flag unnecessary altogether. 

Anyway, my suggestion is to not replace the entire condition with is_float, but instead to only replace the `$order->products[$i]['qty'] > 1` part with an is_float check, like this:

```php
if (is_float($order->products[$i]['qty']) && ($order->products[$i]['qty'] != (int)$order->products[$i]['qty'] || $flag_treat_as_partial)) {
```

Do you mind testing this and then updating the pull request? Feedback welcome too.

drbyte:
Just checking in to see whether this proposed alteration still solves the problem you were initially encountering?

PR #427, opened 2015-05-25
And Another movenext()

Zen4All:
updated to the new iterator

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/427)

PR #426, opened 2015-05-25
Another movenext()

Zen4All:
This time I tried the new iterator. Hope I got it right.

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/426)

PR #424, opened 2015-05-25
Typo, movenext() should be MoveNext().

Zen4All:
[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/424)

PR #430, opened 2015-05-27
Fix iterator syntax errors

drbyte:
Fix some syntax errors in #426 and #427 and #424 and implement the iterator fully in these small files


zcwilt:
:blush: I should have caught that on first review!
:+1:

ajeh:
:+1:

PR #434, opened 2015-06-02
Allow Discount Coupons Amount per Order or Qty

ajeh:
This allows for Discount Coupons based on Amount to be applied either per order or quantity * amount


drbyte:
Needs zc_install .sql update.

drbyte:
:+1:

zcwilt:
:+1:

PR #447, opened 2015-06-19
Fix salt errors on PHP 7

drbyte:
salt param is obsolete in PHP 7


drbyte:
Closing PR. Need to rework why these tests were written this way in the first place.  Punting to @zcwilt :)

zcwilt:
:+1:

ajeh:
:+1:

PR #453, opened 2015-06-26
Paypal introduction removed for orders over 10k, 5.5k

llynix:
While code is in place to hide the PayPal checkout button if orders are too large it doesn't hide the introduction text on the login page leading to a confusing paragraph about PayPal with no button to checkout with.

This fixes that.

### DIFF
--- a/includes/templates/template_default/templates/tpl_login_default.php
+++ b/includes/templates/template_default/templates/tpl_login_default.php
@@ -21,6 +21,10 @@
 <!--BOF PPEC split login- DO NOT REMOVE-->
 <fieldset class="floatingBox back">
 <legend><?php echo HEADING_NEW_CUSTOMER_SPLIT; ?></legend>
+
+<?php //  Paypal doesn't allow over 10k in a single transaction,  This hides the introduction. ?>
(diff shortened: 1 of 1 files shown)


drbyte:
:+1:

zcwilt:
:+1:

drbyte:
In retrospect, I wonder if it would be better to put these conditions into the header_php.php file instead? 

ie, added as the last line of this:

```php
// Check for PayPal express checkout button suitability:
$paypalec_enabled = (defined('MODULE_PAYMENT_PAYPALWPP_STATUS') && MODULE_PAYMENT_PAYPALWPP_STATUS == 'True' && defined('MODULE_PAYMENT_PAYPALWPP_ECS_BUTTON') && MODULE_PAYMENT_PAYPALWPP_ECS_BUTTON == 'On');
// Check for express checkout button suitability:
$ec_button_enabled = ($paypalec_enabled && ($_SESSION['cart']->count_contents() > 0 && $_SESSION['cart']->total > 0));
if ( ($_SESSION['currency'] == 'USD' && $currencies->value($_SESSION['cart']->total, true, 'USD') > 10000) || ($_SESSION['currency'] == 'GBP' && $currencies->value($_SESSION['cart']->total, true, 'GBP') > 5500) ) $ec_button_enabled = false;
```

llynix:
I had wondered after I submitted the pull request if there was a better
place to put it.  The header does seem more appropriate.


Anthony Taylor
---------------------
http://www.fallsgeek.com
(940)228-4580
Wichita Falls, TX
Your connection for everything geek...

PR #456, opened 2015-07-01
Addendum to PayPal intro-text-removal in #453

drbyte:
This removes the business logic from the template, meaning no template changes to get the same result as #453

### DIFF
--- a/includes/modules/payment/paypaldp.php
+++ b/includes/modules/payment/paypaldp.php
@@ -237,8 +237,8 @@ class paypaldp extends base {
 
     // Purchase amount
     if ($this->enabled) {
-      // module cannot be used for purchase > $10,000 USD
-      $order_amount = $this->calc_order_amount($order->info['total'], 'USD');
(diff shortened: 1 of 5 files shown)


includes/modules/payment/paypaldp.php
```diff
@@ -237,8 +237,8 @@ function update_status() {
 
     // Purchase amount
     if ($this->enabled) {
-      // module cannot be used for purchase > $10,000 USD
-      $order_amount = $this->calc_order_amount($order->info['total'], 'USD');
+      // module cannot be used for purchase > $10,000 USD equiv
+      $order_amount = $this->calc_order_amount($order->info['total'], 'USD', false);
```
drbyte: This `false` was added to ensure commas aren't inserted into this number, so that the comparison on the next line works correctly.

zcwilt:
:+1:

ajeh:
:+1:

PR #469, opened 2015-07-17
Copy Product to Duplicate Product sets Date Available to 01/01/0001

drbyte:
This change sets it to null instead of 01/01/0001


ajeh:
:+1:

zcwilt:
:+1:

PR #477, opened 2015-07-17
Remove double `default` clause (breaks PHP7)

drbyte:
This double `default:` is redundant (and actually incorrect), so is a bug. Plus, in PHP7 the duplicate will throw a fatal error.


zcwilt:
:+1:

ajeh:
:+1:

PR #478, opened 2015-07-18
Improvements for Banner Boxes (and fixes SQL error)

drbyte:
- Wraps banners in a `<section class="info-promowrapper">` so can be targeted with CSS styling
- Fixes SQL errors caused by having no active banners in a given group
- Splits logic out of templates more effectively, and simplifies some of the logic


zcwilt:
:+1:

ajeh:
:+1:

PR #479, opened 2015-07-18
Small tidying of tpl_columnar_display.php to output HTML directly

drbyte:
Small tidying of tpl_columnar_display.php to output HTML directly instead of echoed by PHP


zcwilt:
:+1:

ajeh:
:+1:

PR #481, opened 2015-07-18
zc_install - update from Foundation 5.5.1 to 5.5.2

drbyte:
Update Foundation to the latest.

NOTE: There may be some AJAX fixes required for compatibility with the update


zcwilt:
:+1:

ajeh:
:+1:

drbyte:
Hmmm ... need to investigate what's causing the Continue button not to fire. Feels like an ajax call is failing after updating to this new version. Tried 5.5.1 and it works okay, but not with this 5.5.2.

PR #486, opened 2015-07-22
add missing fontawesome-webfont.woff2 and allow in .htaccess

bislewl:


drbyte:
Could you do a quick test to confirm whether all those .htaccess files need to be touched. I suspect only the one in the templates folder needs touching, as there are no font files outside of that folder.

bislewl:
@drbyte you are correct

drbyte:
:+1:

zcwilt:
:+1:

PR #491, opened 2015-07-25
Remove commented lines ending with brackets ... because vi gets confused

scottcwilson:
Remove commented lines ending with brackets. Lines like this break bracket-matching in vi.  If they are genuinely needed, push the bracket to the next line so that the brackets still match.


zcwilt:
:+1:

ajeh:
:+1:

PR #494, opened 2015-07-27
fix issue of: Illegal string offset 'id' in classes/order.php

bislewl:
Fix issue of
PHP Warning:  Illegal string offset 'id' in /public_html/dev/includes/classes/order.php

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -484,7 +484,7 @@ class order extends base {
                                       'products_mixed_discount_quantity' => $products[$i]['products_mixed_discount_quantity']
                                       );
 
-      if (STORE_PRODUCT_TAX_BASIS == 'Shipping' && stristr($_SESSION['shipping']['id'], 'storepickup') == TRUE)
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks for the PR.  What specific steps will trigger the problem? (The last time I saw this it was caused by an outdated shipping module or other changes to core files.)

bislewl:
It seemed to have been created from when displaying the estimated shipping on shopping_cart and having the store tax set to the shipping zone. Using isset() seemed to be a simple solution and it resolved the issue. It's incredibly possible you are correct (and I suspect you are), would it cause any harm, or would it provide some backward functionality?

drbyte:
Not likely to cause any harm at all. Just wanting to be sure to understand the scope in case there's actually a broader compatibility fix that might be even more beneficial.

drbyte:
:+1:  I'm okay with this. 

However, a review of code suggests there are additional places where a similar test is warranted. ie: search for `$_SESSION['shipping']['id']` and there are various places where it references the id or cost elements without first testing for their existence or where they're not confirmed or set in preceding code.

This was an ugly architecture problem inherited from osC years ago, which didn't surface until strict array handling happened in PHP 5.4. Clearly there's still more that could be done to clean it up better.

zcwilt:
:+1:

ajeh:
:+1:

Issue #483, opened 2015-07-21
Add function for adding to TABLE_CONFIGURATION


bislewl:
Almost every single module I write, I have to make some entries to the TABLE_CONFIGURATION, It would be really nice if there was a function that allowed this to happen with out having to type out SQL queries all the time.

Agree? Disagree?

function add_config_value(){
if(!defined()){
$db->Execute("INSERT INTO ". TABLE_CONFIGURATION....
}

}

drbyte:
Hmmm ... I'd written some code for that a long time ago, and clearly it never got included. Right now I've no idea where I left it.

Probably also useful to have some helpers for configuration-groups too, as well as removing config values.

For the purpose of similar naming of functions, I might suggest they start with `zen_config_`, like `zen_config_add_value()`, `zen_config_add_group()`, etc

For the modules you're thinking of, are you speaking in terms of a payment/shipping module where one typically uses the class's `install()` method to put these values in? Or are you referring to other things?

bislewl:
Yes, that's the direction I was thinking. I'm referring to things like Ajax Back In Stock, Zen Light Box, Zencart Product Badges, Etc. Things that only fall into the Configuration Menu.  I guess this is more useful for auto-installers (which there could be a whole discussion on those pros/cons). When creating installers using this: https://www.zen-cart.com/downloads.php?do=file&id=1971 the files each need to go through making inserts into the DB for each value versus if it used a function or a class it would make it quicker. To make a function for a single module to do this is usually more time consuming than just coding it. But if we had a common function it might make the life of some of us dev's a little nicer. If I get time maybe I'll throw it together and do a pull request.

scottcwilson:
This issue is over 5 years old.  If someone wants to put in a PR it will be accepted, but I don't think the issue needs to be kept open.

PR #495, opened 2015-07-27
Expand admin/modules

bislewl:
Interesting thought inspired by #483, what about if the admin/modules.php was expanded to include plugins. The install function is there, along with remove and edit values. What about:

1. This was expanded to support external plugins 
2. It became the "preferred" method for installation/removal of plugins

That  I would think would reduce the need for all this custom code for auto-installers additionally the version number could be stored there as well so we could use the plugin_version_check_for_updates() function as part of it. And since there would be a uninstall function it should solve the need for uninstall files as well, it would also remove the need for the sometimes the 30-40 item in the configuration menu, since it could use the framework already set up. And it would allow for the ability to modify multiple configuration values at a single time. It seems like it was almost originally intended for this...


drbyte:
Do you have a plugin in mind to use as an example of how it would fit into this sort of structure?  Perhaps another github repo we can dialog about or collaborate in?

bislewl:
Sure, I haven't written one for this structure yet. 

I branched Ajax Back in Stock https://www.zen-cart.com/downloads.php?do=file&id=1944  https://github.com/bislewl/ajax_back_in_stock/tree/proposed_160_plugin_structure and added @drbyte  @zcwilt and @ajeh . Do you agree with the general premise?

drbyte:
> Do you agree with the general premise?

We're still discussing it. Willing to explore it before fully committing to it.

drbyte:
@bislewl ... I think it's best to close this PR for the time being. 

We're working on a project to refactor a bunch of things in the Admin, so the code landscape is going to change a bit, and it may be worth revisiting these thoughts after that's done.

Thoughts?

bislewl:
I agree

PR #497, opened 2015-07-27
Remove obsolete SWITCH credit cards

drbyte:
The SWITCH credit card (a former UK card) is now absorbed into Maestro. There is no need to treat it separately. 

NOTE: The define for SWITCH in admin settings is replaced with a "Debit Card" option, for future use by relevant payment integrations, rather than completely removing the option.


zcwilt:
:+1:

ajeh:
:+1:

PR #499, opened 2015-07-28
Add additional notifier hooks to admin orders page

drbyte:
There will be more of these in the future, but I'm submitting this PR now to clear out stale branches.


ajeh:
:+1:

PR #500, opened 2015-07-28
Fixes #422 fractional quantities

drbyte:
Previously it only handled fractions if `> 1`. This change allows handling even for `0.5` etc.


zcwilt:
:+1:

ajeh:
:+1:

PR #503, opened 2015-07-29
Extra underscore in parameter

drbyte:
The extra underscore shouldn't be there.

https://developer.paypal.com/webapps/developer/docs/classic/api/merchant/DoExpressCheckoutPayment_API_Operation_NVP/


ajeh:
:+1:

PR #504, opened 2015-07-29
Make error-handling notifier more functional and fix 84.0 error

drbyte:
The 84.0 implementation applied the PAYMENTINFO_0_ stuff too aggressively in one spot.

This also moves the notifier (added in the 84.0 integration) earlier so that it's more versatile.


ajeh:
:+1:

PR #511, opened 2015-07-31
Wrong product-type lookup on links for Upcoming Products listingBox - Fixes #508

### LINKED ISSUE #508: Upcoming Products wrong Product Type
On Upcoming Product the products_type is not correct and all Products, regardless of products_type are using Product General so they are creating links for product_info

drbyte:
Fixes #508 - wrong product-type lookup on links for Upcoming Products


ajeh:
:+1:

PR #512, opened 2015-07-31
Rearrange PayPal error message email slightly

drbyte:
Rearrange PayPal error message email slightly
The EC emails had info about the affected customer as last-line of the email. This moves it to the top.
Also does a quick sanitize of names, just in case.


ajeh:
:+1:

PR #513, opened 2015-07-31
Ensure Internet Explorer’s Quirks Mode and Compatibility View are disabled

drbyte:
Ensure Internet Explorer’s Quirks Mode and Compatibility View are disabled


ajeh:
:+1:

PR #515, opened 2015-08-01
Change layout-ctrlr buttons to be multi-lang CSS-driven

### LINKED ISSUE #140: Change button type to use Foundation for pr #121
pr #121 is using a standard ZC image button. Once foundation integration or admin has progressed further, change the button to use Foundation. Allowing for simple language control of the button text.

drbyte:
Fixes #140
NOTE: This page should also be converted to LEAD-style at some point
And this also points out that there are lots of other buttons needing reworking as part of a larger task affecting BC.


ajeh:
:+1:

PR #517, opened 2015-08-01
Point CURL testing tools to AWS endpoint

drbyte:
Point CURL testing tools to AWS endpoint
This will help with improving performance and reachability.


ajeh:
:+1:

PR #519, opened 2015-08-01
Fix small styling bug for customers_referral field in tpl_account_edit

drbyte:
Ref: https://www.zen-cart.com/showthread.php?217924-customer-referral-snip-in-edit-account


zcwilt:
:+1:

ajeh:
:+1:

PR #520, opened 2015-08-01
Fix empty-string error in shopping_cart::get_product_id_list

drbyte:
Ref: https://www.zen-cart.com/showthread.php?217915-shopping_cart-get_product_id_list-function

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -611,7 +611,7 @@ class shoppingCart extends base {
   }
   /**
    * Method return a comma separated list of all products in the cart
-   * NOTE: Not used in core ZC, but some plugins make use of it as a helper function
+   * NOTE: Not used in core ZC, but some plugins and shipping modules make use of it as a helper function
    *
    * @return string
    */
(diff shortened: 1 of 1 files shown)


includes/classes/shopping_cart.php
```diff
@@ -611,19 +611,20 @@ function remove_all() {
   }
   /**
    * Method return a comma separated list of all products in the cart
+   * NOTE: Not used in core ZC, but some plugins make use of it as a helper function
    *
    * @return string
-   * @todo ICW - is this actually used anywhere?
```
ajeh: Just a heads up: I actually have shipping modules that use this function so while it appears un-needed it is usefully for obtaining the string of products_id in the cart.

drbyte: Agreed. That's why I removed this comment and added another one above.

drbyte: Even better ... I'll update it to mention shipping modules specifically, for future clarity.

zcwilt:
maybe use a guard test

```php
function get_product_id_list() {
    if (!is_array($this->contents)) {
       return '';
    }
    reset($this->contents);
    $product_id_list = array();
    while (list($products_id, ) = each($this->contents)) {
      $product_id_list[] = $products_id;
    }
    return implode(',', $product_id_list);
  }
```

drbyte:
Good point.

zcwilt:
:+1:

ajeh:
:+1:

PR #521, opened 2015-08-01
Fixes layout_controller lookup of non-php files

drbyte:
Ref: https://www.zen-cart.com/showthread.php?217903-Layout-Controller-detects-files-with-extensions-other-than-php


drbyte:
There are a handful of additional files affected. See next commit.

zcwilt:
:+1:

ajeh:
:+1:

PR #523, opened 2015-08-02
Move message about missing order history to main admin language file

scottcwilson:
If the order history is missing, the invoice and packingslip pages will only show
TEXT_NO_ORDER_HISTORY, since this define is not available to them.


drbyte:
:+1: 

Are there other issues happening if this odd "no order history" scenario happens?  I'm curious why this hasn't been reported previously.

zcwilt:
:+1:

scottcwilson:
Chris: yes, there are other issues; this can't happen organically by using the system.  The situation I was in when I saw this bug was browsing a database created by a cart conversion service.   I think the other cart they were coming from didn't have the concept of order history so there was nothing to import.

PR #525, opened 2015-08-03
Fix formatting in functions-general; rename new function

drbyte:
Fixed line-ending issue
Added comments
Renamed function


ajeh:
:+1:

PR #526, opened 2015-08-03
Add retry button to admin alert_page

drbyte:
The retry link is only available if the admin folder has already been renamed but the zc_install has not. (... because we can't do a retry link if we don't know the name of the new admin folder yet)


ajeh:
:+1:

PR #527, opened 2015-08-03
Force to run zc_install over https if possible

drbyte:
Tests whether zc_install can be done over https, and if so then redirects accordingly.
This helps the proper auto-detection of setup params, and also causes the db password and all other setup data to be transmitted over https instead of in clear text.

This detection only works with domain-specific certificates, not shared certificates.


ajeh:
:+1:

zcwilt:
:+1:

PR #540, opened 2015-08-13
Relocate T&C checkbox to be next to submit button

### LINKED ISSUE #535: Terms and Conditions checked not checked issue
When on the checkout_payment if the Terms and Conditions is checked but the payment module is not, you are given the warning to select a payment method. However the Terms and Conditions are now unchecked.

The message does not show about needing to check the Terms and Conditions the second time through. This is very aggravating especially if selecting credit card as that loses the credit card settings when you see the message to check the Terms and Conditions so you check the terms and conditions and do not notice the credit card isn't setup and now you keep having to fix things until you get it all right.

drbyte:
Fixes #535

Further enhancement could include javascript to disable the submit button until the box is checked ... but only if the feature is enabled


ajeh:
:+1:

zcwilt:
:+1:

PR #542, opened 2015-08-13
Randomize also-purchased results

### LINKED ISSUE #536: Also purchased is not doing a good select method
Regardless of the number of Products purchased together, even if there are 100s of possible Products, only the first X matches ever show.

drbyte:
Closes #536


ajeh:
:+1:

zcwilt:
:+1:

PR #543, opened 2015-08-14
Disambiguation of $$ usage for PHP7 compatibility

drbyte:
These are some uses of indirect variables/properties/methods and globals which need some disambiguation for PHP 7 compatibility.

Ref: http://php.net/manual/en/migration70.incompatible.php


zcwilt:
:+1: 
passes selenium tests :)

ajeh:
:+1:

PR #546, opened 2015-09-01
Fix UTF8 sequence error in ajax call

drbyte:
Ref: https://www.zen-cart.com/showthread.php?216691-PHP-Warning-Invalid-UTF-8-sequence-in-argument-in-ajax-php-on-line-24

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -2,11 +2,11 @@
 /**
  * ajax front controller
  *
- * @package templateSystem
- * @copyright Copyright 2003-2014 Zen Cart Development Team
+ * @package core
+ * @copyright Copyright 2003-2015 Zen Cart Development Team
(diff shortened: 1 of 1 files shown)


drbyte:
Note: Uncertain whether this is entirely the correct fix. If it's really more specific to the PayPal module, then it might be worth making some additional/adjustments there.

zcwilt:
:+1:
I've no problem with this fix for now.

ajeh:
:+1:

drbyte:
Just posting to link to the alternate fix posted by @lat9 at https://www.zen-cart.com/showthread.php?21669124&p=1294488#post1294488

drbyte:
@zcwilt made a PR #613 to address this

PR #547, opened 2015-09-08
Update phpunit and related dependencies

drbyte:


zcwilt:
As noted in our conversation. PHP 5.4 is end of life today, but guess we need to support for a little while longer.
:+1:

ajeh:
:+1:

PR #549, opened 2015-09-08
Add `i` parameter to curltester for advanced debugging

drbyte:
This simply adds additional output to the CURL communications info, specifically about the certificate being connected to.  

This occasionally can be helpful when debugging, but I suspect will seldom be needed.


zcwilt:
:+1:

ajeh:
:+1:

PR #553, opened 2015-09-11
Make T&C "link" open in new window

drbyte:
This will help stop people from losing their place in checkout when they click the link


ajeh:
:+1:

PR #554, opened 2015-09-11
Update version stamping and zc_install to v1.5.5

drbyte:


bislewl:
Is there a release date set?

drbyte:
Not yet. It's premature to be making announcements.

PR #464, opened 2015-07-15
Authorize.net Akamai endpoint change

drbyte:
Per Authorize.net developer bulletin July 14, 2015:

>The new Akamai transaction URLs that are available now are:
`https://api2.authorize.net/xml/v1/request.api`
`https://api2.authorize.net/soap/v1/Service.asmx`
`https://secure2.authorize.net/gateway/transact.dll`

PR #560, opened 2015-09-20
Reverse PR #464 - Authorize.net Akamai Servers

drbyte:
Using "secure2" instead of "secure" was only an intermediate step until they switched everything over to the new system.


zcwilt:
:+1:

ajeh:
:+1:

PR #568, opened 2015-09-24
Remove extraneous code

ajeh:
Remove extraneous code that has no purpose and is never used

### DIFF
--- a/admin/products_to_categories.php
+++ b/admin/products_to_categories.php
@@ -351,8 +351,8 @@ function array_minus_array($a, $b) {
         break;
 
       case 'update_product':
-        $new_categories_sort_array[] = $_POST['current_master_categories_id'];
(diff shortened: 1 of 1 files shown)


admin/products_to_categories.php
```diff
@@ -351,19 +351,11 @@ function array_minus_array($a, $b) {
         break;
 
       case 'update_product':
-        $zv_check_master_categories_id = 'true';
         $new_categories_sort_array[] = $_POST['current_master_categories_id'];
         $current_master_categories_id = $_POST['current_master_categories_id'];
```
drbyte: Both of these POSTs should be (int) here too, because of the way they're used later. (Don't want them accidentally coming in as an array or string when going thru parsing loops, and there are some outputs later that are easier to clean up if `int`'d here.)

drbyte:
:+1:

PR #581, opened 2015-09-30
Enforce additional restrictions on uploaded files

drbyte:
Credits to [curesec.com](http://blog.curesec.com/article/blog/Zen-Cart-154-Code-Execution-and-Information-Leak-54.html) for pointing out possible problems where these restrictions don't exist.


zcwilt:
:+1:

ajeh:
:+1:

PR #582, opened 2015-09-30
Get rid of zc_install test for PHP CGI bug in PHP 5.4.0/5.4.1

drbyte:
While PHP 5.4 is EOL, the ZC v160 code currently still works on PHP 5.4, so the minimum is being set to 5.4.2 for now (since 5.4.2 fixed the CGI bug)


zcwilt:
:+1:

ajeh:
:+1:

PR #583, opened 2015-09-30
zc_install writable dirs

### LINKED ISSUE #405: Should we have zc_install attempt to make writable dirs where needed?
The zc_install inspection process currently only "reports" whether necessary folders are writable, and doesn't actually attempt to make the necessary changes.

Granted, it can't make any changes if permissions won't allow, but in cases where the server does allow it, we should make the necessary changes automatically.

drbyte:
Closes #405

### DIFF
--- a/zc_install/includes/classes/class.systemChecker.php
+++ b/zc_install/includes/classes/class.systemChecker.php
@@ -264,6 +264,13 @@ class systemChecker
   }
   public function checkWriteableDir($parameters)
   {
+    if ($parameters['makeWritable'] === true) {
+      foreach(['666', '766'] as $perm) {
+        if (!is_writeable($parameters['fileDir'])) {
+          chmod($parameters['fileDir'], octdec($perm));
+        }
+      }
+    }
     return is_writeable($parameters['fileDir']);
   }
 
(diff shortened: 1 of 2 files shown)


zc_install/includes/classes/class.systemChecker.php
```diff
@@ -264,6 +264,13 @@ public function dbVersionCheckConfigDescription($db, $dbPrefix, $parameters)
   }
   public function checkWriteableDir($parameters)
   {
+    if ($parameters['makeWritable'] === true) {
+      foreach(['666', '766'] as $perm) {
```
drbyte: Pinging @barco57 for thoughts on a safe progression of perm levels to attempt to set here.  Concern is granting too much, but also granting too little such that one locks oneself out. And of course there's the desire to be as broadly-compatible with various hosting server configurations as possible. 

Feedback welcome.

barco57:
My experience has always been with cPanel servers, and I am forever making sure every bodies perms are set 755 for folders and 644 for files. As I recall, where not declared, the file perms are set too 666 as a default when the applications create a file. I see that with image handler bmz cache files all of the time. In a suPHP environment, I have always stuck with 755/644 as a strict suPHP environment does not allow permissions 666 and 777.

PR #588, opened 2015-10-05
Change db loops in customers.php to new iterator style

### LINKED ISSUE #406: [Task] New db query result loops to use foreach iterator instead of while !EOF...MoveNext
Starting with v1.5.5 looping through database queries no longer requires the complex `while` loop, and can instead use a simple `foreach` iterator. This is more intuitive and more maintainable.

```php
while (!$result->EOF) { 
  $data = $result->fields['filename'];
  //foo
  $result->MoveNext();
}
```

becomes:

```php
foreach($result as $row) {
  $data = $row['fieldname'];
  //foo
}
```

This note isn't suggesting to engage in mass-rewriting. Instead, be mindful of the new construct, and use the iterator approach when writing new code.

scottcwilson:


drbyte:
:+1:

zcwilt:
:+1:

PR #589, opened 2015-10-05
Fixes #69 Guard against modifying product type in URL

### LINKED ISSUE #69: product_type parameter in admin is not validated properly
if the GET parameter for product_type is altered to use an out of range value, the code still does the look up but returns a blank hander name. It then attempts to load product handler code based on an incorrect path.

scottcwilson:
Test Case: 
https://zen.local/admin/index.php?cmd=product_free_shipping&page=1&product_type=999&cPath=64&pID=172&action=new_product

without my change, will issue a debug log and a blank page.  With my fix, will go to the product but ignore the product_type change.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -2727,7 +2727,7 @@ function zen_copy_products_attributes($products_id_from, $products_id_to) {
     $sql = "select type_handler from " . TABLE_PRODUCT_TYPES . " where type_id = '" . (int)$product_type . "'";
     $handler = $db->Execute($sql);
     if ($handler->EOF) { 
-          $messageStack->add('ERROR: Invalid type_handler. Your product_type settings are wrong, incomplete, or damaged.', 'error');
+          $messageStack->add('ERROR: Invalid product_type specified.', 'error');
           return -1;
     }
     return $handler->fields['type_handler'];
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -2722,10 +2722,14 @@ function zen_copy_products_attributes($products_id_from, $products_id_to) {
  */
   function zen_get_handler_from_type($product_type) {
     global $db;
-
+    global $messageStack;
+    
     $sql = "select type_handler from " . TABLE_PRODUCT_TYPES . " where type_id = '" . (int)$product_type . "'";
     $handler = $db->Execute($sql);
-    if ($handler->EOF) return 'ERROR: Invalid type_handler. Your product_type settings are wrong, incomplete, or damaged.';
+    if ($handler->EOF) { 
+          $messageStack->add('ERROR: Invalid type_handler. Your product_type settings are wrong, incomplete, or damaged.', 'error');
+          return -1;
+    }
     return $handler->fields['type_handler'];
   }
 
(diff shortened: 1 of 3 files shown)


admin/includes/init_includes/init_check_parms.php
```diff
@@ -0,0 +1,17 @@
...
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ */
+if (!defined('IS_ADMIN_FLAG')) {
+  die('Illegal Access');
+}
+if (isset($_GET['product_type'])) { 
+  $sql = "select type_handler from " . TABLE_PRODUCT_TYPES . " where type_id = '" . (int)$_GET['product_type'] . "'";
+  $products_type_check = $db->Execute($sql); 
+  if ($products_type_check->EOF) {
```
drbyte: Can I recommend using `zen_get_handler_from_type()` instead of writing code for another query here? 

And the check could probably be just to see whether its return-type is `string` vs `int`.

drbyte:
I guess given that this fix applies to at least the 5 built-in product types, that this universal `init` approach is fine.

scottcwilson:
Chris - happy to do that but the intent was really just to see if the row existed - I could have done select count(*) just as well.  Which do you think is more clear?

drbyte:
My point is to put fewer hard-coded queries into code, and instead re-use functions where the same query is executed already.

scottcwilson:
Had to fix zen_get_handler_from_type() which returned a string in either case.

admin/includes/functions/general.php
```diff
@@ -2722,10 +2722,14 @@ function zen_get_shipping_enabled($shipping_module) {
...
   function zen_get_handler_from_type($product_type) {
     global $db;
-
+    global $messageStack;
+    
     $sql = "select type_handler from " . TABLE_PRODUCT_TYPES . " where type_id = '" . (int)$product_type . "'";
     $handler = $db->Execute($sql);
-    if ($handler->EOF) return 'ERROR: Invalid type_handler. Your product_type settings are wrong, incomplete, or damaged.';
+    if ($handler->EOF) { 
+          $messageStack->add('ERROR: Invalid type_handler. Your product_type settings are wrong, incomplete, or damaged.', 'error');
```
drbyte: Okay, if we're going to display the message to the user, this is a good time to change the message to something more specific, and shorter, like:

`ERROR: Invalid product_type specified.`

admin/includes/init_includes/init_check_parms.php
```diff
@@ -0,0 +1,15 @@
...
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ */
+if (!defined('IS_ADMIN_FLAG')) {
+  die('Illegal Access');
+}
+if (isset($_GET['product_type'])) { 
+  if (zen_get_handler_from_type((int)$_GET['product_type']) == -1) {
+      unset($_GET['product_type']); 
+  }
```
drbyte: To avoid the arrow anti-pattern, let's combine the `if isset` with the `if zen_get_handler_from_type` using `&&`

drbyte:
:+1:

ajeh:
:+1:

PR #592, opened 2015-10-07
Fix error in earlier commit: Remove the double-slashes inserted into …

drbyte:
…the configure.php files by zc_install

Error introduced in PR#585


zcwilt:
:+1:

PR #623, opened 2015-10-18
Language constants: Consolidate defines for Product Model

### LINKED ISSUE #603: Language define consolidation - Model Name
If you look at a string like Model, it's defined 13 times in the catalog and 33 times in the admin.  Now some of these are different from others, but if you just consolidated (say) TEXT_PRODUCTS_MODEL, you'd get a dozen of them.  What if we promote these to the includes/languages/english file in the catalog and admin?

scottcwilson:
Consolidates model and title string for model to one place in catalog, one place in admin, reducing by 42 the number of strings that need to be updated if this name is to be changed.  PRODUCTS_NAME, _DESCRIPTION and _QUANTITY are other strings which could be consolidated in the same way if desired, but MODEL seems like the best choice, since some stores have a strong preference for another name here.  Fixes #603


zcwilt:
Where appropriate, can you update the copyright dates.
Thanks

admin/includes/languages/english/products_to_categories.php
```diff
@@ -14,13 +14,11 @@
 
 define('TABLE_HEADING_PRODUCTS_ID', 'Prod ID');
 define('TABLE_HEADING_PRODUCT', 'Product Name');
-define('TABLE_HEADING_MODEL', 'Model');
```
drbyte: You've removed this, but not re-defined it elsewhere, yet it's used in `category_product_listing.php` and `downloads_manager.php`

zcwilt:
:+1:

drbyte:
:+1:

PR #626, opened 2015-10-19
Backward compatibility for media-copy checkbox handling

drbyte:
Since both 'on' and '1' have been used, and different browsers are giving different results, this change accommodates both, and will allow for potentially-missed file changes in upgrades/plugins.


zcwilt:
:+1:

ajeh:
:+1:

PR #631, opened 2015-10-20
Currency override suggestions from @lat9

drbyte:
Ref: https://www.zen-cart.com/showthread.php?218530-PayPal-Express-Checkout-Rounding-issue-causes-quot-Line-item-subtotals-do-not-add-up-quot&p=1294598#post1294598

Not sure I like this line, which completely overrides the tax-inc/exc calcs in the preceding lines:

```php
    $tax_add = $currencies->value($tax_add, false, $this->currency);
```

UPDATE: After further study, I'm less bothered about this $tax_add line than I initially thought.

### DIFF
--- a/includes/classes/currencies.php
+++ b/includes/classes/currencies.php
@@ -121,6 +121,6 @@ class currencies extends base {
   }
 
   function display_price($products_price, $products_tax, $quantity = 1) {
-    return $this->format (zen_add_tax($products_price * $quantity, $products_tax));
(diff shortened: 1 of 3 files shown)


drbyte:
UPDATE: After further study, I'm less bothered about this $tax_add line than I initially thought.

zcwilt:
:+1:

ajeh:
:+1:

PR #645, opened 2015-11-04
Messages should be shown even if cart is empty

scottcwilson:
If an error occurs but the cart is still empty (perhaps because adding to the cart failed), the messages should still be shown at the top of the shopping cart page.


zcwilt:
:+1:

drbyte:
:+1:

PR #647, opened 2015-11-11
Fix admin timezone chicken/egg problem

drbyte:
Need to load application_top before calling strtotime() so that the timezone overrides are executed first.


zcwilt:
:+1:

ajeh:
:+1:

PR #648, opened 2015-11-20
remove tell a friend remnance

Zen4All:
Shouldn't these two tell a friend variables be removed?

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/648)


drbyte:
I had left those so that people using a tool like the Email Archive Manager could trigger a re-send and still have those lines handle the data properly. Or someone who had adapted Tell A Friend into a plugin could still have this part of the code work despite there being no "hook" to activate it otherwise.

Zen4All:
That makes sense

PR #649, opened 2015-11-20
Change currency-symbol field length to allow font-icons as symbols

drbyte:
ref suggestion by s_mack https://www.zen-cart.com/showthread.php?218821


zcwilt:
:+1:

ajeh:
:+1:

PR #650, opened 2015-11-24
Query can be simpliefied

Zen4All:
The info in this query is already present in $reviews.

[Review on Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/650)

### DIFF
--- a/admin/reviews.php
+++ b/admin/reviews.php
@@ -254,7 +254,7 @@ require('includes/admin_html_head.php');
 
     $order_by = " order by pd.products_name";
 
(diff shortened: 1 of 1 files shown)


drbyte:
Good catch.

In fact, if we simply add this to the `$reviews_query_raw` on line 257, this query you're mentioning can most likely be removed altogether.

`, length(rd.reviews_text) as reviews_text_size`

Zen4All:
The file has been updated

zcwilt:
:+1:

drbyte:
:+1:

PR #656, opened 2015-12-01

jtwebsitedesign:
Update line endings so changes would appear line-by-line instead of replacing entire files. As described here: https://help.github.com/articles/dealing-with-line-endings/


jtwebsitedesign:
Cancelling this request, not all my changes made it into the commit.

PR #659, opened 2015-12-02

jtwebsitedesign:
Update line endings so changes would appear line-by-line instead of replacing entire files. As described here: https://help.github.com/articles/dealing-with-line-endings/


drbyte:
Still debating whether/when we want to incorporate this massive change.

That said, your other proposed changes to admin styling as a separate PR are something we're still interested in seeing. ;)

jtwebsitedesign:
Yes, and sorry for the delay! I was quite busy over the holidays and I've been playing catchup lately. I should have some more free time in the near future, and will submit another PR soon :)

PR #661, opened 2015-12-07
Paypal Refund message text box should be limited to 255 chars

drbyte:
Added char-counter to Refund text box

Ref http://www.zen-cart.com/showthread.php?211173-Admin-Order-Details-Paypal-Refund-message-text-box-is-not-limited-to-255-chars


ajeh:
:+1:

zcwilt:
:+1:

PR #663, opened 2015-12-07
Fix counter_history table primary key to be NOT NULL

drbyte:


drbyte:
Duplicate PR from wrong branch

PR #664, opened 2015-12-07
Indirect-expression fixes for PHP7

drbyte:
These are related to use of $$foo['element'] which need to add the {braces} for specificity


ajeh:
:+1:

zcwilt:
:+1:

PR #676, opened 2015-12-10
CHANGE-465 - Add more notifier hooks to dynamic metatags generator module

drbyte:
These hooks can be used in various ways:
- As in previous versions, hooking `NOTIFY_MODULE_START_META_TAGS` allows one to completely intercept the default metatags generation altogether
- `NOTIFY_MODULE_META_TAGS_BUILDKEYWORDS` allows overriding of just the dynamic generation of keywords
- `NOTIFY_MODULE_META_TAGS_UNSPECIFIEDPAGE` allows the injection of metatags for pages that don't have built-in generation logic in the default module

Also allows an observer on `NOTIFY_MODULE_START_META_TAGS` to change the value of `$metatag_page_name` ... which could force the usual rules to be bypassed (and thus trigger "default"), which can then be intercepted by `NOTIFY_MODULE_META_TAGS_UNSPECIFIEDPAGE` if necessary, to set specific values.

### DIFF
--- a/includes/modules/meta_tags.php
+++ b/includes/modules/meta_tags.php
@@ -8,11 +8,13 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: meta_tags.php drbyte  Modified in v1.6.0 $
  */
+$meta_tags_over_ride = false;
+$metatag_page_name = $current_page_base;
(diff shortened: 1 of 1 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #679, opened 2015-12-11
Fix check for order amount, to handle chicken/egg case when doing validations or ajax

drbyte:
Problem discovered when testing ajax-checkout


zcwilt:
:+1:

ajeh:
:+1:

PR #681, opened 2015-12-11
Fix for: zen_draw_date_selector "times out" in 2018

drbyte:
Ref: https://www.zen-cart.com/showthread.php?219015

Now the default range is 5 years "ago" until 10 years "in future"


zcwilt:
:+1:

ajeh:
:+1:

PR #692, opened 2015-12-28
Add insert-IDs to various order notifier hooks for attribute-related data

drbyte:
These additional insert-id values can aid in tracking inventory related to attributes

Per suggestion from @mc12345678 at https://www.zen-cart.com/showthread.php?215320-order-notifiers


zcwilt:
:+1:

ajeh:
:+1:

PR #698, opened 2015-12-29
Add patch to provide additional aggressive xss filtering on admin side

drbyte:
Add patch to provide additional aggressive `xss` filtering on admin side

NOTE: the admin is already protected against `csrf` issues, so `xss` problems require being logged in with a valid username and password

Credits to Trustwave Security for pointing out these minor risks and helping tighten security overall.


zcwilt:
:+1:

ajeh:
:+1:

PR #701, opened 2015-12-29
Treat all email fields as 'email' type in HTML5 - more mobile friendly

### LINKED ISSUE #625: Code suggestion:  default use of html5 "email" data type
This just came up as I was re-doing my template and had to edit half a dozen forms.

When a user is asked to input an email address, the data type for that field should be "email" not "text".  It is a safe change because the default "text" input type will be the fall-back if the browser doesn't support the new data type ([source](http://www.wufoo.com/html5/types/1-email.html)).

The advantage is a user with a iPhone or Android will have their keyboard default to showing @ and . symbols instead of having to switch keyboard layouts.

Simple change to make.  Anywhere we have:

zen_draw_input_field('email_address' [...]

the 4th parameter, which seems to be blank (and defaulting to "text" in the function), needs to simply be "email".

Alternatively, and perhaps better, would be some logic in the zen_draw_input_field to recognize (based on first parameter) if an email address is being specified, and then automatically apply the "email" type instead of "text" unless, of course, "text" is explicitly stated.

drbyte:
Fixes #625 - Treat all email fields as the more mobile-friendly 'email' type

### DIFF
--- a/admin/admin_account.php
+++ b/admin/admin_account.php
@@ -1,10 +1,10 @@
 <?php
 /**
  * @package admin
- * @copyright Copyright 2003-2013 Zen Cart Development Team
+ * @copyright Copyright 2003-2015 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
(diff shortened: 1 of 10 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #703, opened 2015-12-29
Fix problem of duplicate sort-orders on module settings

drbyte:
Fixed long-standing problem of duplicate sort-orders wiping each other out ... in Admin->Modules screens.


ajeh:
:+1:

zcwilt:
:+1:

PR #708, opened 2015-12-30
Move warnings about zc_install dir etc from catalog to admin, for privacy

drbyte:
Move zc_install and configure.php warnings from catalog to admin, and consolidate all error-detection from /admin/includes/header.php with the rest of the error stuff in /admin/includes/init_includes/init_error.php


zcwilt:
:+1:

ajeh:
:+1:

PR #720, opened 2015-12-31
Fix authorize.net to convert currencies to gateway-supported value

drbyte:
Authorize.net supports multiple currencies, but only one-currency-per-gateway-account. 

Previously there was a bug whereby transactions would be charged the same amount but pass the wrong currency code instead of converting both amount and currency correctly to the single currency supported by the account. This change now forces the storeowner to specify which currency their gateway account is configured for, and converts purchases to that amount before submission.


ajeh:
:+1:

PR #723, opened 2015-12-31
Add the order-status override to SIM using what AIM does

drbyte:
When the module is in Authorize-Only mode, order-status gets overridden to 1-Pending since the payment is indeed actually


zcwilt:
:+1:

ajeh:
:+1:

PR #724, opened 2015-12-31
PayPal's server doesn't seem to be responding. Reverting prior commit.

drbyte:
Apparently although PayPal is promoting their new postback URL, it's still not functioning consistently...

https://www.paypal-knowledge.com/infocenter/index?page=content&id=FAQ1916


ajeh:
:+1:

PR #748, opened 2016-01-03
Migrate this report to new iterator style

scottcwilson:
This report already uses the new report style so I thought
it should have the new iterator too so it's a good template.


drbyte:
:+1:

drbyte:
Thanks Scott.
You did this PR against v160. Are you content with it being merged into v160 and not v155?

zcwilt:
:+1:

PR #750, opened 2016-01-03
Products Price Manager: Add display of taxed prices and specials

mc12345678:
ZC thread: https://www.zen-cart.com/showthread.php?217854-Products-Price-Manager-doesn-t-offer-display-of-prices-and-specials-with-tax
Style used was similar to product_info entry page.


zcwilt:
:+1:

ajeh:
:+1:

PR #755, opened 2016-01-05
Pagination as Unordered List

drbyte:
The generated markup embraces class names compatible with Bootstrap 3 and Foundation For Sites 6, just to prevent the need to alter core files if using those frameworks.


zcwilt:
:+1:

ajeh:
:+1:

PR #757, opened 2016-01-06
Change deprecated clear="all" syntax to class="clearBoth"

drbyte:
(requires `.clearBoth{clear:both}` which is already in default stylesheet)


zcwilt:
:+1:

ajeh:
:+1:

PR #759, opened 2016-01-06
Add additional sanitization to the `request` array

drbyte:
This is to tighten security against plugins which fail to properly sanitize inputs or bypass get/post


zcwilt:
:+1:

ajeh:
:+1:

PR #760, opened 2016-01-06
Remove deprecated `language=javascript`

drbyte:
... and add `type=text/javascript` for compatibility with both HTML5 and HTML4

NOTE: for upgraders, updating these files isn't urgent, but will reduce "validator" errors when doing site-inspections. Functionally, modern browsers will be happy either way.


zcwilt:
:+1:

ajeh:
:+1:

PR #764, opened 2016-01-06
When CSS Buttons are enabled, the 'class' attribute on this zen_image_button showed twice

drbyte:
The extra class in this case needs to be the 4th param instead of the 3rd.


drbyte:
Hmmm ... this needs more work ...

PR #771, opened 2016-01-08
Adaptations for HTML5 compatibility and fix Validator errors

drbyte:
NOTES: 

- altered all tables to use CSS for width and border styling, adding the table css to the main `stylesheet.css`
- changed the `#cartImage` and `#cartProdTitle` IDs to classes: `.cartImage`, `.cartProdTitle` in `stylesheet.css`
- added some designer tools such as code to uncomment for testing messageStack styling, and HTML Validator links in the footer which can be enabled for testing
- fixed validation errors caused by incorrect markup, inefficient nesting, incorrect ordering of tags, etc

Technically, now even `classic` should comply with HTML5 requirements if one were to set the correct doctype and corresponding meta in the head section.


zcwilt:
:+1:

ajeh:
:+1:

PR #780, opened 2016-01-10
Add catchalls in case someone puts the catalog configure.php in admin

drbyte:
Makes guesses in case the essentials from admin configure.php are missing (such as when someone uses a non-admin configure.php in their admin)

These guesses aren't foolproof, but they're reasonable assumptions which should avoid unnecessary errors from occurring if the end-user makes this mistake. Multiple nested subdirectories might pose problems ... in which case the fix is to actually use the proper configure.php files in the proper locations ... ideally what is generated by zc_install.

### DIFF
--- a/admin/includes/defined_paths.php
+++ b/admin/includes/defined_paths.php
@@ -38,6 +38,18 @@ function zen_parse_url($url, $element = 'array')
   }
 }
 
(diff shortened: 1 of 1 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #782, opened 2016-01-10
Add constraint to the width of the admin system-info page

drbyte:
On tablets the sysinfo page was shrinking text to unreadably small size. This constrains it to a reasonable width.


zcwilt:
:+1:

ajeh:
:+1:

PR #787, opened 2016-01-12
Maintain order status that is above default on PayPal batch capture

mc12345678:
Discussion held back on ZC 1.5.1 at https://www.zen-cart.com/showthread.php?214423-Order-status-changes-after-batch-capture-on-PayPal-website where store owner observed that the status of orders that were previously manually updated inside of the ZC admin were lowered once batch capture was performed on PayPal site.

The change proposed above was reported as working at least before the `and` was added to the added `if` statement.  `And` statement intent is to apply/allow the old status to remain if and only if the transaction is one that isn't intended to reduce the status.  For example, refunding, reversing, denying or failing a transaction should not allow the status to remain elevated through this process, but instead take the status intended for those conditions.


ipn_main_handler.php
```diff
@@ -470,6 +470,13 @@
       if ((int)$new_status == 0) $new_status = 1;
       if (in_array($_POST['payment_status'], array('Refunded', 'Reversed', 'Denied', 'Failed'))
            || substr($txn_type,0,8) == 'cleared-' || $txn_type=='echeck-cleared' || $txn_type == 'express-checkout-cleared') {
+        $sql = ("select orders_status from " . TABLE_ORDERS . "
+                 where orders_id = :ordersID:");
```
drbyte: I'd remove the open and close `(` parentheses `)` from this string, since we don't want the string to be treated as a boolean `true/false`.

drbyte:
The concern I have with this change is geared around how some people use (or, rather, how they "add") order-statuses. 

There's a feature-deficiency in Zen Cart presently whereby there's no way to define an "intended progression" of order statuses in a certain direction. Your PR code suggestion is geared on the order-status-id being the determiner of "intended progression".  In reality I find a lot of storeowners just "add more statuses" but don't do the much more technical re-numbering of statuses (and corresponding order history) to ensure that the status-id is relocated into the "desired sequence".

To that end, I think a "sort order" equivalent would be a more reliable determiner of the right progression ... and then your code suggestion in this PR would be very ideal ... in fact it would be a critical necessity.

That said, I've also found that some people are overzealous and way too many order-statuses and just complicate things for themselves. But that's another topic.

What I'd like to see in a future ZC version is the ability to assign "roles" to each (or at least most) statuses, and then things like payment modules (and the kind of update you're talking about in this PR) would quickly know which status to "jump to" when a change is triggered by some system event.

mc12345678:
Had a similar concern about its usage/application.  It was then that I looked at how the $new_status condition was generated.  For the purposes of the originally reported issue if there is a condition that "minimizes" the current possibility of "incorrect" sequencing on such a bulk update then think that it too would be an acceptable answer.

What was seen/expressed by the OP was that items locally processed as some status other than pending that it would be returned to status of 2 which when reviewing the code above this section seemed to be the design of the code to either auto-assign as status 2 or to the status defined as the "goto" status when successfully processing/paid for. It seemed that either the solution would work for the majority of individuals or those that had an issue with it would have had problems for more than one reason and that likely would have sought assistance from the forum  

Have seen the discussion somewhere about status sorting versus status numbers, separating the two features that were merged into one into two separate "variables".  Would seem to be good to me as well...

I did notice that if the code above were incorporated that there is an ipn_notice debug email that would include incorrect information; however, nothing had been done about that as if the $new_status were 0 and adjusted to 1 then that ipn_notice would have already been incorrect.  So not sure if the position of the ipn notice:
 ` ipn_debug_email('IPN NOTICE :: Set new status ' . $new_status . " for order ID = " .  $ordersID . ($_POST['pending_reason'] != '' ? '.   Reason_code = ' . $_POST['pending_reason'] : '') );`
should be moved from its current location just because of the reassignment of its value before storage in the database.

drbyte:
Still trying to figure out the best way to incorporate the sentiment behind this.

drbyte:
Upon further review, I agree with this PR. 

Can you make 2 changes:

1. Redo it against `v160` instead of `v155`
2. As you suggested, above, move the `ipn_debug_email()` call down below your additions, so that it's just above the `ipn_update_orders_status_and_history` line following your additions.

mc12345678:
Absolutely!

I had begun a sort of summary about how the above process might need to be presented so that a different method of managing might be developed, but then it was all lost and I didn't get the umph again to rewrite it. :) 

Will address those requested items in the next day or two.

mc12345678:
Looking at the location to place the ipn_debug_email, one, by moving it out of the main flow path, an indicator is lost to anyone "following" the debug trail, second, there appear to be two places in this area where the status could change and the first occurrence does not necessarily ensure entry into the second if statement area as well, the modification made from status of 0 to status of 1 would not be identified as a "problem"..

Just want to be sure that the goal is to only issue the debug statement when the following is true and then only after all "changes" have been made: 
```php
if (in_array($_POST['payment_status'], array('Refunded', 'Reversed', 'Denied', 'Failed'))
           || substr($txn_type,0,8) == 'cleared-' || $txn_type=='echeck-cleared' || $txn_type == 'express-checkout-cleared') {
```

drbyte:
If I recall correctly, the logging is probably only necessary just before the actual execution of the change of the data in the database.

PR #788, opened 2016-01-12
Display orders in Admin even if orders_status is invalid (with an alert)

drbyte:
Previously any orders whose `orders_status` value didn't match a value in the `orders_status` table would be excluded from display in the Admin.

This now includes those orders, along with an alert which informs the administrator that the order-status is invalid.

(As always, to fix an order in this condition, simply go into the order and do an update to a new order-status using the dropdown menu)


zcwilt:
:+1:

ajeh:
:+1:

PR #795, opened 2016-01-22
Add spans for productBasePrice and productFreePrice

drbyte:
This suggestion from rbarbour adds additional styling targets to prices.


ajeh:
There is a closing </s> shouldn't that be removed?

zcwilt:
:+1:

ajeh:
:+1:

PR #796, opened 2016-01-22
Add ajax admin controller

bislewl:


drbyte:
In 1.6.0 there's already an Ajax handler in the admin.

bislewl:
My Bad. I didn't see one like that on the catalog I assumed there wasn't one. Sorry

PR #799, opened 2016-01-23
Fix attribute controller error when using option names that only have no values

### LINKED ISSUE #800: attribute controller errors with option names that only have no values - issue #799
Tested method to letting attributes controller to allow adding attributes to product when the option names assigned do not have (by design) associated values is addressed in issue #799. 

If recollection is correct this affects more than ZC 1.5.5.  I believe I have seen others in the past experience this.

Alternate non-coding solution is to add an option name that will accept an option value and then to assign an option value to that option name. On next run of the attributes controller no further error messages are displayed about the absence of option values (and still none for option names as now at least two would be in the database).

Issue would be apparent for stores that use only text attributes or file upload attributes. Similar situation would occur if some option name was created by a developer that did not use option values with the associated option name. All such conditions are addressed in the above pull request in admin/includes/functions/general.php (accessible by the admin) as well as for store front use if deemed necessary in includes/functions/functions_lookups.php.

mc12345678:
To address the issue identified by stellarweb at/near https://www.zen-cart.com/showthread.php?219078-FEEDBACK-ON-BETA-of-v1-5-5&p=1302032#post1302032

Effectively, if the populated option_names do not have an associated option_value, then when opening attribute_controller no further action can be performed.  Attached is a function (possibly inappropriately placed in the store side instead of the admin side, but also perhaps useful on both sides) that checks the provided option_id against the option_type and if the option_type is one that inherently does not have an option value then attributes_controller.php can continue with the next check.  If it should have an option_value then the error message is provided.  This additional code
only is run if 1 or more option name is assigned.  

Also a notifier was added so that if a developer has added an option name that has no corresponding option value then they can disable the "alert" received at least on the attributes_controller and processing can continue.  Otherwise, their additional option_name is expected to have generated an option_value.

Fixes #800


mc12345678:
Finally tested and corrected.

Function copied to admin/includes/general.php and corrected in includes/functions/functions_lookups.php for store side use.

mc12345678:
FWIW, the "name" of the notifier hook is not previously used for anything and was a best "guess" at an appropriate designation. No hardship in it being modified/altered. Might suggest a different one for admin and for store side though if they are both to remain.

mc12345678:
See Issue #800 for Issue tracking.  (Supposed to have been referenced by the text in the title of that thread so that they would all be tied together.)

mc12345678:
Have seen elsewhere that there is an expectation of "testing" for provided pull requests.  Wasn't sure what form this testing ought to take, but in words for this pull, I foresee testing to be like the following.

Begin with a fresh install without the sample products.
1. Attempt to access the attributes manager-> expected result is that notified that no option names exist and taken to option names manager.
2. Add an option name at the option names manager, preferably something like a dropdown or radio button.
3. Return to attributes manager -> expected result is to be notified that no option values are present.  This will redirect to the option values manager. 
4. Add an option value to the available option name.
5. Return to attributes manager -> expected result is to be notified that no product exists.  Test complete for attribute that has an associated option value.
6. Return to the option values manager and delete the option value.
7. Return to the option name manager and add an option name that does not include an associated option value such as a text box or a file upload field.
8. Return to the attributes manager -> expected result is to be notified that no product exists.  Test complete to identify that an option name that does not require an option value as presented from the attributes area is recognized as sufficient to not require notification of the absence of option values... (A database full of text option names would never have an option value).
9. Other tests that could be performed but would be anomalous would include a database revision that removed an option name from an option value pair leaving behind option values in the database or if an option value was applied to an option name that does not require an option value such as the text/file upload option name types, and would expect that as long as an option name and an option value existed in each table then the code would continue on to a product error (provided no product have been added).

admin/includes/functions/general.php
```diff
@@ -2135,6 +2135,43 @@ function zen_products_id_valid($products_id) {
...
+      $option_name_id = array($option_name_id);
+    }
+    $sql = "SELECT products_options_type FROM " . TABLE_PRODUCTS_OPTIONS . " WHERE products_options_id :option_name_id:";
+    if (sizeof($option_name_id) > 1 ) {
+      $sql2 = 'in (';
+      foreach($option_name_id as $option_id) {
+        $sql2 .= ':option_id:,';
+        $sql2 = $db->bindVars($sql2, ':option_id:', $option_id, 'integer');
+      }
+      $sql2 = substr($sql2, 0, -1); // Need to remove the final comma off of the above.
```
drbyte: I suggest just using `rtrim($var, ',')` instead of `substr()`

admin/includes/functions/general.php
```diff
@@ -2135,6 +2135,43 @@ function zen_products_id_valid($products_id) {
...
+      $sql2 = ')';
+    } else {
+      $sql2 = ' = :option_id:';
+      $sql2 = $db->bindVars($sql2, ':option_id:', $option_name_id[0], 'integer');
+    }
+    $sql = $db->bindVars($sql, ':option_name_id:', $sql2, 'noquotestring');
+    $sql_result = $db->Execute($sql);
+    foreach($sql_result as $opt_type) {
+      $test_var = true; // Set to false in observer if the name is not supposed to have a value associated
+      $zco_notifier->notify('FUNCTIONS_LOOKUPS_OPTION_NAME_NO_VALUES_OPT_TYPE', array($test_var), $test_var);
```
drbyte: Instead of `array($test_var)` which is basically throwaway, I suggest passing `$opt_type` as the 2nd param. And I rarely use an array in the 2nd param unless I really need to pass multiple immutable items.

admin/includes/functions/general.php
```diff
@@ -2135,6 +2135,43 @@ function zen_products_id_valid($products_id) {
     }
   }
 
+
+/*
+ *  Check if option name is not expected to have an option value (ie. text field, or File upload field)
+ */
+  function zen_option_name_no_values($option_name_id) {
```
drbyte: Does it make better sense to name this function `zen_option_name_NEEDS_no_values`? (caps only for emphasis) ... or something else slightly more semantic? When looking at it I find myself wondering if the name means "has no values?" or "needs no values?" or "should never have any values?", etc.

admin/includes/functions/general.php
mc12345678: Knew there was a better way to do that (not to mention that -1 is not enough to address/remove the comma and space).

admin/includes/functions/general.php
mc12345678: On retrospect, the above (original code) is not useful enough to take advantage directly of the notifier, agree that $opt_type would be significantly beneficial in this (especially as immutable) though also does mean that this solution still does not support older versions of ZC without inclusion of the associated code to support continued use of the modifiable 3rd+ parameter.

admin/includes/functions/general.php
mc12345678: Something is better than what was "quick" generated.  Was trying to keep short, but also not specific to the attribute controller and still make sense for other uses. (May be desired/needed to have an option value as part of the submission process to the shopping cart.  May just need to be a name that when questioned is evaluated by looking at the code/notes.)

mc12345678:
Comments incorporated into branch.  Name chosen for now: zen_option_name_base_expects_no_values.

ajeh:
:+1:

drbyte:
:+1:

judygunderson:
PERFECT!!  Thanks Chad!

mc12345678:
See a problem, fix a problem. :)

Doesn't relate to the current attribute controller issue, but there is currently a limitation on how the added function works with regards to sending an array of option names. As is, will return that an option value is expected if any option name expects a value. This could be improved to provide a varety of results (using an additional optional parameter) to provide results such as how many, if all require, etc... Of course,  a form of divide and concur could be applied in trying to use the function to identify such information... At least it's a start. May the future users of it find ways to improve on it. :)

PR #801, opened 2016-01-28
Log email errors with other PHP errors in /logs/ folder

drbyte:
Email errors will now be logged the same as any other PHP error, in the `/logs/` folder.

This could ultimately deprecate the `EMAIL_FRIENDLY_ERRORS` switch as this basically renders it moot.

Ref: https://www.zen-cart.com/showthread.php?219285-sending-order-emails


zcwilt:
:+1:

ajeh:
:+1:

PR #802, opened 2016-01-29
Update Authorize.net Servers Per DRBYTE Post Dated: 7/15/15

bislewl:
https://www.zen-cart.com/showthread.php?217783-Authorize-net-Networking-Change-Summer-2015

### DIFF
--- a/includes/modules/payment/authorizenet_echeck.php
+++ b/includes/modules/payment/authorizenet_echeck.php
@@ -145,7 +145,7 @@ class authorizenet_echeck extends base {
    * @return string
    */
   function javascript_validation() {
-    $js = '  if (payment_value == "' . $this->code . '") {' . "\n" .
+    $js = '  if (payment_value == "' . $this->code . '") {' . "\n" ;
     $js .= '    var echeck_custname = document.checkout_payment.authorizenet_echeck_bank_accountholder.value;' . "\n";
     $js .= '    var echeck_bank_aba = document.checkout_payment.authorizenet_echeck_bank_aba_code.value;' . "\n";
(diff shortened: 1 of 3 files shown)


drbyte:
I specifically undid that in the latest code-base, because those "server2" addresses were temporary until the main servers were also upgraded ... which I believe happened back in late 2015.

drbyte:
(for the record, I do appreciate the contribution though!)

bislewl:
Interesting..
Today (about 2 hours ago) there was an issue with the response timing out. and once I changed to server2 it worked. That's why I applied that. Thank You!

includes/modules/payment/authorizenet_echeck.php
```diff
@@ -145,7 +145,7 @@ function update_status() {
    * @return string
    */
   function javascript_validation() {
-    $js = '  if (payment_value == "' . $this->code . '") {' . "\n" .
```
bislewl: Also it looks like there is an issue here. That I fixed

drbyte: Yes, that should be fixed. Thanks

drbyte:
Oy … timeouts are disconcerting.

I’d rather not depend on something they don’t have documented in their API docs.

I suppose a failover might be beneficial … but again, it’s risky to code to undocumented specs, lest they shut down the `server2` resources.

bislewl:
You may be correct. 
The very odd thing about the timeouts was that the curltester.php came back with a response. AMEX ran fine but Visa/MC would time out... 

Also I know several other processors that use the this/these file and simply change the URL, not sure if that is a good idea to incorporate the ability to change the Server/URL of the module.

PR #820, opened 2016-02-09
Retain search criteria when sort customer results in admin

drbyte:
Ref: https://www.zen-cart.com/showthread.php?219492-admin-customers-php-sortOrderHeader


zcwilt:
:+1:

ajeh:
:+1:

PR #829, opened 2016-02-14
Null dates need to be passed as `null` instead of `'null'`

drbyte:
Null dates need to be passed as `null` instead of `'null'` 

Same with strings.

USAGE NOTE:  For BindVars, for `date` and `string` bindVarTypes, passed values of a string `null` (case-insensitive) will be converted to a literal MySQL `null` instead of treated as a string of the word `'null'`

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -4,11 +4,11 @@
  * Class used for database abstraction to MySQL via mysqli
  *
  * @package classes
- * @copyright Copyright 2003-2015 Zen Cart Development Team
+ * @copyright Copyright 2003-2016 Zen Cart Development Team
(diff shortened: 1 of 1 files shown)


drbyte:
fixes error:
```
PHP Fatal error:  1292:Incorrect datetime value: 'null' :: 
   UPDATE banners
   SET
   date_scheduled = DATE_ADD('null', INTERVAL '00:00:00' HOUR_SECOND),
   expires_date = DATE_ADD('null', INTERVAL '23:59:59' HOUR_SECOND),
   expires_impressions = null
   WHERE banners_id = 9
 ==> (as called by) /admin/banner_manager.php on line 143
```

zcwilt:
:+1:

ajeh:
:+1:

Issue #270, opened 2015-01-10
Review all uses of htmlspecialchars and zen_db_output


drbyte:
I'm pretty sure that all remaining references to `zen_db_output()` should actually be using `zen_output_string_protected()` for consistency.

... and maybe then temporarily move `zen_db_output()` to the compatibility functions file and redefine it to be a call to `zen_output_string_protected()`, for retirement in a future version.

Further, I think a review of all direct calls to `htmlspecialchars()` should be done to replace them with `zen_output_string_protected()` also ... unless there's a specific reason not to.

Credits to @lat9 http://www.zen-cart.com/showthread.php?215798

mc12345678:
Why does admin: zen_output_string_protected use htmlspecials($string, ENT_COMPAT, CHARSET, FALSE) and catalog side instead use htmlspecials($string, ENT_COMPAT, CHARSET, TRUE) (last factor being to only single encode on the admin and double encode on the store side?) Is this part of what has to be modified/corrected in order to simplify across the two systems? A majority of admin calls directly to htmlspecials seem to use TRUE like on the catalog side, but then calls to zen_output_string_protected on the admin side are not automatically "double-encoded".

lat9:
I dunno, but you can follow the "rules" that I documented [here](http://vinosdefrutastropicales.com/blog/?p=307) to make sure that you don't get double-encoding where you don't want it.

FWIW, I've stopped using the `zen_output_string_protected` function because of the discontinuity between the store-front and admin-console operation.

mc12345678:
I was thinking of that or similar as I was going through it and trying to see what "consistent use" existed for both sides as to why one is one way and the other not and what it would take to align the two...  (Afterall it would be great if all/most functions were one type/style right? Rather than multiple examples of code that accomplishes the "same thing"?)  Seems like the way to go would be the catalog side as that has been the more important side to address, so the question becomes what would it take on the admin side actually to not use the currently defined zen_output_string_protected and instead have it work like the catalog side?

Of course thought this was going to be easier to fix/address than it turned into, but now just trying to see if there is a way to map it forward to use one or the other.

scottcwilson:
It seems like the only files left that still do zen_db_output are: 

admin/reviews.php
admin/packingslip.php
admin/orders.php
admin/invoice.php
includes/classes/order.php

PR #831, opened 2016-02-15
Mc12345678 code for #270 - consolidate zen_db_output

mc12345678:
Associated with issue #270 to remove all uses of zen_db_output, and replace with zen_output_string_protected, then to also address usage of htmlspecialchars where it were to output as a protected string and change that to use the zen_output_string_protected function.

Also as a separate commit, modified affected files to use the new database accessor (Issue #406) and also one for moving html code outside of php calls..

### DIFF
--- a/admin/invoice.php
+++ b/admin/invoice.php
@@ -167,7 +167,7 @@ function couponpopupWindow(url) {
 
       if (isset($order->products[$i]['attributes']) && (($k = sizeof($order->products[$i]['attributes'])) > 0)) {
         for ($j = 0; $j < $k; $j++) {
-          echo '<br><nobr><small>&nbsp;<i> - ' . $order->products[$i]['attributes'][$j]['option'] . ': ' . nl2br(zen_output_string_protected($order->products[$i]['attributes'][$j]['value']));
+          echo '<br><nobr><small>&nbsp;<i> - ' . $order->products[$i]['attributes'][$j]['option'] . ': ' . nl2br(htmlspecialchars($order->products[$i]['attributes'][$j]['value'], ENT_COMPAT, CHARSET, FALSE));
           if ($order->products[$i]['attributes'][$j]['price'] != '0') echo ' (' . $order->products[$i]['attributes'][$j]['prefix'] . $currencies->format($order->products[$i]['attributes'][$j]['price'] * $order->products[$i]['qty'], true, $order->info['currency'], $order->info['currency_value']) . ')';
           if ($order->products[$i]['attributes'][$j]['product_attribute_is_free'] == '1' and $order->products[$i]['product_is_free'] == '1') echo TEXT_INFO_ATTRIBUTE_FREE;
           echo '</i></small></nobr>';
--- a/admin/invoice.php
+++ b/admin/invoice.php
@@ -232,7 +232,7 @@ function couponpopupWindow(url) {
         echo '          <tr>' . "\n" .
              '            <td class="smallText" align="center" valign="top">' . zen_datetime_short($orders_history->fields['date_added']) . '</td>' . "\n";
         echo '            <td class="smallText" valign="top">' . $orders_status_array[$orders_history->fields['orders_status_id']] . '</td>' . "\n";
-        echo '            <td class="smallText" valign="top">' . ($orders_history->fields['comments'] == '' ? TEXT_NONE : nl2br(zen_db_output($orders_history->fields['comments']))) . '&nbsp;</td>' . "\n" .
+        echo '            <td class="smallText" valign="top">' . ($orders_history->fields['comments'] == '' ? TEXT_NONE : nl2br(htmlspecialchars($orders_history->fields['comments'], ENT_COMPAT, CHARSET, TRUE))) . '&nbsp;</td>' . "\n" .
              '          </tr>' . "\n";
         $orders_history->MoveNext();
         if (ORDER_COMMENTS_INVOICE == 1 && $count_comments >= 1) {
--- a/admin/invoice.php
+++ b/admin/invoice.php
@@ -232,7 +232,7 @@ function couponpopupWindow(url) {
         echo '          <tr>' . "\n" .
              '            <td class="smallText" align="center" valign="top">' . zen_datetime_short($orders_history->fields['date_added']) . '</td>' . "\n";
         echo '            <td class="smallText" valign="top">' . $orders_status_array[$orders_history->fields['orders_status_id']] . '</td>' . "\n";
-        echo '            <td class="smallText" valign="top">' . ($orders_history->fields['comments'] == '' ? TEXT_NONE : nl2br(htmlspecialchars($orders_history->fields['comments'], ENT_COMPAT, CHARSET, TRUE))) . '&nbsp;</td>' . "\n" .
+        echo '            <td class="smallText" valign="top">' . ($orders_history->fields['comments'] == '' ? TEXT_NONE : nl2br(zen_output_string_protected($orders_history->fields['comments']))) . '&nbsp;</td>' . "\n" .
              '          </tr>' . "\n";
         $orders_history->MoveNext();
         if (ORDER_COMMENTS_INVOICE == 1 && $count_comments >= 1) {
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -685,7 +685,7 @@ function couponpopupWindow(url) {
           echo zen_image(DIR_WS_ICONS . 'unlocked.gif', TEXT_VISIBLE) . "</td>\n";
         }
         echo '            <td class="smallText">' . $orders_status_array[$orders_history->fields['orders_status_id']] . '</td>' . "\n";
-        echo '            <td class="smallText">' . nl2br(zen_db_output($orders_history->fields['comments'])) . '&nbsp;</td>' . "\n" .
+        echo '            <td class="smallText">' . nl2br(htmlspecialchars($orders_history->fields['comments'], ENT_COMPAT, CHARSET, TRUE)) . '&nbsp;</td>' . "\n" .
              '          </tr>' . "\n";
         $orders_history->MoveNext();
       }
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -607,7 +607,7 @@ function couponpopupWindow(url) {
 
       if (isset($order->products[$i]['attributes']) && (sizeof($order->products[$i]['attributes']) > 0)) {
         for ($j = 0, $k = sizeof($order->products[$i]['attributes']); $j < $k; $j++) {
-          echo '<br /><nobr><small>&nbsp;<i> - ' . $order->products[$i]['attributes'][$j]['option'] . ': ' . nl2br(zen_output_string_protected($order->products[$i]['attributes'][$j]['value']));
+          echo '<br /><nobr><small>&nbsp;<i> - ' . $order->products[$i]['attributes'][$j]['option'] . ': ' . nl2br(htmlspecialchars($order->products[$i]['attributes'][$j]['value'], ENT_COMPAT, CHARSET, FALSE));
           if ($order->products[$i]['attributes'][$j]['price'] != '0') echo ' (' . $order->products[$i]['attributes'][$j]['prefix'] . $currencies->format($order->products[$i]['attributes'][$j]['price'] * $order->products[$i]['qty'], true, $order->info['currency'], $order->info['currency_value']) . ')';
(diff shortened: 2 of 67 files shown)


drbyte:
While I appreciate the efforts of refactoring the `while(!EOF)` to `foreach`, and also the cleaning of HTML-vs-PHP-echoes ... it's important to make PRs more atomic when possible ... ie: fixing one issue instead of mixing multiples.

I realize it's tough to separate them out because the same files are touched in all three different things you've done.

But it's a lot easier to track down future issues if things are more atomic. And also easier to point someone to a PR that shows the changes specific to the issue being addressed. 

Can you split this into 3 PRs instead?

mc12345678:
Absolutely. Didn't want to touch a file to modify it in part and omit other "known" modifications needed for it... Probably could make it cleaner anyways as things seemed to start to fall apart on me at one point...

Was actually going to go the "easy" route and just touch the issue of #270 but thought that I would be doing the code wrong by not going as far/further with it. Got too caught up in that aspect.

drbyte:
> Got too caught up in that aspect.

Not to worry. Sometimes I do too. :)

drbyte:
Note: if you're familiar with `git rebase -i` then you can alter the commits you've made in this branch, and then run `git push --force` to send the revised branch back up to github.

Make a copy of the all the files first in case you want to refer to them later, unless you want to redo the work from scratch.

mc12345678:
'unless you want to redo the work from scratch.' <- H#!! No.. :) I did figure I'd have to make a local mess to fix the public branch and set out the other branches to address the other "issues". 

I've been wanting to practice a little more related to some of the rebase commands, but always when I like what I have as an end result with a less than desirable path. But I like games like tower of hanoii, so all good. :) (once I get to a computer.)

drbyte:
:+1:

zcwilt:
:+1:

admin/orders.php
```diff
@@ -685,7 +685,7 @@ function couponpopupWindow(url) {
           echo zen_image(DIR_WS_ICONS . 'unlocked.gif', TEXT_VISIBLE) . "</td>\n";
         }
         echo '            <td class="smallText">' . $orders_status_array[$orders_history->fields['orders_status_id']] . '</td>' . "\n";
-        echo '            <td class="smallText">' . nl2br(zen_db_output($orders_history->fields['comments'])) . '&nbsp;</td>' . "\n" .
+        echo '            <td class="smallText">' . nl2br(zen_output_string_protected($orders_history->fields['comments'])) . '&nbsp;</td>' . "\n" .
```
mc12345678: Just realized (perhaps a second time) that this substitution will provide inaccurate results in the current arrangement.  zen_db_output forces htmlspecialchars($var, ENT_COMPAT, CHARSET, TRUE) while the admin zen_output_string_protected forces htmlspecialchars($var, ENT_COMPAT, CHARSET, FALSE)... Thus, with this substitution (on the admin side at this time), the $var will not be "double" processed.. (I.e. &amp; will not become &amp;amp;)  This throws a spanner in the works, although could be instead hardcoded to be htmlspecialchars($var, ENT_COMPAT, CHARSET, TRUE);, then go through all of the admin code and basically swap the two sets of declarations because the FALSE version is used far less than the TRUE version seems to be... :) Then can figure out what it takes to convert the remaining FALSE versions instead... Just a thought...

admin/packingslip.php
```diff
@@ -185,7 +185,7 @@
         echo '          <tr>' . "\n" .
              '            <td class="smallText" align="center" valign="top">' . zen_datetime_short($orders_history->fields['date_added']) . '</td>' . "\n";
         echo '            <td class="smallText" valign="top">' . $orders_status_array[$orders_history->fields['orders_status_id']] . '</td>' . "\n";
-        echo '            <td class="smallText" valign="top">' . ($orders_history->fields['comments'] == '' ? TEXT_NONE : nl2br(zen_db_output($orders_history->fields['comments']))) . '&nbsp;</td>' . "\n" .
+        echo '            <td class="smallText" valign="top">' . ($orders_history->fields['comments'] == '' ? TEXT_NONE : nl2br(zen_output_string_protected($orders_history->fields['comments']))) . '&nbsp;</td>' . "\n" .
```
mc12345678: Just realized (perhaps a second time) that this substitution will provide inaccurate results in the current arrangement.  zen_db_output forces htmlspecialchars($var, ENT_COMPAT, CHARSET, TRUE) while the admin zen_output_string_protected forces htmlspecialchars($var, ENT_COMPAT, CHARSET, FALSE)... Thus, with this substitution (on the admin side at this time), the $var will not be "double" processed.. (I.e. &amp; will not become &amp;amp;)  This throws a spanner in the works, although could be instead hardcoded to be htmlspecialchars($var, ENT_COMPAT, CHARSET, TRUE);, then go through all of the admin code and basically swap the two sets of declarations because the FALSE version is used far less than the TRUE version seems to be... :) Then can figure out what it takes to convert the remaining FALSE versions instead... Just a thought...

admin/reviews.php
```diff
@@ -190,7 +190,7 @@
       <tr>
         <td><table width="100%" border="0" cellspacing="0" cellpadding="0">
           <tr>
-            <td valign="top" class="main"><b><?php echo ENTRY_REVIEW; ?></b><br><br><?php echo nl2br(zen_db_output(zen_break_string($rInfo->reviews_text, 15))); ?></td>
+            <td valign="top" class="main"><b><?php echo ENTRY_REVIEW; ?></b><br><br><?php echo nl2br(zen_output_string_protected(zen_break_string($rInfo->reviews_text, 15))); ?></td>
```
mc12345678: Just realized (perhaps a second time) that this substitution will provide inaccurate results in the current arrangement.  zen_db_output forces htmlspecialchars($var, ENT_COMPAT, CHARSET, TRUE) while the admin zen_output_string_protected forces htmlspecialchars($var, ENT_COMPAT, CHARSET, FALSE)... Thus, with this substitution (on the admin side at this time), the $var will not be "double" processed.. (I.e. &amp; will not become &amp;amp;)  This throws a spanner in the works, although could be instead hardcoded to be htmlspecialchars($var, ENT_COMPAT, CHARSET, TRUE);, then go through all of the admin code and basically swap the two sets of declarations because the FALSE version is used far less than the TRUE version seems to be... :) Then can figure out what it takes to convert the remaining FALSE versions instead... Just a thought...

admin/invoice.php
```diff
@@ -232,7 +232,7 @@ function couponpopupWindow(url) {
         echo '          <tr>' . "\n" .
              '            <td class="smallText" align="center" valign="top">' . zen_datetime_short($orders_history->fields['date_added']) . '</td>' . "\n";
         echo '            <td class="smallText" valign="top">' . $orders_status_array[$orders_history->fields['orders_status_id']] . '</td>' . "\n";
-        echo '            <td class="smallText" valign="top">' . ($orders_history->fields['comments'] == '' ? TEXT_NONE : nl2br(zen_db_output($orders_history->fields['comments']))) . '&nbsp;</td>' . "\n" .
+        echo '            <td class="smallText" valign="top">' . ($orders_history->fields['comments'] == '' ? TEXT_NONE : nl2br(zen_output_string_protected($orders_history->fields['comments']))) . '&nbsp;</td>' . "\n" .
```
mc12345678: Just realized (perhaps a second time) that this substitution will provide inaccurate results in the current arrangement.  zen_db_output forces htmlspecialchars($var, ENT_COMPAT, CHARSET, TRUE) while the admin zen_output_string_protected forces htmlspecialchars($var, ENT_COMPAT, CHARSET, FALSE)... Thus, with this substitution (on the admin side at this time), the $var will not be "double" processed.. (I.e. &amp; will not become &amp;amp;)  This throws a spanner in the works, although could be instead hardcoded to be htmlspecialchars($var, ENT_COMPAT, CHARSET, TRUE);, then go through all of the admin code and basically swap the two sets of declarations because the FALSE version is used far less than the TRUE version seems to be... :) Then can figure out what it takes to convert the remaining FALSE versions instead... Just a thought...

drbyte:
I was thinking that the use of TRUE was preferable.

mc12345678:
> I was thinking that the use of TRUE was preferable.

Hmm... How to respond without becoming unnecessarily confusing... (normally it's just necessary. :) )
On the catalog side, zen_output_string_protected does permit double conversion and I forgot in my previous example that the html I provided would be converted by the github system... on the catalog side if `zen_output_string_protected('&amp;x=0')` is used then the result is: `&amp;amp;x=0` 
On the admin side `zen_output_string_protected('&amp;x=0')` will result in `&amp;x=0`.  

zen_db_output on the admin side works the same as the catalog side zen_output_string_protected.

(Somewhat extremely stated:) While it is/would be easy to simply change the letters in the code to call the named function and the named function to operate the same on both sides, that would/could have an impact on operation.  Would prefer to know, understand and address the known user's of the admin code so that when it is changed to double_encode like it does on the catalog side that at least the results of the core code doesn't go sideways.

Preferred to double_encode (use TRUE instead of FALSE), suspected that was the preference, now it is a matter of making it happen everywhere desired/expected.  Had considered creating (perhaps temporarily) an admin function like: zen_catalog_output_string_protected so that all of the applicable htmlspecialchars references could be converted to that or similar and the remaining other uses of zen_db_output and htmlspecialchars could be worked out, then zen_catalog_output_string_protected replaced with zen_output_string_protected... My overall thought though is, best done on my local version and not to do so through a commit...  End result maybe, but not the interim work, else someone may develop code off of the interim change(s).

mc12345678:
Okay, I think I have a plan finally that I've documented below so I don't have to remember it. Been thinking about how to approach this to remove zen_db_output, reduce the number of occurrences of the function definition for zen_output_string_protected (or at least consistently across catalog and admin), and not affect the results of current usage of both in the core.

1. Explode current admin zen_db_output usages to the htmlspecialchars equivalent (ending in TRUE).
2. Explode current admin usage of zen_output_string_protected to htmlspecialchars equivalent (ending in FALSE)
3. Explode current use of admin zen_output_string with protected = true to htmlspecialchars equivalent (ending in FALSE)
4. Verify/reverify that all uses of admin htmlspecialchars use all four fields/parameters.
5. Redefine admin zen_output_string_protected to use htmlspecialchars with TRUE.
6. Redefine all admin htmlspecialchars that end with TRUE to use the new zen_output_string_protected.
7. Square away the zen_output_string_protected function.
8. Set aside a small amount of time to do the above. :)

mc12345678:
Incorporated the removal of zen_db_output from both catalog and admin:
End result is that on admin, zen_db_output no longer exists in the 
base code (compatibility), zen_output_string_protected now double encodes
all uses of htmlspecialchars, zen_output_string_protected, and zen_db_output
have been modified to account for this switch.

mc12345678:
> I like the consolidation. What testing have you done about double-encoding?
I had not intended to change the existing encoding (at the time), but instead to simply perform the replacement, though it soon became clear that there were differences for which I had not yet accounted.  

Now the code has been "inverted" so that the zen_output_string_protected works as "expected", but there remain numerous occurrences of htmlspecialchars($var, ENT_COMPAT, CHARSET, FALSE) (79 occurrences across the entire store at time of above commit and those are all on the admin side).  In above modifications, did not focus fully on the catalog side with all of the "checks", so there may still be some occurrences of htmlspecialchars that are not fully qualified to use all four parameters or all four of the same parameters.

mc12345678:
Just realized that this PR is not fully atomic.... In order to accomplish:
> I'm pretty sure that all remaining references to zen_db_output() should actually be using zen_output_string_protected() for consistency.
> 
> ... and maybe then temporarily move zen_db_output() to the compatibility functions file and redefine it to be a call to zen_output_string_protected(), for retirement in a future version.

To accomplish the above goal outlined in issue #270, the above mixed goal PR was submitted. It does deprecate zen_db_output, but also changes the functionality of zen_output_string_protected also primarily to allow the deprecated version of zen_db_output to reference the existing zen_output_string_protected function. This aspect could be simplified to have zen_db_output simply perform the htmlspecialchars with TRUE and thus the deprecation of zen_db_output could be atomic and leave the refactoring of the admin side zen_output_string_protected to a separate PR... Let me know, and I can split the two out and carry on with the other two PRs described above...

drbyte:
We're looking at refactoring the admin pages fairly significantly to extract a bunch of business logic, and rewrite how the outputs happen. So, gonna close this for now.  

Many thanks for the research ... as the information discussed here helps inform the pending improvements ;)

mc12345678:
More than welcome. Trying to do what I can to realize ZC 1.6.0. :) Goals have been set, was an intriguing issue, and "appeared" easier than it turned out. Sounds like some change for the good is still expected, so yeah! :)

PR #833, opened 2016-02-16
Add error reporting to zc_install for problems writing configure.php files

drbyte:
Add meaningful error messages on-screen if config files couldn't be written properly.


ajeh:
:+1:

Issue #60, opened 2014-07-08
211797: Add notifier to specials, salemaker, and featured product functions


zcwilt:
Original discussion here.
http://www.zen-cart.com/showthread.php?211797-Add-notifier-to-specials-salemaker-and-featured-product-functions

PR #834, opened 2016-02-16
Additional notifier hooks (ref #60)

drbyte:
Additional notifier hooks, as discussed in https://www.zen-cart.com/showthread.php?211797&p=1256923#post1256923


ajeh:
:+1:

PR #835, opened 2016-02-17
Add zen draw label function

bislewl:


zcwilt:
:+1:

drbyte:
I'm okay with the addition of this, but not interested in embracing its use in v155 because it will cause backward compatibility issues with plugins (ie: a plugin upgrades to be based on v155 but then no longer works on v154 because v154 is missing this function)

I'd encourage more use of it starting with v160 though.

Zen4All:
If there were objections, why did this get pushed into 155?

bislewl:
IMHO. Dev's who use the function zen_draw_label() would know it isn't backwards compatible and would probably do they same thing they have done with the plugin_version_check_for_updates() function.

drbyte:
> If there were objections, why did this get pushed into 155?

Not objecting to its introduction. 
Just not wanting to go wild rewriting things in 155 to use it.

Issue #836, opened 2016-02-18
Remove closing php tags for edited files


mc12345678:
Didn't find an associated issue so created one.  As part of other code changes being made, I am trying to address the files in whole rather than just the one part that initiated a change to the file.  As part of that and previously stated goals in the forum, this issue is to "track" the removal of the closing tag of affected files.

drbyte:
> I am trying to address the files in whole rather than just the one part that initiated a change to the file.

Fair, but do keep in mind that commits (and ideally Pull Requests too) should be atomic with respect to a specific change being implemented. This is primarily so that one can look back at PRs to understand specifically what was changed and why. When a file is refactored in numerous ways in one PR then that destroys the ability to follow the bouncing ball.

drbyte:
Not sure this needs to be an issue ticket here.

PR #839, opened 2016-02-22
Remove endtags on changed files of Issue #270, deprecate zen_db_output

mc12345678:
removed the endtags from the files changed in support of Issue #270 (updated by PR #831) and in support of issue #836.  This PR is based off of the end of that commit.  If that commit is not accepted, then I would discourage this one from being accepted as well..

Although there were files in the admin directory that were affected, review of those files showed too much consistency of ending with inclusion of application_bottom.php enclosed by php tags, therefore didn't modify those files in the event that there was a reason all to end with the closing php tag.

But for files that were "internal", the php closing tag was removed as incorporated in this PR.


drbyte:
We're looking at refactoring the admin pages fairly significantly to extract a bunch of business logic, and rewrite how the outputs happen. So, gonna close this for now.

PR #841, opened 2016-02-25
Add @torvista's suggestion for auto-assigning max+1 sort order

drbyte:
Add @torvista suggestion for auto-assigning max+1 sort order

Ref: https://www.zen-cart.com/showthread.php?199251&p=1304539#post1304539


zcwilt:
:+1:

ajeh:
:+1:

PR #849, opened 2016-02-27
Add notifier hooks to make for easier integration of USU plugin

drbyte:


zcwilt:
:+1:

ajeh:
:+1:

drbyte:
IMO this should make it easier for @lhungil for future updates to the USU mod.

PR #854, opened 2016-02-28
Affix admin nav menu to top of page; add odd/even rows on widgets

yama:
Thanks.


zcwilt:
:+1: 
A note, while I like the scroll fixing of the menu bar, it's quite likely we will move to a different menu system in v160. However we can still use the same idea :)

drbyte:
:+1:

PR #857, opened 2016-03-03
Rewrite the handful of queries using 0 to trigger auto-incrementing

drbyte:
Instead of relying on passing 0 to the auto-increment field name, we specifically avoid mentioning it, which causes MySQL to automatically trigger the auto-incrementing rules.

This should help avoid conflicts in cases of odd MySQL configurations such as `NO_AUTO_VALUE_ON_ZERO` in my.cnf

### DIFF
--- a/admin/attributes_controller.php
+++ b/admin/attributes_controller.php
@@ -315,9 +315,8 @@
               $attributes_image_name = $current_image_name;
             }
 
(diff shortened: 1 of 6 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #858, opened 2016-03-03
Serve Admin assets locally for bootstrap, jquery and font-awesome

drbyte:
Serve admin assets locally (instead of via CDN) for both performance and version-proofing benefits.
This also provides the ability for a site maintainer to work offline on a laptop etc, not requiring an internet connection.
Further, it also avoids cases where a chosen CDN might be blocked for various reasons.


zcwilt:
:+1:

ajeh:
:+1:

PR #861, opened 2016-03-04
Clean comments from transaction id field before storing to db

drbyte:
Can't have string stored in bigint field, so remove it before storing, to avoid a db error.

Thanks to @numinix for pointing out the problem.


zcwilt:
:+1:

ajeh:
:+1:

PR #867, opened 2016-03-13
Make Admin messageStack use FontAwesome icons and bootstrap styling

drbyte:
Icons weren't showing correctly, and using FontAwesome is the intended future direction, so makes sense to fix it this way.

Also adding Bootstrap styling to messageStack alerts makes for better consistency.


zcwilt:
:+1:

ajeh:
:+1:

PR #874, opened 2016-03-16
Add CURL fallbacks to support hosting companies that can't do security right

drbyte:
Some hosts don't yet support proper auto negotiation for modern TLS protocols, so this allows a timeout and then tries again by specifying a hard-coded protocol after logging the problem in the myDebug-xxxxx.log logs.


zcwilt:
:+1:

ajeh:
:+1:

PR #880, opened 2016-03-18
Backport fixes for MySQL no-zero-date mode to older zc_install upgrade steps

drbyte:
Fixes chicken/egg problem

Affects upgrades from 150-to-151, and 152-to-153, and already handled 154-to-155


ajeh:
:+1:

PR #887, opened 2016-04-01
Add missing directories

Zen4All:
Since introducing the shared folder, they should also be made searchable
in the developers toolkit.

I added paths for the shared folders where the code is looking for
$template_dir.
I also added some missing paths in the language directories

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/887)


zcwilt:
:+1:

ajeh:
:+1:

PR #889, opened 2016-04-06
Update/add .gitignore to help manage assets that shouldn't be in VCS

drbyte:
Placing `.gitignore` files in relevant subfolders allows git to have more granular control over what gets kept in VCS

This allows, for example, generated/cached images for banner-stats, and session/cache files, etc, to not get put into git unnecessarily.


zcwilt:
:+1:

ajeh:
:+1:

PR #890, opened 2016-04-06
Add Nginx .htaccess Equivalents. Fixes #667

### LINKED ISSUE #667: Add Nginx equivalents to built-in .htaccess files
While Zen Cart runs fine on Nginx, all the added security protections supplied in .htaccess files for Apache have no effect in Nginx.

TO DO: Write up instructions, or alternate nginx config file(s), to supply similar protections to disallow access to inappropriate folders, and to enable appropriate file-caching headers to various assets to improve performance.

dakanji:


drbyte:
FYI - Gonna set up a server for testing this out, before merging.

dakanji:
Makes sense!
>
>
FYI. The CLA email ultimately came back as being undeliverable.

dakanji:
Got my git stuff mixed up a bit but finally got there to revert the added slashes and fix a typo as well as let users know they can use the server for SSL and/or Non SSL Server blocks.

drbyte:
Received CLA from @dakanji

drbyte:
I wonder if your references to `%%http_store_folder%%/` should have the closing `/` removed. 

As it is right now I'm getting double slashes. Here's just a segment as an example, but it's throughout the file

```
## Returns '404 Not Found' http code for external requests
	location ^~ //includes/classes {
		internal;
	}
## Allow all 'GET' and 'HEAD' requests to specific static file types in 'includes'
	location ~ //includes(.+)$ {
		if ( $request_method !~ ^(GET|HEAD)$ ) {
			return 403;
		}
		location ~ //includes(.+)\.(ico|jpe?g|gif|otf|webp|png|swf|flv|pdf|svg?z)$ {
			add_header ETag '';
			add_header Pragma '';
			add_header Last-Modified '';
			add_header Cache-Control 'max-age=864000, public, must-revalidate';
			
			# Optional caching improvement
			# Uncomment if accompanying 'map' block in HTTP section has been added
			#expires $zencart_expires;		
		}
		location ~ //includes/(.+)\.(html|htm|xml|txt|xsl)$ {
			add_header ETag '';
```

Granted, this install was with a VM configured to serve ZC from the root.

drbyte:
Additionally, while editing the various files it would be preferable to refer to `Zen Cart` as two words, instead of all the other forms of "Zencart", "ZENCART", "zencart"  ;)

drbyte:
Is the section for `zc_install` needed? ... ie: this information isn't presented until after zc_install has run, so including directives for it seems kinda moot.

zc_install/includes/nginx_conf/ngx_conf_server.txt
```diff
@@ -0,0 +1,349 @@
...
+		return 418;
+	}
+
+	
+###############################
+# Zencart Catchall
+###############################	
+## Redirect any request not previously explicitly handled to main Zencart index.php file
+	location ~ %%http_store_folder%%(.*)$ {
+		return %%http_store_folder%%/index.php;
```
drbyte: This `return` is coming back as invalid.

But changing this return line to a `try` seems to work: `try_files $uri $uri/ /index.php?$args ;`  (and similarly with the one for admin/index.php 26 lines earlier).

drbyte:
(the merge of v155 code into the main v160 branch has caused a merge conflict with this PR ... but I'm pretty sure it's minor)

dakanji:
> I wonder if your references to %%http_store_folder%%/ should have the closing / removed.

Luckily, Nginx collapses double slashes so that it does not present an issue but will have a look at eliminating this.

> Additionally, while editing the various files it would be preferable to refer to Zen Cart as two words

Noted

> Is the section for zc_install needed?

Without these directives, users will not be able to upgrade/reinstall afterwards.

> This return is coming back as invalid.

I see why. `try_files` is not a good idea for either as it will unlock things. Will fix accordingly.

>  the merge of v155 code into the main v160 branch has caused a merge conflict

Will resolve the conflicts.

drbyte:
I was looking at [this article about (somewhat) similar configuration](https://www.nginx.com/resources/wiki/start/topics/recipes/oscommerce/) but which includes URL rewrite options (assuming there was code installed such as the Ultimate URLs plugin).  Do you mind including some comments in the template about where one should insert such rewrite rules if they were to incorporate a plugin that was using URL formats such as these apache rewrites would do:

```
#Apache Rewrite scenarios:
RewriteRule ^(.*)-c-([0-9_]+)/(.*)-p-([0-9]+)(.*)$ index\.php?main_page=product_info&products_id=$4&cPath=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-c-([0-9_]+)/(.*)-pi-([0-9]+)(.*)$ index\.php?main_page=popup_image&pID=$4&cPath=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-c-([0-9_]+)/(.*)-pr-([0-9]+)(.*)$ index\.php?main_page=product_reviews&products_id=$4&cPath=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-c-([0-9_]+)/(.*)-pri-([0-9]+)(.*)$ index\.php?main_page=product_reviews_info&products_id=$4&cPath=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-p-([0-9]+)(.*)$ index\.php?main_page=product_info&products_id=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-m-([0-9]+)(.*)$ index\.php?main_page=index&manufacturers_id=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-pi-([0-9]+)(.*)$ index\.php?main_page=popup_image&pID=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-pr-([0-9]+)(.*)$ index\.php?main_page=product_reviews&products_id=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-pri-([0-9]+)(.*)$ index\.php?main_page=product_reviews_info&products_id=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-ezp-([0-9]+)(.*)$ index\.php?main_page=page&id=$2&%{QUERY_STRING} [L]
RewriteRule ^(.*)-c-([0-9_]+)(.*)$ index\.php?main_page=index&cPath=$2&%{QUERY_STRING} [L]
```

I'm guessing these would be the nginx equivalents. But where's best to insert them?
```
    location ~ -c-(?<cpath>[0-9_]+)/(.*)-p-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?main_page=product_info&products_id=$id&cPath=$cpath; }
    location ~ -c-(?<cpath>[0-9_]+)/(.*)-pi-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?main_page=popup_image&pID=$id&cPath=$cpath; }
    location ~ -c-(?<cpath>[0-9_]+)/(.*)-pr-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?main_page=product_reviews&products_id=$id&cPath=$cpath; }
    location ~ -c-(?<cpath>[0-9_]+)/(.*)-pri-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?main_page=product_reviews_info&products_id=$id&cPath=$cpath; }
    location ~ -p-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?main_page=product_info&products_id=$id; }
    location ~ -m-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?manufacturers_id=$id; }
    location ~ -pi-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?main_page=popup_image&pID=$id; }
    location ~ -pr-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?main_page=product_reviews&products_id=$id; }
    location ~ -pri-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?main_page=product_reviews_info&products_id=$id; }
    location ~ -ezp-(?<id>[0-9]+)(.*) { rewrite ^ /index.php?main_page=page&id=$id; }
    location ~ -c-(?<id>[0-9_]+)(.*) { rewrite ^ /index.php?cPath=$id; }
```

dakanji:
Wouldn't the instructions for a 3rd party plugin (I assume this is one) be better placed in the installation instructions for that plugin?

If the plugin is an official one distributed with a default installation and supported by the ZC Team (or planned to be), then the directives can just go in now. In which case the location blocks should go in just before the "catchall" location block or more succinctly, the catchall location block can include the following rewrite directives directly converted from the Apache version:

```
rewrite ^(.*)-c-([0-9_]+)/(.*)-p-([0-9]+)(.*)$ index.php?main_page=product_info&products_id=$4&cPath=$2 last;
rewrite ^(.*)-c-([0-9_]+)/(.*)-pi-([0-9]+)(.*)$ index.php?main_page=popup_image&pID=$4&cPath=$2 last;
rewrite ^(.*)-c-([0-9_]+)/(.*)-pr-([0-9]+)(.*)$ index.php?main_page=product_reviews&products_id=$4&cPath=$2 last;
rewrite ^(.*)-c-([0-9_]+)/(.*)-pri-([0-9]+)(.*)$ index.php?main_page=product_reviews_info&products_id=$4&cPath=$2 last;
rewrite ^(.*)-c-([0-9_]+)(.*)$ index.php?main_page=index&cPath=$2 last;
rewrite ^(.*)-p-([0-9]+)(.*)$ index.php?main_page=product_info&products_id=$2 last;
rewrite ^(.*)-m-([0-9]+)(.*)$ index.php?main_page=index&manufacturers_id=$2 last;
rewrite ^(.*)-pi-([0-9]+)(.*)$ index.php?main_page=popup_image&pID=$2 last;
rewrite ^(.*)-pr-([0-9]+)(.*)$ index.php?main_page=product_reviews&products_id=$2 last;
rewrite ^(.*)-pri-([0-9]+)(.*)$ index.php?main_page=product_reviews_info&products_id=$2 last;
rewrite ^(.*)-ezp-([0-9]+)(.*)$ index.php?main_page=page&id=$2 last;
```
Nginx will add the query strings if they exist so no need to account for `%{QUERY_STRING}`.

I'll go for having the rewrites in the catchall block if distributing by default but prefer the individual location blocks if asking users to insert in themselves. As said though, I do think that if something users would be doing if installing the plugin as a 3rd party item, the instructions for this would probably be better with the plugin installation itself.

Thoughts?

dakanji:
FYI. Completed the changes and only item pending is the Ultimate URLs plugin thing. Please advise how you want to proceed with this. I suggest leaving instructions on this this as an issue for the plugin installation (See my previous comment). Kindly confirm so I can push the changes.

drbyte:
I'd put the rewrite stuff into the template, commented-out, as that makes it clear where it should go and easy to enable if desired.

dakanji:
Okay. 

On second thoughts about the positioning, it would be best if someone has specifically installed such a plugin for pretty URLs, that the best configuration would be to use location blocks such as you proposed located right at the start of the set of directives, perhaps just after the php handler to give a fractional speed boost as they would be hit and processed without running through the other directives. 

Will add an appropriate placeholder there.

drbyte:
Ya, that seems sensible to me.

dakanji:
Should be good to go

drbyte:
Okay, thanks!

Will run it through a very basic test on a basic nginx-equipped VM.

Are these changes something you're using "live" on a Zen Cart site? Or is it more black-box?

zc_install/includes/nginx_conf/ngx_conf_server.txt
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+
+	#############################################################################################
+	# ZEN CART 'includes' Folder
+	#############################################################################################	
+	## Allow and fastrack internal requests to 'includes/classes'
+	location ^~ %%store_folder%%/includes/classes {
+		internal;
+	}
```
drbyte: Why the need to be explicit about this `classes` folder? 

Why `classes` and nothing else? Why `classes` at all?

zc_install/includes/nginx_conf/ngx_conf_http.txt
```diff
@@ -0,0 +1,34 @@
+##############################################################
+##############################################################
+##############################################################
+### 		NGINX CONFIGURATION FOR ZEN CART
+###
+### Copy and paste into ‘http’ section of nginx conf file
+### IMPORTANT: 
+###    - Enables optional caching in serve section
+###    - Remember to uncomment relevant lines in server section
```
drbyte: It seems you've removed the `server` section lines to be `uncommented`. Was that intentional? Or should this line be removed if there's no longer a need to uncomment anything?

zc_install/includes/nginx_conf/ngx_conf_http.txt
```diff
@@ -0,0 +1,34 @@
...
+###
+###
+##############################################################
+##############################################################
+##############################################################
+
+
+##############################################################
+# ‘Expires’ map
+# Copy and paste into ‘http’ section of nginx con file
```
drbyte: `nginx con file` ... should that be `conf`?

zc_install/includes/nginx_conf/ngx_conf_http.txt
```diff
@@ -0,0 +1,34 @@
...
+	text/x-javascript		1h;
+	text/css			1h;
+	text/html			1h;
+	image/gif			7d;
+	video/x-flv			7d;
+	application/pdf			7d;
+	image/x-icon			7d;
+	image/jpeg			7d;
+	image/png			7d;
+	text/cache-manifest		2s;
```
drbyte: Seems to be missing the closing `}` to the `map` section.

dakanji:
> Why the need to be explicit about the classes folder?

Good point. Was going to lock down access to PHP files in a fine grained manner but stopped with the classes folder. That location block can be removed without issue I suppose.

> a) Line 9 here talks about uncommenting things ... Was that intentional?

Should now be removed as it seemed better to give simple directions to users.

> b) "Enables optional caching in serve section" ... shouldn't that be server ?

Yes it should. Actually should be amended since the caching is default

> Line 19: nginx con file ... should that be cons?

Assume you meant **`conf`**. Yes, it should.

> Actually, isn't this a duplicate of line 6 above anyway?

Yep. Added the big header afterwards. Probably better without the big header.

> Seems to be missing the closing } to the map section.

Face Palm!

> Are these changes something you're using "live" on a Zen Cart site? Or is it more black-box?

Haven't run the last few changes myself as my test setup is actually slightly different. Will make the edits and actually spin up a specific test installation for this before pushing again.

Thanks for the review.

dakanji:
Seems that the recent merging in of v155 stuff in e71b5594b2b73769ddfea78a41756a3df96ec317 has messed up a few things. 

First thing I noticed was that login.php was not styled and I thought it was the directives messing things up until I saw it was looking for  `includes/stylesheet.css` which doesn't exist in v160.

Suppose some clean up of that merge in is in the works. Testing against the version before the merge for now.

drbyte:
Yes, there are a couple admin-side things that are broken at the moment. A fix is in progress.

dakanji:
Okay … testing against older working version.

drbyte:
Those admin things have been fixed.

But now on nginx the use of font-awesome icons in the Admin is broken ... getting 403 errors when trying to load the font files.

zc_install/includes/nginx_conf/ngx_conf_server.txt
```diff
@@ -0,0 +1,320 @@
...
+	# ZEN CART ADMIN Folder
+	#############################################################################################	
+	## Allow all 'POST', 'GET' and 'HEAD' requests to all PHP files
+	location ~ %%admin_folder%%/(.+)\.php$ {
+		if ( $request_method !~ ^(GET|HEAD|POST)$ ) { return 403; }
+		error_page 418 = @zencart_php_handler;
+		return 418;
+	}
+	## Allow all 'GET' and 'HEAD' requests to specific static file types
+	location ~ %%admin_folder%%(.*)$ {
```
drbyte: Should line 272 have a `/` before the `(.*)` ... like the other directives in this section do?

dakanji:
> Those admin things have been fixed.

Will merge in

> ...the use of font-awesome icons in the Admin is broken ... getting 403 errors when trying to load the font files.

Most likely the path has changed. I had to make a change for them to work when I changed the directives  from working with from v155, which I started with, to v160. Will revert this now.

> Should line 272 have a / before the (.*) ... like the other directives in this section do?

That is intentional. Same structure for the 'includes' section as well. Basically those with nested location blocks which are the two broad sections, 'Catalog' AKA 'includes' folder under root, and 'Admin'.

dakanji:
> Should line 272 have a / before the (.*) ... like the other directives in this section do?

Hmmm. I can see not having the slash would be an issue for your set up where the root folder is just "/".
Need to figure out how to sort that out

> font-awesome icons in the Admin is broken.

Seems not to be a path thing but because `eot` files were not permitted.

drbyte:
The font-awesome icons thing also appears to be aggravated by the fact that they're being loaded with a version number attached, and the regex isn't accounting for that.  

ie: `/admin/includes/template/fonts/fontawesome-webfont.svg?v=4.5.0#fontawesomeregular`

I haven't looked to see if there's a directive/rule for matching on the filename instead of the URI, but wouldn't that be more effective?  Or am I thinking with my PHP brain too much ;)

dakanji:
Arguments don’t play a part in Nginx location matching so everything after the “?” has no effect for location blocks.
Haven’t had a proper look at this today yet but will do shortly. Will be resolved either way as it is trivial.

drbyte:
> Arguments don’t play a part in Nginx location matching so everything after the “?” has no effect for location blocks.

Thanks for clarifying. Nginx is new territory for me, but I'm certainly learning a lot!  Thanks for your input in all this.

dakanji:
Before I push the updates, just curious as to the font awesome issue as it seems to be fine here without any changes apart from adding "eot" files to the regex.

Only place I can see them used so far is the trash in the pages styled with Foundation such as `index.php?cmd=languages` and it is loading fine for me.

dakanji:
Okay, the svg example you gave does not match a location because the location regex had a error.
The regex had `svg?z` instead of `svgz?`. I.E., the "g" was optional instead of the "z" as was intended.

Will wrap up and push for review.

drbyte:
Thanks. I'll review this later today.

The fontawesome icons I was testing for are on the payment modules page. The last column of the table contains some "action" icons. One is (an optional) lightning-bolt indicator for testing communication capability for certain payment modules, and the other is a line-indicator/info button. None of these are loading for me. Getting 403 errors instead.

dakanji:
The lightning bolt icons show up for me with the pushed version. Let's see what you get after the test.

zc_install/includes/nginx_conf/ngx_conf_server.txt
```diff
@@ -0,0 +1,320 @@
...
+	location ~ %%store_folder%%/(editors|docs) {
+		return 403;
+	}
+	
+	
+	#############################################################################################
+	# ZEN CART ADMIN Folder
+	#############################################################################################	
+	## Allow all 'GET' and 'HEAD' requests to specific file types
+	location ~ %%store_folder%%/Index-NoG-daDdy {
```
drbyte: admin folder name?

zc_install/includes/nginx_conf/ngx_conf_server.txt
```diff
@@ -0,0 +1,320 @@
...
+			if ( $request_method !~ ^(GET|HEAD)$ ) { return 403; }
+			expires $zencart_expires;		
+		}
+		# Redirect all requests to ADMIN root to ADMIN/index.php 
+		location ~ %%admin_folder%%(/?)$ {
+			if ( $request_method !~ ^(GET|HEAD)$ ) { return 403; }
+			rewrite ^ %%admin_folder%%/index.php? redirect;
+		}
+		# Deny any other request to '%%store_folder%%/Index-NoG-daDdy' and/or subfolders
+		return 403;
```
drbyte: again, admin folder name?

dakanji:
Question: Noticed in a recent commit about users having custom sql files they might want to run as part of installation and wanted to make sure these are covered. Would this be run by directly inputing the URL to such a file or would the installation process pick it up along with the other sql files in the folder where it is advised to be kept? The former is currently disallowed.

drbyte:
The custom SQL files must be uploaded via FTP. There's no need for anyone to access them via the browser.

zc_install/includes/nginx_conf/ngx_conf_server.txt
```diff
@@ -0,0 +1,320 @@
...
+	#############################################################################################	
+	## Allow all requests to specific static file types
+	location ~ %%store_folder%%/.+\.(md|ico|txt|html)$ {
+		expires $zencart_expires;
+	}
+	## Allow all requests to specific PHP files directly under the ZEN CART ROOT folder
+	location ~ %%store_folder%%/(ajax|index|page_not_found|ipn_main_handler)\.php$ {
+		error_page 418 = @zencart_php_handler;
+		return 418;
+	}
```
drbyte: While I appreciate the easter egg "coffee/teapot" 418 humour, isn't it better practice to actually server a 200 response when in fact there is no error? `4xx` codes are errors. There's no error here.

Or with nginx is the use of 418 becoming a "best practice" despite the semantic confusion between purposes of 2xx and 4xx codes?

dakanji:
> The custom SQL files must be uploaded via FTP. There's no need for anyone to access them via the browser.

I was actually not referring to the upload method as this is separate from the web server. Anyway, I could infer however that  allowed latter statement is in place so all good.

> While I appreciate the easter egg "coffee/teapot" 418 humour, isn't it better practice to actually server a 200 response when in fact there is no error? 4xx codes are errors. There's no error here.

Here is a small primer for what is going on here. Two basic things to bear in mind are:

1. When any http code is returned by a location block, this is first processed internally by Nginx before a response is sent to the client.
2. The PHP handler is specified as what is called a "Named" location block and have the form `@XYZ`. These can only be accessed via internal redirection by Nginx. The upshot being that PHP can only be executed via an internal redirect.

For us, the "invalid" 418 http code is mapped to the php handler in relevant location blocks. When a request hits such a location block, it duly emits a 418 code and because of the mapping, Nginx simply passes the original request (exactly as received) to the mapped location block for processing.

The php handler processes the request and the client gets the relevant http code (200, 204, 404, 403, 500 ... whatever) along with the associated content. If you fire up your dev console in your browser and check, you will see that you will never get a 418 response to a request to one of these location blocks.

There are other ways to achieve the redirection to the named location block but this verbose way allows the user to see exactly what is happening and allows them to extend a location block to do things (further authentication/verification, flood control, malicious url checks etc, etc, etc) that might return standard http codes without conflicting with our redirection since the redirect uses an "invalid" code.

drbyte:
Thanks for the explanation. I did some more reading about the idea after I wrote that last comment. I can see how this approach makes logical sense.

drbyte:
I've been trying out a few different VM configs using various opinionated "defaults" for their nginx config.  One of them makes (what I think is) effective use of `include` features in the nginx.conf and domainxyz.conf files by doing an `include` from other folders for custom configs, particularly for the `http` section and the `server` section. 

This got me to thinking about the fact that this current implementation uses the supplied .txt files as _both_ template _and_ generated/modified output.  What do you think of defaulting to creating *new* files ending in `.conf` instead of updating the existing .txt files in-place?   That way one could just drag-and-drop (or `cp` from CLI with proper permissions, etc) these generated .conf files into the respective nginx config folders for auto-inclusion (assuming the master config is doing `include`s from separate folders for `http` and `server` respectively).

dakanji:
Actually it is a hack (commonly used) due to Nginx not providing a direct way to send people to named location blocks.

You can use try_files as well but as this requires a minimum of two parameters of which the named block has to be the second and as these are tried in turn, the first parameter has to be deliberately made to be invalid so that the second parameter is always used.

So, we could have ...

```
location ~ XYZ {
      try_files /invalid/path @zencart_php_handler;
}  
```

You can read my 2011 discussion with the Nginx author, which also explains the try_files option, here: http://mailman.nginx.org/pipermail/nginx/2011-February/025161.html. He later advised that the error_page option was faster and uses less resources.

dakanji:
Actually the file extension does not have to be conf for including. I am actually including the current .txt files in my tests.

I think that setting out **exactly** how a user should add the directives to their set up might be a bit of a step too far and that creating the files and letting them decide whether to paste or include as they wish is better. Also would suggest not getting involved with moving the files about on the server and let the user handle this. If they are running Nginx, it should be straightforward for them.

We could the however change the file extensions to `conf` or `default` as a cosmetic thing to be closer to usual Nginx files. I chose txt as I was taking a less advanced user into account and went for a file type everyone with a computer can open and I thought more advanced users would know how to go the extra step if desired.

drbyte:
> I think that setting out **exactly** how a user should add the directives to their set up might be a bit of a step too far

Ya, there are lots of things that might need manual intervention, particularly setting the  `fastcgi_pass`

dakanji:
True and they might not be even using fastcgi at all.

Perhaps I should change the files to less vague/generic names as follows:
- ngx_conf_http.txt to **zencart_http.conf**
- ngx_conf_server.txt to **zencart_server.conf**
 
... and add a note the the instructions that these can be opened using notepad or similar?

Would have preferred to have just one file but the map directive can only go in the http section. Alternative is to drop it altogether and just have the one file without the expires settings.

Any thoughts?

drbyte:
> not using fastcgi at all

True, but let's at least assume they are for now. Gotta start somewhere, and they'll be able to tweak it for themselves.

> name zencart_xxxxxx.conf

Agreed.

> note to open with notepad

I'm okay with that

> map directive vs one file without expires settings

Well, the expires thing is useful if someone wants to enable/tweak the caching of assets ... it just makes the site more performant.  

In Apache the defaults are (or at least 'were') not very optimal, and so the general industry guidance is (or at least 'was') to set those specific headers/expiry/etc stuff for various mime types.

(I think we can do without the expires stuff in the zc_install location section though.)

drbyte:
With the latest I'm encountering this problem: 

`example.com/admin/` - redirects to `example.com/admin/index.php` as expected
`example.com/admin` - redirects to `example.com/index.php` ... not desired

... but despite trying several regex alterations I'm not finding a fix.   

Thoughts?

dakanji:
Both should hit the "location ~ %%admin_folder%%(/?)$” block and end up at the same place.
I’ll reinstall the latest and check as my previous install is working as expected.

drbyte:
> Both should hit the "location ~ %%admin_folder%%(/?)$” block and end up at the same place.

I was thinking the same.

I tried changing the `(/?)` to just `/?` (thinking the parentheses weren't really needed).
And also tried removing the `(/?)` altogether. 
Neither made a difference.

drbyte:
Okay, on a whim I rebooted the server instead of just restarting nginx ... and it appears to be fine now.  Gremlins :(

drbyte:
I'm feeling good about where this is at.

What are your thoughts on readiness for merging?

dakanji:
Can be further fine tuned but is a good enough jump off point.
I’ll say go ahead.

drbyte:
Agreed. 

👍

zcwilt:
:+1:

drbyte:
Thanks @dakanji !

drbyte:
Discovered that the `zencart_ngx_server.conf` file from this breaks the use of the `.well-known` URI pattern used by LetsEncrypt when validating a certificate ... at least on a server with only one domain set up, and the letsencrypt script adds its own letsencrypt_server.conf file which is loaded into the server section of the config.

dakanji:
Yes it would … if simply setup without taking care that other non zencart related requests are processed as needed.

The set up provided is as a discrete set of directives to deal with zencart and zencart only. The way it is configured is as a stand alone set of ZC directives that can be dropped into an Nginx setup without messing anything up.

So it can live with other non zencart other directives the user may have/need. It is impossible, and highly undesirable to try,  to second guess what those may be.

So, if a user needs to use letsencrypt, they need to add the required location block for that. Just the same as if the user needs to run Wordpress or Joomla or anything else in the same server block. The zencart provided directives should not try to do this for them in terms of LE just as it wouldn’t for Drupal etc.

As said, the directives are self contained for ZC and ZC only, even the PHP handler is ZC specific. Instruction says to duplicate existing PHP handler but really that is only to avoid confusion as the knowledgeable user would simply point to their existing PHP handling block instead.

So while there may be a temptation to add a LE handling location block here, I would suggest the opposite. The issue only arises from, say, you testing in an artificail setup. You need to consider the LE issue as the same as wanting to install a CMS or any other 3rd party application instance. In this way, it would be apparent that the location blocks and directives for such is not for ZC to distribute.

Note that it says “NGINX CONFIGURATION FOR ZENCART” at the top. This is deliberate ... It is not “NGINX CONFIGURATION FOR EVERYTHING AS SET UP FOR YOU BY ZENCART”

In simple words, the user just has to add a .wellknown location block somewhere after the last ZC specific one but that should be done by them as otherwise, these directives can’t simply be dropped into an already properly setup nginx instance, which may already include a .wellknown location block, without the risk of messing it up. I certainly wouldn’t want that in my Nginx configuration.

Understand that many ZC users may wish to run LE but it is still a separate application that they need to set up to work and this has to be independent from ZC.


Regards,
Dayo

dakanji:
Thinking about this a bit more, it is a bit of a dilemma. The current setup is primarily geared to someone running ZC in a subfolder. It accommodates running it in the root folder but this is not primary target. 

Given that LE is quite important with more and more people using this, and that a user with a setup already configured would most likely easily spot the issue of a conflict, it may not be a bad idea to try to assist those running ZC alone in the root folder and haven't set anything up previously.

However struggling to see a way to accommodate them without introducing other issues and potential confusion. Best I can think of is to change:

```
	#############################################################################################
 	# ZEN CART Catchall
 	#############################################################################################	
 	## Redirect any request not previously explicitly handled to main ZEN CART index.php file
 	location ~ %%slash_folder%% {
 		rewrite ^ %%store_folder%%/index.php? redirect;
 	}
```

To

```
	#############################################################################################
 	# ZEN CART LetsEncrypt Handler
 	#############################################################################################	
 	## Handle Letsencrypt auth here if required and running ZEN CART in server root folder
 	## Needs further user configuration as required
 	location ~ %%store_folder%%/.well-known/acme-challenge {
 	#	Add required directives here
	}
	
	
	#############################################################################################
	# ZEN CART Misc Custom Handler
	#############################################################################################	
	## Misc Custom Directives if required and running ZEN CART in server root folder
	## Insert directives below as required
	#
	#
	#
 	
	#############################################################################################
 	# ZEN CART Catchall
 	#############################################################################################	
 	## Redirect any request not previously explicitly handled to main ZEN CART index.php file
 	location ~ %%slash_folder%% {
 		rewrite ^ %%store_folder%%/index.php? redirect;
 	}
```

The Misc Custom handler block is generated because this is going down the route of showing users where to put 3rd party stuff. Basically LE is one of such but just split out specifically. Both only apply for specific use case of running ZC in root folder and not having a setup in place.  Not sure at all whether this is an advisable path to follow but an option to consider. 

On balance of things considered, I personally would not implement such and would go for an FAQ item to point people to where things can be explained in more detail instead.

dakanji:
BTW, if reviewing this, all the location blocks that contain "internal" can be removed as they as redundant.

PR #891, opened 2016-04-08
Actually set HTTPS link in forms when using SSL.

xbenjii:


zcwilt:
:+1:

drbyte:
:+1:  .... note that the admin will soon only support ENTIRELY SSL or ENTIRELY non-SSL, so some of this will be moot then. But still worth merging.

PR #892, opened 2016-04-11
Multilingual config heading (v160)

Zen4All:
Is the table  "configuration_group" still needed, or can we deprecate it. I ask this, because in my search of making the admin side of Zen Cart more multi lingual I found that the "configuration_group_title" is used for populating the configuration titles
image1
These titles are also defined in admin/includes/languages/english.php, but only used in the dropdown menu, and not for the title.
My solution is to make a call to the "language_key" field in the "admin_pages" table. This way, the constant from the table can be used to translate the titles.

I know table  "configuration_group" if used in other places, but do we really need it, because the data it contains is also present in table "admin_pages"

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/892)


drbyte:
:+1:

zcwilt:
:+1:

PR #899, opened 2016-04-16
change category_tree to __construct

Zen4All:
according to php7

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/899)

### DIFF
--- a/includes/classes/category_tree.php
+++ b/includes/classes/category_tree.php
@@ -19,7 +19,7 @@ if (!defined('IS_ADMIN_FLAG')) {
  */
 class category_tree extends base {
 
-  function zen_category_tree($product_type = "all") {
+  function __construct($product_type = "all") {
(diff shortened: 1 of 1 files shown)


Zen4All:
Sorry, this is wrong. Closing the pull

drbyte:
Right - changing to `__construct` here would have been wrong. 

The change in PHP7 is not to force having a __construct method, but rather that the treatment of a "method named the same as the class" will no longer be automatically considered to be an alias of __construct ... ie: if you want a __construct method, you must be explicit.  
In the case of this class, an explicit constructor isn't necessary.

Zen4All:
Yes, i noticed the mistake after submitting. Early morning coding, and lack of caffeine.

drbyte:
Ha! That reminds me: time for coffee!

PR #900, opened 2016-04-16
Is it corrct to have 2x $cheapest in the notifier?

Zen4All:
Should
`$this->notify('NOTIFY_SHIPPING_MODULE_CALCULATE_CHEAPEST', $cheapest, $cheapest, $rates);`
be ??
`$this->notify('NOTIFY_SHIPPING_MODULE_CALCULATE_CHEAPEST', $cheapest, $rates);`

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zc-v1-series/900)

### DIFF
--- a/includes/classes/shipping.php
+++ b/includes/classes/shipping.php
@@ -214,7 +214,7 @@ class shipping extends base {
           }
         }
       }
-      $this->notify('NOTIFY_SHIPPING_MODULE_CALCULATE_CHEAPEST', $cheapest, $cheapest, $rates);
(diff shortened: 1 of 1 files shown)


drbyte:
The original code is correct.
The first parameter passed is immutable. Subsequent params are passed by reference, and are changeable. The first one could have been passed as a null or empty array, but it was repeated simply for backward compatibility.
Should not be changed.

Zen4All:
Okay thanks for the clarification.

PR #903, opened 2016-04-20
Force change of dob value for old records

drbyte:
Newer MySQL versions which are configured to enforce no zero-dates will fail to allow schema updates to any fields in the customers table if the `customers_dob` datetime field contains zero-date values.


zcwilt:
:+1:

ajeh:
:+1:

PR #904, opened 2016-04-21
Remove double-continue buttons on checkout-shipping-address screen

drbyte:
Also checkout-payment-address screen.

It was displaying 2 continue buttons if no "extra" addresses have been added to address-book.

(Continuation of commit https://github.com/zencart/zc-v1-series/commit/b151f45390c8bd0eb23eb41da3b12480cbe001da )


zcwilt:
:+1:

ajeh:
:+1:

PR #906, opened 2016-04-24
Creates array to clean meta-data

proseLA:
removes `*` in addition to `"` from meta-data.  also allows to remove any character from meta-data, by just adding that character to the array.

### DIFF
--- a/includes/modules/meta_tags.php
+++ b/includes/modules/meta_tags.php
@@ -22,6 +22,8 @@ if (strlen(SITE_TAGLINE) > 1) {
 } else {
   define('TAGLINE', '');
 }
+// array to remove specific characters from meta-data
+$bad_characters = array('"','*');
 $review_on = "";
 $keywords_string_metatags = "";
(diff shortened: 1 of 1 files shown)


includes/modules/meta_tags.php
```diff
@@ -22,6 +22,8 @@
 } else {
   define('TAGLINE', '');
 }
+// array to remove specific characters from meta-data
+$bad_characters = array('"','*');
```
drbyte: Can you move this initialization of `$bad_characters` up to line 16 (replace the comment line of "this should be the first line" with your lines 25-26.

Then add `$bad_characters` to the list of params passed to the notifier on line 17 above. This way people can extend the list by using an observer class.

proseLA:
ok.  i think i made those things happen.... i hope i understood it correctly as i did delete the comment line of "this should be the first line..."

drbyte:
👍 
That's exactly what I was looking for.

zcwilt:
:+1:

PR #908, opened 2016-04-25
The variable names prefix and suffix are incorrectly identified in th…

### LINKED ISSUE #907: Parameter variables incorrectly identified as suffix for prefix content and vice versa
Recently in review of some code, found that the functions zen_get_show_product_switch and zen_get_show_product_switch_name in both the catalog and admin side identify the parameter variable $prefix but it is used as a suffix and the parameter variable $suffix as a prefix.  This simply reverses the assignment(s) so that help offered like on the current development site (doesn't show the code of the declared function) is better understood.

mc12345678:
Fixes #907.  Swaps variable names $suffix and $prefix in two functions each on the catalog and admin side.

### DIFF
--- a/includes/functions/functions_lookups.php
+++ b/includes/functions/functions_lookups.php
@@ -676,8 +676,7 @@
 /*
  * Look up SHOW_XXX_INFO switch for product ID and product type
  */
-    function zen_get_show_product_switch_name($lookup, $field, $prefix= 'SHOW_', $suffix= '_INFO', $field_prefix= '_', 
-                                              $field_suffix='') {
+    function zen_get_show_product_switch_name($lookup, $field, $prefix= 'SHOW_', $suffix= '_INFO', $field_prefix= '_', $field_suffix='') {
       global $db;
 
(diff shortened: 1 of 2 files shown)


includes/functions/functions_lookups.php
```diff
@@ -676,7 +676,8 @@ function zen_get_products_allow_add_to_cart($lookup) {
 /*
  * Look up SHOW_XXX_INFO switch for product ID and product type
  */
-    function zen_get_show_product_switch_name($lookup, $field, $suffix= 'SHOW_', $prefix= '_INFO', $field_prefix= '_', $field_suffix='') {
+    function zen_get_show_product_switch_name($lookup, $field, $prefix= 'SHOW_', $suffix= '_INFO', $field_prefix= '_', 
+                                              $field_suffix='') {
```
drbyte: Other than adding a line-break to wrap to the next line in 4 places in this file, was there anything else changed in this file?

drbyte: Oh, duh.  github highlighting was wonky. I see it now.

drbyte:
👍

zcwilt:
:+1:

PR #909, opened 2016-04-26
Fix extra breadcrumb that was appearing when always-open-with-category is enabled

drbyte:
Example case, according to @ajeh: 
1. Configure store to always-open-with-specific-category
2. Configure to always-show-that-category
3. Navigate to a non-prod-listing page, such as `main_page=featured_products` and see the extra breadcrumbs for the always-open-with category.

This PR fixes that.


ajeh:
:+1:

zcwilt:
:+1:

Issue #594, opened 2015-10-07
Store Pickup & Ship To Address


bislewl:
Something struck me as odd. When Store Pickup is selected. The customer's ship to address is in the order. However some store-owners/employees don't look at the shipping method and just ship regardless.

Shouldn't the TABLE_ORDERS.delivery_\* fields either be left empty or say something like store pickup? Ideally the store they selected (if there is multiple)

drbyte:
Good point. 

Some challenges: 

1. Mustn't store that information until the customer confirms the purchase (else the physical ship-to preference would get lost in case they want to change their mind and ship it). 

2. The `order` class currently uses the customer's selected ship-to address when storing address details into the db. 

3. If the store-pickup address is to be shown on-screen during checkout, there's nothing presently in code to swap the shipping address back-and-forth between store-pickup and the previously-chosen deliver-to address if the customer plays with the shipping method. (nor anything remembering what their previously-chosen deliver-to address was selected as).

4. How do changes to accommodate this impact on other shipping modules?

bislewl:
My thought was for current versions (hackish style none the less) create an observer that after the TABLE_ORDERS.delivery_* fields are saved go back in and wipe those fields it out if the shipping method is store pickup, and add the store selected.

For future versions I have 2 thoughts.
1 Hide the shipping address with JS (when selected)
2 Not save the shipping address when saving the order, sort of like a virtual product that has no shipping, and just store the store selected

The virtual products does have some sort of promise I think even just as a referance point

lhungil:
I think it would be best to place the actual "Ship To" address in the order. So in this case, the actual store's physical address (possibly using an observer).

Short term, probably easiest to have the shipping information be configuration options in the "Store Pickup" shipping module... Which would support multiple different store pickup options without impacting the database or other modules...

proseLA:
i think that it is not as crucial to show the customer the store's address when they select in-store-pickup.  however storing it in the orders table is what i have done in the past by the addition of a function in the orders class.  based on the selection of an in-store pickup shipping method would change the delivery address to the config values for the store.  that way everything is kept in the orders class.

scottcwilson:
Closing issue due to lack of interest and activity.

PR #910, opened 2016-04-28
Populates order table with store address for store_pickups #594

proseLA:
i would prefer a change to the zen_get_configuration_key_value function to add another variable to remove the get_attention.  plus i'm sure there are other config values that may make the new ones superfluous.  however, i do not like the idea of parsing the MODULE_SHIPPING_STOREPICKUP_LOCATIONS_LIST to address this issue.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -438,6 +438,7 @@ class order extends base {
                             'country' => array('id' => $shipping_address->fields['countries_id'], 'title' => $shipping_address->fields['countries_name'], 'iso_code_2' => $shipping_address->fields['countries_iso_code_2'], 'iso_code_3' => $shipping_address->fields['countries_iso_code_3']),
                             'country_id' => $shipping_address->fields['entry_country_id'],
                             'format_id' => (int)$shipping_address->fields['address_format_id']);
+    $this->check_pickup();
 
     $this->billing = array('firstname' => $billing_address->fields['entry_firstname'],
                            'lastname' => $billing_address->fields['entry_lastname'],
@@ -620,6 +621,19 @@ class order extends base {
     }
     $this->notify('NOTIFY_ORDER_CART_FINISHED');
   }
+  function check_pickup() {
+    if (strstr($this->info['shipping_module_code'], 'storepickup') or strstr($_SESSION['shipping']['id'], 'storepickup')) {
+      $this->delivery['firstname'] = zen_get_config_remove_attention('STORE_NAME');
+      $this->delivery['lastname'] = '';
+      $this->delivery['street_address'] = zen_get_config_remove_attention('MODULE_SHIPPING_STOREPICKUP_ADDR1');
+      $this->delivery['suburb'] = zen_get_config_remove_attention('MODULE_SHIPPING_STOREPICKUP_SUBURB');
+      $this->delivery['city'] = zen_get_config_remove_attention('MODULE_SHIPPING_STOREPICKUP_CITY');
+      $this->delivery['postcode'] = zen_get_config_remove_attention('MODULE_SHIPPING_STOREPICKUP_ZIP');
+      $this->delivery['state'] = zen_get_config_remove_attention('MODULE_SHIPPING_STOREPICKUP_STATE');
+      $this->delivery['company'] = '';
+      $this->delivery['country']['title'] = zen_get_country_name(zen_get_configuration_key_value('STORE_COUNTRY'));
+    }
+  }
 
   function create($zf_ot_modules, $zf_mode = FALSE) {
     global $db;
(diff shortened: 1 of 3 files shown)


drbyte:
These changes are predicated on having only _one_ location for store-pickup. 

So, this needs more work to allow for the multiple locations that the `MODULE_SHIPPING_STOREPICKUP_LOCATIONS_LIST` offers. 

The optimal solution probably requires rethinking how multiple locations are handled, and storing location/address data differently than how it's done now.

proseLA:
i will give it some thought.  i'll play around to see how that config value comes into play with multiple locations.  ideally i think the customer should be picking from which location they want to pick up their goods at the time of the order.

what is your feeling with regards to adding a variable to the zen_get_configuration_key_value function, to remove that html?

drbyte:
> what is your feeling with regards to adding a variable to the zen_get_configuration_key_value function, to remove that html?

Not a fan.

drbyte:
> what is your feeling with regards to adding a variable to the zen_get_configuration_key_value function, to remove that html?

Apologies. I misread your question. I had been looking at the code changes where you'd added a new function just to remove the html, and that's what I didn't like.

drbyte:
It looks like the HTML is only added if the $lookup key doesn't have an entry .. and as such is trying to return a warning.

So, that makes me wonder why you were even running into it in the first place. Were you getting the HTML back just because your module hadn't yet been configured properly? If so then it'd be better to have rewritten the logic a bit to drive an earlier self-install of the needed custom fields.

Or am I misreading why you added the html-cleaner stuff in the first place?

drbyte:
@proseLA In v160 I'm not finding any instance of where `zen_get_configuration_key_value()` needs to return that `span`, so I'm inclined to remove it.

Thoughts?

drbyte:
So, there are two issues in this PR:
a) how best to universally handle store-pickup addresses, particularly with multiple locations, how to handle taxation for each location, and whether to show that address as the "shipping" address or not
I think this needs a larger/broader discussion, and might be better done as an `issue` since I think this PR isn't near ready for universal use yet.

b) stripping the HTML from the `zen_get_configuration_key_value` function.
I'd be willing to consider a PR for a change to this function where the whole bit about returning a span is removed. There's some regression research to do regarding that, but I'm 95% sure it's safe to do ... but some testing should be done to be certain.

proseLA:
my apologies.  i have to be better on top of my github activities, as again i have missed your comments.  something is wrong with my process right now, and i'll work on getting better at it.

first off, i agree that there should be a broader discussion about the multiple location situation.  there are a lot of things going on, and i did look at the parsing and the cost functions already built into ZC for multiple locations.  and i have not thought about the taxation issue either.

second, with regards to the zen_get_configuration_key_value function; in developing this PR, i came across the situation where we have all of these config values for address, and there is no guarantee that every field will be populated.  so unbeknownst to me, i would get this span warning, that made no sense in the context of creating a pick-up order address.  so i was forced to remove it.  from a design standpoint, i think html should only be part of the template as opposed to adding html when making database calls.  that said, i'm sure there are places out there that make use of that span, so i would not be so inclined to remove it.

if ok with you, i will create a new PR for what i was thinking with regards to the zen_get_configuration_key_value function.  and you can close and remove this one.  and i'll look into an issue with regards to the multiple address store pickups.

again, i'm sorry about the delayed responses.  something is wrong with my interaction on github and i will work on getting that addressed.

drbyte:
I don't want to add another function such as `zen_get_config_remove_attention` just to remove the HTML from the output generated by another function. 

Question: why even use `zen_get_configuration_key_value()` at all? 

Instead of `zen_get_configuration_key_value('MODULE_SHIPPING_STOREPICKUP_CITY')` you could simply use `MODULE_SHIPPING_STOREPICKUP_CITY` directly.

Every configuration_key value is converted into a define/constant, which is already available in global scope, everywhere in the site (both admin and non-admin). There's no need to use a function to access it. 

Of course, if it weren't defined yet then that would pose some small problems, but in the context in which you appear to be using it it's a moot issue.

drbyte:
@proseLA 
I don't want to add another function such as `zen_get_config_remove_attention` just to remove the HTML from the output generated by another function. 

Question: why even use `zen_get_configuration_key_value()` at all? 

Instead of `zen_get_configuration_key_value('MODULE_SHIPPING_STOREPICKUP_CITY')` you could simply use `MODULE_SHIPPING_STOREPICKUP_CITY` directly.

Every configuration_key value is converted into a define/constant, which is already available in global scope, everywhere in the site (both admin and non-admin). There's no need to use a function to access it. 

Of course, if it weren't defined yet then that would pose some small problems, but in the context in which you appear to be using it it's a moot issue.

proseLA:
learn something new every day.

given that the config values are available, i would agree that a change is not needed.

PR #911, opened 2016-05-02
Update BOC currency parsing to cope with their data changes

drbyte:
- BOC changed their currency codes to `XYZ_NOON` instead of just `XYZ`
- Also fixes a potential division-by-zero condition


zcwilt:
:+1:

ajeh:
:+1:

PR #920, opened 2016-05-13
Add legacy compatibility to adapt for new language-class components

drbyte:
This is only to attempt to avoid breaking very old plugins that haven't been upgraded to the newer more consistent syntax used in v160.


zcwilt:
:+1:

ajeh:
:+1:

PR #926, opened 2016-05-16
Add advanced tip about including plugin sql in zc_install. Fixes #138

### LINKED ISSUE #138: Optional (manually by user) Loading of additional SQL files in Installer post-install
Allow for optional loading of SQL install files. This would encompass things like Zones/Geo Zones etc. 
e.g. see https://github.com/zencart/zc-v1-series/pull/129

This would allow for not bloating the base installation sql, but also allow the store owner to add region data that is specific to their locale. 

Some thought needs to be given to the UI, so that we do not introduce more complexity into the Installer.

drbyte:


zcwilt:
:+1:

ajeh:
:+1:

PR #929, opened 2016-05-17
Multibyte String

AndrewBerezin:


drbyte:
I don't understand. You made two commits, which reverse each other.

AndrewBerezin:
Sorry. I myself do not understand how it happened :(

PR #932, opened 2016-05-20
does not redirect on invalid product_id

proseLA:
when an inactive/non-valid product link is clicked, ZC first issues an HTTP response of 302 and then a 404.  as the page no longer exists, there should NOT be a 302 issued prior to the 404.  the 404 gets handled from includes/modules/pages/product_info/header_php.php.  so i am unclear why a 302 is being issued here and then going to the same page.  

this PR removes the 302 response/redirect when an invalid product_id is selected.

https://www.zen-cart.com/showthread.php?220104&p=1311344#post1311344

### DIFF
--- a/includes/init_includes/init_sanitize.php
+++ b/includes/init_includes/init_sanitize.php
@@ -56,7 +56,7 @@ if (isset($_GET ['products_id']) && isset($_SESSION ['check_valid']) && $_SESSIO
      * do not recheck redirect
      */
     $_SESSION ['check_valid'] = 'false';
-    zen_redirect(zen_href_link($mainPage, 'products_id=' . $_GET ['products_id']));
+    //zen_redirect(zen_href_link($mainPage, 'products_id=' . $_GET ['products_id']));
   }
 } else {
   $_SESSION ['check_valid'] = 'true';


drbyte:
Thanks for this. You raise a good point. 

Question:
Should the redirect still occur if the page requested isn't `product_info` ? ... ie: what about if the `products_id` field is present but `main_page` is set to something else about the product such as reviews etc?

drbyte:
Interestingly that line's been there since as far back as ZC v1.1.0 ... and a lot of handling for 404 has been improved since then but this probably needs some updating ;)

proseLA:
let me take a look at this.  somehow i missed your comment from 4 days ago.  i'll get back to you today.  sorry about the delay.

drbyte:
> this PR removes the 302 response/redirect when an invalid product_id is selected.

Agreed. 👍 

Separately I'm not opposed to a new feature to allow specific configuration from the Admin for global rules about how to handle missing products, which could potentially be overridden by more specific rules where needed ...

zcwilt:
:+1:

proseLA:
cool.
i will think about your suggestion with regards to another configuration on the admin.  frankly i'm still thinking about the best way to incorporate store addresses on in-store pickups for multiple stores...

PR #938, opened 2016-05-21
Remove block of dead code

scottcwilson:


drbyte:
I'm a bit reluctant to remove this, as it was left commented to show a different query for a different strategy on determining the data being retrieved.

Have you looked into its history to confirm that the old will never be wanted again?

scottcwilson:
This giant commented out block of code has been in there, commented out, since Zen Cart 1.3.6.   We have source control.  Why not clean it out and introduce it again only if it is actually needed?

drbyte:
Thanks for checking the history.

👍

zcwilt:
:+1:

ajeh:
:+1:

PR #940, opened 2016-05-21
Maintain order status that is above default on PayPal batch capture

mc12345678:
Submitted to replace a previous pull request that doesn't want to play nice when rebasing from v155 to v160.

Discussion held back on ZC 1.5.1 at https://www.zen-cart.com/showthread.php?214423-Order-status-changes-after-batch-capture-on-PayPal-website where store owner observed that the status of orders that were previously manually updated inside of the ZC admin were lowered once batch capture was performed on PayPal site.

The change proposed above was reported as working at least before the and was added to the added if statement. And statement intent is to apply/allow the old status to remain if and only if the transaction is one that isn't intended to reduce the status. For example, refunding, reversing, denying or failing a transaction should not allow the status to remain elevated through this process, but instead take the status intended for those conditions.

### DIFF
--- a/ipn_main_handler.php
+++ b/ipn_main_handler.php
@@ -470,8 +470,8 @@ Processing...
       if ((int)$new_status == 0) $new_status = 1;
       if (in_array($_POST['payment_status'], array('Refunded', 'Reversed', 'Denied', 'Failed'))
            || substr($txn_type,0,8) == 'cleared-' || $txn_type=='echeck-cleared' || $txn_type == 'express-checkout-cleared') {
-        $sql = ("select orders_status from " . TABLE_ORDERS . "
-                 where orders_id = :ordersID:");
+        $sql = "select orders_status from " . TABLE_ORDERS . "
+                 where orders_id = :ordersID:";
         $sql = $db->bindVars($sql, ':ordersID:', $ordersID, 'integer');
         $old_status = $db->Execute($sql);
         if ($new_status < $oldstatus->fields['orders_status'] && (substr($txn_type, 0, 8) == 'cleared-' || $txn_type=='echeck-cleared' || $txn_type == 'express-checkout-cleared')) {
--- a/ipn_main_handler.php
+++ b/ipn_main_handler.php
@@ -466,11 +466,8 @@ Processing...
           break;
       }
       // update order status history with new information
-      ipn_debug_email('IPN NOTICE :: Set new status ' . $new_status . " for order ID = " .  $ordersID . ($_POST['pending_reason'] != '' ? '.   Reason_code = ' . $_POST['pending_reason'] : '') );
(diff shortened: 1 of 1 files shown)


mc12345678:
Previous direction from DrByte:

> Upon further review, I agree with this PR.

> Can you make 2 changes:

>    Redo it against v160 instead of v155
    As you suggested, above, move the ipn_debug_email() call down below your additions, so that it's just above the ipn_update_orders_status_and_history line following your additions

Looking at the location to place the ipn_debug_email, one, by moving it out of the main flow path, an indicator is lost to anyone "following" the debug trail, second, there appear to be two places in this area where the status could change and the first occurrence does not necessarily ensure entry into the second if statement area as well, the modification made from status of 0 to status of 1 would not be identified as a "problem"..

Just want to be sure that the goal is to only issue the debug statement when the following is true and then only after all "changes" have been made: 
```php
if (in_array($_POST['payment_status'], array('Refunded', 'Reversed', 'Denied', 'Failed'))
           || substr($txn_type,0,8) == 'cleared-' || $txn_type=='echeck-cleared' || $txn_type == 'express-checkout-cleared') {
```

Further discussion from DrByte on the above:
> If I recall correctly, the logging is probably only necessary just before the actual execution of the change of the data in the database.

ipn_main_handler.php
```diff
@@ -467,9 +467,20 @@
...
       if (in_array($_POST['payment_status'], array('Refunded', 'Reversed', 'Denied', 'Failed'))
            || substr($txn_type,0,8) == 'cleared-' || $txn_type=='echeck-cleared' || $txn_type == 'express-checkout-cleared') {
+        $sql = "select orders_status from " . TABLE_ORDERS . "
+                 where orders_id = :ordersID:";
+        $sql = $db->bindVars($sql, ':ordersID:', $ordersID, 'integer');
+        $old_status = $db->Execute($sql);
+        if ($new_status < $oldstatus->fields['orders_status'] && (substr($txn_type, 0, 8) == 'cleared-' || $txn_type=='echeck-cleared' || $txn_type == 'express-checkout-cleared')) {
+          $new_status = $old_status->fields['orders_status'];
+        }
+        ipn_debug_email('IPN NOTICE :: Set new status ' . $new_status . " for order ID = " .  $ordersID . ($_POST['pending_reason'] != '' ? '.   Reason_code = ' . $_POST['pending_reason'] : '') );
```
drbyte: Between lines 469-483 we don't need to have these 3 calls to `ipn_debug_email`. 

Please remove the one on 469 and 472. The one on 483 will suffice.
And restore line 470 to its single-line setting of $new_status.

drbyte:
👍

zcwilt:
:+1:

PR #944, opened 2016-05-23
Fixes #915 sets default for DOB and Gender to false. 

### LINKED ISSUE #915: Suggestion: default for ACCOUNT_DOB, ACCOUNT_GENDER be false
Most people truly don't need either of these, and additional data collection at account creation time is annoying to customers.  Defaulting them to off means people who really want or need these fields still have access to them but others don't have to remember to turn them off.

scottcwilson:
Makes it easier for new users to create carts which prompt for the minimum needed information.


zcwilt:
:+1:

ajeh:
:+1:

PR #950, opened 2016-05-29
Fixes #76 Makes debug logs clickable

### LINKED ISSUE #76: Debug logs widget - make log entries clickable
clicking should then show the actual log entries, either in a modal or a new page

scottcwilson:
Tests:
- missing file
- javascript in file
- tampering with logname parameter.

### DIFF
--- a/admin/view_log.php
+++ b/admin/view_log.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * @package admin
+ * @copyright Copyright 2016 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: $
+ */
+
+  require('includes/application_top.php');
+  require('includes/admin_html_head.php');
+?>
+</head>
+<body>
+<!-- header //-->
+<div class="header-area">
+<?php require(DIR_WS_INCLUDES . 'header.php'); ?>
+</div>
+<!-- header_eof //-->
+
+<!-- body //-->
+<div class="pageHeading"><?php echo HEADING_TITLE; ?></div>
+<br />
+<?php 
+  if (!defined('DIR_FS_LOGS')) define('DIR_FS_LOGS', DIR_FS_CATALOG . 'logs');
+  if (!defined('DIR_FS_SQL_CACHE')) define('DIR_FS_SQL_CACHE', DIR_FS_CATALOG . 'cache');
+   if ($_GET['pathcode'] == 1) { 
+      $filename = DIR_FS_LOGS . "/";
+      $pathname = DIR_FS_LOGS . "/";
+   } else { 
+      $filename = DIR_FS_SQL_CACHE . "/";
+      $pathname = DIR_FS_SQL_CACHE . "/";
+   }
+   $filename = $_GET['logname']; 
+   $file = $pathname . $filename;
+   $rpdir = realpath($file);
+   $rppathdir = realpath($pathname);
+   if ( (substr($rpdir, 0, strlen($rppathdir)) != $rppathdir) || 
+        (strlen($rpdir) < strlen($rppathdir)) ) { 
+      // hacking logname parameter in URL
+      echo FILE_NOT_FOUND; 
+   } else if (file_exists($file)) {
+      echo '<strong>' . PATH . '</strong>' .  $pathname . '<br />'; 
+      echo '<strong>' . FILENAME . '</strong>' .  $filename . '<br /><br />'; 
+      $file_array = @file($file);
+      $file_contents = @implode('', $file_array);
+
+      $file_writeable = true;
+      echo nl2br(zen_output_string_protected($file_contents)); 
+   } else { 
+      echo FILE_NOT_FOUND; 
+   }
+
+?>
+<!-- body_eof //-->
+
+<!-- footer //-->
+<div class="footer-area">
+<?php require(DIR_WS_INCLUDES . 'footer.php'); ?>
+</div>
+<!-- footer_eof //-->
+</body>
+</html>
+<?php require(DIR_WS_INCLUDES . 'application_bottom.php'); ?>
--- a/admin/view_log.php
+++ b/admin/view_log.php
@@ -28,18 +28,15 @@
    $rpdir = realpath($file);
    $rppathdir = realpath($pathname);
    if ( (substr($rpdir, 0, strlen($rppathdir)) != $rppathdir) || 
-        (strlen($rpdir) < strlen($rppathdir)) ) { 
-      // hacking logname parameter in URL
+        (strlen($rpdir) < strlen($rppathdir)) || 
+        (!file_exists($file)) ) {
       echo FILE_NOT_FOUND; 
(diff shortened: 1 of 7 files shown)


scottcwilson:
Please note: I did not test with a file in cache because I have never seen one of those.  If you can send me an example, I can verify it.

zcwilt:
Hmm. So in v160 all logs should be written to DIR_FS_LOGS and never DIR_FS_SQL_CACHE.
If we have code that is writing logs to  DIR_FS_SQL_CACHE. then that's incorrect.
(the installer should take care of testing that DIR_FS_LOGS is writeable)

So I don't think your code needs to worry about  DIR_FS_SQL_CACHE

scottcwilson:
@zcwilt Did you want me to remove the DIR_FS_SQL_CACHE stuff - including in admin/includes/functions/general.php?

includes/library/zencart/DashboardWidget/src/Logs.php
```diff
@@ -34,11 +43,11 @@ public function prepareContent()
     $max_logs_to_list = 20;
 
     $logs = get_logs_data($max_logs_to_list);
-    // keys in $logs are: 'path', 'filename', 'filesize', 'unixtime', 'datetime'
+    // keys in $logs are: 'path', 'filename', 'filesize', 'unixtime', 'datetime', 'pathcode'
 
     foreach ($logs as $log) {
       $tplVars['content'][] = array(
-                                    'text'=> $log['filename'], // @TODO future: add clickable link to ajax-driven viewer here
+                                    'text'=> '<a href="' . zen_href_link(FILENAME_VIEW_LOG, 'logname=' . $log['filename'] . "&pathcode=" . $log['pathcode'], 'NONSSL') . '">' . $this->getDisplayName($log) . '</a>',
```
drbyte: Don't need to pass the 'NONSSL' parameter here.

drbyte:
> Did you want me to remove the DIR_FS_SQL_CACHE stuff - including in admin/includes/functions/general.php?

Yes, please.  The widget and viewer don't need to reflect files in the `DIR_FS_SQL_CACHE` folder anymore.

admin/view_log.php
```diff
@@ -0,0 +1,57 @@
...
+<!-- header_eof //-->
+
+<!-- body //-->
+<div class="pageHeading"><?php echo HEADING_TITLE; ?></div>
+<br />
+<?php 
+  if (!defined('DIR_FS_LOGS')) define('DIR_FS_LOGS', DIR_FS_CATALOG . 'logs');
+   $filename = DIR_FS_LOGS . "/";
+   $pathname = DIR_FS_LOGS . "/";
+   $filename = $_GET['logname'];
```
drbyte: Probably don't need the first `$filename =` here. Just the 2nd one.

admin/view_log.php
```diff
@@ -0,0 +1,57 @@
...
+      echo '<strong>' . PATH . '</strong>' .  $pathname . '<br />'; 
+      echo '<strong>' . FILENAME . '</strong>' .  $filename . '<br /><br />'; 
+      $file_array = @file($file);
+      $file_contents = @implode('', $file_array);
+
+      $file_writeable = true;
+      echo nl2br(zen_output_string_protected($file_contents)); 
+   } else { 
+      echo FILE_NOT_FOUND; 
+   }
```
drbyte: Could this `if ... else if ... else` be simplified to just:

```
if (file_exists) {
//...
} else {
// FILE_NOT_FOUND
}
```

drbyte:
And can the `$file_writeable` be removed?

admin/view_log.php
scottcwilson: I added an admin log to the file hacking case.

admin/view_log.php
scottcwilson: Done.

admin/view_log.php
```diff
@@ -0,0 +1,55 @@
...
+      echo FILE_NOT_FOUND; 
+      zen_record_admin_activity("Possible hacking in log viewer - looking at " . $rppathdir);
+   } else if (file_exists($file)) {
+      echo '<strong>' . PATH . '</strong>' .  $pathname . '<br />'; 
+      echo '<strong>' . FILENAME . '</strong>' .  $filename . '<br /><br />'; 
+      $file_array = @file($file);
+      $file_contents = @implode('', $file_array);
+      echo nl2br(zen_output_string_protected($file_contents)); 
+   } else { 
+      echo FILE_NOT_FOUND;
```
drbyte: Thoughts on simplifying the `if...elseif...else`? 
I'm wordering why there are 2 different paths to an identical `FILE_NOT_FOUND` result.

scottcwilson: The first one really isn't a file not found, it's a "you don't have
permission," but I don't want to say that - or do you think I should?
The first one occurs when they fiddle with the URL to look at a file
outside of the logs directory.

drbyte: As for someone trying to "hack"... how can they access the viewer in the first place if they're not logged in?
And if they're not authorized, then the log-viewer script will never fire, so there's no need to even have something to record the attempt, because the failed login attempt is already logged.

... or am I missing something?

scottcwilson: Just paranoia and anti-bugging.  OK, I will consolidate these cases.

drbyte:
> Just paranoia and anti-bugging. 

Don't get me wrong. I appreciate that you're considering those important angles!

Sometimes I get caught in the mental trap of coding for problems that don't exist, and so I try to spot and prevent those whenever possible. Occasionally less is more ;)

admin/view_log.php
drbyte: > Just paranoia and anti-bugging. 

Don't get me wrong. I appreciate that you're considering those important angles!

Sometimes I get caught in the mental trap of coding for problems that don't exist, and so I try to spot and prevent those whenever possible. Occasionally less is more ;)

scottcwilson: Imagine there is a password leak, and the hacker gets access to this page, then uses this page to attempt to view the configure.php file. Without this protection, they could get to it and see the database password.

drbyte: Fair point. Combine your first two cases into one:

```
if (file is a valid log file ... and file exists) {
//view it
} else {
// not found
}
```

drbyte:
Okay. Looking good.

One more thing:  what are your thoughts on removing the use of `@` so that errors can be logged properly if those fail, instead of them just failing blindly?

scottcwilson:
How about we do that on a separate branch?  This PR is getting a bit long in the tooth.

drbyte:
Understood. I was gonna squash these commits into one when merging.

drbyte:
Future improvements:
- [ ] Add a Back button
- [ ] Instead of just the date showing in the widget, some indicator of what "kind" of file it is would be good ... ie: the filename prefix is good: myDebug, AIM, zcinstall, etc.  When it's just the date and there's no option to clean them up or filter or sort it can be cumbersome to even bother viewing.
- [ ] Maybe even a prev/next button ... but that adds quite a bit of complexity to the code ... unless it becomes a LEAD page in a future life

drbyte:
But for now:
👍

zcwilt:
:+1:

PR #961, opened 2016-06-02
Tidy up inconsistencies in order_weight handling

drbyte:


includes/classes/order.php
```diff
@@ -377,7 +377,7 @@ function cart() {
                         'tax_groups' => array(),
                         'comments' => (isset($_SESSION['comments']) ? $_SESSION['comments'] : ''),
                         'ip_address' => $_SESSION['customers_ip_address'] . ' - ' . $_SERVER['REMOTE_ADDR'],
-                        'order_weight' => ($shipping_weight * $shipping_num_boxes)
+                        'order_weight' => ($shipping_weight * $shipping_num_boxes),
```
drbyte: There is probably a much smarter approach here, instead of multiplying these 2 numbers. The `create()` method uses the calculated shipping weight value as the authoritative source.

zcwilt:
:+1:

ajeh:
:+1:

PR #963, opened 2016-06-03
sqlpatch crashes due to missing notifier declaration

scottcwilson:


admin/sqlpatch.php
```diff
@@ -84,6 +84,7 @@
 // NOTE: this line break is intentional!!!!
 
  function executeSql($lines, $database, $table_prefix = '') {
+   global $zco_notifier; 
    @set_time_limit(1200);
 
    global $db, $debug, $messageStack;
```
drbyte: Please combine the `globals` on one line.

admin/sqlpatch.php
```diff
@@ -282,7 +283,6 @@ function executeSql($lines, $database, $table_prefix = '') {
           default:
             break;
           } //end switch
-//        } // endif $table_prefix
```
drbyte: To remove this requires removing the `if` which opens up at line `114`.

scottcwilson: It's commented out, so it's dead code.  It screws up bracket matching to have things like that.

admin/sqlpatch.php
scottcwilson: fixed

admin/sqlpatch.php
drbyte: Your bracket-matching comment is understood. But you've removed only the 2nd half of the bracket ... so you're actually creating a bracket-matching problem by not removing the opening bracket up at 113/114.

drbyte:
👍

ajeh:
:+1:

PR #964, opened 2016-06-04
Fixes Add Widget when logged out

scottcwilson:
Failure on call to getInstallableWidgets because of login timeout
occurs silently, so the user has no idea what's happening.  This
redirects them to re-authenticate.

Testing: 
- login to admin
- open a new tab on an admin page, and logout on that tab.
- on first tab, press Add Widget button.  With this fix, you'll be redirected to login.


drbyte:
👍

zcwilt:
:+1: 

with a caveat. 
This is a general problem in v160 whenever ajax code is called, and will rear it's ugly head elsewhere (especially LEAD pages). At some point that needs addressing globally.

PR #967, opened 2016-06-05
Consolidate functions files to remove duplication between admin/catalog

drbyte:
This refactoring gets rid of most duplication between admin/catalog, and fixes some disparities that existed as a result of the old duplication.


scottcwilson:
YAY! This has long driven me crazy. :)

drbyte:
me too .... and I think I nearly went crazy while doing this!

PR #968, opened 2016-06-05
order_status_name field is not shown

Zen4All:
the line "'language' => true," needs to be removed, otherwise the when inserting, or editing a order status name the appropriate field is not shown

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/968)


zcwilt:
Hi.
The reason the field does not appear is a CSS issue. 
see https://github.com/zencart/zencart/pull/972

Zen4All:
okay, understood

PR #969, opened 2016-06-05
Making sqlpatch less space sensitive

### LINKED ISSUE #951: Files run through Admin->Tools->Install SQL Patches should ignore additional whitespace
Right now, running a file with the statement 

UPDATE products ADD ADD price_w decimal(15,4) NOT NULL default '0.0000';

succeeds, but if you have two spaces after the UPDATE 

UPDATE  products ADD ADD price_w decimal(15,4) NOT NULL default '0.0000';

it fails.

scottcwilson:
Without this change, sqlpatch commands with two spaces
following the initial verb will fail.  This is confusing
to users and leads to disinformation about how well
sqlpatch works (vs PHPMyAdmin).

Fixes #951


drbyte:
👍

zcwilt:
:+1:

PR #970, opened 2016-06-05
Changes to make test runnable in insolation

scottcwilson:
These changes were required so that this test could be
run stand alone without running the entire suite, i.e.
```
phpunit --filter testAdminLoggingCase testFramework/unittests/testsSundry/AdminFunctionsGeneralTest.php
```
I am writing documentation on how to create and modify tests, and
I want to use this as an example.


zcwilt:
:+1:

drbyte:
👍

PR #976, opened 2016-06-07
changes for multilingual country names

### LINKED ISSUE #374: Multi lingual country names
This may be a to large of a change for ZC 1.6, but I would love to see the country names to be multi lingual in a future release of Zen Cart..
A lot of my customers have shops in multiple languages, and ften ask how and where they can change the county names. Not every one knows the English names for the worlds countries.

Zen4All:
Closes #374


zc_install/sql/install/mysql_zencart.sql
```diff
@@ -393,6 +393,20 @@ CREATE TABLE countries (
...
+#
+
+CREATE TABLE countries_name (
+  countries_id int(11) NOT NULL,
+  language_id int(11) NOT NULL DEFAULT 1,
+  countries_name varchar(64) NOT NULL,
+  UNIQUE countries (countries_id, language_id),
+  KEY idx_countries_name_zen (countries_name)
+) ENGINE=MyISAM;
+
```
drbyte: In this `mysql_zencart.sql` file, shouldn't this also remove `countries_name` and its index from the `countries` table? ... to match the dropping of that column in the `upgrade` file?

includes/classes/order.php
```diff
@@ -267,41 +267,47 @@ function cart() {
     $customer_address_query = "select c.customers_firstname, c.customers_lastname, c.customers_telephone,
                                     c.customers_email_address, ab.entry_company, ab.entry_street_address,
                                     ab.entry_suburb, ab.entry_postcode, ab.entry_city, ab.entry_zone_id,
-                                    z.zone_name, co.countries_id, co.countries_name,
+                                    z.zone_name, co.countries_id, con.countries_name,
```
drbyte: In this file you use `con.` as the alias, but in all the other files you've touched you are using `cn.`
I suggest changing this file to match by using `cn.`

zc_install/sql/install/mysql_zencart.sql
Zen4All: You are right. This should be removed.

includes/classes/order.php
Zen4All: There is actually a valid reason I was using con instead of cn. But I will change it to cn

admin/customers.php
```diff
@@ -1164,7 +1164,7 @@ function check_form() {
 
       if ((!isset($_GET['cID']) || (isset($_GET['cID']) && ($_GET['cID'] == $customer['customers_id']))) && !isset($cInfo)) {
         $country = $db->Execute("select countries_name
-                                 from " . TABLE_COUNTRIES . "
+                                 from " . TABLE_COUNTRIES_NAME . "
                                  where countries_id = '" . (int)$customer['entry_country_id'] . "'");
```
drbyte: Should this include the Language ID? ... else line 1173 is only pulling the first randomly-selected country-name record found.

drbyte:
Looking good ... ;)

admin/customers.php
Zen4All: Duh :blush:

admin/customers.php
```diff
@@ -1165,7 +1165,8 @@ function check_form() {
       if ((!isset($_GET['cID']) || (isset($_GET['cID']) && ($_GET['cID'] == $customer['customers_id']))) && !isset($cInfo)) {
         $country = $db->Execute("select countries_name
                                  from " . TABLE_COUNTRIES_NAME . "
-                                 where countries_id = '" . (int)$customer['entry_country_id'] . "'");
+                                 where countries_id = '" . (int)$customer['entry_country_id'] . "'
+                                 and language_id = " . $_SESSION['languages_id']);
```
drbyte: I hate to be a pain, but it's safer to `(int)` this ;)

Zen4All: haha, I was considering on doing that.

drbyte:
Great!

👍

zcwilt:
:+1:

torvista:
I come to the conclusion there is no mechanism yet for automatically adding localised country names in conjunction with the addition of a new language.
For example on the create account page the country drop-down is empty (when using the new language) as zen_get_countries gets results filtered by the language id and there are no entries in the countries_name table for other than language 1, so no results.
This was tested using my spanish language pack for 1.6.
It should at least duplicate the existing names for the new language id, if there is nothing in the language pack.

Zen4All:
@torvista This should work, There is a class that determines which tables need updating. Do other tables get copied? like product description?
for reference see file: https://github.com/zencart/zencart/blob/458e9572d81c9cafb51a8f4836a4550156baa35b/app/Controllers/admin/Languages.php#L157-L163

torvista:
Results of your test?

Zen4All:
Sorry not tested yet. To busy with other things

PR #978, opened 2016-06-07
Change references to $_SERVER['PHP_SELF'] to $_SERVER['SCRIPT_NAME']

drbyte:
Some server engines don't set `PHP_SELF` reliably, so using `SCRIPT_NAME` appears to be more reliable and compatible.

As for compatibility, here's a simple reference: http://php.about.com/od/learnphp/qt/_SERVER_PHP.htm

### DIFF
--- a/admin/includes/application_bootstrap.php
+++ b/admin/includes/application_bootstrap.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @package admin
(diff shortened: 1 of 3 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #984, opened 2016-06-11
Suggest adding target

Zen4All:
When clicking the link Storefront, the user is directed to the Storefront as expected, but in the same screen, which is not the way it used to work in the old link. I suggest putting the targent="_blank" back in the link

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/984)


drbyte:
👍

PR #985, opened 2016-06-11
Adding missing buid in values

Zen4All:
Using the system inspection page, I came across these values that were missing from the build in arrays

Very handy tool :smile:

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/985)


drbyte:
👍

PR #994, opened 2016-06-13
Link to Store Manager to clean out logs

scottcwilson:
Provide a link in the widget to make it easy to clean everything out.

### DIFF
--- a/includes/library/zencart/DashboardWidget/src/Logs.php
+++ b/includes/library/zencart/DashboardWidget/src/Logs.php
@@ -59,7 +59,7 @@ class Logs extends AbstractWidget
     }
     $tplVars['content'][] = array('text'=> $final_message, 'value'=> '');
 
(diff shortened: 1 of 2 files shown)


zcwilt:
:+1:

drbyte:
👍 
Note: If the user doesn't have permission to access the Store Manager page, they'll get a "Denied" page if they click that link.

ajeh:
Admin should be using zen_admin_href_link vs zen_href_link

drbyte:
True.

ajeh:
:+1:

includes/library/zencart/DashboardWidget/src/Logs.php
```diff
@@ -59,6 +59,9 @@ public function prepareContent()
     }
     $tplVars['content'][] = array('text'=> $final_message, 'value'=> '');
 
+    $clean_message = '<a href="' . zen_href_link(FILENAME_STORE_MANAGER) . '">' . TEXT_CLEANUP_LOGFILES . '</a>';
```
drbyte: href?

PR #997, opened 2016-06-14
Authorize.net has changed their rollout plan for Akamai

drbyte:
According to previous communication, Authorize.net was going to switch to Akamai on June 30 "automatically". But they've delayed that due to many vendors not being prepared.

To benefit from the improved reliability sooner than later, this simple URL change is all that's required in ZC.


zcwilt:
:+1:

ajeh:
:+1:

PR #998, opened 2016-06-15
Logs color

scottcwilson:
Shows the logs widget in normal color with an "all's well" icon when there are no logs, only switching to red with the warning icon if logs exist.

### DIFF
--- a/includes/library/zencart/DashboardWidget/src/Logs.php
+++ b/includes/library/zencart/DashboardWidget/src/Logs.php
@@ -20,6 +20,12 @@ if (!defined('IS_ADMIN_FLAG')) {
  */
 class Logs extends AbstractWidget
 {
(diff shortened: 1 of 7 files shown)


includes/library/zencart/DashboardWidget/src/Logs.php
```diff
@@ -20,6 +20,12 @@
  */
 class Logs extends AbstractWidget
 {
+  public function __construct($widgetKey, $widgetInfo = NULL) 
+  {
+      parent::__construct($widgetKey, $widgetInfo);
+      $count = 0;
```
drbyte: Shouldn't `$count` be `$this->count` ?

And a declaration of the variable as `protected $count;` above the constructor?

zcwilt:
:+1:

drbyte:
👍

zcwilt:
:+1:

PR #999, opened 2016-06-16
Consolidate catalog-side javascript to remove duplication

### LINKED ISSUE #574: Javascript Consolidation
Task: Consolidate javascript and jQuery scripts into fewer shared files in both admin and catalog (except what's shared by both).
Currently many are scattered all over instead of being centralized. 

Ideally:
 - one shared js script for validations that are used by both catalog and admin
 - one admin-specific js script, covering all common needs across admin
 - one catalog-specific js script, which covers all common needs across catalog
 - and a few occasional loaded-on-demand scripts if truly necessary to avoid loading big resources and wasting CPU/bandwidth in relevant cases

drbyte:
Closes #574


drbyte:
NOTE: I've got some conflicts on files, which I'd like to fix before merging.

drbyte:
Replaced by #1023

PR #1001, opened 2016-06-17
Do not display Max # Rows Dropdown if no Pagination

scottcwilson:
Using  'isPaginated' => false in the  listingQuery should
prevent the dropdown for “Max number of items per page” showing up.


drbyte:
👍

zcwilt:
:+1:

Issue #996, opened 2016-06-14
Many admin pages do not have the new look 


scottcwilson:
Admin home page looks good, as do LEAD reports, but admin-> configuration-> anything and admin->customers->customers look bad - pale header bar, hard to read links on the right.

zcwilt:
The ultimate intention is for all admin page to be converted to the new Controller logic, 
However in the mean time some bridging code is probably needed

scottcwilson:
I wonder if we can get some css wizard to style the old pages to look like the new ones as a baby step towards getting them more aligned.  Once I successfully do the Find Duplicate Models report, I'll work on some of the other reports, which are all using the old code at the moment.

zcwilt:
Unfortunately it's not quite as simple as that.

Pages that have been converted to the new controller logic, use the $tplVars array to pass information to the view logic.

legacy pages don;t set any $tplVars, so view that need them will break ( e.g the new adminlte special menu)

That problem isn't insurmountable, but is a code fix rather than a css fix.

dakanji:
I'll like to take this on (at least to make the look of the pages be consistent as a start by moving them to bootstrap).
Actually, I had started the process by setting up a structured process here: https://github.com/dakanji/zencart-style which I was going to discuss with @zcwilt, @drbyte, @ajeh to add to the project for generating the CSS and JS files in a structured.
I can move this over to the template being used on the index page and then start to roll this across to the other pages. 

The template implementation currently has developer only files files in it that should not be distributed and a few other small issues (should be in a stand alone repo and only the outputs from the dist folder should be distributed).

From the way the template has been deployed, perhaps the dev team is not that up to speed with task runners such as Gulp/Grunt or CSS extension languages such Sass/Less but it is a cakewalk once the dev environment is set up.

Bootstrap and the template use Less but Less has basically lost out to Sass. Bootstrap 4 will be Sass based and there is already an official Sass port of Bootstrap 3. So, I will set this up using Sass (which incidentally can take pure CSS code directly so even if not fully understood, can be updated with CSS directly.

If okay with you, I'll get on moving the candidate repo over to generating the template files and we can talk through things as we progress.

drbyte:
@dakanji 

Undecided about where you're headed here.

This issue ticket is about Admin things, but the way you worded things suggests you're aiming at the entire system. Is that an accurate interpretation?


[...]

PR #1002, opened 2016-06-17
Fix HTML Formatting of LEAD pages (top)

scottcwilson:
The part just below adminLeadContainer had spacing that was all
out of alignment.

NOTE: This only affects "View Source"; it would not be visible to normal users who are not looking at the the created HTML.

### DIFF
--- a/admin/includes/template/templates/tplAdminLead.php
+++ b/admin/includes/template/templates/tplAdminLead.php
@@ -25,28 +25,27 @@
 <section class="row" id="adminLeadContainer">
     <aside class="col-md-2">
         <div class="panel">
-            <?php if (count($tplVars['leadDefinition']['actionLinks'])) { ?>
-            <div class="panel">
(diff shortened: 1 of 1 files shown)


drbyte:
Apart from changing white-space, and changing the `a href` link from 3 lines to 2, I don't see what this PR accomplishes. Save for the spaces nothing has changed.

scottcwilson:
When you do a view source things are more aligned.  This is for the
underlying HTML.

drbyte:
Oh, I get it: this is just cosmetically prettifying the raw source, and nothing visual. 
And not fixing the #996 issue you reported previously?

drbyte:
Note: I was puzzled why you left the `a href` on one line, then put the link text and closing `</a>` on the next line, when you were consolidating things into fewer lines.

scottcwilson:
No - sorry for the confusion.  #996 is actually visible to admin users.
This ticket is only visible by those who do a view source.  It's just
jarring because the source above this is beautifully formatted, and this
block was a mess.

PR #1003, opened 2016-06-17
Fix alert to accommodate out-of-stock variations in shopping basket

drbyte:
Ref https://www.zen-cart.com/showthread.php?220471-Mixed-Stocks-in-shopping-cart


ajeh:
:+1:

zcwilt:
:+1:

PR #1004, opened 2016-06-18
Don't use https when building URLs if site does not support it

scottcwilson:
Removes https for localhost installs

### DIFF
--- a/zc_install/includes/functions/general.php
+++ b/zc_install/includes/functions/general.php
@@ -195,16 +195,11 @@ function zen_get_select_options($optionList, $setDefault)
     $documentRoot = zen_get_document_root();
     $url = ($request_type == 'SSL' ? 'https://' : 'http://') . $_SERVER['HTTP_HOST'] . str_replace('/zc_install/index.php', '', $_SERVER['SCRIPT_NAME']);
     $httpServer = zen_parse_url($url, 'host', true);
-    $using_localhost = false;
-    define('LOCALHOST', 'localhost'); 
-    if (strncmp($httpServer, LOCALHOST, strlen(LOCALHOST)) == 0) { 
-       $using_localhost = true;
-    }
     $adminServer = ($request_type == 'SSL') ? 'https://' : 'http://';
     $adminServer .= $httpServer;
     $catalogHttpServer = 'http://' . $httpServer;
     $catalogHttpUrl = 'http://' . $httpServer . '/' . zen_parse_url($url, 'path', true);
-    if (!$using_localhost) { 
+    if ($request_type == 'SSL') { 
       $catalogHttpsServer = 'https://' . $httpServer;
       $catalogHttpsUrl = 'https://' . $httpServer . '/' . zen_parse_url($url, 'path', true);
     } else {
--- a/zc_install/includes/functions/general.php
+++ b/zc_install/includes/functions/general.php
@@ -195,12 +195,22 @@ function zen_get_select_options($optionList, $setDefault)
     $documentRoot = zen_get_document_root();
     $url = ($request_type == 'SSL' ? 'https://' : 'http://') . $_SERVER['HTTP_HOST'] . str_replace('/zc_install/index.php', '', $_SERVER['SCRIPT_NAME']);
     $httpServer = zen_parse_url($url, 'host', true);
+    $using_localhost = false;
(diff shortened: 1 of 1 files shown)


scottcwilson:
hmm... that logic doesn't seem to work on my installation.  But I will rework my fix so that the checks are in place function correctly with MAMP.   Right now when I run curlGetUrl, I get 

    [errmsg] => Unknown SSL protocol error in connection to localhost:-9847

scottcwilson:
The code you are talking about affects whether the zc_install URL is SSL.  The code I'm talking about is building the URLs for the configure.php files.  That's where my change is.

scottcwilson:
I updated the logic so it uses the ping response (which drives whether the page is loaded via SSL or not) in the decision as to whether to create URLs in configure.php with https:// or with http://.

zc_install/includes/functions/general.php
```diff
@@ -199,8 +199,13 @@ function getDetectedURIs($adminDir = 'admin') {
...
     $catalogHttpUrl = 'http://' . $httpServer . '/' . zen_parse_url($url, 'path', true);
-    $catalogHttpsServer = 'https://' . $httpServer;
-    $catalogHttpsUrl = 'https://' . $httpServer . '/' . zen_parse_url($url, 'path', true);
+    if ($request_type == 'SSL') { 
+      $catalogHttpsServer = 'https://' . $httpServer;
+      $catalogHttpsUrl = 'https://' . $httpServer . '/' . zen_parse_url($url, 'path', true);
+    } else {
+      $catalogHttpsServer = 'http://' . $httpServer;
+      $catalogHttpsUrl = 'http://' . $httpServer . '/' . zen_parse_url($url, 'path', true);
+    }
```
drbyte: In the `else` can we just set them to the non-https variable?  This helps it be slightly clearer about the intent (think semantic), and eliminates a bit of duplication.

ie: 
```php
} else {
  $https = $http;
  $httpsUrl = $httpUrl;
}
```

scottcwilson:
That logic doesn't work; it gives the default text in the https locations.

drbyte:
I don't understand.

drbyte:
Why can't lines 206-207 simply point to the values set in 200-201?

scottcwilson:
screen shot 2016-06-19 at 6 08 40 pm

It doesn't fill in the line.  Oh, unless what you were writing was pseudocode and you meant 

```
    } else {
      $catalogHttpsServer = $catalogHttpServer;
      $catalogHttpsUrl = $catalogHttpUrl;
```

drbyte:
Yes, I was just giving an example, instead of writing the code for you ;)

scottcwilson:
LOL sorry about that, chief.  I took you too literally. :) 
Done in a minute.

drbyte:
No worries. Sometimes it's tough to know how specific to get ;)

drbyte:
👍 
Looks good. 
Is this now working the way you expected on your localhost? ie: solving the problem you were originally experiencing?

scottcwilson:
Works like a dream!  
screen shot 2016-06-19 at 6 37 19 pm

zcwilt:
you could simplify code even further and remove the need for the else logic 

e.g. 

     $catalogHttpsServer = $catalogHttpServer;
     $catalogHttpsUrl = $catalogHttpUrl;
     if ($request_type == 'SSL') { 
       $catalogHttpsServer = 'https://' . $httpServer;
       $catalogHttpsUrl = 'https://' . $httpServer . '/' . zen_parse_url($url, 'path', true);
     }

scottcwilson:
Ian: I think the format that was used and merged is actually better because then you can see at a glance that something untoward is happening. In fact, in a future refinement, I'd like to add a message there just in case the person isn't aware that this is a problem.

drbyte:
I’m not sure I agree.

But, that aside, what sort of message are you contemplating?

scottcwilson:
"HEY, we're using http because your site seems not to support https.  This is not best practice, see <link>." 

or something to that effect.

drbyte:
k.

We already display this when zc_install is run over http instead of https and when it can’t detect working ssl to redirect to automatically:

```php
define('TEXT_ERROR_HTTPS', 'PRO TIP: If possible you should already have installed an SSL certificate, and run the installer using https://');
```

PR #1005, opened 2016-06-19
Optimizing some category-lookup-related functions

drbyte:
Reduce some cyclomatic complexity of some functions by removing (some) duplication.
Change treatment of integer values to be integers instead of strings.
Deprecate some duplicates.

(Addendum to #967 which consolidated functions files)


scottcwilson:
👍

zcwilt:
:+1: 

with a reservation 

We have lots of functions that do some sql, and also use some boolean flags 
eg the sql is built differently based on the boolean flag

Lets not do this anymore. 

let's create 2 functions that create different sql, and name them appropriately.

And if we have booleans that create 6 paths through code, let's create 6 functions.

drbyte:
Agreed. But that's a monstrously larger refactoring exercise if we're going to adapt legacy code.

zcwilt:
Yes, wasn't suggesting we refactor legacy code, just moving forward with new code

PR #1010, opened 2016-06-23
Renaming file to remove spaces from filename

scottcwilson:
Files with spaces are problematic for Linux/Mac users.


drbyte:
Hmmmm ... `.psd` ... is that file even used?

scottcwilson:
It's not used in the cart; I figured it was preserved as a documentation artifact.

drbyte:
Eventually we'll probably just delete that `psd` folder.
I don't want to set a precedent that invites a million PRs to clean up the AdminLTE stuff ... cuz we're planning to "lean it out" before final release.

drbyte:
Did you run into this because of a problem?  Or just "oh, I'm using Linux, and spaces are usually a problem, so I should fix that before it becomes a problem"?

scottcwilson:
`find . <some pattern> `
gives me a bunch of errors every time I run it because of this file.  It's quite annoying.

Issue #977, opened 2016-06-07
Considering: Remove admin "newsletters" and "product-notifications" screens


drbyte:
Contemplating removing the archaic and unmaintained "newsletter" page from the admin ... because there are much more reliable and modern newsletter-sending tools available.
Same with product-notifications, since it's really lame :)

We can still "track" the associated opt-ins as we do now, and provide an exporter for extracting relevant data to a 3rd-party system.

Thoughts?  Will anybody "miss" the old tool?

scottcwilson:
:+1:  Do it.

proseLA:
just so i understand what you are asking, you want to remove the following files from the admin side:

newsletters.php
includes/modules/newsletters/newsletter.php
includes/modules/newsletters/product_notification.php
includes/languages/english/newsletters.php
includes/languages/english/modules/newsletters/newsletter.php
includes/modules/newsletters/product_notification.php
includes/languages/english/modules/newsletters/product_notification.php

remove the selection from the admin -> tools menu and do some remaining cleanup.  the customer class/table remains unchanged due to wanting to export to some 3rd party tool, ie constant contact, mailchimp, etc.

is there something else i am missing?

the one thing i will say which can get confusing for the customer is allowing them to unsubscribe from the ZC website and the store owner does not unsubscribe them from the 3rd party tool.  most of the store owners focus on the subscription in my opinion.  and the 3rd party providers generally have good tools for subscribe/unsubscribe.  but that is neither here not there....  just an observation.

drbyte:
> the thing which can get confusing for the customer is allowing them to unsubscribe from the ZC website and the store owner does not unsubscribe them from the 3rd party tool

Thoughts on managing that?

One "partial" way to address it could be to integrate an export tool such as: https://www.zen-cart.com/downloads.php?do=file&id=6

Are additional "hooks" (notifier calls) necessary on the non-admin side to allow "unsubscribes" to auto-trigger the same on 3rd party services?

proseLA:
wow, and to think i coded the email export all from scratch!  ;)

my opinion would be to remove the option completely from a customer being able to change it on ZC.  and to not have an unsubscribe integration to 3rd party services.

[...]

PR #1013, opened 2016-06-24
removes admin newsletters and product notifications #977

proseLA:
if we want to incorporate a 3rd party mailing list add-on; i think it should be part of a separate PR.


drbyte:
Great!

I'd suggest also removing the `newsletters` table from the zc_install/sql/mysql_zencart.sql file.

And also this line from that same file, since it drives the admin menu option which you've removed from the language file:
`('newsletters', 'BOX_TOOLS_NEWSLETTER_MANAGER', 'FILENAME_NEWSLETTERS', '', 'tools', 'Y', 5),`

As for v160 upgrade SQL, it's probably best not to delete the `newsletters` table there, since we don't want to delete peoples' data.
But I think you could safely add something like these to remove the (now broken) menu option:
```sql
delete from admin_pages where page_key = 'newsletters';
delete from admin_pages_to_profiles where page_key = 'newsletters';
```

proseLA:
ok... i think i got that...  best.

proseLA:
interesting that on my local box, the removal of the menu option happened without the sql deletion.

drbyte:
That's because it found no language defines to match. 
And might have even thrown errors recorded in the logs directory.

proseLA:
interesting.  fyi, ZC did not throw any errors in the log directory.

i was getting errors from a missing table (countries_name); which i tracked down to the 'add files' commit.  which frankly took some time...  it seems that the commit message is more important than the branch name, as after a merge, we are left with the commit message.  and i'm not sure if others v160 test stores are getting the error, or they manually added the table as i did.

drbyte:
> interesting. fyi, ZC did not throw any errors in the log directory.

Not surprising... We try to anticipate that so errors aren't logged. 
> i was getting errors from a missing table (countries_name); which i tracked down to the 'add files' commit. which frankly took some time... it seems that the commit message is more important than the branch name, as after a merge, we are left with the commit message. and i'm not sure if others v160 test stores are getting the error, or they manually added the table as i did
> 

Ya, the dev/alpha code won't always provide a clean update path including database schema changes. Those gave to be done manually; or just reinstall regularly. ;)

drbyte:
👍

zcwilt:
:+1:

PR #1014, opened 2016-06-25
Space will fail validation

Zen4All:
The extra space gives an error on code validation (in NetBeans). I am not sure is it will actually fail to load the gradient in IE, but all the examples I examined do not have this extra space.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1014)


drbyte:
I'm content to merge this. NOTE: I will probably completely replace the AdminLTE folder with an updated version.
👍

PR #1025, opened 2016-06-27
Remove legacy stats_products_purchased file in favour of new

scottcwilson:
- removed unused file
- noted location of new logic in the languages file.


drbyte:
👍 

But I'm not sure we need to put the new LEAD file location into every language file.

scottcwilson:
The trouble is, they may not know to look in another file to figure out where things are.  I had trouble figuring this out myself, so I thought others would benefit from this short cut.

zcwilt:
:+1:

PR #1030, opened 2016-06-27
Consolidate admin and catalog order class

drbyte:
Combine the two (nearly identical) classes into one, to reduce duplication and make things easier to maintain.


zcwilt:
:+1:

ajeh:
:+1:

PR #1032, opened 2016-06-28
Split the nginx config into input-template and output-conf files

drbyte:
When zc_install replaces the original file every time it is run, it makes code maintenance complicated. This splits it out into separate files: a template, and an output file.


zcwilt:
:+1:

ajeh:
:+1:

PR #1034, opened 2016-06-28
Reorder load-order of components in checkout_process

drbyte:
In rare and unexplained circumstances tax calculations may not fire properly on shipping.
Changing the load-order to match the order used in other stages of checkout appears to help prevent this.

Long discussion at https://www.zen-cart.com/showthread.php?220280-Shipping-tax-not-added-in-order-confirmation-possible-bug&p=1312461#post1312461


zcwilt:
:+1:

ajeh:
:+1:

PR #1037, opened 2016-06-29
Add Sales Graph report

drbyte:
Adapted from a GPL community contribution, but with major refactoring.

While functional now, it could stand some TLC to pretty it up a lot yet.


zcwilt:
:+1:

PR #1038, opened 2016-06-29
Prevent fatal error if widget class is missing

drbyte:
If user's widget is defined in db table but class is missing, a fatal error was occurring.
Now it just displays an empty widget without crashing.


scottcwilson:
👍

zcwilt:
:+1:

PR #1047, opened 2016-06-30
orderby and groupby clashing in report generator

### LINKED ISSUE #1040: GROUP BY and ORDER BY used in wrong order by LEAD query
If both `groupBys` and `orderBys` are specified in `$listingQuery`, they are used in the wrong order when the query is built, leading to a 1064 SQL Syntax error.

scottcwilson:
fix required by: Stats customers report should be ordered by total descending

Fixes #1040


drbyte:
👍

ajeh:
:+1:

PR #1048, opened 2016-07-02
Warn the user on the system_setup page to get a certificate

scottcwilson:
Uses a more intense color and tells them follow up will be needed:

```
'As soon as possible, you should install an SSL certificate, and update your configure.php files.'
```


zcwilt:
:+1:

ajeh:
:+1:

PR #1052, opened 2016-07-03
Protect against undefined widget/page defines in admin menu functions

drbyte:
Some calls to `constant(foo)` were throwing errors if certain language defines were not present.

These changes simply skip those entries instead of throwing errors. 

(Alternatively a future enhancement could be to throw an exception or log an error instead, to help people troubleshoot why their widgets or menu items aren't showing up.)


ajeh:
:+1:

scottcwilson:
@drbyte If you give me some examples  you used to test with, I can create some documentation on "things to look for when your widget isn't working."

PR #1053, opened 2016-07-03
Ignore git changes to zencart_ngx_server.conf

drbyte:
The `.gitignore` wasn't working previously because the `.conf` file was still under version control.  This removes it from version control.


scottcwilson:
👍

zcwilt:
:+1:

PR #1056, opened 2016-07-04
Automatically disable session.use_trans_sid

drbyte:
Given that `session.use_trans_sid` can be disabled at the application level, let's do that automatically, instead of leaving it up to the end-user to have to figure out how to change `php.ini`


zcwilt:
:+1:

ajeh:
:+1:

Issue #832, opened 2016-02-16
Save order's language in the database


Zen4All:
@lat9

> For multi-lingual stores, it would "be nice" if the language used during an order's checkout was captured in the database as part of the order processing. This would be an aid in follow-on communications with the customer and in the re-construction of the order information. 

quoted from:
[https://www.zen-cart.com/showthread.php?219536-Save-order-s-language-in-the-database&p=1303961#post1303961](url)

This would be the starting point to discus what is needed to accomplish to set the language, where it is being used, viewed, and/or edited

Zen4All:
I think not only the order language should be set, but also the preferred language a customer wants to use for his/her account.

The language should then probably be set in:

- the orders table for order language
- customers or customers_info table for preferred communications language, and maybe default site language

This could be done at account creation and order creation.

From the back end:
- the administrator should be able to see the preferred language in the customers page, and be able to edit the value.
- orders could/should be displayed in the customers language
- invoice and packing should be displayed in the customers language
- the automatic emails after status updates etc, should be in the customers language.

Some thoughts as a strart

drbyte:
I went thru the same thinking process when considering this for 2.0 ... and it actually gets to be monstrous.

So, as far as code implementation, take it in baby-steps.  Small feature steps will be easier to manage.

Zen4All:
first steps are being taken here: [https://github.com/Zen4All/zc-v1-series/tree/v160-Save-order-language](url)
The language used during checkout is being saved with the order in the orders table

torvista:
My opinion.
The only place this really matters (in vanilla zc) is in the sending of emails from an order update.
Orders page: I show the country of the order in the listing and add a UK flag next to the country name if the order was in english as 99% of our orders are in spanish.

More usefully I use a modified info box to send the emails from the listing page and that has a flashing flag next to the "send Email" box if the admin language is not the same as the order language as a warning. 
![admin_orders_email_flag](https://cloud.githubusercontent.com/assets/4391026/13807654/7b410e0a-eb62-11e5-8d09-f595c83150d6.gif)

Same idea in the order details page.
![admin_orders_email_flag_order](https://cloud.githubusercontent.com/assets/4391026/13807692/bcf36532-eb62-11e5-8bb9-dc75c164d77c.gif)

That has proved sufficient and I think worthy of consideration until the whole issue is fixed.

[...]

PR #1058, opened 2016-07-04
Order language

Zen4All:
This update captures the language in which the order was made. This is
displayed in the admin orders page. Also the subsequent emails like
order status updates are send in the language that was used during the
order creation.
Also see #832

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1058)


drbyte:
@Zen4All  you said "Also see #832".  

Do you feel this PR is feature-complete enough to close #832 when it is merged?

admin/orders.php
```diff
@@ -1022,6 +1036,11 @@ function couponpopupWindow(url) {
         if (zen_not_null($oInfo->last_modified)) $contents[] = array('text' => TEXT_DATE_ORDER_LAST_MODIFIED . ' ' . zen_date_short($oInfo->last_modified));
         $contents[] = array('text' => '<br />' . TEXT_INFO_PAYMENT_METHOD . ' '  . $oInfo->payment_method);
         $contents[] = array('text' => '<br />' . ENTRY_SHIPPING . ' '  . $oInfo->shipping_method);
+        $orderLanguageQuery = $db->Execute("SELECT name
+                                            FROM " . TABLE_LANGUAGES . "
+                                            WHERE languages_id = " . $oInfo->language_id);
+        $orderLanguage = $orderLanguageQuery->fields['name'];
+        $contents[] = array('text' => '<br />' . TEXT_INFO_ORDER_LANGUAGE . ' ' . $orderLanguage);
```
drbyte: Similar to the earlier note, this could be optimized to avoid a query by:
```php
$orderLanguage = $lng->get_language_data_by_id($oInfo->language_id);
$contents[] = array('text' => '<br />' . TEXT_INFO_ORDER_LANGUAGE . ' ' . $orderLanguage['name']);
```

admin/orders.php
```diff
@@ -115,6 +115,14 @@
...
         $comments = zen_db_prepare_input($_POST['comments']);
         $status = (int)zen_db_prepare_input($_POST['status']);
+        $order = new order((int)$oID);
+        $emailLanguageQuery = $db->Execute("SELECT directory
+                                        FROM " . TABLE_LANGUAGES . "
+                                        WHERE languages_id = " . (int)$order->info['language_id']);
+        $emailLanguage = $emailLanguageQuery->fields['directory'];
+
+        require(DIR_WS_LANGUAGES . $emailLanguage . '/orders_email.php');
+
```
drbyte: Can I recommend a couple optimizations here?

The query for `$emailLanguageQuery` can be avoided by using something like:
```
$order = new order((intO$oID);
$emailLanguage = $lng->get_language_data_by_id((int)$order->info['language_id']);
zen_load_language_file('orders_email.php', $emailLanguage['directory']);
```

... which depends on also rewriting the `zen_load_language_file()` in `functions_helpers.php` to the following ...
```
/**
 * Lookup specified language file, and load
 *
 * @param string $file filename of language file
 */
function zen_load_language_file($file, $language_directory = '~~') {
    if ($language_directory != '~~') {
      if (file_exists(DIR_WS_LANGUAGES . $language_directory . '/' . $file)) {
        include (DIR_WS_LANGUAGES . $language_directory . '/' . $file);
        return;
      }
    }
    if (file_exists(DIR_WS_LANGUAGES . $_SESSION ['language'] . '/' . $file)) {
        include (DIR_WS_LANGUAGES . $_SESSION ['language'] . '/' . $file);
        return;
    }
    if (file_exists(DIR_WS_LANGUAGES . 'english' . '/' . $file)) {
        include (DIR_WS_LANGUAGES . 'english' . '/' . $file);
    }
}
```

includes/classes/order.php
```diff
@@ -114,6 +114,7 @@ protected function query($order_id = 0) {
                         'last_modified' => $order->fields['last_modified'],
                         'ip_address' => $order->fields['ip_address'],
                         'order_weight' => $order->fields['order_weight'],
+                        'language_id' => $order->fields['language_id']
```
drbyte: Hmmm .... given the suggestions to use `$lng->get_language_data_by_id($id)` earlier, I wonder if it makes sense to set this once into the `order` object ... something like:

```php
$this->info = .... 'language'=>$lng->get_language_data_by_id($order->fields['language_id']),
```
and similarly in the `cart` function too.

Just would have to be careful when writing to the db that we actually store only the `int` of `language_id`;

I'm undecided whether to put both `language` (array) and `language_id` (int) into the `info` array. It wouldn't strictly be needed, as one could easily get it via `$order->info['language']['id']`.

The benefit to this is that we have access to `id`, `directory`, `name`, `code`, and even `icon` as sub-elements of the `['language']` element.


Would it also make sense to add `localized_orders_status_name` directly here (well, up around line `111`), instead of deferring that to the admin `orders.php` app?

admin/orders.php
Zen4All: I was looking for something like this, but did not know the function `get_language_data_by_id()`, that is much cleaner.
I will make this change, as also the change to `zen_load_language_file()` in the `functions_helpers.php`

admin/orders.php
Zen4All: This will also be updated

Zen4All:
This is not the complete feature as mentioned in #832, but a baby step towards it. My next step is to update the preffered language of the customer, which is more complicated

drbyte:
👍

drbyte:
:+1: Thanks. I like those changes.

For future SQL updates (in other PRs you're working on) let's use lowercase for `varchar`.

zcwilt:
:+1:

torvista:
I would suggest also recording the chosen shop language in the customers_basket table.
I use such a column to ensure customers get their email reminder in the same language they browsed in.
I also have had to add my own functions to get the language id from the code, and vice-versa, they would not go amiss...

Zen4All:
@torvista  That is being worked on

PR #1059, opened 2016-07-04
Retire the use of an Admin Configure.php file.

drbyte:
Auto-detect from anything not set in catalog configure.php

If one wishes, an `/admin/includes/local/configure.php` still works, in which one can define `ADMIN_URL` as an override of the auto-detections.

(I suppose one could also put `ADMIN_URL` in the catalog configure.php as an alternative.)


zcwilt:
:+1:

ajeh:
:+1:

drbyte:
Note: I've updated the code to honour the non-admin 'local' configure.php 

And ADMIN URL is replaced by ADMIN HTTP SERVER

PR #1068, opened 2016-07-05
Change cc fields to numbers

bislewl:
When entering Credit Card Numbers on a mobile device a qwerty keyboard appears. By converting the input fields to type="number". The number pad appears instead.


drbyte:
👍 
Note: for backporting to v155 when using the Responsive Classic template, one can add to the jQuery overrides to accomplish the same thing in `jscript_responsive_framework.php`, ie: https://github.com/zencart/zencart/blob/v155/includes/templates/responsive_classic/jscript/jscript_responsive_framework.php

zcwilt:
:+1:

PR #1069, opened 2016-07-06
Find Duplicate Models

scottcwilson:
Find Duplicate Models mod is a new report with edit capabilities.
I'm using it to assist in building the documentation for LEAD pages.


drbyte:
A couple observations/thoughts about consistency:

1. "reports" usually:
 - exist under the Reports menu, not Tools.
 - have a filename which begins as "stats_"

2. I think the menu "key" should be more specific than "findDuplicates", such as including "model" in its name.

Refs:
```
+define('FILENAME_FINDDUPMODELS', 'dup_models');
+define('BOX_REPORTS_FINDDUPMODELS', 'Find Duplicate Models');
```
```
+       ('findDuplicates', 'BOX_TOOLS_FINDDUPMODELS', 'FILENAME_FINDDUPMODELS', '', 'tools', 'Y', 100), 
```

scottcwilson:
Happy to change the key. 
The reason for the positioning in Tools is that it's not just a static report - it offers edit capability so you can fix the duplicates.  Still want it moved?

drbyte:
I have no objection to it being considered a tool. Just don't call it a report elsewhere then ;) ... ie: `BOX_REPORTS_FINDDUPMODELS`

Hmmm, I see you've used `BOX_TOOLS_FINDDUPMODELS` in the install SQL, but not in the language file.

drbyte:
(And it's best-practice to leave off the trailing `?>` at the end of files loaded prior to template-output, so the extra_datafiles file should have it removed.)

zcwilt:
:+1:

drbyte:
👍

PR #1074, opened 2016-07-07
Fixes #1062 misnamed constant in admin_account

### LINKED ISSUE #1062: misnamed constant in admin_account
<th class="name"><?php echo TEXT_NAME ?></th>
should be TEXT_ADMIN_NAME

also on a password change, the constant SUCCESS_PASSWORD_UPDATED is not found as it only exists in customers.

scottcwilson:


drbyte:
If users.php is also loaded on this page, should the pwd define go there instead? Does this problem exist on that screen too?

zcwilt:
:+1: 
While  I agree that there may be a bigger picture regarding some language defines, that should be addressed separately.

scottcwilson:
@drbyte The problem does not exist on users.php because the password isn't changed there - confirmed just now with current 1.6.0 code.

zcwilt:
@scottcwilson 
while this has been merged, I think what @drbyte  was referring to was the original code 
e.g. 

    if (file_exists(DIR_WS_LANGUAGES . $_SESSION['language'] . '/' . 'users.php')) {
      include(DIR_WS_LANGUAGES . $_SESSION['language'] . '/' . 'users.php');
    }

It should also be noted that users should allow for a password change. However we don't have a good partial to do this yet.

scottcwilson:
Gotcha - I just wanted to set his mind at ease that there was not an issue at the current time as far as I can see.  I just did a password change in 1.6.0 admin, and the correct message was issued.

drbyte:
> The problem does not exist on users.php because the password isn't changed there - confirmed just now with current 1.6.0 code.

Thanks. 

We'll have to clean up the awkward sharing of users.php, as @zcwilt mentioned.

Thanks for confirming it's fixed now. I proceeded with the merge for the same reasons you suggested: fix the symptom for now, and we can address the related issues later.

PR #1088, opened 2016-07-11
Who's Online shows indicators but not statistics

scottcwilson:
Fixed to show counts as well as red-yellow-green traffic lights.

### DIFF
--- a/admin/includes/template/dashboardWidgets/tplWhosOnline.php
+++ b/admin/includes/template/dashboardWidgets/tplWhosOnline.php
@@ -17,26 +17,31 @@
 </div>
 <div class="row">
   <div class="col-xs-4"><?php echo WO_REGISTERED; ?></div>
-  <div class="col-xs-2"><?php echo zen_wo_get_visitor_status_icon(0); ?>&nbsp;&nbsp;<?php echo $tplVars['widget']['users'][0]; ?></div>
-  <div class="col-xs-2"><?php echo zen_wo_get_visitor_status_icon(1); ?>&nbsp;&nbsp;<?php echo $tplVars['widget']['users'][1]; ?></div>
-  <div class="col-xs-2"><?php echo zen_wo_get_visitor_status_icon(2); ?>&nbsp;&nbsp;<?php echo $tplVars['widget']['users'][2]; ?></div>
-  <div class="col-xs-2"><?php echo zen_wo_get_visitor_status_icon(3); ?>&nbsp;&nbsp;<?php echo $tplVars['widget']['users'][3]; ?></div>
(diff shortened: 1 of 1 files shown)


admin/includes/template/dashboardWidgets/tplWhosOnline.php
```diff
@@ -17,26 +17,31 @@
...
 <div class="row">
-  <div class="col-xs-6"><?php echo WO_TOTAL; ?> <?php echo $tplVars['widget']['total']; ?></div>
-  <div class="col-xs-6"><?php echo '<a href="' . zen_href_link(FILENAME_WHOS_ONLINE) . '">' . WO_FULL_DETAILS . '</a>'; ?></div>
+  <div class="col-xs-12">&nbsp;</div>
+</div>
+<div class="row">
+  <div class="col-xs-12"><?php echo WO_TOTAL; ?> <?php echo $widget['total']; ?></div>
+</div>
+<div class="row">
+  <div class="col-xs-12"><?php echo '<a href="' . zen_href_link(FILENAME_WHOS_ONLINE) . '">' . WO_FULL_DETAILS . '</a>'; ?></div>
```
drbyte: Why all the additional col-xs-12 rows being added here?

scottcwilson: I think it looks nicer if the summary stats are each on their own line.  If you disagree, I can back this out.

drbyte:
I'm also trying to understand why you changed `$tplVars['widget']` to `$widget` ... when `$widget` is never defined ...

I already see counts in my whos-online widget. I'm not sure I understand what problem is being fixed here.

drbyte:
> I already see counts in my whos-online widget. I'm not sure I understand what problem is being fixed here.

Oy ... well, I did before I did a fresh install. The new install has no counts even after starting some sessions in other windows/browsers.  Okay that's insanely weird.

scottcwilson:
OK.  I have images of before and after if needed, but it sounds like you are convinced.

drbyte:
Ya, definitely weird. Esp since all the counts were showing when I did the original PR for this widget too.

admin/includes/template/dashboardWidgets/tplWhosOnline.php
drbyte: > I think it looks nicer if the summary stats are each on their own line. If you disagree, I can back this out.

I'm not bothered by adding the extra blank line before the totals.
But I think the "more details" link is confusing when it's right under the totals.  Maybe making the class be "col-xs-6 right" for the link might create the separation you're looking for?

drbyte:
👍

zcwilt:
:+1:

PR #1090, opened 2016-07-12
Provides ability to have full-width messages

scottcwilson:
This will be good for headers or footers.  In the case of logs, the
last message was squeezed  because it didn't fit in a single column.


zcwilt:
:+1:

drbyte:
👍

PR #1097, opened 2016-07-14
Add utility sql script to remove demo data from db.

scottcwilson:
This allows iterative testing of demo data.


drbyte:
Caveat: If someone has started with demo data in their store, and then gone live (perhaps after manually deleting or disabling some demo products but not cleaning everything) and then they run this thinking it will remove `only` the demo data, it's gonna wipe out customer data and also the real products/categories/attributes that they've added.

How would you propose is best to handle this caveat?

scottcwilson:
If they have gone live, they will have removed zc_install.   I could put a README file in this directory if you insist, but I don't really see the average person spelunking through this directory.

drbyte:
Fair points.
Maybe just a couple lines of comments at the top of the file would suffice

EXAMPLE ONLY:
```sql
# hey bozo, don't run this unless you want to wipe out all your data
# including categories, products, customers, ezpages, manufacturers, etc
```

zcwilt:
:+1:

drbyte:
I suggest also converting the `DELETE FROM` to `TRUNCATE TABLE`, so that autoincrement counters are properly reset too.

drbyte:
👍

PR #1098, opened 2016-07-15
Fix incorrect selector in checkout_payment form-submit jQuery

drbyte:
Fix posted by  @lat9  ref https://www.zen-cart.com/showthread.php?220658


zcwilt:
:+1:

PR #1099, opened 2016-07-15
Remove `latin1` support from zc_install

drbyte:
`utf8` is pretty much ubiquitious now; there's no need to offer `latin1` at install time; 
Further, the upgrade scripts operate only in utf8 now.


zcwilt:
:+1:

PR #1103, opened 2016-07-16
Fix banner statistics widget

scottcwilson:
Without this change it issues a debug log and shows no graph.
Similar issue as was fixed in PR #1088


drbyte:
👍

Issue #937, opened 2016-05-21
Change use of deprecated "zen_href_link" to "zen_admin_href_link" in Admin Folder


dakanji:
zen_href_link has been deprecated but still found throughout the code and instances should be changed accordingly.

PR #1106, opened 2016-07-16
Stop use of deprecated "zen_href_link" in Admin Folder - Updates #937

dakanji:
A few instances of the deprecated function have crept back in since the original PR.

This cleans these up and additionally, comments out the function with a view to ensure devs do not use it in the core code. This can be uncommented before releasing back into the wild and will make sure it is indeed deprecated in the core.

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -8,8 +8,7 @@
  */
 
   /**
-   * Deprecated and will be removed in later versions. 
-   * Use zen_admin_href_link instead
+   * @deprecated in version 1.6.0. Use zen_admin_href_link instead
    */
   function zen_href_link($page = '', $parameters = '', $connection = 'NONSSL', $add_session_id = true) {
    return zen_admin_href_link($page, $parameters, $add_session_id);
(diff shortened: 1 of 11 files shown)


dakanji:
Looks like the test suite needs updating to "zen_admin_href_link"

zcwilt:
do you feel capable of fixing the test suite. I can send change to do so

zcwilt:
The rest of the PR looks fine

dakanji:
Not worked with the test suite before but looks like it is the "testAdminUrlGeneration" part that needs amending which I can take a punt at. 

Obviously a bit tricky since zen_href_link is valid in the catalog side (perhaps this should be deprecated for zen_catalog_href_link as well at some point later to keep things tidy).

zcwilt:
:+1:

zcwilt:
Another option would be to trigger an E_DEPRECATED within the admin zen_href_link function, along with an @deprecated in the docblock (which might show in some IDE's)   
There is also the question of all of the plugins that might still be using zen_href_link in their code. 

I might open a separate issue/discussion around tidying this all up.

dakanji:
I wouldn't go so far as to emit warnings etc for deprecated API items and it should be enough to simply not use it in core and official plugin code.

There is nothing inherently wrong in 3rd party code using deprecated items as following the semantic versioning specification, deprecated stuff will stay until a v2.0.0 when backward incompatible changes come in.

The @depecated notice is a fair one and should be added.

drbyte:
👍

PR #1109, opened 2016-07-17
Consolidate admin order-update/search logic into shared order class

drbyte:
Move queries out of `/admin/orders.php` and share them from the main `order` class, as part of data automatically queried when an order is retrieved.


zcwilt:
:+1:

ajeh:
:+1:

PR #1114, opened 2016-07-18
Update Test Suite for zen_href_link - Updates #1106

dakanji:
Amended the test suite based on common sense but needs particularly proper checking.

Updates #1106 

Few notes: 

- Amended "spare" function to check for existence of zen_catalog_href_link - please verify
- No checks for zen_catalog_href_link usage found
- Review need for "function testAdminPageSsl" as not sure if still relevant but left in


drbyte:
Agreed: `testAdminPageSsl` segment could be removed.

dakanji:
Is that a confirmation to delete?

Also, perhaps there should still be a test for existence of zen_href_link as just deprecated and not removed.

Finally, not sure how/if the test suite actually tests for usage of a function in the code and fails if used (for deprecated stuff)

drbyte:
> confirmation to delete `testAdminPageSsl`

Yes.

> continue to test for `zen_href_link`

On admin side I wouldn't continue to test it for deprecation.  But the existing catalog-side tests for function of same name should be retained.

dakanji:
In that case, it would seem best to leave the existing zen_href_link tests in place and add equivalents for zen_admin_href_link.

Also means that the earlier suggestion in #1106 by @zcwilt to trigger a PHP error/warning to the zen_href_link function is the way to go. (Should have listened to more experienced counsel).

Looking at the phpunit xml file, it seems `convertErrorsToExceptions` attribute is at the default so that php errors/warnings will throw exceptions. 

This would seem to require adding an override to the `testAdminUrlGeneration` setup function so that the test suite itself does not throw an error while testing zen_href_link functions after adding the trigger_error notice.

drbyte:
Sigh. I don't understand what you've done with that last commit. Nor why. 
I read your comments, but I don't follow the logic.

I was ready to approve the previous commit, save for removing the duplicated admin SSL tests.  As it is now I don't agree. 👎

dakanji:
Fair enough. Unfortunately can't clarify now as responding from phone. The PR
is linked to #1106 which is basically incomplete although already merged.

Will try to clarify thoughts and link later.

zcwilt:
i'm confused as well. 
Before the last commit, I was happy with the PR

dakanji:
Would be too convoluted trying to clarify reasons save that the original aim of preventing the situation whereby usage creeps back in is not resolved with this.

In any case, reverted to the desired state.

drbyte:
👍

zcwilt:
:+1:

PR #1128, opened 2016-07-21
Fix minor code-review inconsistencies

drbyte:
These should fix small issues flagged up by code inspection

 - return types
 - missed array notation on param list
 - indentation inconsistency


zcwilt:
:+1:

PR #1133, opened 2016-07-21
Fix error-log on widget-install

drbyte:
Given that the process of adding a selected widget to the dashboard doesn't require rendering that widget until the page is ready, calling the template isn't necessary. And, since the prepared vars from the widget class haven't been loaded, the vars aren't ready for the template and thus an error is logged.  Alternate fix would be to run the `loadClass` method, but that appears to be wasted CPU activity, hence these lines are simply disabled at the moment.


zcwilt:
:+1:

Just a note, both the addwidget and deleteWidget processes need some updating after the move to using gridstack. In fact the whole dashboard widget stuff needs a good refactoring.

ajeh:
:+1:

PR #1134, opened 2016-07-21
Set minimum tested PHP version to 5.5

drbyte:
PHP 5.5 is officially EOL today, even for security updates. 
We support it only for the sake of convenience for people upgrading. 

... and we might even drop 5.5 support before final release.


scottcwilson:
I don't believe any documentation needs to be updated for this change. 👍

zcwilt:
:+1:

PR #1139, opened 2016-07-23
These messages were all used in admin/users.php

scottcwilson:
They were missing from includes/languages/english/users.php however.


zcwilt:
:+1:

drbyte:
👍

PR #1150, opened 2016-07-25
Retire remaining uses of zen_db_output() with zen_output_string_protected()

### LINKED ISSUE #270: Review all uses of htmlspecialchars and zen_db_output

drbyte:
Closes #270


zcwilt:
:+1:

ajeh:
:+1:

PR #1151, opened 2016-07-25
Fix rare invalid-string-offset errors in functions_email

drbyte:
Only triggered when passing a string for the array of content to HTML emails.

Thanks to @Dave224 https://www.zen-cart.com/showthread.php?220755


zcwilt:
:+1:

ajeh:
:+1:

PR #1153, opened 2016-07-26
Fixes #644 Non superuser can't access Banner Statistics

### LINKED ISSUE #644: A non-superuser Admin gets "Denied" when accessing Banner Statistics page
Permissions for `banner manager` don't also cascade to `banner statistics`.  

(And the `banner stats` widget permission setting isn't tied to the `banner statistics` *page*)

scottcwilson:


zcwilt:
:+1:

drbyte:
👍 okay with this approach for now.  If broader capability is required it can be expanded further at that time.

PR #1158, opened 2016-07-28
Fixes #1026 Add totals/sums to new Admin Reporting infrastructure

### LINKED ISSUE #1026: Add totals/sums to new Admin Reporting infrastructure
Currently the new LEAD reports have no idea about `totaling`.
Something needs to count sums and perhaps other values in order for other kinds of reports to be meaningful.

scottcwilson:


includes/library/zencart/QueryBuilderDefinitions/src/formatters/AbstractFormatter.php
```diff
@@ -116,6 +121,45 @@ public function getFormattedResults()
...
+        }
+        $hasTotal = false; 
+        foreach ($this->outputLayout['listMap'] as $id => $key) {
+           $fieldset = false; 
+           if ( isset($this->outputLayout['fields'][$key]) && 
+                isset($this->outputLayout['fields'][$key]['total']) ) {
+              $total = 0; 
+              if ($this->outputLayout['fields'][$key]['total'] == 'currencySum') {
+                 foreach ($this->formattedResults as $rec) { 
+                    $total += preg_replace("/([^0-9\\.])/i", "", $rec[$key]);
```
drbyte: In this regex, I don't think the `\\` is needed.

admin/includes/template/partials/tplAdminLeadItemRows.php
```diff
@@ -32,4 +32,16 @@
         <?php } ?>
     </tr>
 <?php } ?>
+<?php if ($tplVars['listingBox']['formattedTotals'] != null) { ?>
+<tr>
+<?php $cols = count($tplVars['leadDefinition']['listMap']) + 1; ?>
+<td colspan="<?php echo $cols;?>"><?php zen_draw_separator(); ?></td>
```
drbyte: I'm not sure I understand why you're using `zen_draw_separator` here.

drbyte:
This is looking good.

includes/library/zencart/QueryBuilderDefinitions/src/formatters/AbstractFormatter.php
scottcwilson: Fair enough, it works without that.  I will remove.

admin/includes/template/partials/tplAdminLeadItemRows.php
scottcwilson: The idea was to create a line below the data and the sum.

admin/includes/template/partials/tplAdminLeadItemRows.php
```diff
@@ -32,4 +32,16 @@
         <?php } ?>
     </tr>
 <?php } ?>
+<?php if ($tplVars['listingBox']['formattedTotals'] != null) { ?>
+<tr>
+<?php $cols = count($tplVars['leadDefinition']['listMap']) + 1; ?>
+<td colspan="<?php echo $cols;?>"><?php zen_draw_separator(); ?></td>
+</tr>
+    <tr>
+        <td>&nbsp;</td> <!-- account for checkbox -->
```
zcwilt: Given this column always exists, could we put a title in here 
e.g. Page Total instead of &nbsp; 
I've tried this and it does make the layout look a bit weird but that's a css issue, and we can fix that later

admin/includes/template/partials/tplAdminLeadItemRows.php
drbyte: Okay. Probably best to use CSS for that, such as a corresponding `class` on that cell.

I've never mentioned it yet, but I'd like to retire `zen_draw_separator` because it's using the old-school approach of images for separation and spacers.

scottcwilson: Is there anything available now?  I didn't see anything obvious.

drbyte: probably `td class="underline"`

and then add to main stylesheet:
```css
td.underline {
    border-bottom: 1px solid #000;
    padding: 0px 7px;
}
```

dakanji:
> I'd like to retire zen_draw_separator because it's using the old-school approach of images for separation and spacers.

How about using HTML in some cases?

I was going to have some convenience functions such as

```
  function zen_horizontal_html_spacer($spaces = 2, $class=null) {
  	
  	$html = zen_not_null($class) ? '<span class="' . zen_output_string(trim($class)) . '">' : '<span>';
  	for ($x = 1; $x <= $spaces; $x++) {
  		$html .= "&nbsp;";
    }
  	$html .= "</span>";
    return $html;
  }

  function zen_vertical_html_spacer($spaces = 2, $class=null) {
  	$class = zen_not_null($class) ? ' ' . zen_output_string(trim($class)) : '';
  	for ($x = 1; $x <= $spaces; $x++) {
  		$html .= '<div class="row' . $class . '">&nbsp;</div>';
    }
    return $html;
  }
```

These are then called like so:

```
<?php 
echo zen_vertical_html_spacer(1);
echo zen_vertical_html_spacer(1, "hidden-xs");
?>
```

In this example, one vertical space is inserted for small devices and two for larger screens.

I'll have this in my proposal for the login form which I'll push soon. Perhaps you can review then.

drbyte:
@dakanji I'm not understanding the complete context about how often you'd need to use such helpers, but it really has nothing to do with this PR.

zcwilt:
@scottcwilson 
you are doing <?php zen_draw_separator(); ?>
this has no effect on output as you would have to do
<?php echo zen_draw_separator(); ?>

given that, I would juts remove it

includes/library/zencart/QueryBuilderDefinitions/src/formatters/AbstractFormatter.php
```diff
@@ -134,7 +134,7 @@ public function getFormattedTotals($outputLayout)
               $total = 0; 
               if ($this->outputLayout['fields'][$key]['total'] == 'currencySum') {
                  foreach ($this->formattedResults as $rec) { 
-                    $total += preg_replace("/([^0-9\\.])/i", "", $rec[$key]); 
+                    $total += preg_replace("/([^0-9\.])/i", "", $rec[$key]);
```
drbyte: FYI, inside the square brackets of a regex pattern like this the only chars that would need escaping would be `]` (because it ends the square-bracket pattern), or a `-` if it isn't the last character (because it might be confused with specifying a character range like a-z etc).  So technically the `.` doesn't need the `\` in this case.

zcwilt:
:+1:

drbyte:
👍

PR #1159, opened 2016-07-28
Upgrade Font-Awesome in Admin to v4.6.3

dakanji:
Updating this as first step in admin styling.


drbyte:
👍 
It's a moving target, and like all other packages that are integrated, will need another final update before final release.

PR #1164, opened 2016-07-29
 On update in-cart quantity remains if qty still correct 

### LINKED ISSUE #1163: In cart quantity modified on shopping cart page update
In an update performed in commit https://github.com/zencart/zencart/commit/7f975f39e154a75fac5e63f9cfcca9040ceb8e84
as a part of [pull 92](https://github.com/zencart/zencart/pull/92) an attempted fix for notifying customers that the quantity in the cart had been modified was applied.  The code works fine when product is added to the cart, but when the cart is updated (pressing the "refresh" button while in the cart). If the quantity of product that was in the cart plus the quantity that has been entered (even if unchanged) is greater than the existing stock of that product, then the quantity would be modified (reduced). If there were two product in the cart when this update was performed a sort of "chase" condition would occur where one is "corrected" the other would go wrong, try to fix, and role would reverse.

This is discussed on the ZC site at [this thread](https://www.zen-cart.com/showthread.php?220786-Update-Button-of-Shopping-Cart-Reduces-Quantity-in-Cart).

mc12345678:
Fixes #1163

In an update performed to notify customers that the quantity in the cart had been modified (primarily addressing when adding product to the cart), the same code had been applied to the actionUpdateProduct function which if before the refresh the quantity in the cart plus the quantity to which it is to be set is greater than the available stock, then the quantity to be presented would be modified.

IE: store is setup to not allow going "negative" (allow checkout if out-of-stock is false), if a product has 10 in stock, 9 are added to the cart, the cart is then visited, and while on the cart if the refresh button is pressed, the quantity of that product would become 1. 

The above change removes that code but maintains notification to the customer if the quantity were adjusted for reasons such as attempting to add more than what is in stock or more than the maximum allowable or in other words if the quantity has been modified to be the quantity that currently existed in the cart or the maximum allowed.


includes/classes/shopping_cart.php
```diff
@@ -1808,6 +1800,13 @@ function actionUpdateProduct($goto, $parameters) {
           default:
             $adjust_max= 'false';
           }
+          
+// bof: notify about adjustment to new quantity to be same as current in stock or maximum to add
+          if ($adjust_max == 'true') {
+            $messageStack->add_session('shopping_cart', ($_SESSION['cart']->display_debug_messages ? 'FUNCTION ' . __FUNCTION__ . ': ' : '') . WARNING_PRODUCT_QUANTITY_ADJUSTED . zen_get_products_name($_POST['products_id'][$i]), 'caution');
```
drbyte: Curious: Why did you change to using `$_SESSION['cart']->display_debug_messages` here, when `$this` already refers to `$_SESSION['cart']` because it's the self-same object

mc12345678: Because of the environment I was in at the time of retypying "the line" where I have the above "incorrect" variable used in code outside of a class... 

Will correct and update the branch.

drbyte:
In chatting with @ajeh it seems this PR doesn't fully fix the problem.

mc12345678:
Which issue(s)/condition(s) remain unresolved?

drbyte:
Consider Die Hard:
If you put 1 in the cart, and then try to edit, you can't change it to 10, cuz it forces it back to either 1, or down to 9 (maybe it's replacing with the difference between the 1 already in and the 10 trying to be changed to?).

Ajeh can describe it better ... ;)

ajeh:
Try this ...

Go to Product and add 1 to cart
Now on Shopping Cart try to change to 10 and it goes to 1

Now remove product from cart
Go to Product and add 10 to cart it changes to 9
Try to change on shopping cart to 10 and it changes to 9

So it has inconsistent results and is not a complete fix

ajeh:
I do see 1 advantage to your fix ...

If you have an amount under 9 and try to go over 9 it returns back to what you had in the cart, rather than go to 1 but if you try to go over 9 it doesn't go to 9. However at least you get the message that it was adjust because you went over the minimum.

Example: start with 7 and try to go to 22 and you go back to 7. Then try 8 and you go to 8. Then try 9 you go to 9.

mc12345678:
So, first die hard when I went to it didn't have a max value assigned, but once I set/restored it then this issue comes into play.

The way the "max" checks work currently is that they effectively remove the customer's change from the playing field. So as @ajeh pointed out, adding 1 to the cart then attempting to modify the total amount to a value outside of the allowable returns the result back to what was originally entered, 1. Same would be true if 5 were initially added and then the total quantity was to be adjusted to 10, it would return to 5. 

It was not clear that the desire was to set the final quantity to the max if the customer tried to enter a value greater than the max available when doing an update. 

A problem I can see with doing that is when mixed product are involved and two or more mixed product are already in the cart, then if the first of those is taken to/above the max then it will win the quantity(ies) but the other(s) will go to zero for those that follow. The current arrangement would leave them as is... So, with that in mind, would the goal be to only have the currently viewed behavior of returning to previous entry when 2 or more mixed product (of the same product) are in the cart, and in other cases set the amount to the max allowable/available?

ajeh:
I agree that this is a better solution than being broken. 
:+1: 

Note: additional consideration should be thought about to allow the Max to be the result when attempting to go over that amount, rather then go back to the quantity that the customer had.

drbyte:
👍

mc12345678:
I have been continuing to consider such an additional restoration; however, there are some considerations to be given for mixed quantities, which if any individual entries changed, whether they increased or decreased, total resulting affect, etc... I have something that steps closer, but was working on a possible "ideal" solution yesterday. So far have identified 3 new functions to add to the shopping_cart class and some internal logic to be applied to the mixed quantity cases.

PR #1165, opened 2016-07-29
Switching to non-recursive category tree.

scottcwilson:
Stores with large numbers of categories can error during recursion.

### DIFF
--- a/includes/classes/category_tree.php
+++ b/includes/classes/category_tree.php
@@ -157,24 +157,22 @@ class category_tree extends base {
         }
         $this->box_categories_array[$row]['path'] = $cPath_new;
 
-        if (isset($cPath_array) && in_array($catID, $cPath_array)) {
-            $this->box_categories_array[$row]['current'] = true;
-        } else {
-            $this->box_categories_array[$row]['current'] = false;
-        }
+        $this->box_categories_array[$row]['current'] = 
+            (isset($cPath_array) && in_array($catID, $cPath_array)); 
 
         // display category name
-        # if parent is 0 don't add a prefix to the category name, otherwise repeat CATEGORIES_SUBCATEGORIES_INDENT (int)$this->tree[$catID]['level'] amount of times and add CATEGORIES_SEPARATOR_SUBS before the category name
-        $this->box_categories_array[$row]['name'] = ($this->tree[$catID]['parent'] == (int)TOPMOST_CATEGORY_PARENT_ID ? '' : str_repeat(CATEGORIES_SUBCATEGORIES_INDENT, (int)$this->tree[$catID]['level']) . CATEGORIES_SEPARATOR_SUBS) . $this->tree[$catID]['name'];
+        if ($this->tree[$catID]['parent'] == (int)TOPMOST_CATEGORY_PARENT_ID) { 
+           $this->box_categories_array[$row]['name'] = $this->tree[$catID]['name']; 
+        } else { 
+           $this->box_categories_array[$row]['name'] = str_repeat(CATEGORIES_SUBCATEGORIES_INDENT, (int)$this->tree[$catID]['level']) . CATEGORIES_SEPARATOR_SUBS . $this->tree[$catID]['name'];
+        }
 
         // make category image available in case needed
-        $this->box_categories_array[$row]['image'] = $this->tree[$catID]['image'];
+        $this->box_categories_array[$row]['image'] = 
+            $this->tree[$catID]['image'];
 
-        if (zen_has_category_subcategories($catID)) {
-            $this->box_categories_array[$row]['has_sub_cat'] = true;
-        } else {
-            $this->box_categories_array[$row]['has_sub_cat'] = false;
-        }
+        $this->box_categories_array[$row]['has_sub_cat'] = 
+            zen_has_category_subcategories($catID); 
 
         if (SHOW_COUNTS == 'true') {
             $products_in_category = zen_count_products_in_category($catID);
--- a/includes/classes/category_tree.php
+++ b/includes/classes/category_tree.php
@@ -157,22 +157,18 @@ class category_tree extends base {
         }
         $this->box_categories_array[$row]['path'] = $cPath_new;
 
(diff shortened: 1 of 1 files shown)


drbyte:
I'm okay with the logic here.
I'm inclined to change the code style a bit to remove the arrow-pattern of nested if/else stuff when all that's being done is set true/false, and split long ternary usage into separate statements particularly where one result adds an empty string, and then after the ternary is done another string is added to the overall result.

scottcwilson:
:+1: Will do.

includes/classes/category_tree.php
```diff
@@ -157,24 +157,22 @@ function zen_show_category($first_element) {
...
+            (isset($cPath_array) && in_array($catID, $cPath_array)); 
 
         // display category name
-        # if parent is 0 don't add a prefix to the category name, otherwise repeat CATEGORIES_SUBCATEGORIES_INDENT (int)$this->tree[$catID]['level'] amount of times and add CATEGORIES_SEPARATOR_SUBS before the category name
-        $this->box_categories_array[$row]['name'] = ($this->tree[$catID]['parent'] == (int)TOPMOST_CATEGORY_PARENT_ID ? '' : str_repeat(CATEGORIES_SUBCATEGORIES_INDENT, (int)$this->tree[$catID]['level']) . CATEGORIES_SEPARATOR_SUBS) . $this->tree[$catID]['name'];
+        if ($this->tree[$catID]['parent'] == (int)TOPMOST_CATEGORY_PARENT_ID) { 
+           $this->box_categories_array[$row]['name'] = $this->tree[$catID]['name']; 
+        } else { 
+           $this->box_categories_array[$row]['name'] = str_repeat(CATEGORIES_SUBCATEGORIES_INDENT, (int)$this->tree[$catID]['level']) . CATEGORIES_SEPARATOR_SUBS . $this->tree[$catID]['name'];
+        }
```
drbyte: Can this collapse further? 
ie:

```php
if ($this->tree[$catID]['parent'] != (int)TOPMOST_CATEGORY_PARENT_ID) { 
    $this->box_categories_array[$row]['name'] = str_repeat(CATEGORIES_SUBCATEGORIES_INDENT, (int)$this->tree[$catID]['level']) . CATEGORIES_SEPARATOR_SUBS;
}
$this->box_categories_array[$row]['name'] .= $this->tree[$catID]['name']; 
```

includes/classes/category_tree.php
```diff
@@ -157,24 +157,22 @@ function zen_show_category($first_element) {
         }
         $this->box_categories_array[$row]['path'] = $cPath_new;
 
-        if (isset($cPath_array) && in_array($catID, $cPath_array)) {
-            $this->box_categories_array[$row]['current'] = true;
-        } else {
-            $this->box_categories_array[$row]['current'] = false;
-        }
+        $this->box_categories_array[$row]['current'] = 
+            (isset($cPath_array) && in_array($catID, $cPath_array));
```
drbyte: This can still be one line.

includes/classes/category_tree.php
```diff
@@ -157,24 +157,22 @@ function zen_show_category($first_element) {
...
+        if ($this->tree[$catID]['parent'] == (int)TOPMOST_CATEGORY_PARENT_ID) { 
+           $this->box_categories_array[$row]['name'] = $this->tree[$catID]['name']; 
+        } else { 
+           $this->box_categories_array[$row]['name'] = str_repeat(CATEGORIES_SUBCATEGORIES_INDENT, (int)$this->tree[$catID]['level']) . CATEGORIES_SEPARATOR_SUBS . $this->tree[$catID]['name'];
+        }
 
         // make category image available in case needed
-        $this->box_categories_array[$row]['image'] = $this->tree[$catID]['image'];
+        $this->box_categories_array[$row]['image'] = 
+            $this->tree[$catID]['image'];
```
drbyte: again, one line is fine here

... and again below too

includes/classes/category_tree.php
```diff
@@ -137,64 +137,60 @@ function zen_category_tree($product_type = "all") {
...
-    if ($this->tree[$counter]['next_id'] != false) {
-      $ii++;
-      $this->zen_show_category($this->tree[$counter]['next_id'], $ii);
+        if (SHOW_COUNTS == 'true') {
+            $products_in_category = zen_count_products_in_category($catID);
+            if ($products_in_category > 0) {
+                $this->box_categories_array[$row]['count'] = $products_in_category;
+            } else {
+                $this->box_categories_array[$row]['count'] = 0;
+            }
```
drbyte: Could this be shortened to:

```php
            $this->box_categories_array[$row]['count'] = 0;
            $products_in_category = zen_count_products_in_category($catID);
            if ($products_in_category > 0) {
                $this->box_categories_array[$row]['count'] = $products_in_category;
            }
```
or even just:
```php
            $this->box_categories_array[$row]['count'] = max(0, zen_count_products_in_category($catID));
```

scottcwilson: The first form does an unnecessary assignment when products in category is non zero. 
The second is at least as confusing as a ternary operator.
Why are you saying these are better than what's there now?

drbyte: Avoiding the unnecessary if/else

Will `zen_count_products_in_category()` ever give a value less than 0 ?   Is the check for 0 even needed?

drbyte:
👍 

I'm curious about this comment: 
>        # NOTE: $this->tree[$catID]['parent'] is a string, but would make more sense as integer - consider forcing (int) when setting the var

... Everywhere `'parent'` is set it's coming as a result from the db, where it is stored in an `integer` field in MySQL. Are you saying that MySQL is returning the int as a string? Or are you saying something else in this class is converting it to a string first?

scottcwilson:
I took this whole function straight from a forum post.  I didn't edit it because you said it was good to merge as long as I used the constant instead of 0 for the top level cat.

drbyte:
No worries. 
I'm fine with it as-is, although it inspires further cleanup of the whole class.

Thanks for submitting it, and indulging the cleanup.

Issue #404, opened 2015-05-01
[Proposal] Stop requiring /admin/ folder rename


drbyte:
Feedback welcome ...

Proposal: Instead of throwing a roadblock up if the /admin/ folder is not renamed, instead just display a warning message inside the admin after logging in.

wiztechinc:
A roadblock it may be but it's dang effective. A message? Hmm, I got
clients who would start yelling at me and I've seen clients who almost
never log into admin - they fill orders from the order email!
ᐧ

Delia Wilson Lunsford
WizTech, Inc., (formerly Delia Wilson Design, LLC.)
434-202-4307
Terms and Conditions for working with WizTech, Inc.
(http://www.teamwiztech.com/terms-conditions.php)

barco57:
Why would your clients yell at you Delia...shouldn't you have already renamed their admin folders for them...

wiztechinc:
Oh, no, not because I did anything wrong!!

zcwilt:
Just some background here, and reasons to support the issue.

A long time ago, ZC admin was pretty insecure :(
It had little in the way of xss/xsrf protection and was subject to some other horrible vulnerabilities.
Obfuscating the admin directory was, at the time a quick way of mitigating against this.

Since then the admin code has been security hardened (and tested thru multiple PA-DSS validations) 

The intent of this issue is not to stop store owners changing the admin directory, but to not `roadblock` the admin if they haven't.

There are a number of reasons why.

1. Security by obscurity is bad, and can lead to complacency in managing security issues.
2. Easier to document.
3. Moving to use shared configs/controllers for admin/store

wiztechinc:
Easier to upgrade, I must admit, if the folder is named admin. One thing that not having a standard admin folder name is that it cuts down on folks trying to hack in. So it's easier on the server (I suspect) since they are just getting a 404 in response now. Don't know if that really makes a difference or not but since I host sites as well, I appreciate them not finding what they are looking for.

Since plugins and custom code can lessen security, do we really want to open up the possibility of hacking again? I assume zen cart admin can be made vulnerable by those who don't know any better.

drbyte:
> @wiztechinc wrote: 
> open up the possibility of hacking again

Sigh. That choice of wording can be sorely misinterpreted by alarmists.

[...]

PR #1166, opened 2016-07-29
Enhance Developer Mode Installation Fixes #1162

### LINKED ISSUE #1162: zc_install folder removal requirement maintained in developer mode
In transitioning to using the developer mode outlined by @drbyte in https://github.com/zencart/zencart/issues/404#issuecomment-234328537, I notice that the zc_install folder removal requirement is still in force leading to gitignore issues.

This requirement should be overridden in developer mode.

dakanji:
Fixes #1162 
- Override zc_install folder removal requirement in developer mode
 - Introduce option to select developer mode during installation
 - Can live side by side with other options such as process outlined in https://github.com/zencart/zencart/issues/404#issuecomment-234328537

### DIFF
--- a/admin/includes/init_includes/init_admin_auth.php
+++ b/admin/includes/init_includes/init_admin_auth.php
@@ -18,7 +18,9 @@ $val = getenv('HABITAT');
 $habitat = ($val == 'zencart' || (isset($_SERVER['USER']) && $_SERVER['USER'] == 'vagrant'));
 
 // admin folder rename required
-if ((!defined('ADMIN_BLOCK_WARNING_OVERRIDE') || ADMIN_BLOCK_WARNING_OVERRIDE == '') && ($habitat == false))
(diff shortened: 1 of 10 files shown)


zcwilt:
This is a whole lot of code to do what we already do. 
e.g 
We have a DEVELOPER_MODE constant that is used in the installer to bypass admin rename and set a standard password.

We have a ADMIN_BLOCK_WARNING_OVERRIDE constant to bypass checks in admin if the zc_install directory has not been renamed.

Use of both of these relies on a physical act on behalf of a developer to enable them.

Having an option in the Installer means non developers might inadvertently  set it, and ultimately compromise their install. 

As such, without some more persuasion, I don't think I can agree with this PR.

:-1:

dakanji:
That appears to be a fairly hasty decision and is a bit disappointing but nevermind.  
Closing this PR and moving on.

PR #1178, opened 2016-08-02
Refactor admin customers queries and code to shared functions

drbyte:
Move a bunch of processing logic into functions file instead of admin/customers.php


zcwilt:
:+1:

ajeh:
:+1:

PR #1180, opened 2016-08-02
Include generic About Us page by default

drbyte:
Currently defaults to using a Define-Page, but contains notifier hooks to allow for an observer class to intercept and output something from EZ-Pages instead.


scottcwilson:
:+1: I am really glad to see "About Us" become a first class citizen in Zen Cart.

zcwilt:
:+1:

PR #1182, opened 2016-08-03
Fixes #229 Removes redirect from ot_coupon.

### LINKED ISSUE #229: order_total collect_posts redirects before allowing validation of other errors
I am wondering about the invalid coupon checking code under `collect_posts` in `ot_coupon`.
The redirect code prevents further `collect_posts` functions from being ran.

The example I have in this case is that we run coupon codes, rewards points and gift vouchers. Ideally you would have each run and determine whether there are any errors and then redirect back.

This is already somewhat covered in `checkout_confirmation` by this code:

``` php
if ($messageStack->size('checkout_payment') > 0) {
  zen_redirect(zen_href_link(FILENAME_CHECKOUT_PAYMENT, '', 'SSL'));
}
```

scottcwilson:
Doing a redirect within collect_posts() prevents other OT modules
from running their collect_posts() methods, as noted in the issue report.

### DIFF
--- a/includes/classes/order_total.php
+++ b/includes/classes/order_total.php
@@ -160,13 +160,20 @@ class order_total extends base {
   function collect_posts() {
     if (MODULE_ORDER_TOTAL_INSTALLED) {
       reset($this->modules);
+      $collect_posts_rc = true; 
       while (list(, $value) = each($this->modules)) {
         $class = substr($value, 0, strrpos($value, '.'));
         if ( $GLOBALS[$class]->credit_class ) {
           $post_var = 'c' . $GLOBALS[$class]->code;
           if ($_POST[$post_var]) $_SESSION[$post_var] = $_POST[$post_var];
-          $GLOBALS[$class]->collect_posts();
+          $rc = $GLOBALS[$class]->collect_posts();
+          if (isset($rc) && $rc != true) { 
+               $collect_posts_rc = false; 
+          }
         }
+      }
+      if (!$collect_posts_rc) { 
+            zen_redirect(zen_href_link(FILENAME_CHECKOUT_PAYMENT, '', 'SSL', true, false));
       }
     }
   }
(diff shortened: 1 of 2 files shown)


zcwilt:
@scottcwilson sorry meant to look at this earlier in the week.
One problem I see is that there now appears no where that does a redirect to checkout_payment if an error occurs in the collect_posts function.

scottcwilson:
Yes you are right and there were other misses of redirects too.

scottcwilson:
Should be good now.

includes/classes/order_total.php
```diff
@@ -160,14 +160,21 @@ function update_credit_account($i) {
...
-          $GLOBALS[$class]->collect_posts();
+          $rc = $GLOBALS[$class]->collect_posts();
+          if (isset($rc) && $rc != true) { 
+               $collect_posts_rc = false; 
+          }
         }
       }
+      if (!$collect_posts_rc) { 
+            zen_redirect(zen_href_link(FILENAME_CHECKOUT_PAYMENT, '', 'SSL', true, false));
+      }
```
drbyte: I'm guessing, but does your use of `rc` mean "return code"?

I think semantically you mean to track whether any error occurred when processing `collect_posts()` in the specified module.
In that context would it make the code more readable if `$collect_posts_rc` was `$collect_posts_success` and `$rc` was `$no_errors`?

Also, these changes to `order_total.php` will affect *all* order_total modules that have a `collect_posts()` method, but only `ot_coupon` has been updated in this PR to do set these return values. Shouldn't `ot_gv` also be updated similarly?

NOTE: `ot_gv` does its zen_redirect to CHECKOUT_FLOW instead of CHECKOUT_PAYMENT. I suspect `ot_coupon` should have been using the same CHECKOUT_FLOW redirect. Fortunately you've moved it here to order_total, but probably should update it to redirect to `FILENAME_CHECKOUT_FLOW step=payment`.

scottcwilson:
On reflection, I wonder if this is truly worth doing, since it can be broken by anyone coming along and introducing an order total module which does a redirect.  What @robryanx is asking for is really more of an architectural improvement (gather all error messages at once, rather than one by one) than a true defect.
d

drbyte:
Agreed

scottcwilson:
OK.  Closing this PR.  Please close issue #229.

PR #1183, opened 2016-08-04
Labels in checkout steps should not be links

scottcwilson:
They don't take you to the page shown, so it's misleading to
display them as links.


drbyte:
👍

zcwilt:
:+1:

PR #1184, opened 2016-08-06
Click on on admin link when timed out gives security error

scottcwilson:
Repro: Allow admin to timeout (or logout from a separate tab).
Then click on a command - for example
    admin/index.php?cmd=template_select
This redirects to URL:
    admin/index.php?cmd=login&camefrom=denied
And puts you on the security error page once you login.


zcwilt:
@scottcwilson I'd like to close this issue as the changes needed are more complex. 
Note, I actually introduced the bug with https://github.com/zencart/zencart/pull/1168
The problem is that we don;t want to do any redirects in admin auth if we get here via an ajax call. 
 As I said, my previous PR didn't address this completely.

scottcwilson:
ok

scottcwilson:
What if I resubmitted this, but with redirectto set to the home page?
Would that be ok?
It's just a bit startling to see the permission denied message.

zcwilt:
see https://github.com/zencart/zencart/pull/1187

PR #1189, opened 2016-08-09
Fixes #794 - different traffic light for installed but disabled

### LINKED ISSUE #794: status indicator(s) of admin/modules.php for shipping
Clarification of the "definition" of the various shipping status icons (green, yellow, red) or consistency among various admin screens with similar status indicators is encouraged.  

Note to self: Yellow identifies more than one potential condition, installed but not enabled shipping module (non-functional results) or module with zone definitions (functional but limited results).  A similar yellow condition occurs in other admin screens but typically as a caution that operation may not be possible or that something is missing, etc...

Focus of this issue identified here: https://www.zen-cart.com/showthread.php?219322-US-and-rest-of-the-world-shipping-configuration/page2

scottcwilson:


drbyte:
👍

zcwilt:
:+1:

PR #1190, opened 2016-08-10
session_write_close is a void function

scottcwilson:
It should not be used as a return value lest someone "check it".


zcwilt:
:+1:

PR #1193, opened 2016-08-11
Mc12345678 restore paypal incontext operation

### LINKED ISSUE #1192: PayPal Express InContext not fully implemented
In ZC 1.5.5, the newly added code for PayPal Express checkout to support PayPal In Context checkout does not properly implement the popup and remnant shadow box expected to be seen/tested as shown at http://mc12345678.com which was independently implemented by November 2014.  This can be compared to a base ZC 1.5.5 installation related to selecting PayPal Express at either the shopping_cart or login screen.

The ZC files do not include the javascript call expected by PayPal to support action on the button press page (shopping_cart and login when directed to login as part of the checkout) nor the resulting page displayed to customers upon return from PayPal (by ZC 1.5.5 base code: potentially checkout_shipping, checkout_payment, checkout_confirmation, checkout_success, time_out and index.  

This issue is further discussed on ZC forum at: https://www.zen-cart.com/showthread.php?220814-Paypal-In-context-popup-does-not-load-content-then-proceeds-to-full-page-payment

mc12345678:
Fixes #1192

### DIFF
--- a/includes/auto_loaders/config.paypalwpp_incontext.php
+++ b/includes/auto_loaders/config.paypalwpp_incontext.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * Autoloader array for PayPal Express In Context functionality. Makes sure that PayPalWPP In Context is instantiated at the
+ * right point of the Zen Cart initsystem.
+ * 
+ * @package     paypal_incontext
+ * @copyright   Copyright 2003-2016 Zen Cart Development Team
+ * @copyright   Portions Copyright 2003 osCommerce
+ * @copyright   Portions Copyright 2012-2016 mc12345678
+ * @license     http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version     $Id: config.paypalwpp_incontext.php xxxx 2016-08-20 11:15:10Z mc12345678 $
+ */
+
+    $autoLoadConfig[0][] = array(
+        'autoType' => 'class',
+        'loadFile' => 'observers/class.paypalwpp_incontext.php'
+    );
+    $autoLoadConfig[199][] = array(
+        'autoType' => 'classInstantiate',
+        'className' => 'paypalwpp_incontext',
+        'objectName' => 'paypalwpp_ic'
+    ); 
--- a/includes/classes/observers/class.paypalwpp_incontext.php
+++ b/includes/classes/observers/class.paypalwpp_incontext.php
@@ -1,0 +1,73 @@
+<?php
+
+/**
+ * @package     paypal_incontext
+ * @copyright   Copyright 2003-2016 Zen Cart Development Team
+ * @copyright   Portions Copyright 2003 osCommerce
+ * @copyright   Portions Copyright 2012-2016 mc12345678
+ * @license     http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: mc12345678 Sat Aug 21 12:16:37 2016 -0500 Added in v1.6.0 $
+ */
+class paypalwpp_incontext extends base {
+
+  
+  /*
+   * This is the observer for the PayPal return URIs when the PayPal payment is being processed at the end of the purchase.
+   */
+  function __construct() {
+    
+    $attachNotifier = array();
+
+    $attachNotifier[] = 'NOTIFY_HEADER_START_CHECKOUT_SHIPPING';
+    $attachNotifier[] = 'NOTIFY_HEADER_START_CHECKOUT_PAYMENT';
+
+    $this->attach($this, $attachNotifier);
+  }
+
+
+  // Clear the paypal_ec_token at the PayPal return URL (RETURNURL) when the
+  //  payment is cancelled at PayPal.  Without other process modifications, 
+  //  this is seen as necessary so that at the revisit to the payment 
+  //  confirmation page that PayPal Express In Context can again operate.
+  //    $attachNotifier[] = 'NOTIFY_HEADER_START_CHECKOUT_SHIPPING';
+  function updateNotifyHeaderStartCheckoutShipping(&$callingClass, $notifier) {
+    if (isset($_GET['ec_cancel']) && $_GET['ec_cancel'] == '1' 
+     && isset($_SESSION['paypal_ec_token']) && !empty($_SESSION['paypal_ec_token'])
+     && defined('MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE') && MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE == 'InContext' 
+     && defined('MODULE_PAYMENT_PAYPALWPP_MERCHANTID')    && MODULE_PAYMENT_PAYPALWPP_MERCHANTID    != '' 
+     && defined('MODULE_PAYMENT_PAYPALWPP_STATUS')        && MODULE_PAYMENT_PAYPALWPP_STATUS        == 'True') {
+      unset($_SESSION['paypal_ec_token']);
+    }
+  }
+
+  // Clear the paypal_ec_token at the PayPal return URL (RETURNURL) when the
+  //  payment is cancelled at PayPal.  Without other process modifications, 
+  //  this is seen as necessary so that at the revisit to the payment 
+  //  confirmation page that PayPal Express In Context can again operate.
+  //    $attachNotifier[] = 'NOTIFY_HEADER_START_CHECKOUT_PAYMENT';
+  function updateNotifyHeaderStartCheckoutPayment(&$callingClass, $notifier) {
+    if (isset($_GET['ec_cancel']) && $_GET['ec_cancel'] == '1' 
+     && isset($_SESSION['paypal_ec_token']) && !empty($_SESSION['paypal_ec_token'])
+     && defined('MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE') && MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE == 'InContext' 
+     && defined('MODULE_PAYMENT_PAYPALWPP_MERCHANTID')    && MODULE_PAYMENT_PAYPALWPP_MERCHANTID    != '' 
+     && defined('MODULE_PAYMENT_PAYPALWPP_STATUS')        && MODULE_PAYMENT_PAYPALWPP_STATUS        == 'True') {
+      unset($_SESSION['paypal_ec_token']);
+    }
+  }
+
+  /*
+   * Generic function that is activated when any notifier identified in the observer is called but is not found in one of the above previous specific update functions is encountered as a notifier.
+   */
+  function update(&$callingClass, $notifier, $paramsArray) {
+
+    if ($notifier == 'NOTIFY_HEADER_START_CHECKOUT_SHIPPING') {
+      $this->updateNotifyHeaderStartCheckoutShipping($callingClass, $notifier);
+    }
+  
+    if ($notifier == 'NOTIFY_HEADER_START_CHECKOUT_PAYMENT') {
+      $this->updateNotifyHeaderStartCheckoutPayment($callingClass, $notifier);
+    }
+    
+  } //end update function - mc12345678
+} //end class - mc12345678
+
--- a/includes/modules/pages/checkout_confirmation/jscript_paypal_incontext.php
+++ b/includes/modules/pages/checkout_confirmation/jscript_paypal_incontext.php
@@ -1,0 +1,17 @@
+<?php
+/**
+ * Javascript enabler for providing and handling the PayPal 
+ *   Express In Context shopping experience.
+ *
+ * @package     paypal_incontext
+ * @copyright   Copyright 2003-2016 Zen Cart Development Team
+ * @copyright   Portions Copyright 2003 osCommerce
+ * @copyright   Portions Copyright 2012-2016 mc12345678
+ * @license     http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: Author: mc12345678  Wed Aug 24 20:20:09 2016 -0500 Modified in v1.6.0 $
+ */
+ 
+    $buttonContainer = '';
+        
+    require(DIR_WS_MODULES . 'payment/paypal/paypalwpp_incontext.php');
+
--- a/includes/modules/pages/login/jscript_paypal_incontext.php
+++ b/includes/modules/pages/login/jscript_paypal_incontext.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * Javascript enabler for providing and handling the PayPal 
+ *   Express In Context shopping experience.
+ *
+ * @package     paypal_incontext
+ * @copyright   Copyright 2003-2016 Zen Cart Development Team
+ * @copyright   Portions Copyright 2003 osCommerce
+ * @copyright   Portions Copyright 2012-2016 mc12345678
+ * @license     http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: Author: mc12345678  Wed Aug 24 20:20:09 2016 -0500 Modified in v1.6.0 $
+ */
+ 
+    $buttonContainer = ($ec_button_enabled) ? 'ecicButton' : '';
+        
+    require(DIR_WS_MODULES . 'payment/paypal/paypalwpp_incontext.php');
--- a/includes/modules/payment/paypal/paypalwpp_incontext.php
+++ b/includes/modules/payment/paypal/paypalwpp_incontext.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * paypal EC In-Context button display template
+ *
+ * @package     paypal_incontext
+ * @copyright   Copyright 2003-2016 Zen Cart Development Team
+ * @copyright   Portions Copyright 2003 osCommerce
+ * @copyright   Portions Copyright 2012-2016 mc12345678
+ * @license     http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: Author: mc12345678  Wed Aug 24 20:20:09 2016 -0500 Modified in v1.6.0 $
+ *
+ *   $buttonContainer is either the 'id' for an HREF <a> tag or <button> tag.
+ *     it can be represented as a string or as a javascript element.
+ */
+
+if (   defined('MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE') && MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE == 'InContext' 
+    && defined('MODULE_PAYMENT_PAYPALWPP_MERCHANTID')    && MODULE_PAYMENT_PAYPALWPP_MERCHANTID    != '' 
+    && defined('MODULE_PAYMENT_PAYPALWPP_STATUS')        && MODULE_PAYMENT_PAYPALWPP_STATUS        == 'True') {
+    
+    if (isset($buttonContainer) && zen_not_null($buttonContainer) && $buttonContainer != '') {
+
+        if (isset($_SESSION['payment']) && !empty($_SESSION['payment'])) {
+            $payMod = $_SESSION['payment'];
+        } else {
+            $payMod = null;
+        }
+
+        require_once(DIR_WS_CLASSES . 'payment.php');
+        $paypalwpp_module = 'paypalwpp';
+        // init the payment object
+        $classPay = new payment($paypalwpp_module);
+        if (zen_not_null($payMod)) {
+            $_SESSION['payment'] = $payMod;
+        } else {
+            unset($_SESSION['payment']);
+        }
+        unset($payMod);
+?><script type="text/javascript">
+window.paypalCheckoutReady = function () {
+  paypal.checkout.setup('<?php echo MODULE_PAYMENT_PAYPALWPP_MERCHANTID; ?>', {
+    locale: '<?php echo ${$paypalwpp_module}->getLanguageCode('incontext'); ?>',
+<?php
+    echo (MODULE_PAYMENT_PAYPALWPP_SERVER == 'live' ? '' : "\n          environment: 'sandbox',"); ?>
+    buttons: [{
+      container: '<?php echo $buttonContainer; ?>',
+      type: 'checkout',
+      color: 'gold', <?php /* possible colors are 'gold', 'blue', or 'silver' echo (defined('MODULE_PAYMENT_PAYPALWPP_INCONTEXTCOLOR') && MODULE_PAYMENT_PAYPALWPP_INCONTEXTCOLOR != '' ? MODULE_PAYMENT_PAYPALWPP_INCONTEXTCOLOR : 'gold'); */ ?>
+      size: '<?php echo (defined('MODULE_PAYMENT_PAYPALWPP_INCONTEXTSIZE') && MODULE_PAYMENT_PAYPALWPP_INCONTEXTSIZE != '' ? MODULE_PAYMENT_PAYPALWPP_INCONTEXTSIZE : 'small'); ?>',
+      shape: '<?php echo (defined('MODULE_PAYMENT_PAYPALWPP_INCONTEXTSHAPE') && MODULE_PAYMENT_PAYPALWPP_INCONTEXTSHAPE != '' ? MODULE_PAYMENT_PAYPALWPP_INCONTEXTSHAPE : 'pill'); ?>'
+    }],
+    click: function() {
+      paypal.checkout.initXO();
+
+      var action = $.post('/set-express-checkout');
+
+      action.done(function (data) {
+        paypal.checkout.startFlow(data.token);
+      });
+
+      action.fail(function () {
+        paypal.checkout.closeFlow();
+      });
+    }
+  });
+};
+</script>
+<?php 
+    }
+?>
+<script async="async" src="//www.paypalobjects.com/api/checkout.js" type="text/javascript"></script>
+<?php
+}
--- a/includes/modules/payment/paypalwpp.php
+++ b/includes/modules/payment/paypalwpp.php
@@ -280,6 +280,16 @@ class paypalwpp extends base {
     $confirmation = array('title' => '', 'fields' => array());
     return $confirmation;
   }
+  
+  function process_form_params() {
+    if (   defined('MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE') && MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE == 'InContext' 
+        && defined('MODULE_PAYMENT_PAYPALWPP_MERCHANTID')    && MODULE_PAYMENT_PAYPALWPP_MERCHANTID    != '' 
+        && defined('MODULE_PAYMENT_PAYPALWPP_STATUS')        && MODULE_PAYMENT_PAYPALWPP_STATUS        == 'True') {
+      return ' data-paypal-button="true" data-paypal-id="' . MODULE_PAYMENT_PAYPALWPP_MERCHANTID . '"';
+    } else {
+      return false;
+    }
+  }
   /**
    * Prepare the hidden fields comprising the parameters for the Submit button on the checkout confirmation page
    */
@@ -296,17 +306,30 @@ class paypalwpp extends base {
     if ($this->use_incontext_checkout == false) return '';
 
     // send the PayPal-provided javascript to trigger the incontext checkout experience
-    return "      <script>
+    return "      <script type=\"text/javascript\">
         window.paypalCheckoutReady = function () {
         paypal.checkout.setup('" . MODULE_PAYMENT_PAYPALWPP_MERCHANTID . "', {
           //locale: '" . $this->getLanguageCode('incontext') . "',"
           . (MODULE_PAYMENT_PAYPALWPP_SERVER == 'live' ? '' : "\n          environment: 'sandbox',") . "
           container: 'checkout_confirmation',
-          button: 'btn_submit'
+          button: 'btn_submit',
+          click: function() {
+            paypal.checkout.initXO();
+
+            var action = $.post('/set-express-checkout');
+
+            action.done(function (data) {
+              paypal.checkout.startFlow(data.token);
+            });
+
+            action.fail(function () {
+              paypal.checkout.closeFlow();
+            });
+          }
         });
       };
       </script>
-      <script src=\"//www.paypalobjects.com/api/checkout.js\" async></script>";
+      <div data-paypal-button=\"yes\" data-paypal-id=\"" . MODULE_PAYMENT_PAYPALWPP_MERCHANTID . "\"></div>";
   }
   /**
    * Prepare and submit the final authorization to PayPal via the appropriate means as configured
@@ -686,6 +709,8 @@ if (false) { // disabled until clarification is received about coupons in PayPal
 
     $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, date_added) values ('PayPal Merchant ID', 'MODULE_PAYMENT_PAYPALWPP_MERCHANTID', '', 'Enter your PayPal Merchant ID here. This is used for the more user-friendly In-Context checkout mode. You can obtain this value by going to your PayPal account, clicking on Profile and navigating to the My Business Info section; You will find your Merchant Account ID on that screen. A typical merchantID looks like FDEFDEFDEFDE11.', '6', '25', now())");
     $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('Use InContext Checkout?', 'MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE', 'InContext', 'PayPal now offers a newer friendlier InContext (in-page) checkout mode (Requires that you enter your MerchantID in the Merchant ID Setting above). Or you can use the older checkout style which offers Pay Without Account by default but with a full-page-redirect.', '6', '25', 'zen_cfg_select_option(array(\'InContext\', \'Old\'), ', now())");
+    $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('InContext Button Size?', 'MODULE_PAYMENT_PAYPALWPP_INCONTEXTSIZE', 'small', 'When PayPal\'s newer friendlier InContext (in-page) checkout mode is enabled PayPal offers various PayPal button sizes. Select the size of the button to use.', '6', '25', 'zen_cfg_select_option(array(\'small\', \'tiny\', \'medium\'), ', now())"); 
+    $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('InContext Button Shape?', 'MODULE_PAYMENT_PAYPALWPP_INCONTEXTSHAPE', 'pill', 'When PayPal\'s newer friendlier InContext (in-page) checkout mode is enabled the PayPal button can take different shapes. Select the button shape to use.', '6', '25', 'zen_cfg_select_option(array(\'pill\', \'rect\'), ', now())"); 
 
     $this->notify('NOTIFY_PAYMENT_PAYPALWPP_INSTALLED');
   }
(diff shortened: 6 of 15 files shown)


drbyte:
👍

drbyte:
I find it curious that PayPal's demos use a different file, `checkout.js` instead of `paypal.js` and none of their docs mention the `paypal.js` file.

mc12345678:
Have found today that api/checkout.js will work as well, but there is still an issue with implementation where in process_button, because when this is called in later processing and the token is set or when cancel is selected (causing the token to be set as well as ec_cancel to be set to 1, that the javascript code provided does not get "redisplayed".

The Web address that I have that documented it's use redirects to the current in-context documentation. Further by following the path of that content, it in part directs to a version of the checkout.js file.

Recommend an alternate location to capture the inclusion of the applicable js code or a "work-around" in process_button or elsewhere so that process_button is "smarter".

mc12345678:
FYI, I have a revised version that is more in line with currently published information. Currently not in a position to force replacement of the existing pull request. Can push as is (with history) or at a later time force push a "one time" solution.) Please let me know which is preferred.

mc12345678:
FYI, I have a revised version that is more in line with currently published information. Currently not in a position to force replacement of the existing pull request. Can push as is (with history) or at a later time force push a "one time" solution.) Please let me know which is preferred.

mc12345678:
FYI, I have a revised version that is more in line with currently published information. Currently not in a position to force replacement of the existing pull request. Can push as is (with history) or at a later time force push a "one time" solution.) Please let me know which is preferred.

drbyte:
I'd recommend:

a) make a new commit which updates this PR by removing unnecessary changes and adding additional changes. 

b) then your proposed changes can be reviewed here

c) then later once everything's determined to be the correct complete solution, we can consider having you rebase the entire PR into one nice clean commit which you force-push to update this PR (which is the preferable approach, since all the separate earlier commits you made to this PR aren't very atomic; they're too granular) ;
or we could consider squashing them on our end before merging (this is less clean but is better than doing a direct merge)

But, since you said you're not able to do a "force push" "currently", for whatever reason, at least start with adding yet another commit to update this PR with your latest proposed strategy.

drbyte:
It's throwing a JS error: `Undefined variable: paypal` when trying to call `paypal.checkout.setup(the merchant id)`

mc12345678:
Looks like was also missing the modification that I expected to exist related to the button information... Adding to and forcing to the commit momentarily. (Able to work locally with a git rather than through the github website only.)

drbyte:
While you're working with a local git client, do you mind rebasing all the commits into one, and then doing a force push?

mc12345678:
Force push finally done... Added installation keys to offer a little bit of modification capability, Addressed PayPal's implementation which adds a button regardless of an existing image within a link such that if/when possible to display the in context window that if the visitor doesn't have javascript enabled that a button is still displayed for checkout/payment. 

Instead of duplicating the button code for every file implementation (which would also easily allow each presentation of the button to be unique), a common file was generated that implements the PayPal default settings if not fully/properly installed.

includes/modules/pages/checkout_confirmation/jscript_paypal_incontext.php
```diff
@@ -0,0 +1,21 @@
...
+ * @version $Id: Author: mc12345678  Sat Aug 13 11:35:35 2016 -0400 Modified in v1.6.0 $
+ */
+ 
+  if (   defined('MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE') && MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE == 'InContext' 
+      && defined('MODULE_PAYMENT_PAYPALWPP_MERCHANTID')    && MODULE_PAYMENT_PAYPALWPP_MERCHANTID    != '' 
+      && defined('MODULE_PAYMENT_PAYPALWPP_STATUS')        && MODULE_PAYMENT_PAYPALWPP_STATUS        == 'True') {
+        
+        require(DIR_WS_MODULES . 'payment/paypal/paypalwpp_incontext.php');
+?>
+<!-- merchant site HTML page --><script async="async" src="//www.paypalobjects.com/api/checkout.js"></script>
```
drbyte: What's the purpose of the `<!-- merchant site HTML page -->` comment? Or did that just come from some sample code?

includes/modules/payment/paypal/paypalwpp_incontext.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+    require_once(DIR_WS_CLASSES . 'payment.php');
+    $paypalwpp_module = 'paypalwpp';
+    // init the payment object
+    $classPay = new payment($paypalwpp_module);
+?><script>
+window.paypalCheckoutReady = function () {
+  paypal.checkout.setup('<?php echo MODULE_PAYMENT_PAYPALWPP_MERCHANTID; ?>', {
+    //locale: '<?php echo ${$paypalwpp_module}->getLanguageCode('incontext'); ?>',
```
drbyte: I'm not sure the `$paypalwpp_modue` variable is needed. Couldn't it just be a string directly in the `new payment()` call?

And, syntactically, shouldn't `${$paypalwpp_module}->getLanguageCode(` actually be `${$classPay}->getLanguageCode(`?

includes/modules/payment/paypalwpp.php
```diff
@@ -686,6 +686,9 @@ function install() {
...
+    $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('InContext Button Size?', 'MODULE_PAYMENT_PAYPALWPP_INCONTEXTSIZE', 'small', 'When PayPal\'s newer friendlier InContext (in-page) checkout mode is enabled PayPal offers various PayPal button sizes. Select the size of the button to use.', '6', '25', 'zen_cfg_select_option(array(\'small\', \'tiny\', \'medium\'), ', now())"); 
+    $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('InContext Button Shape?', 'MODULE_PAYMENT_PAYPALWPP_INCONTEXTSHAPE', 'small', 'When PayPal\'s newer friendlier InContext (in-page) checkout mode is enabled PayPal the PayPal button can take different shapes. Select the button shape to use.', '6', '25', 'zen_cfg_select_option(array(\'pill\', \'rect\'), ', now())");
```
drbyte: I "get" why you've done this. 
I'm curious: what's the likelihood of people needing to customize this? What's your gut say? 1 in 5?, 1 in 25? or 1 in 500?

(I'm mindful that we used to offer color customization options this way, but later moved them to the language file because nobody used them, and then later even PayPal deprecated and abandoned them completely.)

drbyte:
Don't get me wrong: my comments are not criticisms. Just standard code-review questions ;)

includes/modules/pages/checkout_confirmation/jscript_paypal_incontext.php
mc12345678: Originally was from sample code, then thought to leave to make discovery within code "easier", then thought for some unknown reason that would find itself near the actual new button that is created, but in retrospect serves no real purpose.

That said, even though I originally had the required file directly in the one jscript_ file, seems like one of the "methods" to find ZC content is to reference some type of "comment" in each "section". Kind of wondering if shouldn't add a comment at the beginning and end of the file outside the if statement to support "diagnosis" of why the page(s) don't respond as expected.

Thoughts? (ie. Remove the above comment and add two minor ones, one at file opening before <?php and one after close of if referencing in-context.)

includes/modules/payment/paypal/paypalwpp_incontext.php
mc12345678: So, here was trying to align various code to existing usage. I need to relocate from where this was taken and also verify that on first usage (shopping_cart) that the language identifier is returned. Reason for this verification is that when first implemented, received an odd error with a section of notifier code where the issues appeared unrelated. But once the require of the payment class was made require_once, the issue went away. It was then seen as odd that ${$paypalwpp_module} actually worked. So in a way this syntactical oddity was left to in a way make a point and possibly address a separate issue. If $classPay really represents the resulting class assignment, then it seems that the corrected code should be $classPay->.

To further "explain", I had found something that worked in the heat of solving the problem and hadn't gone back to perform further potential cleanup such as this. It just doesn't seem right that in calling to declare a class that suddenly the global space now includes the variable by that modules class. Not saying it's a problem, just seemed unusual and another reason why the code wasn't presented in a more "syntactical" fashion. :)

includes/modules/payment/paypalwpp.php
mc12345678: So, this was a last minute feature add (bell or whistle in my opinion) with the thought of a few things. First could completely omit any such designations in the original file as the values provided are the PayPal defaults (at this time), but also trying to think about all of the discussions in the forum about presentation and ability to modify aspects to suit. So I went all out with configurable menu options which brought to new ZC users the option to easily locate and then modify the setting(s) for the selection(s). I feel that at the moment (though pure speculation) that talking about the 1 in 500 end users, but... also not totally thinking that the end user is the one to likely adjust/consider the option as much as the one that is installing/upgrading the site. I do know that a test site on which this was applied that the default PP gold button looks terrible, the blue works but the silver makes it almost disappear but really matches the template (but it doesn't pop).

By moving all calls to this formatting into a shared file, I was trying to eliminate any such use, but then got to thinking about the DIYers and was trying to settle on a sort of happy medium considering PP seemed to think this "feature" of at least all three options to be necessary. 

Now that said, if you're thinking to just move color to a definition or similar and keep the other two as options, I could go with that and leave a note about the current potential other "styles". It all becomes mute if the shared file is brought into the jscript_ in order to support various representations of the PayPal button.

Essentially, while I'm satisfied that there is working code and believe that ZC needs anything it can get to regain the traction lost in the last 2 years or so. Don't know if it's a culture shift (generationally related) or if there is/was some ruffled feathers, but ZC in recent e-commerce reviews doesn't seem to even show. When commenters bring it up, the response is often not positive. :(

mc12345678:
> Don't get me wrong: my comments are not criticisms. Just standard code-review questions ;)

No worries on my part, they're valid questions in an attempt to minimize changes and in the case of jscript the amount of unnecessary information passed to the browser, past experience with related code, and general kitten misfortune curiosity.  

If nothing else, almost everything asked about was on my irk (with no visible/noticeable impact on operation) list, though expected more also about the separation of the button action code from the checkout.js line.  Also somewhat expected something about that the PP javascript checkout.js code is called rather than "included".  Though by review of the code it seems like it is still being periodically updated.  From an end user perspective, seems best to keep it referenced instead of included... But, one might then also say the same about other software such as jQuery. :)

includes/modules/pages/checkout_confirmation/jscript_paypal_incontext.php
drbyte: Skip the comments IMO. 
If there's an issue loading the js files then it won't be just with this one, so sending extra text to the browser 100% of the time is unnecessary clutter.

mc12345678: I was thinking more from the standpoint that in source code review, one sees that there is this in context option, but no active code, searches and finds that their issue might be the missing the merchant ID. Once entered, then assuming JS related issues that it now "works". Somewhat trying to address the potential/likely questions as best as possible.

includes/modules/payment/paypal/paypalwpp_incontext.php
mc12345678: Found the code snippet from which this originated...

ipn_main_hander.php:

```php
  // See if the paypalwpp module is enabled.
  if (defined('MODULE_PAYMENT_PAYPALWPP_STATUS') && MODULE_PAYMENT_PAYPALWPP_STATUS == 'True') {
    $paypalwpp_module = 'paypalwpp';
    // init the payment object
    $payment_modules = new payment($paypalwpp_module);
    // set the payment, if they're hitting us here then we know
    // the payment method selected right now.
    $_SESSION['payment'] = $paypalwpp_module;
    // check to see if we have a token sent back from PayPal.
    if (!isset($_SESSION['paypal_ec_token']) || empty($_SESSION['paypal_ec_token'])) {
      // We have not gone to PayPal's website yet in order to grab
      // a token at this time.  This will send the customer over to PayPal's
      // website to login and return a token
      $$paypalwpp_module->ec_step1();
    } else {
      // This will push on the second step of the paypal ec payment
      // module, as we already have a PayPal express checkout token
      // at this point.
      $$paypalwpp_module->ec_step2();
    }
  }
```

While the above code is a "snippet" from the entire file, $payment_modules is not used at all in the remainder of that code other than to be assigned or potentially assigned again elsewhere later in the same file.  Further, the use of $$paypalwpp_module instead of ${$paypalwpp_module} while apparently clear enough for PHP 7.0 and doesn't currently cause an issue, seems like perhaps could lead to a problem in the future if not considered, but not trying to make changes just to do so. (Like the missing semi-colon in the tpl_ec_button file in the non-replaced link line.)

includes/modules/pages/checkout_confirmation/jscript_paypal_incontext.php
drbyte: I completely understand your logic.

But not everyone uses PayPal, and there's no need to clutter every site with unnecessary comments. 
Someone who would trace back thru source as you described is equally likely to check the admin config settings for the module and discover the same requirement.

Additionally, while I grok that lots of people have resorted to injecting a plethora of comments into every template segment, I believe it'd be better to be able to also turn off all those comments, displaying them only optionally if a master config switch were enabled for troubleshooting purposes. But that's a change that requires editing a massive number of individual files.
And, now this discussion digresses far from the topic at hand.

mc12345678: Comment to be removed and incorporated by forced push again, unless desired to "build" on this "one" at a time? :/ (Or at least to make the same change to all of the files at once instead of something like 6 times over.) Either way will be something I will get to in the next day or two based on accessibility to the desired software/hardware combination. Please continue to provide comments, insightful.

drbyte: rebased force push is ideal.

includes/modules/payment/paypal/paypalwpp_incontext.php
drbyte: My bad. I had the global mixed around in my head. (Not happy with all the global vars we use in ZC, but we're working on changing that. Separate topic.)
You're right. No need to switch to ${$classPay}.

If we decide later to clean up the string to avoid assigning another unused var, we can do that in the snippet you quoted as well, at that time.

drbyte:
> expected something about that the PP javascript checkout.js code is called rather than "included". 

No. The `checkout.js` file should definitely be called from paypal directly. It's bad practice to self-host any PayPal assets. They should always be loaded from PayPal's servers.

includes/modules/payment/paypal/paypalwpp_incontext.php
mc12345678: So should I change the use of variables in this section? Ie instead of ${$paypalwpp_module] just use $paypalwpp? I hadn't looked into the code significantly deeper to see what all happens in the function regarding the passed variable and had just considered the code reuse as "easy" to address when/if some sort of global change was to occur.

drbyte: No. I'm assuming that even though you've commented-out the `locale` part you intend to keep it. In that case leave this section as-is.

drbyte:
I do wonder though: there's a lot of copy-paste duplication in all those jscript files. I can't think of any reason for them all to be duplicates of each other. Lots of room for error there.

Why not just make them all do:

```php
<?php
    require(DIR_WS_MODULES . 'payment/paypal/paypalwpp_incontext.php');
```
... and move all the `if` conditional logic into the `require`d file?

includes/modules/payment/paypal/paypalwpp_incontext.php
mc12345678: Chicken/egg. It was commented out in the code of paypalwpp. I haven't had a good environment in which to test it, so left it for those wishing to get more from it. Similar thing of also trying to minimize future documentation. :)

mc12345678:
Had contemplated that, but was also thinking that perhaps code reuse for other purposes might exist. The central settings in the required file with the reason to use them outside of it... 

That said and several hours on which to ponder. Putting all in one file, if someone wished to specialize one or another versions of the file, they could copy the contents of the shared file into the individual file and modify to suit.

Arm sufficiently twisted. :)

includes/modules/pages/checkout_confirmation/jscript_paypal_incontext.php
```diff
@@ -0,0 +1,27 @@
...
+ 
+  if (   defined('MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE') && MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE == 'InContext' 
+      && defined('MODULE_PAYMENT_PAYPALWPP_MERCHANTID')    && MODULE_PAYMENT_PAYPALWPP_MERCHANTID    != '' 
+      && defined('MODULE_PAYMENT_PAYPALWPP_STATUS')        && MODULE_PAYMENT_PAYPALWPP_STATUS        == 'True') {
+        
+//        require(DIR_WS_MODULES . 'payment/paypal/paypalwpp_incontext.php');
+if (isset($_SESSION['paypal_ec_token']) && !empty($_SESSION['paypal_ec_token'])
+    && isset($_SESSION['paypal_ec_markflow']) && $_SESSION['paypal_ec_markflow'] = 1) {
+
+}?>
```
drbyte: Not understanding why all this is here ...

mc12345678: Non-essential if the overall concept/operation is accepted, something of an alternate path that would need to be used/taken advantage of if the provided method of "clearing" the paypal_ec_token (left the paypal_ec_markflow unchanged) is considered undesirable/unacceptable when ec_cancel=1 is a parameter of the url. 

Is primarily a remnant that can be removed for operational sake, but the interaction of all these parts is pretty extensive.

The vital part if all else is accepted is the $buttonContainer and require of the file. Otherwise thus far haven't found a way (outside of a potential init_ file or other file loaded on page load/redirect) that would allow accomplishing the same functionality already offered by ZC when pressing the confirm order button while a token is assigned to an "active" connection.

mc12345678: I meant to write something earlier about the change(s) provided, might help with some of the understanding.

Currently, process_button in paypalwpp.php only provides information related to in context process if there is no previous paypal_ec_token session data and the appropriate action/settings have been applied to allow PayPal in context to be used. In straight forward processing this is fine because either the customer visited PayPal earlier and established all of the necessary information to permit shipping verification and payment collection or they haven't yet and therefore at the checkout_confirmation page are ready to goto PayPal (in this discussion In Context) where they apply the necessary credentials to complete payment. The thing is, if at any point during this process they select the cancel option in the PayPal window, then they are redirected to the cancelurl which was set depending on the logged in status of the user before attempting purchase. Well, at the point of checkout_confirmation, if I understand correctly the user either as themselves or a temporary user is logged in and therefore the cancel url is back to the checkout_shipping page. But, there is nothing at the checkout_shipping page nor forward towards checkout_confirmation that modifies the paypal_ec_token other than potentially recreating one when the payment is again posted at the checkout_confirmation page and found to be incorrect for the current purchase. 

So, on second and subsequent trips to payment_confirmation, process_button will push nothing related to the use of in context, including if the user were to go back as far as the shopping cart and then use in context payment if then immediately delivered to the checkout_confirmation page.

So two options at that point seemed possible. Make it so that some or all of the session data was removed before getting to that point, or remove the check of the existing session data and lose the operational benefit of having/maintaining that data. From the notes in paypalwpp.php it appears that the goal is to maintain the session data, so the only thing that seemed viable was to remove the session data when it would appear as most likely to need to be rebuilt anyways (cancel). So, then the issue of the checkout.js file's involvement, well, if the session data is to prevent use of in context on the checkout_confirmation page, then it is possible that it was used in advance of that page but that page is/was to be shown and the lightbox removed. To accomplish the later part requires the checkout.js file somewhere on page, but with it inside of the form code, it never seemed to work. Once pulled out, it seemed to work though came across some issues with firefox and chrome. Chrome outright didn't like the iframe if it was hosted through https and the source site was not. Firefox was not as direct about the issue.

So, perhaps that answers the other side of the potential confusion. (not sure to which aspect the question was expected to cover)

includes/classes/observers/class.paypalwpp_incontext.php
```diff
@@ -0,0 +1,76 @@
...
+  //  this is seen as necessary so that at the revisit to the payment 
+  //  confirmation page that PayPal Express In Context can again operate.
+  //    $attachNotifier[] = 'NOTIFY_HEADER_START_CHECKOUT_SHIPPING';
+  function updateNotifyHeaderStartCheckoutShipping(&$callingClass, $notifier) {
+    if (isset($_GET['ec_cancel']) && $_GET['ec_cancel'] == '1' 
+     && isset($_SESSION['paypal_ec_token']) && !empty($_SESSION['paypal_ec_token'])
+     && defined('MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE') && MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE == 'InContext' 
+     && defined('MODULE_PAYMENT_PAYPALWPP_MERCHANTID')    && MODULE_PAYMENT_PAYPALWPP_MERCHANTID    != '' 
+     && defined('MODULE_PAYMENT_PAYPALWPP_STATUS')        && MODULE_PAYMENT_PAYPALWPP_STATUS        == 'True') {
+      unset($GLOBALS[_SESSION]['paypal_ec_token']);
```
drbyte: I don't understand why you and others lately are referencing `$GLOBALS[_SESSION]` when $_SESSION is a superglobal and therefore `$_SESSION` is all that's needed.

includes/auto_loaders/config.paypalwpp_incontext.php
```diff
@@ -0,0 +1,27 @@
...
+ * @package     paypalwpp_incontext
+ * @author      mc12345678 
+ * @copyright   Copyright 2008-2016 mc12345678
+ * @copyright   Copyright 2003-2007 Zen Cart Development Team
+ * @copyright   Portions Copyright 2003 osCommerce
+ * @link        http://www.zen-cart.com/
+ * @license     http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version     $Id: config.paypalwpp_incontext.php xxxx 2016-08-20 11:15:10Z mc12345678 $
+ *
+ * Stock by Attributes 1.5.4
```
drbyte: A lot of unnecessary docblock components here. 

And "Stock by Attributes" can probably be removed ;)

includes/modules/pages/checkout_confirmation/jscript_paypal_incontext.php
drbyte: The conditional code in this file seems unnecessary here. Particularly when there's nothing inside this nested `if`.

... and (unless I'm missing something here despite your explanation) then the outer `if` is repeated in the `require`d file anyway, so feels like it could be excluded and just left to the `require`d file instead.

includes/auto_loaders/config.paypalwpp_incontext.php
mc12345678: Welcome suggestion of making it shorter, what minimal information is suggested if any?

Thought I got rid of that even though I was using it as a template. :) oops.

includes/classes/observers/class.paypalwpp_incontext.php
mc12345678: Comes from a comment in the php manual related to unsetting the $_SESSION variable. Of course note was made 11 years ago. Based on the ZC posted related comment, and the consideration that this file could be called within a function (even though it is normally loaded in the global space) was taking that code "requirement" into consideration. Thus far in review of other areas of the php manual it discusses the "availability" of the superglobal variables but not necessarily the impact of modifying the associated data such as unsetting it.

Basically comes down to following another's lead in the face of the statement there and here about it being a superglobal and unnecessary to refer back to the $GLOBALS variable in such a case.

includes/classes/observers/class.paypalwpp_incontext.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  /*
+   * Generic function that is activated when any notifier identified in the observer is called but is not found in one of the above previous specific update functions is encountered as a notifier.
+   */
+  function update(&$callingClass, $notifier, $paramsArray) {
+  global $db;
```
mc12345678: Agree. Will strike. Good catch and thank you.

includes/modules/payment/paypal/paypalwpp_incontext.php
```diff
@@ -0,0 +1,70 @@
...
+      }
+
+      require_once(DIR_WS_CLASSES . 'payment.php');
+      $paypalwpp_module = 'paypalwpp';
+      // init the payment object
+      $classPay = new payment($paypalwpp_module);
+      if (zen_not_null($payMod)) {
+        $_SESSION['payment'] = $payMod;
+      } else {
+        unset($GLOBALS[_SESSION]['payment']);
```
drbyte: $_SESSION

includes/auto_loaders/config.paypalwpp_incontext.php
drbyte: I guess it stood out because the docblock segment is longer than the lines of code.

I'd remove `@link`. 

`@package` probably should probably be more consistent across all the files you're adding. Right now you've got `paypalwpp_incontext`, `paypalwpp`, `paymentMethod`, `page`.   

I'm inclined to suggest `paymentMethod`, or if you feel the distinction is beneficial then maybe `paypal_incontext` ... for all the files you're adding via this PR.

includes/modules/payment/paypal/paypalwpp_incontext.php
mc12345678: DOH!

includes/auto_loaders/config.paypalwpp_incontext.php
mc12345678: Considering that some of the files could be removed (are needed only to support paypal_incontext)  and the overall functionality would remain (payment possible through PayPal), I plan on modifying to paypal_incontext for newly added files.

Will address other inconsistencies with the docblocks as well.  Expect to be able to provide updated (and corrected) within the next 14 hours. I encourage not giving up on the review and continuing to identify other issues no matter how "small".

drbyte:
> @mc12345678 " I encourage not giving up on the review and continuing to identify other issues no matter how "small". There are plenty of details to consider, don't want things to die out in the end. :) "

Agreed.

mc12345678:
FYI, have already considered use of four spaces for new code. Other formatting criteria for new code that has been missed?

includes/modules/payment/paypalwpp.php
```diff
@@ -306,7 +306,7 @@ function process_button() {
         });
       };
       </script>
-      <script src=\"//www.paypalobjects.com/api/checkout.js\" async></script>";
+      <div data-paypal-button=\"yes\" data-paypal-id=\"" . MODULE_PAYMENT_PAYPALWPP_MERCHANTID . "\"></div>"
```
drbyte: Is there a missing `;` here?

includes/classes/observers/class.paypalwpp_incontext.php
```diff
@@ -0,0 +1,75 @@
...
+   */
+  function update(&$callingClass, $notifier, $paramsArray) {
+
+    if ($notifier == 'NOTIFY_HEADER_START_CHECKOUT_SHIPPING') {
+      $this->updateNotifyHeaderStartCheckoutShipping($callingClass, $notifier);
+    }
+  
+    if ($notifier == 'NOTIFY_HEADER_START_CHECKOUT_PAYMENT') {
+      $this->updateNotifyHeaderStartCheckoutPayment($callingClass, $notifier);
+    }
```
drbyte: Upon further review it strikes me that this entire `function update()` section is redundant, since the observer system automatically fires the corresponding `updateFooBarBiz()` method whenever the `FOO_BAR_BIZ` notifier point is triggered.

includes/classes/observers/class.paypalwpp_incontext.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+/**
+ * Description of class.paypalwpp_incontext: This class is used to support order information related to PayPal Express In Context.  This way reduces the modifications of the various header files to nearly nothing in support of troubleshooting.
```
drbyte: This description feels redundant.

(You did say "no matter how trivial" ;) )

includes/modules/payment/paypalwpp.php
```diff
@@ -686,6 +686,8 @@ function install() {
...
+    $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('InContext Button Size?', 'MODULE_PAYMENT_PAYPALWPP_INCONTEXTSIZE', 'small', 'When PayPal\'s newer friendlier InContext (in-page) checkout mode is enabled PayPal offers various PayPal button sizes. Select the size of the button to use.', '6', '25', 'zen_cfg_select_option(array(\'small\', \'tiny\', \'medium\'), ', now())"); 
+    $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('InContext Button Shape?', 'MODULE_PAYMENT_PAYPALWPP_INCONTEXTSHAPE', 'small', 'When PayPal\'s newer friendlier InContext (in-page) checkout mode is enabled PayPal the PayPal button can take different shapes. Select the button shape to use.', '6', '25', 'zen_cfg_select_option(array(\'pill\', \'rect\'), ', now())");
```
drbyte: Seems to be an extra mention of the word `PayPal` in the 2 places where you insert `MODULE_PAYMENT_PAYPALWPP_INCONTEXTSHAPE`:

> 'When PayPal\'s newer friendlier InContext (in-page) checkout mode is enabled **PayPal** the PayPal button can take different shapes.

includes/templates/template_default/templates/tpl_checkout_confirmation_default.php
```diff
@@ -175,7 +175,7 @@
         ?>
 
         <?php
-        echo zen_draw_form('checkout_confirmation', $formActionLink, 'post', 'id="checkout_confirmation" onsubmit="submitonce();"');
+        echo zen_draw_form('checkout_confirmation', $formActionLink, 'post', 'id="checkout_confirmation" onsubmit="submitonce();"' . (defined('MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE') && MODULE_PAYMENT_PAYPALWPP_CHECKOUTSTYLE == 'InContext' && defined('MODULE_PAYMENT_PAYPALWPP_MERCHANTID')    && MODULE_PAYMENT_PAYPALWPP_MERCHANTID    != '' && defined('MODULE_PAYMENT_PAYPALWPP_STATUS')        && MODULE_PAYMENT_PAYPALWPP_STATUS        == 'True' ? ' data-paypal-button="true" data-paypal-id="' . MODULE_PAYMENT_PAYPALWPP_MERCHANTID . '"':''));
```
drbyte: I'm still struggling over the 2 changes in this file. They're completely PayPal-specific, and it would be preferable to leave payment-module-specific stuff out of the templates.

Thoughts?

mc12345678: A separate function would/could provide the desired result, but seems would have to test for the presence of that function unless going to add it to all of the existing payment modules... Is a little "selfish" for it to just be for PayPal to be able to add to the innerhtml of the tags.

includes/modules/payment/paypalwpp.php
mc12345678: Was trying to make sure that the text did not look the same for each entry and mucked up a copy/paste. First instance of Paypal removed.

includes/classes/observers/class.paypalwpp_incontext.php
mc12345678: True for ZC 1.5.3 and above, but not if applied to older ZC versions.  Written with the concept that with the thankful addition of the individual function that there is the possibility of removal of the basic update function and the code would continue to support that removal. If the goal is/was to make it only fully applicable to ZC 1.5.5 then need to change it to an auto.loaded_class.php instead of the auto_loader/observer combination.

includes/modules/payment/paypalwpp.php
mc12345678: Yes, in my original code it was on the next line with a lot of commented out code between the end of that line and the end of the one below.

Added back to the end of that line.

includes/templates/template_default/templates/tpl_checkout_confirmation_default.php
mc12345678: Working on adding process_form_params as a payment class function. Adding to paypalwpp.php payment module. Should this commit include addition of it to all payment modules or leave that as separate item/issue to address?

includes/modules/pages/login/jscript_paypal_incontext.php
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @package     paypal_incontext
+ * @copyright   Copyright 2003-2016 Zen Cart Development Team
+ * @copyright   Portions Copyright 2003 osCommerce
+ * @copyright   Portions Copyright 2012-2016 mc12345678
+ * @license     http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: Author: mc12345678  Wed Aug 24 20:20:09 2016 -0500 Modified in v1.6.0 $
+ */
+ 
+    $buttonContainer = 'ecicButton';
```
mc12345678: Causes a javascript error when page is accessed prior to the cart having contents. ($ec_enabled variable or similarly named not providing a false response)

Updating to identify the button the same way the template code is written or '' if not true. All within a line instead of a formal if/else construction.

includes/templates/template_default/templates/tpl_checkout_confirmation_default.php
drbyte: Given that the `payment` class methods all (if the pattern is followed) fire the method in the respective module only if it's defined, there should be no need to touch any of the other modules.

mc12345678: That was the consideration being applied. Saw that at some point it seems (from visual inspection) that the other modules had the "other" functions included even if returning false. Wanted to be sure to address that open "issue" and related consistency.

mc12345678:
One thing I was thinking about. In the middle of the shipping module paypalwpp.php there is a statement that as of Dec 2015 there is a need to use Mark instead of Sole for using PayPal In Context. Was trying to figure out the conditions in which that was fully necessary and if it shouldn't have additional controls to allow Sole "on occasion".

drbyte:
> there is a statement that as of Dec 2015 there is a need to use Mark instead of Sole for using PayPal In Context.

Background:

`SOLUTIONTYPE` is used to control the default display style of the  PayPal landing/login page:
- `Mark` = "PayPal Login" - screen opens with login prompt
- `Sole` = "Credit Cards" - screen opens with Pay By Credit Card prompt

PayPal makes more money when people use funds in their PayPal account, instead of purchasing with credit cards. So, their primary goal is to get people to create accounts, and their secondary goal is to get them to transfer PayPal funds between PayPal accounts (where they get their 3% cut just for brokering it).

Thus, they default to "Mark" mode, and in my experience with various testing scenarios they don't fully support "Sole" in all of their application "flows"/"routes". Sometimes "Sole" just breaks w/o explanation. And their tech staff never have an answer why.  I've determined that the answer is: They assume and prefer "Mark" at all times, except for few places where "Sole" has been accommodated.

At the time I wrote that comment my testing was revealing that they weren't supporting Sole (the API calls were being rejected with errors, unless changed to Mark).

drbyte:
> Was trying to figure out the conditions in which that was fully necessary and if it shouldn't have additional controls to allow Sole "on occasion".

There are already way too many settings to configure. I'm very reluctant to add "more controls" without a solid business case.  What "additional controls" are you thinking of?, and "on what occasions"?

mc12345678:
> What "additional controls" are you thinking of?, and "on what occasions"?

I was thinking more internally than user/operator. My interpretation of the need to use mark as a `SOLUTIONTYPE` was that it was needed for allowing the purchase to go straight through or some unique condition that just wasn't followed up with in more detail; however:
> Background:

> SOLUTIONTYPE is used to control the default display style of the PayPal landing/login page:

> - Mark = "PayPal Login" - screen opens with login prompt
- Sole = "Credit Cards" - screen opens with Pay By Credit Card prompt
PayPal makes more money when people use funds in their PayPal account, instead of purchasing with credit cards. So, their primary goal is to get people to create accounts, and their secondary goal is to get them to transfer PayPal funds between PayPal accounts (where they get their 3% cut just for brokering it).

> Thus, they default to "Mark" mode, and in my experience with various testing scenarios they don't fully support "Sole" in all of their application "flows"/"routes". Sometimes "Sole" just breaks w/o explanation. And their tech staff never have an answer why. I've determined that the answer is: They assume and prefer "Mark" at all times, except for few places where "Sole" has been accommodated.

> At the time I wrote that comment my testing was revealing that they weren't supporting Sole (the API calls were being rejected with errors, unless changed to Mark).

The above clearly indicates that a transaction expected to meet a certain set of criteria could not be determined (the criteria that is) and therefore and in light of some independently observed inconsistencies so for the time being there may not be a use "sole" for this set of cart conditions and use "mark" for the rest with which one can code. 

I have no specific desired need one way or another, was trying to bring back in line with existing PP EC settings where/if able to eliminate one potential variable in why something may end up not working..

drbyte:
> where/if able to eliminate one potential variable in why something may end up not working

Understood.

mc12345678:
Discovered on a fresh install that 1) an error detected at process_confirmation if performing checkout using paypal incontext was not handled properly (return uri was undefined) and 2) that the new button shape configuration setting did not have an accurate default value so would be "blank" during setup until chosen and then saved.

drbyte:
@mc12345678 it's been awhile since we had dialog here on this. 
Have you been using these changes in production anywhere?

drbyte:
@mc12345678  I guess I'm mainly looking for confirmation that you're satisfied that these changes are complete.

mc12345678:
@drbyte, yes it has been a while.

I have been using it on my hosted site as well I have recently been contacted by two independent individuals about applying it to other versions of ZC.  Most recently with a little massaging it was applied to a ZC 1.3.9 site.  The issue encountered or needed to be overcome was clearing of the token when PayPal Express was selected by an individual not yet logged in and the transaction cancelled.  Attempting to then checkout at that point using PayPal Express caused the site to redirect to the standard checkout instead of the InContext window. It appears that something in the code path between 1.3.9 and ZC 1.5x changed that required additional action to clear the token (I assume it was the ipn_main_handler code or the applicable paypalwpp.php Step1/Step2 code that changed)

I recently fully implemented the provided code (in place of what I had previously in place) on my site and just had a purchase through PayPal. Perhaps I have a product misconfiguration, but they chose PayPal Express from the shopping_cart, completed payment through PayPal, but then had to come back/continue through the checkout_payment and checkout_confirmation pages (checkout_shipping was expected to be skipped because the content is digital (though with attributes) and therefore no shipping address to be expected).  They did not cancel the paypal incontext window (which they were presented from shopping_cart), and were not asked for paypal credentials a second time once they arrived at checkout_confirmation.

Another inquirer was following some of the posts in the ZC forum and came across this code, asked a question about it, but has not responded back on success/failure of use/implementation.

drbyte:
Yes, the way the redirects were done (for step1/step2) in v139 have been changed since v150.

mc12345678:
> @mc12345678 I guess I'm mainly looking for confirmation that you're satisfied that these changes are complete.

Ahh, so, there is a little bit of the "rub"... There is function and there is implementation.

One suggestion made was to simply apply the javascript in a way/manner that is throughout instead of on particular pages.  Ie. method of implementation.  This in part stemmed from the application of the paypal "button" at the site to have it available on each and every page instead of on/in the checkout structure only.  The code is maleable enough to handle that, but it requires "user" understanding in order to apply (or someone to apply it for them).  The same is true if the template names a button on a given page with a different designation than provided in the default template(s). 

So... I don't have significant input on applying the code in order to consider application to non-default ZC stores as "complete".

drbyte:
> So... I don't have significant input on applying the code in order to consider application to non-default ZC stores as "complete".

No worries.  Those sorts of adaptations evolve over time.

drbyte:
In reviewing the code for this PR today I'm satisfied to merge it, unless you have any objections.

👍

mc12345678:
From an "inclusion" perspective, I thought that I had (somewhere?) identified the criteria needed to implement on a site that is/was perhaps modified away from a default version, ie. in effect to know the return points, origination points, and what "data" is needed to support the overall operation.  

In a way I used two other "popular" ZC plugins as a frame of reference for implementation such that the visitor is provided the associated code/data only when it has been enabled in the admin with correct data and the specific page(s) have been identified as pages applicable to operate/process the InContext code. Those plugins have javascript files installed on the applicable page, instead of being present in a central location with a "switch" to identify whether to enable/disable or what button is in question/should be sought. Though neither of those two plugins have quite the same "need" of the code... Ie. in principle if a cancel destination doesn't have the js code, the customer could possibly end up with a "locked" page, whereas the others, if they don't have the code then an added feature isn't made available. Little different. :)

mc12345678:
> In reviewing the code for this PR today I'm satisfied to merge it, unless you have any objections.

:+1:
I do not have any objections, I keep wondering what/if there is some unforeseen PayPal operation that isn't done the way that ZC has previously done it and to that I have to count on those potentially more familiar with the way of PayPal.  This in my mind provides an alternate method by which to communicate with PayPal and an improved user experience/perception.

zcwilt:
:+1:

PR #1195, opened 2016-08-13
FIXES #276 Admin tax rounding not the same as catalog

### LINKED ISSUE #276: Admin Tax Rounding not the same as Catalog
Scenario
Tax is set at 20%. 
Product price including tax is 6.99 which when entered into Products Price (Gross) on the product entry page is calculated as 5.825 in the Products Price (Net) entry field. 
Preview page changes this to 2 decimal places making it 5.83. 
Tax added to 5.83 is 7.00.
Tax added to 5.825 is 6.99

The product list in admin shows the product as 7.00 as do any order pages and invoices. 
Store side shows the correct value of 6.99

scottcwilson:
Using the example from the issue, bot the catalog and admin
show the price at $6.99.

### DIFF
--- a/includes/classes/currencies.php
+++ b/includes/classes/currencies.php
@@ -91,7 +91,7 @@ class currencies extends base {
       }
       $currency_value = zen_round($number * $rate, $this->currencies[$currency_type]['decimal_places']);
     } else {
-      $currency_value = $number; 
+      $currency_value = zen_round($number, $this->currencies[$currency_type]['decimal_places']);
     }
 
     return $currency_value;
--- a/includes/classes/currencies.php
+++ b/includes/classes/currencies.php
@@ -91,7 +91,7 @@ class currencies extends base {
       }
       $currency_value = zen_round($number * $rate, $this->currencies[$currency_type]['decimal_places']);
     } else {
-      $currency_value = zen_round($number, $this->currencies[$currency_type]['decimal_places']);
+      $currency_value = $number; 
     }
 
     return $currency_value;
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -497,7 +497,7 @@ class order extends base {
                                       'tax_groups'=>$taxRates,
                                       'tax_description' => zen_get_tax_description($products[$i]['tax_class_id'], $taxCountryId, $taxZoneId),
                                       'price' => $products[$i]['price'],
-                                      'final_price' => zen_round($products[$i]['price'] + $basket->attributes_price($products[$i]['id']), $decimals),
+                                      'final_price' => $products[$i]['price'] + $basket->attributes_price($products[$i]['id']),
                                       'onetime_charges' => $basket->attributes_price_onetime_charges($products[$i]['id'], $products[$i]['quantity']),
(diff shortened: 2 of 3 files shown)


includes/functions/functions_taxes.php
```diff
@@ -166,9 +166,9 @@ function zen_add_tax($price, $tax) {
 
     if (IS_ADMIN_FLAG === true) {
       if (DISPLAY_PRICE_WITH_TAX_ADMIN == 'true') {
-        return zen_round($price, $currencies->currencies[DEFAULT_CURRENCY]['decimal_places']) + zen_calculate_tax($price, $tax);
+        return $price + zen_calculate_tax($price, $tax);
       } else {
-        return zen_round($price, $currencies->currencies[DEFAULT_CURRENCY]['decimal_places']);
+        return $price;
```
drbyte: I'm assuming these changes are here to mirror the `else` condition on the following lines.

includes/classes/order.php
```diff
@@ -497,7 +497,7 @@ protected function cart(shoppingCart $basket, $customer_id = null, $sendto_addre
                                       'tax_groups'=>$taxRates,
                                       'tax_description' => zen_get_tax_description($products[$i]['tax_class_id'], $taxCountryId, $taxZoneId),
                                       'price' => $products[$i]['price'],
-                                      'final_price' => zen_round($products[$i]['price'] + $basket->attributes_price($products[$i]['id']), $decimals),
+                                      'final_price' => $products[$i]['price'] + $basket->attributes_price($products[$i]['id']),
```
drbyte: why was this change made?

includes/classes/currencies.php
```diff
@@ -91,7 +91,7 @@ function value($number, $calculate_currency_value = true, $currency_type = '', $
       }
       $currency_value = zen_round($number * $rate, $this->currencies[$currency_type]['decimal_places']);
     } else {
-      $currency_value = zen_round($number, $this->currencies[$currency_type]['decimal_places']);
+      $currency_value = $number;
```
drbyte: Why this change?

includes/functions/functions_taxes.php
scottcwilson: Correct

includes/classes/order.php
scottcwilson: Rounding here is premature. You still have to multiply by quantity.

includes/classes/currencies.php
scottcwilson: Sorry, that was a failed experiment I should have removed.  Fixed.

drbyte:
I'd like to hear @zcwilt feedback after running the webtest suite against this.

zcwilt:
:+1:
I'm still having problems running webtests due to an incompatibility between the latest firefox version and the selenium/marionette driver. However unit tests do contain some pricing checks (mainly discount coupon price/tax stuff) 
given I cant see any probs logic wise, lets pass this pr. I'll look to move some other pricing checks from web tests to unit tests shortly.

scottcwilson:
Can we merge?

drbyte:
👍

PR #1196, opened 2016-08-14
Fix upgrade sql differences from v155 to v160

scottcwilson:
Currently, upgrade from 1.5.5 creates multiple logs


scottcwilson:
The symptom which initially alerted me to a problem: after upgrade, admin user 1 had no widgets.  Then I noticed several logs were created by the update.  Both issues are resolved by this change.

drbyte:
👍

Issue #1138, opened 2016-07-23
"There was a security error" msg NOT shown after excessive logins


scottcwilson:
I was suspicious about this, so I wanted to check it before working on #1135.  I confirmed that it does not work as expected.

Repro steps:
- Login with a bad password 8x.  You won't see the error on the 9th or 10th time.

Analysis: 
- admin/includes/functions/admin_access.php line 399 sets the redirect variable, which causes admin/login.php to do a header().  If this is taken out, the message will appear when a bad password is used, BUT the screen which is redrawn does not work - the $_POST variables aren't set when you submit it, so you don't get the normal login behavior.  (Expected: error message would be issued again.)

- Because admin/login.php does not look in $messageStack for a message, you can't save the message for later display, then do the redirect.  (Of course, this logic could be added if it makes sense.)

Thoughts on what to do?

netcrop:
1, Inside admin_access.php, function zen_session_destroy() will cause redraw of admin/index.php. Therefore no need to call zen_redirect() inside login.php in this case.
2, inside login.php Change if ( $redirect != '') zen_redirect($redirect) To if ($message == '' && $redirect != '') zen_redirect($redirect).
3, I set FILENAME_DEFAULT='https://absolutpath/admin/index.php'.

netcrop:
I have verified my suggestion and it works, the suggestion is:
1, inside file admin_access.php inside block
if ($_SESSION['login_attempt'] > 6 || $result['failed_logins'] > 6)
{
Add this line
 $message = ERROR_SECURITY_ERROR;
}
2, inside file login.php inside block
 if ($admin_name == '' && $admin_pass == '') {
...
} else {
Change 
 if ($redirect != '') zen_redirect($redirect);
To
 if ($message == '' && $redirect != '') zen_redirect($redirect);
}
3, During my test, the admin login page shows "There was a security error when trying to login." after 6 wrong passwords attempts.
4, Needs further investigation: 
Inside admin_access.php, function zen_session_destroy() flush the browser cache, the warning message don't show up on 8, 10, 12 ... woring passwords attempts.

netcrop:
I have verified my suggestion and it works, the suggestion is:
* inside file admin_access.php replace block
```
if ($_SESSION['login_attempt'] > 6 || $result['failed_logins'] > 6) { ... }
```
With this block
```
if ($_SESSION['login_attempt'] == 7 || $result['failed_logins'] == 7)
{
[...]

PR #1203, opened 2016-08-18
Fixed:#1138 "There was a security error" msg NOT shown after excessive logins

netcrop:
//To guaranty the redirection will not happen in this case.
$error = true;

Do I need submit same fix for v1.6.0 ?


drbyte:
I'd prefer that you redid this as a PR against v160, not against v155.

netcrop:
Yes, please reject this PR and I'll submit into v160 instead.

Issue #1202, opened 2016-08-17
AJAX: Session variables' changes aren't saved


lat9:
See [this ](https://www.zen-cart.com/showthread.php?220887-AJAX-Session-variable-changes-not-saved&p=1317262#post1317262)Zen Cart posting for details.

Essentially, the `/ajax.php` script needs to end with a require of `includes/application_bottom.php` so that any session-related changes made by an AJAX handler are properly recorded in the session.

drbyte:
Assigned to @zcwilt to determine the best approach for v160

drbyte:
Ref #1298 for an update to the v155 version of this

PR #1205, opened 2016-08-18
#1202: Save session after AJAX request

lat9:
Include application_bottom.php so that any session-related changes made
by an AJAX handler are properly saved.

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -50,3 +50,5 @@ if (isset ($_GET['act'])&&isset ($_GET['method'])) {
         }
     }
 }
+
+require ('includes/application_bottom.php');


drbyte:
👍

zcwilt:
:+1:

ajax.php
```diff
@@ -50,3 +50,5 @@ function utf8_encode_recurse($mixed_value) {
         }
     }
 }
+
+require ('includes/application_bottom.php');
```
drbyte: Question: Doesn't the `exit()` on line 47 negate the calling of application_bottom?  

So, second question: is this actually fixing the original thing you intended it to?

PR #1206, opened 2016-08-19
Fixes duplicate array keys

scottcwilson:


drbyte:
👍

admin/ezpages.php
```diff
@@ -243,9 +243,7 @@ function zen_set_ezpage_status($pages_id, $status, $status_field) {
                         'status_header' => '',
                         'status_sidebox' => '',
                         'status_footer' => '',
-                        'status_toc' => '',
-                        'page_open_new_window' => '',
-                        'page_is_ssl' => ''
+                        'status_toc' => ''
```
drbyte: This is a good opportunity to point out that it is good practice in the PHP world to end even the last element of an array with a `,` 

For example, if `'status_toc' => ''` had kept a comma at the end of the line, then that would have been 2 less lines involved in this diff.

Javascript can't handle the extra commas in its array syntax (json too). But PHP does, and this is one reason why it is increasingly embraced.

PR #1209, opened 2016-08-20
Fixed:#1138 "There was a security error" NOT shown after excessive logins

netcrop:
a careful solution that require least change in the code.
* On every seventh invalid attempt ```zen_session_destroy()``` will be called and that is 86% of times.
* To prevent redirection that cause the warning message not shows up: ``` $redirect = ''```


netcrop:
Here is the purpose by using ```zen_session_destroy()``` in admin_access.php

* ```zen_session_destroy()``` could be replaced with                           ```$_SESSION['login_attempt'] = 0;``` since one important aspect is to prevent ```$_SESSION['login_attempt']``` grown beyond a barrier that trigger the             ```header('HTTP/1.1 406 Not Acceptable')``` inside password_forgotten.php.
But what if this trigger was made on purpose as a slam prevention? That would  require a ordinary user to restart the browser for requesting a password       reset. Not so convenient. Therefore ```zen_session_destroy()``` **is** been used.

* The same scenario happens after user had required password reset, but before she activate the reset link via email. When she again tries to login the site  with several invalid attempt that trigger the slam prevention. And that is why **no**```zen_session_destroy()``` been used this time.

PR #1210, opened 2016-08-20
Call uses 1 parameter, 0 expected

scottcwilson:


drbyte:
👍 I'll accept this, but note that there is a broader scope involved with the affected code.

PR #1221, opened 2016-08-24
Fixes #1207 widgets_description should not have language_id field

### LINKED ISSUE #1207: Dashboard widgets description table should not have a languages_id column
Widgets do have language defines of course but the widgets should work in any language.  This column should be removed.

scottcwilson:


zc_install/sql/install/mysql_zencart.sql
```diff
@@ -3400,45 +3398,45 @@ INSERT INTO get_terms_to_filter VALUES ('record_company_id', 'TABLE_RECORD_COMPA
...
-('counter-history', 'COUNTER_HISTORY', '', 1),
-('new-orders', 'NEW_ORDERS', '', 1),
-('logs', 'LOGS', '', 1)
+INSERT INTO dashboard_widgets_description (widget_key, widget_name) VALUES
+('general-statistics', 'GENERAL_STATISTICS'),
+('order-summary', 'ORDER_SUMMARY'),
+('new-customers', 'NEW_CUSTOMERS'),
+('counter-history', 'COUNTER_HISTORY'),
+('new-orders', 'NEW_ORDERS'),
+('logs', 'LOGS')
```
drbyte: Given that the `description` table is now only used to list the name of the language-constant, perhaps it should be retired, and the language-constant moved into the `dashboard_widgets` table?

drbyte: This PR looks good. I'm torn as to whether my suggestion to drop this extra table should be part of this PR or separate on its own.  
In the end either is fine.

scottcwilson: Changes like this are quite time consuming to test (because you have to do both create and upgrade) so I'd rather just stay with this update if you don't feel strongly.

drbyte: /me punts to @zcwilt

drbyte:
👍 
I think this can stand on its own.
If there are tweaks to be done to this, we can do them as another PR.

ajeh:
:+1:

PR #1240, opened 2016-08-27
Fixes styling that caused the messageStack to hide the nav menu

### LINKED ISSUE #1225: Displaying a header alert on admin home page wipes out menus
Repro steps: On an updated 160 site, go to admin->configuration->website maintenance to true.  (This would be done automatically on an upgrade).  

You will then notice your header menu items are obfuscated.  See image. 
bad_menu

barco57:


drbyte:
👍  ... but I'd like to see this put into the main stylesheet `/admin/includes/template/css/stylesheet.css` instead of `adminLTE.css` since the adminLTE file might get replaced at any point by applying an update.

Using `inherit` seems to apply the same effect, in my tests.
```css
.main-header {max-height: inherit;}
```

Whatcha think?

barco57:
What is the parent you are inheriting from?

drbyte:
I dunno ... But since adminLTE.css is loaded before stylesheet.css it seemed logical to just override whatever adminlte.css set by telling it to inherit instead of have the specific 100px limit

barco57:
well, renaming AdminLTE.css to AdminLTE.xxxxxxxxx and NOT putting anything in the main stylesheet also makes the hiding problem go away... 
So I suppose the better question is it more likely that in the future AdminLTE may go away completely or just be updated?

drbyte:
Well, I've been toying with removing it, but there are some other dependencies that would need changing too.

I can merge what you've got, and try to remember to override it myself later if we decide to keep it. 

It's just best-practice to never update the files of another imported framework directly .... just like we tell people to never do direct-edits to Zen Cart's template_default files.

PR #1242, opened 2016-08-28
TABLE_DASHBOARD_WIDGETS_GROUPS also does not need a languages id column

scottcwilson:


drbyte:
👍

zcwilt:
:+1:

barco57:
language_id Primary Key still needs to be removed on the table schema. Currently, an install stops at that point with a database problem.

```
DROP TABLE IF EXISTS dashboard_widgets_groups;
CREATE TABLE IF NOT EXISTS dashboard_widgets_groups (
  widget_group varchar(64) NOT NULL,
  widget_group_name varchar(255) NOT NULL,
  PRIMARY KEY (widget_group,language_id)
) ENGINE=MyISAM;
```

scottcwilson:
Thanks @barco57.  PR #1252 submitted.

PR #1251, opened 2016-08-28
Fix gravatar image resolution

### LINKED ISSUE #1241: Large Gravatar image distorted
I noticed while playing around that the gravatar in the popup when you hover the little gravatar in the menu is hideous. Very distorted because it is using the same image for both, 20px x 20px

large_gravatar_fugly

Looking at the code it seems that only the 20px x 20px version is requested from gravatar.com using this function
```
protected function getGravatar( $email, $s = 20, $d = 'mm', $r = 'g', $img = false, $atts = array() )
```
So, either there needs to be two functions, one to pull a 90px x 90px and one for the 20px x 20px or the function needs to be changed to pull the 90px version and then styling changes added to downsize it to 20px.

drbyte:
Fixes #1241 - credits to @barco57

### DIFF
--- a/admin/includes/template/css/stylesheet.css
+++ b/admin/includes/template/css/stylesheet.css
@@ -640,3 +640,4 @@ td.underline {
     border-bottom: 1px solid #000;
     padding: 0px 7px;
 }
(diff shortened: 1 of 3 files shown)


zcwilt:
:+1:

PR #1254, opened 2016-08-29
Removes deleted field from index

scottcwilson:
Forgot to remove language_id from index when removing from
dashboard_widgets_groups table.


scottcwilson:
Thanks to @barco57 for catching this error.

PR #1255, opened 2016-08-29
fixes banner statistics widget-key in dashboard_widgets_to_users

barco57:
the current widget_key for banner statistics is incorrect (banner_statistics versus banner-statistics) causing the banner statistics dashboard widget to not display


drbyte:
Thanks

PR #1262, opened 2016-08-31
Based on call in orders.php line 562, function should be static.

scottcwilson:


zcwilt:
:+1: 
while you are editing this function you could get rid of the 
    $new_fields = '';
line

ajeh:
:+1:

PR #1263, opened 2016-09-01
Fixing erroneous extra parameter

scottcwilson:
create_add_products function from orders class only has 1 parameter.


zcwilt:
:+1:

ajeh:
:+1:

PR #1265, opened 2016-09-02
Function re-uses loop variables for inner and outer loop

scottcwilson:
- inner loop at line 282 - use different variables than are used on line 81.
- same at line 364 vs line 81.
This is not a true bug but someone could accidentally create one by not
being aware of the reuse.

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -78,7 +78,7 @@
     if ($from_email_name == $from_email_address) $from_email_name = STORE_NAME;
 
     // loop thru multiple email recipients if more than one listed  --- (esp for the admin's "Extra" emails)...
-    foreach(explode(',',$to_address) as $key=>$value) {
+    foreach(explode(',',$to_address) as $value) {
       if (preg_match("/ *([^<]*) *<([^>]*)> */i",$value,$regs)) {
         $to_name = str_replace('"', '', trim($regs[1]));
         $to_email_address = $regs[2];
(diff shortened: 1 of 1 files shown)


drbyte:
Ya, the re-use is harmless in this case, since `$key` from line 81 is never used anywhere later.

So, a simpler fix would be to just change line 81 and change `$key => $value` to just `$value`

scottcwilson:
That makes no sense. You are still leaving one potential bug in place.  I have spent HOURS debugging issues with this very same root cause.

drbyte:
I must not be understanding what you mean then. 

From what I can see:
- `$key` on 81 isn't being used purposefully anywhere, so is irrelevant. Dropping it from 81 adds some safety by removing ambiguity.
- `$key` on 282 is likewise not used; changing `$key => $val` to `$attachment` (and on subsequent lines) adds the safety by removing ambiguity. (I didn't suggest this one earlier, but should have.)
- the loop on 364 is really tiny, and perfectly suited for using `$key=>$val` ... and since $key and $val are no longer used above, there's no clash.

Or am I missing what you're trying to say?

scottcwilson:
Oh I see - your comment earlier was not to use `$attachment` but just `$value`. 

screen shot 2016-09-04 at 2 51 19 am

So I'm ok with the line 282 change but can we keep my change to line 364, since this preserves the uniqueness of variables across the outer loop?  PR Updated.

drbyte:
> but can we keep my change to line 364, since this preserves the uniqueness of variables across the outer loop? 

I had also meant to change line 81 (the outer loop) so that there would be no clash at 364

```diff
-     foreach(explode(',',$to_address) as $key=>$value) {
+     foreach(explode(',',$to_address) as $value) {
```

scottcwilson:
Got it - thanks for clarifying.

drbyte:
👍

zcwilt:
:+1:

PR #1273, opened 2016-09-05
Add option to bypass archiving "send admin copy of" emails

scottcwilson:


barco57:
hmmm... If I agree that this is a good idea I am sure to come to regret it in the future. I can not count the number of times a site owner has come to me and said "I am not getting the order email from the cart" and I have gone to the email archive and been able to see that the cart did its part by sending the email.

wiztechinc:
Just went thru that with a cart so yeah, maybe not a good idea.

barco57:
If the goal is to try and trim down on the size that the email archive can get to over time because people forget it is there. I would consider moving the email archive viewer mod in as core, add a size/time alert to the header (like for the admin activity log) with store owner adjustments for how big they are willing to let it get or if how long they hang on to archived emails is more important. It could even have a widget to show the last X number of emails sent by the cart on the homepage.

drbyte:
@scottcwilson what was the inspiration behind submitting this?

scottcwilson:
@drbyte A client's email archive was horribly clogged.  This seemed like a reasonable countermeasure. 

@barco57 @wiztechinc Why would having an email copy in the archive be more convincing than having the original customer acknowledgement in the archive?  There are still a ton of things that could have gone wrong and caused the customer not to receive the email. 

It just seems wasteful to archive a copy.   What if I made it switchable in admin->configuration->email options?

drbyte:
> What if I made it switchable in admin->configuration->email options?

If adding it as a switch, I recommend adapting the existing `EMAIL_ARCHIVE` switch, and change the choices from `true/false` to `All / All except store owner / No`

One caveat:  In the case of order-confirmations, the store generates 2 messages. But in the case of contact-us, it only generates one ... and if the contact-us email address is the same as EMAIL_FROM then those won't get archived and there'll be no record of them in the event of email failure.

scottcwilson:
@drbyte Would you mind if I added a parameter to zen_mail that would indicate if the email was being sent as a copy (i.e. in accordance with a SEND_EXTRA* configuration item)?  It's a lot more changes but then I could just check on this plus the EMAIL_ARCHIVE switch.  

Otherwise, as you pointed out, this could fail to archive in cases where it would be useful to archive. 

I am ok abandoning this if it's unpopular.

drbyte:
> Would you mind if I added a parameter to zen_mail that would indicate if the email was being sent as a copy

It already does: The `$module` param contains the word `_extra` in those cases.

scottcwilson:
Nice!  I had forgotten that.  (There is one defect where `includes/library/zencart/Controllers/src/admin/GvQueue.php` does not observe this convention; will fix.)

drbyte:
👍  ... Do you mind squashing into one commit?

scottcwilson:
Not at all, if you feel it's good to go.

zcwilt:
:+1:

PR #1282, opened 2016-09-09
Remove stray line from long ago

### LINKED ISSUE #1174: shopping_cart/header_php.php switch break on default and $cur_row++
```
  switch (true) {
    case (SHOW_SHOPPING_CART_DELETE == 1):
    $buttonDelete = true;
    $checkBoxDelete = false;
    break;
    case (SHOW_SHOPPING_CART_DELETE == 2):
    $buttonDelete = false;
    $checkBoxDelete = true;
    break;
    default:
    $buttonDelete = true;
    $checkBoxDelete = true;
    break;
    $cur_row++;
  } // end switch
```
Looks to me like the break on default should not be there and the cur_row is in the wrong place...if even used at all.

scottcwilson:
Fixes #1174

### DIFF
--- a/includes/modules/pages/shopping_cart/header_php.php
+++ b/includes/modules/pages/shopping_cart/header_php.php
@@ -79,7 +79,6 @@ for ($i=0, $n=sizeof($products); $i<$n; $i++) {
     $buttonDelete = true;
     $checkBoxDelete = true;
     break;
-    $cur_row++;
   } // end switch
   $attributeHiddenField = "";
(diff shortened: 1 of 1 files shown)


drbyte:
Is this for #1174 ?

scottcwilson:
LOL I had it on my list of little fixes and forgot that it had already been reported.  Yes Fixes #1174.

zcwilt:
:+1:

PR #1284, opened 2016-09-11
Small fixes to zone validation function

scottcwilson:


includes/library/zencart/CheckoutFlow/src/AccountFormValidator.php
```diff
@@ -191,15 +191,14 @@ protected function checkStateZones($fieldName)
...
+            $foundExactIsoMatch = $this->tryFindExactZoneMatch($zone, $foundExactIsoMatch);
         }
         if ($foundExactIsoMatch) {
             $this->addressEntries['state']['zone_id'] = $zone->fields['zone_id'];
             $this->addressEntries['state']['zone_name'] = $zone->fields['zone_name'];
 
             return false;
-        }
-        if (!$foundExactIsoMatch) {
+        } else {
```
drbyte: Strictly speaking, it'd be even better to remove the `else` wrapper, like this:

OLD:
```php
if (foo) {
  blah;
  return false;
}
if (!foo) {
  blah;
  return true;
}
```

is optimized like this:
```php
if ($foo) {
  blah;
  return false;
}
blah;
return true;
```

Notice that the `else` wrapper is removed, and the code is just outdented. 
The reason this is ideal is due to the early-return that's done inside the first `if`, and since there's only one other execution path following it (first if was "true?", and last was "false?", with no other code after it), removing the wrapper around the `else` makes the code more readable, particularly because it's clear there's no other condition that should have been checked.

includes/library/zencart/CheckoutFlow/src/AccountFormValidator.php
```diff
@@ -191,15 +191,14 @@ protected function checkStateZones($fieldName)
         $zone = $this->getZoneFromDatabase();
         $foundExactIsoMatch = ($zone->RecordCount() == 1);
         if ($zone->RecordCount() > 1) {
-            $foundExactIsoMatch = tryFindExactZoneMatch($zone, $foundExactIsoMatch);
+            $foundExactIsoMatch = $this->tryFindExactZoneMatch($zone, $foundExactIsoMatch);
```
drbyte: nice catch ;)

zcwilt:
:+1:

ajeh:
:+1:

PR #1291, opened 2016-09-15
Fixes #1275 checks ini_set function on install

### LINKED ISSUE #1275: add check in zc_install for @ini_set
see: https://goo.gl/SmTrPy

ini_set was a disabled function and therefore no myDEBUG-.... logs were getting created.

proseLA:
i think this is correct.

Fixes #1275

### DIFF
--- a/zc_install/includes/systemChecks.yml
+++ b/zc_install/includes/systemChecks.yml
@@ -49,6 +49,12 @@ systemChecks:
         parameters:
           version: 5.6.0
           versionTest: ge
+  checkIniSet:
+    runLevel: always
+    errorLevel: WARN
+    mainErrorText: <? echo TEXT_ERROR_INI_SET . PHP_EOL; ?>
+    methods:
+      checkIniSet:
   checkZCversion:
     runLevel: always
     errorLevel: WARN
--- a/zc_install/includes/systemChecks.yml
+++ b/zc_install/includes/systemChecks.yml
@@ -52,7 +52,7 @@ systemChecks:
   checkIniSet:
     runLevel: always
     errorLevel: WARN
-    mainErrorText: <? echo TEXT_ERROR_INI_SET . PHP_EOL; ?>
+    mainErrorText: <?php echo TEXT_ERROR_INI_SET . PHP_EOL; ?>
     methods:
       checkIniSet:
   checkZCversion:
--- a/zc_install/includes/classes/class.systemChecker.php
+++ b/zc_install/includes/classes/class.systemChecker.php
@@ -690,4 +690,15 @@ if ($errnum != 0) error_log('CURL Connect: ' . $errnum . ' ' . $errtext . "\n" .
       }
     }
   }
+  public function checkIniSet() {
+      $orig = ini_get('error_log');
+      if ($orig !== 'test_log') {
+          ini_set('error_log', 'test_log');
+      } else {ini_set('error_log', 'test_log456');}
+      if ($orig !== ini_get('error_log')) {
+          return true;
+      } else {
+          return false;
+      }
+  }
 }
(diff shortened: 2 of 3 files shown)


proseLA:
i have thoughts....
#1 - point taken and concur.  i should pay more attention and at some point get more familiar with IDE. ;)
#2 - as far as i can tell, we are talking about constructing a new function - get_disabled_funcs, which is fine, but i think would be dependent on ini_get.  we already have an ini_get check, and i was thinking about possibly tying these two functions somehow together.  i am not that familiar with yaml; but without ini_get, i'm not sure we will be able to tell if ini_set is disabled.... perhaps we should work with these together or make this check dependent on that check?  
#3 - i thought about that after my pull request.  concur!
#4 - sure.

drbyte:
> second - as far as i can tell, we are talking about constructing a new function - get_disabled_funcs, which is fine, but i think would be dependent on ini_get. we already have an ini_get check, and i was thinking about possibly tying these two functions somehow together. i am not that familiar with yaml; but without ini_get, i'm not sure we will be able to tell if ini_set is disabled.... perhaps we should work with these together or make this check dependent on that check?

No, I wouldn't create another function.  Just do the 2 checks in the same function.

I mentioned "get_disabled_funcs" as a concept. There might even be a PHP function for that already. If not, do an ini_get('disabled_funcs') (I'm talking off the top of my head without researching the actual name here).

```
function checkIniSet() {
// a. check for ini_set disabled in PHP
// b. if not disabled, test whether changing error_log works
}
```

proseLA:
sounds good...  i'll get to it soon.  no worries.

i think i'm going to want to some more testing esp. if ini_get is disabled.  there is no php function for disabled functions that i have found.  

best.

drbyte:
http://www.php.net/manual/en/ini.core.php#ini.disable-functions

So maybe consider something like:
```php
/**
 * return true if ini_set is reliable; false if it is disabled
 */
function checkIniSet() {
  // test if ini_set is disabled in php.ini
  if (strpos(ini_get('disable_functions'), 'ini_set') !== false) return false;

  // if we get here, ini_set is enabled, so lets test it
  $orig = ini_get('error_log');
  ini_set('error_log', 'a_random_place');
  if (ini_get('error_log') != 'a_random_place') return false;
  ini_set('error_log', $orig);
  return true;
}
```

proseLA:
i have tested with ini_get disabled.  there are many php errors sent to the screen, that are, IMHO, outside the scope of this PR.  i do not see the nicely formatted ZC error related to it on zc_install page, but pretty apparent to me there are problems.  i do think, however, that this addresses the ini_set issue.
screenshot from 2016-09-15 10-56-11

zc_install/includes/systemChecks.yml
```diff
@@ -49,6 +49,12 @@ systemChecks:
         parameters:
           version: 5.6.0
           versionTest: ge
+  checkIniSet:
+    runLevel: always
+    errorLevel: WARN
+    mainErrorText: <? echo TEXT_ERROR_INI_SET . PHP_EOL; ?>
```
drbyte: `<?` should be `<?php`

drbyte:
> i have tested with ini_get disabled.

Ya, that'll cause problems.

And there are a hundred functions we could do explicit testing for (ie: every function the app uses), but that's kinda unrealistic. 

I could be wrong, but IMO it's pretty reasonable to expect ini_get to not be a disabled function.

drbyte:
>  i do think, however, that this addresses the ini_set issue.

Agreed.

drbyte:
Just need to fix that `<?` vs `<?php` error.

proseLA:
yes i agree on ini_get....

drbyte:
👍

zcwilt:
:+1:

ajeh:
:+1:

PR #1294, opened 2016-09-21
removes duplicate code in extra_configures dir

proseLA:
in the interest of D.R.Y., on the admin side, loads all code from includes/extra_configures, and then loads code from admin/includes/extra_configures.  i have removed any duplicates from admin/includes/extra_configures, as well as preserved the *-adm-* naming feature for logging.  one consideration would be code loaded from includes/extra_configures that one would NOT want loaded from the admin.

also, 2 file changes were reformatted according to the PSR-2 standard, so changes look a bit more substantial than they are.


drbyte:
Hmm ... I think this is likely to cause some problems, due to differences between catalog and admin.

- media_manager.php -- note how you're loading the catalog one first, which sets a define, which is then later superceded by the admin media_manager.php file which tries to redefine, which isn't possible, and is going to throw an error. 

- the pci and security patches on catalog-side look like they might interfere with the admin sanitization infrastructure, and so could be at cross-purposes with one another.


An alternate approach could be to backtrack and focus on simply the enable_error_logging file being relocated, and follow the same method as used by the set_catalog-time_zone file ... which basically just explicitly loads the corresponding catalog file instead of duplicating code on the admin side. That accomplishes DRY just as well, and is more explicit about what exactly is to be done instead of wildcarding the loading of all the catalog files into admin without limits.

proseLA:
Ok, I have no problem doing it as you suggested.  However, what if we were to completely eliminate the admin/includes/extra_configures folder, load everything from the includes/extra_configures folder; and then explicitly use what we need based on the IS_ADMIN_FLAG?  We can accomplish D.R.Y. as well as hopefully simplify the structure for this added includes directory?  In development, I make use of  this dBug tool (http://dbug.ospinto.com/), and i’m forced to symlink on the admin side to the catalog side.  I am not sure how many other people are adding things here in this manner.  I also have a file that contains client specific modifications in the includes/extra_datafiles/ folder that instead of symlinking it on the admin side, I have modified the admin/includes/application_top.php to explicitly include that file from the catalog side.

It is just an idea, and some food for thought, as I think it would be cleaner.  But I have no problem doing it as you suggested.

With regards to your other comments, I do not make use of the media manager, but perhaps I’m missing something.  Both media_manager.php files only contain one definition The admin is defining DIR_FS_CATALOG_MEDIA:

    define('DIR_FS_CATALOG_MEDIA', DIR_FS_CATALOG . 'media/');

The catalog is defining DIR_WS_MEDIA:

    define('DIR_WS_MEDIA', 'media/');

I did test this code and no errors were thrown.  So I think I do not understand where the redefine is.  What am I missing here?

With regards to the pci and security patches, I think you have a valid point.  I do remember the problems associated with the admin sanitizer.  I think this is another opportunity to sanitize all in one place as opposed to having it in more than one place.  There should be 1 sanitizer class that handles everything.  Or at least have the admin sanitizer extend the base sanitizer class (or the other way around…  I’m not sure which side would be more restrictive and need additional functionality).  For v1.6, is it necessary to carry over patches as opposed to incorporating it into the base code?

drbyte:
> Both media_manager.php files only contain one definition

Oops. My bad. You're right, there's not duplication there. 

> For v1.6, is it necessary to carry over patches as opposed to incorporating it into the base code?

Good point. Most of the time we tend to keep patches until the next major release, at which time we roll them into core code. You'll see in several cases we leave the file behind but empty it out and replace its contents with comments stating that it is no longer needed, just to be explicit, and to mitigate against ill-effects of keeping the old patch file around if logic is already in core. 
In v155 the admin sanitization was a bit of a bigger leap than we'd normally like to do without releasing a major version bump, so there is indeed a bit of disparity between admin/catalog there.  But now we're getting into a completely different topic.

drbyte:
> In development, I make use of this dBug tool (http://dbug.ospinto.com/), and i’m forced to symlink on the admin side to the catalog side. I am not sure how many other people are adding things here in this manner.

I have a small handful of dev-specific extra-configures files that I use to quickly enable various debugging or dev-only components. I don't find it that onerous to copy them into corresponding admin and non-admin folders, but I understand your point. 
I also name them with a `dev-` prefix, so that the `.gitignore` rules auto-exclude them from getting committed into my git repos. It also makes them easy to spot.
I'm not sure you need to symlink, but if that's simpler for you, feel free.

drbyte:
> what if we were to completely eliminate the admin/includes/extra_configures folder, load everything from the includes/extra_configures folder; and then explicitly use what we need based on the IS_ADMIN_FLAG? 

Hmmm ... I'd like to ponder this awhile.  I see several pro's and a few con's.  I'd welcome others' input. 

But I'd prefer that the consolidation of the error_logging code be treated as its own PR.

proseLA:
i did notice the leftover empty file in one of those directories....

feel free to ponder....  i will modify this PR as you have suggested...

proseLA:
git is freaking killing me right now....

proseLA:
next time i will try and do separate commits; one for the code changes and one for any reformatting to the PSR-2 standard.

zcwilt:
:+1:

PR #1298, opened 2016-09-27
Make ajax handler respond only to ajax calls

### LINKED ISSUE #1296: ajax.php: Should $robotsNoIndex be set?
I've been chasing an issue for the past year on one of my commercial plugins that makes use of the Zen Cart's default ajax.php handling.  That plugin's AJAX handler performs some input validation and I've seen multiple reports of the plugin logging the fact that its parameters aren't properly set.

I believe that this is due to search engines finding and calling the ajax.php function.  As such, I was thinking that having the ajax.php module include (prior to its load of application_top.php) the following statement:

`$robotsNoIndex = true;`

might mitigate that invalid usage.

drbyte:
Make ajax handler respond only to ajax calls
And clean up the code so that sessions are closed properly under all execution paths.

Fixes #1296

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -9,46 +9,67 @@
  * @version $Id: Author: zcwilt  Thu Dec 31 19:12:00 2015 +0000 Modified in v1.5.5 $
  */
 
-require ('includes/application_top.php');
+// Abort if the request was not an AJAX call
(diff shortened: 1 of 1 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #1299, opened 2016-09-28
Removed unnecessary <br /> tags in the order information table

GaryH21:
I've removed the unnecessary <br /> tags in the order information table in admin/orders.php


zcwilt:
:+1:

PR #1304, opened 2016-10-06
removed code from application_top that is repeated in extra_configure…

llynix:
While reading through the source I noticed the time zone is set twice.  Once in application_top and then shortly after in includes/extra_configures/set_time_zone.php

This removes the code from application_top as it is more configurable via set_time_zone.php


zcwilt:
:+1:

ajeh:
:+1:

PR #1306, opened 2016-10-11
Add instructions for alternate admin order-confirm email template

drbyte:
Added comments to demonstrate how to specify an alternate admin-only order confirmation email template, directly in order.php class.


zcwilt:
:+1:

ajeh:
:+1:

PR #1308, opened 2016-10-11
Move inline error message text to language file

drbyte:
(since v1.3.9)

Ref https://www.zen-cart.com/showthread.php?220775

Thanks @lat9


zcwilt:
:+1:

ajeh:
:+1:

PR #1311, opened 2016-10-12
Fix: if fax disabled in config, should also disable in account-edit tpl

drbyte:
Ref: https://www.zen-cart.com/showthread.php?221142-Fax-number-on-tpl_account_edit_default-php


zcwilt:
:+1:

ajeh:
:+1:

PR #1312, opened 2016-10-12
Fix javascript selector ids for collects-card-data field

drbyte:
Ref: https://www.zen-cart.com/showthread.php?220961-collectsCardDataOnsite-payment-class-not-creating-correctly-named-hidden-fields

Thanks @lat9


drbyte:
Thanks @lat9

zcwilt:
:+1:

ajeh:
:+1:

PR #1313, opened 2016-10-12
Enable the "back" button on checkout address edit pages

drbyte:
Ref: https://www.zen-cart.com/showthread.php?221119-Checkout-shipping-payment-addresses-Can-t-cancel-if-only-1-address

Thanks @lat9


WebsiteRob:
I noticed the files to be changed refer to "responsive_classic" template files and yet, it uses the same applicable code as "template_default" files. No mention though of changing "template_default" files.

lat9:
@WebsiteRob, it was only an issue with the `responsive_classic` template; `template_default` required no change.

PR #1316, opened 2016-10-14
Fix: checkout_shipping: No method chosen after change virtual to mixed

drbyte:
Fix checkout_shipping: No method chosen after cart goes from virtual to mixed

Thanks @lat9

Ref https://www.zen-cart.com/showthread.php?220727-checkout_shipping-No-method-chosen-after-cart-goes-from-virtual-to-mixed


ajeh:
:+1:

PR #1321, opened 2016-10-18
Fix Chrome browser incompatibility with javascript function name

drbyte:
... because evidently Chrome 54.0.2840.59 doesn't like javascript functions named 'animate'.


zcwilt:
:+1:

ajeh:
:+1:

PR #1323, opened 2016-10-18
Rearrange function declarations in ajax handler for compatibility

### LINKED ISSUE #1318: ajax.php: Functions need to be defined before use (155 branch)
When using the ajax.php file on a ZC 1.5.5a installation, I receive an "Error 500" when the interface is used:

`PHP Fatal error:  Call to undefined function utf8_encode_recurse()`

The issue is that the utf8_encode_recurse function doesn't exist and the fall-back declaration (as well as the ajaxAbort declaration) occurs **after** the functions' invocation.

drbyte:
Fixes #1318


zcwilt:
:+1:

ajeh:
:+1:

PR #1325, opened 2016-10-19
Fixes #1264 Admin toplevel links don't work

### LINKED ISSUE #1264: Admin mobile view menu broken
mobile view menu opens to top level but sub menu items do not open.

scottcwilson:


drbyte:
This now makes the old Alt-Nav menu appear whenever one clicks on the top-level menu items ... which in mobile mode completely skips the contextual submenu from ever displaying. 

:(

scottcwilson:
What is the contextual submenu?  Can you explain how you think it should work?

zcwilt:
So here is an example
https://getbootstrap.com/examples/navbar/

Note the horizontal `dropdown` menu has contextual entries.

If you shrink the window, you will get the standard `hamburger` however the `dropdown` entry will still expand to show contextual entries.

zcwilt:
My css/html foo is poor, but it really looks like we don't just have the right `collapse` classes on the menu entries.

scottcwilson:
Oh yeah sorry I suck at that stuff.  I just wanted to at least get it to the level of 1.5.5.

PR #1331, opened 2016-11-01
Add shopping cart methods for calc in-cart details for category-lookups

drbyte:
(Backported from v160)

Also removes `number_of_uploads` from URI on add-to-cart redirects.


ajeh:
:+1:

zcwilt:
:+1:

PR #1336, opened 2016-11-12
replace order processing button with processing.gif at confirmation

proseLA:
ok...  i'm trying to not overthink it...  but...

the question remains how to override the gif?  right now it is hardcoded into the jquery with template default.  but ideally we should be able to override it in a particular template....  or is that too much?

let me know and i can change it accordingly.

best.


drbyte:
Ya, accommodating template-specific overrides is a little more challenging, particularly since you have to use PHP to do some fallback detection.

One approach could be: 

a) add something like this to the top of `jscript_zen_general_top.php`:

```php
<script> var $processingGif = '<?php echo zen_image(DIR_WS_TEMPLATE_IMAGES . 'processing.gif', 'Processing ...', '', '', 'id="btn_submit" border="0"'); ?>'; </script>
(followed by the existing content of the file)
```

b) then your `replaceWith` logic becomes something like:

```javascript
  if ($processingGif == undefined) $processingGif = '<img id="btn_submit" src="includes/templates/template_default/images/processing.gif" border="0" alt="Processing" title="Processing">';
  $("#btn_submit").replaceWith($processingGif);
```

proseLA:
that works...

drbyte:
Great!
👍 

Thanks for submitting.

 /me punts to @zcwilt for review

zcwilt:
I think @drbyte punted to me as I have some code in the admin to allow for setting javascript variables as part of the template structure, and because  I'm in the process of refactoring our unit/acceptance tests. 

I don't think at the moment these have relevance, and on review of code happy to :+1:

PR #1337, opened 2016-11-13
Add invalid email address message fixes #1295

### LINKED ISSUE #1295: error meesage on login - opinions wanted....
if a user enters the wrong credentials, ZC provides the following error message:

Error: Sorry, there is no match for that email address and/or password.

i have clients that want ZC to notify the user if it is the username or password that is the problem.  i am curious what the community thinks with regards to telling the user which one it is.

proseLA:
Closes #1295 

i'm sorry; i was not sure about the proper way to add a new config value into the database.

i have it in a new temp file in the zc_install directory.  

please tell me the best way to handle this if the PR is on its way to being accepted and i will modify accordingly.

best.

### DIFF
--- a/includes/modules/pages/password_forgotten/header_php.php
+++ b/includes/modules/pages/password_forgotten/header_php.php
@@ -20,61 +20,73 @@ $_SESSION['navigation']->remove_current_page();
 
 if (isset($_GET['action']) && ($_GET['action'] == 'process')) {
 // Slam prevention:
-  if ($_SESSION['login_attempt'] > 9)
-  {
-    header('HTTP/1.1 406 Not Acceptable');
-    exit(0);
-  }
-  // BEGIN SLAM PREVENTION
-  if ($_POST['email_address'] != '')
-  {
-    if (! isset($_SESSION['login_attempt'])) $_SESSION['login_attempt'] = 0;
-    $_SESSION['login_attempt'] ++;
-  } // END SLAM PREVENTION
+    if ($_SESSION['login_attempt'] > 9) {
+        header('HTTP/1.1 406 Not Acceptable');
+        exit(0);
+    }
+    // BEGIN SLAM PREVENTION
+    if ($_POST['email_address'] != '') {
+        if (!isset($_SESSION['login_attempt'])) {
+            $_SESSION['login_attempt'] = 0;
+        }
+        $_SESSION['login_attempt']++;
+    } // END SLAM PREVENTION
 
-  $email_address = zen_db_prepare_input($_POST['email_address']);
+    $email_address = zen_db_prepare_input($_POST['email_address']);
 
-  $check_customer_query = "SELECT customers_firstname, customers_lastname, customers_password, customers_id
+    $check_customer_query = "SELECT customers_firstname, customers_lastname, customers_password, customers_id
                            FROM " . TABLE_CUSTOMERS . "
                            WHERE customers_email_address = :emailAddress";
 
-  $zco_notifier->notify('NOTIFY_PASSWORD_FORGOTTEN_CHECK_CUSTOMER_QUERY', array(), $check_customer_query);
+    $zco_notifier->notify('NOTIFY_PASSWORD_FORGOTTEN_CHECK_CUSTOMER_QUERY', array(), $check_customer_query);
 
-  $check_customer_query = $db->bindVars($check_customer_query, ':emailAddress', $email_address, 'string');
-  $check_customer = $db->Execute($check_customer_query);
+    $check_customer_query = $db->bindVars($check_customer_query, ':emailAddress', $email_address, 'string');
+    $check_customer = $db->Execute($check_customer_query);
 
-  if ($check_customer->RecordCount() > 0) {
+    if ($check_customer->RecordCount() > 0) {
 
-    $zco_notifier->notify('NOTIFY_PASSWORD_FORGOTTEN_VALIDATED', $email_address);
+        $zco_notifier->notify('NOTIFY_PASSWORD_FORGOTTEN_VALIDATED', $email_address);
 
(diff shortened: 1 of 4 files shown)


drbyte:
proseLA wrote: 
> i was not sure about the proper way to add a new config value into the database.

Generally we try to avoid adding any additional configuration settings unless they are essential. There's a lot of template-complexity and code-complexity that exist because of dozens of rarely-used configuration settings that were added for single-user requests back in early days. It's created some technical debt in the code, and we try to be very judicious about whether to add new ones.

But when essential, they go into the mysql_zencart.sql file (for new installs) and the mysql_upgrade_vZZZ.sql file (for upgrades)

drbyte:
Question: Why did you include all the commits for #1336 in this Pull Request?

proseLA:
with regards to #1336, probably because i messed up my git branch. i'm guessing i created a branch off that branch which is why it ended up here....  

i can fix that....  i think....

with regards to the added configuration value, i can key my changes off the PADSS_ADMIN_SESSION_TIMEOUT_ENFORCED value (or some other judicious chosen config value).  i'm with you on the sheer number of config values...  

let me know and i'll attempt to fix this PR.

best.

drbyte:
> with regards to #1336, probably because i messed up my git branch. i'm guessing i created a branch off that branch which is why it ended up here....

> i can fix that.... i think....

Thanks.

> with regards to the added configuration value, i can key my changes off the PADSS_ADMIN_SESSION_TIMEOUT_ENFORCED value (or some other judicious chosen config value). 

Yes, I see that you followed that pattern. I'm okay with that.

proseLA:
sorry about the git messup....  i think i have addressed that...  i have also used the PADSS_ADMIN_SESSION_TIMEOUT_ENFORCED variable as the key/trigger.  i do not think it is ideal, but i completely comfortable with it....  as someone who has struggled with the number of config values in ZC, i'm all on board....

includes/modules/pages/password_forgotten/header_php.php
```diff
@@ -66,14 +66,26 @@
...
+      } else {
+          $messageStack->add_session('password_forgotten', EMAIL_ADDRESS_NOT_FOUND);
+      }
   }
 
-  $messageStack->add_session('login', SUCCESS_PASSWORD_SENT, 'success');
+  /*$messageStack->add_session('login', SUCCESS_PASSWORD_SENT, 'success');
 
-  zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
+  zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL')); */
```
drbyte: Given that you've relocated these lines above, let's not leave them here commented-out.

includes/modules/pages/password_forgotten/header_php.php
```diff
@@ -66,14 +66,26 @@
...
+      }
+      zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
 
   } else {
     $zco_notifier->notify('NOTIFY_PASSWORD_FORGOTTEN_NOT_FOUND', $email_address);
+      if (PADSS_ADMIN_SESSION_TIMEOUT_ENFORCED) {
+          $messageStack->add_session('login', SUCCESS_PASSWORD_SENT, 'success');
+          zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
+      } else {
+          $messageStack->add_session('password_forgotten', EMAIL_ADDRESS_NOT_FOUND);
```
drbyte: Does this `add_session` actually work as expected? Given that you're bypassing the `zen_redirect` here, you probably need to use `add` instead of `add_session` (since `add_session` is for making the message show up on the "next" page, not the current page).

I suggest testing the use of `add` vs `add_session`, to determine whether you're getting the desired result.

includes/modules/pages/password_forgotten/header_php.php
proseLA: no problem...

includes/modules/pages/password_forgotten/header_php.php
proseLA: drB, i assume you are talking about line 82?  i have no problem changing that, but it does work as expected and i did test it.  i did change the class to `password_forgotten`, as in this situation where the user enters an invalid email address, we could like to stay on this page, which it currently does. 

i did now just try it with `add` as opposed to `add_session`.  they both work fine.  i have no problem changing it to `add`.

drbyte:
Looking good. 

Do you mind rebasing (& squashing) all your commits into one?

proseLA:
i've attempted to squash and rebase....  let me know if it looks ok...  first time in this arena....

drbyte:
Hmmm .... no; ideally it would end up turning the 8 separate commits into a single commit.

Process:
- "rebase interactive"
- mark all except the first commit as squashed
- update the actual final commit message to capture the meaning of the overall commit
- force-push the branch back to github

proseLA:
now??

drbyte:
Ya, that's more like it. 

You must have also used another editor, because several unrelated lines have been reformatted; while I"m not very bothered about that (usually it's good), it always irritates me when IDEs mangle things by wrapping code to new lines and actually make the code harder to read. lol

@zcwilt are you okay with this?

proseLA:
on the squash and rebase, i think the 1st time i did, git told me i needed to pull down as opposed to doing a force-push.  it got rather frustrating as the 2nd time (when i got it right), i had to resolve a number of conflicts... but i will try and remember the force-push.  and with regards to the squash and rebase, do you use something like this: 
`git reset --soft HEAD~3 &&
git commit`
or something like this:
`git rebase -i HEAD~4`

finally, with regards to that whole code change, again i apologize.  as i said, there was a bunch of rebasing conflicts that git needed me to address; and i'm using the same editor....  if anything it looks like PhpStorm was applying the PSR-2 coding standard as i was forced to address those conflicts.

zcwilt:
:+1:

drbyte:
:+1:

PR #1339, opened 2016-11-20
Additional order notifiers

lat9:
Add some notifiers to allow:
1) Additional action-buttons to the order-edit view.
2) Additional icons to the order-listing legend
3) Additional icons, based on the above legend, to prefix an order's listing row
4) Additional "action" icons to be added to the order's "action" column.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -115,6 +115,9 @@
         $order->doVoid();
         zen_redirect(zen_admin_href_link(FILENAME_ORDERS, zen_get_all_get_params(array('action')) . 'action=edit'));
         break;
+      default:
+        $zco_notifier->notify('NOTIFY_ADMIN_ORDERS_DEFAULT_ACTION', $oID, $order);
+        break;
     }
   }
 require('includes/admin_html_head.php');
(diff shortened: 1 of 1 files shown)


drbyte:
👍

drbyte:
(would prefer to have this rebased so all the branch merges aren't included .... is that something you're comfortable doing?)

lat9:
I'm still a GitHub neophyte, but I can follow instructions ... sometimes!  Point me to an instruction-set and I'll give it a try!

drbyte:
Usually it entails something akin to:
1. Do a `git rebase --interactive`.  This should show you all the commits in the list.
2. In the list of commits, you can set the letter in the first column of each row to `s` to squash it to be included with the commit after it. 
3. You can also edit the overall commit message (usually to remove the "merged this, pulled this, merged again, deleted, etc" parts and just have the main message that you want to be in the commit
4. When the rebase is done, then you do a `git push --force` to replace what's up here on github with the cleaned-up version.

If you're not used to using the command-line to work with git, then your GUI tool probably has a "rebase iteractively" option, which allows the same process.

lat9:
Uncle!  I'll (hopefully) follow the process you outlined in the future.  I think that the issue I created for myself was that I pulled from the master zencart github into my github instead of using my local repository as a hand-off point.

drbyte:
Ya, it does look like that.

This is more drastic, but I once (intentionally) just deleted my personal clone of the zencart repo from my own github account, then re-forked the main one from zencart/zencart, and then pulled it down.  It accomplishes the same thing in the end. The downside is that it loses any branches I'd made (but if I don't need to keep them then it's a non-issue).

drbyte:
I've cleaned it up.

What I did:
- checkout the `v160` branch
- pull down and merge all the `v160` updates so that your own v160 branch is fully current
(The above steps are best done before creating a new feature branch, to avoid merge challenges.)
- change to the additional_notifiers branch
- `git rebase --interactive v160` ... to tell it to show all the commits that are different between this branch and the v160 branch
- `pick` and `squash` commits accordingly
- `git push --force` to update github

I was only able to update this branch, because there's an open PR for it. I can't touch your v160 branch. That's up to you :)

lat9:
Thanks, DrByte.  I appreciate (as always) your help!

zcwilt:
:+1:

ajeh:
:+1:

PR #1352, opened 2016-12-20
Move product-type detection logic up a step, so it cascades to other types

drbyte:
Product-type detection logic wasn't cascading to delete/copy modules. 
Moving it up a step, and removing the interception in products.php makes it more universal.

Ref: https://www.zen-cart.com/showthread.php?221596-product_type-incorrectly-set-in-admin-product-php&p=1323232#post1323232


zcwilt:
:+1:

ajeh:
:+1:

PR #1354, opened 2016-12-24
Prevent DB_CHARSET from being set to blank during upgrades

drbyte:
The constant was added in v150, so upgrades from prior versions were getting this set to blank, resulting in errors.
Ref https://www.zen-cart.com/showthread.php?221539-upgrade-DB_CHARSET-is-blank

This commit also includes same safety check for empty db-type.


zcwilt:
:+1:

ajeh:
:+1:

PR #1356, opened 2016-12-28
Change transaction_id field for authorizenet history to varchar

drbyte:
While the Authorize.net specifications indicate the transaction_id will be a numeric BIGINT value, other emulators (such as E-Processing) use string values, which trigger type errors. Changing to `varchar(32)` to accommodate this emulator anomaly.


zcwilt:
:+1:

ajeh:
:+1:

PR #1358, opened 2016-12-30
Standardize TLS connectivity for email

drbyte:
For consistency, TLS should always be port 587, and SSL always port 465.
(More and more hosts are converting to TLS, because it's way more secure than SSL.)

Also removed X-Mailer header signature, to avoid unnecessary disclosure.

Various code commenting updated/added.

Also updated related docs:
https://www.zen-cart.com/wiki/index.php/Troubleshoot_-_Email_Problems
https://www.zen-cart.com/wiki/index.php/Troubleshoot_-_Understanding_Causes_Of_Lost_Emails

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -2,25 +2,35 @@
 /**
  * functions_email.php
  * Processes all outbound email from Zen Cart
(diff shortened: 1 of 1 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #1360, opened 2017-01-01
TravisCI: echo mysql version & skip admin config file (no longer used)

drbyte:
(Travis doesn't currently display the installed version of MySQL in its list of installed resources.)


zcwilt:
:+1:

ajeh:
:+1:

PR #1361, opened 2017-01-01
Tidy the SQL query in the NewestOrders widget

drbyte:
Remove some unnecessary aliases which make the query harder to read


zcwilt:
:+1:

ajeh:
:+1:

PR #1370, opened 2017-01-20
Forces visitor to https if full site is setup for https and http is e…

### LINKED ISSUE #1369: Contact us with full site SSL does not redirect to https if http entered
In [How do I enable SSL after I have installed Zen Cart?](https://www.zen-cart.com/content.php?56-how-do-i-enable-ssl-after-i-have-installed-zen-cart) it is stated that if a store owner wishes to set their entire store to SSL that ENABLE_SSL should be set to false; however, in such a condition a customer could enter a non-https address to reach the contact_us page and it would be first "provided" as http.  When ENABLE_SSL is set to true and such an address is entered, then the site redirects to https.

While this condition requires a user to make that entry, it would seem that the same treatment should be offered for the full site set as SSL as is performed for when SSL is enabled just for the necessary pages.

This type of topic has been discussed a few times as a "debate"; however, it was today that I realized this issue existed.  Most recently discussed is: https://www.zen-cart.com/showthread.php?221740-How-can-I-change-all-links-in-the-store-to-be-https-if-the-site-accessed-via-https

mc12345678:
…ntered.

Per https://www.zen-cart.com/content.php?56-how-do-i-enable-ssl-after-i-have-installed-zen-cart,
if one is to set the entire site to https, then it is recommended that
SSL_ENABLED be set to false, but if the uri entered begins with http, then
the redirect to force the page to load via SSL is not activated and the
customer could enter enter data to the webpage without using the SSL.

Fixes #1369

### DIFF
--- a/includes/modules/pages/contact_us/header_php.php
+++ b/includes/modules/pages/contact_us/header_php.php
@@ -91,7 +91,7 @@ if (isset($_GET['action']) && ($_GET['action'] == 'send')) {
 } // end action==send
 
 
-if (ENABLE_SSL == 'true' && $request_type != 'SSL') {
+if ($request_type != 'SSL' && (ENABLE_SSL == 'true' || substr(HTTP_SERVER, 0, 5) == 'https')) {
   zen_redirect(zen_href_link(FILENAME_CONTACT_US, '', 'SSL'));
 }
 


mc12345678:
To anyone reading this (assuming it will survive my next push, so read "quickly") and to introduce some non-techy discussion into the midst of the more computer science/programming part of this, there are two parts to the above recommendation that was provided:
1) There is a concern of order of operations related to using `&&` (comparison of two items where they both must be true in order for that portion to be considered true) and `||` (comparison of two items where if either is true then that portion is true).  Now the discussion offered related to the human understandable `'and'` and `'or'` which also work in a similar way, but PHP says that those comparison operators are lower level than other values: see the [PHP manual on operator precedence](http://php.net/manual/en/language.operators.precedence.php).  Use of those two operators (`'and'` and `'or'`) can introduce some very quirky results without applicable parentheses to group operations.  The suggested code, though strictly used `&&` and `||`.
2) There are multiple comparisons performed when the same/expected result can be obtained with less code and therefore less system "strain".

Discussion of item 1) will help with discussion of item 2).  PHP in an `&&` comparison will stop evaluating any further argument within the `&&`ing group when any value is found to be false.  PHP will also stop processing an `||` condition from left to right if any "value" is found to be true.  So, in the proposed code with 2 `&&`s on the left, they will be checked first.  If that is true, then the operations on the right will be ignored.  If either of the first two are found to be false, then processing will begin on the right and again if any of the 3 are found false, then the entire comparison is identified as false.  If all are true, then the comparison is considered true.

So item 2) well, here's where the "fun" is... In both comparison sides, there is at least one commonality: `$request_type != 'SSL'`. So that allows the code to be "shortened" to 
`$request_type != 'SSL' && (ENABLE_SSL == 'true' || ENABLE_SSL != 'true' && substr(HTTP_SERVER, 0, 5)== 'https')`
Okay, but then that discussion of how `||` operates... If the left side is false then the right side is processed... Well.. The way the right side was written was in one way to support continued operation even though in one of the "upgrade" paths ENABLE_SSL might become blank '' instead of specifically 'false'.  Instead of "digging" into how `== 'false'` might be expected to resolve the same as `== ''`, by it being `!= 'true'` it fit any other possibility.  The right side was also written to make it "clear" what comparison(s) were being done and in a way "why" they were being done. But, that is also not a "clean" method of coding/logic.

So, again, the way `||` operates, if the first operator is false, then by inference, the first test on the right side is already true and known to be true... Therefore there is no need to "test" if `ENABLE_SSL != 'true'` because that MUST be the case... :) Ergo the simplified code proposed by DrByte to take a potential five condition test down to 3 conditions which is an extension of one condition from the original code and provides the final code:
`$request_type != 'SSL' && (ENABLE_SSL == 'true' || substr(HTTP_SERVER, 0, 5)== 'https')`

In the end, the doctor saw something that "wasn't right" (which is true), but perhaps couldn't explain exactly why. :)

drbyte:
👍

zcwilt:
:+1:

PR #1374, opened 2017-01-24
Provide maximum product on shopping cart product update.

### LINKED ISSUE #1373: On cart update product quantity doesn't resolve to maximum
In correcting issue #1163 with #1164, if a quantity of a product was entered that exceeded a maximum (ie. $add_max, or increased above the total quantity), then the result for that/those unique products would return back to the value that was present before attempting to change it.

It was explained as desirable to allow the cart to resolve to the maximum quantity of each entered quantity when pushing an update of the cart quantity where the quantity exceeds said possible maximum.

mc12345678:
In #1163, an issue with updating the cart was identified.

The solution to that issue (#1164) prevented assigning the maximum product
to the cart on update `added:`(when trying to increase the in-cart quantity above the maximum allowed) `:`
and instead would result in the last acceptable
quantity remaining in the cart instead of going "out-of-bounds".

This corrects the issue(s) discussed in those two threads by minorly
rearranging the actionUpdateProduct code to perform the cart decrease actions `added:`with no revision of the decrease functionality (ie. no change in cart quantity resolution if add less than the minimum/unit combination)`:`
before any cart increase actions and specifically to support mixed product as
that is a situation where two or more cart products can/do depend on
each other.

Created a function that obtains details about the posted product data in order to support processing
that data systematically.

Added some "session" message blockers because this solution revises the sequence of how 
the cart functions, the prevent duplicate messages is targeted at one feature/operation and therefore had to add some flags to otherwise prevent duplicate messages being displayed.

The over cart process could be modified for "consistency" (ie. all actions to address decrease or low level insertions before increase), but it seems that it is only in the update condition that the 
changes made are possibly of true benefit.

Fixes #1373

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -70,6 +70,7 @@ class shoppingCart extends base {
    */
   var $display_debug_messages = FALSE;
   var $flag_duplicate_msgs_set = FALSE;
+  var $flag_duplicate_quantity_msgs_set = FALSE;
   /**
    * constructor method
    *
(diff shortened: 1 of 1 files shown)


ajeh:
So far, this appears to be working correctly.

mc12345678:
Yeah for that!

drbyte:
Okay 👍

ajeh:
:+1:

PR #1375, opened 2017-01-25
Server Info:  Add database section

lat9:
Update the Tools->Server Info to include database configuration

### DIFF
--- a/app/Controllers/admin/ServerInfo.php
+++ b/app/Controllers/admin/ServerInfo.php
@@ -27,6 +27,7 @@ class ServerInfo extends AbstractInfoController
         $this->buildPHPInfoSection();
         $this->buildSystemInfo();
         $this->buildVersionInfo();
+        $this->buildDatabaseInfoSection();
 
     }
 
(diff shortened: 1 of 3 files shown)


app/Controllers/admin/ServerInfo.php
```diff
@@ -116,4 +117,23 @@ private function buildVersionInfo()
...
+    
+    private function buildDatabaseInfoSection ()
+    {
+        $databaseInfo = [];
+        $system = zen_get_system_information();
+        $databaseInfo['heading'] = TITLE_DATABASE_VARIABLES . $system['db_version'] . ($system['mysql_strict_mode'] == true ? '<em> ' . TITLE_MYSQL_STRICT_MODE . '</em>' : '');
+        
+        $databaseInfo['fields'] = [];
+        $show_variables = $this->dbConn->Execute ("SHOW VARIABLES");
+        while (!$show_variables->EOF) {
```
drbyte: Since v1.5.5 we now support, and prefer, a `foreach` syntax when looping thru db queries, instead of the `while( !EOF) ... MoveNext...` syntax.

So this would look more like:

```php
        $results = $this->dbConn->Execute ("SHOW VARIABLES");
        foreach($results as $result) {
            $databaseInfo['fields'][] = array (
                'name' => $result['Variable_name'],
                'value' => zen_not_null ($result['Value']) ? $result['Value'] : '&nbsp;'
            );
        }
```

drbyte:
👍

lat9:
As an aside, when I was making that change I'd initially messed up the variable name that I was using to extract the information, using $show_variable**s** (the primary return) rather than the individual row ($show_variable, no **s**).  My result was a white-screen with no log generated.

I haven't (yet) had the chance to dig deeper to understand what's going on, but wanted to report that.

zcwilt:
:+1:

PR #1377, opened 2017-01-31
Refactor each() to foreach() for compatibility with PHP 7.2

drbyte:
An RFC for PHP 7.2 has been voted on for inclusion in PHP 7.2.  It proposes to deprecate the `each()` function for PHP 7.2 and removes it in 8.0.

The `each()` function is exponentially slower and wastes resources vs `foreach()`, so we're refactoring it now whether PHP 7.2 finally does adopt it or not.

Things to note when refactoring: 
1. `foreach()` doesn't need a `reset()` to be called before it runs, so those can be removed.
2. There are 3 syntax formats, depending on how the parameters are presented in the `list()` call:

a) `while(list($key, $value) = each($foo))`
This becomes `foreach($foo as $key => $value)`

b) `while(list(, $value) = each($foo))`
This becomes `foreach($foo as $value)`

c) `while(list($key, ) = each($foo))`
This becomes `foreach($foo as $key => $value)`

### DIFF
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -312,8 +312,7 @@
           }
         }
 
-        reset($products);
-        while (list($key, $value) = each($products)) {
+        foreach($products as $key => $value) {
           $category_ids = '';
 
           for ($i=0, $n=sizeof($value['categories']); $i<$n; $i++) {
@@ -336,8 +335,7 @@
           zen_remove_category($categories[$i]['id']);
         }
 
-        reset($products_delete);
-        while (list($key) = each($products_delete)) {
+        foreach($products_delete as $key) {
           zen_remove_product($key);
         }
       }
--- a/admin/coupon_admin.php
+++ b/admin/coupon_admin.php
@@ -782,8 +782,7 @@ function check_form(form_name) {
                 <td>
 <?php
 /* Re-Post all POST'ed variables */
-    reset($_POST);
-    while (list($key, $value) = each($_POST)) {
+    foreach($_POST as $key => $value) {
       if (!is_array($_POST[$key])) {
         echo zen_draw_hidden_field($key, htmlspecialchars(stripslashes($value), ENT_COMPAT, CHARSET, TRUE));
       }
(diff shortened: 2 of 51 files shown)


zcwilt:
:+1:

ajeh:
:+1:

drbyte:
Note: the various imported vendor packages we leverage will require updates; will cross that bridge when they push new versions.

dakanji:
@drbyte, How would you suggest tackling instances where there is no while loop involved?

Examples:
a) list($key, $value) = each($foo)

b) list(, $value) = each($foo)

c) list($key, ) = each($foo)

drbyte:
I'm not aware of any such use cases in actual Zen Cart code, old or new.

Reading http://php.net/manual/en/function.each.php and http://php.net/manual/en/function.list.php offer some insight, particularly the fact that `each` only handles numeric indices, not named indices. I suspect you'll end up just referring to `$foo[0]` and `$foo[1]` etc, or some variation of that.

dakanji:
Thanks. Came across these elsewhere and was looking for correct solutions when I found your post here.

For the record,

```
a) list($key, $value) = each($foo) is equivalent to:
$key = array_keys($foo)[0];
$value = array_values($foo)[0];


b) list(, $value) = each($foo) is equivalent to:
$value = array_values($foo)[0];

c) list($key, ) = each($foo) is equivalent to:
$key = array_keys($foo)[0];
```

... and list($key) = each($foo) is equivalent to Example C

drbyte:
Agreed. Thanks for posting these. 

It's unfortunate there's no shorter and equally expressive version as `each()` when used outside a loop.

torvista:
I'm using php7.2 for development, PHPstorm marks "each" as deprecated.
My production site/hosting is already on 7.1.7.
Given the widespread use of "each" in 1.55 (>200 instances), and 7.2 due in November, what is the plan?

drbyte:
We were discussing that earlier this week. We'll be tagging a new release.

PR #1380, opened 2017-02-02
Tax rate and description reversed for cases where there is no active tax rate

robryanx:
I don't actively use these tax groups but came to my attention due to "A non-numeric value encountered" notice in PHP 7.1
Effective result should be the same as the text would have evaluated to 0


drbyte:
👍  Thanks!

ajeh:
:+1:

PR #1381, opened 2017-02-02
Add __destruct method to query-factory

### LINKED ISSUE #1344: Should the queryFactory class include a destructor?
As more and more sites upgrade to a version of Zen Cart that uses the mysqli interfaces, I'm seeing a significant number of sites that have logs similar to the following

> PHP Warning:  mysqli_connect(): (HY000/1203): User xxx already has more than 'max_user_connections' active connections in /file_path/includes/classes/db/mysql/query_factory.php on line 64

and I'm wondering if perhaps the `mysqli` interface is more picky than the `mysql` interface regarding closing down the database connection once a page's rendering is complete.

These logs are generated across Zen Cart versions 1.5.2 through 1.5.5b and on various hosted platforms, which led me to the common  point of the missing class destructor.  That could also be exacerbated by the issue that was present in the AJAX handler's missing load of `/includes/application_bottom.php`.

drbyte:
Closes #1344

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -132,6 +132,10 @@ class queryFactory extends base {
     unset($this->link);
   }
 
+  function __destruct() {
+    $this->close();
+  }
+
   function set_error($zp_err_num, $zp_err_text, $dieOnErrors = true) {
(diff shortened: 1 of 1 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #1385, opened 2017-02-03
Support table prefixes for 1.3.8 upgrade in zc_installer

### LINKED ISSUE #1386: v155 zc_install does not fully support table prefixed upgrade of 1.3.8
The upgrade script from 1.3.8 to 1.3.9 includes a query that is not fully supported by the zc_install process as rewritten for ZC 1.5.5.

This issue is further discussed in the ZC forum at [Error: Database upgrade from 1.3.8a](https://www.zen-cart.com/showthread.php?221788-Error-Database-upgrade-from-1-3-8a).

mc12345678:
Prior to ZC 1.5.5, the zc_installer included code to process the upgrade
script from ZC 1.3.8 to 1.3.9; however, in the redevelopment for ZC 1.5.5,
operations did not include processing SQL commands such as only DROP TABLE
though DROP TABLE IF EXISTS was supported.  This code was replaced and
expanded to support DROP TABLE and DROP TABLE IF EXISTS.

Also the code did not support the SQL command of RENAME TABLE.

To support the revised code the ZC 1.3.9 sql (upgrades 1.3.8 to 1.3.9)
required minor modification as well.

No other related/earlier SQL was identified as needing this change.
Fixes #1386

### DIFF
--- a/zc_install/includes/classes/class.zcDatabaseInstaller.php
+++ b/zc_install/includes/classes/class.zcDatabaseInstaller.php
@@ -37,7 +37,7 @@ class zcDatabaseInstaller
     $this->dieOnErrors = isset($options['dieOnErrors']) ? (bool)$options['dieOnErrors'] : FALSE;
     $this->errors = array();
     $this->basicParseStrings = array(
-    'DROP TABLE IF EXISTS ',
+    'DROP TABLE ',
     'CREATE TABLE ',
     'REPLACE INTO ',
     'INSERT INTO ',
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks @mc12345678 !

drbyte:
👍

mc12345678:
> Thanks @mc12345678 !
More than welcome.  

I feel that I should have caught that in testing because I did experience a similar database issue that had to be worked around by first removing the table prefix.  At the time I didn't understand the installer as well as I do now, so I didn't realize what it would have taken to correct.

zcwilt:
:+1:

PR #1391, opened 2017-02-04
Remove authnet test from v139 upgrade step

drbyte:
... because this field has changed and requiring it to be of the old schema may throw false errors during upgrades


zcwilt:
:+1:

ajeh:
:+1:

PR #1401, opened 2017-02-25
Space before parenthesis multi

mc12345678:
Add right parenthesis separation
    
    In the upgrade sql for from version 1.3.8 to 1.3.9 as provided in ZC 1.5.5, there are a few sql queries that include an internal sql query.  When evaluating a line that reads `FROM this_table);`.  The existing code will remove the ending semicolon; however, then it considers the right parenthesis as part of the table name.  It appears that there are three potential solutions:
    1) place the parenthesis on the next line.
    2) add a space into the sql query file between the table name and the following characters within the sql query itself.
    3) add a space into the sql query as read from the file.
    
    This commit addresses internally adding one space at the end of the sql
    query prior to each parenthesis that ends the string as it was read from
    the file.  It does not consider that there might be multiple spaces before
    the parenthesis, just that there is a non-right parenthetical value to the
    left of the last character before adding back the parentheses with a
    preceding space.
    
    There is still a limitation on this solution, but it is a solution to an
    observed issue.  The concept can be carried forward to insert more spaces,
    or to effectively find the first character that needs the first space in it.

### DIFF
--- a/zc_install/includes/classes/class.zcDatabaseInstaller.php
+++ b/zc_install/includes/classes/class.zcDatabaseInstaller.php
@@ -103,7 +103,16 @@ class zcDatabaseInstaller
       if ( substr($this->line,-1) ==  ';')
       {
         if (substr($this->newLine,-1)==' ') $this->newLine = substr($this->newLine,0,(strlen($this->newLine)-1));
-        if (substr($this->newLine,-1)==')') $this->newLine = substr($this->newLine,0,(strlen($this->newLine)-1)) . ' )';
+        $numRight = 0;
+        while (substr($this->newLine,-1)==')')
+        {
+          $numRight++;
+          $this->newLine = substr($this->newLine,0,(strlen($this->newLine)-1));
+        }
+        for ($i = 0; $i < $numRight; $i++)
+        {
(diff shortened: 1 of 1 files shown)


drbyte:
I'm confused. 

Your descriptions for #1400 and #1401 a repetitive, and code commit descriptions seem similar.

Is there duplication here?

Is there intentional repetition?

Is one supposed to have been abandoned?

mc12345678:
Offering two options for processing. Either fix a single instance of a parenthesis being used, or the possibility that there could be nested parentheses.

I thought about presenting two completely independent PRs, then didn't want to cause a problem/conflict, so instead I pushed two where basically one is dependent on the other. 

I felt like patching the one parentheses was a band-aid, rather than a fix. I also didn't want to spend too much time looking into all possible SQL format possibilities and make the simple but effective code complicated.

Yes, they are related to each other, one should ultimately go away or be merged with the other. I'm not averse to either/neither being used as is. They relate more to development/upgrade than operation so not sure how involved the queries are planned/expected to be or if they have in some way been simplified because of the interpreter instead of what is truly needed. Etc...

drbyte:
Closed by merging #1400

PR #1403, opened 2017-03-03
Restore price manager entry of specials by percent

mc12345678:
In issue #750, the price manager data entry was modified to support display
of the specials price for both gross and net costs instead of just for
one or the other.  Unfortunately in that proposal, the ability to enter
a value with a percentage was not maintained.  Today when trying to test
other operations, I realized my oversight and worked to correct it.

This commit will disregard tax differences in data entry if a percent is
entered and apply the same percentage to both fields, otherwise if the
data entered is a number, then the appropriate "conversion" as applicable
will be performed while typing.


drbyte:
👍

zcwilt:
:+1:

PR #1404, opened 2017-03-03
Special and SaleMaker price-by-attributes sales correction

mc12345678:
When adding a product to the cart that has been priced by attributes, with
a price factor applied to the attribute and a special applied to the
product, the shopping_cart page reflects all of the correct prices; however,
once the customer moves forward to checkout, the price does not calculate
correctly.

This is heavily discussed and explained at this ZC post:
https://www.zen-cart.com/showthread.php?221940-attributes-price-factor-special-price-incorrect-totals&p=1326843#post1326843

There are other issues related to the application of the price factor and/or specials/sales,
but those are to be addressed separately.


mc12345678:
Will update this PR with the code to correct this globally in ZC. Current PR only addresses a small fix. This comment will I think get deleted upon my reset of the PR.

mc12345678:
I'd like to point out. In the first attempt to resolve the issue of prices not showing correctly for a product priced by attributes, a price factor applied and a special applied to the product, the solution suggested involved "correcting" the math that had been done upstream and doing so at the point of "need". 

This was done for at least two reasons:
1) minimize changes to other areas of code.
2) address the issue as seen at the point where the problem existed/was seen.

In trying to identify just why there was a discrepancy, it was discovered that some "earlier" values were not adjusted as would be needed to support the desired/expected outcome. Then of course once that area was addressed further testing/adjustments identified other areas to correct. 

All fine and well to an extent, but now the changes have resulted in functions acting differently than before and a change to the structure. My thing is that it seems/looked like there had been a desire to keep these additional calculations outside of the functions in which I propose. 

I'm not sure if that remains the case or if the fix is something desirable. If anything it seems like the attributes_controller has a column designed/expecting to show how much of a discount an attribute is providing. While that "feature" wasn't really working under a basic ZC install, it did seem to highlight when the proposed fix might be necessary. At any rate, there doesn't appear to be a way to calculate such a discount amount as a result of an attribute's discount, and further, I believe the current proposal assumes that the desire is to discount attributes with sales and specials regardless of the constant that exists in the cart section of the shopping_cart class file...

Guess looking to see from the maintainers of if they 1) recommend an alternative method to solve the problem, 2) recommend developing such a function to identify the discount associated with an attribute to display such value in the attributes_controller and 3) if more should be done/offered regarding enabling/disabling aspects of discounts/sales being applied to attributes.

mc12345678:
To add a little more discussion.  I realize that the movement of the zen_get_discount_calc code and checks for whether the product is priced by attribute into the zen_get_attributes_price_final function involve additional queries at each call. While that can be factored down by providing those values up front as additional parameters to the function, I'm somewhat trying to get an idea of the suggested way forward in the grander scheme of things.

I.e. should a completely new function(s) be created to address this issue, should additional parameters be added which if left unassigned would result in the function providing the same as "always" result/operation, etc...

One thing that can be seen in this rewrite is that it looks like there is an automatic assumption that the attribute prices follow the discounting of the product although the applicable functions seem to determine the need for discount based on price by attribute in the shopping cart class there is the "test" against a constant (ATTRIBUTES_PRICE_FACTOR_FROM_SPECIAL) of whether the price factor for the attribute "discount" should be against the normal price or the special price.

includes/classes/shopping_cart.php
```diff
@@ -786,7 +786,8 @@ function calculate() {
                 }
               if ($attribute_price->fields['attributes_discounted'] == '1') {
                 // calculate proper discount for attributes
-                $new_attributes_price = zen_get_discount_calc($product->fields['products_id'], $attribute_price->fields['products_attributes_id'], $attribute_price->fields['options_values_price'], $qty);
+                $new_attributes_price = zen_get_discount_calc($product->fields['products_id'], $attribute_price->fields['products_attributes_id'], /*$attribute_price->fields['options_values_price']*/ $attribute_price->fields["options_values_price"] + (zen_get_products_price_is_priced_by_attributes($attribute_price->fields["products_id"]) ? zen_products_lookup($attribute_price->fields["products_id"], 'products_price') : 0), $qty);
```
drbyte: I "get" that you've commented-out the "old" and then added new, but to comment-out the exact same variable and then add it back in again is redundantly redundant.

```diff
-$attribute_price->fields["options_values_price"]
+/*$attribute_price->fields['options_values_price']*/ $attribute_price->fields["options_values_price"] + foo...
```

Same on line 1019

drbyte:
I'm torn about leaving in some of the other commented-out-"old" lines that you're replacing with completely different code.

Also torn whether this should go into v155

includes/classes/shopping_cart.php
mc12345678: I believe at the moment that this upgrade affecting commenting is more about the development changes rather than the final product by which others would use to finally get the expected prices displayed/processed when working with attributes (minus perhaps unique tax classes on an attribute). I know that for me it has provided a sort of marker that the code has been touched in that area and what it originally was. For such inline replacement I would think the redundant commented out code should be removed for final producr. 

With the breadth of impact overall (additional processing when displaying values, additional queries, etc.) that some amount of supporting review/operations need to be confirmed. I've tried to methodically go through variations, but I wouldn't doubt that I have missed some expectancy.

mc12345678:
So there are no database modifications to the code changes if that helps any.  

There does remain a few files to be similarly be touched in the same way as presented. If the proposed operational process is considered sound, then I will update/include those as well. (ie. Admin side to see a sample of the results comes to mind).

drbyte:
It's a delicate juggle between touching files that are often changed by plugins, vs giving everyone (who upgrades) the improvements "now" vs "later".

And also considering db changes as you mentioned.

mc12345678:
> It's a delicate juggle...

Agree.

Personally, I didn't realize that these issues existed and it was only after seeing the call to inquiry on two separate topics and looking into it that spurred action. I don't know if others have independently (commercially maybe) handled this or ignored it, but seeing some of the more core posters asking about it seems to be a good thing in a way, that more are using more features and seeking assistance rather than ditching out.

mc12345678:
BTW, I do not yet suggest that this be incorporated into core code quite yet.  I haven't chased it down, but I have seen an occurrence related to a specials price either being higher than a base price or a sale price being higher than a specials price or some similar situation where the condition was contrary to what a reasonable individual (customer) would expect to see, but it causes some odd response.  Yes understand that one should reasonably enter data so that either each additional "feature" is a lower cost and not make it look like hey have this great deal for you that costs more than the base price!  Some of that is/may be a need for appropriate validation at each location of change, but I think what I was seeing is more related to program flow and how it should respond versus how it did respond.

I'm planning on creating some sort of test "plan" to try all of the options that appear to be possible and methodically work through them. Would be nice if when that is posted, the  work can be divided and conquered.

mc12345678:
Been pondering the condition of a product that is on sale with a "constant" discount such as x amount off. There are notes in the code about how this shouldn't be done with attributes; however, there is nothing that either prevents it or notifies the store admin about placing a constant price reduction against a product that has attributes.

That aside, I've identified how a product with a single attribute that affects price (in particular one priced by attributes); however, if a product has more than one price affecting attribute (something that has to be "discovered") then the total reduction (assuming the option value(s) in each option name is affected by "specials") is not to exceed the total discount and as well, the deduction regardless of attribute choice "should" be what has been entered which may not be possible if the price of an attribute is less than a "balanced" portion of the total discount... In other words: the complexity of getting such a discount right goes up significantly without a predetermined knowledge of what one would expect based on product information.

Currently the code determines some "semblance" of a reduction and it is possible to apply the discount to just the product's price (for non-priced-by-attribute product) assuming that there is "excess" after the reduction, otherwise run into the same "determination" issue as described above. 

Further what this means is that I will be focussing on the remaining sale/special combinations and will just have to leave the constant reduction style correction to a later date if at all...

drbyte:
@mc12345678 Where does this PR stand?

mc12345678:
Not as complete as desired. There are some aspects that have been corrected/addressed, others that if not changed from their current state would provide the same discrepancies as they have for some time. So I could update the areas discussed above about "replacement" formatting and what has been improved upon could be incorporated with the likelihood of knowing that more could/should be done for it to operate properly when dealing with multiple discounts and prices on a displayed product.

drbyte:
I think I understand you to be saying that this is kinda like "phase 1", and after merging then "phase 2" can be done later. 

In that light, yes, if you wanna clean it up for merge, that'd be great. 

Mind if I re-point it at 1.5.6 instead of 1.5.5?

mc12345678:
> Mind if I re-point it at 1.5.6 instead of 1.5.5?

No, do not mind.
> I think I understand you to be saying that this is kinda like "phase 1", and after merging then "phase 2" can be done later.

Basic operational testing with the existing changes have eluded to this possibility. 

I believe some of the changes proposed in ZC 1.6.0 regarding the price display determination will provide further needed improvements, though at last review a few months ago one item found during this modification but not yet addressed, remained not addressed in the 1.6.0 path.  

Basically, its the "simple" state of how a discount from a sale or special is determined or against what the discount is calculated (ie. if there is a sale and a special on a product, should the sale's "percent off" be the percent off of the special or the percent off of the base product price? At last review, it was displayed as "always" off of the base product price.).  

Text in choosing the sale "criteria" eludes to how the discount should be calculated, but it does not appear that associated code is incorporated and in some ways with all of the various available options, it can not be so "easily" satisfied.  (I.e. a product with more than one attribute each affecting the base price of a product where the product is on sale by a straight price reduction. How is that price reduction to be applied "uniformly" on initial base display when each such attribute could have its own price and such that selection of the "cheapest" attributes of a product within an on-sale master category could result in the product being undesirably free...) 

Anyways it is some of this later part that can be addressed in a subsequent modification.

drbyte:
Okay, repointed at v156. 
I'm pretty sure I resolved the merge conflicts correctly.

mc12345678:
> I'm pretty sure I resolved the merge conflicts correctly.

Knock on wood, will find out shortly. :)

mc12345678:
Made it through the code and current status.  Updated comments and modified some of the changed code structure to match previous recommendations/requests. ie:
````php
$variable = $value;
if (condition == true) {
  $variable = $new_value;
}
````
instead of using an if/else.

Other changes made to clarify comments and maybe make upcoming modifications easier. Changes added a single commit to the changes last made.  Some of this did take away "notes" of some potential progress; however, some of those were possibly misleading.

PR #1411, opened 2017-03-19
Add semicolon to includes/classes/payment.php javascript

mc12345678:
Line is missing a closing semi-colon.

### DIFF
--- a/includes/classes/payment.php
+++ b/includes/classes/payment.php
@@ -155,7 +155,7 @@ class payment extends base {
        $js =  $js .'    alert(error_message);' . "\n";
        $js =  $js . '    return false;' . "\n";
        $js =  $js .'  } else {' . "\n";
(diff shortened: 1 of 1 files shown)


drbyte:
Is there a specific error this is causing?

JavaScript is different in that it often treats an end-of-line the same as PHP would treat a semicolon, so isn't always strictly required. 

But, if there's a problem happening without the semicolon I'd be interested in the details.

mc12345678:
Will take a little doing to validate is/was a problem causing an issue. Was found when working on the PayPal incontext improvement. Is already incorporated in 1.6.0, but will need to see if it is an issue with the same applied in 1.5.5 as is (missing the end semi-colon).

drbyte:
👍

zcwilt:
:+1:

PR #1412, opened 2017-03-19
Add function to identify the presence of a date span overlap

mc12345678:
zen_datetime_overlap offers the ability to identify the overlap relationship
between two provided date ranges with consideration of the frame of
reference by which to make the comparison.  Ie. should dates be compared
with concern of their overlap in the future, in the past only or an
overlap existing at all.

The function was originally considered to support the addition of specials
to a product such that it would not be possible from the admin entry point
to insert a new special that does not overlap with an existing (active)
special.  At least initially until it is determined/decided that specials
can be applied to specials for example.


drbyte:
Can you please pull this against v160?

Pulls to v155 are for critical bugs only.

mc12345678:
Will do.

mc12345678:
Closing because in moving this to ZC 1.6.0, the original base did not get updated as well and has generated errors unrelated to the action of the pull request.

PR #1413, opened 2017-03-20
Add function to identify the presence of a date span overlap

mc12345678:
zen_datetime_overlap offers the ability to identify the overlap relationship
between two provided date ranges with consideration of the frame of
reference by which to make the comparison.  Ie. should dates be compared
with concern of their overlap in the future, in the past only or an
overlap existing at all.

The function was originally considered to support the addition of specials
to a product such that it would not be possible from the admin entry point
to insert a new special that does not overlap with an existing (active)
special.  At least initially until it is determined/decided that specials
can be applied to specials for example.

This is/was also discussed in the ZC forum at: https://www.zen-cart.com/showthread.php?222038-Can-there-be-more-than-one-quot-Special-quot-for-a-product


lat9:
A note:

When you use `trigger_error` to log the issues, that processing already includes the filename and line number from which the error was reported, so using __FILE__ and __LINE__ is redundant.

mc12345678:
> When you use trigger_error to log the issues, that processing already includes the filename and line number from which the error was reported, so using FILE and LINE is redundant.

Had initially considered alternate methods of notification, which required identifying the additional information (if wanting to ease trouble shooting), finally decided on trigger_error and hadn't really considered the need for modifying that.  Double information in the end, but has been updated to remove that confusing duplication.

zcwilt:
Still reviewing this in light of other work that is happening  in 1.6
I had planned in converting the specials page to the new listing/lead code.

mc12345678:
Can respect that. Would think that the function would still play a part in that new process where perhaps additional information would/could be captured by use of the listing/lead code, though I still need to get caught up in the new changes, processes, and significant internal flow changes.

zcwilt:
yes, whether it's in a function or helper class, it should be part of the updated code.

zcwilt:
As such would like to leave this PR live, just as a reminder :)

mc12345678:
> As such would like to leave this PR live, just as a reminder :)

Understood that the comment is more of a "public" note, but also that I believe it means for me not to modify it away, but allow it to persist until resolved. Regardless, got it, I think. :)

drbyte:
👍

mc12345678:
Consideration of incorporating into ZC 1.5.6?

drbyte:
Okay, submit a PR for it.

PR #1414, opened 2017-03-23
Add cache-bypass forms of Execute method

lat9:
There are many times when a SELECT query needs to pull from the database, bypassing the cache.  The current interface is pretty clumsy:

`$db->Execute ("SELECT ...", false, false, 0, true);`

This update adds cache-bypass forms of the Execute and ExecuteRandomMulti methods.


zcwilt:
:+1:

drbyte:
👍

PR #1417, opened 2017-04-04
Incorporate sanitizer code related to manufacturers.

mc12345678:
The image file is not sanitized similar to the products image as well the
manufacturers_url will convert an ampersand `&` to ampersand `&amp;`.  With this
sanitization, an ampersand (for example in a URI) will not be rewritten.

This was discovered while helping out with a manufacturer plugin and is
discussed in the ZC forum at:
https://www.zen-cart.com/showthread.php?222079-Cannot-modify-header-information-headers-already-sent&p=1328124#post1328124

Although, there a different/additional sanitization was suggested for manufacturers_name to be equivalent to products_name.

### DIFF
--- a/admin/includes/init_includes/init_sanitize.php
+++ b/admin/includes/init_includes/init_sanitize.php
@@ -177,7 +177,7 @@ $group = array(
 );
 $sanitizer->addSimpleSanitization('CONVERT_INT', $group);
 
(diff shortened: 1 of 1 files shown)


drbyte:
👍

zcwilt:
:+1:

Issue #213, opened 2014-10-21
Validate STOCK_CHECK clause in shopping_cart page's header_php.php file


drbyte:
lat9 post from http://www.zen-cart.com/showthread.php?214533-Validate-STOCK_CHECK-clause-in-shopping_cart
The test of the returned result doesn't match the return signature.

PR #1423, opened 2017-04-14
Reduce db hits during product-stock checking

### LINKED ISSUE #1346: Repetitive Queries: zen_get_products_stock()
Mainly just curious if this is a current design choice or just a leftover from the past. I notice this for a lot of things but in this specific case for getting product stock in the cart.

```php
if (STOCK_CHECK == 'true') {
    $qtyAvailable = zen_get_products_stock($products[$i]['id']);
    // compare against product inventory, and against mixed=YES
    if ($qtyAvailable - $products[$i]['quantity'] < 0 || $qtyAvailable - $_SESSION['cart']->in_cart_mixed($products[$i]['id']) < 0) {
      $flagStockCheck = '<span class="markProductOutOfStock">' . STOCK_MARK_PRODUCT_OUT_OF_STOCK . '</span>';
      $flagAnyOutOfStock = true;
    }
  }
```

This will run a query for each product in `zen_get_products_stock` to get the current stock count. What could be done though is using the cart function `function get_products` which is already querying the products table and returning the quantity in `$products_array[]`

proseLA:
Closes #1346 

it also seems the `$products_id` would no longer be needed in the `zen_check_stock` function.  although i did leave it there.

in exploring this situation, i also came across times when ZC will automatically adjust the quantity in cart to the amount on hand.  and other times when it would alert the user that the current quantity is too much without giving any additional info about the quantity.  it seems like ZC should automatically adjust in both cases.

### DIFF
--- a/includes/classes/ajax/zcAjaxPayment.php
+++ b/includes/classes/ajax/zcAjaxPayment.php
@@ -120,7 +120,7 @@ class zcAjaxPayment extends base
     $stock_check = array();
     if (STOCK_CHECK=='true') {
       for($i = 0, $n = sizeof ($order->products); $i<$n; $i++) {
-        if ($stock_check[$i] = zen_check_stock ($order->products[$i]['id'], $order->products[$i]['qty'])) {
+        if ($stock_check[$i] = zen_check_stock ($order->products[$i]['id'], $order->products[$i]['qty'], $order->products[$i]['quantity_on_hand'])) {
           $flagAnyOutOfStock = true;
         }
       }
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -494,6 +494,7 @@ class order extends base {
       $this->products[$index] = array('qty' => $products[$i]['quantity'],
                                       'name' => $products[$i]['name'],
                                       'model' => $products[$i]['model'],
+                                      'quantity_on_hand' => $products[$i]['products_quantity'],
                                       'tax_groups'=>$taxRates,
                                       'tax_description' => zen_get_tax_description($products[$i]['tax_class_id'], $taxCountryId, $taxZoneId),
                                       'price' => $products[$i]['price'],
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -1182,7 +1182,7 @@ class shoppingCart extends base {
       $products_query = "select p.products_id, p.master_categories_id, p.products_status, pd.products_name, p.products_model, p.products_image,
                                   p.products_price, p.products_weight, p.products_tax_class_id,
                                   p.products_quantity_order_min, p.products_quantity_order_units, p.products_quantity_order_max,
-                                  p.product_is_free, p.products_priced_by_attribute,
+                                  p.product_is_free, p.products_priced_by_attribute, p.products_quantity,
                                   p.products_discount_type, p.products_discount_type_from, p.products_virtual, p.product_is_always_free_shipping,
                                   p.products_quantity_mixed, p.products_mixed_discount_quantity
                            from " . TABLE_PRODUCTS . " p, " . TABLE_PRODUCTS_DESCRIPTION . " pd
@@ -1342,6 +1342,7 @@ class shoppingCart extends base {
                                   'image' => $products->fields['products_image'],
                                   'price' => ($products->fields['product_is_free'] =='1' ? 0 : $products_price),
         //                                    'quantity' => $this->contents[$products_id]['qty'],
+                                  'quantity_on_hand' => $products->fields['products_quantity'],
                                   'quantity' => $new_qty,
                                   'weight' => $products->fields['products_weight'] + $this->attributes_weight($products_id),
                                   // fix here
--- a/includes/functions/functions_lookups.php
+++ b/includes/functions/functions_lookups.php
@@ -259,8 +259,8 @@ function zen_get_zone_class_title($zone_class_id) {
  * @param int $products_quantity  Quantity to compare against
  * @return string status message
 */
-  function zen_check_stock($products_id, $products_quantity) {
-    $stock_left = zen_get_products_stock($products_id) - $products_quantity;
+  function zen_check_stock($products_id, $products_quantity, $products_quantity_on_hand) {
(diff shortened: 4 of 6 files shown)


drbyte:
In addition to $products_id no longer being used, I'm struck that the returned string from zen_check_stock is never used.

It leaves me torn about whether to even call an external function to do a simple "if A < B" check. Well, if the function were a method on a related class, and not receiving unused variables, and were returning just a true/false, it'd make good sense.

proseLA:
i agree.... i think....  i am all for making it simpler.  i am also not sure this function is the place for the css styling.  i suppose a question remains if external plugins were making use of this function, how much will changing it break their plugins?

drbyte:
IMO the css is irrelevant. Outputting a string of information would be a higher concern than whether that string contains any CSS.

It's already been refactored somewhat to remove some unnecessary duplication and wasted/confusing output -- see #213 and forum URL for context.

And, while in this case I'm hardly worried about plugins, I think a new function with a specific name would be more helpful than refactoring `zen_check_stock`.

I'm inclined to suggest we create something like `zen_has_stock_on_hand()`, and have it simply return true/false by taking the first part of the `zen_check_stock` function logic.  Then have `zen_check_stock` call `zen_has_stock_on_hand` to determine whether to display the HTML string or not.

Then anywhere we just want to check whether we have stock on hand or not, we call `zen_has_stock_on_hand`, and where we want the HTML message we call `zen_check_stock`

Thoughts?

proseLA:
drB, i am still thinking about this...  i wanted to help with this PR as i felt we could reduce server load (and improve site speed) by avoiding unnecessary hits on the DB.  but now we are talking about bigger changes and more philosophical changes which for me requires some thought....

in thinking about it from a cursory standpoint, adding yet another function for checking inventory makes little sense to me....  there are already 2 and the answer to the question as to whether we need them both is not so clear to me.  so i am not thrilled about the idea of adding a 3rd...

as i said in one of my earlier posts, at times ZC corrects the cart quantity and other times it does not....  so my primary question is what are we trying to do?  and if the answer is to only allow a customer to check out with available inventory, perhaps we can change the shopping cart class and the order class to include the logic.  perhaps that makes it more complicated (not my goal), but at least everything is in one place....  and relatively easy to find...

as the design considerations are a bit bigger here, i just need to give myself a little more time to think about what is the best way to proceed...

mc12345678:
> in exploring this situation, i also came across times when ZC will automatically adjust the quantity in cart to the amount on hand. and other times when it would alert the user that the current quantity is too much without giving any additional info about the quantity. it seems like ZC should automatically adjust in both cases.

Having proposed some of the recent changes for quantity modification, I did see some of this issue as well, but the bigger question was what was the "right" way to address/modify the quantity and also, *when* should the quantity be adjusted? And in relation to issue #1346 and this current code change, the two do not appear to be related.  

@proseLA, I don't know where you are at with your contemplation, but yes having an additional function does add another function; however, the concern of having a third should be weighed against the 2 existing functions.  One which is a sort of "user" function to specifically retrieve the quantity of stock "normally" tracked by ZC, the second, to output html that is basically unused if the quantity requested exceeds the quantity available and that the presence of that output is, can be, and has been previously used to indicate a "true/false" condition.  

In a way (somewhat in a vacuum), the existing zen_check_stock function could go away and be replaced with the newly proposed `zen_has_stock_on_hand` with the output string placed into the file(s) where it is desired to be used which could also ultimately be a language file, or the code completely modified to not return the string and the string incorporated into the "one" function in which it is used; however, that seems like at this point it could be unnecessarily harmful to other software out there (ie. maybe in time it would go away, maybe not). 

I'm not sure though I follow:
> perhaps we can change the shopping cart class and the order class to include the logic. perhaps that makes it more complicated (not my goal), but at least everything is in one place.... and relatively easy to find...

The proposed code, while "dependent" on one another (up until a purchase is made, the order class depends on the shopping_cart class with the shopping_cart class pulling details together about the product.  The order class pulls the person data as well as the product data together) works towards an improved interface with the database, actually expands some functionality possibilities (that currently have to be rewritten or replaced) and focuses the code to be more object/data oriented.  The shopping_cart class obtains the available quantity from the database whenever it is asked for the products assigned to the cart, the order makes sure that there is a shopping cart and also gets the product so that the shopping_cart associated with that order is up-to-date with the available quantity, and there is a sort of centralized method to evaluate the quantities without any additional code. Anyways, regarding the above statement I'm not sure if you are/were thinking that each the shopping_cart and order classes should have these same functions internally (one place), that actually seems to add more of an issue than to have centralized functionality that is called upon by whatever object needs it.  

From another perspective, it could seem comical to actually have a function that subtracts two numbers and identifies if one is larger than the other, but it is being suggested to "provide" a common tool that could be used in many places and would remain up-to-date through future versions. Further, the current zen_check_stock function would actually have an additional parameter that isn't used, so a new function actually also seems like the right thing to do especially if all it is going to do is take `b - a` and evaluate to `true if a < b` or `false if a > b`.

The difference in cart response based on adding of product is a separate issue.  Perhaps it could use improvement/modification some of which may further reduce the repetitive queries, but it appears that the suggested further modifications would offer a gain of improvement directly related to the original issue and it is great to see that such a change is being furthered.

proseLA:
sorry for the delayed response.

i am inclined to just delete this PR.  i do not think functions are the place for the html.  and as drB pointed out in his initial comment, there is only 1 place where that is used in:

/tpl_ajax_checkout_confirmation_default.php#L134

and template files are where the html styling should be done, IMO.

trying to wrap my head around 3 different functions for a simple A > B inventory check makes little sense to me.

using a different logic during the checkout procedure than when adding to the cart also makes little sense to me.  i have looked at the logic, and if one adds something to the cart or updates the cart with a new quantity, then ZC will adjust to the max allowed, ie qty on hand, in this case.  if during the checkout procedure, someone else purchases that inventory ahead of the current customer, then ZC sets the $flagAnyOutOfStock to on, and it is up to the customer to determine how many he can add.  

i think the logic should be the same here.  

while i am all for minimizing database hits, i am not for unnecessary complexities.  which i think is where we are heading here.

drbyte:
Agreed.

PR #1424, opened 2017-04-24
Attribute option sort ordering

lat9:
Adds the secondary sort by option name when sorting also by option sort-order.

See [this ](https://www.zen-cart.com/showthread.php?222184-Product-attribute-sorting)Zen Cart posting for additional information.


ajeh:
:+1:

drbyte:
👍

PR #1426, opened 2017-05-02
Sync parameters for setting/testing cookies

drbyte:
When testing to set Force Cookie Use capability, we're not passing all the parameters that are also used by the logic which sets the session cookie. This syncs them to be the same.


zcwilt:
:+1:

ajeh:
:+1:

PR #1429, opened 2017-05-05
Patch language path folder for PHPMailer

drbyte:
After relocating PHPMailer files to the "vendor" folder, the language files are no longer found in the "classes/support" directory. This code change lets them be loaded from the default location instead of triggering a file-not-found condition.


zcwilt:
:+1:

ajeh:
:+1:

PR #1433, opened 2017-05-07
Refactor to use new request class vars instead of direct $_GET vars

drbyte:
Most use of $_GET should be accessed via zcRequest ... or use other vars already sanitized by the Request class or other filters.


zcwilt:
:+1:

ajeh:
:+1:

PR #1437, opened 2017-05-30
Fix bug: when not logged in, tax zone defaults gave string instead of zero

drbyte:
Resolves: `PHP Warning:  A non-numeric value encountered in /includes/functions/functions_taxes.php on line 172`


zcwilt:
:+1:

ajeh:
:+1:

PR #1442, opened 2017-06-06
Add minor-version match specificity flag to plugin version checking

drbyte:
Now, by default, it will ignore any non-numeric version characters (ie: v155b will match v155e, because letters are stripped before comparing)


zcwilt:
:+1:

ajeh:
:+1:

PR #1449, opened 2017-07-03
Assign default language in an odd/rare situation.

mc12345678:
Addresses an odd and rare situation that could maybe exist.
Basically, if the language selection is set to the browser, the browser
returns a language that is not installed, and if the store no longer has
a language assigned to a language['id'] of 1 (as could be seen if
installed a new language and then removed the base language).

Other than a specific/direct database edit that could cause an issue such
as a default no longer being defined or to disrupt the language table
assignment of an appropriate language_id to the default language, then this
should ensure that the language['id'] would be set if the browser does not
come back with a favorable result.

### DIFF
--- a/includes/init_includes/init_languages.php
+++ b/includes/init_includes/init_languages.php
@@ -20,6 +20,7 @@ if (!isset($_SESSION['language']) || isset($_GET['language'])) {
   } else {
     if (LANGUAGE_DEFAULT_SELECTOR=='Browser') {
       $lng->get_browser_language();
+      if (!zen_not_null($lng->language['id'])) $lng->set_language(DEFAULT_LANGUAGE);
     } else {
       $lng->set_language(DEFAULT_LANGUAGE);
     }


includes/init_includes/init_languages.php
```diff
@@ -20,6 +20,7 @@
   } else {
     if (LANGUAGE_DEFAULT_SELECTOR=='Browser') {
       $lng->get_browser_language();
+      if (!zen_not_null($lng->language['id'])) $lng->set_language(DEFAULT_LANGUAGE);
```
drbyte: Debating whether to incorporate this fallback into get_browser_language() directly.

mc12345678: Had considered as well, but thought the "purity" of the function should be preserved (ie. what information is actually provided from the browser) and based on the response of that "object" to take the applicable action although some of what is "taken" is directly related to what is in the store's database, so certainly on the line with that "tough question". :) 

There are even alternative directions that could be taken, as the "default" code that follows basically assumes a vanilla install (english, 1, en) where a similar "search" could be performed for "those" specific values (expecting english to remain unchanged) and if not found then fallback on to the default language (assuming that its definition hasn't been lost.)

Sorry, I had a number of potential directions to consider as a solution, but determined that all other paths seemed to lead to the DEFAULT_LANGUAGE and there exists code to handle it therefore if the specific language_id isn't assigned then tried to be sure a language_id was identified without entering an infinite loop.

Do see that the admin side somewhat already forces DEFAULT_LANGUAGE even for browser detection.

drbyte:
👍  for 1.5.5

For v1.6.0, because language logic has been consolidated into 1 class, the ideal fix will be to add this sort of check to the `language` class, just before the return from `set_language()`.

zcwilt:
:+1:

PR #1450, opened 2017-07-04
Database: Add affectedRows function

lat9:
Add a function to the database class that returns the number of affected rows in the previous SELECT, UPDATE, REPLACE or DELETE query.


drbyte:
👍

zcwilt:
:+1:

PR #1451, opened 2017-07-04
Order notifiers special taxes

lat9:
I know I've got this messed up (pulling in two changes), but I don't understand how to coerce GitHub and SourceTree to reflect that.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -241,6 +241,9 @@ class order extends base {
 
       $this->info['tax_groups']["{$this->products[$index]['tax']}"] = 1;
 
+      $this->notify('NOTIFY_ORDER_QUERY_ADD_PRODUCT', $product, $index);
+
+
       $index++;
     } // end loop $orders_products as $product
 
@@ -412,6 +415,7 @@ class order extends base {
           $taxCountryId = $tax_address->fields['entry_country_id'];
           $taxZoneId = $tax_address->fields['entry_zone_id'];
         }
+        $this->notify('NOTIFY_ORDER_CART_TAX_ADDRESS', $tax_address);
       }
     }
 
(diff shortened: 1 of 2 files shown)


includes/classes/order.php
```diff
@@ -412,6 +415,7 @@ protected function cart(shoppingCart $basket, $customer_id = null, $sendto_addre
           $taxCountryId = $tax_address->fields['entry_country_id'];
           $taxZoneId = $tax_address->fields['entry_zone_id'];
         }
+        $this->notify('NOTIFY_ORDER_CART_TAX_ADDRESS', $tax_address);
```
drbyte: Is the goal here to merely allow the observer to "inquire" or "watch" what the $tax_address query-result contains?  Or to also actually allow changes to any of the calculated values?  

If it's to also allow changes, I'd add some more parameters, such as `$taxCountryId, $taxZoneId` ... just as additional params, not as a nested array.

includes/classes/order.php
```diff
@@ -687,6 +691,8 @@ public function do_calculate_taxes($taxRates, $index, $taxCountryId, $taxZoneId)
               $this->info['tax_groups'][$taxDescription] = $taxAdd;
           }
       }
+      
+      $this->notify('NOTIFY_ORDER_CART_TAX_CALCULATIONS_END', array('index' => $index, 'tax_add' => $tax_add, 'shown_price' => $shown_price, 'taxRates' => $taxRates));
```
drbyte: Are you passing this as an array because you primarily intend to use the observer for logging/reporting/recording what was done? ie: just watching? 

(Just curious.)

drbyte:
Oh. To split it out, you would have wanted to change back to the v155 branch *before* creating the new notifiers branch. Looks like you branched directly off of the affectedRows branch, thus including its commits as part of this branch. 

No worries. When we merge the other one, that'll likely already sync up with this one, without conflicts.

includes/classes/order.php
lat9: Yes, at this point in the processing, the data is _informational_ only; the observer can use the information (specifically the `index`) to make any updates needed to the associated product's tax.

includes/classes/order.php
lat9: At this point in the processing, only the `$tax_address` is needed/recorded in the observer since that address defines the taxable location.  Special processing might be needed for certain "zones" (e.g. states, countries or zipcodes).

drbyte:
👍

zcwilt:
:+1:

PR #1458, opened 2017-07-09
Allow Admin to disable all emails to certain customers

drbyte:
Ref https://www.zen-cart.com/showthread.php?190076-Disabling-Customers-Email


zcwilt:
Just a note. Need to think how to do this better if customer management in v160 is converted to LEAD.

but for now :+1:

ajeh:
:+1:

PR #1459, opened 2017-07-10
Alternative ensurance that attributes have values

### LINKED ISSUE #1454: Starting at text may not be displayed
While reviewing code to answer a ZC forum question, discovered that the includes/functions/functions_lookups.php file function zen_has_attributes_values could falsely indicate that a product does not have cost modifying values assigned to it.

The function as found in ZC 1.5.5 but shown as existing since the original commit to github is:

````php function zen_has_product_attributes_values($products_id) {
    global $db;
    $attributes_query = "select sum(options_values_price) as total
                         from " . TABLE_PRODUCTS_ATTRIBUTES . "
                         where products_id = '" . (int)$products_id . "'";
    $attributes = $db->Execute($attributes_query);
    if ($attributes->fields['total'] != 0) {
      return true;
    } else {
      return false;
    }
  }
````

The problem would be visible if say there were three attributes associated with a product having a price modification of: -x, 0 and +x where x is the same "number" on both sides. Unfortunately the function doesn't also take into account other attribute related price differences that could exist independent from the "lower left corner" attribute price such as attribute quantity discounts (starting at would likely need to calculate the lowest possible price which would likely be based on ordering the largest possible quantity within the discount group), but I also think that this factor is for something else and not necessarily this function.

There is nothing in the attributes controller to force the value entered to always be positive and there probably should not be. It is possible to enter a - for the price prefix against a negative number to add the value, but regardless the question that appears to need answered by this function is: is there an attribute that can modify the price of this product when selected?

Recommend this function be rewritten as:
````php
 function zen_has_product_attributes_values($products_id) {
    global $db;
    $attributes_query = "select count(options_values_price) as total, options_values_price as price
                         from " . TABLE_PRODUCTS_ATTRIBUTES . "
                         where products_id = '" . (int)$products_id . "'";
    $attributes = $db->Execute($attributes_query);
    if ($attributes->fields['total'] == 0) {
      return false;
    } else {
      foreach ($attributes as $attribute) {
        if ($attribute['price'] != 0) {
          return true;
        } 
      }
      return false;
    }
  }
````

There are some other minor improvements that can be made to the formatting of the query and/or if the same "total" should be used or if the function should otherwise be written to support/simplify upgrading... Haven't looked at that aspect of the modification yet.

mc12345678:
This is a modified version of the code proposal that fixes #1454.

The operation of the testing has been simplified to having only two returns instead of a series of logic "tests" of when false should be returned providing true if and only if a price `options_values_price` is identified as being non-zero, otherwise false.

### DIFF
--- a/includes/functions/functions_lookups.php
+++ b/includes/functions/functions_lookups.php
@@ -285,17 +285,20 @@
  */
   function zen_has_product_attributes_values($products_id) {
     global $db;
-    $attributes_query = "select sum(options_values_price) as total
+    $attributes_query = "select distinct count(options_values_price) as total, options_values_price as price
                          from " . TABLE_PRODUCTS_ATTRIBUTES . "
-                         where products_id = '" . (int)$products_id . "'";
+                         where products_id = " . (int)$products_id;
 
     $attributes = $db->Execute($attributes_query);
 
     if ($attributes->fields['total'] != 0) {
-      return true;
-    } else {
-      return false;
-    }
+      foreach ($attributes as $attribute) {
+        if ($attribute['price'] != 0) {
+          return true;
+        }
+      }
+    } 
+    return false;
   }
 
 /*


drbyte:
Is the `distinct` actually necessary? it seems odd to be asking for distinct on the `count()` function.

mc12345678:
It's unnecessary for the `count()` function but reduces the results on the `price` results, in particular when there is no price entered (one row returned). Otherwise more results are provided to be parsed until either all have been reviewed or a non-zero result is located.

drbyte:
👍

zcwilt:
:+1:

PR #1473, opened 2017-08-04
Reduce db queries for prices

mc12345678:
In several of the price functions, both the special and sales price is sought.
Both are obtained through the same function `zen_get_products_special_price` with a switch to either return only the specials price (true) or to return the overall special price (false) (sale price if present otherwise special price if a special, and false if neither)
If a product is not on special and does not have a sale then the returning result is `false` and therefore there is no further determination of the special necessary (ie. no further database query needed). If the result is anything other than `false`, then there may be a special and the value of that special (if present) remains to be determined.

While there remain other means of gaining efficiency in this process (such as returning all data collected from a one time database search/review), this at least reduces the redundancy of querying the database for results of the first `zen_get_products_special_price` call.  Though perhaps caching also helps in this overall concept, but the recommended modification mitigates the need to use the cache.

### DIFF
--- a/includes/functions/functions_prices.php
+++ b/includes/functions/functions_prices.php
@@ -493,13 +493,7 @@
     }
 
     $new_products_price = zen_get_products_base_price($product_id);
-    $new_sale_price = zen_get_products_special_price($product_id, false);
-    if ($new_sale_price !== false) {
-      $new_special_price = zen_get_products_special_price($product_id, true);
-    } else {
-      $new_special_price = false;
-    }
-
+    $new_special_price = zen_get_products_special_price($product_id, true);
     $discount_type_id = zen_get_products_sale_discount_type($product_id);
 
     if ($new_products_price != 0) {
--- a/includes/functions/functions_prices.php
+++ b/includes/functions/functions_prices.php
@@ -997,19 +997,19 @@ If a special exist * 10
     $product_check = $db->Execute("select products_tax_class_id, products_price, products_priced_by_attribute, product_is_free, product_is_call from " . TABLE_PRODUCTS . " where products_id = '" . (int)$products_id . "'" . " limit 1");
 
     // If Free, Show it
-    if ($product_check->fields['product_is_free'] == '1') {
+    if ((int)$product_check->fields['product_is_free'] === 1) {
       return 0;
     }
 
     $display_sale_price = zen_get_products_special_price($products_id, false);
 
-    if ($display_sale_price) {
+    if ($display_sale_price !== false) {
       return $display_sale_price;
     }
 
     $display_special_price = zen_get_products_special_price($products_id, true);
 
-    if ($display_special_price) {
+    if ($display_special_price !== false) {
       return $display_special_price;
     }
 
--- a/includes/functions/functions_prices.php
+++ b/includes/functions/functions_prices.php
@@ -201,15 +201,15 @@
       return '';
     }
 
-    $show_display_price = '';
     $display_normal_price = zen_get_products_base_price($products_id);
     $display_sale_price = zen_get_products_special_price($products_id, false);
+
     if ($display_sale_price !== false) {
       $display_special_price = zen_get_products_special_price($products_id, true);
     } else {
       $display_special_price = false;
     }
-    
+
     $show_sale_discount = '';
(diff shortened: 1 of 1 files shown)


includes/functions/functions_prices.php
```diff
@@ -489,8 +493,12 @@ function zen_get_discount_calc($product_id, $attributes_id = false, $attributes_
     }
 
     $new_products_price = zen_get_products_base_price($product_id);
-    $new_special_price = zen_get_products_special_price($product_id, true);
     $new_sale_price = zen_get_products_special_price($product_id, false);
+    if ($new_sale_price !== false) {
```
drbyte: Curiously, `$new_sale_price` isn't used in this function anywhere. 
Let's revert this section (line 492) back, and then this new if/else can be removed too.

includes/functions/functions_prices.php
```diff
@@ -990,8 +998,12 @@ function zen_get_products_actual_price($products_id) {
 
     $show_display_price = '';
     $display_normal_price = zen_get_products_base_price($products_id);
-    $display_special_price = zen_get_products_special_price($products_id, true);
```
drbyte: I wonder if this is a more efficient version of this actual-price function?

```php
  function zen_get_products_actual_price($products_id) {
    global $db;
    $product_check = $db->Execute("select products_tax_class_id, products_price, products_priced_by_attribute, product_is_free, product_is_call from " . TABLE_PRODUCTS . " where products_id = '" . (int)$products_id . "'" . " limit 1");

    // If Free, Show it
    if ($product_check->fields['product_is_free'] == '1') {
      return 0;
    }

    $display_sale_price = zen_get_products_special_price($products_id, false);

    if ($display_sale_price) {
      return $display_sale_price;
    }

    $display_special_price = zen_get_products_special_price($products_id, true);

    if ($display_special_price) {
      return $display_special_price;
    }

    $display_normal_price = zen_get_products_base_price($products_id);
    return $display_normal_price;
  }
```

drbyte:
@mc12345678 I've pushed some commits to this branch. Thoughts?

includes/functions/functions_prices.php
```diff
@@ -985,29 +987,28 @@ function zen_get_products_sale_discount($product_id = false, $categories_id = fa
...
+      return 0;
+    }
 
-    $products_actual_price = $display_normal_price;
+    $display_sale_price = zen_get_products_special_price($products_id, false);
 
-    if ($display_special_price) {
-      $products_actual_price = $display_special_price;
-    }
     if ($display_sale_price) {
```
mc12345678: Basically just from a code review, shouldn't this and the others be evaluated against `!== false` or something? Ie. what if there is a price of "0" intentionally returned as compared to the few cases in `zen_get_products_special_price` where `false` is returned and there are a few where `0` is returned (if calculated sale price is less than 0)?

includes/functions/functions_prices.php
```diff
@@ -490,8 +494,6 @@ function zen_get_discount_calc($product_id, $attributes_id = false, $attributes_
 
     $new_products_price = zen_get_products_base_price($product_id);
     $new_special_price = zen_get_products_special_price($product_id, true);
-    $new_sale_price = zen_get_products_special_price($product_id, false);
```
mc12345678: Yes, in the spirit of reducing queries, this would be one to be removed from the code as it currently stands; however, as identified in #1404, there are some discrepancies that would need this value to be "returned" for use unless an alternative solution has been developed.

includes/functions/functions_prices.php
drbyte: I suppose. For the last 10 years in production it hasn't been something complained about. Not that that's the only factor to consider. It just informs about the significance of the issue.

If one were to explicitly test for strict falsey values, what would the logic be? ie: what's the fallback action?

includes/functions/functions_prices.php
drbyte: Sigh. Why do we have 4 PRs all about the same subject, all going in different directions?

includes/functions/functions_prices.php
drbyte: Okay, I think you're right. Let's do the comparison `!== false` for safety.

includes/functions/functions_prices.php
drbyte: Okay, I'm gonna back out this change.

PR #1474, opened 2017-08-04
Update internal notes of is_coupon_valid_for_sales

mc12345678:
Found while reviewing other code and seeing this new functionality.

The action taken within the function `is_coupon_valid_for_sales` did not coincide with some of the associated code comments.  These have been updated to reflect either the action or the meaning for reaching/being at that area.


drbyte:
Thanks

PR #1477, opened 2017-08-07
Restore no-default on tinytext fields - partial reversal of #1470

drbyte:
strict schema rules don't like a "default" value on text/blob fields, including tinytext


zcwilt:
:+1:

ajeh:
:+1:

PR #1479, opened 2017-08-08
Store Manager - improve repair of invalid master_categories_id

drbyte:
Rework the `master_categories_id` reset to use original insert order, for greater accuracy during resets.

~~Stores that use 3rd-party import tools to add/delete products and categories can end up with products linked to invalid master-categories.
This new repair button finds those products and sets them to the first-found category to which that product is linked.~~

### DIFF
--- a/admin/store_manager.php
+++ b/admin/store_manager.php
@@ -1,10 +1,10 @@
 <?php
 /**
  * @package admin
- * @copyright Copyright 2003-2014 Zen Cart Development Team
+ * @copyright Copyright 2003-2017 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version GIT: $Id: Author: DrByte  Jun 30 2014 Modified in v1.5.4 $
+ * @version GIT: $Id: Author: DrByte  Aug 2017 Modified in v1.5.6 $
  */
 
   require('includes/application_top.php');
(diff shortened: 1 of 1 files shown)


ajeh:
What is wrong with the existing:
Reset ALL Products Master Categories ID
to be used for Linked Products and Pricing: 

in the Store manager?

drbyte:
The existing one resets ALL products, whether their linked master_categories_id is valid or not.

The added option here ONLY resets those with an invalid master_categories_id, so is not as volatile as the other option.

drbyte:
I wonder if an even better option is to (also) provide a list of clickable (admin) links to products with an invalid `master_categories_id` so the storeowner could manually investigate each?

torvista:
While an extra tools are welcome, the problem is in knowing that there is a problem to fix.
Hence my post: [https://www.zen-cart.com/showthread.php?192247-Finding-all-products-wit-quot-invalid-master-categorie-id-quot&p=1333807#post1333807](url)
With respect to the suggestion, if a product is unlinked/it only has a single entry in product_to_categories, there is no issue with resetting it's mcat to the single entry in product_to_categories with a button.
Stepping further back from the problem, is there any reason not to **automatically** correct this when detected, for example when zen_get_master_categories_pulldown gets used, with a corresponding messagestack. 
And/Or, if an invalid mcat is detected/fixed as above on the one product, a global check gets run to find  and correct any others (automatically where possible) and provide a message for linked products to be corrected manually.
I think that if there is no issue with making some things automatic, they should be.

ajeh:
Turn off the index and you will get the first Category that a product was added to when there are multiple Categories as that is "more" likely to be the original Master ...

drbyte:
Haha! 

Let's not be telling people to delete their indexes. That's more of a problem than it is a solution.

drbyte:
Update: reworked to bypass the index, per @ajeh suggestion.

ajeh:
:+1:

zcwilt:
:+1:

PR #1480, opened 2017-08-14
Fixes #1457 (M'ple GC's purchased ...)

### LINKED ISSUE #1457: Multiple GC purchased not recorded if GVs aren't queued
ZC1.5.5e.  Configure the GC order-total so that GCs aren't queued (not, IMO, a good idea ... but).

Using the demo data, a customer has an existing GC balance of $10.00 and purchases 2 x $5.00 cards and 1 x $10.00 GC in a single order.  Upon order-completion, the customer's account reflects a GC balance of $20.00 instead of $30.00.

The issue lies in /includes/modules/order_total/ot_gv.php's `user_has_gv_account` function:
```
  function user_has_gv_account($c_id) {
    global $db;
    $gv_result = $db->Execute("select amount from " . TABLE_COUPON_GV_CUSTOMER . " where customer_id = '" . (int)$c_id . "'");
    if ($gv_result->RecordCount() > 0) {
      return $gv_result->fields['amount'];
    }
    return 0; // use 0 because 'false' was preventing checkout_payment from continuing
  }
```
Since there are multiple GCs in the order, the function is called multiple times to determine the customer's GC balance.  Unfortunately, the database cache "remembers" only the first value (10.00); that's why the full GC purchase isn't applied.

To correct for the ZC1.5.5 series:
```
  function user_has_gv_account($c_id) {
    global $db;
    $gv_result = $db->Execute("select amount from " . TABLE_COUPON_GV_CUSTOMER . " where customer_id = '" . (int)$c_id . "'", false, false, 0, true);
    if ($gv_result->RecordCount() > 0) {
      return $gv_result->fields['amount'];
    }
    return 0; // use 0 because 'false' was preventing checkout_payment from continuing
  }
```

For the ZC1.6.0 series, replace the $db->Execute function with $db->ExecuteNoCache.

lat9:


drbyte:
👍

zcwilt:
:+1:

PR #1486, opened 2017-08-15
Rename "in stock"/"out of stock" to "enabled"/"disabled"

### LINKED ISSUE #1476: Admin: TEXT_PRODUCT_AVAILABLE/TEXT_PRODUCT_NOT_AVAILABLE still refer to stock
I still think these should be Enabled/Disabled:
https://www.zen-cart.com/showthread.php?215759-quot-in-stock-out-of-stock-quot-should-be-quot-enabled-disabled-quot

I see these defined in language product.php and in categories.php but don't see where the category one would get used.

drbyte:
Closes #1476


zcwilt:
:+1:

PR #1487, opened 2017-08-15
Remove unused `final_price` field from customers_basket table.

### LINKED ISSUE #1285: delete unused final_price from customer_basket table
In the table customer_basket there is a field called final_price. It seems to me after searching through the code that this field is not used (any more).

drbyte:
Closes #1285


zcwilt:
:+1:

PR #1490, opened 2017-08-15
Remove SHOW_GV_QUEUE define from english.php into ot_gv settings

### LINKED ISSUE #1420: SHOW_GV_QUEUE in english.php
in the admin side language file english.php on line 522 (ZC 1.5.5e), there is the define
```
// check GV release queue and alert store owner
  define('SHOW_GV_QUEUE',true);
```
Just curious as to why this is in a language define and not in the database.
I suspect it is there as some legacy from Osc,

drbyte:
Fixes #1420


zcwilt:
:+1:

PR #1492, opened 2017-08-17
Ignore sub-arrays in zen_get/post_all_get_params

drbyte:
Nested arrays for get/post values are not used internally in any way that needs re-echoing to the page. Skipping them to avoid throwing false errors when bots/probers do unauthorized stuff.
Ref https://www.zen-cart.com/showthread.php?222694


zcwilt:
:+1:

ajeh:
:+1:

PR #1494, opened 2017-08-17
Fix confusing text related to copying qty discounts when cloning products

drbyte:
Ref https://www.zen-cart.com/showthread.php?222731


zcwilt:
:+1:

ajeh:
:+1:

PR #1505, opened 2017-09-04
Allow admin javascript to have its own file instead of altering core

drbyte:
Custom admin javascript can now be added to `/admin/includes/javascript_loader.php` as needed.


drbyte:
Good point. I've updated to harmonize them a bit. 

A wholesale admin rewrite with a complete templating system would make this simpler; developers love that idea, but storeowners/upgraders and plugin authors despise it. :(

barco57:
anything that forces updates of add-on modules I am all for, Break all the things!

zcwilt:
:+1:

ajeh:
:+1:

PR #1506, opened 2017-09-08
Split TEXT_VISTORS_CART to create TEXT_CART_HELP for shopping_cart page

### LINKED ISSUE #1080: Fix wording of Visitors Cart text notice
`TEXT_VISITORS_CART` could use some tweaks to make it more explicit that it applies only to customers with an account.

And, it's not the same in `login.php` and `shopping_cart.php` ... so it might be worth splitting them into different defines because they're quite different.

drbyte:
Closes #1080

### DIFF
--- a/includes/languages/english/shopping_cart.php
+++ b/includes/languages/english/shopping_cart.php
@@ -1,10 +1,10 @@
 <?php
 /**
  * @package languageDefines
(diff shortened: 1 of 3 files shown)


zcwilt:
:+1:

ajeh:
:+1:

PR #1507, opened 2017-09-10
Negate '/' match after store folder in location regex

dakanji:
Previous regex would match /store/downloads/xyz.txt for instance and given position early in conf file, would be executed before the downloads folder location.

### DIFF
--- a/zc_install/includes/nginx_conf/ngx_server_template.txt
+++ b/zc_install/includes/nginx_conf/ngx_server_template.txt
@@ -67,7 +67,7 @@
 		return 418;
 	}
 	## Allow all requests to specific static file types
-	location ~ %%store_folder%%/[^/]+\.(md|ico|txt|html) {
+	location ~ %%store_folder%%/.+\.(md|ico|txt|html)$ {
 		expires $zencart_expires;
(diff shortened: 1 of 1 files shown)


drbyte:
Hmmm ... the net "change" here is null.   Is something missing from the commits?

zc_install/includes/nginx_conf/ngx_server_template.txt
```diff
@@ -67,7 +67,7 @@
 		return 418;
 	}
 	## Allow all requests to specific static file types
-	location ~ %%store_folder%%/.+\.(md|ico|txt|html)$ {
+	location ~ %%store_folder%%/[^/]+\.(md|ico|txt|html)$ {
```
drbyte: Should this same tweak be applied everywhere in this file?

Like lines 79, 84, 89, 106, etc.

dakanji:
Had a look and all the others are for "XYZ and/or subfolders".

Issue with the fixed line was that it was supposed to be limited to items directly under the root folder and not include any subfolders.

drbyte:
Fair. Thx

PR #1512, opened 2017-09-21
Correct array operation when calling zen_option_name_base_expects_no_…

mc12345678:
…values

The way the code ended up being written was that if an array is presented
to the function then the code would error because the last internal
line of code would set the $sql2 to just a right parentheses instead
of adding a parenthesis at the end of the string.


drbyte:
👍

drbyte:
Thanks

PR #1514, opened 2017-09-21
Mc12345678 remove zen_get_buy_now_qty from template (#13)

mc12345678:
Update product main_template_vars to include the buy_now_qty and template file to reference previously collected data.

Adding the result of the zen_get_buy_now_qty into the product main_template_vars removes the processing (and potential modifications necessary to support alternate processes) from the template page.  This allows the template to present the information that is needed/expected without relying on additional processing or dependency on internal functions (that may need modification to support expected/desired result).

This has also been considered as a way to solve an issue that can occur when using the product option type Grid as offered by Products attributes grid where specifically the use of units or minimum quantities are intended to only impact the individual attribute selections not the overall "product".


drbyte:
👍

zcwilt:
:+1:

PR #1515, opened 2017-09-22
Support ISO-8601 style date-of-birth date

mc12345678:
Support date of birth collection from "cell phones" as suggested in the forum: https://www.zen-cart.com/showthread.php?222560-New-customer-using-smartphone-Date-of-Birth-info-not-accepted&p=1333443#post1333443

Source of the issue appears to be by the use of `<input type="date" />` in the template file which returns an ISO-8601 styled date.  This date style may not conform with the date format "recommended" to the visitor.  This change does not prevent a guest from entering the data in the ISO-8601 (or expanded as provided by this code) when typing in the number, but it does support the use of "javascript" style date pickers to provide the date in a format understandable for further processing.

An ISO-8601 specifically formatted date would be one that has a dash (-) only between dates and does not include the additional characters supported by the incorporated regex.  To make this regex only recognize ISO-8601 "alternate" formats, change `(|-|\/)` to `(-)`, otherwise the existing regex permits all numbers to be "pressed" together, to have a dash (-) or to have a forward slash (/); however, it is expected that the same symbol/method is used for each separator (or lack of), otherwise the date is considered incorrectly entered.

### DIFF
--- a/includes/modules/pages/document_general_info/main_template_vars.php
+++ b/includes/modules/pages/document_general_info/main_template_vars.php
@@ -110,6 +110,7 @@
 
   $products_qty_box_status = $product_info->fields['products_qty_box_status'];
   $products_quantity_order_max = $product_info->fields['products_quantity_order_max'];
+  $products_get_buy_now_qty = zen_get_buy_now_qty($_GET['products_id']);
 
   $products_base_price = $currencies->display_price(zen_get_products_base_price((int)$_GET['products_id']),
                       zen_get_tax_rate($product_info->fields['products_tax_class_id']));
--- a/includes/modules/pages/document_product_info/main_template_vars.php
+++ b/includes/modules/pages/document_product_info/main_template_vars.php
@@ -110,6 +110,7 @@
 
   $products_qty_box_status = $product_info->fields['products_qty_box_status'];
   $products_quantity_order_max = $product_info->fields['products_quantity_order_max'];
+  $products_get_buy_now_qty = zen_get_buy_now_qty($_GET['products_id']);
 
   $products_base_price = $currencies->display_price(zen_get_products_base_price((int)$_GET['products_id']),
                       zen_get_tax_rate($product_info->fields['products_tax_class_id']));
--- a/includes/modules/pages/product_free_shipping_info/main_template_vars.php
+++ b/includes/modules/pages/product_free_shipping_info/main_template_vars.php
@@ -110,6 +110,7 @@
 
   $products_qty_box_status = $product_info->fields['products_qty_box_status'];
   $products_quantity_order_max = $product_info->fields['products_quantity_order_max'];
+  $products_get_buy_now_qty = zen_get_buy_now_qty($_GET['products_id']);
 
   $products_base_price = $currencies->display_price(zen_get_products_base_price((int)$_GET['products_id']),
                       zen_get_tax_rate($product_info->fields['products_tax_class_id']));
--- a/includes/modules/pages/product_info/main_template_vars.php
+++ b/includes/modules/pages/product_info/main_template_vars.php
@@ -111,6 +111,7 @@
 
   $products_qty_box_status = $product_info->fields['products_qty_box_status'];
   $products_quantity_order_max = $product_info->fields['products_quantity_order_max'];
+  $products_get_buy_now_qty = zen_get_buy_now_qty($_GET['products_id']);
 
   $products_base_price = $currencies->display_price(zen_get_products_base_price((int)$_GET['products_id']),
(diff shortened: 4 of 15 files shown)


includes/modules/create_account.php
```diff
@@ -110,6 +110,10 @@
 
   if (ACCOUNT_DOB == 'true') {
     if (ENTRY_DOB_MIN_LENGTH > 0 or !empty($_POST['dob'])) {
+      if (preg_match('/^([0-9]{4})(|-|\/)([0-9]{2})\2([0-9]{2})$/', $dob)) {
+        $_POST['dob'] = date(DATE_FORMAT, strtotime($dob));
+        $dob = date(DATE_FORMAT, strtotime($dob));
```
drbyte: Thanks. 

NOTE: These lines can be converted into one, like this:

```diff
-        $_POST['dob'] = date(DATE_FORMAT, strtotime($dob));
-        $dob = date(DATE_FORMAT, strtotime($dob));
+        $_POST['dob'] = $dob = date(DATE_FORMAT, strtotime($dob));
```

(Note: when doing double assignments like this, the innermost one is done first, so $dob is assigned the result of date() and then $_POST[dob] is assigned what $dob was given. This is seldom "important", but worth knowing.)

mc12345678: Tried to make the change so that it would be a squashed commit; however, wasn't successful online.

includes/modules/create_account.php
```diff
@@ -110,6 +110,9 @@
 
   if (ACCOUNT_DOB == 'true') {
     if (ENTRY_DOB_MIN_LENGTH > 0 or !empty($_POST['dob'])) {
+      if (preg_match('/^([0-9]{4})(|-|\/)([0-9]{2})\2([0-9]{2})$/', $dob)) {
```
drbyte: The logic here as I'm reading it is "4 digits, plus a delimiter of either `-` or `/`, plus 2 digits, and same delimiter, and 2 digits", right?

Looking at the regex closer, is the first `|` necessary here?
`(|-|\/)`

mc12345678: Somewhat correct, the first emptiness before the first pipe (`|`) indicates no divider, such that would be 4 digits, no delimiter, a `-`, or a `/`, 2 digits, and same excluded or included delimiter, and then 2 digits.

Necessary may be a strong statement; however, it allows supporting not having a divider between the 4-2-2 pairing as identified in the php manual ISO8601 section for 3 of the 4 presented methods: http://php.net/manual/en/datetime.formats.date.php

The first pipe (`|`) inspects the entire content as an eight digit number by having no "filter" on it.  Without it, some testing/testers failed the data as invalid.

I thought that trying to incorporate the 4th method would cause an issue as it presents YY-MM-DD which when speaking of historical dates, if any, would be problematic with all date types if say referring to 11-07-10 which could be interpreted in any way because all numbers are less than the maximum number of 12 for month. :/ THAT will be an issue that may have to be addressed later in some other way...

drbyte: Usually to make an item optional, a `?` is used (`?` means "0 or more times").

Would this give the same results?

```diff
-'/^([0-9]{4})(|-|\/)([0-9]{2})\2([0-9]{2})$/'
+'/^([0-9]{4})(-|\/)?([0-9]{2})\2?([0-9]{2})$/'
```

mc12345678: Not exactly: the `\2?` makes the 2nd "entry" optional such that the following tests all appear as acceptable (random date chosen but consistently):
2017-04-09 (OK)
20170409 (OK)
2017/04/09 (OK)
2017-0409 (not)
2017/0409 (not)

drbyte: Ya, I wondered about that after posting. Thanks for testing that possibility.

drbyte:
👍

includes/modules/create_account.php
mc12345678: No problem, appreciate the review especially when it comes to the wild animal of regex.

drbyte:
Are you closing this because you don't want it merged?

drbyte:
Oh, I see what's happened. You did this PR from your `v156` branch, instead of a separate feature-branch (ie: something with a different name) .... but a few hours ago you reset your `v156` branch, which then wiped out this PR because it hasn't been merged yet.

drbyte:
I've recreated it as #1518 ... perhaps you can double-check that I reconstructed it correctly from memory :)

mc12345678:
On it boss. Sorry, yeah I was cleaning up my local copy to bring it up to official and then make my online copy match. Knew there would be a problem with my accidental start from the main v156 branch.

PR #1516, opened 2017-09-22
Refactor update_product for strict data types

drbyte:


admin/includes/modules/update_product.php
```diff
@@ -15,40 +15,33 @@
...
+            'products_quantity' => convertToFloat($_POST['products_quantity']),
+            'products_type' => (int)$_GET['product_type'],
+            'products_model' => zen_db_prepare_input($_POST['products_model']),
+            'products_price' => convertToFloat($_POST['products_price']),
+            'products_date_available' => $products_date_available,
+            'products_weight' => convertToFloat($_POST['products_weight']),
+            'products_status' => (int)$_POST['products_status'],
+            'products_virtual' => (int)$_POST['products_virtual'],
+            'products_tax_class_id' => (int)$_POST['products_tax_class_id'],
+            'manufacturers_id' => convertToFloat($_POST['manufacturers_id']),
```
mc12345678: Shouldn't this (manufacturers_id) be an (int) cast instead (unless there is a move afoot to have something like a Dewey Decimal Classification system)? (not that it is likely to cause a problem (famous last words) by being a float.)

drbyte: Yup. Good catch. Got lost in the sea of fields...

drbyte: Updated.

zcwilt:
:+1:

ajeh:
:+1:

PR #1519, opened 2017-10-03
CDN protocol updates to some admin js assets

Zen4All:
- Added protocol to links
- change moment.min.js to moment-with-locales.min.js in favor of multi lingual use

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1519)


drbyte:
👍

PR #1520, opened 2017-10-03
Add a Simple array to sanitize operation.

mc12345678:
This also added transfer of the parameterDefinition through the
processCustom function to allow the custom function to know the associated
definition.

### DIFF
--- a/admin/includes/classes/AdminRequestSanitizer.php
+++ b/admin/includes/classes/AdminRequestSanitizer.php
@@ -72,6 +72,7 @@ class AdminRequestSanitizer extends base
         $this->doStrictSanitization = false;
         $this->getKeysAlreadySanitized = array();
         $this->postKeysAlreadySanitized = array();
+        $this->arrayName = '';
         $this->debugMessages[] = 'Incoming GET Request ' . print_r($_GET, true);
         $this->debugMessages[] = 'Incoming POST Request ' . print_r($_POST, true);
     }
@@ -176,10 +177,12 @@ class AdminRequestSanitizer extends base
                 $result = $this->findSanitizerFromRequestMethod($parameterName, $parameterDefinitions);
             }
             if ($result) {
+                $this->arrayName = '';
                 $this->runSpecificSanitizer($parameterName, $result);
             }
         }
         if ($this->doStrictSanitization) {
+            $this->arrayName = '';
             $this->filterStrictSanitizeKeys();
             $this->filterStrictSanitizeValues();
         }
@@ -299,7 +302,7 @@ class AdminRequestSanitizer extends base
     {
         $func = $this->adminSanitizerTypes[$parameterDefinition['sanitizerType']]['function'];
         $this->debugMessages[] = 'SANITIZER CUSTOM == ' . $sanitizerName;
-        $func($this, $parameterName);
+        $func($this, $parameterName, $parameterDefinition);
     }
 
     /**
@@ -312,8 +315,10 @@ class AdminRequestSanitizer extends base
             $this->getKeysAlreadySanitized[] = $parameterName;
         }
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING NULL ACTION(POST) == ' . $parameterName;
-            $this->postKeysAlreadySanitized[] = $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING NULL ACTION(POST) == ' . $this->arrayName;
+            $this->postKeysAlreadySanitized[] = $this->arrayName;
         }
 
     }
@@ -329,8 +334,10 @@ class AdminRequestSanitizer extends base
             $_GET[$parameterName] = preg_replace('/[^\/ 0-9a-zA-Z_:@.-]/', '', $_GET[$parameterName]);
         }
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING SIMPLE_ALPHANUM_PLUS(POST) == ' . $parameterName;
-            $this->postKeysAlreadySanitized[] = $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING SIMPLE_ALPHANUM_PLUS(POST) == ' . $this->arrayName;
+            $this->postKeysAlreadySanitized[] = $this->arrayName;
             $_POST[$parameterName] = preg_replace('/[^\/ 0-9a-zA-Z_:@.-]/', '', $_POST[$parameterName]);
         }
     }
@@ -341,9 +348,11 @@ class AdminRequestSanitizer extends base
     private function filterConvertInt($parameterName)
     {
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING CONVERT_INT (POST) == ' . $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING CONVERT_INT (POST) == ' . $this->arrayName;
             $_POST[$parameterName] = (int)$_POST[$parameterName];
-            $this->postKeysAlreadySanitized[] = $parameterName;
+            $this->postKeysAlreadySanitized[] = $this->arrayName;
         }
         if (isset($_GET[$parameterName])) {
             $this->debugMessages[] = 'PROCESSING CONVERT_INT (GET) == ' . $parameterName;
@@ -360,9 +369,11 @@ class AdminRequestSanitizer extends base
     {
         $filedirRegex = '~[^0-9a-z' . preg_quote('.!@#$%^& ()`_+-~/' . '\\', '~') . ']~i';
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING FILE_DIR_REGEX == ' . $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING FILE_DIR_REGEX == ' . $this->arrayName;
             $_POST[$parameterName] = preg_replace($filedirRegex, '', $_POST[$parameterName]);
-            $this->postKeysAlreadySanitized[] = $parameterName;
+            $this->postKeysAlreadySanitized[] = $this->arrayName;
         }
 
     }
@@ -374,9 +385,11 @@ class AdminRequestSanitizer extends base
     {
         $alphaNumDashUnderscore = '/[^a-z0-9_-]/i';
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING ALPHANUM_DASH_UNDERSCORE (POST) == ' . $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING ALPHANUM_DASH_UNDERSCORE (POST) == ' . $this->arrayName;
             $_POST[$parameterName] = preg_replace($alphaNumDashUnderscore, '', $_POST[$parameterName]);
-            $this->postKeysAlreadySanitized[] = $parameterName;
+            $this->postKeysAlreadySanitized[] = $this->arrayName;
         }
         if (isset($_GET[$parameterName])) {
             $this->debugMessages[] = 'PROCESSING ALPHANUM_DASH_UNDERSCORE (GET) == ' . $parameterName;
@@ -393,9 +406,11 @@ class AdminRequestSanitizer extends base
     {
         $prodNameRegex = '~<\/?scri|on(load|mouse|error|read|key)(up|down)? ?=|[^(class|style)] ?= ?(\(|")|<!~i';
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING WORDS_AND_SYMBOLS_REGEX (POST) == ' . $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING WORDS_AND_SYMBOLS_REGEX (POST) == ' . $this->arrayName;
             $_POST[$parameterName] = preg_replace($prodNameRegex, '', $_POST[$parameterName]);
-            $this->postKeysAlreadySanitized[] = $parameterName;
+            $this->postKeysAlreadySanitized[] = $this->arrayName;
         }
         if (isset($_GET[$parameterName])) {
             $this->debugMessages[] = 'PROCESSING WORDS_AND_SYMBOLS_REGEX (GET) == ' . $parameterName;
@@ -411,15 +426,18 @@ class AdminRequestSanitizer extends base
     {
         $prodDescRegex = '~(load=|= ?\(|<![^-])~i';
         if (isset($_POST[$parameterName])) {
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
             $this->debugMessages[] = 'PROCESSING PRODUCT_DESC_REGEX == ' . $parameterName;
             if (is_array($_POST[$parameterName])) {
                 foreach ($_POST[$parameterName] as $pKey => $pValue) {
+                    $currentArrayName = $this->setCurrentArrayName($pKey);
                     $_POST[$parameterName][$pKey] = preg_replace($prodDescRegex, '', $_POST[$parameterName][$pKey]);
-                    $this->postKeysAlreadySanitized[] = $parameterName;
+                    $this->postKeysAlreadySanitized[] = $currentArrayName;
                 }
             } else {
                 $_POST[$parameterName] = preg_replace($prodDescRegex, '', $_POST[$parameterName]);
-                $this->postKeysAlreadySanitized[] = $parameterName;
+                $this->postKeysAlreadySanitized[] = $this->arrayName;
             }
         }
     }
@@ -430,10 +448,13 @@ class AdminRequestSanitizer extends base
     private function filterMetaTags($parameterName)
     {
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING META_TAGS == ' . $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING META_TAGS == ' . $this->arrayName;
             foreach ($_POST[$parameterName] as $pKey => $pValue) {
+                $currentArrayName = $this->setCurrentArrayName($pKey);
                 $_POST[$parameterName][$pKey] = htmlspecialchars($_POST[$parameterName][$pKey], ENT_COMPAT, 'utf-8', false);
-                $this->postKeysAlreadySanitized[] = $parameterName;
+                $this->postKeysAlreadySanitized[] = $currentArrayName;
             }
         }
     }
@@ -444,10 +465,11 @@ class AdminRequestSanitizer extends base
     private function filterSanitizeEmail($parameterName)
     {
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING SANITIZE_EMAIL (POST) == ' . $parameterName;
-            $result = filter_var($_POST[$parameterName], FILTER_SANITIZE_EMAIL);
-            $_POST[$parameterName] = $result;
-            $this->postKeysAlreadySanitized[] = $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING SANITIZE_EMAIL (POST) == ' . $this->arrayName;
+            $_POST[$parameterName] = filter_var($_POST[$parameterName], FILTER_SANITIZE_EMAIL);
+            $this->postKeysAlreadySanitized[] = $this->arrayName;
         }
         if (isset($_GET[$parameterName])) {
             $this->debugMessages[] = 'PROCESSING SANITIZE_EMAIL (GET) == ' . $parameterName;
@@ -463,10 +485,11 @@ class AdminRequestSanitizer extends base
     private function filterSanitizeEmailAudience($parameterName)
     {
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING SANITIZE_EMAIL_AUDIENCE (POST) == ' . $parameterName;
-            $result = htmlspecialchars($_POST[$parameterName]);
-            $_POST[$parameterName] = $result;
-            $this->postKeysAlreadySanitized[] = $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING SANITIZE_EMAIL_AUDIENCE (POST) == ' . $this->arrayName;
+            $_POST[$parameterName] = htmlspecialchars($_POST[$parameterName]);
+            $this->postKeysAlreadySanitized[] = $this->arrayName;
         }
     }
 
@@ -477,14 +500,17 @@ class AdminRequestSanitizer extends base
     {
         $urlRegex = '~([^a-z0-9\'!#$&%@();:/=?_\~\[\]-]|[><])~i';
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING PRODUCT_URL_REGEX == ' . $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName);
+            $this->debugMessages[] = 'PROCESSING PRODUCT_URL_REGEX == ' . $this->arrayName;
             foreach ($_POST[$parameterName] as $pKey => $pValue) {
+                $currentArrayName = $this->setCurrentArrayname($pKey, $this->arrayName);
                 $newValue = filter_var($_POST[$parameterName][$pKey], FILTER_SANITIZE_URL);
                 if ($newValue === false) {
                     $newValue = preg_replace($urlRegex, '', $_POST[$parameterName][$pKey]);
                 }
                 $_POST[$parameterName][$pKey] = $newValue;
-                $this->postKeysAlreadySanitized[] = $parameterName;
+                $this->postKeysAlreadySanitized[] = $currentArrayName;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


drbyte:
Can you describe the "use case" which led you to come up with this? ie: how do you intend this to be used?

mc12345678:
So the use initiated from: https://www.zen-cart.com/showthread.php?222894-Adding-Adminer-for-MySQL-vs-the-ZC-Admin-sanitizer

Which upon further review identified two things: 
1) the MULTI_DIMENSIONAL type required an intermediate array between the initial key and the "final" value and would not work for the information presented.
2) After application of this change submitted, it also broke MULTI_DIMENSIONAL.

So I have been working by internal modification to the filterSimpleArray function to correct the issue created by modification of the traverseStrictSanitize function.

While I have identified a method that will make this work (adding 'STRICT_SANITIZE_VALUES' as a potential builtin sanitizer), I am trying to avoid adding it.

There is an additional modification that was made in the provided commit which was to expand the capability of the processCustom function which appears would also satisfy the condition; however, it still seems that the filterStrictSanitizeValues function could/would need additional support information or the sanitization provided by it would need to be duplicated in the filterSimpleArray.

The post fields array data:
````html
Array
(
    [securityToken] => obscured
    [fields] => Array
        (
            [number] => 1<onload
            [url] => https://mydomain.com/index.php?main_page=product_info&products_id=1
        )
)
````
When posted does not get processed through the strictsanitizer for `fields[url]` even when that particular field is given  `array('sanitizerType' => 'STRICT_SANITIZE_VALUES')`

Ultimately it was a stumping situation which I haven't been able to give up because of my drive to solve a problem. :) 

the `fields[number]` was given a sanitizer of 'WORDS_AND_SYMBOLS_REGEX' and does process correctly through the function.  So I guess ultimately I am now trying to see how to ensure that when using the proposed style that other keys and values within the array will be strictly sanitized when that condition is expected or explicitly identified to the field (which appears to be an issue with any of the fields in the sanitizer that if it is desired to strictly sanitize regardless of the base setting, then the specific strict sanitizer does not get executed.

File: admin/includes/classes/AdminRequestSanitizer.php [hunk 1]
```diff
@@ -555,6 +555,45 @@ private function filterMultiDimensional($parameterName, $parameterDefinition)
...
+                $type = $parameterDefinition['params'][$pkey]['sanitizerType'];
+                $params = isset($parameterDefinition['params'][$pkey]['params']) ? $parameterDefinition['params'][$pkey]['params'] : null;
+                $newParameterDefinition = array('sanitizerType' => $type, 'params' => $params);
+                $this->runSpecificSanitizer($pkey, $newParameterDefinition);
+                $requestPost[$parameterName][$pkey] = $_POST[$pkey];
+                if ($type != 'STRICT_SANITIZE_VALUES' && $type != 'STRICT_SANITIZE_KEYS') {
+                    $this->postKeysAlreadySanitized[] = $pkey;
+                }
+            }
+            if ($this->doStrictSanitization) {
```
mc12345678: The code in this area looks like it could use some additional work to fully incorporate STRICT_SANITIZE_KEYS. 

Ie. `unset` the associated key in the `$requestPost` variable up front, then encapsulate the assignment in an `if` statement to see if `$key` is set or exists in `$_POST`.

zcwilt:
My apologies. I have meant to have a good look at this, but have had some family stuff to take care of. 
My initial read of the PR seems to look that it's all OK, but would like to just do a few more tests.

mc12345678:
No issues, I actually figured it was a desire to hang back a little until it seemed like I was done.

Let me know if it would be preferred to compress the commit or otherwise split out any other features. (ie. modification to the custom method or other items).

drbyte:
> No issues, I actually figured it was a desire to hang back a little until it seemed like I was done.

There's definitely that too ;)

zcwilt:
are you done :)

mc12345678:
I am done playing with the code the only change I was potentially considering was if you wanted me to squash the commits into one.

I've not identified any issues with the way this processes, the ability to not strictly sanitize is present, and all sanitize options seem to work as might be expected.

drbyte:
>  want me to squash the commits into one?

Yes, if you're comfortable doing that, please do. 
(summary: squash with a rebase, then force-push the new commit back to this branch)

mc12345678:
Be a few hours, but not a problem and glad to.

mc12345678:
Hrmm.. :/ looks like there is a potential code improvement (reduction of code duplication) if:
````php
            if (!isset($this->arrayName) || $this->arrayName == '') {
                $this->arrayName = $parameterName;
            } else {
                $this->arrayName = $this->arrayName . '[' . $parameterName . ']';
            }
````
were incorporated as a separate internal function possibly identified as `function updateArrayName($parameterName)` with whatever value is needed to perform the update sent to the function.

Thoughts?

BTW, I saw the extra `}` that existed within `function updateSimpleArray` and corrected it.

**Comment by drbyte** [comment]
A couple additional ways to simplify it include:

```php
if (empty($this->arrayName)) {
                $this->arrayName = $parameterName;
            } else {
                $this->arrayName .= '[' . $parameterName . ']';
            }
```

or 
```php
$this->arrayName = (empty($this->arrayName)) ? $parameterName : $this->arrayName . '[' . $parameterName . ']';
```

mc12345678:
Thought about using empty, but if the first "element" or tag name was "0" then on the next sanitization the first level would be lost on the second "use" as `empty($this->arrayName)` would evaluate as true (empty) and `$this->arrayName` would get reassigned to the new/next $parameterName.

drbyte:
Argh, ya, that's the one edge case where that falsy response kills the simplicity.

zcwilt:
Another simplification would be to remove the else clause

e.g.

    
     $result = $this->arrayName . '[' . $parameterName . ']';
     if (!isset($this->arrayName) || $this->arrayName == '') {
          $result = $parameterName;
     }
     $this->arrayName = $result;

also have no problem with making it a small function and personally prefer setCurrentArrayName

zcwilt:
btw if turning this into a function I would pass in both this->arrayName and $parameterName as  function parameters, and then do 
$this->arrayName = setCurrentArrayName($this->arrayName, $parameterName) 

and the function would then look like 

    function setCurrentArrayName($currentArrayName, $parameterName) 
    {
         $result = $currentArrayName . '[' . $parameterName . ']';
         if (!isset($currentArrayName) || $currentArrayName == '') {
              $result = $parameterName;
         }
        return $result;
    }

zcwilt:
Doh.
Just realized that function won't work well if $this->arrayName is not set.
Just need to flip the test 

    $result = $parameterName;
    if (isset($this->arrayName) && $this->arrayName != '') {
        $result = $this->arrayName . '[' . $parameterName . ']';
    }

mc12345678:
Darn, there appears to also be three different styles of the operation, been trying to determine the best way forward with that.  I had already determined that if a function would be used that `setCurrentArrayName` would be the preferred name of the two listed, but was trying to come up with preferably one function that would be easy to understand for new filters/users.  As noted and independently tested, a variable would have to be set/declared in order to pass it to the function (or the function receive the variable by reference).  While it is not an issue with the existing code (it is always set even if to empty quotes), but there is a sort of issue of consideration for future use/modification.  

Unless considered/identified as a practice of concern, the `$currentArrayName` could be received by reference (`&$currentArrayName`) and still be tested for being set or not and the calling use would not error (online tested in PHP 5.2.17 through PHP 7.1). Yesterday was considering some different options: a potentially "complicated" function passing in things like a modifiable "result" variable as well as the two things to use for "addition" to allow the function to perform all of the operations internally and the associated data modified back in the calling function, a function that just addresses $this->arrayName with a function that handles the other two conditions via a return value, a single function that returns a value without modifying anything in the class, and/or trying to see if the operations can be rewritten to simplify the process using the above:

"Styles":
1. (Modifies `$this->arrayName` based on `$this->arrayName`) This could be used with a single parameter to a single function regardless the state of `$this->arrayName` or to a function that accepts the `$currentArrayName` by reference and accepts the `$parameterName` with a return result to be assigned to `$this->arrayName`.
````php
           if (!isset($this->arrayName) || $this->arrayName == '') {
               $this->arrayName = $parameterName;
           } else {
               $this->arrayName = $this->arrayName . '[' . $parameterName . ']';
           }
````
2. (Modifies a local variable based on the local variable) would be best either to use/switch to an alternate internal variable to allow temporary modification or to be in a single function that accepts `$currentArrayName` when it is not set by receiving by reference (`&$var`) with the result sent back to the function to be assigned to `$currentArrayName`.
````php
       if (!isset($currentArrayName) || $currentArrayName == '') {
           $currentArrayName = $parameterName;
       } else {
           $currentArrayName = $currentArrayName. '[' . $parameterName. ']';
       }
````
3. (Modifies $this->arrayName to be based on a local variable) This is a combination of types 1 and 2, but could be accomplished with a single function that permits $currentArrayName to be undefined which is possible if received by reference. 
````php
           if (!isset($currentArrayName) || $currentArrayName == '') {
               $this->arrayName = $k;
           } else {
               $this->arrayName = $currentArrayName . '[' . $k . ']';
           }
````
Conditions 2 and 3 could presumably be performed with a single function and removal of the `!isset` condition with `$currentArrayName` received **not** by reference.  Expectation (as would be described in the function notes) is that the build on data or base `$currentArrayName` would be set prior to use of the function even if set to empty quotes. otherwise the function would result in an error by trying to pass an undefined variable. 

Though the proposed single function would work if the variable was received by reference (`&$currentArrayName`) even if the value is not purposefully modified in the function.

Based on the recommendations provided and barring a concern of using the by reference use, looks like the preferred method would be to return the value that is to be set to the applicable result and allow the local function to identify what value is to be modified.

So function to incorporate would be:
````php
private function setCurrentArrayName(&$currentArrayName, $parameterName)
{
    $result = $parameterName;

    if (isset($currentArrayName) && $currentArrayName != '') {
        $result = $currentArrayName . '[' . $parameterName . ']';
    }

    return $result;
}
````
Then in the places where the three styles above are used the existing left side assignment would equal the function result. I.e.:

1.
````php
$this->arrayName = setCurrentArrayName($this->arrayName, $parameterName);
````
2.
````php
$currentArrayName = setCurrentArrayName($currentArrayName, $parameterName);
````
3.
````php
$this->arrayName = setCurrentArrayName($currentArrayName, $k);
````

**Comment by zcwilt** [comment]
So A bit confused.
Why the need to receive $currentArrayName by reference. Would switching the parameters around  
private function setCurrentArrayName($parameterName, $currentArrayName = null)
be better 

regardless, receiving $currentArrayName by reference seems to complicate things, not a fan of doing 'magic', regardless of the fact there is still a lot of that in ZC :(

**Comment by zcwilt** [comment]
btw I do understand your concerns, especially in terms of adding functions that help noob users. However, we have done that in ZC in the past. 
e.g functions that take multiple parameters and the do lots of different things. In the end it really does complicate stuff.

**Comment by mc12345678** [comment]
> So A bit confused.
Why the need to receive $currentArrayName by reference. Would switching the parameters around
private function setCurrentArrayName($parameterName, $currentArrayName = null)
be better
>
>regardless, receiving $currentArrayName by reference seems to complicate things, not a fan of doing 'magic', regardless of the fact there is still a lot of that in ZC :(

In testing, it didn't matter what the "default" value was when attempting to call the function, if the variable was not declared then the php was failing at the point of making the call not at the point of receiving the information.  By passing by reference, the error did not occur and the resulting array was built as expected for either case of not defined or having existing data.

**Comment by zcwilt** [comment]
Again, if users are going to call the function knowing that $currentArrayName does not exist, then would prefer the option of switching parameter order to set it to null in the function call 
private function setCurrentArrayName($parameterName, $currentArrayName = null)

where they would then only pass in the $parameterName

**Comment by mc12345678** [comment]
>Again, if users are going to call the function knowing that $currentArrayName does not exist, then would prefer the option of switching parameter order to set it to null in the function call
private function setCurrentArrayName($parameterName, $currentArrayName = null)
>
>where they would then only pass in the $parameterName

Yes, now that makes sense. Will switch to support purposefully formatting `$parameterName` in absence of a `currentArrayName`.

**Comment by zcwilt** [comment]
I notice you are still receiving the `$currentArrayName` as a reference in your function parameters.

e.g.

    private function setCurrentArrayName($parameterName, &$currentArrayName = NULL)


My previous posts were meant to take us away from receiving by reference.
Will try and explain my reasoning a bit more succinctly.

As I see it, a function signature is a contract. It's telling a developer exactly what we expect.

if I have a function like 

   private function setCurrentArrayName($parameterName, $currentArrayName)

then I am telling other developers that they MUST pass in 2 parameters and both those parameters must have a value.

If the values they are trying to pass in don't have values, then they either have to set some defaults before passing the values in, or just not use this function.

With the signature   

    private function setCurrentArrayName($parameterName, $currentArrayName = null)

The contract is now that you can either pass in both parameters, and both those parameters have a sane value, or if you know you don't have a sane value for `$currentArrayName`, don't pass anything in 

e.g. 

    $result = $this->setCurrentArrayName($parameterName, $currentArrayName)

or 

    $result = $this->setCurrentArrayName($parameterName)

Also as a developer, when I see a function receiving a parameter by reference, what do I expect that function to do. Personally I would assume that the function is going to manipulate that parameter and change it within the function, as that is what referenced parmeters are meant to do. If the function does not do that then it's breaking the contract that I expect.


BTW I am pretty much happy with the rest of the PR, and while I may be coming across as a bit anal/pedantic around the new function, I do want this PR to get merged.

**Comment by mc12345678** [comment]
As might be noticed in the forum, I am certainly not against discussion at any/all levels. So, my initial application/usage of the by reference passing was a mix of trial and error and suggestion. I normally wouldn't have even considered it if there hadn't been a discussion of failure by directly passing an unset variable which would then require an inplace repeated series of logic. 

So at the time to quickly and easily diagnose I took to one of those online php testing platforms. Supplied the function, then started using different initial values and in some cases non-values. Mind you I didn't initially have clean display operations to where each value to be presented was on its own line. How hard can it be to tell the difference between one thing and the next in a display of maybe three things?

Standard cases working fine, move to an edge case, although had a value for the currentarrayname, I've now unset it. Sending the unset value, with currentarrayname either first or later having a default value to support not sending it at all, I get a long message and what appears to be a non-display of any information related to the parameter that was sent in. Ie. It looks like the function has completely failed and user would be stumped as to what has happened, what's going on. 
Add a by reference call, and we'll, even without the the direct manipulation of currentArrayName like one might expect because of the nature of its common use, the function itself returns the "modified" value, but... the one thing about it is basically whatever currentarrayname is, it's condition/value has been brought into the function's space rather than to provide message from the calling location.

Ok, so there's that part of discussion, but... a difference in environment is ultimately was caused the current solution.

The long message was a notice. And as such, the function did execute without trouble and the value that would be expected by passing in a null to the currentarrayname would have provided the same effect (without the message). ZC basically makes life easier for the base user by at least ignoring notices, which if the code had been tested in a standard ZC install wouldn't have appeared at all and we wouldn't be discussing (I wouldn't be trying to learn this little nugget). 
So, notices: when to accept them occurring in the code, when to modify the code to prevent them, and to what extreme? I mean in a way I recognize that by directly preventing the notice from being generated the only way someone would know that they were trying to pass a truly undefined value to the function by code "monitoring" would be to insert some sort of variable recording/display either just before sending to the function or within it (modification of it) basically. The removal of the by reference in a standard ZC would cause no additional "messaging/logging" because notices would be ignored, it would require an alternate setting to allow notices to be logged, but then the question is how important in a way is that notice considering inside the function is a test for the value to determine if it is set which is intended to address the condition that the variable was "provided" but the variable has not been defined, ie. the function is intended to test and handle the case that is causing the notice in the first place, but nothing is preventing the generation of the notice other than basically ignoring it. 

I'm cool with allowing the generation of the notice that is basically set to be ignored, I thought in reading about how some go through and inspect the ZC code that I might try to keep the function off the radar so-to-speak even if the issue were found by "accident". 

I don't consider the explanation to be anything more than a means to express understanding and possibly put across a core concept. I hope that response is not "because I told you to" :)

**Comment by mc12345678** [comment]
FWIW, I do have a local commit that removes the by reference and instead of using `isset` for testing uses `!is_null` which without the by reference pass seems more appropriate. 

Would still be curious about the idea of handling/preventing a notice unless the notice (for ZC general building purposes) in this case would be akin to a pre-planned debug log for the case when a non-defined variable is attempted to be passed to a function.

**Comment by drbyte** [comment]
I firmly believe one should code so that no `notice` or `warning` or any other messages are triggered under normal circumstances, and if certain cases (such as relying on code from other developers/contributors, including plugins) could cause issues, then code defensively so that they can be informed of the problem.

Every time PHP has to trigger a problem (whether `notice` or otherwise), even if the `error_reporting` level is set to exclude it, it slows things down. 

Developers should code with `error_reporting(E_ALL)` set so that they deal with all issues caused by their code. In production certain error levels are suppressed because you don't want to interrupt the end-user's experience, and that's why possible problem scenarios should be coded-for.

One should also be careful to not intentionally code in such a way that unnecessary "debug logs" are generated simply to point out a minor information point. And also should particularly avoid repeat logs that would trigger thousands of times a day. Failing gracefully is always preferred over millions of duplicate info-log messages.

---

Your stumbling upon using pass-by-reference as a possible workaround for a separate limitation is good in itself, as it helps expose more information to you in your decision-making process while coding. That doesn't always mean it's the right solution, but it does add helpful information on-the-journey to the right solution.

As discussed earlier, in a not-completely-object-oriented code base, PHP lacks an elegant mechanism to handle a combined test of "is set + and is not null + and is not falsey + and is not intentionally '0' (whether int or string)". 
That said, the fact that this problem exists may be exposing an opportunity for additional architectural improvements. 

And then that brings up the question of whether this particular section of code (both before and following your PR) will be relevant for v160, or require significant overhaul to suit the architectural changes therein.

**Comment by mc12345678** [comment]
So first want to address the last: 
> And then that brings up the question of whether this particular section of code (both before and following your PR) will be relevant for v160, or require significant overhaul to suit the architectural changes therein.

I'm certainly under no false impression that any of this is expected to last as is.  It is proposed as a solution to the problem observed in the current state with the current method of sanitization.  When it was first announced I was a little surprised at its inclusion, but the reasoning behind addressing it was sound, so the results of the process are seen.  How it is accomplished and whether any effort associated with it will be deconstructed/revamped is the nature of accomplishing the goal.

I reread the above multiple times applying emphasis at different parts of the discussion.  Taking in one of the last revision suggestions and the above philosophy, the two of you have led me to revisit the need for having three different styles of building the resulting arrayName.  I.e.:
>  Just realized that function won't work well if $this->arrayName is not set.
Just need to flip the test
>
> ````php 
> $result = $parameterName;
> if (isset($this->arrayName) && $this->arrayName != '') {
>     $result = $this->arrayName . '[' . $parameterName . ']';
> }
> ````

The above actually made me think that the suggestion was to not include a function and simply replace all previous usage of the series of code with the above in one way or another (substituting `$result` as necessary).  Maybe that was the direction headed at the time, but I was determined to make a function that would add support and centralize the operation without too much confusion.

At any rate, while I think I have found a solution to our issues, I think I need to again test the simple array, the strict sanitize value, and multi_dimensional functions based on the potential solution.

Basically, I think I have identified how all calls to `setCurrentArrayName` can be made such that the result is built off of `$this->arrayName` and therefore there is no need to pass the "base" value to the function and that it would only need to accept the one parameter `$parameterName`.  This is perceived as allowing `isset($this->arrayName)` to remain (not changed to `!is_null`) and to eliminate the notice response of passing an undefined value to the `setCurrentArrayName` function.  It doesn't prevent or address attempting to pass `$parameterName` as an undefined variable, but that seems to be a condition that would be identified as a problem by the one writing new code to use the function when reviewing the results (either the sanitizer debug logs or the mydebug logs themselves). I mean, unless there is a thought that such a condition when identified within the function warrants a response action (debug log) because basically what I see as potential outcome is that either `$result` will remain empty or it would have an empty sub-array (`value[]`).  The empty sub-array condition seems to be addressed when strict sanitization is in place for both the simple array and multi-dimensional conditions by not building on the empty key, but is not addressed elsewhere.

Thoughts?

Biggest take away by me, thanks for the subtle beating. :) I think that if the additional testing (created a [temporary branch](https://github.com/mc12345678/zc-v1-series/commit/745414fd00f97e984fdc8caabe20e430c99d9555) to test the latest commit prior to squashing, rebasing and force-pushing) proves successful then the result is more inline with what the group has discussed (no magic, elimination of earlier possible notice generation, reusable code, general template for other future similar options, etc...). More comment is welcome.

Btw, I noticed that there are class helper functions such as `setPostKeyAlreadySanitized` and `setGetKeyAlreadySanitized` that aren't used in the code.  I almost incorporated the applicable into `filterSimpleArray` in this latest edit/update to attempt to more closely approach OOP, but wasn't sure if the thought was that those shouldn't be used, that they were expected to be but in the large amounts of changes that had occurred they were overlooked, or for some other unthought of reason not used.  I had also begun to incorporate their use throughout the code but thought that without discussion it would just be something "else" in the way of potentially incorporating this change.

File: admin/includes/classes/AdminRequestSanitizer.php [hunk 2]
```diff
@@ -411,15 +426,18 @@ private function filterProductDescRegex($parameterName)
     {
         $prodDescRegex = '~(load=|= ?\(|<![^-])~i';
         if (isset($_POST[$parameterName])) {
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName, $this->arrayName);
             $this->debugMessages[] = 'PROCESSING PRODUCT_DESC_REGEX == ' . $parameterName;
             if (is_array($_POST[$parameterName])) {
                 foreach ($_POST[$parameterName] as $pKey => $pValue) {
+                    $currentArrayName = $this->arrayName . '[' . $pKey . ']';
```
mc12345678: Argh, an instance of needing to substitute the function to build `currentArrayName`.  Amazing what is discovered when changes are colorized.

admin/includes/classes/AdminRequestSanitizer.php
```diff
@@ -430,10 +448,13 @@ private function filterProductDescRegex($parameterName)
     private function filterMetaTags($parameterName)
     {
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING META_TAGS == ' . $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName, $this->arrayName);
+            $this->debugMessages[] = 'PROCESSING META_TAGS == ' . $this->arrayName;
             foreach ($_POST[$parameterName] as $pKey => $pValue) {
+                $currentArrayName = $this->arrayName . '[' . $pKey . ']';
```
mc12345678: Here again.

admin/includes/classes/AdminRequestSanitizer.php
```diff
@@ -477,14 +500,17 @@ private function filterProductUrlRegex($parameterName)
     {
         $urlRegex = '~([^a-z0-9\'!#$&%@();:/=?_\~\[\]-]|[><])~i';
         if (isset($_POST[$parameterName])) {
-            $this->debugMessages[] = 'PROCESSING PRODUCT_URL_REGEX == ' . $parameterName;
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName, $this->arrayName);
+            $this->debugMessages[] = 'PROCESSING PRODUCT_URL_REGEX == ' . $this->arrayName;
             foreach ($_POST[$parameterName] as $pKey => $pValue) {
+                $currentArrayName = $this->arrayName . '[' . $pKey . ']';
```
mc12345678: And again.

admin/includes/classes/AdminRequestSanitizer.php
```diff
@@ -523,7 +553,11 @@ private function filterMultiDimensional($parameterName, $parameterDefinition)
         if (!isset($requestPost[$parameterName])) {
             return;
         }
+        // Add the parameterName to the base arrayname.
+        $this->arrayName = $this->setCurrentArrayName($parameterName, $this->arrayName);
+        $currentArrayName = $this->arrayName;
```
mc12345678: This and the above line can be combined into a single statement with the far left being `$currentArrayName` which will be assigned the value to `$this->arrayName` which will be assigned the result of `$this->setCurrentArrayName`. Perhaps though so many variables aren't necessary?

admin/includes/classes/AdminRequestSanitizer.php
```diff
@@ -535,22 +569,98 @@ private function filterMultiDimensional($parameterName, $parameterDefinition)
...
+                if (array_key_exists($pkey, $_POST)) {
+                    $this->filterStrictSanitizeValues();
+                    $newKey = $_POST[$parameterName];
+                    $this->arrayName = $newArrayName;
+                    $requestPost[$parameterName][$newKey] = $hacked;
+                } else {
+                    continue; // Key is "unclean" and therefore should use the next key.
+                }
             }
+            $newArrayName = $newArrayName . '[' . $newKey . ']';
```
mc12345678: Arghh.. Another version of the assignment with this being a manual version that did not fit the previous pattern for replacement which means more revision in this area to make sure it works as expected/desired when/if replaced with `$newArrayName = $this->setCurrentArrayName($newKey)`.

admin/includes/classes/AdminRequestSanitizer.php
```diff
@@ -560,10 +670,13 @@ private function filterProductNameDeepRegex($parameterName)
     {
         $prodNameRegex = '~<\/?scri|on(load|mouse|error|read|key)(up|down)? ?=|[^(class|style)] ?= ?(\(|")|<!~i';
         if (isset($_POST[$parameterName])) {
+            // Add the parameterName to the base arrayname.
+            $this->arrayName = $this->setCurrentArrayName($parameterName, $this->arrayName);
             $this->debugMessages[] = 'PROCESSING PRODUCT_NAME_DEEP_REGEX == ' . $parameterName;
             foreach ($_POST[$parameterName] as $pKey => $pValue) {
+                $currentArrayName = $this->arrayName . '[' . $pKey . ']';
```
mc12345678: For some reason I think I did this on purpose; however, because I don't know and didn't document the specific purpose, it seems like this should be rewritten to use the above `$this->setCurrentArrayName` function to assign the value back to `$currentArrayName`.

admin/includes/classes/AdminRequestSanitizer.php
```diff
@@ -588,19 +704,33 @@ private function filterStrictSanitizeValues()
...
+                        }
+                    }
                 }
             }
             if (!$inner) {
                 if ($type == 'post') {
-                    if (!in_array($k, $this->postKeysAlreadySanitized)) {
-                        $this->postKeysAlreadySanitized[] = $k;
+                    if (!in_array($this->arrayName, $this->postKeysAlreadySanitized)) {
+                        $this->postKeysAlreadySanitized[] = $this->arrayName;
```
mc12345678: Why in filter's such as this is `setPostKeyAlreadySanitized($this->arrayName);` not used instead of manipulating the array directly? I some what get why `addKeyAlreadySanitized` wasn't selected to be used (if the key were already present then it wouldn't be added an additional time and it seems like it is desired to allow it to be added multiple times in the place(s) incorporated).

zcwilt:
I guess my take on this at the moment is that the commit you have in your [temporary branch](https://github.com/mc12345678/zc-v1-series/commit/745414fd00f97e984fdc8caabe20e430c99d9555) 
is fine, and would be happy to merge as is. 

Agree that there is probably further work/optimisations that can be made in the future,

mc12345678:
While I'd prefer the other things worked out before committing the changes, I understand and agree that on the whole it works but could use a few more tweaks. Will do the above to create a single commit against the current ZC 1.5.6 branch I expect later today.

drbyte:
@mc12345678 and @zcwilt 
The last comments suggested this may still be incomplete. So, checking in to ask both of you about the status of this PR.

mc12345678:
I pushed what appeared to functionally test. While there could always be more done, I believe the changes accomplished the desired goal and offered some improvement to the overall process. I have no further modification to it at this time.

zcwilt:
:+1:

drbyte:
Okay. 
👍 

Thanks for all the work on this.

PR #1525, opened 2017-10-06
redirects non-https pages to https when whole site is https

proseLA:
i am proposing no http landing pages when the whole site is set for https.  this would eliminate the need for store owners to create apache redirects in the .htaccess file.

unfortunately, i was not able to see how to get the autoloader on the admin side make use of the init_script on the store facing side of the code.  if somone were to enlighten me, i will modify this PR accordingly.


dakanji:
Appreciate the sentiment here but personally think the whole SSL and NON SSL page infrastructure, a hold out from the early days of the program when SSL was expensive both in terms of fees and server load, should be removed entirely.

Ideally, web applications should simply provide content in an agnostic manner and not try to do the job of webservers. So will suggest there should be a way to simply bypass the whole thing and just let the webserver handle what http protocol to use.

Until not long ago with Joomla for instance, when you have a site that simply turns all pages into https, common now with letsencrypt, the program  would basically "fight" you as it tried to do all kinds of things ... especially when not using Apache. 

See my SO post: https://stackoverflow.com/a/41299376/891636

mc12345678:
May also benefit to identify the thread at which this discussion/concept initiated (last): https://www.zen-cart.com/showthread.php?222916-Proposal-use-PHP-to-redirect-to-https

mc12345678:
> So will suggest there should be a way to simply bypass the whole thing and just let the webserver handle what http protocol to use.

If by this meaning to allow the webserver to address whether to use https: instead of http:, the webserver would still be able to do a redirect prior to ZC attempting to take any action.  The primary issue (with the provided implementation) is that the software could be put into a loop because of an incorrectly configured server and there is no direct logic to disable the feature.  It would require ftp or similar process to alter the file loading sequence or remove the applicable file(s).

As far as "tight" control ie, keeping things within not only a specific protocol (http instead of ftp or other method) but also within some "group" of that, well attempts have been made for example in the admin for the admin side to use one server communication method rather than switching between definitions, on the catalog side direction had been provided to remove any such "switch" from one to the other, but not all of the standard code within ZC had been updated to support so that guidance appears to have been reverted.  Basically, I see this aspect as one of the features that has allowed ZC to provide the level of security that it has especially when compared to other systems.

drbyte:
> Ideally, web applications should simply provide content in an agnostic manner and not try to do the job of webservers. 

Agreed. If only webservers were trustworthy and configurable by the layperson. The average small storeowner has no technical foo, and so they buy shared-hosting, which as you know is configured for the lowest-common-denominator, and in practice cannot be configured easily by the layperson.

> So will suggest there should be a way to simply bypass the whole thing.

There is. Almost. Simply define both HTTP_SERVER and HTTPS_SERVER to the correct 'https' URL for the domain. That will stop ZC from switching in-and-out-of-https.

The only remaining "configuration" after that (which this PR attempts to handle) is basically the same as HSTS but in PHP. (ie: rewrite incoming http URIs to https before passing along to the application). 
I agree that it's better to configure the server to do it, and would prefer that that approach be taken. Unfortunately every shared-server is configured differently, and different rewrite rules are required in each case. In the absence of reliable support staff, with Apache at least .htaccess can be played with and updated by the end-user with some ability to figure out what works. With nginx, well, the average user has no access to make adjustments there.

drbyte:
As to this PR, I'm still undecided where I stand on implementing it as a default. 
Might end up merging it but disabling it as an optional thing one can quickly re-enable if the preferred approach of configuring at the server level is deemed impossible.

proseLA:
interesting discussion.

first, i still do not see an answer to whether the admin can load the same init_script as the customer facing.  if not possible, i think we should change that.

second, i agree with discussion with regards to shared hosting, configuration, technical expertise of store owner, etc.  the question remains how much server config should/can ZC do when considering where the majority of ZC installs are hosted?

third, perhaps adding another config switch is the answer, but perhaps we need to re-think the configure.php file.  current practice of setting HTTP_SERVER to https is so counter-intuitive to a layperson.  i suppose there is a holdover from when the base url for HTTPS_SERVER was different from HTTP_SERVER, but are there many of those installs still out there?  and is there a need to support them?  (or am i wrong and is there another reason for having 2 defines in the configure.php file?)  we currently do not have separate defines on the admin side.

best.

admin/includes/auto_loaders/config.https.php
```diff
@@ -0,0 +1,3 @@
+<?php
+$autoLoadConfig[15][] = array('autoType' => 'init_script',
+                              'loadFile' => 'init_https.php');
```
drbyte: To share it from catalog side:
```diff
-                              'loadFile' => 'init_https.php');
+                              'loadFile'=> DIR_FS_CATALOG . 'includes/init_includes/init_https.php');
```

includes/init_includes/init_https.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+if (((HTTPS_SERVER == HTTP_SERVER) || (IS_ADMIN_FLAG)) && (strpos(HTTP_SERVER, 'https:') !== false)) {
```
drbyte: Since Admin doesn't technically define HTTPS_SERVER, switching the order of these would save a constant-not-defined error:

```diff
-if (((HTTPS_SERVER == HTTP_SERVER) || (IS_ADMIN_FLAG)) && (strpos(HTTP_SERVER, 'https:') !== false)) {
+if ((IS_ADMIN_FLAG || HTTPS_SERVER == HTTP_SERVER) && (strpos(HTTP_SERVER, 'https:') !== false)) {
```
Also, removing unnecessary parentheses (as shown) makes it more readable.

drbyte:
> third, perhaps adding another config switch is the answer, but perhaps we need to re-think the configure.php file. current practice of setting HTTP_SERVER to https is so counter-intuitive to a layperson. i suppose there is a holdover from when the base url for HTTPS_SERVER was different from HTTP_SERVER, but are there many of those installs still out there? and is there a need to support them? (or am i wrong and is there another reason for having 2 defines in the configure.php file?) we currently do not have separate defines on the admin side.

Perhaps, but that's a separate topic.

dakanji:
> Simply define both HTTP_SERVER and HTTPS_SERVER to the correct 'https' URL for the domain. That will stop ZC from switching in-and-out-of-https.

Fair enough but ...

> The average small storeowner has no technical foo, and so they buy shared-hosting, which as you know is configured for the lowest-common-denominator, and in practice cannot be configured easily by the layperson.

Perhaps an option is to get users to use htaccess as Apache is the default supported web server and htaccess readily available on shared hosting. Wouldn't suggest trying for Nginx users.

Maybe even provide the htaccess (with a .txt extension for instance) which users can rename to get this if needed similar to many open source projects.

Something like this in the ZC root perhaps?

```
RewriteEngine On
RewriteCond %{HTTPS} off
RewriteRule ^(.*)$ https://%{HTTP_HOST}%{REQUEST_URI} [L,R=301]
```

Should then be possible to strip out the whole infrastructure from the codebase .... just a thought

dakanji:
> question remains how much server config should/can ZC do

Can see that is something the dev team juggles. Always a challenge to draw the line at an appropriate point. Suppose it is something that needs to be revisited now and then to make sure what is being done is at an appropriate level ... not too much and not too little.

admin/includes/auto_loaders/config.https.php
proseLA: not true.  as i suspected there is no way to share an init_script from the customer side to the admin side.  you must change the autoType.  from autoload_func.php:

      case 'init_script':
      $baseDir = DIR_WS_INCLUDES . 'init_includes/';
      if (file_exists(DIR_WS_INCLUDES . 'init_includes/overrides/' . $entry['loadFile'])) {
        $baseDir = DIR_WS_INCLUDES . 'init_includes/overrides/';
      }
      /**
       * include an init_script as specified by autoloader array
       */
      require($baseDir . $entry['loadFile']);
      $debugOutput .= 'require(\'' . $baseDir . $entry['loadFile'] . '\');' . '<br />';
      break;


one must change the autoType to either require or include.  is there a preference?  i'mn guessing require as that is what the init_script autoType does...

drbyte:
Okay, for now, since v160 is in dev stage, gonna accept this for now, and allow for some in-the-wild feedback from devs working with it.
👍 

May still disable it by default prior to release, depending on feedback.

zcwilt:
:+1:

includes/init_includes/init_https.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/** 
+ * Do automatic redirect of http URIs to https, if https is given in HTTP_SERVER in configure.php
+ *
+ */
+
+if ((IS_ADMIN_FLAG || HTTPS_SERVER == HTTP_SERVER) && (strpos(HTTP_SERVER, 'https:') !== false)) {
```
mc12345678: I've been trying to find a way to express the below based on previous conversations. I've spent too much time worrying about that so here's my thoughts on the structure of the check(s) to lead to the redirect action.

Unless there is a specific reason to have the `$request_type` check occur after all of the other checks (as provided which for a site that is being accessed via https would be performed on each and every page load), then would suggest moving the last `if` up to the beginning of the first `if`.  Then, with regards to the `HTTPS_SERVER` to `HTTP_SERVER` comparison, this seems like it is only important if `ENABLE_SSL` is set to `true` otherwise in theory nothing else would need to access/use `HTTPS_SERVER` and related constants.  Then would also need to check for the condition where `ENABLE_SSL != 'true'` and if all of those things are `true` by this point, then seems should be checking for https: to begin the `HTTP_SERVER` constant, not just be somewhere in the string. (could have an internal server configuration that places some variation of https at the end of the string and use a port such that the string would end with '...https:XXX' but yet begin with http:).  All this making the above line:
````diff
- if ((IS_ADMIN_FLAG || HTTPS_SERVER == HTTP_SERVER) && strpos(HTTP_SERVER, 'https:') !== false) {
+ if ($request_type !== 'SSL' && (IS_ADMIN_FLAG || ENABLE_SSL == 'true' && HTTPS_SERVER == HTTP_SERVER || ENABLE_SSL != 'true') && strpos(HTTP_SERVER, 'https:') === 0) {
`````

drbyte: I think I understand your thinking here, and agree with the suggestion to rearrange the order of things. 

But I wonder whether the check for `ENABLE_SSL` is necessary.  
My personal preference is to dump that define altogether: the thinking being that since the google gods are driving the web industry to use https everywhere, then there's no need to even use ENABLE_SSL at all.  And probably no need for HTTPS_SERVER either.  A single define of HTTP_SERVER would suffice for the entire site. There'd be no need for switching between http/https either. 
I don't see the need for supporting these extra defines in either production or dev environments. 

Apart from backward-compatibility, is there a technical or business reason to keep them at all?

mc12345678: The only thing that seems to come to mind and likely should be something independently addressed outside of the core of ZC would be some sort of "alternate" server address from which "non-https" content (under the guise of either using the alternate define of what the main store uses and then using or not using https) could be pulled/obtained. I have no such need nor have I worked with anyone that has shown such a need. Other plugins have offered "remote" server related code for those that say split their images off to another server.

To me it would seem that if the thought is all https: or all not as appears to be the direction of the pendulum, that there is no future need for a second define nor an indicator of the ability to switch and from that a need to switch to/from the same server but via a different "mode" of communication.  (ie. be gone HTTPS_SERVER and ENABLE_SSL) I don't know what forum of nor group of people need to be canvassed to elicit more of a technical/business need response.  

Historical conversations were that encrypting added a system burden (burden = slowdown = reduced rankings) and that where there wasn't a need for some form of increased data protection then don't use https. That's perhaps the old mantra and ZC has been handling switching between the two states well, but as pointed out and as the discussions over the last couple of years have gone, the way suggested to go is all one way all of the time with "preference" of https. 

Makes me wonder if it is worth digging into all of the code to remove the flips and flops or would a few key extractions address the issue and allow someone to revive it if the need resurfaced? Ie. bite the bullet or hoard?

PR #1530, opened 2017-10-19
v160 image handler

lat9:
Applies the same notifier updates to the v160 branch as previously pushed to v156, allowing Image Handler and any large-image plugins (e.g. Fual Slimbox, ColorBox) to also benefit from the change.


drbyte:
👍

zcwilt:
:+1:

PR #1535, opened 2017-10-25
Coupon link not https/mixed content

mc12345678:
When a coupon has been entered,
the store is not 100% https (HTTP_SERVER is not https:),
and the checkout_payment page is viewed, the page link for the coupon
information (popup link) is generated as http: only and can cause mixed
content warnings.


drbyte:
👍

zcwilt:
:+1:

PR #1536, opened 2017-10-25
Improve handling of zero quantity added to cart

### LINKED ISSUE #1364: Quantity of 0 added to cart reports successful addition
From the product information page, ZC defaults to add some quantity of product to the cart.  Modifying the quantity to 0 or blank requires user action.  The response when clicking add to cart is that the product has been successfully added to the cart; however, by going to the cart, the cart is empty (sure zero items added to zero items is still addition, but...).  To back up one step, this can be seen when the setting is to not go to the cart after adding a product.

In comparison, selecting the add selected items to the cart option from a category listing with no product quantities entered will on the same store setup take the customer to the shopping cart where the cart is identified as being empty.

I am totally open to suggested/desired cart response and am willing to put forth a pull request to address.  

My initial thought regarding a 0 quantity added to the cart from the product information page was to provide a "unique" message; however, when attempting to add a negative quantity as a test, the message related to a unit error.  As such instead of generating a new unique message, I considered using the same one as it seems to apply.  Adding a new message well, allows easier future adaptation, but also adds at least another constant to the defines.

How was this discovered? Well, in updating the plugin [Product Attribute Grid](https://www.zen-cart.com/downloads.php?do=file&id=618), the operation of the add-to-cart code was mirrored.  As a result, if no quantity of attribute identified product is entered while on the product information page and the add-to-cart button is pressed, then the response is the same as the default ZC store (successfully added to cart, but there is nothing that has been added).  I plan to modify that code to account for a response similar to what is suggested here, but why only fix that plugin and not offer/suggest an equivalent solution/result for the overall project? :)

mc12345678:
Fixes #1364.

Adds a message for a zero (or empty) quantity being added to the cart which is most important for a store that does not display the shopping cart after adding product to the cart because the `actionAddProduct` function will display success for such a condition.  This also added a check of the `$_POST` array for existence of a key `shopping_cart_zero_or_less` such that if the key exists and is not set to `true`, then the success message will still be displayed.
- If the store is set to show the shopping cart after adding a product:
  - If the `$_POST['cart_quantity']` entered is non-numeric or is less than or equal to zero and `$_POST['shopping_cart_zero_or_less']` is not set, then a message is displayed about an error about adding the product.
  - If the `$_POST['cart_quantity']` entered is non-numeric or is less than or equal to zero and `$_POST['shopping_cart_zero_or_less']` is set to anything then no message is displayed.

- If the store is set to **NOT** show the shopping cart after adding a product:
  - If the `$_POST['cart_quantity']` entered is non-numeric or is less than or equal to zero and `$_POST['shopping_cart_zero_or_less']` does not exist, then a message is displayed about an error adding the product to the cart and also the success message is **not** displayed.  
  - If the `$_POST['cart_quantity']` entered is non-numeric or less than or equal to zero and `$_POST['shopping_cart_zero_or_less']` is set to something other than true, then the success message is displayed.
  - If the `$_POST['cart_quantity']` entered is non-numeric or less than or equal to zero and `$_POST['shopping_cart_zero_or_less']` is set to true, then no message is displayed about either condition.

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -1858,10 +1858,16 @@ class shoppingCart extends base {
           }
         }
       }
-      if (!is_numeric($_POST['cart_quantity']) || $_POST['cart_quantity'] < 0) {
+      if (!is_numeric($_POST['cart_quantity']) || $_POST['cart_quantity'] <= 0) {
         // adjust quantity when not a value
-        $chk_link = '<a href="' . zen_href_link(zen_get_info_page($_POST['products_id']), 'cPath=' . (zen_get_generated_category_path_rev(zen_get_products_category_id($_POST['products_id']))) . '&products_id=' . $_POST['products_id']) . '">' . zen_get_products_name($_POST['products_id']) . '</a>';
-        $messageStack->add_session('header', ERROR_CORRECTIONS_HEADING . ERROR_PRODUCT_QUANTITY_UNITS_SHOPPING_CART . $chk_link . ' ' . PRODUCTS_ORDER_QTY_TEXT . zen_output_string_protected($_POST['cart_quantity']), 'caution');
+        // If use an extra_cart_actions file to prevent processing by this function,
+        //   then be sure to set $_POST['shopping_cart_zero_or_less'] to a value other than true
+        //   to display success on add to cart and not display the below message.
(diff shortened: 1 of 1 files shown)


drbyte:
👍

zcwilt:
:+1:

PR #1538, opened 2017-10-28
coupon javascript info link type does not match page load.

mc12345678:
When a coupon has been entered,
the store is not 100% https (HTTP_SERVER is not https:),
and the checkout_payment page is viewed, the page link for the coupon
information (popup link) is generated as http: only and can cause mixed
content warnings.


zcwilt:
:+1:

PR #1542, opened 2017-11-02
Update PHPMailer from 5.2.24 to 5.2.25

drbyte:
PHPMailer has this is listed as the last official 5.2.x release. But we're retaining it here for broader backward compatibility with older PHP versions ... for now.


drbyte:
Thanks for the heads-up.

PR #1545, opened 2017-11-05
Fix HTML email messages with contents $EMAIL_MESSAGE_HTML

scottcwilson:


drbyte:
Thanks for the suggested code change.

Is this inspired from issues with a plugin such as Email Archive Viewer?
Or is it inspired from storeowners wanting alert messages to be in HTML format? 

I've always felt that alert-style messages are better in text-only format since they're less likely to get flagged as spam, and also less likely to be misunderstood as customer/sale-related instead of important internal alerts ... which is probably why this was coded to not supply an HTML message component.
Hence my questions above, to understand the context behind your suggestion.

scottcwilson:
The trouble is if you have Email Admin Format = HTML, you get this unset variable ($EMAIL_MESSAGE_HTML) when admins are removed.  I would be OK with forcing text format on specific admin notifications, but that's currently not what's happening.

drbyte:
Okay.  
Can't replicate the problem, but will merge for now.

drbyte:
👍

zcwilt:
:+1:

ajeh:
:+1:

PR #1549, opened 2017-11-09
removing obsolete menuBox function

Zen4All:
Looks like this has not been used since Zen Cart 1.0 was released, and can be safely removed
It used to build the old Osc menu structure

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1549)


drbyte:
👍

PR #1551, opened 2017-11-11
Add Notifiers associated with account_newsletters

mc12345678:
Incorporated notifiers into includes/modules/pages/account_newsletters/header_php.php to support the newsletter plugin and not further modifying this file.

3 Notifiers added: 
NOTIFY_HEADER_START_ACCOUNT_NEWSLETTERS
NOTIFY_HEADER_ACCOUNT_NEWSLETTER_UPDATED
NOTIFY_HEADER_END_ACCOUNT_NEWSLETTERS


drbyte:
👍 

NOTE: Don't count on `$newsletter_general` being a string, despite it being passed as a string currently.  It really should be an integer value, and will probably be changed to one in a future version, so ensure your plugin can accommodate it by casting to `(int)`.

zcwilt:
:+1:

PR #1556, opened 2017-11-18
Provide maximum product on shopping cart product update.

### LINKED ISSUE #1373: On cart update product quantity doesn't resolve to maximum

mc12345678:
In #1163, an issue with updating the cart was identified.

The solution to that issue (#1164) prevented assigning the maximum product
to the cart on update and instead would result in the last acceptable
quantity remaining in the cart instead of going "out-of-bounds".

This corrects the issue(s) discussed in those two threads by minorly
rearranging the `actionUpdateProduct` code to perform the cart decrease actions
before any cart increase actions and specifically for mixed product as
that is a situation where two or more cart products can/do depend on
each other.

Fixes #1373 for ZC 1.5.6. This was corrected for ZC 1.6.0 via https://github.com/zencart/zencart/pull/1374

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -70,6 +70,7 @@ class shoppingCart extends base {
    */
   var $display_debug_messages = FALSE;
   var $flag_duplicate_msgs_set = FALSE;
+  var $flag_duplicate_quantity_msgs_set = FALSE;
   /**
    * constructor method
    *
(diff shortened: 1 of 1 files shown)


drbyte:
👍 Thanks

ajeh:
:+1:

PR #1560, opened 2017-11-24
zc156-Align admin zen_get_prid/zen_get_uprid, fixes #1559.

### LINKED ISSUE #1559: zen_get_uprid functionality different admin/storefront
What the heck!  I've been chasing my tail on this for a while now, to find that the processing of the subject function is **_significantly_** different in the admin implementation (functions/general.php) from that on the storefront (functions/functions_general.php).

Admin side:
```
  function zen_get_uprid($prid, $params) {
    $uprid = $prid;
    if ( (is_array($params)) && (!strstr($prid, '{')) ) {
      while (list($option, $value) = each($params)) {
        $uprid = $uprid . '{' . $option . '}' . $value;
      }
    }

    return $uprid;
  }
```
Storefront:
```
  function zen_get_uprid($prid, $params) {
//print_r($params);
    $uprid = $prid;
    if ( (is_array($params)) && (!strstr($prid, ':')) ) {
      while (list($option, $value) = each($params)) {
        if (is_array($value)) {
          while (list($opt, $val) = each($value)) {
            $uprid = $uprid . '{' . $option . '}' . trim($opt);
          }
        } else {
        //CLR 030714 Add processing around $value. This is needed for text attributes.
            $uprid = $uprid . '{' . $option . '}' . trim($value);
        }
      }      //CLR 030228 Add else stmt to process product ids passed in by other routines.
      $md_uprid = '';

      $md_uprid = md5($uprid);
      return $prid . ':' . $md_uprid;
    } else {
      return $prid;
    }
  }
```

I'd suggest "aligning" the admin version to reflect the storefront version and potentially modifying both to remove the use to the to-be-deprecated **each()**.

lat9:
Ensure that the admin versions of the functions are the same as their storefront counterparts.

Also re-factors the functions and changes the use of the deprecated "each()" function to use their foreach() equivalent.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -609,24 +609,30 @@
       return $default_zone;
     }
   }
-
-  function zen_get_uprid($prid, $params) {
+  
+function zen_get_uprid($prid, $params) 
+{
     $uprid = $prid;
-    if ( (is_array($params)) && (!strstr($prid, '{')) ) {
-      while (list($option, $value) = each($params)) {
-        $uprid = $uprid . '{' . $option . '}' . $value;
-      }
+    if (is_array($params) && strpos($prid, ':') === false) {
+        foreach ($params as $option => $value) {
+            if (is_array($value)) {
+                foreach ($value as $opt => $val) {
+                    $uprid .= ('{' . $option . '}' . trim($opt));
+                }
+            } else {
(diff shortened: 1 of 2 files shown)


lat9:
Note that the **each()** refactoring was also applied to the storefront version of the function!

drbyte:
👍 
Thanks!

PR #1561, opened 2017-11-27
add  autocapitalize="none" and spellcheck="false" to Login fields

Zen4All:
These 2 tags will help a user when accessing the admin on a mobile phone or tablet.
Normally on these devices the first character will automatically be a capital. This is very annoying (to me) having to think about this every time I log in.
Also no spell check should be done.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1561)


drbyte:
👍 
Thanks!

PR #1562, opened 2017-11-28
Wrong place of closing bracket - breaks category search in admin

Zen4All:
The place of the closing bracket prevents search parameter from being added to the path

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1562)


drbyte:
Thanks!

PR #1563, opened 2017-11-28
Turn error reporting on if autoloader-debug is on

### LINKED ISSUE #1557: ALL ZC versions.... change of debugging....
one of the issues i have with ZC is debugging.  and in looking at application_top.php, this became a little clearer to me.

if i were to set DEBUG_AUTOLOAD to true (not something one would do on a production server), why would it not just set the error_reporting on as well?  instead, i have to remember that there is another variable there called STRICT_ERROR_REPORTING and set that also to true.  why not turn that on, if DEBUG_AUTOLOAD is already set to true?

thoughts??

proseLA:
Adds php error reporting if `DEBUG_AUTOLOAD` (used for debugging files in the `auto_loaders` folder) is turned on. 
Fixes #1557


drbyte:
Thanks

PR #1565, opened 2017-12-01
Simplify conditions in admin/product.php

Zen4All:
This piece of code can be made simpler by removing the outer 'if' end 'elseif' conditions. The only thing they really do is take a pile of stuff, and split it two piles, before making the pile even smaller.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1565)


drbyte:
👍 Thanks

PR #1572, opened 2017-12-28
Replace phplot library with flot

drbyte:
Previously the admin banner graph/stats pages relied on an outdated library phplot, which also required storing its generated graphs to the filesystem.
This puts those graphs on-screen in real-time, and removes the deprecated code dependencies.


zcwilt:
:+1:

PR #1574, opened 2017-12-28
Lengthen fields for image paths

### LINKED ISSUE #1564: Products Attributes Image length
I just ran into an issue where the product attribute image wouldn't load and it was coming back to a 404 because the image wasn't present. Upon looking at it, it appears that it was looking for something like: "attributes/revenge_of_the_nerds_2-greek_clubs-lambda_lambda_lambda.j" of course I noticed it was missing the "peg" or "pg" from the end. Which drilled down to be that, the products_attributes.attributes_image had the same value and was limited to 68.  However I'm wondering if there was something we can do? Perhaps there should be some sort of check during the upload that would either a) Fail/Reject the upload and or b) rename and notify of such change?  The other solution of course be to increase the size of the products_attributes.attributes_image field beyond the current 68, but then how long should it be?

**TEMPORARY WORK AROUND:**
For this install as a quick work around I increased the size of the field. However that isn't a current perfect solution as someone could always upload something with a super long long name and it would still show a 404 to the end customer, however it would show a good upload in the admin.
  
##  ##
  
*NOTE: The filename has been been changed to protect the innocent*

drbyte:
Allows for longer path names (or optionally URLs if plugins are doing that).

Fixes #1564

### DIFF
--- a/zc_install/sql/install/mysql_zencart.sql
+++ b/zc_install/sql/install/mysql_zencart.sql
@@ -2,10 +2,10 @@
 # * Main Zen Cart SQL Load for MySQL databases
 # * @package Installer
 # * @access private
-# * @copyright Copyright 2003-2016 Zen Cart Development Team
+# * @copyright Copyright 2003-2017 Zen Cart Development Team
 # * @copyright Portions Copyright 2003 osCommerce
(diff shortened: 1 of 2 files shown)


ajeh:
:+1:

zcwilt:
:+1:

PR #1575, opened 2017-12-28
Add helper function to limit filename length for images too long for db

drbyte:
Compliments #1574 
And addresses deeper issue in #1564


ajeh:
:+1:

zcwilt:
:+1:

drbyte:
I guess in hindsight this should be called by the Upload class so that it truncates the filename before saving ...

PR #1577, opened 2017-12-30
Under review: Align admin attributes_preview with catalog

mc12345678:
Changes have been made to the store side attributes display file but have not been carried over to the admin "equivalent".  This aligns them.

### DIFF
--- a/admin/includes/attributes_preview.php
+++ b/admin/includes/attributes_preview.php
@@ -184,11 +184,11 @@ if (!defined('IS_ADMIN_FLAG')) {
               $products_options_details = '';
             }
             if ($products_options_names->fields['products_options_images_style'] >= 3) {
-              $products_options_details .= $products_options_display_price . ($products_options->fields['options_values_price'] != 0 ? '<br />' . $products_options_display_weight : '');
-              $products_options_details_noname = $products_options_display_price . ($products_options->fields['options_values_price'] != 0 ? '<br />' . $products_options_display_weight : '');
+              $products_options_details .= $products_options_display_price . ($products_options->fields['products_attributes_weight'] != 0 ? '<br />' . $products_options_display_weight : '');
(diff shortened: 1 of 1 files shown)


drbyte:
In v160 we removed the attributes-preview altogether ... because it can't really ever display the same as it does catalog-side. 

Any objection to removing it in 156 too?

Zen4All:
😏 I already removed the link in my refactored attibutes_controller.php file

mc12345678:
Well, I had noticed that functionally the attributes_preview provided incorrect information which was some of the same incorrect information presented on the attributes_controller and the product pages but some of those issues are being resolved through #1404. That coupled with the lack of "support" by plugins to consider the potential of seeing how the attributes will be populated. (preview "format" certainly doesn't display like it does on a product page, but then again a product page "preview" as written/provided won't necessarily work either if a product to be previewed is disabled.)

I did notice the preview was removed from the attributes_controller, but didn't stop me from trying to bring the code up to snuff in absence of the link from the attributes_controller page. 

I don't know of anything specific that must have the feature, It seems though that it is being removed for the wrong reason and without an equivalent replacement/option (unless I missed seeing a link that would specifically display the product as seen on the store front.) 

If it's destined to be removed, might as well start sooner rather than later so that either those that miss it can speak up and/or find a solution that works for them. Obviously prefer something being said as can't read minds. :)

drbyte:
Agreed: removal "just because" isn't always fair. 

It was removed in 160 because of its non-parity to catalog and complete inability to ever make it be so, without refactoring it to be a modal/popup that pulls in the entire catalog-template using substituted previewable variables.  Humongous task. 

Similar reason for targeting removal of admin "preview" page for products .. something which I don't think we've committed to 160 yet. The preview doesn't serve much of a purpose, other than to be "one extra step" every time someone updates a product. (It has a technical dependency in that it facilitates uploading of files, but that can be reworked.)

PR #1579, opened 2017-12-30
V156 strict zen get products display price

mc12345678:
Strict reporting correction for functions_prices


drbyte:
I'm confused. 
Why 4 new PRs with all the same files being changed in the same ways? 

I think whatever branch you're forking from is based on something really old.

mc12345678:
Picked up from the branch on dealing with the specials, sales, and attributes which was to be based off of ZC 1.5.6. I can rebase back to ZC 1.5.6 and force push independently.

drbyte:
okay, that's probably best

PR #1580, opened 2017-12-30
typo

Zen4All:
arrya should be array

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1580)


drbyte:
Oy! 
Thanks!

PR #1592, opened 2018-01-11
[v156] Add MyAccount and ContactUs buttons to checkout-success page

drbyte:
Instead of making the language-string contain tiny text-only links, add specific obvious buttons so people know how to take action.

screen shot 2018-01-11 at 1 28 23 pm


zcwilt:
:+1:

ajeh:
:+1:

PR #1596, opened 2018-01-14
Core schema changes to accommodate utf8mb4 index length limits

drbyte:
Note: the cases where the field length was changed instead of the index are either primary keys or fields whose length can afford to be shorter.

 - [x] field lengths where shorter will suffice
 - [x] lengths on hybrid indexes
 - [x] Session Length: PHP allows for 256

Supporting 256 chars in db requires MySQL 5.0.3 or newer.

### DIFF
--- a/zc_install/ajaxLoadMainSql.php
+++ b/zc_install/ajaxLoadMainSql.php
@@ -2,9 +2,9 @@
 /**
  * ajaxLoadMainSql.php
  * @package Installer
- * @copyright Copyright 2003-2016 Zen Cart Development Team
+ * @copyright Copyright 2003-2018 Zen Cart Development Team
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: Author: DrByte  Fri Oct 9 15:32:07 2015 -0400 New in v1.5.5 $
+ * @version $Id: Author: DrByte  Modified in v1.5.6 $
  */
 define('IS_ADMIN_FLAG', false);
 if (!defined('__DIR__')) define('__DIR__', dirname(__FILE__));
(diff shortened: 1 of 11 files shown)


torvista:
In phpmyadmin, on the sessions table it will not allow varchar(256) with Innob, even with utf8_general_ci (10.1.29-MariaDB).

drbyte:
Some reading I've done further suggests it may require MySQL higher than 5.7.7 or MariaDB higher than 10.2.2 

What error message/s do you get?

torvista:
clipboard01

torvista:
I note the ZC156 installer appears to complete successfully despite having failed to create this table: no error flagged (subsequently the admin login fails with a debug).

drbyte:
Did `whos_online` throw any errors? 
From my reading I understand that `unique key`s need to also support the full length * 4, so would expect the same error to occur when creating the `session_id` unique index on the `whos_online` table, too.

torvista:
whos_oline is created ok: has varchar(256) and is still MyISAM.
clipboard01

drbyte:
Puzzling

I have mariadb 10.2.12 installed, so my testing results are a bit different.

torvista:
I've installed ZC158 many times on Xampp having created the DB with
CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_520_ci;
with no problem.
But I've just started to tryout Laragon and the install did not create table products.description: needing both varchars at 250 to create the table (they are set differently at 254/255).
Subsequently when playing with this value in the database admin tool Heidi, it won't accept more than 250. The error is the same as that created in the installer debug.

Clipboard04

This is php 7.4.16, mysql 5.7.24.

drbyte:
MySQL has significantly tinier capacity (1000) than MariaDB does (over 3000) on index lengths.

The issue is that if you want to support mb4 on MySQL then you can NEVER have columns longer than 250 characters, because each character requires 4 bytes. 250 * 4 = 1000 which is MySQL's maximum index size.

The problem is compounded many times over when indexing multiple columns.
And even worse when storeowners want to stuff even longer values in those columns.

You have some options:
- don't use MySQL (use MariaDB instead)
- drop (or skip creating) indexes on those longer fields (yes, you'll have to edit the .SQL files before installing)
- use shorter column lengths (yes, you'll have to edit the .SQL files before installing)

drbyte:
For v158 I had increased the length of the field in question due to common request for longer product names. Unfortunately that blew things out when using MySQL.
I've reverted the default install to use a shorter length

PR #1598, opened 2018-01-14
zc_install to recognize "local/configure.php" for offline dev use

drbyte:


torvista:
I used today's 156 zip file with /local configures, overwrote with the eight files referenced here and imported my 1.55e database. 
The installer found, flagged and used the correct local configures and upgraded the database to 1.56.
Works for me!

mc12345678:
This appears to address the use of local/configure.php to identify an alternate local database within local/configure.php to support updating the database, but not address the use of say an older includes/configure.php file from outside the local environment (paths may be different in those files and in the sequence of store load (includes/local/configure.php, includes/configure.php, includes/defined_paths.php) the outside store "directories" (not included in the ZC 1.5.5 style configure.php) may be defined in that file and not support the local environment.

An issue (potentially) related to this (though not "providing credit" to this modification) was identified at https://www.zen-cart.com/showthread.php?223450-Logs-not-being-written-after-upgrade-to-1-5-5e-f and I had a post to add to that but had to start a new thread at: https://www.zen-cart.com/showthread.php?223459-Logs-not-being-written-after-upgrade-to-1-5-5e-f

drbyte:
> This appears to address the use of local/configure.php to identify an alternate local database within local/configure.php to support updating the database, but not address the use of say an older includes/configure.php file from outside the local environment (paths may be different in those files and in the sequence of store load (includes/local/configure.php, includes/configure.php, includes/defined_paths.php) the outside store "directories" (not included in the ZC 1.5.5 style configure.php) may be defined in that file and not support the local environment.

There's a lotta "words" there, but it's hard to discern what's really being said. 

Do you mean "if some additional custom defines had been added, they're ignored during the file-upgrade"? And is this really isolated to "local" configure.php files?

Issue #1583, opened 2018-01-01
ZC 156/160: Refactoring admin categories / products and listing page


Zen4All:
What is your opinion on refactoring admin categories / products and listing page?
Currently we have:
- admin/categories.php
- admin/product.php
- admin/includes/modules/category_product_listing.php
These files share code, contain duplicate (unused) code,

I propose to make the listing a separate page and not a module file. this would mean moving it to the admin folder form includes/modules.
code will need to be moved around between the three file, but will make it much easier to find and modify things in the future.

drbyte:
No objection here.  Those pages do need an overhaul.

Keep in mind there are multiple product types supported, so consideration should be given there too.

Zen4All:
Ok, i'll get to it. Yes I will remember product types. I was working on a way to eliminate the need for new files for each product type, but that is a long way from being ready.

drbyte:
> I was working on a way to eliminate the need for new files for each product type

Thumbs up 👍 to that!

The product-free-shipping was only added as an example of how to slightly clone the product-general for some customization.
The others serve similarly.
But it's definitely long overdue for a simpler way to have multiple types without so much annoying duplication! (and so many near-duplicate files in core code, and having to be handled during upgrades, and for some plugins, etc etc).

Zen4All:
The setup I am testing right now is:
- Put all fields in a database table
- connect these fields to a product type
- read the fields from the database
- build a screen dynamically, according to the fields that are connected to a specific product type.

This setup works, and gives a nice clean file. The "problem", you need to put a lot of info per field in the database. The advantage is, adding new fields is childs play, using a simple screen interface.

But all this if for later.
First things first

drbyte:
Oh. You mean having a database of database field names/types/constraints/etc?
Definitely pros and cons with that.

Zen4All:
Yes exactly
![product_fields](https://user-images.githubusercontent.com/2766593/34512919-a008c2aa-f065-11e7-97bf-a707fc503182.png)

drbyte:
k - may require some metrics testing, since it forces db integrity to be handled by code instead of by the db, which can cause performance issues. It also introduces reporting challenges. 
Not opposed to it, just aware that it'll need to be treated carefully. 

As you said, it's a separate project.

Zen4All:
Nothing's final yet. It started out with a simple table, but go tout of hand rather quickly 😆 . An other possibility for defines etc is to use a file-based system, but I have to investigate on that.

[...]

PR #1604, opened 2018-01-17
v156: Admin products and categories facelift/refactoring

Zen4All:
see #1583 for more information

@drbyte edit:
Closes #1631


drbyte:
@Zen4All ... thoughts on readiness for review/merge?

Zen4All:
Hoping to be ready end of this week

drbyte:
I hope to review this tomorrow.

Zen4All:
Okay, I have some latest css changes am bug-fixes, will commit them this evening. After that, there may be one or two defines that are missing, but that is it.

drbyte:
Have you any thoughts on how best to handle the floating Preview button?

image 2018-03-30 at 3 37 25 pm
image 2018-03-30 at 3 38 38 pm

Zen4All:
Sorry, I do not understand what you want me to do.

drbyte:
I was wondering your opinion on how much the overlap may be a problem for users, and what options could be presented as a ‘solution’. 

In practical use I’m guessing it’s only a minor annoyance. While there’s always someone who will complain, it can always be addressed later.

Zen4All:
The idea of making a floating button came from one of my customers who has a 3 language store. He got tired of scrolling up or down. Since then I implemented this on multiple stores, and nobody complained about the overlap.
One option could be to make the space between the bottom of the screen an bottom of the button.

zc_install/sql/updates/mysql_upgrade_zencart_156.sql
```diff
@@ -127,7 +127,16 @@ SELECT products_attributes_id FROM products_attributes WHERE products_id NOT IN
...
+## alter admin_pages for new product listing pages
+UPDATE admin_pages
+SET language_key = 'BOX_CATALOG_CATEGORY',
+    main_page = 'FILENAME_CATEGORY_PRODUCT_LISTING',
+    display_on_menu = 'N',
+    sort_order = 18
+WHERE page_key = 'categories';
 
+INSERT INTO admin_pages (page_key, language_key, main_page, page_params, menu_key, display_on_menu, sort_order)
+VALUES ('categoriesProductListing', 'BOX_CATALOG_CATEGORIES_PRODUCTS', 'FILENAME_CATEGORY_PRODUCT_LISTING', '', 'catalog', 'Y', 1);
```
drbyte: Why the insertion of a new key, instead of renaming-and-hiding?

Zen4All: I always thought the pages had to be in the admin_pages  table, so you can allow or disallow them in the profiles.
That is why I added a new item for the listing, and updated the old item so it represents the category page.

zcwilt:
:+1:

PR #1607, opened 2018-01-19
js tweak to state-selector dropdown

llynix:
With state dropdowns selected on the edit address page (address book process) the users state is not selected as default.  This adds a default zone if zone_id is set to the update_zone piece of javascript.


includes/functions/functions_general.php
```diff
@@ -1674,7 +1674,9 @@ function zen_js_zone_list($country, $form, $field, $showTextField = true) {
       $num_state = 1;
       while (!$states->EOF) {
         if ($num_state == '1') $output_string .= '    ' . $form . '.' . $field . '.options[0] = new Option("' . PLEASE_SELECT . '", "");' . "\n";
-        $output_string .= '    ' . $form . '.' . $field . '.options[' . $num_state . '] = new Option("' . $states->fields['zone_name'] . '", "' . $states->fields['zone_id'] . '");' . "\n";
+        $output_string .= '    ' . $form . '.' . $field . '.options[' . $num_state . '] = new Option("' . $states->fields['zone_name'] . '", "' . $states->fields['zone_id'] . '"';
+        if($states->fields['zone_id'] == $zone_id) $output_string .= ',1,1';
```
drbyte: What does the `,1,1` refer to here?

includes/templates/responsive_classic/jscript/jscript_addr_pulldowns.php
```diff
@@ -74,7 +74,7 @@ function update_zone(theForm) {
     theForm.zone_id.options[NumState] = null;
   }
   // build dynamic list of countries/zones for pulldown
-<?php echo zen_js_zone_list('SelectedCountry', 'theForm', 'zone_id'); ?>
+<?php echo zen_js_zone_list('SelectedCountry', 'theForm', 'zone_id', isset($zone_id) ? $zone_id : -1); ?>
```
drbyte: I'm not sure I understand this addition. The `zen_js_zone_list()` function does accept a 4th parameter, but not the one you're passing here.

includes/functions/functions_general.php
llynix: https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option

It makes the option being added selected.

includes/templates/responsive_classic/jscript/jscript_addr_pulldowns.php
llynix: When viewing your address inside the my account address book the zen_js_zone_list doesn't select your state as default, instead defaulting to Please Select.  This patch fixes that.

PR #1609, opened 2018-01-23
Restore display of featured, new, special, and upcoming product to ca…

mc12345678:
…tegories.

Two items addressed in this commit.  1: to restore the display of
featured, new, special, and upcoming products on category pages that are
not the main page.  This is a caused by initially referencing an empty
string element as an array which builds the string up with additional
characters to increase the size of the string.

This sets/resets the $categories_products_id_list to be an array in the
event that it is not when calling one of the above modules or if otherwise
calling the zen_get_categories_products_list function.

### DIFF
--- a/includes/modules/upcoming_products.php
+++ b/includes/modules/upcoming_products.php
@@ -13,7 +13,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 }
 
 // initialize vars
-$categories_products_id_list = '';
+$categories_products_id_list = array();
 $list_of_products = '';
 $expected_query = '';
 
--- a/includes/functions/functions_categories.php
+++ b/includes/functions/functions_categories.php
@@ -481,6 +481,11 @@
   function zen_get_categories_products_list($categories_id, $include_deactivated = false, $include_child = true, $parent_category = '0', $display_limit = '') {
     global $db;
     global $categories_products_id_list;
+
+    if (!isset($categories_products_id_list) || !is_array($categories_products_id_list)) {
+      $categories_products_id_list = array();
+    }
+
     $childCatID = str_replace('_', '', substr($categories_id, strrpos($categories_id, '_')));
 
     $current_cPath = ($parent_category != '0' ? $parent_category . '_' : '') . $categories_id;
--- a/includes/modules/featured_products.php
+++ b/includes/modules/featured_products.php
@@ -13,7 +13,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 }
 
 // initialize vars
-$categories_products_id_list = '';
+$categories_products_id_list = array();
 $list_of_products = '';
 $featured_products_query = '';
 $display_limit = '';
@@ -89,4 +89,4 @@ if ($num_products_count > 0) {
     $zc_show_featured = true;
   }
 }
-?>
+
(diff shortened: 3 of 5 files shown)


mc12345678:
The problem appears to occur in a PHP version greater than 7.0.24 (though likely not until 7.1.x observed to be a problem at 7.1.10).  There are no error logs generated, it just simply doesn't work like "expected" because of "emptying" the array to a string instead of an array.

includes/modules/upcoming_products.php
```diff
@@ -22,7 +22,7 @@
 $limit_clause = "  order by " . (EXPECTED_PRODUCTS_FIELD == 'date_expected' ? 'date_expected' : 'products_name') . " " . (EXPECTED_PRODUCTS_SORT == 'asc' ? 'asc' : 'desc') . "
                    limit " . (int)MAX_DISPLAY_UPCOMING_PRODUCTS;
 
-if ( (($manufacturers_id > 0 && $_GET['filter_id'] == 0) || $_GET['music_genre_id'] > 0 || $_GET['record_company_id'] > 0) || (!isset($new_products_category_id) || $new_products_category_id == '0') ) {
+if ( (($manufacturers_id > 0 && $_GET['filter_id'] == 0) || (isset($_GET['music_genre_id']) && $_GET['music_genre_id'] > 0) || (isset($_GET['record_company_id']) && $_GET['record_company_id'] > 0)) || (!isset($new_products_category_id) || $new_products_category_id == '0') ) {
```
drbyte: You can use `! empty($_GET['music_genre_id'])` here instead of the longer `(isset($_GET['music_genre_id']) && $_GET['music_genre_id'] > 0)`

drbyte:
Thanks for the work in investigating this!

includes/modules/upcoming_products.php
mc12345678: I had contemplated that, but wasn't confident that `empty` could correctly be used here.

Based on the one comment, revisited the code to provide an additional commit addressing that one specific comment and other related factors.

This branch currently contains two commits, my original revision to address the issue and then other modifications that have been seen elsewhere (though did not tackle future potential PHP 7.2 related items such as `sizeof` -> `count`).  I have individual modifications on a separate branch though it was chaotic in change this/change that, oh yeah, what about this...

Let me know how if anything should be pulled out/rewritten, etc...

Also, wanted to revise the `while (!$var->EOF)` structure to `foreach ($var as $var_item)` for a few of them, but the `NextRandom()` style method was used and I didn't know how to properly incorporate that other than as provided.

drbyte: Haha! I can appreciate the challenge of wanting to "do everything at once"! 

Since this is against the v155 branch, which ideally is taking only bugfixes, can we reduce this PR down to just what's required to fix the presenting bug, and then do a separate PR to v156 for the broader cleanup?

drbyte: > though did not tackle future potential PHP 7.2 related items such as sizeof -> count.

FWIW the change from sizeof->count isn't a requirement for PHP 7.2. It's just something that shows up in some code review tools where a rule for "avoid alias functions" is enabled.  You'll see that in most cases i reverted back from making that sweeping change.

drbyte:
Thanks!

PR #1613, opened 2018-01-24
Additional admin/orders notifiers

lat9:
Allows an admin observer to include its additional "legend" icons and add to the left-hand legend-icons on a per-order basis.


drbyte:
Thanks!

PR #1614, opened 2018-01-25
Restore display of featured, new, special, and upcoming product to ca…

mc12345678:
…tegories.

Two items addressed in this commit.  1: to restore the display of
featured, new, special, and upcoming products on category pages that are
not the main page.  This is a caused by initially referencing an empty
string element as an array which builds the string up with additional
characters to increase the size of the string.

This sets/resets the $categories_products_id_list to be an array in the
event that it is not when calling one of the above modules or if otherwise
calling the zen_get_categories_products_list function.

Also: Reworked mySql to use all capital letters for mySql commands as well
as modifying incorporation of an integer value by removing the single
quote(s) around the cast value.  This reduces the supported processing of
then taking the integer cast value, converting it to a string that the
database engine must now/again convert the string to a number.

Replaced isset($var) && sizeof($var) > 0 with !empty($var)

Replaced !isset($var) || sizeof($var) == 0 with empty($var)

Primarily modified to address reducing notices generated.

### DIFF
--- a/includes/functions/functions_categories.php
+++ b/includes/functions/functions_categories.php
@@ -481,6 +481,11 @@
   function zen_get_categories_products_list($categories_id, $include_deactivated = false, $include_child = true, $parent_category = '0', $display_limit = '') {
     global $db;
     global $categories_products_id_list;
+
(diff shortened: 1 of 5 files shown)


drbyte:
Thanks!

PR #1615, opened 2018-01-25
Fixed Mysql Structure

drillsar:
Moved some configuration files to proper place


zc_install/sql/install/mysql_zencart.sql
```diff
@@ -2242,7 +2242,9 @@ VALUES ('configMyStore', 'BOX_CONFIGURATION_MY_STORE', 'FILENAME_CONFIGURATION',
        ('configAllListing', 'BOX_CONFIGURATION_ALL_LISTING', 'FILENAME_CONFIGURATION', 'gID=23', 'configuration', 'Y', 22),
        ('configIndexListing', 'BOX_CONFIGURATION_INDEX_LISTING', 'FILENAME_CONFIGURATION', 'gID=24', 'configuration', 'Y', 23),
        ('configDefinePageStatus', 'BOX_CONFIGURATION_DEFINE_PAGE_STATUS', 'FILENAME_CONFIGURATION', 'gID=25', 'configuration', 'Y', 24),
-       ('configEzPagesSettings', 'BOX_CONFIGURATION_EZPAGES_SETTINGS', 'FILENAME_CONFIGURATION', 'gID=30', 'configuration', 'Y', 25),
+       ('configCheckoutSettings', 'BOX_CONFIGURATION_CHECKOUT_SETTINGS', 'FILENAME_CONFIGURATION', 'gID=26', 'configuration', 'Y', 26),       
+       ('configWidgets', 'BOX_CONFIGURATION_WIDGET', 'FILENAME_CONFIGURATION', 'gID=27', 'configuration', 'Y', 27),
+       ('configEzPagesSettings', 'BOX_CONFIGURATION_EZPAGES_SETTINGS', 'FILENAME_CONFIGURATION', 'gID=28', 'configuration', 'Y', 28),
```
mc12345678: Curious, is there a reason or specific need other than to set Ez-Pages as the next assignment as to why it was changed from 30 to 28? It was added back in ZC 1.3.0 as gID=30 and has been so ever since without apparent issue (nor specific rhyme or reason other than leaving room for other options to be included).

Also in the effort to make these changes why was the sort order not consistently set as one less than the gID like the other previous assignments?

drillsar: Well the way it is now if you add to the database it would go to 31 and so on. The way it is now next add will go to gid 29. I think it makes it much easier to read configuration. No big deal I guess.

PR #1617, opened 2018-01-30
Align previous commit #1613 with v160

lat9:
Align the notifiers previously added, using the interface currently present in the v160 version of the orders.php script.


lat9:
Thanks, DrByte!

drbyte:
Thanks for checking the consistency. Helps keep us both sane! :)

PR #1619, opened 2018-02-01
remove unnecessary preg_replace Fixes #1616

### LINKED ISSUE #1616: init_templates.php
i was hating the admin title in v155, and thought i would look into changing it for v160; and it looks like it was already done!  that said, i'm not sure if it is worth it, but the code on line could be changed from:

`$pagename = preg_replace('/\.php$/', '', basename($PHP_SELF));`

to 

`$pagename = basename($PHP_SELF, '.php');`

a bit cleaner IMHO, but a very minor detail.  i am also a big fan of breaking out the html header as currently done in v160.

proseLA:
small change.  little cleaner.

### DIFF
--- a/admin/includes/init_includes/init_templates.php
+++ b/admin/includes/init_includes/init_templates.php
@@ -38,7 +38,7 @@ if (isset($_GET['cmd'])) {
   }
 }
 if ($pagename == '') {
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks!

PR #1621, opened 2018-02-01
zc156-downloads-for-guests

lat9:
Update the download page's processing, with some re-factoring, to enable a download to be gathered via order's email address, in support of guest checkout.


drbyte:
👍

PR #1622, opened 2018-02-05
v156-customer-logged-in-guest

lat9:
Additional functions, allowing core-code and plugins to consistently identify whether or not a customer is logged in and whether/not a guest-checkout is active.


drbyte:
👍

PR #1625, opened 2018-02-11
V155: Update for PHP7

drillsar:
Removed deprecated magic_quotes support; updated constructor


mc12345678:
What issues are seen with the removed code when php 7.x is used?

Won't that removal cause problems if the php version is less than 5.4 and therefore change the php version applicability list for ZC 1.5.5?

This comment is not directed at the change of the `CardinalXMLParser` class constructor.

PR #1626, opened 2018-02-15
Simplify queries on creating a new language

Zen4All:
These three queries can be simplified. The selection of the id can be
done in the same table as the selection of name/description.
This reduces load on the database significantly. For the categorie
table, with the default demo categories, from 0.0018 to 0.0005 seconds.
This is over 3 times faster.

Old Query:
originalquery

New query:
newquery

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1626)

### DIFF
--- a/admin/languages.php
+++ b/admin/languages.php
@@ -39,10 +39,9 @@ if (zen_not_null($action)) {
         }
 
 // create additional categories_description records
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks!

PR #1631, opened 2018-02-25
Incorporate changes across product types for consistency

mc12345678:
There have been changes that have been made to the various product types
supplied with ZC; however, there have either been html validation errors
or inconsistencies applied to each product type.  This commit takes the
common/newest usage and spreads it across the 5 default product types.

Some changes were made in ZC 1.5.6 to one product type but not all, other
changes appear to have been made before ZC 1.5.0 and about anywhere in
between.

### DIFF
--- a/admin/document_general.php
+++ b/admin/document_general.php
@@ -1,10 +1,10 @@
 <?php
 /**
  * @package admin
- * @copyright Copyright 2003-2016 Zen Cart Development Team
+ * @copyright Copyright 2003-2017 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: Author: DrByte  Sun Oct 18 02:03:48 2015 -0400 Modified in v1.5.5 $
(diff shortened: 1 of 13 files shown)


Zen4All:
This commit may not be necessary, as I am working on an update which makes these files obsolete. Only product.php is needed, and each file type has its own subfolder containing the usual files like collect_info.

See #1604

mc12345678:
If not needed then fine if it is ignored.  Didn't really see where any one product type was fully html validation ready, had all of the "recent" improvements, etc...

Would ask that if such a "single" product format is going to be used that consideration be included to support additional product types.  I know that for the rescue I work with we have created a product type to support display of rescue related data for the animals.  There likely re other cases of product types out there to also remain considered.

Zen4All:
As you know the current categories, products and listing files are a spaghetti mess of (duplicate) code.
I created three "new" files.
1. category / product listing
2. category
3. product.

The structure of these files is much simpler, and the html validates. I removed a lot of redundant tables.

The product.php loads the necessary product type files, but the product type specific file (like document_general.php) is not needed any more.

drbyte:
Closing since #1604 has been merged, and created a dozen conflicts. 
Feel free to re-submit if a proposed fix was missed in that PR.

PR #1632, opened 2018-02-25
v156-Enhance log formatting

lat9:
Due to the way that the log records are currently formatted, it's sometimes "difficult" to connect the actual error to the backtrace information.  This change modifies the formatted output, grouping the debug-backtrace information with the PHP information, making it easier to debug/determine what's going on.

Also includes some re-factoring.


drbyte:
Thanks!

PR #1633, opened 2018-03-01
v156 Remove integer key data from iterator results

mc12345678:
In ZC 1.5.5, the iterator class was implemented allowing the use of foreach
to cycle through query results.  To support this operation, the rewind
module was written to use the Move($position) method which was written to
allow query results to be returned as both a key=>value response as well
as a position=>value response.  Therefore, the initial reset of the query
results would produce an array of both the key=>value pair and position=>value
pair.  Then when the MoveNext method was used, the key=>value pair would
be updated; however, the position=>value pair remained constant.

This commit adds continued support for code that intentionally used the
Move($position) to access the integer based field result and to also
prevent adding the integer based results to the Move($value) method returned
values.

A similar condition was identified in the ExecuteRandomMulti function.

Incorporation of this modification added a parameter to each function with
a default value that causes the function to operate as it has for all of
ZC 1.5.x.

This was ultimately an issue because mysqli_fetch_array (and the legacy
mysql_fetch_array) when not provided a resulttype as described in the
php manual (http://php.net/manual/en/mysqli-result.fetch-array.php) defaults
to providing both the integer based result and the field key based result.

Perhaps this was done to support some yet older code; however, with the
implementation of the iterator class to provide only field results
(consistently at each next result) a modification was needed at least against
the foreach implementation.

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -262,7 +262,7 @@ class queryFactory extends base {
         if ($obj->RecordCount() > 0) {
           $zp_result_array = mysqli_fetch_assoc($zp_db_resource);
           if ($zp_result_array) {
-            $obj->fields = array_replace($obj->fields, mysqli_fetch_assoc($zp_db_resource));
+            $obj->fields = array_replace($obj->fields, $zp_result_array);
             $obj->EOF = false;
           }
         }
@@ -284,7 +284,7 @@ class queryFactory extends base {
         return $this->Execute ($zf_sql, false, false, 0, true);
     }
 
-  function ExecuteRandomMulti($zf_sql, $zf_limit = 0, $zf_cache = false, $zf_cachetime=0, $remove_from_queryCache = false, $int_key = true) {
+  function ExecuteRandomMulti($zf_sql, $zf_limit = 0, $zf_cache = false, $zf_cachetime=0, $remove_from_queryCache = false) {
     $this->zf_sql = $zf_sql;
     $time_start = explode(' ', microtime());
     $obj = new queryFactoryResult($this->link);
@@ -310,13 +310,7 @@ class queryFactory extends base {
         $zp_ii = 0;
         while ($zp_ii < $zf_limit) {
           $obj->result[$zp_ii] = array();
-          if ($int_key === true) {
-            $obj->result[$zp_ii] = @mysqli_fetch_array($zp_db_resource);
-          } else if ($int_key === 'only') {
-            $obj->result[$zp_ii] = @mysqli_fetch_row($zp_db_resource);
-          } else {
-            $obj->result[$zp_ii] = @mysqli_fetch_assoc($zp_db_resource);
-          }
+          $obj->result[$zp_ii] = @mysqli_fetch_assoc($zp_db_resource);
           if (!$obj->result[$zp_ii]) {
             unset($obj->result[$zp_ii]);
             $obj->limit = $zp_ii;
@@ -532,7 +526,7 @@ class queryFactoryResult implements Countable, Iterator {
    *
    * @var array of field => value pairs
    */
-  public $fields;
+  public $fields = array();
 
   /**
    * Indicates if the result is cached.
@@ -682,7 +676,7 @@ class queryFactoryResult implements Countable, Iterator {
    *
    * @param int $zp_row the row to move to
    */
-  public function Move($zp_row, $int_key = true) {
+  public function Move($zp_row) {
     global $db;
     if ($this->is_cached) {
       if($zp_row >= sizeof($this->result)) {
@@ -694,13 +688,7 @@ class queryFactoryResult implements Countable, Iterator {
         $this->EOF = false;
       }
     } else if (@mysqli_data_seek($this->resource, $zp_row)) {
-      if ($int_key === true) {
-        $this->fields = array_replace($this->fields, @mysqli_fetch_array($this->resource));
-      } else if ($int_key == 'only') {
-        $this->fields = array_replace($this->fields, @mysqli_fetch_row($this->resource));
-      } else {
-        $this->fields = array_replace($this->fields, @mysqli_fetch_assoc($this->resource));
-      }
+      $this->fields = array_replace($this->fields, @mysqli_fetch_assoc($this->resource));
       $this->cursor = $zp_row;
       $this->EOF = false;
     } else {
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -201,9 +201,7 @@ class queryFactory extends base {
       $obj->result = $zp_result_array;
       if (sizeof($zp_result_array) > 0 ) {
         $obj->EOF = false;
-        foreach($zp_result_array[0] as $key => $value) {
-          $obj->fields[$key] = $value;
-        }
+        $obj->fields = array_replace($obj->fields, $zp_result_array[0]);
       }
     } elseif ($zf_cache) {
       $zc_cache->sql_cache_expire_now($zf_sql);
@@ -222,28 +220,19 @@ class queryFactory extends base {
         if ($zp_rows > 0) {
           $zp_ii = 0;
           while ($zp_ii < $zp_rows) {
-            $zp_result_array = mysqli_fetch_array($zp_db_resource);
-            if ($zp_result_array) {
-              $obj->result[$zp_ii] = array();
-              foreach($zp_result_array as $key => $value) {
-                if (!preg_match('/^[0-9]/', $key)) {
-                  $obj->result[$zp_ii][$key] = $value;
-                }
-              }
-            } else {
+            $obj->result[$zp_ii] = array();
+            $obj->result[$zp_ii] = mysqli_fetch_assoc($zp_db_resource);
+            if (!($obj->result[$zp_ii])) {
+              unset($obj->result[$zp_ii]);
               $obj->limit = $zp_ii;
               break;
             }
             $zp_ii++;
           }
-          foreach($obj->result[$obj->cursor] as $key => $value) {
-            if (!preg_match('/^[0-9]/', $key)) {
-              $obj->fields[$key] = $value;
-            }
-          }
+          $obj->fields = array_replace($obj->fields, $obj->result[$obj->cursor]);
           $obj->EOF = false;
         }
-        unset($zp_ii, $zp_result_array, $key, $value);
+        unset($zp_ii);
       }
       $zc_cache->sql_cache_store($zf_sql, $obj->result);
       $obj->is_cached = true;
@@ -271,13 +260,9 @@ class queryFactory extends base {
       } else {
         $obj->resource = $zp_db_resource;
         if ($obj->RecordCount() > 0) {
-          $zp_result_array = mysqli_fetch_array($zp_db_resource);
+          $zp_result_array = mysqli_fetch_assoc($zp_db_resource);
           if ($zp_result_array) {
-            foreach($zp_result_array as $key => $value) {
-              if (!preg_match('/^[0-9]/', $key)) {
-                $obj->fields[$key] = $value;
-              }
-            }
+            $obj->fields = array_replace($obj->fields, mysqli_fetch_assoc($zp_db_resource));
             $obj->EOF = false;
           }
         }
@@ -324,23 +309,22 @@ class queryFactory extends base {
         mysqli_data_seek($zp_db_resource, $zp_start_row);
         $zp_ii = 0;
         while ($zp_ii < $zf_limit) {
-          $zp_result_array = @mysqli_fetch_array($zp_db_resource);
-          if ($zp_result_array) {
-            $obj->result[$zp_ii] = array();
-            foreach($zp_result_array as $key => $value) {
-              $add_key = ($int_key === true ? true : !preg_match('/^[0-9]/', $key)); // mc12345678 support controlling presence of the integer key result
-
-              if (!$add_key) continue;
-
-              $obj->result[$zp_ii][$key] = $value;
-            }
+          $obj->result[$zp_ii] = array();
+          if ($int_key === true) {
+            $obj->result[$zp_ii] = @mysqli_fetch_array($zp_db_resource);
+          } else if ($int_key === 'only') {
+            $obj->result[$zp_ii] = @mysqli_fetch_row($zp_db_resource);
           } else {
+            $obj->result[$zp_ii] = @mysqli_fetch_assoc($zp_db_resource);
+          }
+          if (!$obj->result[$zp_ii]) {
+            unset($obj->result[$zp_ii]);
             $obj->limit = $zp_ii;
             break;
           }
           $zp_ii++;
         }
-        unset($zp_ii, $zp_result_array, $key, $value);
+        unset($zp_ii);
         $obj->EOF = false;
 
         $obj->result_random = array_rand($obj->result, count($obj->result));
@@ -630,20 +614,14 @@ class queryFactoryResult implements Countable, Iterator {
       if ($this->cursor >= sizeof($this->result)) {
         $this->EOF = true;
       } else {
-        foreach($this->result[$this->cursor] as $key => $value) {
-          $this->fields[$key] = $value;
-        }
+        $this->fields = array_replace($this->fields, $this->result[$this->cursor]);
       }
     } else {
-      $zp_result_array = @mysqli_fetch_array($this->resource);
+      $zp_result_array = @mysqli_fetch_assoc($this->resource);
+      $this->fields = array_replace($this->fields, $zp_result_array);
       if (!$zp_result_array) {
         $this->EOF = true;
-      } else {
-        foreach($zp_result_array as $key => $value) {
-          if (!preg_match('/^[0-9]/', $key)) {
-            $this->fields[$key] = $value;
-          }
-        }
+        unset($this->fields);
       }
     }
   }
@@ -654,12 +632,7 @@ class queryFactoryResult implements Countable, Iterator {
   public function MoveNextRandom() {
     $this->cursor++;
     if ($this->cursor < $this->limit) {
-      $zp_result_array = $this->result[$this->result_random[$this->cursor]];
-      foreach($zp_result_array as $key => $value) {
-        if (!preg_match('/^[0-9]/', $key)) {
-          $this->fields[$key] = $value;
-        }
-      }
+      $this->fields = array_replace($this->fields, $this->result[$this->result_random[$this->cursor]]);
     } else {
       $this->EOF = true;
     }
@@ -716,20 +689,17 @@ class queryFactoryResult implements Countable, Iterator {
         $this->cursor = sizeof($this->result);
         $this->EOF = true;
       } else {
-        foreach($this->result[$zp_row] as $key => $value) {
-          $this->fields[$key] = $value;
-        }
+        $this->fields = array_replace($this->fields, $this->result[$zp_row]);
         $this->cursor = $zp_row;
         $this->EOF = false;
       }
     } else if (@mysqli_data_seek($this->resource, $zp_row)) {
-      $zp_result_array = @mysqli_fetch_array($this->resource);
-      foreach($zp_result_array as $key => $value) {
-        $add_key = ($int_key === true ? true : !preg_match('/^[0-9]/', $key)); // mc12345678 support controlling presence of the integer key result
-
-        if (!$add_key) continue;
-
-        $this->fields[$key] = $value;
+      if ($int_key === true) {
+        $this->fields = array_replace($this->fields, @mysqli_fetch_array($this->resource));
+      } else if ($int_key == 'only') {
+        $this->fields = array_replace($this->fields, @mysqli_fetch_row($this->resource));
+      } else {
+        $this->fields = array_replace($this->fields, @mysqli_fetch_assoc($this->resource));
       }
       $this->cursor = $zp_row;
       $this->EOF = false;
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -299,7 +299,7 @@ class queryFactory extends base {
         return $this->Execute ($zf_sql, false, false, 0, true);
     }
 
-  function ExecuteRandomMulti($zf_sql, $zf_limit = 0, $zf_cache = false, $zf_cachetime=0, $remove_from_queryCache = false) {
+  function ExecuteRandomMulti($zf_sql, $zf_limit = 0, $zf_cache = false, $zf_cachetime=0, $remove_from_queryCache = false, $int_key = true) {
     $this->zf_sql = $zf_sql;
     $time_start = explode(' ', microtime());
     $obj = new queryFactoryResult($this->link);
@@ -328,6 +328,10 @@ class queryFactory extends base {
           if ($zp_result_array) {
             $obj->result[$zp_ii] = array();
             foreach($zp_result_array as $key => $value) {
+              $add_key = ($int_key === true ? true : !preg_match('/^[0-9]/', $key)); // mc12345678 support controlling presence of the integer key result
+
+              if (!$add_key) continue;
+
               $obj->result[$zp_ii][$key] = $value;
             }
           } else {
@@ -667,7 +671,7 @@ class queryFactoryResult implements Countable, Iterator {
   public function rewind() {
       $this->EOF = ($this->RecordCount() == 0);
       if ($this->RecordCount() !== 0) {
-          $this->Move(0);
+          $this->Move(0, false); // mc12345678 eliminate return of integer based key
       }
   }
 
@@ -705,7 +709,7 @@ class queryFactoryResult implements Countable, Iterator {
    *
    * @param int $zp_row the row to move to
    */
-  public function Move($zp_row) {
+  public function Move($zp_row, $int_key = true) {
     global $db;
     if ($this->is_cached) {
       if($zp_row >= sizeof($this->result)) {
@@ -721,6 +725,10 @@ class queryFactoryResult implements Countable, Iterator {
     } else if (@mysqli_data_seek($this->resource, $zp_row)) {
       $zp_result_array = @mysqli_fetch_array($this->resource);
       foreach($zp_result_array as $key => $value) {
+        $add_key = ($int_key === true ? true : !preg_match('/^[0-9]/', $key)); // mc12345678 support controlling presence of the integer key result
+
+        if (!$add_key) continue;
+
         $this->fields[$key] = $value;
       }
       $this->cursor = $zp_row;


mc12345678:
Looking at the code a little more from a future state perspective, would entertain changing the interim variable from `$add_val` to `$add_key` if considered to be "clearer".

mc12345678:
The provided code change comes from a ZC forum discussion found here: https://www.zen-cart.com/showthread.php?223658-db-SELECT-returning-numeric-indexes

Also, this PR is similar to ones submitted for ZC 1.6.0 #1634 and ZC 1.5.5 #1635.

drbyte:
I'm trying to understand "when" this "started to be a problem" ... because I'm pretty sure the iterator was added long before.

mc12345678:
Not entirely sure how to answer "when", but I can recall and can see that if/when `$result->Move(x);` is called, and then the "fields" (`$result->fields`) iterated without addressing/accessing a specific field, then based on the php manual anytime using php 5.x or above, the use of `mysql_fields_array` with no final parameter will export both the associated field name and result and then an indexed list of the same result. The `MoveNext()` method would then replace/update the associative keyed data without touching the numerically indexed data. 

ZC core code does not appear to reveal this issue at least related to using the `Move` method because ZC specifically uses/references the data being iterated/combined. It does not "blindly" use the result data. 

As to when in history it was introduced, I have not reviewed past code (pre-1.5.0) to see the history of the `query_factory` class. Though there is/was a function related to providing random results that previously was used for randomizing data for lists such as new items that had the potential of providing the same product/item twice because of the use of the `Move` method's incorporation although that function has been abandoned in place with alternate code/logic provided to get the desired random value. That too has not been further investigated other than its observation in ZC 1.5.5.

As to how/when it was identified as something suggested to be modified to something like suggested in this pull request, that is discussed in the associated forum thread and becomes noticeable if all of the key/value pairs are used within a foreach loop with an expectation that only the associative array is returned like seen when performing a query followed by stepping through the results one time.

There have been several occasions over the last couple of years where the suggestion has been made by the core team to reinitiate an iteration of a query's results by use of the `Move` with a response comment related to something like yes, but it doesn't give the desired result(s) or it has its own problems. I myself can recall a thread I started that I kept pure to the issue of discussion rather than bringing up the one issue and also seeing that more than the associative array was provided back. As a result of seeing that I did instead modify how I was going to approach the problem it could have caused by only referencing known/expected query results instead of simply iterating through the results or querying the database a second time to avoid using the `Move` method and/or potentially the query cache.

All-in-all I would suggest that actually in addition to the above that database fields be able to be returned with the naming convention supported by mysql at least for/as unquoted fields. Currently fields that begin with a number are excluded from the returned fields list when using the results directly obtained from the `Execute` method followed by the `MoveNext` method, but are available if continuously using the `Move` method provided none of the fields is just a number (requiring quoting to obtain the result) and it's position in the query's list of results is not the same as its field name (number).

proseLA:
i am not sure when this started to be a "problem"...  but it is definitely a problem.  and i'm unsure about the solution.
the problem, to me, is that the data is wrong.
there seems to be 2 solutions, 1-> remove the numeric keys and their associated values; 2-> correct the values associated with the numeric keys in the array.
i am not sure if anyone uses the array with the numeric keys, and if they did, i think we would have heard about this problem sooner.  because in my testing on v155 with php7.1, that data is wrong.  and i tried it with multiple tables.
whether this "problem" manifests itself or not is immaterial to me.  querying the db should not provide inaccurate results.

drbyte:
> querying the db should not provide inaccurate results.

Agreed. I'm still not convinced that the results are inaccurate. The code was written to specifically retrieve both the named and numeric indices. That doesn't make the existence of the numeric indices "inaccurate".

mc12345678:
> I'm still not convinced that the results are inaccurate. The code was written to specifically retrieve both the named and numeric indices. That doesn't make the existence of the numeric indices "inaccurate".

The "issue" is to some extent two parted.

Calling the `Move` method repetitively and incrementing the index would consistently provide both the associative and indexed data throughout. Calling just the `MoveNext` method after an initial query (that is not already cached with the associative and indexed data in the fields) will return only the associative data provided that no field starts with a number nor is a number itself. The "inaccuracy" comes when the `Move` method is first called (as in the `Rewind` method) which provides both the associative and indexed record data, followed by a `MoveNext` call that only updates the associative data, leaving the indexed data from the first record regardless the number of `MoveNext` calls that follow and any record that begins with or is only a number.

A solution to address the `MoveNext` continued presence of the indices or fields that begin with a number is to clear the record before making the assignment; however, that does not address the initial call to `Move` (as made by `Rewind`). Therefore for all records to respond as if the query were first made, the `Move` method needs some form of modification or the `Rewind` method needs to call a new method that returns the first record in the manner that all `MoveNext` method calls respond. Instead of adding an additional method, this commit modifies the capability of the `Move` method to default to its "original" design (earliest implementation found to be in ZC 1.2.0d as no other earlier 1.2.0 versions appear to be publicly available) and to offer the ability to ignore the fields that are ignored in the `MoveNext` method.

Even with that change, the `MoveNext` method appears to unnecessarily ignore fields that begin with a number regardless the length of the field name. Even ignoring a field that is just a number is slightly extensive as it is possible, but per the mysql "instruction" a field can be a number provided it is quoted (back-tick) which still offers its own issues if the existing `Move` method is to continue to permit return of the index of a field as the data response may not line up the field name with its position within the fields requested and therefore make the returned index portion of the list wrong and the associative portion would be unbalanced as compared to the indexed portion (ie. There would be one less associative record than indexed list for each field in the query that was named a number only.)

drbyte:
> followed by a `MoveNext` call that only updates the associative data

Okay. Got it.

What are the implications of just removing the exclusion done by `if (!preg_match('/^[0-9]/', $key)) {` in both `Move()` and `MoveNext()` ?

mc12345678:
> What are the implications of just removing the exclusion done by `if (!preg_match('/^[0-9]/', $key)) {` in both `Move()` and `MoveNext()` ?

First let's talk about the theoretical issue that could occur in open source:
if the `if (!preg_match('/^[0-9]/', $key)) {` was removed, then data from fields in tables that are named `1_life` or `2` or `100_balloons` could be returned and processed. Note though that `2` is a single number.  Now assume that a table has say 4 fields in it: `first_id`, `2`, `next_id`, `final_value`.  Then in a query the let's say three of the four values are to be returned where `first_id` just happens to be the primary key.  This would be performed most likely by the following:
`SELECT '2', next_id, final_value FROM new_table WHERE first_id = 5;`

The associative results would be: [2] = x, ['next_id'] = y, ['final_value'] = z
The indexed results would be [0] = x, [1] = y, [2] = x
with a complete "array" effectively of [2], ['next_id'], ['final_value'], [0], [1] (five elements) and note how the indexed result of [2] is the same as [0] because the index response key would be the same as the associative response key in one of the values.
Okay, so what if the order were "proper", then:
`SELECT next_id, final_value, '2' FROM new_table WHERE first_id = 5;`
The associative results would be: ['next_id'] = x, ['final_value'] = y, [2] = z 
The indexed results would be [0] = x, [1] = y, [2] = z
but the complete array would effectively be: ['next_id'], ['final_value'], [0], [1], [2] (again five elements); however, there would not be a definitive "third" associative key that is known.  In this case it may be considered to be [2] because there were two associative keys already provided and therefore all other "numbers" would be considered to be "properly" sequenced (if they were) and therefore the number of keys that follow the number of associative keys are also associative.  But, that only works in a properly formatted query where each element is in the sequence corresponding with the number and assumes that the results are actually in the order provided above.

So, what about the less theoretical "outlier".  If the restriction were removed, then every query and `MoveNext` would return both the associative and indexed result(s).  Looping on all results within a specific record would cause the value to be processed twice, also the amount of memory used would double to store each individual record.  If one were to attempt to attempt to process through each "field" of the query and specifically looking for/at the key, then if attempting to address the associative keys, the information would need to in some way be split "in half" assuming that none of the fields are number alone (for a single numeric field half of the quantity would result in a decimal number, for an even number of numeric fields, half of that quantity would not be respresentative of the associative field quantity as for each pair of numbers corresponding to an existing position within the query list two associative fields are lost while the number of index fields remains the same.  A feature that can be obtained by the current exclusion of fields beginning with or are a number would be lost in that it is not possible to then identify things like the total number of fields in the table (`SELECT * FROM this_table LIMIT 1`) identify the count of `fields` from that result and that identifies information about the table.  Also, if trying to loop through all of the fields in the result and possibly take action based on key=>value, occurs twice  unless the data is to be manipulated in a specific way when a specific key is encountered which could result in the associative and indexed data sets within a given query to have different values which if then also evaluated could cause different operations to occur than expected (as one value was updated but not the other "matching" value)... 

Again, this is based on taking blind action to the returned data set "trusting" that what is returned is understood/known and sensical (unlike the effect of the "theoretical" issue above).  Perhaps @lat9 could further identify an effect of consistently including both the associative and indexed results (albeit maintaining the indexed value(s) with the equivalent associative value(s) unlike how it currently works).

lat9:
What value are those numerical indexes?  They're unnamed and, if used, will result in code that's not particularly readable.

After a query like

```
      $product = $db->Execute("select pd.products_name, pd.products_description, pd.products_url,
                                      p.products_id, p.products_quantity, p.products_model,
                                      p.products_image, p.products_price, p.products_virtual, p.products_weight,
                                      p.products_date_added, p.products_last_modified,
                                      date_format(p.products_date_available, '%Y-%m-%d') as
                                      products_date_available, p.products_status, p.products_tax_class_id,
                                      p.manufacturers_id,
                                      p.products_quantity_order_min, p.products_quantity_order_units, p.products_priced_by_attribute,
                                      p.product_is_free, p.product_is_call, p.product_is_local_delivery, p.products_quantity_mixed,
                                      p.product_is_always_free_shipping, p.products_qty_box_status, p.products_quantity_order_max,
                                      p.products_sort_order,
                                      p.products_discount_type, p.products_discount_type_from,
                                      p.products_price_sorter, p.master_categories_id
                              from " . TABLE_PRODUCTS . " p, " . TABLE_PRODUCTS_DESCRIPTION . " pd
                              where p.products_id = '" . (int)$_GET['pID'] . "'
                              and p.products_id = pd.products_id
                              and pd.language_id = '" . (int)$_SESSION['languages_id'] . "'");
```
..., I'd much prefer to debug/modify code that used constructs like
```
if ($product['products_model'] != '') {
   ...
}
```
than
```
if ($product[5] == '') {
   ...
}
```
In addition, the inclusion of those numerically-indexed values essentially doubles the payload returned by the query, which could result in premature out-of-memory conditions when large report-related queries are run.

My preference would be to either remove those numeric keys altogether or, at least, provide a different function that returns them so that the values don't overload those large report-type queries.

drbyte:
> In addition, the inclusion of those numerically-indexed values essentially doubles the payload returned by the query, which could result in premature out-of-memory conditions when large report-related queries are run.

Fair point. But remember this is not a "new" thing. It's been returning the same payload since this class was written.

So, if we change from `mysqli_fetch_array()` to `mysqli_fetch_assoc()` to get the named columns only and therefore also the smaller payload, what will break?

lat9:
I'd never seen, albeit I never looked, those numeric keys returned before zc1.5.5.  My belief (i.e. I don't absolutely know) is that changing to get the named columns only won't break anything (or anyone!).

The concept of using `foreach` rather than `while (!$query->EOF)` has not, to my knowledge, been widely adopted by plugins so any affect would be seen first by the early-adopters (like the Zen Cart base code, itself).

mc12345678:
One thing that could help with memory (regardless of the use of the indices or not) would be to clear the `$this->fields` array prior to assignment at least within the `MoveNext()` method by changing:
````php
        foreach($zp_result_array as $key => $value) {
          if (!preg_match('/^[0-9]/', $key)) {
            $this->fields[$key] = $value;
          }
        }
````
to:
````php
        $this->fields = array();
        foreach($zp_result_array as $key => $value) {
          if (!preg_match('/^[0-9]/', $key)) {
            $this->fields[$key] = $value;
          }
        }
````

The reason for this additional action (perhaps should be addressed in a separate pull) is that internally as variables are assigned, PHP will generate an internal variable (using memory) to which values are assigned while maintaining the original variable before replacing the old value with the new value.  Thus if the memory used by the original variable required the same amount as the new variable, the total memory used at that point is twice the memory of either variable plus any additional overhead.  By assigning the variable to an empty array, the new assignment(s) would use only the memory of the new variable and any overhead. 

As to the preference of using the index as compared to the associative value, I could see that one may generate an array of fields to be queried:
````php
$field_array[] = 'pd.products_name';
$field_array[] = 'p.products_id';
// etc...
````
with the field list generated by imploding with a comma and then using the resulting data; however, I can also see how if that portion of the data were wanted and the change(s) of this PR are implemented that one could still obtain the indexed data within the `foreach` loop by calling the `Move()` method with the cursor of the current query result.  

A problem (difficulty?) that results by including the indexed array (or a field that begins with a number) would be seen in trying to work with the results as objects, such as `$obj_result->field_name`.  A variable/field that begins with a number (but it seems not itself a number) must be accessed by use of curly brackets around the quoted field_name. ie: `$obj_result->{'2e'}` which will return a result; however, it appears that `$obj_result->{'2'}` will return a Notice 'Notice:  Undefined property: stdClass::$2 in [...][...] on line XXX' and not work with the number.

So an added "impact" on removing the

drbyte:
I think the simplest is to:

1. change from `mysqli_fetch_array()` to `mysqli_fetch_assoc()`
and
2. remove the no-longer-necessary ` if (!preg_match('/^[0-9]/', $key)) {` filter.

The benefits here:
- smaller payloads from mysql, therefore smaller memory footprint
- less CPU due to less re-processing of results by looping thru keys
- fixes the bug/inconsistency caused by the `MoveNext` skipping updates because of the looping
- No core code has been relying on those numeric-indexed keys.

The cons:
- unknowns of what 3rd-party code might have been relying on numeric indices for looping, particularly those not manifesting problems because they're still using the old `while/!eof/MoveNext` constructs.

lat9:
Simple is good.

mc12345678:
> So, if we change from mysqli_fetch_array() to mysqli_fetch_assoc() to get the named columns only and therefore also the smaller payload, what will break?

If replaced in all locations (which would be necessary to fully support the `foreach` loop as otherwise written and not creating any additional functions) would result in the loss of ability to reference the data using an index method.  This appears to have no direct impact to the core code of ZC (meaning no code change necessary elsewhere within ZC); however, has the undisclosed affect of preventing the use of those indices in code that others have written but not gotten engaged in discussions similar to this for whatever reason.

As to payload, there are two payloads involved, the memory usage of a single row of data internal to the `MoveNext` method (assignment to `$zp_result_array`) and then the assignment to `$this->fields` if the key to `$zp_result_array` doesn't begin with a number (approximately half of the result being stored).

If the change is made basically only to `MoveNext` to use `mysqli_fetch_assoc` or `mysqli_fetch_array` (where `mysqli_fetch_array` would use  the `MYSQLI_ASSOC` parameter) then the ability to obtain the indexed result(s) would still be possible through the `Move` method if desired (having to add the proposed additional perimeter) and the `MoveNext` method would immediately use about 2/3 less memory if the result of that function is set directly to `$this->fields` and any related modification made to support setting the `EOF`.  Further reduction in memory use would be obtained if before assignment of the result to `$this->fields` that `$this->fields` was set to an empty array.

mc12345678:
> 2. remove the no-longer-necessary `if (!preg_match('/^[0-9]/', $key)) {` filter.

It never was "necessary", but was included/incorporated for a reason, possibly because of the issue(s) with the accessing the field(s) as an object and eliminating that concern by not permitting fields that begin with a number.

Simple is not always best.

> - fixes the bug/inconsistency caused by the MoveNext skipping updates because of the looping

fixes the bug/inconsistency caused by the `MoveNext` skipping updates because of the looping when `Move` is called before `MoveNext` as part of the current query processing. MoveNext called immediately after a query did not result in a bug or inconsistency.  Now with the implementation of `Rewind` (calling `Move`) as part of the `foreach` iterator there is this bug/inconsistency.

Just to be sure I understand the summary and perhaps make a rather blunt statement: instead of maintaining the current capability AND reducing processing resource usage, the thought is to unnecessarily remove existing capability and reduce the operational capability of Zen Cart while still (potentially) achieving the same reduction in processing resource usage because it is "easy".

drbyte:
> Now with the implementation of Rewind (calling Move) as part of the foreach iterator there is this bug/inconsistency.

Yes, I could have included the mention of `Rewind` to be more specific that that's the bug I was referring to. Not that we're talking about any other bugs here.

> Further reduction in memory use would be obtained if before assignment of the result to $this->fields that $this->fields was set to an empty array.

I'm open to that.

> because it is "easy".

Not because it's easy. But because it makes the code less complex, and therefore easier to maintain. That it happens to also be more performant as a by-product is a pleasant boon.

lat9:
`Simple !== Easy`

Coming from 30 years of commercial programming, keeping the code "simple" directly affects its cost of maintenance, as DrByte indicated.

mc12345678:
> 1. change from `mysqli_fetch_array()` to `mysqli_fetch_assoc()`

Understand also that this change will have the potential of increasing the need for help support for code that is written outside of the current base ZC code that uses number beginning fields.  Call it a risk if you like that may need to be addressed when/if query results are converted to objects and fields are used that begin with a number (though it appears that number only fields can not be accessed as an object).

>>     Further reduction in memory use would be obtained if before assignment of the result to $this->fields that $this->fields was set to an empty array.

> I'm open to that.

I'll make the change, though, having given an inch I plan to take a mile as relates to my other/further suggestion on the issue.


>>    because it is "easy".

> Not because it's easy. But because it makes the code less complex, and therefore easier to maintain. That it happens to also be more performant as a by-product is a pleasant boon.

I'll also say this, even beginning when I first started using ZC, outside reviews of ZC had not been "positive", some statements because of the "age" of the code, others because of how code was (or was not) managed.  Such reviews have changed and those reviewers appear to be providing more realistic points of view.  I hesitate on removing a capability just cuz, especially since I believe a few years ago I recognized, while coding, that the "additional" fields were returned when using the `Move` method and made specific coding decisions based on that instead of modifying or requesting the modification of the core code.  As far as I could see, it was *my* problem to deal with not something that caused a ZC problem.

I still think that although no such programmers have come out of the woodwork to identify the need/desire to allow `Move` method to continue to provide the entire dataset, that it could if not should still support that operation with a rewrite such as:
````php
  public function Move($zp_row, $assoc_only = false) {
    global $db;
    if ($this->is_cached) {
      if($zp_row >= sizeof($this->result)) {
        $this->cursor = sizeof($this->result);
        $this->EOF = true;
      } else {
        $this->fields = array();
        $this->fields = $this->result[$zp_row];
/*        foreach($this->result[$zp_row] as $key => $value) {
          $this->fields[$key] = $value;
        }*/
        $this->cursor = $zp_row;
        $this->EOF = false;
      }
    } else if (@mysqli_data_seek($this->resource, $zp_row)) {
      $this->fields = array();
      if ($assoc_only === false) {
        $this->fields = @mysqli_fetch_array($this->resource);
      } else {
        $this->fields = @mysqli_fetch_assoc($this->resource);
      }
/* Does $this->fields need to be evaluated for action if @ causes an issue that would have been "caught" within the foreach loop? */
/*      $zp_result_array = @mysqli_fetch_array($this->resource);
      foreach($zp_result_array as $key => $value) {
        $this->fields[$key] = $value;
      }*/
      $this->cursor = $zp_row;
      $this->EOF = false;
    } else {
      $this->EOF = true;
      $db->set_error(mysqli_errno($this->link), mysqli_error($this->link), $db->dieOnErrors);
    }
  }
````

And in my 36 years of commercial programming experience removing functionality without a technical reason is a disservice to the users (leading to loss of such users, poor reviews, and decline in public image) especially when describing a section of code that has exhibited the effects of the described problem for 13+ years.  I can only plead for those that don't know about this discussion or are hesitant to speak up for one reason or another. Ultimately the central maintainers own the code.

lat9:
Can you point to a plugin (or Zen Cart base code) that's making use of those numerical indexes?

I don't accept your assertion that removing those numerical indexes from the query-return is a disservice to users, as it's (a) reducing the payload/memory requirements and (b) ensuring that the results returned are "consumable" without intervention (like needing to weed those indexes out).

It would be one thing if the interface (and its limitations) were known and adopted by plugins (like the pre-zc155 `Move(0)+MoveNext()` requirement to "reset" a query), but I'm unconvinced that that scenario exists.

proseLA:
i'm a bit confused.  to summarize:
- there is a bug, that has been there for 13+ years, which returns wrong data in the numeric indexes portion of a returned array.
- the suggested solution is to remove the numeric indexes part of the array.
- this solution has the potential to break some plugins, that make use of the numeric indices for looping; but those plugins do NOT make use of the data contained in the numeric indices as that has been wrong for 13+ years (or whatever time period).
- a further fringe scenario exists, where someone has created a field within a table whose field name is a number; that data returned for that field, in all likelihood has also been wrong.
- someone needs to speak up for those developers who are ignorant of this thread.  although, we are unsure if there are any in-use plugins for v155 that exhibit this behavior.

did i miss anything?  other than the memory payload implications?

hard to imagine, but i think we should correct the numeric indices and keep them.  i can imagine a store's worst nightmare.  they have a small problem; post it on the board; some person on the board says you really need to upgrade; they upgrade; and now their crucial plugin no longer works.  

i truly do not feel strongly about it.  but i do feel strongly about not returning inaccurate data; which is the way it stands now.  removing the numeric indices is one way to solve this problem, if it ends up breaking someone's store, it keeps IT people in business.

is v160 still slated for using laravel?  if so, i suppose new plugins will not have this problem...  

best.

proseLA:
wow.  i ran some 'simple' tests, here is what i found:

- on a test v155 site, i did what drByte suggested here:

change from mysqli_fetch_array() to mysqli_fetch_assoc()
remove the no-longer-necessary if (!preg_match('/^[0-9]/', $key)) { filter.

and some basic testing suggested to me that everything worked.  this included the lat9 basic test, as well as a working test ZC site with significant modifications.  and i could not create a problem.  in addition, i added 2 fields to a table, 1 field started with a number, the 2nd field's complete name was a number.  no problems.

- in looking at that modified table, i can safely say that anyone that has fields numbered so, out in the wild, using current code could get some pretty problematic results.  especially with regard to the test field that was solely a number. 

- i then replaced the complete file with the new modified file in this PR.  errored out.  ton of warnings with regards to array_replace and php fatal error on duplicate field name.  but, alas, this was a v155 site and a v156 query_factory.

- i configured and booted up a basic v156 site and replaced the query_factory one with the code contained in this PR.  does not work.  at all.  again, a ton of warnings with the array_replace, but no fatal errors.  but the data returned is wrong.  by a mile.  referencing back the simple test from the discussion on the forum, where we retrieve 2 records and look at them, there is just a bunch of numbers.  no data.  

ok.  this was tested on a working machine capable of running v160, but i did notice it was running php5.6.  i will try 1 last test and get php7.1 up and running and see if it works, but my current testing suggests that the code in this PR does not work.

proseLA:
ok.  last test: with a v156 shop running php7.1, this modified version of query_factory returns similar garbage.

perhaps someone else could test and confirm?

my bottom line: 
- i could not get the code in this PR to return anything useful.
- i'm curious if this code was in fact tested; or perhaps i have done something wrong.
- i found no problems with removing the numeric indices from the query_factory in a working v155 installation that contains numerous plugins and modified code.  i accomplished this by implementing the drByte's suggestions.
- after these tests, i am now of the opinion to remove the numeric indices from the return of the array.  

best.

mc12345678:
> ok. last test: with a v156 shop running php7.1, this modified version of query_factory returns similar garbage.
>
> perhaps someone else could test and confirm?

The latest commit to the associated branch (not pushed to any of the other branches) has been tested and as provided herein did/does not work requiring additional modification (preliminarily pushed elsewhere; however, also requires additional work to capture the currently proposed improvements). There is nothing about this modification that is PHP 7.x specific.  In fact, the addition of the `foreach` iterator could have been fully incorporated starting in Zen Cart 1.3.9, but that is a completely separate topic.  Further, it is because this commit is still in revision that it was not force merged to the initial functional commit that provides a `foreach` result that mitigates the originally expressed problem.

>
> my bottom line:
>
> -   i could not get the code in this PR to return anything useful.
> -   i'm curious if this code was in fact tested; or perhaps i have done something wrong.

Yes the latest commit was tested and found to be problematic, but has not yet been updated here.  If anything I preemptively included development code in the main branch of this pull request instead of branching off of it like I did just after that. 

> -    i found no problems with removing the numeric indices from the query_factory in a working v155 installation that contains numerous plugins and modified code. i accomplished this by implementing the drByte's suggestions.

As previously stated, core Zen Cart code (with core database data or pretty much for that matter at this point any existing functional plugin that has not yet incorporated the use of `foreach` to iterate query results) would not be affected by permitting the use of fields that begin with a number or by simply dropping the array that is returned when issuing `Move(X)` or as specifically applied against `foreach`, `Move(0)`.  In fact core code operates just fine with no modification at all.

> -    after these tests, i am now of the opinion to remove the numeric indices from the return of the array.

What is it about "these tests" had any impact on whether to continue to support operational capability that has existed for 13+ years? I don't see how these "tests" have any impact what-so-ever on the decision to support operations that have existed for others to use (whether packaged for public use or not).  Code originally proposed supported the removal of the results that affected the newly incorporated `foreach` method.

I've refrained from mapping out all of the use capabilities because of the time involved with documenting them; however, it appears necessary to incorporate now. Note, the examples below demonstrate the capability of the software and the demonstrative code offers a way to execute the task not that the specific task is the important part (ie. use of a date field is an example field that could be used):

Use of the `Move(x)` method to return data not available by the `MoveNext()` method:

````php
// $additional_filtering defined to obtain refined results in perhaps a sorted sequence, limited number of items, etc... or no additional filtering if left as empty quotes.
$result = $db->Execute("SELECT countries_name, countries_id, countries_iso_code_2, countries_iso_code_3 FROM " . TABLE_COUNTRIES . $additional_filtering);

// Perform operations using the $result data.

// Desire to reuse the query's results which may include using fields that begin with a number or are a number without querying again nor using the cache.
$result->Move(0);
while (!$result->EOF) {
  // Process data from $result->fields which includes all associative fields and the indices for the results as well as fields that begin with a number (if such have been added to the table beyond the ZC default fields).
  // This process could also use/add additional array elements to the fields array as described below.
  if ($result->fields[2] == $result->fields[3]) {
    // Perform action if third and fourth fields are the same, though not exactly the same ('0' and false would evaluate equivalent)
  } 


  $result->Move($result->cursor + 1);
}
````

Use of the `fields` array as part of processing records to not obliterate all record data (ie. sustain data through each iteration as currently supported): 
````php
$result = $db->Execute("SELECT products_id, products_date_added FROM " . TABLE_PRODUCTS);

$selected_date = CONFIGURATION_CUT_OFF_DATE; // Value possibly stored in the database or otherwise defined.

// Perform operations using the $result data.
while (!$result->EOF) {
  // Process data from $result->fields which uses data from the query to populate fields within the result set.

  if ($result->fields['products_date_added'] < $selected_date) {
    $result->fields['total_before'] .= $result->fields['products_id'] . '_';
  } else {
    $result->fields['total_at_and_after'] .= $result->fields['products_id'] . '_';
  }
  // Potentially take action on the added field(s).
  // Use the fields from the array to perform some operation.

  $result->MoveNext();
}

// Potentially take action on the added field(s).
````

Using the new `foreach` iterator to also retrieve the data available in the `Move(X)` method:
````php
// $additional_filtering defined to obtain refined results in perhaps a sorted sequence, limited number of items, etc... or no additional filtering if left as empty quotes.
$result = $db->Execute("SELECT countries_name, countries_id, countries_iso_code_2, countries_iso_code_3 FROM " . TABLE_COUNTRIES . $additional_filtering);

// Perform operations using the $result data.
foreach ($result as $key =>$value) {

  $all_data = $db->Move($result->cursor);

  if ($all_data->fields[2] == $all_data->fields[3]) {
    // Perform action if third and fourth fields are the same, though not exactly the same ('0' and false would evaluate equivalent)
  } 
  // Use the fields from the array(s) to perform some operation.
}
````

I believe there are other usage capabilities including if the `queryFactoryResult` class were modified to provide some form of default/modified `fields` data.

mc12345678:
Okay, looking for suggestion/guidance on how to present the variations that have been developed.

Tested with 5 different versions of the query_factory.php class running ZC 1.5.6 with strict reporting on PHP 7.1.15, MySQL 5.6.34, Linux 3.2.61-grsec-modsign
1) Original file.
2) Original file modified to use mysqli_fetch_assoc for all cases of mysqli_fetch_array in the file and removal of the if statement (keeping the for loop surrounding it).
3) Original file modified to use mysqli_fetch_assoc for all cases and no for loop for assignment, just to assign the value directly to the variable where able.
4) Modifications made basically as I had described supporting the existing capability but without the identified problem. 
5) Similar to the above but used various methods of clearing data before assignment.

Long and short of it, Generally speaking with repetitive testing the worst case was changing the assignments to all just mysqli_fetch_assoc and not incorporating some additional form of memory clearing before assignment. (Used twice as much memory as any of the previous results).
While most the rest had equivalent impact on memory (within perhaps 1 kilobyte of one another in both memory used and peak memory used.) Though only one really offers the flexibility of the existing operation with little to no observable difference in operation but also includes supporting some additional features if other parts of the store software were overloaded as needed for whatever reason.

At any rate, I'd like to offer/present the various sets of code but without things getting too convoluted and looking for a suggestion of method to do that in a "meaningful" way.  Therefore others can test/try/review the code and either come up with a similar conclusion or find something else that may have been missed.

proseLA:
option 2.  it is what drByte suggested 20 days ago.  it is what @lat9 would prefer.

there are 5 replacements of mysqli_fetch_array with  cases of mysqli_fetch_assoc.  and i count 5 removals of the if filter.

it removes the numeric indices and it keeps all added numeric named fields.  it returns correct data.

i have tested the code done that way and could not find a problem.  let's put this to bed and move on.

drbyte:
Let's go ahead with the switch to mysqli_fetch_assoc, and also include the memory/performance improvements @mc12345678 has been championing here.

mc12345678:
To attempt to be clear of the way proposed forward:

> Let's go ahead with the switch to mysqli_fetch_assoc, 

That at any time the database is queried using ZC built-in queries that the only data returned is from the associated style of query (whether one has chosen to use the `$db` to "hide" a value or not and that there is to be no way to query using ZC code to retrieve the indexes. That if such data were to be gathered it would require using raw database queries. Further that this change is beyond the original reason for this pull request which was to specifically address iterator functionality. (conversation in some cases has been specifically about the iterator operation and in some cases about general query operation. I want to be sure that this latest proposed direction is to be all inclusive and not just about the iterator.)

drbyte:
> Let's go ahead with the switch to mysqli_fetch_assoc,

> That at any time the database is queried using ZC built-in queries that the only data returned is from the associated style of query

Yes.

>  and that there is to be no way to query using ZC code to retrieve the indexes. That if such data were to be gathered it would require using raw database queries.

Yes, since Zen Cart core code only utilizes associated array values. At this point I'm not bothered about any 3rd-party code that's operating on integer values, because the odds are extremely rare that they're doing that, and if they are then they are getting questionable results if they've half-embraced the new foreach() syntax anyway. They can, and should, update their code for compatibility.

>  Further that this change is beyond the original reason for this pull request which was to specifically address iterator functionality.

Er, well, the title of this post has always been about iterator functionality.

If you feel strongly about moving the small changes for memory-improvement to a separate PR, I"m okay with that.

mc12345678:
Had me up until:
>and if they are then they are getting questionable results if they've half-embraced the new foreach() syntax anyway

If they were previously using `MoveNext()` in conjunction with `Move()` and transitioned to `foreach()`, they would not need to make any other changes than what is necessary to support `foreach()` and the code would still work as it did.  Under the path directed, other changes would be necessary either in the code leading up to the loop or within the loop itself (such as using customized code).  Alternatively they would need to implement their own version of the `query_factory` class which by some is disclaimed as hacking up the code on one hand but is done regularly on the other. 

>> Further that this change is beyond the original reason for this pull request which was to specifically address iterator functionality.
> Er, well, the title of this post has always been about iterator functionality.

Whether correctly referenced or not, the post has been about addressing the `foreach` functionality as an iterator because of its observed effects.  It calls the `Move()` method which will need rewriting to support the overall direction advised for the `query_factory` class to remove all possibility of returning the query result index.  This is instead of either modifying the `Move()` method to work well with `foreach` (first commit of this PR) or adding an additional method that would only use `mysqli_fetch_assoc` to `reset` the query results to the beginning. It is this ultimate truncation of functionality from all query results that I was referring as being beyond the original reason for the pull request.  I see the memory "improvements" as integral with making this change as they are not only expected to be improvements to the operations, but also drive the final usage/layout.

While I do not agree with the direction, the direction has been set.  I will prep the code to meet the proposed direction (and likely document the alternatives for future review) as soon as I can.

drbyte:
> or adding an **additional method** that would only use `mysqli_fetch_assoc`

It's this additional method (and adding additional parameters to several other methods) that I specifically wish to avoid.

drbyte:
I've updated the PR with a couple fixes:
- pre-declare the `$fields` property as an `array` so `array_replace` has an array to start with
- re-use the `$zp_result_array`

And removed the int-fallback.

PR #1634, opened 2018-03-01
v160 Remove integer key data from iterator results

mc12345678:
In ZC 1.5.5, the iterator class was implemented allowing the use of foreach
to cycle through query results.  To support this operation, the rewind
module was written to use the Move($position) method which was written to
allow query results to be returned as both a key=>value response as well
as a position=>value response.  Therefore, the initial reset of the query
results would produce an array of both the key=>value pair and position=>value
pair.  Then when the MoveNext method was used, the key=>value pair would
be updated; however, the position=>value pair remained constant.

This commit adds continued support for code that intentionally used the
Move($position) to access the integer based field result and to also
prevent adding the integer based results to the Move($value) method returned
values.

A similar condition was identified in the ExecuteRandomMulti function.

Incorporation of this modification added a parameter to each function with
a default value that causes the function to operate as it has for all of
ZC 1.5.x.

This was ultimately an issue because mysqli_fetch_array (and the legacy
mysql_fetch_array) when not provided a resulttype as described in the
php manual (http://php.net/manual/en/mysqli-result.fetch-array.php) defaults
to providing both the integer based result and the field key based result.

Perhaps this was done to support some yet older code; however, with the
implementation of the iterator class to provide only field results
(consistently at each next result) a modification was needed at least against
the foreach implementation.

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -299,7 +299,7 @@ class queryFactory extends base {
         return $this->Execute ($zf_sql, false, false, 0, true);
     }
 
-  function ExecuteRandomMulti($zf_sql, $zf_limit = 0, $zf_cache = false, $zf_cachetime=0, $remove_from_queryCache = false) {
+  function ExecuteRandomMulti($zf_sql, $zf_limit = 0, $zf_cache = false, $zf_cachetime=0, $remove_from_queryCache = false, $int_key = true) {
     $this->zf_sql = $zf_sql;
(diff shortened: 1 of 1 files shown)


mc12345678:
The provided code change comes from a ZC forum discussion found here: https://www.zen-cart.com/showthread.php?223658-db-SELECT-returning-numeric-indexes

Also, this PR is similar to ones submitted for ZC 1.5.6 #1633 and ZC 1.5.5 #1635.

drbyte:
Closing - not planning to retain numeric indices.

PR #1639, opened 2018-03-03
Move media define

Zen4All:
It is more logical to add the define to the defined_paths.php file af this is the location of all paths.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1639)


drbyte:
I'm undecided about doing this, mainly because it's part of example product-types, vs root core functionality.

Also, is the same define set on the Admin side? If so, then similar change should be done there.

Also, your "if" condition is testing for the wrong constant (copy/paste error).

Zen4All:
:blush:  yes it was done in a hurry. I'll check the admin,  fix the if, and point taken on the sample product type.

PR #1644, opened 2018-03-11
Zero-weight/freeshipping Admin Alert might not appear

lat9:
The message "Warning: 0 weight is configured for Free Shipping and Free Shipping Module is not enabled" is intended to warn the admin that the setting _Configuration->Shipping/Packaging->Order Free Shipping 0 Weight Status_ is enabled but the `freeshipper` shipping module is not enabled.

Unfortunately, the message does **not** currently appear if `freeshipper` is installed but disabled.  This PR makes that correction in both the admin's configuration.php and modules.php scripts.


drbyte:
Good catch. Thanks. 👍

PR #1645, opened 2018-03-12
[v156] Freshen Docs Styling

dakanji:
Misc Other Updates:
1. Update Zen Cart Website Protocol to https
2. Added placeholders for some v1.5.6 files to be finalised post release
3. Updated instructions dating from v1.3.x about renaming Admin folder and so on


proseLA:
is it possible to move away from in-line css to 1 css file that gets imported for all of the instruction pages?

dakanji:
This commit was supposed to do what I think you are suggesting ... Can you clarify?

barco57:
well, that is the point of this pull request...
Though I find myself wondering why oxygen.css?
The admin is using bootstrap, the zc_install is using foundation, and now adding a third css framework for another part...

dakanji:
Oxygen.css is just a name I gave the css file and is not a framework albeit it draws on a subset of Hydrogen css >> https://github.com/pimbrouwers/hydrogen ... hence the play on words as in breathe some oxygen into the old staid styling the docs have. 

Could just as well have called it docs.css or anything else.

barco57:
ah, I thought it had something to do with http://oxygencss.com/

dakanji:
Aha! I thought I was being original with the name. 

Well, if the changes are fine in principle to the devs, the name of the css file can be changed to a boring but unambiguous "docs.css" or similar.

proseLA:
i'm sorry.  my error.  i could blame it on 4/20, but my comment was from 4/19....

i admit, i also looked up the oxygen css as well...  but my comment was clearly misinformed, as i must have been looking at the diff incorrectly.

dakanji:
Changed the name of the css file to "zen.css"

dakanji:
Got too messy

PR #1646, opened 2018-03-18
Ez page updates

Zen4All:
- remove double setting of parameters
- remove parameters that are defined twice
- minor layout update
- some bug fixes (missing closing form tag, and typo in button class)

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1646)

### DIFF
--- a/admin/ezpages.php
+++ b/admin/ezpages.php
@@ -288,7 +288,7 @@ if (zen_not_null($action)) {
         ?>
 
         <div class="form-group">
-          <div class="col-sm-12"><?php echo (($form_action == 'insert') ? '<button type="sumit" class="btn btn-primary">' . IMAGE_INSERT . '</button>' : '<button type="sumit" class="btn btn-primary">' . IMAGE_UPDATE . '</button>') . ' <a href="' . zen_href_link(FILENAME_EZPAGES_ADMIN, (isset($_GET['page']) ? 'page=' . $_GET['page'] . '&' : '') . (isset($_GET['ezID']) ? 'ezID=' . $_GET['ezID'] : '')) . '" class="btn btn-default" role="button">' . IMAGE_CANCEL . '</a>'; ?></div>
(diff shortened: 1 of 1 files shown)


scottcwilson:
👍  I thought "sumit" was the "Southern variant" of submit. :)

admin/ezpages.php
```diff
@@ -283,186 +282,102 @@ function init() {
...
+          <div class="col-sm-9 col-md-6"><?php echo zen_draw_input_field('pages_title', htmlspecialchars($ezInfo->pages_title, ENT_COMPAT, CHARSET, TRUE), zen_set_field_length(TABLE_EZPAGES, 'pages_title') . ' class="form-control"', true); ?></div>
         </div>
         <div class="form-group">
             <?php echo zen_draw_label(TABLE_HEADING_PAGE_OPEN_NEW_WINDOW, 'page_open_new_window', 'class="col-sm-3 control-label"'); ?>
-          <div class="col-sm-9">
-            <label class="radio-inline"><?php echo zen_draw_radio_field('page_open_new_window', '1', $is_page_open_new_window) . TEXT_YES; ?></label>
-            <label class="radio-inline"><?php echo zen_draw_radio_field('page_open_new_window', '0', $not_page_open_new_window) . TEXT_NO; ?>
+          <div class="col-sm-9 col-md-6">
+            <label class="radio-inline"><?php echo zen_draw_radio_field('page_open_new_window', '1', ($ezInfo->page_open_new_window == 1 ? true : false)) . TEXT_YES; ?></label>
```
mc12345678: Interested in understanding and curious, is there a particular reason/thought behind using this style: 
````php
($ezInfo->page_open_new_window == 1 ? true : false)
````
instead of this style:
````php
($ezInfo->page_open_new_window == 1)
````
?

Zen4All: I used this style, because it is clear what it does. 
The really shorthand style is not clear enough for me.

mc12345678: Makes sense to me and also provides one an easy way to modify the "default" setting if they wish to do so.  Sorry to bother you about it and thank you for responding.

Zen4All: I'm open for discussion, it is open source 😃 .
Maybey @zcwilt, @ajeh  , or @drbyte  can have a vote in this. As this form pf code is used in multiple files, like banner_manger.php , and collect_info.php
I Have no problem in changing the code is that is what the majority wants

drbyte: My preference is the shorter syntax: `($ezInfo->page_open_new_window == 1)`, since it returns a boolean true/false, which is exactly what is needed in that position.

While one could also remove the surrounding parentheses, that affects readability even more, so you'll see that we very often include those, even though this isn't an order-of-operations case where parentheses are usually implemented.

Zen4All: Ok, i'll make it the short version.

drbyte:
Thanks!

PR #1647, opened 2018-03-21
v156-notifiers-for-base-functions

lat9:
Add notifiers to various "base" functions, enabling observers to modify (or in some cases replace) those functions' processing and return values.  This will enable many plugins to remove the need for core-file overwrites, allowing for an easier Zen Cart upgrade path.

Please note that many of the notifier strings are in use by various plugins; please don't change the names and/or "payload".

### DIFF
--- a/includes/functions/functions_lookups.php
+++ b/includes/functions/functions_lookups.php
@@ -157,10 +157,26 @@
 */
   function zen_get_products_stock($products_id) {
     global $db;
+    
+    // -----
+    // Give an observer the chance to modify this function's return value.
+    //
     $products_id = zen_get_prid($products_id);
+    $products_quantity = 0;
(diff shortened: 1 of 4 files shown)


drbyte:
I'm a little worried about the reliance on notifiers in `functions_prices.php`, since the v1.6.0 version of that file has overhauled how pricing data is generated, now producing an array of values and a subtemplate which can be used to style the output.

lat9:
The "payload" for the v1.6.0 notifications could be modified from its v1.5.6 version to provide the same interface, but supplying the array-element values.  That would keep the interface consistent and downward-compatible.

drbyte:
Will merge for now. But this will need review for future compatibility at some point.

PR #1652, opened 2018-04-03
 Fix: checkout_shipping: No method chosen after selected shipping is …

mc12345678:
…no longer available

Changes in available shipping methods can occur for a number of reasons, shopping cart items added/deleted, prices changed, coupons applied, etc... When a previously selected shipping option is no longer available, ZC reports that change; however, does not then force a new selection.  As a result, the previously available selection (now removed from the options) remains internally active and could result in incorrect shipping fees to be collected (too high or too low).

This commit will stage the shipping option selection to select a new lowest price.


drbyte:
Thanks

PR #1670, opened 2018-04-23
Include store-defined message in all order confirmation emails

scottcwilson:
Also provide notifier for additional customization.


drbyte:
Thanks!

PR #1671, opened 2018-04-26
Correct FirstData HCO price hash comparison

mc12345678:
As identified in this forum thread:
https://www.zen-cart.com/showthread.php?223791-Payeezy-HCO-HASH-error
payments appear to be collected; however, not recorded in ZC because
the hash value comparison does not pass.  Per technical support when
performing the hash comparison, the total price of the purchase is to
be incorporated into the hash with two decimal places applied.

The current code will round to 2 decimal places but if the value ends with
a zero, then that zero is dropped from right to left up to the decimal
point. This corrects that specific comparison.


drbyte:
Thanks!

PR #1675, opened 2018-04-29
Undefined offset $this->data[$category_id]

Zen4All:
When a category has no siblings, $this->data[$category_id] is not set

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1675)


drbyte:
Thanks

PR #1676, opened 2018-04-29
Trying to get property of non-object in /includes/classes/payment.php

Zen4All:
php is complaining about not used modules.
PHP Notice:  Trying to get property of non-object in /includes/classes/payment.php on line 172
PHP Notice:  Undefined index: freecharger in //includes/classes/payment.php on line 172

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1676)


drbyte:
Thanks

PR #1678, opened 2018-05-01
v156-correct-order-notifier-parm

lat9:
Throwing a debug-log on PHP 7.1+:
```
PHP Notice:  Undefined variable: products in C:\xampp\htdocs\zc156\includes\classes\order.php on line 242
```

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -239,7 +239,7 @@ class order extends base {
 
       $this->info['tax_groups']["{$this->products[$index]['tax']}"] = '1';
 
-      $this->notify('NOTIFY_ORDER_QUERY_ADD_PRODUCT', $product, $index);
+      $this->notify('NOTIFY_ORDER_QUERY_ADD_PRODUCT', $this->products, $index);
 
       $index++;
       $orders_products->MoveNext();


drbyte:
Did you mean to pass the entire `$this->products` array, instead of just the current product?

I don't see any additional context in the original PR: https://github.com/zencart/zencart/commit/3905d0792140c2e65d5793210c2ddc2a70e43425

lat9:
I'll be darned if I can find the customization that uses this notification; it's probably "good enough" to just pass `$this->products[$index]`.  Shall I resubmit, or will you take care of the change?

drbyte:
I'll update it.

drbyte:
It also makes me wonder about the purpose of passing `$index` in that I can't see why someone would want to modify that value, or what they'd do when reading it?  I suppose the entire `$this->products` array is accessible inside the observer though, so maybe it'd use that to read prior entries.

mc12345678:
I could see the passing of the `$index` (editable or not) to be used to evaluate the product that was just added (and as indicated to be able to evaluate where within the list of products processing is currently) and editable to allow removal of the item(s) from the list, to skip the next `X` product, etc... 

Passing `$this->products[$index]` supports passing the current product but provides little context within the entire list of products.  Without the context, one may have to cycle through all product to find the one that matches the desired condition of matching the latest product.  Passing just `$index` as a parameter (first or later) would support determination of `$this->products[$index]` without specifically passing `$this->products[$index]`. `$this->products` is accessible from within the observer and with `$index` then so is `$this->products[$index]`.

scottcwilson:
I suspect `$this->products[$index]` was the original intent. Thanks for catching this, Cindy.

PR #1679, opened 2018-05-02
v156 Eliminate some of the strict error reports

mc12345678:
Unsure of which if any of these have already been reported/fixed because
of the number of items identified.  Advise which to remove or request
to rebase which should remove any already incorporated at that time (even
if it means for me to remove merge errors).

### DIFF
--- a/admin/includes/modules/update_product.php
+++ b/admin/includes/modules/update_product.php
@@ -15,7 +15,7 @@ if (isset($_GET['pID'])) {
 }
 if (isset($_POST['edit_x']) || isset($_POST['edit_y'])) {
   $action = 'new_product';
-} elseif ($_POST['products_model'] . $_POST['products_url'] . $_POST['products_name'] . $_POST['products_description'] != '') {
+} elseif ((isset($_POST['products_model']) ? $_POST['products_model'] : '') . (isset($_POST['products_url']) ? implode('', $_POST['products_url']) : '') . (isset($_POST['products_name']) ? implode('', $_POST['products_name']) : '') . (isset($_POST['products_description']) ? implode('', $_POST['products_description']) : '') != '') {
   $products_date_available = zen_db_prepare_input($_POST['products_date_available']);
   $products_date_available = (date('Y-m-d') < $products_date_available) ? $products_date_available : 'null';
 
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -273,8 +273,8 @@ class order extends base {
                                    from (" . TABLE_CUSTOMERS . " c, " . TABLE_ADDRESS_BOOK . " ab )
                                    left join " . TABLE_ZONES . " z on (ab.entry_zone_id = z.zone_id)
                                    left join " . TABLE_COUNTRIES . " co on (ab.entry_country_id = co.countries_id)
-                                   where c.customers_id = '" . (int)$_SESSION['customer_id'] . "'
-                                   and ab.customers_id = '" . (int)$_SESSION['customer_id'] . "'
+                                   where c.customers_id = " . (!empty($_SESSION['customer_id']) ? (int)$_SESSION['customer_id'] : 0) . "
+                                   and ab.customers_id = " . (!empty($_SESSION['customer_id']) ? (int)$_SESSION['customer_id'] : 0) . "
                                    and c.customers_default_address_id = ab.address_book_id";
 
     $customer_address = $db->Execute($customer_address_query);
@@ -287,8 +287,8 @@ class order extends base {
                                    from " . TABLE_ADDRESS_BOOK . " ab
                                    left join " . TABLE_ZONES . " z on (ab.entry_zone_id = z.zone_id)
                                    left join " . TABLE_COUNTRIES . " c on (ab.entry_country_id = c.countries_id)
-                                   where ab.customers_id = '" . (int)$_SESSION['customer_id'] . "'
-                                   and ab.address_book_id = '" . (int)$_SESSION['sendto'] . "'";
+                                   where ab.customers_id = " . (!empty($_SESSION['customer_id']) ? (int)$_SESSION['customer_id'] : 0) . "
+                                   and ab.address_book_id = " . (!empty($_SESSION['sendto']) ? (int)$_SESSION['sendto'] : 0);
 
     $shipping_address = $db->Execute($shipping_address_query);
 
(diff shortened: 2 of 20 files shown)


includes/classes/shopping_cart.php
```diff
@@ -745,11 +745,11 @@ function calculate() {
           $sale_maker_discount = '';
 
           // bottom total
-          if ($attribute_price->fields['product_attribute_is_free'] == '1' and zen_get_products_price_is_free((int)$prid)) {
+          if (!$attribute_price->EOF && $attribute_price->fields['product_attribute_is_free'] == '1' and zen_get_products_price_is_free((int)$prid)) {
```
drbyte: Would it be safe to move these `! EOF` checks up to line 739 as a single `if ($attribute_price->EOF) continue;` ?

includes/classes/order.php
```diff
@@ -272,8 +272,8 @@ function cart() {
                                    from (" . TABLE_CUSTOMERS . " c, " . TABLE_ADDRESS_BOOK . " ab )
                                    left join " . TABLE_ZONES . " z on (ab.entry_zone_id = z.zone_id)
                                    left join " . TABLE_COUNTRIES . " co on (ab.entry_country_id = co.countries_id)
-                                   where c.customers_id = '" . (int)$_SESSION['customer_id'] . "'
-                                   and ab.customers_id = '" . (int)$_SESSION['customer_id'] . "'
+                                   where c.customers_id = '" . (int)(!empty($_SESSION['customer_id']) ? $_SESSION['customer_id'] : 0) . "'
+                                   and ab.customers_id = '" . (int)(!empty($_SESSION['customer_id']) ? $_SESSION['customer_id'] : 0) . "'
```
drbyte: For these cases I think the `(int)` should be moved inside, like:

```diff
- (int)(!empty($_SESSION['customer_id']) ? $_SESSION['customer_id'] : 0) 
+ (!empty($_SESSION['customer_id']) ? (int)$_SESSION['customer_id'] : 0) 
```

admin/includes/modules/update_product.php
```diff
@@ -12,7 +12,7 @@
   if (isset($_GET['pID'])) $products_id = zen_db_prepare_input($_GET['pID']);
   if (isset($_POST['edit_x']) || isset($_POST['edit_y'])) {
     $action = 'new_product';
-  } elseif ($_POST['products_model'] . $_POST['products_url'] . $_POST['products_name'] . $_POST['products_description'] != '') {
+  } elseif ((isset($_POST['products_model']) ? $_POST['products_model'] : '') . (isset($_POST['products_url']) ? implode('', $_POST['products_url']) : '') . (isset($_POST['products_name']) ? implode('', $_POST['products_name']) : '') . (isset($_POST['products_description']) ? implode('', $_POST['products_description']) : '') != '') {
```
drbyte: Using `implode()` here is an interesting approach. Mind clarifying your "why"?

mc12345678: Ultimately in a multi-language environment, the fields `products_url`, `products_name`, and `products_description` all are passed as an `array` to include the language identifier to finish off the posted information.   (Even in a single language situation, the source code will still present the language identifier for each of those fields.) 
So, instead of evaluating the language(s) implemented (doesn't add value to further evaluate knowing that there is always a language identifier), determining whether the value has been passed as an array (may be an issue in future PHP versions) and knowing that the `implode()` function will generate a text item representing the single text item or the list of items separated by the `glue` (an empty quote), combining that data to evaluate the presence of a character fits soundly. Perhaps some other comparison (regex?) of that combination would be a better fit but if someone wants to store a blank or space, why should that be prevented? 

But, the idea was to use the existing approach in a stricter php manner based on the type(s) of input expected.

drbyte: Thanks. I like it!

includes/classes/order.php
mc12345678: Wasn't sure how best to "support" the upgrade process and also had a "concern" about someone choosing to use some non-integer style variable in lieu of the 0.  By placing the `(int)` up front, seemed sounder for such individual modification; however, also see that the 0 sufficiently meets the expectation.

I don't mind moving the `(int)` to inside, leaving any modifications that may be made to be properly implemented by the modifier/user.

includes/classes/shopping_cart.php
mc12345678: So couple of things.
First, I had not stepped back to truly evaluate the possibility of "merging" these changes into a single `continue` operation; however, I had considered if that evaluation should be done and the excuse I identified at the time was, what if there were notifiers incorporated in this loop that truly expected to be executed (because previous versions supported stepping through each line of the loop).  There currently are no such notifiers in this area and the apparent purpose of this code section supports that simplification, but that was a reason that I "released" these changes for review.  It appears though that in absence of that possibility/concern that the current code supports the recommended simplification of continuing through the loop if there is no price data associated. (All occurrences of adding to the total resolve to adding 0 to an existing value).

I do not really have a preference either way though I also am not sure about modifications made by others to this code/area of code...

mc12345678:
Unable at moment to squash and rebase.  Will do so later unless independently addressed.

Note that one of the issues that was addressed in admin/includes/modules/update_product.php still exists in admin/includes/modules/update_product_meta_tags.php. Though that file was not previously touched in this issue, the file has similar code that has been removed from admin/includes/modules/update_product.php and is missing the same check(s) that were corrected in this commit. (Ie. there appear to be more places to address strict notifications.)

Specifically changing:
````php
  $sql_data_array = array( 
      'metatags_title' => zen_db_prepare_input($_POST['metatags_title'][$language_id]), 
      'metatags_keywords' => zen_db_prepare_input($_POST['metatags_keywords'][$language_id]), 
      'metatags_description' => zen_db_prepare_input($_POST['metatags_description'][$language_id])); 

````
to:
````php
  $sql_data_array = array( 
      'metatags_title' => isset($_POST['metatags_title'][$language_id]) ? zen_db_prepare_input($_POST['metatags_title'][$language_id]) : '', 
      'metatags_keywords' => isset($_POST['metatags_keywords'][$language_id]) ? zen_db_prepare_input($_POST['metatags_keywords'][$language_id]) : '', 
      'metatags_description' => isset($_POST['metatags_description'][$language_id]) ? zen_db_prepare_input($_POST['metatags_description'][$language_id]) : ''); 
````

I do not suggest using a !empty() test as the text itself could be (for example) '0' which would evaluate to empty and cause the '0' to be converted to two single quotes with nothing between them.

drbyte:
Thanks!

PR #1680, opened 2018-05-09
Fix error found during review for strict errors.

mc12345678:
Haven't reviewed past installations to determine how this came to be different; however, the maximum was to be compared against `$j` instead of `$n`. That said, almost wonder if foreach should be used instead.


drbyte:
Thanks

PR #1681, opened 2018-05-09
Capture more changes for strict error prevention.

mc12345678:


drbyte:
Can a bunch of these, like `$_SESSION['layoutType']` and `$flag_disable_right` etc, just be fixed by actually declaring them?  Several of those are already taken care of in the `main_template_vars.php` master file, so shouldn't be an issue.

drbyte:
Thanks

PR #1685, opened 2018-05-14
Fix errors where old sites are missing new email css file

scottcwilson:
Don't try to pull in css file if it doesn't exist.


drbyte:
Thanks

lat9:
Using E_USER_ERROR on the `trigger_error` function will result in (er) a fatal error.  Did you want to change that to use E_USER_WARNING?

The focus of this PR (I think) was to provide continued operation in the absence of that file.

scottcwilson:
That's a really good idea.  Submitted.

PR #1693, opened 2018-05-24
Remove some unused code from shipping estimator module

scottcwilson:
Looks bigger than it is because I removed the control-Ms as well.


drbyte:
thanks

PR #1694, opened 2018-05-24
Change Salutation (gender) and DOB default status to false

scottcwilson:
To help GDPR compliance and stop shopowners from collecting information they don't really need, false would be a better default for both the salutation and DOB field.


drbyte:
Agreed.
A lot of settings are "on by default" in order to demonstrate what features are built-in, which eliminates a lot of questions about "can it do this?". Early research suggested people were able to grok functionality by seeing it enabled first, and then simply explored whether they could turn it off.

But I agree, GDPR etc are changing the landscape, somewhat for the good, but now requires merchants to work harder.

Thanks for the submission.

PR #1695, opened 2018-05-27
Make order shipping_cost float 

scottcwilson:
Casting the value as float prevented some php 7.1 error logs


drbyte:
Here's my reason: if there's something earlier in the chain that's setting a non-numeric value into that variable, it'd be better to fix that in case there are other knock-on effects (such as unexpected string values, even if empty string instead of 0), etc.

drbyte:
In chatting with @zcwilt I think changing the `isset()` to `! empty()` and removing the `(float)` is probably a safer "fix" or catch-all for buggy (read: less-strictly-written) shipping modules. Floats can cause some penny-rounding issues.
And, since I think the issue you were needing to "fix" was probably an 'empty string' or a non-existent array element, the `empty()` test is adequate to cover both cases.

zcwilt:
:+1:

Issue #1697, opened 2018-05-30
PHP Warning thrown during admin currency_cron processing


lat9:
I'm seeing the following logged each time the /admin/currency_cron.php script runs:
```
PHP Warning:  include(/site_root/admin/includes/languages/english): failed to open stream: No such file or directory in /site_root/admin/includes/init_includes/init_languages.php on line 38
```
That comes from the following code-block
```php
// include the language translations
  require(DIR_WS_LANGUAGES . $_SESSION['language'] . '.php');
  $current_page = basename($PHP_SELF);
  if (file_exists(DIR_WS_LANGUAGES . $_SESSION['language'] . '/' . $current_page)) {
    include(DIR_WS_LANGUAGES . $_SESSION['language'] . '/' . $current_page);
  }
```
... where line 38 is that last `include`.  The issue's root-cause is the fact that `$current_page` isn't defined during that cron processing, so the `file_exists` function's return is based on the existence of the language **directory**, not a **file**.

The issue can be corrected by using `is_file` instead of `file_exists`.

drbyte:
I suspect setting `$PHP_SELF` in `currency_cron.php` may also resolve it:

```php
$PHP_SELF = __FILE__;
```

Granted, perhaps `is_file` may be a more universal fix.

lat9:
Which one do you prefer?  I'm staging a PR to use `is_file` right now.

PR #1698, opened 2018-05-30
#1697-v156-correct-cron-warning

lat9:
Update the languages' initialization script to use `is_file` instead of `file_exists`.


drbyte:
Thanks!

Fixes #1697

PR #1700, opened 2018-06-04
Prevent exceeding length of file search developers toolkit.

mc12345678:
If item sought is near the end of the file and the search attempts to view more lines above and below the found line, then the default store would attempt to find data beyond the length of the file and result in a strict PHP notice, this corrects that issue.


drbyte:
Thanks

PR #1701, opened 2018-06-04
Remove duplicate define for sales-report-graph

mc12345678:
…es.php

That definition is not needed on the admin side because of its equivalent catalog definition.

Found via strict related reporting where reporting was set to not report this type of issue; however, still came through for some reason.  Removing the define on the admin or conditionally defining it on the catalog side resolved the issue.


drbyte:
Thanks. 
I removed the commented-out lines, as they're probably not serving much additional value. 
The only reason I can see for keeping them would be for merging into older versions, but even then, as you say, it's already in `filenames.php`

PR #1702, opened 2018-06-04
Strict error corrections

mc12345678:
Found some more items reported when operating in strict mode as have continued to use v1.5.6 and running various tests.


drbyte:
Similar to #1681 can some of these be fixed by just declaring the values earlier in the code workflow?

drbyte:
Thanks

PR #1703, opened 2018-06-04
Correct html validation and tag alignment.

mc12345678:
I understand that perhaps the individual product type file is going to go away and that issue #1631 already has some items, but for some reason it did not seem to have this modification so am including it here and will move as requested.  Didn't want to disrupt review of that pull request.


drbyte:
This seems to be redundant now that #1604 has been merged.

PR #1705, opened 2018-06-06
Fix typos from previous refactoring

Zen4All:
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1705)


drbyte:
Good catch.

PR #1706, opened 2018-06-07
Add dash and spaces to admin categories pulldown for readability

Zen4All:
- added  a dash and spaces between the categorie name and id
- clean up query -> removed single quotes 
- used foreach in favour of while loop

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1706)


drbyte:
Thanks!

PR #1669, opened 2018-04-20
[v156] Freshen Docs & "Install Suggested" Styling

dakanji:
- Freshens the styling on the "Install Suggested" page which is currently has a very dated look.
- Same style similarly used to freshen files in docs folder
- Makes docs mobile friendly

Misc Other Updates:
- Update Zen Cart Website Protocol to https
- Added placeholders for some v1.5.6 files to be finalised post release
- Updated instructions dating from v1.3.x about renaming Admin folder and so on

All css directives contained in ./docs/zen.css.
zen.css is a 3.0kb tweaked subset of Hydrogen.css (https://github.com/pimbrouwers/hydrogen, Copyright 2018 Pim Brouwers and Licensed under MIT).

Replaces #1645

PR #1707, opened 2018-06-08
[v156] Updates to #1669 (Documentation files)

dakanji:
Minor Tweaks and add missing changes in "Install Suggested" page.

Before: https://www.screencast.com/t/mEceXJGG87i
After: https://www.screencast.com/t/whmYahWLiG

Updates #1669

### DIFF
--- a/docs/0.about_zen_cart.html
+++ b/docs/0.about_zen_cart.html
@@ -1,10 +1,10 @@
-!DOCTYPE html>
+<!DOCTYPE html>
 <!--
 	/**
 	 * @package documentation
-	 * @copyright Copyright 2003-2016 Zen Cart Development Team
-	 * @license https://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
-	 * @version $Id: Author: DrByte  Fri Jan 1 17:56:22 2016 -0500 Modified in v1.5.5 $
+	 * @copyright Copyright 2003-2018 Zen Cart Development Team
+	 * @license https://www.zen-cart.com/license/2_0.txt GNU Public License v2.0
+	 * @version $Id: 0.about_zen_cart.html  Modified in v1.5.6 $
 	 */
 	//-->
 <html>
@@ -23,6 +23,7 @@
 	<link rel="stylesheet" href="./zen.css">
 </head>
 <body>
+
 	<section id="pageHeader">
 		<div class="h-g">
 			<div class="h-u-md-6-12 h-u-sm-1-12">
@@ -31,7 +32,7 @@
 				</a>
 			</div>
 			<div class="h-u-md-6-12 h-u-sm-hidden h-text-right">
-				<a class="h-btn" href="./index.html">
+				<a class="h-btn h-box-shadow" href="./index.html">
 					Documentation Home
 				</a>
 			</div>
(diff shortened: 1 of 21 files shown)


drbyte:
Which app are you using that's doing the code-formatting and indentation? 
(I'm not convinced I like all of it, but I'm not interested in squabbling about it either. Just curious your tool of choice.)

dakanji:
Hi,

I mainly use Smultron: https://www.peterborgapps.com/smultron. I see from the link that it is now a paid app but I downloaded it about 10 years ago when it was free and still use it.

It is just an advanced text editor which I chose because it allows for regex based searches. It doesn't indentations and any issues from that are purely mine in applying my preferred style of coding HTML. I find this easier to read but understand everyone has their preferences. I tend to like tabs and for PHP, the one true brace style but everyone has a preference.

When starting from a lump of non-indented HTML, I first run it though jsbeautifier (http://jsbeautifier.org/), which also does HTML, and select tab indents.

I also have Github's more powerful Atom: https://atom.io which can auto indent with psr styles (which I dislike because of the "anti tab" stance they take) and such but I generally just use Smultron with jsbeautifier.

Any particular issues with the output you want changed?

dakanji:
I now understand why you mentioned squabbling. Debate between Tabers and Spacers can get heated. Unfortunately, Spacers seem to control the damned standards - based on archaic considerations if I may say .... but let's not squabble!

As for the tag per line thing, it is just something I do for readability, as I miss a fair few things especially when dealing with code from others, and for me to properly understand complex child/parent relationships. Also think it helps reviewers see exactly what has been done. Not a hard and fast thing as there are a few tags not so treated.

drbyte:
Squabbling aside ... Spacers rule!  :D
(That said, I wish the "standard" was "2 spaces", not "4 spaces" for common indentation. Of course, Tabbers love that they can configure that in their editor and not worry about it. I get it. Really I do. But I still hate tabs. :D )

Smultron -- that was the first code-editor I used after switching to Mac years ago. Used it for a long time until I needed more. I think it was unmaintained for awhile, and "some feature" I needed was offered in another app, so I switched. I don't remember what "to" anymore.

Nowadays I use Sublime Text and phpStorm. Tons of power in these apps.

dakanji:
So I managed to royally screw up the squashing attempt on the other PR ... fingers crossed for this one!

dakanji:
Looks like it worked ... Used command line instead of sourcetree

drbyte:
👍

dakanji:
@drbyte, 

< Squabble >
Since we had a little chat about Tabs and Spaces, I had been worrying looking around wondering if the mad spacers had totally taken over despite all reason screaming that the old adage "Tab to Indent, Space to Align", from back in the typewriter days, makes the most sense.

I find that the "hatred" of tabs always invariably comes from trying to use tabs to align, which never works, and then deciding all tabs are bad and to be avoided.

Anyway, it was a relief to find one sane beacon out there ... the Wordpress Coding Standards: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation.

It was particularly gratifying to see the old adage paraphrased as a rule of thumb. 

Reason lives on!
< /Squabble >

PR #1709, opened 2018-06-11
Add missing quotes to array key in coupon_restrict

Zen4All:
The array key is missing the single quotes around the word "text". It is apparently now a major bug, as the code also seems to work fine without them. So no need to fix in earlier versions

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1709)


drbyte:
Ya, newer PHP versions are getting more and more strict about things like this.
Thanks

PR #1712, opened 2018-06-11
Preserve address_book_id through notification

### LINKED ISSUE #1711: Preserve added address_book_id through notification
As identified in [this ](https://github.com/lat9/vat4eu/issues/2)VAT4EU issue, the notifier `NOTIFY_MODULE_CHECKOUT_ADDED_ADDRESS_BOOK_RECORD` is issued just after creating a base address-book record ... but that new address-book-id isn't preserved over the notification.

Since most observers of that notice are going to be performing database activity (and rendering the module's subsequent call to `$db->Insert_ID();` invalid -- it returns 0).

lat9:
... so it's preserved over the notification, assuming that an observer is likely to be performing additional database actions of its own.

Fixes #1711


drbyte:
Thanks!

PR #1716, opened 2018-06-19
Fixes #1667 - $fields variable is already initialized

### LINKED ISSUE #1667: $fields is already empty
I this code block the second $new_fields (line 906) is not needed. It is already set to be empty.
https://github.com/zencart/zencart/blob/8100fcb709853a0e4bf668dfb2c9e28adf3b329d/admin/orders.php#L901-L916

and this line should have a dot in front of the = like $new_fileds .=
https://github.com/zencart/zencart/blob/8100fcb709853a0e4bf668dfb2c9e28adf3b329d/admin/orders.php#L928

scottcwilson:
Removed second initialization (and a block of commented out code); changed subsequent set of variable to use concatenation instead of just assignment.


drbyte:
Thanks.

PR #1717, opened 2018-06-20
Remove commented code which causes bracket matching to fail

scottcwilson:


drbyte:
meh -- may not merge this

Which editor are you using that ignores code comments and thus gets confused by commented-out-brackets/braces?

scottcwilson:
vim.  What's the issue with merge?

includes/classes/order.php
```diff
@@ -781,7 +781,6 @@ function create_add_products($zf_insert_id, $zf_mode = false) {
           //            $this->products[$i]['stock_value'] = $stock_values->fields['products_quantity'];
 
           $db->Execute("update " . TABLE_PRODUCTS . " set products_quantity = '" . $stock_left . "' where products_id = '" . zen_get_prid($this->products[$i]['id']) . "'");
-          //        if ( ($stock_left < 1) && (STOCK_ALLOW_CHECKOUT == 'false') ) {
```
drbyte: Here, please just remove the bracket. The alternate logic should remain.

includes/modules/attributes.php
```diff
@@ -242,7 +242,6 @@
                       break;
                       case '3':
                       $tmp_attributes_image_row++;
-                      //                  if ($tmp_attributes_image_row > PRODUCTS_IMAGES_ATTRIBUTES_PER_ROW) {
```
drbyte: Here, remove the bracket and leave the line. Serves as a reference for an alternate approach.

The others below can be removed.

Issue #1673, opened 2018-04-27
zc156: Attributes Controller, can't add new attribute


lat9:
With an install of the demo products, I created a copy of the _**Microsoft Internet Keyboard**_ product and went to the Attributes Controller to add some attributes for testing.

I selected the "Color" option, the "Black" color and pressed the `Insert` button.  Upon click, the screen returned to the main Attributes Controller display with the "Please select a category ..." message.

Looking in the /logs directory, I found the following log:
```
[27-Apr-2018 10:36:07 America/New_York] PHP Notice:  Undefined offset: 2 in C:\xampp\htdocs\zc156\admin\includes\modules\products_previous_next.php on line 95
[27-Apr-2018 10:36:07 America/New_York] PHP Notice:  Undefined index: products_id in C:\xampp\htdocs\zc156\admin\attributes_controller.php on line 252
[27-Apr-2018 10:36:07 America/New_York] PHP Notice:  Undefined index: products_id in C:\xampp\htdocs\zc156\admin\attributes_controller.php on line 263
[27-Apr-2018 10:36:07 America/New_York] PHP Notice:  Undefined index: products_id in C:\xampp\htdocs\zc156\admin\attributes_controller.php on line 379
[27-Apr-2018 10:36:07 America/New_York] PHP Notice:  Undefined index: products_id in C:\xampp\htdocs\zc156\admin\attributes_controller.php on line 381
[27-Apr-2018 10:36:07 America/New_York] PHP Notice:  Undefined index: current_category_id in C:\xampp\htdocs\zc156\admin\attributes_controller.php on line 381
```

On inspection of the database, a `products_attributes` record was added ... with a `products_id` of **0**.

lat9:
Yes, unfortunately that error "note" was made to an issue that was closed around 6 months ago; that's why I didn't see it!

Zen4All:
You can't keep track of everything

Zen4All:
I'll make a pr with the fix this week

drbyte:
Closed by #1720

PR #1531, opened 2017-10-23
Reduce complexity of code and refresh look of Attributes Controller

Zen4All:
These are the changes done to the attributes controller so far.
Maybe I have overdone it, but I tried to reduce the complexity of the file, ans also refresh its looks.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1531)

Issue #1511, opened 2017-09-21
What is the future of attributes_controller.php


Zen4All:
I am writing a new module for Zen Cart, in which the product collect info page is completely rewritten. The data is presented in tabs, and should be more user friendly. (see: [Zen-Cart-Collect-Info-Through-Tabs-In-New-Style](https://github.com/Zen4All-nl/Zen-Cart-Collect-Info-Through-Tabs-In-New-Style)). This also brings the possibility of adding extra info/fields to the page by dropping in extra tabs to a special folder..

Which brings me to the title of this issue.

On inspecting the code of the admin/attributes_controller.php file:
- I see a lot of html errors
- The html table structure is a mess, some tags are not closed or opened, or are in the wrong location in the code.
- Use of hard-coded input fields, instead of  using the zen cart functions
- Database query's that ask for all information in a table where only 3 or 4 field of the 27 are needed.

This can easily be "repaired", but I would like to know if I should spend time on it, or wait for Zen Cart 1.6

drbyte:
I agree: It needs some work. 
I'd welcome a PR that cleans it up a bit. 

If it's ready before the end of the month, you can PR it against the `v156` branch.

Zen4All:
👍 I'll see what I can do

mc12345678:
FYI, the last time I tried to correct the html validation on the admin/attributes_controller.php page for ZC 1.5.5, it broke the responsive nature of the display and pushed content out across the display instead of keeping it "tight".  At that point I didn't feel like doing a comparison between the modified version and a non-modified version to correct the associated issue(s), but look forwards to your ideas.

Zen4All:
Well, like I said I was doing some major rework for my own module ![example](https://github.com/Zen4All-nl/Zen-Cart-Collect-Info-Through-Tabs-In-New-Style/blob/master/Documentation/images/new%20look.png), so I already located the problems 😄 .

The biggest problem is that some tables are not closed, and row are not properly opened.

Zen4All:
This will not be ready before the end of the moth.

Zen4All:
Looking through the file I see no way of deleting/removing a download form an attribute.
Once a file is assigned to an attribute the only way to remove it, is to delete the attribute. Leaving the filename filed empty has no effect, because there is a zen_not_null check.

I am going to change this in the update function

Zen4All:
https://github.com/Zen4All/zencart/blob/v156_attributes_controller/admin/attributes_controller.php

[...]

PR #1720, opened 2018-07-05
Fix missing product_id, current_category_id, and attribute_id

Zen4All:
Attributes were not saved properly due to missing product_id, current_category_id, and attribute_id. also see #1673, #1531, #1511

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1720)


drbyte:
Thanks!

PR #1723, opened 2018-07-09
[v156] Refine Error Heading Display

dakanji:
The install index page always says there are issues even when the only thing that shows up is a 'Pro Tip'. This PR adds logic to only show the header saying there is an issue when there are actual issues itemised as opposed to tips.


drbyte:
Thanks!

PR #1725, opened 2018-07-10
[Dev] Align Nginx Config with v156

dakanji:
This aligns the Nginx Config in Dev by bringing in edits done when back porting the config to v156. In some cases this just means indicating that the changes were introduced in v1.5.6.

### DIFF
--- a/zc_install/includes/languages/lngEnglish.php
+++ b/zc_install/includes/languages/lngEnglish.php
@@ -100,14 +100,14 @@ define('TEXT_ADMIN_SETUP_NEWSLETTER_OPTIN', 'Opt In: ');
 define('TEXT_PAGE_HEADING_COMPLETION', 'Setup Finished');
 define('TEXT_COMPLETION_HEADER_MAIN', '');
 define('TEXT_COMPLETION_INSTALL_COMPLETE', 'Installation completed.');
-define('TEXT_COMPLETION_INSTALL_LINKS_BELOW', 'You can now access your store front and admin backend using the links below.');
+define('TEXT_COMPLETION_INSTALL_LINKS_BELOW', 'You can now access your Storefront and Admin are using the links below.');
 define('TEXT_COMPLETION_UPGRADE_COMPLETE', 'Congratulations, your upgrade is now complete.');
 define('TEXT_COMPLETION_ADMIN_DIRECTORY_WARNING', 'Your admin directory could not be renamed automatically, you will need to rename your admin directory before accessing it');
 define('TEXT_COMPLETION_INSTALLATION_DIRECTORY_WARNING', "Please delete the 'zc_install' folder now");
 define('TEXT_COMPLETION_INSTALLATION_DIRECTORY_EXPLANATION', "To prevent unauthorized persons from overwriting your installation, administering your installation will be disabled until the folder has been deleted.");
 
-define('TEXT_COMPLETION_CATALOG_LINK_TEXT', 'Your Store Front');
-define('TEXT_COMPLETION_ADMIN_LINK_TEXT', 'Your Admin Backend');
+define('TEXT_COMPLETION_CATALOG_LINK_TEXT', 'Your Storefront');
+define('TEXT_COMPLETION_ADMIN_LINK_TEXT', 'Your Admin Area');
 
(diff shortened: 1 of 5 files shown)


zc_install/includes/languages/lngEnglish.php
```diff
@@ -179,11 +182,11 @@
 define('TEXT_ERROR_CURL', 'Problems with the CURL extension - PHP is reporting that CURL is not present.');
 define('TEXT_ERROR_UPLOADS', 'Upload Extension for PHP is not enabled');
 define('TEXT_ERROR_XML', 'XML Extension in PHP is not enabled');
-define('TEXT_ERROR_GZIP', 'Gzip Extension in PHP is not enabled');
+define('TEXT_ERROR_GZIP', 'The GZip Extension in PHP is not enabled<br>[ <i><b>NOTE:</b> This may not be relevant if you are using Nginx and handling GZip within Nginx.<i> ]');
 define('TEXT_ERROR_EXTENSION_NOT_LOADED', '%s extension does not seem to be loaded');
 define('TEXT_ERROR_FUNCTION_DOES_NOT_EXIST', 'PHP function %s does not exist');
 define('TEXT_ERROR_CURL_LIVE_TEST', 'Could not use CURL to contact a live server');
-define('TEXT_ERROR_HTTPS', 'PRO TIP: If possible you should already have installed an SSL certificate, and run the installer using https://');
+define('TEXT_ERROR_HTTPS', 'PRO TIP: If possible, install an SSL certificate and run the installer using https://');
```
mc12345678: This rewording doesn't capture the fact of an SSL certificate already being installed.  In that condition the message would imply to the user that it is not installed and needs to be installed.  If it is installed and this message appears the issue may have simply been that they did not access the installer using https:.

Suggest that if going to modify this text that some more consideration be given to rewording.

dakanji: Is it not that case that tip only shows up when it is detected that HTTPS is not being used?

mc12345678: Sorry for the late response.  Regarding the detection and the verbiage being changed from:
> PRO TIP: If possible you should already have installed an SSL certificate, and run the installer using https://

to:
> PRO TIP: If possible, install an SSL certificate and run the installer using https://

The checks performed are to first see if the site was accessed using https:.  If it wasn't, then if curl is present (to support subsequent checking) it is to then attempt to access the site using https with the same uri as attempted.  If that is successful then to go to the site using https to perform the install and basically that ends the concern.  

However, if the site were accessed without https (although it has been installed) and curl was not operational (curl issue is addressed as a separate statement), then this error message would be displayed.  Problem with that scenario is that the SSL certificate would have been installed already (although the new verbiage says it needs to be installed) and would be incorrect.  Alternatively, if the https: method of accessing the site is not by changing http: to https: only, then the statement of the need to install a SSL certificate could be wrong as well... Example:

Site name: mydomain dot com
Site is hosted on a shared domain type account where an individual SSL certificate has not been purchased.  The method to connect securely is: https: // www dot hosteddomain dot com /mydomain.  There are several other possible variations of that arrangement.  Again in this scenario, there is no specific SSL certificate that must be installed to support https communication.  It would be preferred from a business perspective, but https would still be available.  Basically then, at least in those situations the statement that a SSL needs to be installed would be false and indicate somewhat from a security standpoint that it doesn't know what is going on and raises doubt in the faith of its proper operation. 

Lastly and unfortunately more frequent than desired, the SSL may have been installed, but there is some other reason that it is not working correctly or as expected.  Reasons may include server configuration, partial installation, etc... 

By directing to install the SSL instead of suggesting that it may or may not be installed, that it is a good idea to do so and further to attempt to access the installation using https: (using whatever method/criteria needed to access it) the software is not wrong about the condition.  The software just is unable to specifically identify what it takes to establish SSL style communication. 

In absence of establishing communication the logic and program flow is suggesting that because https was not used, that it would be good to be used and at the least to know/understand that such usage requires some level of installation.  Ultimately installation or further resolution may be something done by the individual, by the host, etc...

dakanji: Hi, can't say I read the entire missive lol, but the edit looks reasonable.

drbyte:
Thanks

PR #1726, opened 2018-07-11
micropayments for pp express Fixes #1618

### LINKED ISSUE #1618: paypal express and micropayments.
includes/modules/payment/paypal/paypal_curl.php:526

AMT does not look to be part of paypal express.  i recommended a hack as seen in the code here:

https://www.zen-cart.com/showthread.php?223396-can-t-switch-between-PayPal-and-Micropayments-accounts&p=1341528#post1341528

in thinking about it, i'm not sure it is better to add the AMT field in the express module.  not sure if that will get sent to paypal or not.  as i have not really done much work with paypal.  

to me, it makes sense to provide support for all paypal modules for micropayments.

thoughts?

proseLA:
i branched against v156.


drbyte:
Thanks!

PR #1730, opened 2018-07-17
Correct attributes' display "randomness" - Fixes #1729

### LINKED ISSUE #1729: zc156: Attributes can be "randomized" on order-class query
Since the `order` class' `query` method doesn't specify an `ORDER BY` clause when retrieving an order's products' attributes, the attributes can be displayed in a random (based on MySQL caching) order.

lat9:
... by specifying the sort-order for the query.


drbyte:
Thanks

PR #1737, opened 2018-08-09
fix #1736

### LINKED ISSUE #1736: v156 New product not added to correct category
@Zen4All,
 
With the new edited categories/products code somewhere before or near the commit: https://github.com/zencart/zencart/commit/2e7ddadcb9fe4e67d4130906f6aac6772d357e2b
it appears that if have navigated to a category then selected to add a product, the product ends up being added to the root of the store rather than the current category.  It must then be moved to the desired category to place it where desired.

Historically, once a user has navigated to a category and then selects to add a product, the category at that point becomes the master category for the product.  That does not appear to be happening now.

Zen4All:
With the new edited categories/products code somewhere before or near
the commit: 2e7ddad
it appears that if have navigated to a category then selected to add a
product, the product ends up being added to the root of the store rather
than the current category. It must then be moved to the desired category
to place it where desired.

Historically, once a user has navigated to a category and then selects
to add a product, the category at that point becomes the master category
for the product. That does not appear to be happening now.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1737)

### DIFF
--- a/admin/includes/modules/document_general/collect_info.php
+++ b/admin/includes/modules/document_general/collect_info.php
@@ -130,13 +130,13 @@ for ($i = 0, $n = sizeof($tax_class_array); $i < $n; $i++) {
 </script>
 <div class="container-fluid">
     <?php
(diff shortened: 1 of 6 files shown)


drbyte:
Thanks!

PR #1739, opened 2018-08-18
Show errors created during db test connection

scottcwilson:
Attempted install from scratch with admin/includes/local/configure.php containing 

define('ADMIN_BLOCK_WARNING_OVERRIDE', '1');

This breaks the install, but no error is given because the installer is looking at `$errorList['newDatabaseCheck']` which is not set.


drbyte:
Thanks!

PR #1743, opened 2018-08-22
v156 fix incorrect variables in admin/ezpages.php

Zen4All:
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1743)


drbyte:
Thanks!

PR #1748, opened 2018-08-25
MySQL 5.7 requires order by fields to be specified in select

scottcwilson:


scottcwilson:
Should have been more specific: when DISTINCT is used, this is required.   Here's the error: 

[25-Aug-2018 08:31:17 America/New_York] PHP Fatal error:  3065:Expression #1 of ORDER BY clause is not in SELECT list, references column 'purelypoultry.c.sort_order' which is not in SELECT list; this is incompatible with DISTINCT :: SELECT DISTINCT c.categories_id, cd.categories_name FROM categories c, categories_description cd WHERE c.categories_id = cd.categories_id AND c.parent_id= '0' AND cd.language_id='1' AND c.categories_status='1' ORDER BY c.sort_order, cd.categories_name; ==> (as called by) /Users/scott/Sites/xxx/includes/templates/purely2014/templates/tpl_modules_categories_tabs.php on line 22 <== in /Users/scott/Sites/xxx/includes/classes/db/mysql/query_factory.php on line 167

PR #1750, opened 2018-09-02
PHPMailer 5 handler backward-compatibility break

mc12345678:
The _autoload function has been deprecated in PHP 7.2.x, unfortunately this
file is associated with a larger project, but in PHP 7.2.x this caused an
issue.

### DIFF
--- a/includes/classes/vendors/PHPMailer/PHPMailerAutoload.php
+++ b/includes/classes/vendors/PHPMailer/PHPMailerAutoload.php
@@ -37,13 +37,13 @@ if (version_compare(PHP_VERSION, '5.1.2', '>=')) {
     } else {
         spl_autoload_register('PHPMailerAutoload');
     }
-} else {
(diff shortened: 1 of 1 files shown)


Zen4All:
For php 7.2 tha lates version of phpmailer is required https://github.com/PHPMailer/PHPMailer/releases/tag/v6.0.5
from the phpmailer readme

> Legacy versions
PHPMailer 5.2 (which is compatible with PHP 5.0 - 7.0) is no longer being supported for feature updates, and will only be receiving security updates from now on. You will find the latest version of 5.2 in the 5.2-stable branch, and future versions of 5.2 will be tagged with 5.2.x version numbers, so existing Composer configs should remain working. If you're using PHP 5.5 or later, we recommend you make the necessary changes to switch to the 6.0 release.

drbyte:
Will merge; however, will also upgrade PHPMailer to v6, since it is compatible with PHP 5.5.0+ which is the current-stated-minimum PHP ZC 1.5.6 aims to support.

PR #1751, opened 2018-09-02
Modify handling of version check errors to provide text only results

mc12345678:
the array $data['error'] itself can be an array, a boolean, or text.
The use of print_r outputting as text will allow the full set of data
to be pushed which in this case is a triggered error.


drbyte:
Thanks

PR #1752, opened 2018-09-02
Align product type code across all types

mc12345678:
While reviewing product types for potential strict reporting issues,
identified that other product types did not appear to properly capture
the $products_manufacturer information which led to other aspects in
need of update.


drbyte:
Grr - I've always hated this blatant duplication of code. It's a bear to maintain.
Thanks.

PR #1753, opened 2018-09-02
Remove extra quotes from integers in address lookup query

mc12345678:
Remove a strict notice from zen_address_format and other in file cleanups.

While addressing the strict notice report also adjusted sql statement(s)
to form the sql statement to not have quotes when assigning integers.


drbyte:
Thanks

PR #1754, opened 2018-09-02
Push page direction after initial view to next page.

### LINKED ISSUE #1747: v156 When adding a non-basic product can not preview info.
@Zen4All 
 Changes made to support #1736 did not fully resolve the ability to add product to the starting category for product that are not the default general product.  

Now, after filling in information about the product and attempting to select the preview button, the form attempts to load the product type admin file for the chosen product type which has been removed from the overall fileset.

In each product type's `collect_info.php` file, recommend changing `$type_handler` to `FILENAME_PRODUCT` in the new_product zen_draw_form located at or around line 168.

mc12345678:
As part of the new product management admin design, the specific file
for each product type has been removed.  To move forward in processing
non-basic product types, the associated form needs to redirect to
an existing "controller" which is the FILENAME_PRODUCT location.

Fixes #1747.

### DIFF
--- a/admin/includes/modules/document_general/collect_info.php
+++ b/admin/includes/modules/document_general/collect_info.php
@@ -130,7 +130,7 @@ for ($i = 0, $n = sizeof($tax_class_array); $i < $n; $i++) {
 </script>
 <div class="container-fluid">
     <?php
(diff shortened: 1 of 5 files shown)


drbyte:
Thanks

PR #1756, opened 2018-09-02
Strict error corrections

mc12345678:
Also revised the looping of sql results to align with the new available
process.  Removed some duplicative operations and ensured variables expected
to be available later have a value assigned.


drbyte:
Thanks

PR #1757, opened 2018-09-07
Show message on die in case zen_redirect fails

scottcwilson:
In my cart, the redirect is failing which means the customer sees a blank page with no message.  This just adds the message in.

### DIFF
--- a/includes/modules/payment/paypalwpp.php
+++ b/includes/modules/payment/paypalwpp.php
@@ -2946,7 +2946,7 @@ if (false) { // disabled until clarification is received about coupons in PayPal
           if ($response['L_ERRORCODE0'] == 10422 || $response['L_ERRORCODE0'] == 10486) {
             header("HTTP/1.1 302 Object Moved");
             zen_redirect($this->ec_redirect_url);
-            die();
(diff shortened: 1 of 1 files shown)


drbyte:
Curious: why is the redirect failing? Is something wrong in that part too?

scottcwilson:
Not sure why - I suspect because there was data sent prior to the call.  That's generally what causes it to fail.

drbyte:
Oh. “Headers already sent”?

While a more detailed “die” is a possible (albeit poor) fallback, the root cause of the broken redirects is worth solving.

scottcwilson:
Right.  The trouble is this is a hard to duplicate problem.  I'm not sure what causes Paypal to get into this state.  Are the headers sent because PayPal is in debug mode perhaps?

scottcwilson:
possibly the cause of the failure is ipn_main_handler.php line 85-87? 

```
<html>
Processing...
</html>
```

scottcwilson:
Or perhaps it's a `setcookie` call that was done.   In any event, the person who coded this knew this could happen, which is why a `die()` was added after the header.

drbyte:
The person who coded this knew PayPal can be flakey. They also knew that the "cause" could be varied hence the non-specific (empty) output. 

Will merge; however my gut tells me there's something else that ought to be corrected if headers-already-sent is happening on that specific site.

PR #1758, opened 2018-09-10
Use styling the $EXTRA_INFO block

scottcwilson:
Styling exists for this block in email_common.css but it was not applied in any of the templates.


drbyte:
Thanks!

PR #1759, opened 2018-09-11
removed underscore from tiny_mce

llynix:
I'm the maintainer of the tinymce plugin.  It comes in a folder without an underscore.  Removing the underscore from the main code will make things a little easier to maintain.


drbyte:
Thanks

PR #1761, opened 2018-09-11
Email-css fix

drbyte:
Sometimes the CSS file wasn't being processed, due to order of operations logic


drbyte:
Hmmm ... still needs better fallback logic...

mc12345678:
What about:
````php
// initialize array element.
$block['EMAIL_COMMON_CSS'] = '';

// if $css_lang_folder is a language or empty quote and the file exists then assign.
if (file_exists(DIR_FS_EMAIL_TEMPLATES . $css_lang_folder . 'email_common.css')) {
  $block['EMAIL_COMMON_CSS'] = file_get_contents(DIR_FS_EMAIL_TEMPLATES . $css_lang_folder . 'email_common.css');
}

// if no assignment has been made AND the default file does not exist, then error ELSE
//   if there is no assignment yet (file exists and there was a language assigned), then error about missing language file and assign to the default file.
if ($block['EMAIL_COMMON_CSS'] == '' && !file_exists(DIR_FS_EMAIL_TEMPLATES . 'email_common.css')) {
  // language may have or may not have been assigned, but there is no file available.
  trigger_error('Missing common email CSS file(s): ' . DIR_FS_EMAIL_TEMPLATES . $css_lang_folder . 'email_common.css' . (zen_not_null($css_lang_folder) ? ' and ' . DIR_FS_EMAIL_TEMPLATES . 'email_common.css' : ''), E_USER_WARNING);
} else if ($block['EMAIL_COMMON_CSS'] == '') {
  // language was assigned, the language override does not exist, but the default does.
  trigger_error('Missing common email CSS file: ' . DIR_FS_EMAIL_TEMPLATES . $css_lang_folder . 'email_common.css', E_USER_WARNING);
  $block['EMAIL_COMMON_CSS'] = file_get_contents(DIR_FS_EMAIL_TEMPLATES . 'email_common.css');
}
````

Or is there an expectation that no matter what, if the default version of the file doesn't exist then to provide an error? (I.e., the above errors if there is no file found or if the override doesn't exist, where the "original" seemed to error if either the language override was not found or if the default file was not found...)

If it is expected that an error log be generated each time this section of code is executed and there is no default version of the file (expected to be consistent with the previous code's error messaging), then:
````php
// initialize array element.
$block['EMAIL_COMMON_CSS'] = '';
// Test for existence of default file
$default_css_file_exists = file_exists(DIR_FS_EMAIL_TEMPLATES . 'email_common.css');

// if $css_lang_folder is a language or empty quote and the file exists then assign.
if (file_exists(DIR_FS_EMAIL_TEMPLATES . $css_lang_folder . 'email_common.css')) {
  $block['EMAIL_COMMON_CSS'] = file_get_contents(DIR_FS_EMAIL_TEMPLATES . $css_lang_folder . 'email_common.css');
}

// If there was a language assigned, but the file did not exist (based on above test).
if ($block['EMAIL_COMMON_CSS'] == '' && zen_not_null($css_lang_folder)) {
  // Missing the language override file and there was a language override detected, want to notify here first as is the higher priority than a non-language override condition.
    trigger_error('Missing common email CSS file: ' . DIR_FS_EMAIL_TEMPLATES . $css_lang_folder . 'email_common.css', E_USER_WARNING);

  // But if the default exists then assign the contents.
  if ($default_css_file_exists) {
    $block['EMAIL_COMMON_CSS'] = file_get_contents (DIR_FS_EMAIL_TEMPLATES . 'email_common.css');
  }
}

// if the default file does not exist regardless of having an override, then error. Placed in this sequence to further reflect by log the override possibility such that the missing override (if detected as needed) would be identified first and then the default.
if (!$default_css_file_exists) {
  // language may have or may not have been assigned, but the common file is not present and possible not the override file.
  trigger_error('Missing common email CSS file: ' . DIR_FS_EMAIL_TEMPLATES . 'email_common.css', E_USER_WARNING);
}

````

drbyte:
@mc12345678 thanks for your thoughts.
I decided to rework the logic to match the pattern below it for selecting the actual template file, since they're doing similar processing.

I don't want to see any fatal (halting) errors triggered if the css file isn't found, however I think it should still be logged if none is found.

drbyte:
@lat9 would you mind peeking at this PR?

lat9:
@drbyte, that looks like some good fall-back processing.  As always, I've got a question ...

Since the `$_SESSION['languages_code']` is the 2-character language code, what's the purpose of that ` $langfolder_short` addition?

mc12345678:
No problem, the intention of the error reporting was to only provide notification.  In this "simplified" cycle, shouldn't there be a test out of verifying the same file exists three times if $lang_folder === ''?

mc12345678:
> I think it requires more code to check whether both are duplicate strings (when lang_folder is empty) than it does to just loop through the list.

Code, yes, but unless file presence is cached, drive cycles are slower and resource intensive than code.

Issue #1585, opened 2018-01-09
admin customers addresses popup link, cosmetic change


torvista:
In the customers listing., when the customer has multiple addresses, all of 
"| 1 of 2"
is made into a link including the separator, rather than just the "1 of 2" or perhaps just the "2" as would be expected.
This is due to the constant TEXT_INFO_ADDRESS_BOOK_COUNT containing the separator
`define('TEXT_INFO_ADDRESS_BOOK_COUNT', ' | 1 of  ');`

corresponding line in customers.php:
`<td class="dataTableContent" align="right"><?php echo $customer['customers_id'] . ($zc_address_book_count == 1 ? TEXT_INFO_ADDRESS_BOOK_COUNT . $zc_address_book_count : '<a href="' . zen_admin_href_link(FILENAME_CUSTOMERS, 'action=list_addresses' . '&cID=' . $customer['customers_id'] . ($page > 0 ? '&page=' . $page : '')) . '">' . TEXT_INFO_ADDRESS_BOOK_COUNT . $zc_address_book_count . '</a>'); ?></td>`

drbyte:
Am I correct in reading this that you're recommending the pipe (`|`) symbol be made something separate?

torvista:
Yes. It is always there and is not language dependant.

drbyte:
Closed via PR #1767

PR #1767, opened 2018-09-12
issue #1585 cosmetic change admin customers addresses

zcwilt:
While not a fan of having html in languages stuff, makes sense here.
Also don't see the need for text when there is only one addressbook link
but have allowed for that with TEXT_INFO_ADDRESS_BOOK_COUNT_SINGLE
fixes issue #1585

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -1,10 +1,10 @@
 <?php
 /**
  * @package admin
- * @copyright Copyright 2003-2016 Zen Cart Development Team
+ * @copyright Copyright 2003-2018 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: Author: zcwilt  Fri Apr 15 Modified in v1.5.5 $
+ * @version $Id: Author: zcwilt  Wed Sept 18 Modified in v1.5.6 $
  */
 require('includes/application_top.php');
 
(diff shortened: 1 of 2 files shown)


admin/customers.php
```diff
@@ -1276,7 +1276,7 @@ function init() {
                     $zc_address_book_count_list = zen_get_customers_address_book($customer['customers_id']);
                     $zc_address_book_count = $zc_address_book_count_list->RecordCount();
                     ?>
-                <td class="dataTableContent text-right"><?php echo $customer['customers_id'] . ($zc_address_book_count == 1 ? TEXT_INFO_ADDRESS_BOOK_COUNT . $zc_address_book_count : '<a href="' . zen_href_link(FILENAME_CUSTOMERS, 'action=list_addresses' . '&cID=' . $customer['customers_id'] . ($_GET['page'] > 0 ? '&page=' . $_GET['page'] : '')) . '">' . TEXT_INFO_ADDRESS_BOOK_COUNT . $zc_address_book_count . '</a>'); ?></td>
+                <td class="dataTableContent text-right"><?php echo $customer['customers_id'] . ($zc_address_book_count == 1 ? TEXT_INFO_ADDRESS_BOOK_COUNT_SINGLE : sprintf(TEXT_INFO_ADDRESS_BOOK_COUNT, zen_href_link(FILENAME_CUSTOMERS, 'action=list_addresses' . '&cID=' . $customer['customers_id'] . ($_GET['page'] > 0 ? '&page=' . $_GET['page'] : '')), $zc_address_book_count) . '</a>'); ?></td>
```
mc12345678: Looks like the final ` . </a>` that is added to all entries could come out of this line as well.  

This is based on considering that `TEXT_INFO_ADDRESS_BOOK_COUNT` contains the full html for the link, `TEXT_INFO_ADDRESS_BOOK_COUNT_SINGLE` doesn't contain any html and the mentioned content would be applied to whichever information is provided.

zcwilt: oops. you are right,

PR #1774, opened 2018-09-18
Pass product_type as form field with product collect_info.php

### LINKED ISSUE #1773: v156 not shown individual product_type preview_info.php
@Zen4All,
 
In #1736, the admin/product.php file was modified to assign the product type based on first the url containing an assigned products_id (pID), and if that is not present, then to pick up on a posted value (product_type). 

Then in #1747, the problem identified was that it was not possible to transition from a non-basic product type to the preview of information. The solution proposed supported previewing information; however, because of the above logic and posting the data instead of clicking a link, the product type defaulted to a standard product in admin/product.php.

This works for product that have already been assigned/created and for the first click from the categories_to_products listing to create a new product. However, a product's id is not assigned until the product is finally inserted.  The rewrite also modified the program flow which includes the central focus of the admin/product.php file needing to be provided information to be able to know/determine the product_type (and other necessary information), which can be accomplished a few other ways as well, but the majority of the process flow is by way of posted data instead of url related data.

mc12345678:
…fication.

In #1736, the admin/product.php file was modified to assign the product
type based on first the url containing an assigned products_id (pID), and if
that is not present, then to pick up on a posted value (product_type).

Then in #1747, the problem identified was that it was not possible to
transition from a non-basic product type to the preview of information.
The solution proposed supported previewing information; however, because of
the above logic and posting the data instead of clicking a link, the
product type defaulted to a standard product in admin/product.php.

This works for product that have already been assigned/created and for the
first click from the categories_to_products listing to create a new product.
However, a product's id is not assigned until the product is finally
inserted.  The rewrite also modified the program flow which includes the
central focus of the admin/product.php file needing to be provided information
to be able to know/determine the product_type (and other necessary
information), which can be accomplished a few other ways as well, but the
majority of the process flow is by way of posted data instead of url related
data.

This commit adds a hidden field to the collect info page to capture the
product_type expected, removes the product_type designation from the
uri for all default Zen Cart product types, and modifies the logic to first
act on the posting of data instead of information on the uri.

Fixes #1773

### DIFF
--- a/admin/product.php
+++ b/admin/product.php
@@ -11,7 +11,7 @@ require('includes/application_top.php');
 $action = (isset($_GET['action']) ? $_GET['action'] : '');
 require(DIR_WS_CLASSES . 'currencies.php');
 $currencies = new currencies();
-$product_type = (isset($_GET['pID']) ? zen_get_products_type($_GET['pID']) : (isset($_POST['product_type']) ? $_POST['product_type'] : 1));
+$product_type = (isset($_POST['product_type']) ? $_POST['product_type'] : (isset($_GET['pID']) ? zen_get_products_type($_GET['pID']) : 1));
 $type_handler = $zc_products->get_admin_handler($product_type);
 $current_category_id = (isset($_POST['cPath']) ? $_POST['cPath'] : (isset($_GET['cPath']) ? $_GET['cPath'] : 0));
 $zco_notifier->notify('NOTIFY_BEGIN_ADMIN_PRODUCTS', $action);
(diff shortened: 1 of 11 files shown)


admin/product.php
```diff
@@ -11,7 +11,7 @@
 $action = (isset($_GET['action']) ? $_GET['action'] : '');
 require(DIR_WS_CLASSES . 'currencies.php');
 $currencies = new currencies();
-$product_type = (isset($_GET['pID']) ? zen_get_products_type($_GET['pID']) : (isset($_POST['product_type']) ? $_POST['product_type'] : 1));
+$product_type = (isset($_POST['product_type']) ? $_POST['product_type'] : (isset($_GET['pID']) ? $_GET['pID'] : 1));
```
drbyte: This should still call `zen_get_products_type()`

mc12345678: Correct... Sorry and corrected.

drbyte:
Thanks!

PR #1777, opened 2018-09-19
fix broken css link

### LINKED ISSUE #1776: New install css is broken
When doing a new  install I do not include the docs folder.
This breaks the new layout for the install pages, as it refers to the css files in the docs folder.

Zen4All:
This fixes the broken css link to the docs folder, as there should be no docs folder in a production environment.

Fixes #1776


drbyte:
Thanks!

Issue #1778, opened 2018-09-19
Sub-categories can be added to categories containing products which are all disabled


lat9:
To replicate, within the admin's _**Catalog->Categories/Products**_:

1. Create a new sub-category.
2. Add a product to that category and disable that product.

Upon return to the "categories/products listing", see:
![image](https://user-images.githubusercontent.com/2685585/45774609-6a65de80-bc1b-11e8-92fe-7a4447c56bf5.png)

The admin is, incorrectly, given the opportunity to add a category to a category that contains products &mdash; which can lead to unexpected results during store front operations.

------

The issue lies in the following code block, located in` /admin/categories.php`:
```php
      // check for which buttons to show for categories and products
      $check_categories = zen_has_category_subcategories($current_category_id);
      $check_products = zen_products_in_category_count($current_category_id, false, false, 1);

      $zc_skip_products = false;
      $zc_skip_categories = false;

      if ($check_products == 0) {
        $zc_skip_products = false;
        $zc_skip_categories = false;
      }
      if ($check_categories == true) {
        $zc_skip_products = true;
        $zc_skip_categories = false;
      }
      if ($check_products > 0) {
        $zc_skip_products = false;
        $zc_skip_categories = true;
      }
```

The 2nd parameter in the call to `zen_products_in_category_count` requests a count of products _that are not deactivated_, which is the root-cause of the issue.  Changing that statement to

```php
 $check_products = zen_products_in_category_count($current_category_id, true, false, 1);
```

corrects the unwanted behavior.

drbyte:
Closed by #1779

mc12345678:
With this "correction", what is the proposed method to move product into one or more sub-categories of a main category where the sub-category(ies) do not yet exist? 

Further, does this mean that all of the other move and copy options are going to be revised to prevent product and sub-categories from existing in a category? 

Ie. copy (duplicate) a product from a branched sub-category to the parent category, the product will be disabled but be within the category listing that also has a sub-category. On the store front, the central area when trying to view that main category shows no sub-categories nor product; however, the categories sidebox shows the additional category(ies).  This can be seen by going to DVD Movies->Action and duplicating Speed 2: Cruise Control to the category DVD Movies.  Then from the catalog side going to: index.php?main_page=product_info&cPath=3  

[...]

PR #1779, opened 2018-09-19
#1778: Correct categories+products in categories

lat9:
Note that the module that requires update has changed for the 1.5.6 release!


drbyte:
Thanks!

PR #1781, opened 2018-09-19
Revise product type check

mc12345678:
Saw this code section that looked like it could use "improvement".  I haven't asked anyone if there is/would be problems with others code caused by making this change instead of just leaving it alone.  Looked liked something that could be improved upon and so I changed it.

There is duplication in the code to identify the product types that can be removed by reworking some of the variables and operations.  The effect of this modification is that there is no longer a variable created called `$product_types_array` that would only exist if there were no product restrictions, there is no longer a specific variable generated `$restrict_types` to identify the specific restricted product types (or absence of them).

Also applied the PHP 5.6+ use of `[]` for array declaration instead of `array()` throughout.


drbyte:
Thanks!

PR #1782, opened 2018-09-20
Remove unnecessary code from refactor of Move in method rewind

mc12345678:
In #1633, there had been some ideas proposed, one of which was to
add a parameter to the call to the Move method that would specifically
rewind the query results and not offer the integer based field keys.

The functionality within the Move method was basically returned to its
previous state and the additional parameter was identified as unnecessary.

This removes the parameter and associated comment for "cleanliness".


drbyte:
Thanks!

PR #1783, opened 2018-09-20
Support adding debug backtrace data to DB Transaction log

mc12345678:
While attempting to identify the cause, path, and history of queries
in a plugin, I found it necessary to identify the source of the
query(ies) beyond what is provided by the default logging.  At one point
I had corrected a relatively obvious source, but in so doing I lost any
additional markers of the flow path.  Adding the backtrace information
to this log (location within output is welcome to be moved as necessary)
greatly added useful information.

While such logs are already lengthy, this has been incorporated at the
moment as a feature that can be enabled by adding a constant and defining
it to not be falsey.  If this feature is considered accepted then I think
that in the logging section of the configuration menu, it would be worth
adding an additional option to include this data instead of "requiring"
a file modification.

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -180,7 +180,20 @@ class queryFactory extends base {
       if (strtoupper(substr($zf_sql,0,6))=='SELECT' /*&& strstr($zf_sql,'products_id')*/) {
         $f=@fopen(DIR_FS_LOGS.'/query_selects_' . $current_page_base . '_' . time() . '.txt','a');
         if ($f) {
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks

mc12345678:
> Since the config constants are always strings, and the proposed backtrace output is directly dependent on STORE_DB_TRANSACTIONS, how about expanding the existing STORE_DB_TRANSACTIONS, from just true/false to true/false/backtrace choices? (instead of introducing another config setting that would be irrelevant if the other setting was false anyway)

Somehow missed this comment. 100% agree with the recommendation and revision as pushed, thank you for taking the time with it.

PR #1785, opened 2018-09-20
Fix refactor error in attributes_controller

mc12345678:
Appears that in the rebuild of the attributes_controller that a variable
was written out incorrectly ($products_options_array was used instead of
$products_options_query). Based on further search of the file and
the just-in-time application of the code, this commit appears to be the
best/right fit for this issue that had thrown warnings while in strict
mode.


drbyte:
Thanks!

PR #1787, opened 2018-09-20
Downloads: Redirect to timeout using SSL

lat9:
As @mc12345678 so kindly pointed out on the OPC repo (https://github.com/lat9/one_page_checkout/issues/151), there's a discrepancy between the `includes/modules/pages/download/header_php.php` provided by the plugin and zc156 (on which the plugin's downloads-handling is based).

That is due to the fact that the zc156 repo's version of the module hasn't been updated to provide that handling, which this PR corrects.


drbyte:
Thanks!

Issue #1416, opened 2017-04-03
Isn't it time to make ez-pages multilingual?


Zen4All:
This has been something that annoyed me since I started with Zen Cart, in the 1.3.8 era.  The ez-pages are only single language in the core.
Is there any interest to bring this add-on: [Multi-Language EZ-Pages](https://www.zen-cart.com/downloads.php?do=file&id=36), to the core code?

drbyte:
I've often considered it, but then when I go to inspect how stable the plugin is it seems someone else is posting that they're having pain with it.

I'd consider a PR for v160

Zen4All:
I have installed this module on multiple installations, and can't remember I ever had a problem with it, but I'll install and test it on a zc160 testbed, and see what happens.
If nothing out of the ordinary happens, I'll prepare a pull for the 1.6 branch

drbyte:
I'd welcome a PR. 
Even for v156 as long as the schema changes aren't terribly drastic or complex regarding old data.

Zen4All:
I'll make a pr next week.
The scheme changes are not that complex.

drbyte:
Hi @Zen4All what are your thoughts on readiness of this for v156?

drbyte:
Closed by PR #1789

PR #1789, opened 2018-09-24
Add multi language support to ez-pages

Zen4All:
These are the changes needed to make EZ-pages multi lingual. I have kept the database tables the same as the ML addon, so if already installed it wil not give any problems.

see #1416

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1789)


drbyte:
Thanks!

PR #1790, opened 2018-09-25
Replace $_GET['product_type'] with $_POST['product_type']

Zen4All:
$_GET['product_type'] is empty, and $_POST['product_type'] is passed from the preview page.


Zen4All:
Wrong branch

PR #1795, opened 2018-09-25
Add admin-level, common AJAX support

### LINKED ISSUE #1738: v156: Include base AJAX processing in the admin?
Is it time to update the admin to use the base AJAX processing that's included in the storefront?  

I'm thinking that an `/admin/ajax.php` would `require` the version in the storefront root, the admin header would require the storefront `/includes/template_default/jscript/jscript_framework.php` and the admin's `classes` directory would now include an `ajax` sub-directory ... mimicking the storefront behavior.

lat9:
Closes #1738

### DIFF
--- a/admin/includes/init_includes/init_admin_auth.php
+++ b/admin/includes/init_includes/init_admin_auth.php
@@ -10,6 +10,22 @@ if (!defined('IS_ADMIN_FLAG')) die('Illegal Access');
 
 define('SUPERUSER_PROFILE', 1);
 
+// -----
+// Special handling for AJAX requests.  Return a 'logged_out' error if no admin session
+// is active; otherwise, bypass the remainder of the authorization checks.
+//
+if (basename($PHP_SELF) != FILENAME_AJAX . '.php') {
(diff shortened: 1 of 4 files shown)


admin/includes/init_includes/init_admin_auth.php
```diff
@@ -10,6 +10,22 @@
 
 define('SUPERUSER_PROFILE', 1);
 
+// -----
+// Special handling for AJAX requests.  Return a 'logged_out' error if no admin session
+// is active; otherwise, bypass the remainder of the authorization checks.
+//
+if (basename($PHP_SELF) != FILENAME_AJAX . '.php') {
```
drbyte: Should this be `==` instead of `!=` ?

lat9: It's not my month ... of course it should be.

lat9:
That's most likely going to correct that json output that I saw on admin session-timeout, too.

drbyte:
Thanks!

PR #1800, opened 2018-09-28
Invoice/packingslip

Zen4All:
- Remove over complex table structure
- remove unnecessary order_check query, as this data is already in $order
- add container class, to make the width smaller on wide screens, better resembling a paper sheet

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1800)


drbyte:
Thanks!

PR #1806, opened 2018-10-08
Call to zen_draw_radio_field should have boolean third parameter

scottcwilson:
Value was set to 'selected'.  

See https://www.zen-cart.com/showthread.php?224305-Call-to-zen_draw_radio_field-in-includes-modules-attributes-php


drbyte:
Thanks.
In older PHP versions, since the string `'selected'` is `truthy`, it was evaluating to `true`. But in newer PHP versions which are more strict about type, the boolean is definitely better.

PR #1807, opened 2018-10-09
Move markup to code files for TEXT_PRODUCTS_URL_WITHOUT_HTTP constant

scottcwilson:


drbyte:
What business problem is this solving? 

Having the `<small>` class in the language file means people don't have to touch core code in 5 places if they want to tweak that.

Zen4All:
Or just remove the `<small>` tag completely. Or add a class, so it can be changed in the css in 1 place

scottcwilson:
Compliance with the coding standards, which specify that markup and strings should be separate. 

https://docs.zen-cart.com/Contributing/main/coding_standards

drbyte:
Does this mean your next plan is to change every `<span>` and `<strong>` tag in all the language files?
I don't want to see that.

scottcwilson:
Why don't we close this out then and revisit in 2.0.

torvista:
Many years ago I suggested this field should accept urls with/without protocols and just check and strip them off if necessary. I added that necessary code at the time
Why?
Well in my case at least, that url is always copied and pasted from a browser, so always has the protocol prefix.
Or, it is imported from a csv, which would be built by copying and pasting urls, again with protocols.
So, I think it should be protocol agnostic, and then you can get rid of the explanatory text altogether.

drbyte:
> I think it should be protocol agnostic, and then you can get rid of the explanatory text altogether.

Ya, when stepping back as you've done, I agree that that's probably the most user-intuitive direction to take.

torvista:
Note that this field and the url for the manufacturer website should use the same function.

PR #1817, opened 2018-10-10
Indicate what the Product's MetaTag Title is affected by

### LINKED ISSUE #1555: Mark What the Product's Meta Tag Title Should Include
This baffles me.
When entering data in the admin\includes\modules\product\collect_info_metatags.php

You can toggle the radiobuttons to include or exclude from the meta-tag. When setting values to exclude this is saved in the database as a 0 value (be sure to leave the input fields empty). So far all goes well.

Now re-open the page and all radio buttons are "reset" to include (value = 1), although the database contains 0 values.
The trigger for this is an if statement that checks for the meta_description and metatags_keywords being empty.
https://github.com/zencart/zencart/blob/783637669f95c66a460c348830a1da4f616c3186/admin/includes/modules/product/collect_info_metatags.php#L63-L68
What is the logic behind this? because now on screen you have other data then is in the database .

drbyte:
Alerts what is required for metatags data to activate

Fixes #1555

### DIFF
--- a/admin/includes/languages/english/product.php
+++ b/admin/includes/languages/english/product.php
@@ -2,10 +2,10 @@
 
 /**
  * @package admin
- * @copyright Copyright 2003-2017 Zen Cart Development Team
+ * @copyright Copyright 2003-2018 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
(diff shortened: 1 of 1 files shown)


drbyte:
screen shot 2018-10-10 at 12 57 48 am

PR #1821, opened 2018-10-12
Search button too small 

scottcwilson:
When the responsive classic template does not use CSS buttons, the  search button needs to be wider.  Otherwise the text gets cut off.


drbyte:
Since the buttons needed some extra space in both templates, I've moved it to template_default.

Thanks!

scottcwilson:
Cool.  No sense in duplicating the file if it can easily be avoided.

PR #1832, opened 2018-10-28
Display name of database in Version check

scottcwilson:
Showing the name of the database currently in use by Zen Cart in the version check is a nice timesaver for clients who have many databases in their account.

### DIFF
--- a/admin/includes/languages/english/server_info.php
+++ b/admin/includes/languages/english/server_info.php
@@ -1,10 +1,10 @@
 <?php
 /**
  * @package admin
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks. 

I've relocated the output to the top of the server-info section, next to the server's IP, as these are (essentially the only) "identifying" information that might be unsafe to copy/paste in a support forum. Keeping them together simplifies "not forgetting to scrub" when posting.

scottcwilson:
+1

PR #1833, opened 2018-11-04
Fix error messages from upload class when used by admin

scottcwilson:
"File too large" failure messages not appearing during admin product edit.


scottcwilson:
I could do the redirect on upload failure if you think that's a better approach.

drbyte:
Is this "product edit" the only place the problem occurs?

scottcwilson:
As far as I know - but I only stumbled on this issue.

PR #1834, opened 2018-11-05
remove unused variable $languages_id in site-map

Zen4All:
And $load_from_database = true can probably also be removed as it is also unused.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1834)


drbyte:
Thanks!

PR #1835, opened 2018-11-07
Refactor Contact_us header_php.php

mc12345678:
- Adjusted the spacing/alignment of code, though only to two spaces each.
- Applied PHP 7.x style array handling changing from `array()` to `[]` in all instances.
- Reworked the generation of `$send_to_email` and `$send_to_name` to remove the dependency on the source email from having the format of: `NAME TO DISPLAY <email@address>` where if the section of `<>` was missing then no email would be sent and a debug log would likely be created.  This modification generally expects that a proper email address has been entered in either `EMAIL_FROM` or `CONTACT_US_LIST`.  There is no change to the presence/absence of a validation check of the email address being properly entered.
- Converted use of logic and to && and added parentheses to maintain historical identification and clarity of the operator precedence.
- Added a test for the presence of posting the `send_to` field when `CONTACT_US_LIST` is populated for strict handling.
- Github identified a mix of line endings and reported that they would all be changed to "Windows Style" by normalizing them to CR|LF.

### DIFF
--- a/includes/modules/pages/contact_us/header_php.php
+++ b/includes/modules/pages/contact_us/header_php.php
@@ -3,10 +3,10 @@
  * Contact Us Page
  *
  * @package page
- * @copyright Copyright 2003-2016 Zen Cart Development Team
+ * @copyright Copyright 2003-2018 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: Author: DrByte  Sun Oct 18 23:02:01 2015 -0400 Modified in v1.5.5 $
+ * @version $Id: Author: DrByte modifier: mc12345678 Modified in v1.5.6 $
  */
 
 // This should be first line of the script:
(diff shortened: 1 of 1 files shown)


lat9:
Wouldn't this be an opportunity to apply [PSR-2](https://www.php-fig.org/psr/psr-2/) styling to the module?

mc12345678:
Yes it is, provided the code changes are/will be considered acceptable. I am working on a writeup regarding some caveats to what the proposed change has with a potential minor alternate consideration.

includes/modules/pages/contact_us/header_php.php
```diff
@@ -26,57 +26,65 @@
 
   $zc_validate_email = zen_validate_email($email_address);
 
-  if ($zc_validate_email and !empty($enquiry) and !empty($name) && $error == FALSE) {
+  if (($zc_validate_email && !empty($enquiry) && !empty($name)) && $error == FALSE) {
```
drbyte:
```suggestion
  if ($zc_validate_email && !empty($enquiry) && !empty($name) && $error == FALSE) {
```

Thanks for considering the parentheses which denote the prior meaning; however in this case they are unnecessary.

drbyte:
Save for the one minor suggestion above, I'm okay with what's proposed so far.

PR #1838, opened 2018-11-09
Refactor Contact_us header_php.php

mc12345678:
- Adjusted for PSR-2: four spaces for indenting, line endings UNIX style, one space after commas
  of parameters, spaces before/after assignment or comparison, etc...
- Applied PHP 7.x style array handling changing from `array()` to `[]` in all instances.
- Reworked the generation of `$send_to_email` and `$send_to_name` to remove the dependency on the source email from having the format of: `NAME TO DISPLAY <email@address>` where if the section of `<>` was missing then no email would be sent and a debug log would likely be created.  This modification generally expects that a proper email address has been entered in either `EMAIL_FROM` or `CONTACT_US_LIST`.  There is no change to the presence/absence of a validation check of the email address being properly entered.
- Converted use of logic `and` to `&&`.
- Added a test for the presence of posting the `send_to` field when `CONTACT_US_LIST` is populated for strict handling and to default to `EMAIL_FROM` and `STORE_NAME` when `send_to` is not posted or is not recognized in `CONTACT_US_LIST`.

### DIFF
--- a/includes/modules/pages/contact_us/header_php.php
+++ b/includes/modules/pages/contact_us/header_php.php
@@ -12,119 +12,121 @@
 // This should be first line of the script:
 $zco_notifier->notify('NOTIFY_HEADER_START_CONTACT_US');
 
-require(DIR_WS_MODULES . zen_get_module_directory('require_languages.php'));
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks!

Issue #1853, opened 2018-11-15
Additional notification needed for /admin/orders.php


lat9:
This (https://github.com/zencart/zencart/commit/7e59597a6038fb2939bb2eae36a3e295a03798f7#diff-f12e30c1c80e1e48222bff824ac6902b) commit added the following notifier to enable observers to add columns to the orders' listing display:

```php
$zco_notifier->notify('NOTIFY_ADMIN_ORDERS_LIST_EXTRA_COLUMN_DATA', (isset($oInfo) ? $oInfo : array()), $orders->fields);
```
That's a good thing, but there's no matching notification to allow an observer to add the table-heading associated with the data-item insert.

PR #1854, opened 2018-11-15
Add notifier for admin/orders extra listing headings

lat9:


drbyte:
Thanks!

Closes #1853

PR #1859, opened 2018-11-16
Shorten display of long config values

scottcwilson:
Only show the first 40 chars of a long config value so screen won't get messed up.


drbyte:
Thanks

PR #1861, opened 2018-11-17
Allow EZPage to show even if not in header,footer,sidebox

### LINKED ISSUE #1823: Proposed: remove the filter for EZ page display
Sometimes I want to create some EZ page content just for display in a specific page (the product info page, for example), and don't want it in the header, footer or sidebox. But the current system doesn't allow me to do this - a check in `includes/modules/pages/page/header_php.php` on line 30 will stop it.  Now I know you can comment out the check, but I think this ought to be the default. Thoughts?

scottcwilson:
Allow ezpage to be marked visible even if not show in header, footer or sidebox. 

Fixes #1823

### DIFF
--- a/admin/ezpages.php
+++ b/admin/ezpages.php
@@ -73,6 +73,7 @@ if (zen_not_null($action)) {
       }
       $page_open_new_window = (int)$_POST['page_open_new_window'];
       $page_is_ssl = (int)$_POST['page_is_ssl'];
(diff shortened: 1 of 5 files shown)


drbyte:
This looks good. 
I wonder if there should be an input/checkbox for this in the Edit screen as well?

scottcwilson:
Yep, good idea.

drbyte:
Great! Thanks!

PR #1862, opened 2018-11-17
Admin-orders notifications "future-proofing"

### LINKED ISSUE #1858: Possible change to admin/orders.php notifications?
After pushing that header-notification update for #1853, it occurred to me that the current interfaces used for the `NOTIFY_ADMIN_ORDERS_LIST_EXTRA_COLUMN_HEADING` and `NOTIFY_ADMIN_ORDERS_LIST_EXTRA_COLUMN_DATA` notifications make them very version-dependent.

I was looking to incorporate, for instance, back into zc155f, but that version of `orders.php` is a table-based rendering while the zc156 (and presumably follow-on versions) use a div-based display.

Would you accept a PR to make those notifications more future-proofed?

lat9:
Closes #1858


drbyte:
Nicely done. Thanks!

PR #1871, opened 2018-11-21
Updates to Spiffy

Zen4All:
I know this script will be removed in the future, but Zen Cart 1.5.6 will be around for some time. So I found it necessary to update the layout, so it matches the bootstrap theme better.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1871)


drbyte:
Thanks!

PR #1874, opened 2018-11-24
Layout updates for reports

Zen4All:
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1874)

### DIFF
--- a/admin/stats_products_purchased.php
+++ b/admin/stats_products_purchased.php
@@ -1,5 +1,3 @@
-<?php
-/**
  * @package admin
  * @copyright Copyright 2003-2016 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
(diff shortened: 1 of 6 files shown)


admin/stats_products_purchased.php
```diff
@@ -6,257 +6,223 @@
...
+                $rows = '0' . $rows;
+              }
+              if ($products_filter != '') {
+                // products_id
+                $cPath = zen_get_product_path($products_filter);
+              } else {
+                // products_name or products_model
+                $cPath = zen_get_product_path($orders_products['products_id']);
+              }
+              $type_handler = $zc_products->get_admin_handler($product['products_type']);
```
drbyte: Is `$product` already defined here? And `$zc_products` too?

Zen4All: I had to add that line, because $type_handler was undefined, and gave an error on the page

drbyte: I see $product defined in similar code later in 195, 202, 206, but not up here.

drbyte: I see `$zc_products` is part of the init autoloaders. So that part's fine.

The `$product` var is not. (The code may "work" because of fallbacks use to ensure a sane response is obtained.)

I think you'll need to change `$product['products_type']` to `zen_get_products_type($orders_products['products_id'])` in this case.

drbyte: And also on `202`

drbyte:
Apart from fixing those, the rest of the refactorings here look good.

Thanks again!

Zen4All:
Change made.

drbyte:
Thanks!

PR #1881, opened 2018-11-28
Layout update for music genre files

Zen4All:
Media-manager will follow in an up coming PR/commit, as it needs some more work.


scottcwilson:
Has this stuff been tested?  Was it ready to merge?

PR #1883, opened 2018-11-29
Fix products purchased report

scottcwilson:


Zen4All:
Thats strange, it was there when i pushed the PR 😕

PR #1889, opened 2018-12-02
zc156: Square jQuery no-load

lat9:
If the Square payment method is enabled, but not loaded (e.g. the 'freecharger' payment method is active) or if Square has zone restrictions (and the current billing zone is outside of the restrictions), don't load the payment method's jQuery processing, since it will "pollute" the jQuery pool.


lat9:
Noting that I've created a similar [PR ](https://github.com/zencart/payeezyjs/pull/5)for the Payeezyjszc payment-method.  That update should also get rolled into v156.

drbyte:
Thanks!

PR #1894, opened 2018-12-13
Fix admin javascript errors in customer javascript messages

scottcwilson:
Caused by embedded single quote.  Occurs in Admin->Customers->Customers, edit customer.


drbyte:
Shouldn't these entire defines be wrapped in double-quotes?  That does 2 things: it removes the need to escape the single-quotes, and makes the `\n` into a proper line-break. 
Or does that need to also translate over into JS where these strings are echoed out inside single-quoted strings already?

PR #1895, opened 2018-12-13
Embedded break tag not needed

scottcwilson:
Embedded carriage return means message is not properly formatted on mobile.  Carriage return not required; works fine without it.


scottcwilson:
webstore_product

PR #1896, opened 2018-12-13
Hide some sections in admin product listing to make more mobile friendly

scottcwilson:


drbyte:
Since these pages are styled with Bootstrap, shouldn't we just use the built-in `.hidden-md` class?

Ref: https://getbootstrap.com/docs/3.3/css/#helper-classes

scottcwilson:
LOL I looked and looked for that and couldn't find it.  Will do.

drbyte:
Thanks!

Zen4All:
I am not a fan of putting the hidden-md class here, as it will hide stuf on a 1024px wide screen, which is not necessary. Otherwise great fix.

admin/category_product_listing.php
```diff
@@ -648,12 +648,12 @@ function init() {
                 }
                 ?>
                 <tr onclick="document.location.href = '<?php echo zen_href_link(FILENAME_CATEGORY_PRODUCT_LISTING, zen_get_path($category['categories_id'])); ?>'" role="button">
-                  <td class="text-right"><?php echo $category['categories_id']; ?></td>
-                  <td><?php echo zen_image(DIR_WS_ICONS . 'folder.gif', ICON_FOLDER); ?>&nbsp;<strong><?php echo $category['categories_name']; ?></strong></td>
-                  <td class="text-center">&nbsp;</td>
-                  <td class="text-right"><?php echo zen_get_products_sale_discount('', $category['categories_id'], true); ?></td>
-                  <td class="text-center">&nbsp;</td>
-                  <td class="text-right">
+                  <td class="text-right hidden-md hidden-sm hidden-xs"><?php echo $category['categories_id']; ?></td>
```
Zen4All: Id column is not hidden in product, and header.

admin/category_product_listing.php
```diff
@@ -648,12 +648,12 @@ function init() {
...
+                  <td hidden-md hidden-sm hidden-xs><?php echo zen_image(DIR_WS_ICONS . 'folder.gif', ICON_FOLDER); ?>&nbsp;<strong><?php echo $category['categories_name']; ?></strong></td>
```
Zen4All: This should be removed

PR #1902, opened 2018-12-15
Fix current category setting when adding new products in Admin

### LINKED ISSUE #1901: 1.5.6: Product created in subcategory has wrong master categories id
Repro steps: navigate to hardware->monitors.  Create a product.  Product will not be visible in monitors category; it was created in hardware category. 

Issue seems related to the fact that master category shown during creation is ID# 1_6 Hardware instead of ID# 6_Monitors.   Investigating.

scottcwilson:
Fixes #1901.
Testcase provided in that issue.

### DIFF
--- a/admin/includes/init_includes/init_category_path.php
+++ b/admin/includes/init_includes/init_category_path.php
@@ -12,6 +12,8 @@ if (!defined('IS_ADMIN_FLAG')) {
 // calculate category path
   if (isset($_GET['cPath'])) {
     $cPath = $_GET['cPath'];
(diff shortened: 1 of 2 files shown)


scottcwilson:
Root cause: in 1.5.5, the file admin/includes/modules/category_product_listing.php added the cPath to the URL on line 330, which meant it could be retrieved out of $_GET data.  In 1.5.6, this logic was moved to category_product_listing.php line 1209, where cPath was added as a form hidden field, moving it to $_POST data.

PR #1903, opened 2018-12-15
remove product music type stuff from install

zcwilt:
The intention here is to test any new plugin infrastructure so that it would allow something complex like the music type to be easily installed as an addon


zcwilt:
I did think about this, but decided that we could probably fix in an upgrade script. 1.6 will cause so much disruption regarding upgrading anyway, that this is probably the least of those problems

PR #1919, opened 2018-12-19
Fixes #1917 - PHPMailer failure in 2.0

### LINKED ISSUE #1917: Running an order with emails on and mail transport = PHP gives blank page in 1.6
PHPMailer not properly referenced in includes/functions/functions_email.php

[19-Dec-2018 06:16:17 America/New_York] PHP Fatal error:  Class 'PHPMailer\PHPMailer\PHPMailer' not found in /Users/scott/Sites/demo_160/includes/functions/functions_email.php on line 194

scottcwilson:
Checkout and other operations requiring email send will crash if send emails = true and transport = PHPMailer.


includes/functions/functions_email.php
```diff
@@ -191,6 +191,7 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
       $zco_notifier->notify('NOTIFY_EMAIL_AFTER_EMAIL_FORMAT_DETERMINED');
 
       // now lets build the mail object with the phpmailer class
+      require DIR_APP_LIBRARY . 'PHPMailer/src/PHPMailer.php';
```
drbyte: @zcwilt shouldn't the class autoloader take care of this, so we don't need to do a manual `require` here?

scottcwilson: includes/auto_loaders/config.core.php 241-248 should be updated in any event.

scottcwilson: But I saw other manual requires for classes such as in includes/application_top.php: 
`DIR_APP_LIBRARY . 'aura/autoload/src/Loader.php';`

scottcwilson: I know for a fact that what's in this PR does work, because I had to test it a bunch of times to get it to work.

drbyte: Will merge for now. May update later to make less rigid. (Preferring to allow class autoloaders do their own requires)

Issue #1914, opened 2018-12-18
Convert remaining pages from older style top menu bar to newer


scottcwilson:
A number of pages in 1.6.0 have not yet been updated with the newer style headers.  Compare 

admin/index.php?cmd=categories

to

admin/index.php?cmd=sqlpatch

(see "new" and "old" images attached)

I could go ahead and take care of this.

![new](https://user-images.githubusercontent.com/4391638/50178772-69545400-02d3-11e9-8067-a7ea1057feed.png)

![old](https://user-images.githubusercontent.com/4391638/50178794-7a04ca00-02d3-11e9-9a10-d402477ecf33.png)

scottcwilson:
About 25% of these are now done.

scottcwilson:
Last PR for this submitted.

PR #1926, opened 2018-12-19
Fixing top part of admin page

scottcwilson:
First block of fixes for #1914.

### DIFF
--- a/admin/attributes_controller.php
+++ b/admin/attributes_controller.php
@@ -687,7 +687,6 @@ function translate_type_to_name($opt_type) {
     return $value_string;
   }
 
-require('includes/admin_html_head.php');
 ?>
 <script type="text/javascript">
 function go_option() {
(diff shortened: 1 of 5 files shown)


admin/attributes_controller.php
```diff
@@ -699,10 +698,14 @@ function popupWindow(url) {
   window.open(url,'popupWindow','toolbar=no,location=no,directories=no,status=no,menubar=no,scrollbars=no,resizable=yes,copyhistory=no,width=600,height=460,screenX=150,screenY=150,top=150,left=150')
 }
 </script>
+<?php
```
Zen4All: This puts the above JavaScript outside the html tags

scottcwilson: Update your branch and do a view source - it looks ok to me.

Zen4All: Is see the same errors in some of the files in #1935 where `require_once('includes/template/common/tplHtmlHeadLegacy.php');`
is put after the javascripts

scottcwilson:
ARGH.  You are correct -I see what you mean now.

Zen4All:
Maybe the scripts should be lowered to be inside body tag

scottcwilson:
It's ok - I rearranged the inclusion of these files and it works now.  The trouble is, the example I used for how to do this was wrong, so I carried the error forward. :(  But the newly submitted files are fixed.

PR #1925, opened 2018-12-19
check length of attention field before inserting into db

scottcwilson:
I have seen this fail when the string was too long.

PR #1928, opened 2018-12-19
Move fix from #1925/#1927 into v2

scottcwilson:


scottcwilson:
How do I run the unit tests?

zcwilt:
It's the use of zen_field_length that is causing the problem, as that tries to access global $db
which isn't set in the tests. would need to do a simple mock here.

zcwilt:
@scottcwilson  re tests
in the root of where you have files checked out on your computer you would need to do
``composer update``, assuming you have composer installed
then you should just be able to run 
``composer tests``

scottcwilson:
Hmmm... in AdminLoggingTest.php line 110 I added 

`
        global $db;
        $db = $this->getMockBuilder('queryFactory')->getMock();
`

but get the same failure.

PR #1931, opened 2018-12-20
Allow admin to support page load language overrides.

### LINKED ISSUE #1930: Admin: unable to override core language definitions
While applicable to ZC 1.5.6, it is also applicable to earlier versions.  If a language define exists in the main language file of the admin (`admin/includes/languages/English.php`), then it can not be overridden by any other language define in a default zen cart store.  To perform the override, additional software would have to be added to pre-load the associated language file(s) before `admin/includes/init_includes/init_languages.php` at load point 70.  

The current load sequenbce in `admin/includes/init_includes/init_languages.php` is:
- load the main language file
- load the language file for the current page
- load the languages/extra_definitions file(s)

As a result any definition that is in the main language file can not be overridden in a current page file and in order to provide a language override on one page, the definition(s) of the main language file has to be transferred to each and every current page language file that could possibly want to use that definition.  This is basically reverse of how the catalog side is loaded and is something that has been seen for some time on the admin side with regards to product types.

I have a suggested revision; however, also have seen how there have been some tests added in Zen Cart 1.5.6 to prevent some sort of double loading.

mc12345678:
Fixes #1930

### DIFF
--- a/admin/includes/init_includes/init_languages.php
+++ b/admin/includes/init_includes/init_languages.php
@@ -4,7 +4,7 @@
  * @copyright Copyright 2003-2018 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: lat9 Wed May 30 15:34:35 2018 -0400 Modified in v1.5.6 $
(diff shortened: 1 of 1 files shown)


drbyte:
I agree with this.

In fact, I thought we'd already done this ... but evidently it was just a dream :)

Thanks!

PR #1933, opened 2018-12-20
v156 post before get on cPath assignment

mc12345678:
Generally speaking, it seems that when given the opportunity to read
something from the uri and from the posted page data, that the information
to keep/capture should come and normally comes from the posted page data.

As such, this commit modifies the sequence of testing to assign the `$cPath`
on page load such that a modification to the page's uri would not affect
the cPath.

### DIFF
--- a/admin/includes/init_includes/init_category_path.php
+++ b/admin/includes/init_includes/init_category_path.php
@@ -10,10 +10,10 @@ if (!defined('IS_ADMIN_FLAG')) {
   die('Illegal Access');
 }
 // calculate category path
-  if (isset($_GET['cPath'])) {
-    $cPath = $_GET['cPath'];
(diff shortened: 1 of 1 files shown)


drbyte:
This switches around the order of steps added in #1902 

I tend to agree with the proposed change.

mc12345678:
Yes, apologies for not referencing that commit/issue. An alternate solution for the modification of that commit was to again update the product info `collect_info` pages to capture the provided category information; however, the proposed change seems to centralize the assignment more appropriately. Only really an issue then when trying to upgrade to using new product types... (both methods could maybe be employed in such files for backwards compatibility as determined necessary by the updater.)

scottcwilson:
I am open to this change.  I didn't have a particular order in mind when I submitted #1902, I just wanted the POST case to be covered.

PR #1938, opened 2018-12-20
fixes ordering issue from #1926

scottcwilson:


scottcwilson:
Thanks to @Zen4All for reporting this bug.

PR #1941, opened 2018-12-20
prevent empty ezpages links list from throwing warnings in PHP 7.2+

scottcwilson:


drbyte:
Clarification: you're just adding this for "completeness", not because it's actually still needed right?

Ref previous commit: https://github.com/zencart/zencart/commit/d8e60d70fd1f2a06171cea25192386f630f6edfa

scottcwilson:
No, it's needed.  I got logs without it.

drbyte:
Hmm, maybe it needs to be initialized as a double-nested array, given the context.  Argh.

Can you test the following adjustment to your PR?  Instead of `!empty && sizeof > 0`, just `!empty()` should suffice, since `empty()` correctly treats a "sizeof == 0" as empty.

scottcwilson:
Yep - that works.

Issue #1944, opened 2018-12-20
Update error logging for PHP-FPM/nginx


lat9:
The current `includes/extra_configures/enable_error_logging.php` doesn't come into play on sites that are hosted on a PHP-FPM/nginx webhost due to that configuration's disallowing of a PHP `set_error_handler` override.

PR forthcoming.

scottcwilson:
Thanks for doing this, Cindy.

drbyte:
Hmmm ... I've been using php-fpm on nginx for ages, and it records errorlogs fine.
Will explore the PR.

barco57:
Cindy, is this related to cnc-specialty?

lat9:
> 
> 
> Hmmm ... I've been using php-fpm on nginx for ages, and it records errorlogs fine.
> Will explore the PR.
The previous version will record the backtrace information, but not the actual error.

lat9:
> 
> 
> Cindy, is this related to cnc-specialty?

Yes.

barco57:
Ok, I assume the title of this "logging for PHP-FPM/nginx" means PHP-FPM or nginx because nginx is not present on that server they are on.

PR #1945, opened 2018-12-20
#1944: Update error logging for PHP-FPM/nginx

lat9:
Fully take over the logging of errors, using the PHP error_log function to direct all non-fatal errors via the set_error_handler function.

For fatal errors, use a different function (since they don't go through the set_error_handler), using the register_shutdown_handler function.

### DIFF
--- a/includes/extra_configures/enable_error_logging.php
+++ b/includes/extra_configures/enable_error_logging.php
@@ -18,20 +18,56 @@ function zen_debug_error_handler ($errno, $errstr, $errfile, $errline)
     if (!(error_reporting() & $errno)) {
         return;
     }
-    ob_start();
-    if (version_compare(PHP_VERSION, '5.3.6') >= 0) {
-        debug_print_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
-    } else {
-        debug_print_backtrace();
+
+    $handled = true;                    //-For known error types, we'll handle them here.
+    switch ($errno) {
+        case E_NOTICE:
+        case E_USER_NOTICE:
+            $error_type = 'Notice';
+            break;
+        case E_DEPRECATED:
+        case E_USER_DEPRECATED:
+            $error_type = 'Deprecated';
+            break;
+        case E_WARNING:
+        case E_USER_WARNING:
+            $error_type = 'Warning';
+            break;
+        default:
+            $handled_here = false;      //-Unknown error type, let PHP's built-in handler do its thing.
+            break;
     }
-    $backtrace = ob_get_contents();
-    ob_end_clean();
-    // The following line removes the call to this zen_debug_error_handler function (as it's not relevant)
-    $backtrace = preg_replace ('/^#0\s+' . __FUNCTION__ . "[^\n]*\n/", '', rtrim($backtrace), 1);
-    $message = date('[d-M-Y H:i:s e]') . ' Request URI: ' . $_SERVER['REQUEST_URI'] . ', IP address: ' . (isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : 'not set') . PHP_EOL . $backtrace;
-    error_log(PHP_EOL . $message . PHP_EOL, 3, $GLOBALS['debug_logfile_path']);
+
+    if ($handled) {
+        ob_start();
+        if (version_compare(PHP_VERSION, '5.3.6') >= 0) {
+            debug_print_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+        } else {
+            debug_print_backtrace();
+        }
+        $backtrace = ob_get_contents();
+        ob_end_clean();
+        // The following line removes the call to this zen_debug_error_handler function (as it's not relevant)
+        $backtrace = preg_replace ('/^#0\s+' . __FUNCTION__ . "[^\n]*\n/", '', rtrim($backtrace), 1);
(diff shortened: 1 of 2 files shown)


proseLA:
@lat9 u know i love you and your work.  total respect.  

onto my pet peeve:
can't we put all of the code into the catalog facing side, and then on the admin side, include that file?  use the IS_ADMIN (or whatever that var is) as the differentiator for the file name.

it is SO frustrating to see the same code on both sides and this GREAT addition to ZC seems ripe to enforce better coding practices to me.  i know for the sites that i maintain, i do this exact thing that i preach.  it's just SO much easier to only have any function defined once.

lat9:
@proseLA, I agree and will see if I've got time over the next couple of days to do that merge.

includes/extra_configures/enable_error_logging.php
```diff
@@ -52,18 +99,30 @@ function zen_debug_error_handler ($errno, $errstr, $errfile, $errline)
...
+$debug_logfile_path = DIR_FS_LOGS . $log_prefix . $log_date->format('Ymd-His-u') . '.log';
+unset($log_prefix, $log_date);
 
 /**
  * Error reporting level to log
  * Default: E_ALL ^E_NOTICE
  */
 $errors_to_log = (version_compare(PHP_VERSION, 5.3, '>=') ? E_ALL & ~E_DEPRECATED & ~E_NOTICE : version_compare(PHP_VERSION, 5.4, '>=') ? E_ALL & ~E_DEPRECATED & ~E_NOTICE & ~E_STRICT : E_ALL & ~E_NOTICE);
-
+$errors_to_log = -1;
```
drbyte: Did you mean to set this to `-1` ? If so, then the line above it is unneeded.

includes/extra_configures/enable_error_logging.php
```diff
@@ -52,18 +99,30 @@ function zen_debug_error_handler ($errno, $errstr, $errfile, $errline)
...
+ *      - ii ..... is the 2-digit minute
+ *      - ss ..... is the 2-digit second
+ *      - xxxxx ... is the time in milliseconds
+ *
  *    (or if you don't have a /logs/ folder, it will use the /cache/ folder instead)
  */
-$debug_logfile_path = DIR_FS_LOGS . '/myDEBUG-' . time() . '-' . mt_rand(1000,999999) . '.log';
+$log_prefix = (IS_ADMIN_FLAG) ? '/myDEBUG-adm-' : '/myDEBUG-';
+$log_date = new DateTime();
+$debug_logfile_path = DIR_FS_LOGS . $log_prefix . $log_date->format('Ymd-His-u') . '.log';
```
drbyte: Is your use of milliseconds providing greater entropy than the 4-6-digit random number? The goal here is prevent unauthorized reading of the files.

includes/extra_configures/enable_error_logging.php
lat9: Nope, debug code.  See update.

includes/extra_configures/enable_error_logging.php
lat9: Since the milliseconds are, in essence, a 6-digit pseudo-random number, I believe that it's providing uniqueness.  The .htaccess in the /logs directory is also preventing that unauthorized reading.

drbyte: > The .htaccess in the /logs directory is also preventing that unauthorized reading.

Yes, on apache the .htaccess prevents access, but nginx ignores .htaccess so one can't depend on those directives.

> Since the milliseconds are, in essence, a 6-digit pseudo-random number, I believe that it's providing uniqueness.

I'll accept that for now.

includes/extra_configures/enable_error_logging.php
```diff
@@ -12,26 +12,73 @@
...
+        case E_ERROR:
+        case E_USER_ERROR:
+            $error_type = 'Fatal error';
+            break;
+        default:
+            $handled = false;      //-Unknown error type, let PHP's built-in handler do its thing.
+            break;
+    }
+
+    if ($handled) {
```
drbyte: One of the things we're embracing more and more now in ZC code is trying to avoid the arrow-antipattern (unnecessary `if` or nested conditionals), especially where blocks of code are wrapped inside those conditions when an early-return would help make the logic easier to read.

Do you mind updating this to something like:
```php
if (! $handled) {
  return false;
}
// ... the rest of the code that was formerly inside the `if` loop
```

lat9: No problems; I'll get that updated in a bit.

lat9: There you go ...

includes/extra_configures/enable_error_logging.php
```diff
@@ -52,18 +96,29 @@ function zen_debug_error_handler ($errno, $errstr, $errfile, $errline)
 
 /**
  * The path where the debug log file will be located
- * Default value is: DIR_FS_LOGS . '/myDEBUG-999999-00000000.log'
- * ... which puts it in the /logs/ folder:   /logs/myDEBUG-999999-00000000.log  (where 999999 is a random number, and 00000000 is the server's timestamp)
+ * Default value is: DIR_FS_LOGS . '/myDEBUG-yyyymmdd-hhiiss-xxxxx.log'
```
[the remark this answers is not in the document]
lat9: You're right, it's a picky comment; feel free to submit a PR to make that change.

BTW, that text wasn't trying to hide from anything nor did it _allude_ to the fact that the file's generated name in the admin is slightly different.

PR #1947, opened 2018-12-21
More admin top line fixes

scottcwilson:


proseLA:
i have not really looked at this too close; however, would it make more sense to put:

`require_once('includes/template/common/tplHtmlHead.php');
  require_once('includes/template/common/tplHtmlHeadLegacy.php');`

into:

admin/includes/application_top.php

?

zcwilt:
No. For 2 reasons.

First, application_top.php should be used purely for bootstrapping code.
Second, we currently have 2 types of pages in admin. 
Legacy pages, which these commits cover and new controller based pages that use a different method for including template code. 

Strictly speaking we could have foregone these type of commits in favour of converting all pages to new controller stuff. However that may be a longer term goal based on what we decide we need for a candidate release.

PR #1948, opened 2018-12-21
Add early returns in shopping_cart class to avoid errors from empty queries

mc12345678:


drbyte:
By "notification generation", are you referring to PHP Notices?

mc12345678:
There was a condition experienced a while back where not having some aspects of product in the cart (attributes or just a product in general) that a complaint would be logged when trying to access undefined/unassigned variables.  Had incorporated these modifications into a commit that touched several files, but in recent updates appears to have gotten lost (didn't see anything that explained why these few changes should not be incorporated).

As to the question of how they came to be known about and whether notices/notifications: some time ago based on previous suggestion I placed my server in strict mode to test my own code and to run Zen Cart.  Seeing how PHP is tightening up on the expectations of operation (and of course ideally to use positive controls on code and data condition) it just seems like these issues should be addressed otherwise there will be a flood of requests from those that continue to use the software or an abandonment.  I do not recall what form of announcement was made through the mydebug logs, but after initially testing `isset` against everywhere within each of those methods and the discussion of suggesting to instead do the test one time, I had resubmitted the pull to prevent the generation of mydebug log files while in strict php mode.  It was a few months ago when I had detected the issue(s) and yes, in a way was holding back.  There have been a lot of people helping make improvements and I would have expected that some of these would have been addressed ages ago.  But with the release of ZC 1.5.6 and PHP 7.3 being pickier than ever, I certainly wanted to get these out to support the next release.  If determined not to be incorporated then so be it at least it would have been fairly judged.  Thing is, I doubt much of the above rhetoric was necessary in order to explain whether notification or notice. :) Scrub what is needed. :)

drbyte:
I should have been more clear: I was really inquiring whether these "early return" `continue` statements were preventing generation of errors/alerts/notices/whatever, as your choice of words didn't make it clear, nor did the code changes. 

I had hoped you might just respond with more details or examples of what problem it was solving.

Armed with that information it's easier to do the code review, and to update the title to be more specific.

mc12345678:
To tighten up on the conversation, suggest reviewing the comments of #1679 for the choice of path taken; however, I believe I may have lost the initiating event.  I know that in some other code generation I've been trying to tighten up on the submission of attribute information such that if attribute information is submitted that does not exist that it is handled.  

Right now at least where these "early returns" occur the general thought is that there would be no data if and only if that designated "record" (`$option=>$value`) didn't exist.  

Perhaps I also saw something in the `$this->contents[$products_id]['attributes']` that didn't lead to an appropriate pairing when dealing with one of the various attribute types (yeah potential goose chase here?).

I do know and still have commented out that I had inserted several `isset` tests for the weight, price prefix, price_discounted, etc... where the reasoning was that there was no data returned from the initial query.  Looking at the current default install, it seems that these fields have default values so if the query returned any row then there would still be an expected value by which to compare and execute the lines below the code inserted.  But not having the value/data assigned to the `'attributes'` array item, that would cause the absence of data returned and at least while in php strict mode the messages that were provided.

BTW, would think that could have actually been less specific: what caused the issue being corrected? :)

PR #1949, opened 2018-12-21
Address some strict notifications surrounding admin sanitizers.

mc12345678:


drbyte:
Would it make sense to just make sure the array entry exists, instead of adding a bunch of `isset` stuff to the code?

mc12345678:
I'm assuming the question is not one of being a smart alec. (e.g. if wanting to see why the heater isn't running have you checked that it's cold outside?) But instead of one asking what code modification might be best to address the situation.

Are you asking if there should be something like:
````php
    private function runSpecificSanitizer($parameterName, $parameterDefinition)
    {
        if (!isset($this->adminSanitizerTypes[$parameterDefinition['sanitizerType']])) {
            // Perhaps provide some sort of notification then exit this code because nothing else can be done.
            //  Isset used here because there would be an expectation of this sanitizerType to have
            //    some value appended to not cause it to be null and because isset has been considered faster
            //    than in_array.  
            return;
        }
        if ($this->adminSanitizerTypes[$parameterDefinition['sanitizerType']]['type'] === 'builtin') {
            $this->processBuiltIn($parameterDefinition['sanitizerType'], $parameterName, $parameterDefinition);
        }
        if ($this->adminSanitizerTypes[$parameterDefinition['sanitizerType']]['type'] === 'custom') {
            $this->processCustom($parameterDefinition['sanitizerType'], $parameterName, $parameterDefinition);
        }
    }
````

Call it perhaps over optimizing, but general research has shown that `isset` is faster than `in_array` and historically can be used alone if there is no expectation of the value being tested of actually being set to `null` and still being present/detected.  If such a case were possible then the generally (historically) suggested method is to `||` (or) `isset` with `in_array` so that if the `isset` determines `false` because the value is `null` that `in_array` will return `false` only if the field truly is not present in the array but will return `true` for the field that is present but set to `null`. So, then on the optimization aspect, well, for general use of the admin from the manual admin perspective, either method of `isset` or `in_array` is going to have minimal impact.  But I think I came across this when working on an automated method of updating database data even if by accident of typing something in incorrectly and getting a load of errors. :)

drbyte:
> I'm assuming the question is not one of being a smart alec. 

Correct assumption. 

I'm trying to make sure that we don't add unnecessary code. And I really hate having unnecessary `isset` wrappers "just because".  It makes the code harder to read, and IMO if we need to litter the code with `isset` then it suggests a code-smell that should be fixed instead of bandaged. Granted, I understand that we're limited by certain architectural constructs in ZC's core, so sometimes the desired cleanups are not practical.

drbyte:
By suggesting to make sure the array entry already exists, I meant that "much earlier in the code" we could actually initialize it, so that it's reasonable to expect its existence instead of always having to test for it. I mean, our code depends on these being set, so why don't we make sure of that earlier, instead of having to test it way down the line in this function.

drbyte:
In studying the code for awhile it seems that the only way for `['type']` to not be set is for custom registrations of additional sanitizers.

While testing with `isset` is one way to protect against invalid submissions, I wonder if the logic should be augmented further to check whether it's an incoming `custom` sanitizer? ie: if `['type']` isn't set (mistakenly), but `['function']` is present, then treat it as custom.

But, before going further, can you supply some example code that was triggering the need to add these `isset` protections? I'd like to see a use-case, since I'm not finding one in core code.

mc12345678:
Well, I think that I came across it when I was trying to add the filterSimpleArray where I considered creating it as a custom filter but then.. well basically I found it better to have worked on the inner core of the sanitization than to try to work around the way it was.  I have spent some time looking through my systems to see if I still had remnants of what I started, but I don't seem to be coming up with any of the old code which I would have expected to have just renamed so as not to load.

Yes, though I believe that was where/when it was discovered.  In part I'd be fine with tabling this one as I believe I have just maintained it as part of my filset for when/if I ever write a custom type filter. considering that the function is private in my opinion further supports the evaluation that it is custom filter dependent.

mc12345678:
I will say that the custom filter method (as well as the core filter) was in its infancy and there was little to no documentation on how to create one's own custom filter, so yes having some sort of "verification" earlier in the process rather than having `isset` inside `runSpecificSanitizer` is definitely a good idea. Amplifying the expectations of how to develop a proper custom filter would be of help as well (at least then could point at something to say this is why all these messages are being created.

I guess at the time I recognized that there was an issue, discussions had eluded to this method of sanitization as a temporary measure, and I didn't know if the development had stopped at this function with the idea that when/if someone wanted to create a custom filter that it would then be flushed out or what.  Basically I didn't want to devote time to something that was to be wasted or already developed  by someone else in the team (the time of each of you is important as well). 

So... I now I have "dropped the mic". :)

PR #1950, opened 2018-12-21
Catch some possibly missing variables in tpl_index_product_list.php

mc12345678:


includes/templates/template_default/templates/tpl_index_product_list.php
```diff
@@ -65,7 +65,7 @@
   if (isset($_GET['typefilter']) && $_GET['typefilter'] != '') echo zen_draw_hidden_field('typefilter', $_GET['typefilter']);
 
   // draw manufacturers_id if not already done earlier
-  if ($get_option_variable != 'manufacturers_id' && isset($_GET['manufacturers_id']) && $_GET['manufacturers_id'] > 0) {
+  if (!(isset($get_option_variable) && $get_option_variable == 'manufacturers_id') && isset($_GET['manufacturers_id']) && $_GET['manufacturers_id'] > 0) {
```
drbyte: The second half of this line could be converted to an `empty` check as well.

PR #1951, opened 2018-12-21
Simplify checks of whether customer is logged in, during shipping-estimator

mc12345678:
testing for `$_SESSION['customer_id']` being php empty or not.


includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
```diff
@@ -85,7 +85,7 @@
     }else{
 ?>
 <table id="seQuoteResults">
-<?php if ($_SESSION['customer_id'] < 1 ){ ?>
+<?php if (!isset($_SESSION['customer_id']) || $_SESSION['customer_id'] < 1 ){ ?>
```
drbyte: Doesn't `empty` work here too? If not, why?

PR #1952, opened 2018-12-21
V156 shipping strict

mc12345678:
This modified a file that was modified in a previous commit, but wasn't entirely sure how to "separate" but combine all items.  So, result is that if this were committed separately from the previous commit then there would be merge errors involved which I was trying to avoid for all.

### DIFF
--- a/includes/modules/shipping_estimator.php
+++ b/includes/modules/shipping_estimator.php
@@ -48,6 +48,7 @@ if ($_SESSION['cart']->count_contents() > 0) {
   // deprecated; to be removed
   if (file_exists(DIR_WS_CLASSES . 'http_client.php')) require_once(DIR_WS_CLASSES . 'http_client.php'); // shipping in basket
 
+  $sendto = 0;
 
   if (!empty($_SESSION['customer_id'])) {
     // user is logged in
@@ -88,7 +89,7 @@ if ($_SESSION['cart']->count_contents() > 0) {
       //add state zone_id
       $_SESSION['cart_zone'] = $state_zone_id;
       $_SESSION['cart_zip_code'] = $zip_code;
-    } elseif (!empty($_SESSION['cart_country_id'])){
+    } elseif (!empty($_SESSION['cart_country_id']) && !empty($_SESSION['cart_zip_code'])){
       // session is available
       $_SESSION['country_info'] = zen_get_countries($_SESSION['cart_country_id'],true);
       $country_info = $_SESSION['country_info'];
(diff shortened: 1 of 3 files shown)


includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
```diff
@@ -85,11 +85,11 @@
     }else{
 ?>
 <table id="seQuoteResults">
-<?php if ($_SESSION['customer_id'] < 1 ){ ?>
+<?php if (!isset($_SESSION['customer_id']) || $_SESSION['customer_id'] < 1 ){ ?>
```
drbyte: `empty`?

includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
```diff
@@ -98,14 +98,17 @@
        <th scope="col" id="seTotalHeading"><?php echo CART_SHIPPING_METHOD_RATES; ?></th>
      </tr>
 <?php
+      if (empty($extra)) {
+        $extra = '';
+      }
       for ($i=0, $n=sizeof($quotes); $i<$n; $i++) {
-        if(sizeof($quotes[$i]['methods'])==1){
+        if(isset($quotes[$i]['id']) && sizeof($quotes[$i]['methods'])==1 && isset($quotes[$i]['methods'][0]['id'])){
```
drbyte: Should `$thisquoteid` be initialized here, similar to what you added later?

drbyte:
@mc12345678 BTW, thanks for catching all these.

includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
mc12345678:
```suggestion
<?php if (empty($_SESSION['customer_id'])){ ?>
```
Certainly makes sense to me. I couldn't think of a specific case where the customer_id would be non-integer and less than 0 considering this had specifically evaluated to less than 1....  Don't know impact on anything else that has "used" this code.

includes/modules/shipping_estimator.php
```diff
@@ -150,10 +151,10 @@ function shipincart_submit(){
   }
   // begin shipping cost
   if(!$free_shipping && $_SESSION['cart']->get_content_type() !== 'virtual'){
-    if (zen_not_null($_POST['scid'])){
+    if (isset($_POST['scid']) && zen_not_null($_POST['scid'])){
```
drbyte: FYI - I've also been converting "isset && zen_not_null" to `empty`, unless I'm specifically expecting a `falsy` value, which is rare.

includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
mc12345678: Depends on overall code structure goal.  The else of this `if` statement is where an initialization of `$thisquoteid` was added.  The first step within this first `if` statement was to make an assignment to the variable `$thisquoteid`.  

So, if there is a goal to perhaps follow some other programming languages where every variable that is to be used is to be declared at the onset of the function and possibly provided an "initial" or default value, then yes would suggest initializing `$thisquoteid` to an empty string early on and/or possibly on each `for` loop. Probably would be more important to have the value `$thisquoteid` defined to something if there were any other location later in the code that needed to use the variable.

So, even though I think by this point the code has already been incorporated, doesn't mean that further changes aren't possible/necessary. :)

drbyte: While there's no specific goal to pre-initialize every possible variable, PHP's embracing of strictness causes the need for all these `isset` checks if the variable is going to be passed or its contents evaluated.

I had observed your pattern, and wondered if you had intentionally skipped initializing it in the first instance when you had initialized it in the second.

PR #1954, opened 2018-12-22
Change spiders.txt to regex rules pkg

drbyte:


drbyte:
I really wanted this to make the init_sessions code shorter. 
But all the `require` statements bloat it. 

Why aren't we just doing PSR-4 autoloading?

includes/autoload_namespaces.php
```diff
@@ -41,6 +42,7 @@
 define('URL_CHECKOUTFLOW', 'zencart/CheckoutFlow/src/');
 define('URL_VIEW', 'zencart/View/src/');
 define('URL_ADMINUSER', 'zencart/AdminUser/src/');
+define('URL_CRAWLER_DETECT', 'jaybizzle/crawler-detect');
 define('URL_PHPMAILER', 'PHPMailer/src/');
```
zcwilt: Shouldn't this point to the src  directory

zcwilt: probably why autoloading is failing

PR #1959, opened 2018-12-22
Handle bad cmd= value

scottcwilson:
Go to admin home rather than creating a debug log/blank page when command is misspelled.

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -34,9 +34,9 @@ if ($foundAction) {
 } else {
     $filename = preg_replace('/[^a-zA-Z0-9_-]/', '', $cmd) . '.php';
     if (file_exists($filename)) { 
-       require(preg_replace('/[^a-zA-Z0-9_-]/', '', $cmd) . '.php');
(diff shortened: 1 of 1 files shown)


admin/index.php
```diff
@@ -32,5 +32,11 @@
 if ($foundAction) {
     exit(0);
 } else {
-    require(preg_replace('/[^a-zA-Z0-9_-]/', '', $cmd) . '.php');
+    $filename = preg_replace('/[^a-zA-Z0-9_-]/', '', $cmd) . '.php';
+    if (file_exists($filename)) { 
+       require(preg_replace('/[^a-zA-Z0-9_-]/', '', $cmd) . '.php');
```
drbyte: can this second `preg_replace` be replaced with `$filename`?

Also, `require` doesn't require parentheses. We're progressively removing parentheses for require/include statements and their `_once` equivs.

scottcwilson:
Thanks Chris.  Looks good!

Issue #1970, opened 2018-12-23
Specials date validation fails to work


scottcwilson:
Javascript errors occur when entering invalid dates by typing into a date field. 

<img width="1333" alt="screen shot 2018-12-23 at 2 46 29 pm" src="https://user-images.githubusercontent.com/4391638/50387041-ab114000-06c1-11e9-9cd9-977f39374490.png">

drbyte:
Is this new to v1.5.6?

scottcwilson:
Yes.  Fine in 1.5.5.

scottcwilson:
Argh, I'm sorry that's not true.  It *was* a bug in 1.5.5 as well - I just didn't notice it.

scottcwilson:
The date validation seems not to work at all.  If you type 13/25/2018 in one of the dates and press update, it gives you a log and a white page.  I have submitted a fix for this.

scottcwilson:
Fixed by #1978 except for Products Price Manager.  New issue #1982 opened for that work. 
Please close this ticket.

PR #1978, opened 2018-12-23
Fix date check where spiffyCal is used

scottcwilson:
Date check not correctly done in specials.  (Same bug exists in other places; I can fix if you accept this.)

### DIFF
--- a/admin/banner_manager.php
+++ b/admin/banner_manager.php
@@ -305,7 +305,7 @@ if (zen_not_null($action)) {
 
         <div class="row">
             <?php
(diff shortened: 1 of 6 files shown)


scottcwilson:
Fixes #1970 plus other bugs.
To test: use an invalid date (13/25/2018) in a special start or end date.  Without this fix it will produce a debug log and white page.

scottcwilson:
Other admin functions which likely have this same problem:
./banner_manager.php
./coupon_admin.php
./featured.php
./products_price_manager.php
./salemaker.php

drbyte:
The proposed fix looks good. Go ahead with the additional pages.

scottcwilson:
OK I did all of them except products price manager.  Please merge this and create a new ticket for PPM and I'll do that next week. It's more complicated.

PR #1981, opened 2018-12-27
V156 zc install - add parser for left join queries

mc12345678:
There have recently been multiple posts to the ZC forum about the upgrade process not completing successfully even though the screen message alludes to everything being done.  While some of the issues appear to be related to date or datetime fields having '0000-00-00' stored in them (beyond the date_modified field already addressed) also identified a few items as captured in the comments for each commit of this pull request.  One of the issues was that the sql statement to left join a table did not account for the `DB_PREFIX`.

EDIT: Recommend a method to evaluate the table data stored to see if the table's field definition is date or datetime and then if it could support a value of `0000-00-00` and if not then to modify the value either to `null` (if supported) or `0001-01-01`, though also depends on what caused it and what can be done to handle the reset value.  At the very least, suggest offering more/some information about the issue on the zc_install screen..

### DIFF
--- a/zc_install/includes/classes/class.zcDatabaseInstaller.php
+++ b/zc_install/includes/classes/class.zcDatabaseInstaller.php
@@ -139,6 +139,7 @@ class zcDatabaseInstaller
   private function parseLineContent()
   {
     $this->lineSplit = explode(" ",(substr($this->line,-1)==';') ? substr($this->line,0,strlen($this->line)-1) : $this->line);
+    if (!isset($this->lineSplit[3])) $this->lineSplit[3] = "";
     if (!isset($this->lineSplit[4])) $this->lineSplit[4] = "";
     if (!isset($this->lineSplit[5])) $this->lineSplit[5] = "";
     foreach ($this->basicParseStrings as $parseString)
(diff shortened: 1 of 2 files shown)


zc_install/includes/classes/class.zcDatabaseInstaller.php
```diff
@@ -352,20 +352,31 @@ public function parserRenameTable()
...
       }
     }
   }
+  public function parserLeftJoin()
+  {
+    if (!$this->tableExists($this->lineSplit[2]))
+    {
+      if (!isset($result)) $result = sprintf(REASON_TABLE_NOT_FOUND, $this->lineSplit[2]).' CHECK PREFIXES!';
+      $this->writeUpgradeExceptions($this->line, $result, $this->fileName);
+      error_log($result . "\n" . $this->line . "\n---------------\n\n");
```
drbyte: Did you mean to leave this `error_log` call?

drbyte:
Thanks for these updates!

zc_install/includes/classes/class.zcDatabaseInstaller.php
mc12345678: Now able to respond in this location, I had intended the error_log to be here so that a `zcInstallDEBUG-xxxxxxxxxx-xxxxxx` file would be created in addition to populating a `zcInstallLog_xx-xx-xxxx_xx-xx-xx-xxxxxx` file though both are populated with information in this situation.

PR #1983, opened 2018-12-28
Fix link in music product-type admin files

mesnitu:
Update to the form of the product type musica, IF this was the spectacle procedure


drbyte:
Thanks!

PR #1383, opened 2017-02-02
Fix calc for attr qty prices onetime

drbyte:
Fixes #1252

PR #1985, opened 2018-12-28
Fix price-onetime function call parameters

drbyte:
Wrong parameters passed. 

As reported by @lat9 at: https://www.zen-cart.com/showthread.php?224245-Shopping-cart-class-invalid-function-parameters


drbyte:
Merging as it's essentially a backport of #1383

PR #1987, opened 2018-12-29
Refresh Admin Login Page Styling

### LINKED ISSUE #1898: admin login page hard to read in 1.5.6
Could some smart guys with CSS skillz please fix this? @Zen4All ? 
screen shot 2018-12-14 at 7 13 15 am

dakanji:
Refresh Admin Login Page Styling. Fixes #1898


drbyte:
Tell me, why are you adding "yet another stylesheet" to maintain?

Surely the same outcome can be accomplished by using the existing CSS framework.

dakanji:
The existing stylesheet needs a complete overhaul.
Thought best way to get there was to go in steps. such that this alt one ultimately takes over and renamed back to just "stylesheet".

Couldn't be sure of other impacts if the existing one was just amended within this context.

dakanji:
Knowing that additional stylesheets might be an issue, what I have started, but will make as a separate item, will be to move the "zen.css" file from the docs folder and merge these into that file (v156). 

Then build on this to refresh the general admin styling (v200). Basically picking up on #996

drbyte:
> move the "zen.css" file from the docs folder and merge these

I was afraid you'd say that. I'm still not a fan of what was done there.

"Fixing" the admin login should be the sole focus of this PR.  (granted, I still haven't seen an actual reason why it's "broken")

Bootstrap has been embraced in the Admin, for better or for worse. Let's stick with that. We're not going to overhaul the CSS in v156 patch releases.

dakanji:
You need to be a bit more precise.

When you say you are not a fan of what was done, I take it you are not saying you do not like the resulting PRESENTATION (Look/Feel a user gets) but that you specifically mean unease about the method of how RESPONSIVENESS (Break point definition and nomenclature) was achieved. 

If that is indeed the case, then RESPONSIVENESS is something that can be drawn from Bootstrap easily and needs to be done for consistency with the rest of the admin.  

Can you confirm?

drbyte:
Okay. The longer-term plan helps with understanding context and what to expect next.

Steps 1 and 2 are okay for now.

In chatting with @zcwilt, for v2 we'd like the Admin UI changes to be done iteratively, starting with an Issue including screenshots of the proposed change (before and after might be helpful). Given that making the changes my require doing all the coding and therefore have code for a PR anyway, you could submit it as a PR instead of an Issue. But it's important that the reasoning behind the changes be described, as well as the screenshots.

Additionally, it's probably fair to mention now that we'd like to move the v2 Admin over to Bootstrap 4, to benefit from its broader breakpoint range (better for landscape mobile devices) and flexbox features. The reason for sticking with Bootstrap is to avoid having to write and maintain a complete set of docs on how to do styling and breakpoints etc for a bespoke CSS framework of our own making -- there's no need for that.

drbyte:
For this PR, given that you've done some of the Bootstrap combining work for the Docs side, do you want to do that inside this PR? Or should we merge as-is and you can do the combining separately? Doing it in this PR is still canonical about the login page subject at hand.

dakanji:
Think it is best to do things in discrete chunks in terms of PRs so will tackle combining separately and this can go ahead as it stands.

Bootstrap is a convenient and well understood framework so makes sense to adopt. 

As for BS4, as you will see from [this forum post](https://www.zen-cart.com/showthread.php?219608-The-future-Bootstrap-or-Foundation-And-why&p=1305141#post1305141), I already thought BS4 should have been used for v200 all the way back in 2016!

PR #2003, opened 2019-01-01
Admin define TEXT_IMAGE_NONEXISTENT missing in product.php

scottcwilson:
The products page in Admin->Catalog->Categories products needs this
define.


drbyte:
For v156, okay.
For v2 we should consolidate this define into one place.

scottcwilson:
@drbyte Do you want a wholesale de-duplication of admin defines in v2?  Or should I just handle this one?

drbyte:
Just this one for now. If you want to create an Issue exposing the scope of duplication that currently exists in v2, we can talk there about which things to retain duplicates for (and why), etc.

drbyte:
Given that there are 2 strings used in the duplicates for this current define, let's settle on one string, not multiple defines.
This seems to be the most common: 
`define('TEXT_IMAGE_NONEXISTENT', 'IMAGE DOES NOT EXIST');`

PR #2012, opened 2019-01-02
Add extra sql queries for ez pages during language deletes

Zen4All:
Since the ez-page are now multilingual, the new table should also be updated when adding or removing a language

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2012)

### DIFF
--- a/admin/languages.php
+++ b/admin/languages.php
@@ -173,7 +173,7 @@ if (zen_not_null($action)) {
           $db->Execute("INSERT INTO " . TABLE_EZPAGES_CONTENT . " (pages_id, languages_id, pages_title, pages_html_text)
                         VALUES ('" . (int)$coupon['pages_id'] . "',
                                 '" . (int)$insert_id . "',
-                                '" . zen_db_input($ezpages['pages_title']) . "',
+                                '" . zen_db_input($ezpage['pages_title']) . "',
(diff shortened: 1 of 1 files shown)


admin/languages.php
```diff
@@ -164,6 +164,19 @@
...
+        $ezpages = $db->Execute("SELECT pages_id, pages_title, pages_html_text
+                                 FROM " . TABLE_EZPAGES_CONTENT . "
+                                 WHERE languages_id = " . (int)$_SESSION['languages_id']);
+
+        foreach ($ezpages as $ezpage) {
+          $db->Execute("INSERT INTO " . TABLE_EZPAGES_CONTENT . " (pages_id, languages_id, pages_title, pages_html_text)
+                        VALUES ('" . (int)$coupon['pages_id'] . "',
+                                '" . (int)$insert_id . "',
+                                '" . zen_db_input($ezpages['pages_title']) . "',
+                                '" . zen_db_input($ezpage['pages_html_text']) . "')");
```
drbyte: Shouldn't these `$ezpages[]` vars be singular `$ezpage[]` to match the foreach?

Zen4All: yes

drbyte:
Thanks!

PR #2016, opened 2019-01-03
Fix broken admin login form

drbyte:
Removed HTML markup from inside function parameter, so submit button can work.

Reformatted HTML to put line-breaks in the right spots.


scottcwilson:
Tested this AM; works perfectly.

PR #2018, opened 2019-01-03
Correct field type in zc_install check, allowing upgrade to complete

scottcwilson:
Without this fix, the 1.5.6->2.0 upgrade does not work; thinks it is still needing 1.5.5->1.5.6. 

Upgrade to version 1.5.6 completed.
Could not update to version 1.6.0. We detect that you currently have v1.5.5, and must perform the updates to get to version 1.5.6 first.


drbyte:
eeks - I'd forgotten that it wants uppercase for those checks.

PR #2021, opened 2019-01-03
v156 admin/includes/modules/update_product.php non-posted master_cate…

mc12345678:
…gory

For a product that is not linked, it will not have a master_category
to post to the preview_info page and therefore will not have
a master_category field to post to update_product.php.  When that
field is not posted and strict processing is active, trying to
cast that to an integer will cause a PHP notice.  This detects
the presence of the posted value and then in abscence of it
expects that the master_categories_id will be posted (even if 0)
and will perform the type conversion on that expected posted value.


drbyte:
Oh how I wish we could just use PHP 7's null-coalesce operator ... so much tidier.

mc12345678:
Yup.  Thought of some things like that when touching files.

drbyte:
Does this same thing need to be done to the `update_product.php` in the music product-type?

PR #2026, opened 2019-01-04
Fix rejected coupon amounts for percent-off

scottcwilson:
See https://www.zen-cart.com/showthread.php?224853-Coupon-not-working-1-5-6&p=1352909#post1352909


scottcwilson:
hang on ... testing another approach ...

PR #2042, opened 2019-01-06
Non-existent variable `$languages_id` used.

mc12345678:
Discovered this while reviewing mydebug files generated while
strict processing was enabled and as a result of verifying a
recent correction made for the buttons on the category listing
page.

replaced $languages_id with $_SESSION['languages_id']

Had done a search of at least the admin with no result found for
$languages_id other than in this file:
admin/includes/category_product_listing.php


drbyte:
Good catch.

PR #2043, opened 2019-01-06
Refactored zen_get_categories_name_from_product.

mc12345678:
Identified by a self-implemented coding error that attempting to
access this function with a non-existent products_id caused a
mydebug notice (in strict mode) to be logged for
`$check_products_category->fields['master_categories_id']`
not existing.  Execution continued apparently, but it seems that
it could be improved.  Also removed the text casting quotes around
the values and ensured each were cast as an integer.


drbyte:
Thanks

PR #2044, opened 2019-01-06
Remove duplicated functions from admin.

scottcwilson:


Zen4All:
You can not replace zen_draw_form as the admin side has a different parameter pattern then the front side.

drbyte:
Since this is for v2, maybe it's time to update the v2 admin files to use the same catalog-side function? Which means updating every zen_draw_form call in the admin to ensure its parameters are submitted in the right order.

scottcwilson:
Thanks for catching that @Zen4All.  I tested a bunch of things but didn't test a form submission. 

@drbyte Maybe ... but this is not a high value change.  Can we get focused on the key issues like building a template for v2?

lat9:
... or for downward compatibility, re-define the function as taking a variable number of arguments (using the splat parameter) and then process them in a different order for the admin vs storefront.

PR #2048, opened 2019-01-06
Additional centerbox notifiers for plugin use

lat9:
... to allow a plugin to make modifications to the generated list-box-contents.


lat9:
Thanks, DrByte; that will help in future updates to a couple of my plugins.

PR #2055, opened 2019-01-08
Force config key addition if missing in authorizenet

scottcwilson:


scottcwilson:
If approved I will make the same change to authorizenet.php, which has the same issue.

drbyte:
Approved.

PR #2061, opened 2019-01-08
Prevent "Admin Admin" from being in a page title

scottcwilson:
Check for the string "Admin" at the start of the name before adding it as a prefix.

### DIFF
--- a/admin/includes/init_includes/init_templates.php
+++ b/admin/includes/init_includes/init_templates.php
@@ -43,5 +43,9 @@ $pagename = ucwords($pagename);
 if ($pagename == '') {
   $pagename = STORE_NAME;
 }
-$title = TEXT_ADMIN_TAB_PREFIX . ' ' . $pagename;
+if (strncmp(TEXT_ADMIN_TAB_PREFIX, $pagename, strlen(TEXT_ADMIN_TAB_PREFIX)) == 0) {
+   $title = $pagename;
+} else {
+   $title = TEXT_ADMIN_TAB_PREFIX . ' ' . $pagename;
+}
 define('TITLE', $title);


dakanji:
How does this impact non English set ups?

scottcwilson:
It uses the define values, not the actual string.

admin/includes/init_includes/init_templates.php
```diff
@@ -43,5 +43,9 @@
 if ($pagename == '') {
   $pagename = STORE_NAME;
 }
-$title = TEXT_ADMIN_TAB_PREFIX . ' ' . $pagename;
+if (strncmp(TEXT_ADMIN_TAB_PREFIX, $pagename, strlen(TEXT_ADMIN_TAB_PREFIX)) == 0) {
+   $title = $pagename;
+} else {
+   $title = TEXT_ADMIN_TAB_PREFIX . ' ' . $pagename;
+}
```
mc12345678: Isn't the current recommendation for if/else style code to keep the original line but if it meets the proposed condition reassign it?
I.e.
```diff
$title = TEXT_ADMIN_TAB_PREFIX . ' ' . $pagename;
+if (strncmp(TEXT_ADMIN_TAB_PREFIX, $pagename, strlen(TEXT_ADMIN_TAB_PREFIX)) == 0) {
+    $title = $pagename;
+}
```

scottcwilson:
Generally that's a good policy but in this case, for clarity, I think the current logic is better and easier to read.  This style of titling based on page name is new for Zen Cart and people might want to change it.

admin/includes/init_includes/init_templates.php
scottcwilson: If you got to the home page and look at the title in your browser tab you'll see it says "Admin Admin Home."  That was the driver for this change.

mc12345678: Ok, I do think I've seen it, but thought it "had always been that way". :)

mc12345678:
Thought that in either code "style" that would be possible, I do somewhat like being able to have the additional information, though.

PR #2066, opened 2019-01-09
Cast the year in zen_date_short to int

mc12345678:
Recently a forum member identified a problem with ZC 1.5.6 where an error was thrown I think on line 208 indicating that parameter 6 was expected to be an integer; however, a string was provided.  Now that I have had a look at the software on a larger screen, it appears that the issue stems from `$year` not being cast as an integer.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -196,7 +196,7 @@
   function zen_date_short($raw_date) {
     if ( ($raw_date == '0001-01-01 00:00:00') || ($raw_date == '') ) return false;
 
-    $year = substr($raw_date, 0, 4);
+    $year = (int)substr($raw_date, 0, 4);
     $month = (int)substr($raw_date, 5, 2);
     $day = (int)substr($raw_date, 8, 2);
     $hour = (int)substr($raw_date, 11, 2);


mc12345678:
So, yesterday when I was trying to look over the issue reported, I discovered that the 'parameter 6' error was not at all related to these date functions but instead the zen_date_diff function.  Unfortunately the individual has not provided back any data to further identify the issue, but I could see that perhaps the lack of the integer casting was omitted as a sort of "future" test? Like, has php changed enough to prevent this from working? Sure, it seems annoying that it is not consistent, but does is that inconsistency hurting anything?

What I'm saying is that if there is some thought that there perhaps was a reason for it to not be cast as an `int`, then it almost seems like it could stay that way.  The problem I thought I was correcting didn't exist in this function but was observed in another and that one doesn't even try to cast the data to numbers.

drbyte:
After a bunch of testing (not on a Windows version of PHP though) I don't think there'll be a problem either way (keeping as-is or casting to int).

mc12345678:
Agree that that difference seemed unusual.  We are ever approaching the possibility of using 2037 (less than 20 years away), it would seem that the Unix Epoch side step should be applied elsewhere additionally.  Something that is/was interesting as I was researching the recent questions was that when I tried an online php tool, it did not have trouble returning a date well into the future in any php environment.  Makes me wonder how they did that, I mean even a year such as 19720 came back fine. :/ Certainly expect that the machine was more than a 32 bit device, but...

The other side to that was when I entered a date of `mkdate(0, 0, 0, 0, 0, 0)` and presented the year it came back as 1999....  

I realize by email that this issue is closed, but I wanted to finish off this response, regardless. :)

PR #2067, opened 2019-01-09
Return false from date functions evaluated to 0000-00-00 00:00:00

mc12345678:
I have been trying to track down how a `date` field in a plugin ended up being stored with `0000-00-00` and at least the last place where assignment was made, a Zen Cart date function was used in a similar manner as other `date`/`datetime` functions.  Looking into the associated admin date functions, it appears that if the `datetime` of `0001-01-01 00:00:00` or an empty string (possibly by loose comparison) was fed into the functions then false is returned; however, there is no similar review/response if the result were `0000-00-00 00:00:00` which at least currently for `date` and `datetime` fields is considered to not be acceptable.

Perhaps a similar review should be done for a year of 1, month of 1, day of 1 and empty for the other three fields. Reason? A `date` field may have `0001-01-01` but will not include the time portion and therefore will evaluate back in the `DATE_TIME_FORMAT` as equivalent to `0001-01-01 00:00:00`... Just some thoughts.

I would not be disappointed if the reason this is not incorporated is that perhaps there is a non-database reason to actually return `0000-00-00 00:00:00` and therefore not be incorporated/included.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -176,7 +176,7 @@
   }
 
   function zen_date_long($raw_date) {
-    if ( ($raw_date == '0001-01-01 00:00:00') || ($raw_date == '') ) return false;
+    if (empty($raw_date) || $raw_date <= '0001-01-01 00:00:00') return false;
 
     $year = (int)substr($raw_date, 0, 4);
     $month = (int)substr($raw_date, 5, 2);
@@ -184,8 +184,6 @@
     $hour = (int)substr($raw_date, 11, 2);
     $minute = (int)substr($raw_date, 14, 2);
     $second = (int)substr($raw_date, 17, 2);
-    
-    if (empty($year) && empty($month) && empty($day) && empty($hour) && empty($minute) && empty($second)) return false;
 
     return strftime(DATE_FORMAT_LONG, mktime($hour, $minute, $second, $month, $day, $year));
   }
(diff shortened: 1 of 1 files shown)


drbyte:
In a perfect world we'd never use zero-dates. We'd store/read null instead. But nulls can be tricky, and a bunch of code relies on zeros (or now 0001-etc) to equate to empty/unspecified.

There are 93 fields in the default schema that have date/time/datetime, and most weren't built to handle null. It's a major overhaul to convert to null. But it is the best solution IMO.

Do you think `return false` in this PR is safe everywhere?

drbyte:
In looking further, I think I'd prefer that this be changed to update the code earlier in these functions for:
```php
$raw_date == '0001-01-01 00:00:00'
```
from using `==` to using less-than-or-equal: `<=`

(and remove the (pointless) extra (duplicate) parentheses in the `if` statement)

This becomes less code to maintain, and feels like it captures the same overall meaning.

mc12345678:
That would involve removing the check against `== ''` as well correct?

Feel like need to also evaluate if need to check against `!isset($rawdate)` resulting in this condition check: `!isset($rawdate) || $raw_date <= '0001-01-01 00:00:00`?

drbyte:
Since `$rawdate` is a required parameter to the function, it's gonna be `set`. However, a test for `empty` is probably good, since that catches null too. Both `empty` and `<` will essentially catch the empty string.

So `if (empty($raw_date) || $raw_date <= '0001-01-01 00:00:00') return false;` probably suffices.

PR #2068, opened 2019-01-09
Incorporate date corrections.

mc12345678:
Incorporates the `date`/`datetime` sql posted by @drbyte at Zen Cart post: https://www.zen-cart.com/showthread.php?224861-db-upgrade-does-not-fix-all-incorrect-datetime-values&p=1353021#post1353021

Several seem to have had success with this sql and hadn't seen it yet proposed for incorporation here in github.

### DIFF
--- a/zc_install/sql/updates/mysql_upgrade_zencart_156.sql
+++ b/zc_install/sql/updates/mysql_upgrade_zencart_156.sql
@@ -6,7 +6,7 @@
 # * @copyright Copyright 2003-2019 Zen Cart Development Team
 # * @copyright Portions Copyright 2003 osCommerce
 # * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
-# * @version $Id: DrByte 2019 Jan 04 Modified in v1.5.6a $
+# * @version $Id: DrByte 2019 Jan 09 Modified in v1.5.6b $
 #
 
 ############ IMPORTANT INSTRUCTIONS ###############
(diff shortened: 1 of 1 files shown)


drbyte:
I haven't included it yet, because if someone is upgrading from another version prior to v156 then putting this in the v156 SQL doesn't catch the cases where people are upgrading from say 150-to-151 where other datetime fields are updated, and those would fail if there was bad data in them.

There's no hook currently in zc_install to pre-run sql updates. Only post.

mc12345678:
Doh...

Understood, in order to perform the update to 1.5.6, whatever previous update sql needs to be run and will experience a problem at the associated table(s).  

Did take a look through the code as well and agree don't necessarily see a great built in way to pre-apply the common sql code for action other than perhaps duplicating a little bit of code in the update path as loaded in includes/extra_configures though there does not appear to be an init_includes style path, the notfier system while loaded doesn't appear to help at the pre-load point.

I know it is typically difficult for me to answer a similar question, is there anything I can do to help and not duplicate any effort?

drbyte:
I haven't tested it, and I have some reservations about being too aggressive with force-updating fields not added by Zen Cart (ie: those from plugins or external tools), but [the link @Zen4All posted](https://stackoverflow.com/questions/23385555/remove-all-zero-dates-from-mysql-database-across-all-tables) contains this MySQL possibility, where the only replacement to insert would be the database-name. (Er, well, if we want to isolate to prefixed tablenames, we might have to add some other matching rules too)
... but it's worth exploring:
```sql
select
concat('UPDATE ',TABLE_NAME,' SET ',COLUMN_NAME,'=NULL WHERE ',COLUMN_NAME,'=0;')
from information_schema.COLUMNS 
where TABLE_SCHEMA = 'DATABASE_NAME' and DATA_TYPE in ('datetime', 'date', 'time');
```

drbyte:
Closing this. Gonna do it differently, as part of every version-upgrade step in zc_install.

PR #2070, opened 2019-01-11
Admin sanitizer htmlspecialchars parameters

### LINKED ISSUE #2064: Should additional parameters for htmlspecialchars be included in admin sanitizer?
https://github.com/zencart/zencart/blob/f9752c8f01d031756c0bb31de0c43a35ef716531/admin/includes/classes/AdminRequestSanitizer.php#L782

Was looking over the strict sanitizer values code to support identification of what characters are sanitized and noticed that at the above file location, `htmlspecialchars($string)` is used "alone" and without the added parameters seen throughout much of the catalog side where it typically is used as: `htmlspecialchars($string, ENT_COMPAT, CHARSET, TRUE)`.  Was wondering if that should be updated?

The php manual identifies that in absence of providing the above parameters, that effectively for ZC 1.5.6 having a published lower band PHP of 5.6 that the same is executed, although the CHARSET defaults to UTF-8 regardless of the characterset setting for the site (i.e. still using ISO-8859-1 or other).

Meanwhile at: https://github.com/zencart/zencart/blob/f9752c8f01d031756c0bb31de0c43a35ef716531/admin/includes/classes/AdminRequestSanitizer.php#L456 the encoding is forced to `'utf-8'` and at: https://github.com/zencart/zencart/blob/f9752c8f01d031756c0bb31de0c43a35ef716531/admin/includes/classes/AdminRequestSanitizer.php#L491 there also is no additional parameters for function `filterSanitizeEmailAudience` which supposedly again would default to `'UTF-8'`.

mc12345678:
Fixes #2064.

### DIFF
--- a/admin/includes/auto_loaders/config.core.php
+++ b/admin/includes/auto_loaders/config.core.php
@@ -1,31 +1,36 @@
 <?php
 /**
  * @package admin
- * @copyright Copyright 2003-2018 Zen Cart Development Team
+ * @copyright Copyright 2003-2019 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: lat9 Mon Oct 22 13:19:39 2018 -0400 Modified in v1.5.6 $
+ * @version $Id: mc12345678 Sun Jan 20 01:27:39 2019 -0400 Modified in v1.5.6b $
  */
 if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
 /**
  * autoloader array for admin application_top.php
+ * Where DIR_WS_CLASSES is used alone in commented text, the file loads relative
+ *   to the admin side folder.
 **/
 /**
- * require('includes/version.php');
- * require(DIR_WS_CLASSES . 'class.base.php');
- * require(DIR_WS_CLASSES . 'class.notifier.php');
- * $zco_notifier = new notifier()'
- * require(DIR_WS_CLASSES . 'sniffer.php');
+ * require(DIR_FS_CATALOG . DIR_WS_INCLUDES . 'version.php');
+ * require(DIR_FS_CATALOG . DIR_WS_CLASSES . 'class.base.php');
+ * require(DIR_FS_CATALOG . DIR_WS_CLASSES . 'class.notifier.php');
+ * $zco_notifier = new notifier();
+ * require(DIR_FS_CATALOG . DIR_WS_CLASSES . 'sniffer.php');
  * require(DIR_WS_CLASSES . 'logger.php');
- * require(DIR_WS_CLASSES . 'shopping_cart.php');
- * require(DIR_WS_CLASSES . 'products.php');
+ * require(DIR_FS_CATALOG . DIR_WS_CLASSES . 'shopping_cart.php');
+ * require(DIR_FS_CATALOG . DIR_WS_CLASSES . 'products.php');
  * require(DIR_WS_CLASSES . 'table_block.php');
  * require(DIR_WS_CLASSES . 'box.php');
  * require(DIR_WS_CLASSES . 'message_stack.php');
  * require(DIR_WS_CLASSES . 'split_page_results.php');
  * require(DIR_WS_CLASSES . 'object_info.php');
- * require(DIR_WS_CLASSES . 'class.phpmailer.php');
- * require(DIR_WS_CLASSES . 'upload.php');
+ * require(DIR_FS_CATALOG . DIR_WS_CLASSES . 'class.phpmailer.php');
+ * require(DIR_FS_CATALOG . DIR_WS_CLASSES . 'upload.php');
+ * require(DIR_FS_CATALOG . DIR_WS_CLASSES . 'class.zcPassword.php');
+ * zcPassword = new zcPassword();
+ * require(DIR_WS_CLASSES . VersionServer.php');
  */
   $autoLoadConfig[0][] = array('autoType'=>'require',
                                'loadFile'=> DIR_FS_CATALOG . DIR_WS_INCLUDES .  'version.php');
@@ -94,14 +99,11 @@ if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
  */
   $autoLoadConfig[20][] = array('autoType'=>'init_script',
                                 'loadFile'=> 'init_db_config_read.php');
-  $autoLoadConfig[20][] = array('autoType'=>'init_script',
-                                'loadFile'=> 'init_sanitize.php');
 /**
  * Breakpoint 30.
  *
  * require('includes/init_includes/init_gzip.php');
  * $sniffer = new sniffer();
-
  *
  */
   $autoLoadConfig[30][] = array('autoType'=>'init_script',
@@ -147,14 +149,21 @@ if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
   $autoLoadConfig[60][] = array('autoType'=>'init_script',
                                 'loadFile'=> 'init_sessions.php');
 /**
+ * Breakpoint 65.
+ *
+ * require('includes/init_includes/init_languages.php');
+ * Expecting nothing loaded before init_sanitize to require $_POST/$_GET sanitization.
+ */
+  $autoLoadConfig[65][] = array('autoType'=>'init_script',
+                                'loadFile'=> 'init_languages.php');
+/**
  * Breakpoint 70.
  *
- * require(DIR_WS_FUNCTIONS . 'admin_access.php');
- * require('includes/init_includes/init_languages.php');
+ * require('includes/init_includes/init_sanitize.php');
  *
  */
   $autoLoadConfig[70][] = array('autoType'=>'init_script',
-                                'loadFile'=> 'init_languages.php');
+                                'loadFile'=> 'init_sanitize.php');
 /**
  * Breakpoint 80.
  *
@@ -167,8 +176,7 @@ if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
  * Breakpoint 90.
  *
  * $zc_products = new products();
- * require(DIRWS_FUNCTIONS . 'localization.php');
- * require(DIRWS_FUNCTIONS . 'validations.php');
+ * require(DIR_WS_FUNCTIONS . 'localization.php');
  *
  */
   $autoLoadConfig[90][] = array('autoType'=>'classInstantiate',
@@ -183,7 +191,6 @@ if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
  *
  */
   $autoLoadConfig[100][] = array('autoType'=>'classInstantiate',
-                                'classPath'=>DIR_WS_CLASSES,
                                  'className'=>'messageStack',
                                  'objectName'=>'messageStack');
 /**
--- a/admin/includes/classes/AdminRequestSanitizer.php
+++ b/admin/includes/classes/AdminRequestSanitizer.php
@@ -453,7 +453,7 @@ class AdminRequestSanitizer extends base
             $this->debugMessages[] = 'PROCESSING META_TAGS == ' . $this->arrayName;
             foreach ($_POST[$parameterName] as $pKey => $pValue) {
                 $currentArrayName = $this->setCurrentArrayName($pKey);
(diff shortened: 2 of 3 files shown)


mc12345678:
Problem is (as I believe wilt found earlier) `CHARSET` isn't defined until the language is loaded, which (without digging in) appears to be after the sanitizer is accessed/needs the htmlspecialchars sanitizer.  So, smidge of a chicken/egg thing to address.

Perhaps an "internal" variable and if by chance the constant is set that the variable equals the constant? At least it would make its use available for search?

mc12345678:
Yes, would be easier.  Concerned though of allowing as much opportunity to use unsanitized data as that would do. There is currently only one `$_GET` in all of the default language files as shown by the ZC 1.5.6 search of the language files for `$_GET` or `$_POST` in ZC 1.5.6; however, there is also at least one other in the assignment of the language to the session. On the admin side as you most likely know at least before the actual language file is loaded there is: admin/includes/languages/YOUR_LANGUAGE/CURRENT_PAGE and admin/includes/languages/YOUR_LANGUAGE/extra_definitions/*.php

What would seem to make the most sense would be a separation of the language file so that it would have the setup portion and the actual language portion with one loaded before the other allowing the sanitization the applicable setup but not have the same potential of allowing posted data to be assigned without sanitizing first.

But, like I said moving the init_sanitize to after the language loading would be easy and involves a little chicken/egg, but seems like it would be late in the process. Of course, that is all to be done to try to load a single variable assignment: `define('CHARSET', 'utf-8');` early enough to support other operations.

lat9:
I didn't say that the move would be _easier_ &mdash; I said it would be _better_, since the store's defined `CHARSET` would be used in the sanitization.

From what I've seen in the admin's `init_languages.php`, its primary requirement is that it be loaded post-session initialization.  The `language.php` class does its own job of sanitization, not allowing an undefined language to be loaded.

Perhaps the solution lies, as you've intimated, in splitting the load of the primary (e.g. /includes/languages/english.php) language file from all others.  Load the primary language file (which gets the `CHARSET` defined), then load the sanitizer's initialization file, and finally load the remaining language files ... all on/around CP 70 where the language-files' initialization currently resides.

mc12345678:
> I didn't say that the move would be easier — I said it would be better, since the store's defined CHARSET would be used in the sanitization.
>
> From what I've seen in the admin's init_languages.php, its primary requirement is that it be loaded post-session initialization. The language.php class does its own job of sanitization, not allowing an undefined language to be loaded.
>
> Perhaps the solution lies, as you've intimated, in splitting the load of the primary (e.g. /includes/languages/english.php) language file from all others. Load the primary language file (which gets the CHARSET defined), then load the sanitizer's initialization file, and finally load the remaining language files ... all on/around CP 70 where the language-files' initialization currently resides.

Even in simply splitting the loading of `admin/includes/languages/english.php` (as currently populated) from other language files, it should load *after* other such language files so that they can provide an override as was done in issue #1930 and pull issue #1931.  This would likely entail splitting the file so that specific definitions applicable to the language concept can be loaded in preparation for work to be done with data provided. Then the content of the presented information shown to the user.  All of that so that the marriage between database and display can be sustained as part of the sanitization. I believe it also more closely approximates PSR style implementation.

This would be a significant change for any admin language developer/provider.

While perhaps the contents/lines of this "first" file could be modified (not so sure that defines such as `ICON_METATAGS_ON`, `ICON_METATAGS_OFF`, `TEXT_LEGEND_META_TAGS`, `TEXT_INFO_META_TAGS_USAGE`, or possibly even `TEXT_ADMIN_TAB_PREFIX` need to load as part of the language initialization in order to support subsequent consistencies on the formatting of the language), my initial thought of the first to load language file would be similar to:
````php
<?php
/**
 * @package admin
 * @copyright Copyright 2003-2019 Zen Cart Development Team
 * @copyright Portions Copyright 2003 osCommerce
 * @version $Id: DrByte 2019 Jan 04 Modified in v1.5.6a $
 */
if (!defined('IS_ADMIN_FLAG'))
{
  die('Illegal Access');
}
define('CONNECTION_TYPE_UNKNOWN', '\'%s\' is not a valid connection type for generating URLs' . PHP_EOL . '%s' . PHP_EOL);
// added defines for header alt and text
define('HEADER_ALT_TEXT', 'Admin Powered by Zen Cart :: The Art of E-Commerce');
define('HEADER_LOGO_WIDTH', '200px');
define('HEADER_LOGO_HEIGHT', '70px');
define('HEADER_LOGO_IMAGE', 'logo.gif');
// look in your $PATH_LOCALE/locale directory for available locales..
$locales = array('en_US', 'en_US.utf8', 'en', 'English_United States.1252');
@setlocale(LC_TIME, $locales);
define('DATE_FORMAT_SHORT', '%m/%d/%Y');  // this is used for strftime()
define('DATE_FORMAT_LONG', '%A %d %B, %Y'); // this is used for strftime()
define('DATE_FORMAT', 'm/d/Y'); // this is used for date()
define('PHP_DATE_TIME_FORMAT', 'm/d/Y H:i:s'); // this is used for date()
define('DATE_TIME_FORMAT', DATE_FORMAT_SHORT . ' %H:%M:%S');
// for now both defines are needed until Spiffy is completely removed.
define('DATE_FORMAT_SPIFFYCAL', 'MM/dd/yyyy');  //Use only 'dd', 'MM' and 'yyyy' here in any order
define('DATE_FORMAT_DATE_PICKER', 'yy-mm-dd');  //Use only 'dd', 'mm' and 'yy' here in any order
define('ADMIN_NAV_DATE_TIME_FORMAT', '%A %d %b %Y %X'); // this is used for strftime()
////
// Return date in raw format
// $date should be in format mm/dd/yyyy
// raw date is in format YYYYMMDD, or DDMMYYYY
function zen_date_raw($date, $reverse = false) {
  if ($reverse) {
    return substr($date, 3, 2) . substr($date, 0, 2) . substr($date, 6, 4);
  } else {
    return substr($date, 6, 4) . substr($date, 0, 2) . substr($date, 3, 2);
  }
}
// // include template specific meta tags defines
//   if (file_exists(DIR_FS_CATALOG_LANGUAGES . $_SESSION['language'] . '/' . $template_dir . '/meta_tags.php')) {
//     $template_dir_select = $template_dir . '/';
//   } else {
//     $template_dir_select = '';
//   }
//   require(DIR_FS_CATALOG_LANGUAGES . $_SESSION['language'] . '/' . $template_dir_select . 'meta_tags.php');
// used for prefix to browser tabs in admin pages
define('TEXT_ADMIN_TAB_PREFIX', 'Admin ');
// if you have multiple stores and want the Store Name to be part of the admin title (ie: for browser tabs), swap this line with the one above
//define('TEXT_ADMIN_TAB_PREFIX', 'Admin ' . STORE_NAME);
// meta tags
define('ICON_METATAGS_ON', 'Meta Tags Defined');
define('ICON_METATAGS_OFF', 'Meta Tags Undefined');
define('TEXT_LEGEND_META_TAGS', 'Meta Tags Defined:');
define('TEXT_INFO_META_TAGS_USAGE', '<strong>NOTE:</strong> The Site/Tagline is your defined definition for your site in the meta_tags.php file.');
// Global entries for the <html> tag
define('HTML_PARAMS','dir="ltr" lang="en"');
// charset for web pages and emails
define('CHARSET', 'utf-8');
````

I was thinking that the first five defines may need to remain to possibly support the page (browser tab) having some level of content by which one might determine where/when a problem occurred (wishful thinking?!).

With the remainder of the file (including items moved from above) to be loaded (as a separate file) after the other override language files have been loaded. Again refer to issue #1930 and pull issue #1931.

Perhaps the "main" file as presented above would remain as the languaged file: `admin/includes/languages/english.php` with the subsequent content to be in an additional file.  The additional file is still to be identified as to where/when it would be loaded.  This way, if someone identifies that they are unable to override a primary language's define, it would be easier to identify that the language file layout had not yet been adopted and that they could split the remainder of the existing file to the second file.

Propose the second file to have a name including the camelcase of `defines` such that the base structure would be:
`admin/includes/languages/english.php`
`admin/includes/languages/englishDefines.php`
with no other such current use of the camelcase of `defines` proposed.

The header (comment area) of each file would provide some level of description of the expectation of content within the file and suggestion of which file should contain which content for the desired result. (e.g. any define in `english.php` would not be overrideable in other language files, where `englishDefines.php` would be loaded later to where `current_page` and/or `extra_definitions` could provide an alternate define.)

The "hard" part of incorporating/accepting the word `defines` into the filename is that there obviously are some defines needed to support the goal, but the idea proposed is to minimize the heartache of transitioning to this from version to version and to sustain override capability. It already would involve some level of effort for those refusing to fully upgrade but continue to cherry pick newer code.

drbyte:
Note that the v2 language defines split off locale-specific stuff into a `locale.php` file, which can be processed earlier.

lat9:
> 
> 
> I'm thinking loading it at point 75 is 95% safe.
> 
> Is this all moot after PHP 5.6 support is dropped?

Not moot, since the primary issue (sanitizer using CHARSET prior to definition) isn't a PHP version-dependent issue.

drbyte:
> > Is this all moot after PHP 5.6 support is dropped?

> Not moot, since the primary issue (sanitizer using CHARSET prior to definition) isn't a PHP version-dependent issue.

The addition of CHARSET came from https://github.com/zencart/zencart/issues/2064 where it was mentioned that older PHP versions set a different default.

My comment above was in relation to: if we're no longer using/supporting those older PHP versions which used a conflicting default charset, isn't this chicken/egg issue no longer an issue, and therefore this PR is moot?

drbyte:
Maybe we should just leave this alone then?

mc12345678:
I would say no on the leave it alone for the mere fact that there is currently a hard coded `utf-8` as it has been.

Seems to me that the split of the language file to the development version of a locale.php file would be the best fit, with the alternate being a later load point where the site could still have the potential problem that caused the sanitization process to be developed. That problem being possible based on development of software by others outside of the core Zen Code.

lat9:
As @mc12345678 found, the current sanitizer class uses a combination of PHP version-specific defaults or a hard-coded `'utf-8'` ... not-good things.  The considerations for the correction should take into place, IMO, preexisting installations that currently _appear_ to work.  My take is that going to the split of `locale.php` from the base language file is a bit too extreme for a 'point' release of zc1.5.6.

While many of the plugin sanitizers (e.g. Edit Orders) load their specific sanitizations via an `extra_datafile` include, the use of `htmlspecialchars` by the sanitizer class doesn't come into play until `init_sanitize.php` invokes the `runSanitizers`  class-method.

Given those conditions, I suggest a two-pronged approach:

1. Update the `/admin/includes/classes/AdminRequestSanitizer.php` module to use a fully-formed call to each use of `htmlspecialchars`, using the currently-defined `CHARSET` as the encoding basis.
2. Modify `/admin/includes/auto_loaders/config.core.php` as follows:
    1. Move `init_languages.php` from CP 70 to CP 71, moving it 'closer' to the completion of `init_sessions.php` which is its prerequisite.
    2. Move `init_sanitize.php` from CP 20 to CP 72, just after its `init_languages.php` prerequisite.

That grouping and the splitting of the loaded check-points will allow plugins to insert functionality between each base initialization file while minimizing the 'distance' to keep things kosher.

mc12345678:
1. How is moving `init_languages.php` from CP 70 to CP 71 moving it 'closer' to `init_sessions.php` CP 60?
2. And why load `init_sanitize.php` so close to `init_languages.php` that if someone wanted to properly load something between the two but had issues with alphabetization or other expected sequencing that they would unnecessarily have to modify the `config.core.php` file which pretty much couldn't be any more core to Zen Cart operation.

lat9:
1. 60/70 ... typos.
2. That's why I suggested (typo removed) 60 (init_sessions), 61 (init_languages), 62 (init_sanitize).  Can you really conceive a case where multiple plugins have to have their additions ordered in a specific sequence and also specifically between the language- and sanitization-initializations?

mc12345678:
> Can you really conceive a case where multiple plugins have to have their additions ordered in a specific sequence and also specifically between the language- and sanitization-initializations?

Doesn't really answer the question of why lock out the ability to load something in between and/or require modification to insert? It doesn't really matter what is currently conceived as much as it does in planning/preparing for future use/application. 

For example, I don't personally have a need for the changes made in this pull, but I have a feeling of responsibility.  I came across something that was incorrect mostly by it not matching the plans expressed by the core developers and the possibility it could affect the database for those using an alternate `CHARSET`.  I proposed the change.  How someone uses it, works, or works with it, is up to them.

The thing is that there is plenty of space in between, it just doesn't make sense to me why to force consecutive values when there is no expressed need for one to immediately follow another. I.e. In the admin isn't it suggested to add items to a sorted list giving them a sort order that jumps every 10 numbers or so?

PR #2075, opened 2019-01-14
v156 correct display of attribute price for priced-by-attribute

mc12345678:
Identified in https://www.zen-cart.com/showthread.php?224923-Attribute-pricing-display-all-fouled-up, if a product is `priced by attribute`, the default resulting price (of each attribute) included the `products_price`.  Removing the `products_price` from the attribute price response restores the previous default functionality; however, doing that for all situations causes the result of the name of this function (`zen_get_attributes_price_final`) to not be possible.  Added a parameter that defaults to the historical ZC response of not including the `products_price` in the final result.  This solution resolved the observed condition and may maintain the overall desired resolution made by pull #1404.  If not, then the subtraction/addition should be that of zen_get_discount_calc without the `options_values_price` factored into that price difference.

### DIFF
--- a/admin/includes/functions/functions_prices.php
+++ b/admin/includes/functions/functions_prices.php
@@ -1010,8 +1010,8 @@ If a special exist * 10+9
 
     $attributes_price_final = 0;
 
-    if ($pre_selected == '' or $attribute != $pre_selected->fields["products_attributes_id"]) {
-      $pre_selected = $db->Execute("select pa.* from " . TABLE_PRODUCTS_ATTRIBUTES . " pa where pa.products_attributes_id= '" . (int)$attribute . "'");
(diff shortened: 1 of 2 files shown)


mc12345678:
Sorry Ajeh for the additional modification, but at least on the store side the latest commit appears to correct the issue most recently discussed (in depth) in: https://www.zen-cart.com/showthread.php?224923-Attribute-pricing-display-all-fouled-up

As identified in the latest commit, it also appears to resolve the following (having tested each of them):
- https://www.zen-cart.com/showthread.php?219366-SaleMaker-only-discounts-base-price-not-attributes
- https://www.zen-cart.com/showthread.php?221940-attributes-price-factor-special-price-incorrect-totals

The second thread has two related issues described in it.

Still need to update the admin side to reflect the changes made in the catalog side, but the store side was a little more important.

PR #2077, opened 2019-01-15
Change keepalive loginURL to camefrom v index on timeout

proseLA:
i have tested and verified this works...  it creates the camefrom url with 1 `?` and then uses the `&` for seperating parameters.  it also trims the last `&`.

as @Zen4All previously commented on, the `<?=` directive is NOT affected by the short tag being turned on or off since php 5.4.  so i opted to not change all of that code and leave it as is....  which seems clean to me.

in addition, i did move the `'` for the camefrom var to outside of the `$camefrom` as suggested by @drbyte and being consistent with the existing code.


drbyte:
Thanks

PR #2078, opened 2019-01-15
addresses error when shipping module comes back w empty response

proseLA:
testing `empty()` v `!is_array()`.  empty seemed to work fine.

i would love to get all files into unix line feeds as opposed to windows....  but it seems so easy for me to screw things up in git.  so apologies about the line change on the time stamp.  i blame my IDE!

### DIFF
--- a/includes/classes/shipping.php
+++ b/includes/classes/shipping.php
@@ -183,6 +183,9 @@ class shipping extends base {
         $class = substr($value, 0, strrpos($value, '.'));
         if (isset($GLOBALS[$class]) && is_object($GLOBALS[$class]) && $GLOBALS[$class]->enabled) {
           $quotes = $GLOBALS[$class]->quotes;
+          if (empty($quotes['methods'])) {
+            return;
+          }
           $size = sizeof($quotes['methods']);
           for ($i=0; $i<$size; $i++) {
             if (isset($quotes['methods'][$i]['cost'])){


drbyte:
Thanks!

mc12345678:
There's no need to test if `$rates` is `empty` as `$rates` was already identified as an `array`. If at the end of whatever processing has previously occurred, no additional rate has been added, then `sizeof($rates)` will be `zero` and `zero` is not `< zero`, therefore nothing within the second `for` loop would be processed.  So don't understand why anything further needs to be tested if there are no results to be added to `$rates`.

mc12345678:
> adding the additional conditional is the answer for what this PR was trying to accomplish.

That makes it harder to read.  The preferred method (as expressed in other pulls) is to operate like you had with:
````php
if (empty($quotes['methods'])) {
  continue;
}
````
Accomplishes the same task but easier to read and understand.

So, since you seem to want to point out issues by saying how wrong something is, should I do the same?

proseLA:
> So, since you seem to want to point out issues by saying how wrong something is, should I do the same?

@mc12345678 let's recount.

@drByte suggested a PR based on a discussion on the forum.  i went ahead and did that PR, as suggested by him.  

you then pointed out a problem with that PR. i spent some time looking at what you stated, and acknowledged the problem you pointed out.  however, i also stated that your ALREADY merged solution was wrong.  that code would NOT come up with the cheapest quote in all situations.  you found a problem in the code, and your solution did not address the problem.   (perhaps you could have looked at the code closer?)

i then agreed with you with regards to you comment on the `$rates` array.

i came up with a solution to add another conditional; you came up with `continue`.  i said `continue` would work.

now you are asking some sort of rhetorical question?  about style?

previous PRs have talked about the arrow anti-pattern.  which i tried to avoid. 

allow me to be very clear: my clients do NOT make use of the cheapest function.  i am attempting to contribute to make the product better.  i could care less which solution gets implemented.  they both work.  i do care if the code is wrong.  sorry if i did not express that in a manner amenable to you.

now, let's look at one quote you made:

> I evaluated the situation from what the departing action was, instead of what was happening. That was **likely** in error.

there is NO likely in your statement.  that was in error.  do you ever have a full-on mea culpa?  or are you never fully wrong?  (also rhetorical, i'm not really looking for an answer.)

perhaps prior to typing at the keyboard, you can examine the situation just a little bit better.  

feel free to have the last word.  i'm done.

PR #2082, opened 2019-01-15
GV text

scottcwilson:


drbyte:
Argh. Is this really necessary?

Embedded links make language files more brittle.

scottcwilson:
The problem is if you don't have it, how do they know where to get it?    Not everyone runs the shopping cart sidebox.

PR #2084, opened 2019-01-17
Do not show home address as delivery address for pickup orders

scottcwilson:
Currently, the address shown as the delivery address for an order to be shipped using Store Pickup is the customer's default address.  This is misleading.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -162,7 +162,7 @@ class order extends base {
                             'format_id' => $order->fields['delivery_address_format_id']);
 
     if (($order->fields['shipping_module_code'] == 'storepickup') || 
-        empty($this->delivery['name']) && empty($this->delivery['street_address'])) {
+        (empty($this->delivery['name']) && empty($this->delivery['street_address']))) {
       $this->delivery = false;
     }
 
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -161,7 +161,8 @@ class order extends base {
                             'country' => $order->fields['delivery_country'],
                             'format_id' => $order->fields['delivery_address_format_id']);
 
-    if (empty($this->delivery['name']) && empty($this->delivery['street_address'])) {
+    if (($order->fields['shipping_module_code'] == 'storepickup') || 
+        empty($this->delivery['name']) && empty($this->delivery['street_address'])) {
       $this->delivery = false;
     }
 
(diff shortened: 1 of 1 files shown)


includes/classes/order.php
```diff
@@ -161,7 +161,8 @@ function query($order_id) {
                             'country' => $order->fields['delivery_country'],
                             'format_id' => $order->fields['delivery_address_format_id']);
 
-    if (empty($this->delivery['name']) && empty($this->delivery['street_address'])) {
+    if (($order->fields['shipping_module_code'] == 'storepickup') || 
+        empty($this->delivery['name']) && empty($this->delivery['street_address'])) {
```
drbyte: Is this nesting of parentheses correct?  Shouldn't the extra opening `(` be moved from 164 to beginning of 165?

scottcwilson: No - and binds more tightly than or.  But I can add parens if you find this confusing to read.

drbyte: So then the first extra parentheses is redundant, right?

scottcwilson: You're right.

scottcwilson: Yes, but I think it's a bit easier to read how it is now.

drbyte: Fair. I'm always trying to avoid unnecessary visual debt, including unnecessary parentheses.

drbyte:
How does this affect things when the storepickup module is defined with multiple pickup locations? Wouldn't it be desirable to have the pickup location included? And wouldn't that pickup location make sense to be displayed as the delivery address?

scottcwilson:
I considered that but the issue is that the pickup locations are entered in plain, unvalidated text so they're not amenable to printing via `zen_address_label`.  My main concern was not making the order confirmation (and subsequent display) suggest that the order would be delivered to the person's home address, which is how it looks now.

drbyte:
Removing confusion is valid.

Doesn't removing the pickup-location address make it even more likely that one would assume it's being delivered to their home, since that'd be the only address listed?

How about simply changing the title of "Delivery Address" to "Pickup Location"?

The longer more involved "fix" for the unvalidated address text is an improved multi-location store-pickup module. Unfortunately that really needs its own added database table for store addresses .... which seems dumb if one has only one location.  Chicken/Egg.

scottcwilson:
This change does not remove the pickup-location address.  It removes the customer's home address from being shown as the delivery location - which is misleading and incorrect.

lat9:
The underlying issue, IMO, is that the `storepickup` processing _should_ replace the order's delivery address with the pickup address selected by the customer.

scottcwilson:
This is a much larger fix; see Chris' comments above.

torvista:
Something else to bear in mind is that is it ever more common for customers to choose a delivery service and then also a pickup location (shop, newsagent, hairdresser...) that has an arrangement with that particular courier. People don't like to wait in for non-arriving couriers and hopefully all these insane **individual** deliveries will be banned anyway due to environmental concerns.
So, the use of the phrase "pickup location" will be relevant for all delivery methods in the near future, along with facilitating api methods to get such pickup data from the couriers during ordering...

Issue #2057, opened 2019-01-08
Zen Cart V2 Admin Roadmap Overview


zcwilt:
Zen Cart V2 Admin Roadmap Overview
==================================

While the github [V2 Public Road Map](https://github.com/zencart/zencart/projects/1) project board helps, it's more for managing issues/pr's on a day to day basis.

So for those interested, we thought it would be useful to give a broader overview of the higher level issues that need to be addressed to get V2 to a release candidate.

To start off, we are just going to address the Admin for now.
There will be a similar overview done for Catalog soon.

For some idea of where we are already at you can take a look at the `docs/whatsnew_2.0.html` file within the release.

We also suggest [downloading the V2 branch](https://github.com/zencart/zencart/archive/develop.zip) to look at how the new code is structured.
 
Admin Styling/Templates/JS etc.
-------------------------

Currently the v2 admin is using bootstrap 3 and adminlte, although the reliance on adminlte css is fairly minimal. 

The plan is to convert to bootstrap 4, and some tests we have done on this show that it should be fairly easy with minimal changes required to current admin templates to embrace this.

Reducing dependancies: 
Currently we use around 42 lines of code to define the css and js required for an admin view/page, and in  the main do that for every view regardless of if those dependancies are required or not. Now it may be for admin that this isn't really a problem. Admin is less heavily used than the catalog and the lag caused by loading some/many resources may not be a problem for most people.

New V2 Admin Controllers
------------------------

[Admin Controllers Docs](https://docs.zen-cart.com/Developer_Documentation/v2/code_docs/admin_action_classes/introduction)

[Lead Controller Docs](https://docs.zen-cart.com/Developer_Documentation/v2/code_docs/admin_lead_pages/introduction)

Note. The docs are somewhat out of date, but provide an idea of the usefulness of the new infrastructure.

While a fair number of pages have been converted to use the new admin controllers, there are quite a number (mostly more complex) pages still left to convert.

It's also certain that there will need to be some core infrastructure enhancements/refactoring to allow these endpoints to be easily built.

As the output html is completely separate from logic, converting templates for some pages can be challenging (e.g.  the attributes controller).

Consideration needs to be given to revamping the UI for these types of pages, to make them more intuitive.

A list of those pages awaiting conversion will be added as an addendum to this issue.

Plugin Infrastructure
--------------------

The new Admin Infrastructure also brings benefits to plugin authors.
The main one being that plugins authored for v2 need never touch core code.
[...]

PR #2088, opened 2019-01-19
Initial Commit for plugin infrastructure

zcwilt:


lat9:
Is there going to be some documentation on how a plugin developer might use these interfaces?

scottcwilson:
Yes!

lat9:
Yes, there will be, or Yes, I agree?

scottcwilson:
Both!  The trouble is "Speedy" @zcwilt  codes faster than I can write documentation!  So you have to be patient.

drbyte:
It's a bit early to document this. The code is still in development, hence "Initial commit" in the title.

You'll also note that this PR includes some docs.

I envision us creating a few plugins as examples as well.

lat9:
Yes, I was looking for .md documents rather than .html ones.  I'll also note that for an infrastructure change of this magnitude, I would have anticipated that documentation be developed (and reviewed by the community) _first_, rather than as an after-the-fact.  It's a lot easier to review the written word than written code and then everyone (developers and users) would have a common reference point if/when different interpretations arise.

zcwilt:
I guess this is my fault. The plug-in infrastructure will eventually rely on lots of other changes to core code. An example would be how we load language definitions.  In order to develop that I needed the very basic structure that this PR gave. 

There are also a lot of other core infrastructure changes that plugins will use as well, and the intention is to do these as separate PR's and with those PR's to add to the dev docs.

Again as an example, hopefully I will have a PR shortly for language stuff and pretty much spent this morning writing developer docs for that, which hopefully we will get feedback on.

As those PR's come through the dev docs for plug-ins will be updated as well.

zcwilt:
e.g.

https://raw.githubusercontent.com/zcwilt/zencart/v2-new-language-structure/dev_docs/_build/html/_sources/languageloading.rst.txt

zcwilt:
I guess I should also note https://github.com/zencart/zencart/issues/2057

If you have some comment about how plug-ins etc should work, then you can do it there as well.

PR #2089, opened 2019-01-19
Fix GV_NAME define

scottcwilson:
See https://www.zen-cart.com/showthread.php?224966-GV_NAME-GV_NAMES-defined-too-late for an explanation.


proseLA:
3 of the files have closing php tags at the end of the file.  might as well remove them while u r here....

scottcwilson:
I'd rather not make changes like this in a dot release.  If there's a 1.5.7, we can do it there.

lat9:
You realize that migrating those constants out of the main language file could break existing admin-level plugins that expect those definitions to be preexisting, right?

scottcwilson:
Agreed.  Let's pull the file into english.php if it hasn't already been done.

PR #2102, opened 2019-01-20
creates logfile prior to die on error in zen_href_link

proseLA:
to better help debugging, creates log file using ZC logging function.

also removes global variables no longer used and left over when addressed ENABLE_SSL_ADMIN constant.  from commit 2cb2c07

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -10,8 +10,10 @@
 ////
 // The HTML href link wrapper function
   function zen_href_link($page = '', $parameters = '', $connection = 'SSL', $add_session_id = true) {
-    global $request_type, $session_started, $http_domain, $https_domain;
+    global $session_started;
     if ($page == '') {
+      trigger_error("zen_href_link($page, $parameters, $connection), unable to determine the page link.",
+            E_USER_ERROR);
       die('</td></tr></table></td></tr></table><br><br><font color="#ff0000"><b>Error!</b></font><br><br><b>Unable to determine the page link!<br><br>Function used:<br><br>zen_href_link(\'' . $page . '\', \'' . $parameters . '\', \'' . $connection . '\')</b>');
     }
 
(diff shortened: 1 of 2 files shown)


drbyte:
What's the use-case/situation that drove the need to trigger an error at this stage? 

ie: normal core code doesn't trigger this scenario. What code are you finding is passing an incorrect type, and therefore needs to halt execution?

(I realize that one wouldn't ever know this was happening to end-users if it wasn't reported by them, or logged. But ... how did you stumble on this? ie: are there specific common customizations that are triggering it? specific plugins that are sending bad parameters? specific version-upgrade incompatibilities causing this? etc etc ...)

proseLA:
we have a die statement in the code.  if someone were to encounter a die statement as indicated, why not provide BETTER logging to help solve this rare situation?  it only makes sense to me.

i ran into this situation, and then i had to start digging to find it.  with a log file, i would have saved time.

with regards to normal code not triggering it, sure i suppose it would be the case; HOWEVER the base ZC install already accounts for this error.  someone else must have run into it which is why it is here.  what does it matter how?  i ran into it as part of an upgrade to v156.  why not give me and other users the tools to more easily find the problem?

i don't buy the log fills up we run out of disk space argument on today's web with the hosting that people have.  and if so, that is on the store owner; there are plenty of threads in the forum about log directories getting loaded up because someone does not maintain their shop.  log files are there for a reason.  to point out errors.  this error to me is FAR better suited with a log file than a generic die statement without a backtrace.

see:
https://goo.gl/xw7BJR

lat9:
As @proseLA implied, I've never seen this during "normal" operation but I _have_ chased my tail during plugin development and/or customizations when a call to `zen_href_link` when awry.  

I also don't think that it's worthy of a switchable-type operation since `zen_href_link` is going to `die` (pun intended) regardless and the change is, essentially, a developer aid.

PR #2105, opened 2019-01-20
Correct for strict notifications when preview_info page URI entered

mc12345678:
If a preview_info page url was provided by itself, there is no
data posted and variables beyond it are accessed, but not set.

This sets those variables and provides control to prevent
notifications while in strict PHP mode.


mc12345678:
Discovered that this commit does not fully resolve the issue.  Preparing amplified changes to return the desired/necessary data. Starting with the `product` type first.  The associated variables that were added: $products_name, $products_description, and $products_url are all language dependent arrays and this commit did not reconstruct the array from the query.

PR #2106, opened 2019-01-21
Correct upload warning when no file uploaded

mc12345678:
Found a log being generated if attempt to access a page that would
normally be the one just after an upload, but where it was
landed at without posting data.  As a result, this patch
appears to resolve the logged messages and would assign a
value if one existed.

Example to reproduce would be something like accessing:
admin/product.php?cPath=4&pID=1&action=new_product_preview


drbyte:
Thanks for this. 
I'm thinking a better fix would be to do an early-return if both the $_FILES and $this->file are falsey/invalid.
Debating the best approach in a mid-point-release.

scottcwilson:
My thoughts on this issue are here: 
https://www.zen-cart.com/showthread.php?224327-Believed-bug-includes-classes-upload-php-line-84

mc12345678:
@scottcwilson, I had meant to try to reference that post, thank you for adding.

What is the effect of applying the proposed method of this PR (I understand that there has been an alternate solution considered, but am wondering what testing the proposed method discovered)?

Using the method from that post with the link identified in the OP of this pull resulted in defaulting to the double single quotes as `$products_image_tmp_name` was undefined and therefore not set.

drbyte:
Opted for early-return since there's no need to process globals.

PR #2110, opened 2019-01-22
New v2 language loading

zcwilt:
see docs in dev_docs folder
Still some work to do for template overrides
tests added, although need to test helper function


drbyte:
I'm hoping I don't find myself in a maze of twisty little passages ...

PR #2112, opened 2019-01-23
v156 Build language dependent variables when preview_info URI entered

mc12345678:
Previous correction applied initiated the concept of building the expected language dependent variables; however, did not incorporate the `language_id` into the variable and therefore an array was not created.  This provides the array needed to continue processing and show the information that existed about the product when it was accessed directly.

### DIFF
--- a/admin/includes/modules/document_general/preview_info.php
+++ b/admin/includes/modules/document_general/preview_info.php
@@ -32,9 +32,12 @@ if (zen_not_null($_POST)) {
 
   $pInfo = new objectInfo($product->fields);
   $products_image_name = $pInfo->products_image;
-  $products_name = $pInfo->products_name;
-  $products_description = $pInfo->products_description;
-  $products_url = $pInfo->products_url;
+
+  foreach($product as $prod) {
+    $products_name[$prod['language_id']] = $prod['products_name'];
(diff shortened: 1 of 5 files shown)


drbyte:
Should lines 15-17 also be treated as an array?
If so, then I'm inclined to consider refactoring to remove these arrays altogether and just use the $pInfo object.

But then again, this is not in the midst of a major release cycle, so none of these changes should be "sweeping" insomuch as potentially breaking plugins.

mc12345678:
I forgot to mention what was seen using the previous pull. Because the returned data from the query was a string, when the later processing attempted to use the first part of the array (haven't made any changes to the language_id), on screen a single character was placed for each of the applicable fields. It dawned on me I think because of helping someone with their code modifications to try to access a string as an array, that I thought back to this issue to try to eliminate the strict notifications. It would seem that in removing the message associated with strict processing, that an expected result was not in mind. When I thought about it a moment more I realized that I should have seen whatever data already existed for the item.

That's when I tested my original thought of trying to use $pInfo and found the two scenarios described above and opted for just using the query results.

The resulting array is processed a few lines down from there in the if/else section, so the only state change was in $product. Perhaps at the end of the foreach, $product should be reset so it would remain in the condition "expected", though there does not appear to be any further use of it in the base code.

PR #2114, opened 2019-01-24
Return the company name to address layouts

mc12345678:
Identified in https://www.zen-cart.com/showthread.php?224980-Customer-company-name-not-showing-up-on-Edit-Invoice-or-Packing-Slip-pages
the company name was not appearing wherever zen_address_format
was used.  This fixes that issue.

### DIFF
--- a/includes/functions/functions_customers.php
+++ b/includes/functions/functions_customers.php
@@ -83,8 +83,8 @@ function zen_address_format($address_format_id = 1, $incoming = array(), $html =
     // do the substitutions
     $address_out = str_replace(array_keys($address), array_values($address), $fmt);
 
-    if (ACCOUNT_COMPANY == 'true' && !empty($address['company']) && false === strpos($fmt, '$company')) {
-        $address_out = $address['company'] . $address['cr'] . $address_out;
+    if (ACCOUNT_COMPANY == 'true' && !empty($address['$company']) && false === strpos($fmt, '$company')) {
+        $address_out = $address['$company'] . $address['$cr'] . $address_out;
     }
 
     // -----


mc12345678:
Yup, did test.  Not sure what you have for line 85, but it is a blank line in the pull request.  

Yes, strict notifications came in on the variables of lines 99 ($firstname) through 108 ($hr) where:
````php
    $GLOBALS['zco_notifier']->notify(
        'NOTIFY_END_ZEN_ADDRESS_FORMAT',
        array(
            'format' => $fmt,
            'address' => $address,
            'firstname' => $firstname,
            'lastname' => $lastname,
            'street' => $street,
            'suburb' => $suburb,
            'city' => $city,
            'state' => $state,
            'country' => $country,
            'postcode' => $postcode,
            'cr' => $cr,
            'hr' => $hr,
        ),
        $address_out
    );
````
begins at line 94.  If not mistaken all of the passed variables that are reported actually are assigned within the `$address` variable which in this situation is not mutable (as being the first parameter of a notify), but `$address_out` is mutable.

The lines above the modification, look like this:
````php
    // convert into $-prefixed keys
    foreach ($address as $key => $value) {
        $address['$' . $key] = $value;
        unset($address[$key]);
    }
````
This takes the key of the `$address` array and prepends it with a dollarsign: `$address['company']` becomes `$address['$company']`.  Thus, when executing this portion of the code:
````php
    // do the substitutions
    $address_out = str_replace(array_keys($address), array_values($address), $fmt);
    if (ACCOUNT_COMPANY == 'true' && !empty($address['$company']) && false === strpos($fmt, '$company')) {
        $address_out = $address['$company'] . $address['$cr'] . $address_out;
    }
````
The first line goes through and replaces a key of the `$address` array with the value of that key.
The if loop, checks to be sure that the company name should be included using a "default for everyone" format.  (Is it expected to be a part of the address, is there a company name to be included, and did the address format not have the company name in it already.

Then instead of a substitution, the specific value of the specific keys are prepended.  Because of the above loop, `$address['$cr']` is defined and set, along with the other keys that were prepended with a dollar sign $. 

Note that `$address['$company']` is different than `$address[$company]`.  The second example would cause a problem as there is no key that is defined to have the value of the variable `$company` AND `$company` is not defined.

It does look like the notifier probably should be touched up as now the field `'address'` will be an an array which may not be expected by software previously written to use this notifier.

mc12345678:
Here is the `print_r($address, true)` of demo data just before the check for company which helped lead to the proposed modification:
````php
[$statecomma] => California,
[$firstname] => Bill Smith
[$lastname] =>
[$postcode] => 12345
[$streets] => 123 Any Avenue
[$company] => JustaDemo
[$country] => United States
[$suburb] =>
[$street] => 123 Any Avenue
[$state] => California
[$city] => Here
[$zip] => 12345
[$cr] => <br>
[$CR] => <br>
[$hr] => <hr>
[$HR] => <hr>
````

mc12345678:
Proposal for the notifier would be:
````php
$GLOBALS['zco_notifier']->notify(
        'NOTIFY_END_ZEN_ADDRESS_FORMAT',
        array(
            'format' => $fmt,
            'address' => $address,
            'firstname' => $address['$firstname'],
            'lastname' => $address['$lastname'],
            'street' => $address['$street'],
            'suburb' => $address['$suburb'],
            'city' => $address['$city'],
            'state' => $address['$state'],
            'country' => $address['$country'],
            'postcode' => $address['$postcode'],
            'cr' => $address['$cr'],
            'hr' => $address['$hr'],
        ),
        $address_out
    );
````
$address['company'] 
    

and add on the following array assignments that are/were not directly captured above. (available still through the `$address` array:
````php
    'zip' => $address['$zip'],
    'streets' => $address['$streets'],
    'statecomma' => $address['$statecomma'],
````

Note that 'zip' would be expected to have the same value as 'postcode' but could be programmed to be different so figured should be passed as well.

PR #2117, opened 2019-01-25
Correct zen_address_format notification

lat9:
As originally pushed for #2046 and modified for #2114.


mc12345678:
Why not pass the `$address` array and provided parameters instead of targeting each of the associated fields that may/may not be used/needed? Ie. if something were added/modified in the function or a `$fmt` used a variable that is defined in the function but not passed, then the notifier may need to be modified if not at least evaluated for modification.  

Thinking of potential uses where the other parameters passed to the function (besides `$incoming`) might be of value. For example whether to use html or not to support determination of appropriate alternate data for creating `$address_out`? Similar with `$address_format_id` to be able to apply different rules based on that value? 
Etc...

lat9:
Because the current use, e.g. Vat4EU, presumes that the array elements are present without the leading `$` characters.

mc12345678:
Okay, then to ask the question another way, why not pass the `$address` array and provided parameters in addition to targeting each of the associated fields?

lat9:
Because it's redundant.

mc12345678:
And yet would support the current use and would offer a way for other code to make use of this notifier in the core code without having to request a change to it or modify it.

lat9:
Updated to include the additional formatted fields.

drbyte:
Is the Vat4EU mod doing its own data-sanitization? Passing the `$incoming` array and using that for processing would result in bypassing htmlspecialchars. Is that intentional?

lat9:
That was the design, with the sanitized versions being passed as additional fields within the $p1 array. 

Vat4EU simply appends any gathered VAT number to the address, using only the ['cr'] element of that passed array.  I might have other customizations _in the wild_ that also make use of that notifier.

PR #2140, opened 2019-01-29
v156-admin-mail-strict

mc12345678:
Addressing strict notifications


mc12345678:
Already incorporated from the looks of it.  Closing the PR.

PR #2142, opened 2019-01-29
v156 strict notifications stats_sales_report_graphs.php

mc12345678:
Addressed strict notifications; however, appears that a section of code is not even executed.


drbyte:
Duplicate of #2127

PR #2143, opened 2019-01-29
v156 coupon_admin: Text about free shipping not displayable in commit…

mc12345678:
… area.

in evaluation of the following code, the variable `$coupon_name` does not have the field `coupon_type` in its results.  as a result, this appears to always evaluate to true based on the second condition of loosely comparing `$coupon_name->fields['coupon_type'] == '0' (string containing the value zero)

`($coupon_name->fields['coupon_type'] == 'E' || $coupon_name->fields['coupon_type'] == '0' ? TEXT_FREE_SHIPPING . '<br />' : '')`

### DIFF
--- a/admin/coupon_admin.php
+++ b/admin/coupon_admin.php
@@ -398,7 +398,7 @@
       }
       break;
     case 'update_confirm':
-      if ( ($_POST['back_x']) || ($_POST['back_y']) ) {
+      if ( (!empty($_POST['back_x'])) || (!empty($_POST['back_y'])) ) {
         $_GET['action'] = 'new';
       } else {
         $coupon_type = 'F'; // amount off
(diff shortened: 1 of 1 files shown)


mc12345678:
There was no thoughtful reason.

I do see though in execution of the change I proposed that I should not have removed the parentheses around the equality test. I apparently incorrectly remembered the order of operations and presumed that the equality test would be performed against the value of the `$_GET` instead of against the result of the `or` of being set and the value.

So on line 1337 the parentheses around the equality test needs to be restored and the error-suppression can be removed.

mc12345678:
Can't correct my previous post from my device. I should have said: performed against the value of the `$_GET` instead of against the result of the "or" (`||`) of being set and the value.

The result would have been the same with `or` or `||` in this case, but the `or` operator wasn't used here.

admin/coupon_admin.php
```diff
@@ -33,7 +33,7 @@
   if (isset($_GET['reports_page'])) $_GET['reports_page'] = (int)$_GET['reports_page'];
   if (isset($_GET['status'])) $_GET['status'] = preg_replace('/[^YNA]/','',$_GET['status']);
   if (isset($_GET['codebase'])) $_GET['codebase'] = preg_replace('/[^A-Za-z0-9\-\][\^!@#$%&*)(+=}{]/', '', $_GET['codebase']);
-  if (($_GET['action'] == 'send_email_to_user') && ($_POST['customers_email_address']) && (!$_POST['back_x'])) {
+  if (isset($_GET['action']) && ($_GET['action'] == 'send_email_to_user') && ($_POST['customers_email_address']) && (!$_POST['back_x'])) {
```
drbyte: Since you're adding multiple `isset($_GET['action'])` calls, why don't we just make sure it's actually set to an empty string at the top. Ref: line 13 where it's already done conditionally.

admin/coupon_admin.php
```diff
@@ -1331,11 +1336,14 @@ function check_form(form_name) {
...
+    if ($cc_list->EOF && (empty($_GET['cid']) || ($_GET['cid'] == $cc_list->fields['coupon_id'])) && empty($cInfo)) {
+      $cInfo = new objectInfo($cc_list->fields);
+    }
     while (!$cc_list->EOF) {
-      if (((!$_GET['cid']) || (@$_GET['cid'] == $cc_list->fields['coupon_id'])) && (!$cInfo)) {
+      if ((empty($_GET['cid']) || ($_GET['cid'] == $cc_list->fields['coupon_id'])) && empty($cInfo)) {
         $cInfo = new objectInfo($cc_list->fields);
       }
-      if ( (is_object($cInfo)) && ($cc_list->fields['coupon_id'] == $cInfo->coupon_id) ) {
+      if ( isset($cInfo) && (is_object($cInfo)) && ($cc_list->fields['coupon_id'] == $cInfo->coupon_id) ) {
```
drbyte: Can the `is_object($cInfo)` be replaced by `isset($cInfo)`?
(I recall when we added the `is_object()` check that it was for basically the same goal, just didn't use isset at the time.)

PR #2145, opened 2019-01-29
initialize some vars in preview_info pages

mc12345678:
If the page is loaded without posting information, notifications
may get thrown for attempting to assign an array to an undefined
value, this pre-sets the items allowing the values to have
some pre-existing value to carry through if present otherwise
to assign to an empty array.

### DIFF
--- a/admin/includes/modules/document_general/preview_info.php
+++ b/admin/includes/modules/document_general/preview_info.php
@@ -10,6 +10,10 @@ if (!defined('IS_ADMIN_FLAG')) {
   die('Illegal Access');
 }
 $languages = zen_get_languages();
+if (empty($products_description)) $products_description = [];
+if (empty($products_name)) $products_name = [];
+if (empty($products_url)) $products_url = [];
+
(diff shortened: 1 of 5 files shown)


drbyte:
What's the workflow that's loading the preview_info page without any POSTed data?
If no data is POSTed, maybe it's better to do an abort, and skip processing the page at all?

mc12345678:
An "accidental" flow path would be a time out while sitting on the preview page and attempting to reload the page.
A more purposeful path would be to view the data of the product using the parameter on the uri of `read=only` or coming into this from something else that is intended to pass along any of those three variables having additional data already incorporated into the array when loading the page with the parameter `action=new_product_preview`.  This could be done/used as an interception of the posted data as well to unset the post and only support specific information going forwards.  In either way, while addressing the need to ensure that the values were set as an array, I didn't want to unnecessarily wipe out any existing data for no reason.

I have nothing specific but was evaluating the impact and ways that the information/code could be used as well as how the remaining code on the page might be affected or override the option(s).

PR #2152, opened 2019-02-03
V156 admin index dashboard strict

mc12345678:
Address notifications made when using admin/index_dashboard.php while in strict mode.

### DIFF
--- a/admin/index_dashboard.php
+++ b/admin/index_dashboard.php
@@ -115,11 +115,14 @@ foreach ($whos_online as $session) {
   }
 
   if (empty($session['session_id'])) {
+    if (empty($spider_array[$status])) $spider_array[$status] = 0;
     $spider_array[$status] += 1;
   } else {
     if ($session['full_name'] == "&yen;Guest") {
+      if (empty($guest_array[$status])) $guest_array[$status] = 0;
       $guest_array[$status] += 1;
     } else {
+      if (empty($user_array[$status])) $user_array[$status] = 0;
       $user_array[$status] += 1;
     }
   }
--- a/admin/index_dashboard.php
+++ b/admin/index_dashboard.php
@@ -64,7 +64,7 @@ $report = 4;
 require_once DIR_WS_CLASSES . 'stats_sales_report_graph.php';
 $endDate = mktime(0, 0, 0, date("m"), date("d"), date("Y"));
 //$startDate = mktime() - (365 + 182) * 3600 * 24;
-$startDate = mktime() - (365 * 2) * 3600 * 24;
+$startDate = time() - (365 * 2) * 3600 * 24;
 
 //$startDate = mktime() - (365)*3600*24;
 $report = new statsSalesReportGraph($report, $startDate, $endDate);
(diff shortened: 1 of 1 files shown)


admin/index_dashboard.php
```diff
@@ -342,24 +345,24 @@ function drawCharts() {
...
-          <div class="col"><?php echo zen_image(DIR_WS_IMAGES . 'icon_status_yellow.gif'); ?>&nbsp;&nbsp;<?php echo $user_array[1]; ?></div>
-          <div class="col"><?php echo zen_image(DIR_WS_IMAGES . 'icon_status_red.gif'); ?>&nbsp;&nbsp;<?php echo $user_array[2]; ?></div>
-          <div class="col"><?php echo zen_image(DIR_WS_IMAGES . 'icon_status_red_light.gif'); ?>&nbsp;&nbsp;<?php echo $user_array[3]; ?></div>&nbsp;
+          <div class="col"><?php echo zen_image(DIR_WS_IMAGES . 'icon_status_green.gif'); ?>&nbsp;&nbsp;<?php echo (!empty($user_array[0]) ? $user_array[0] : ''); ?></div>
+          <div class="col"><?php echo zen_image(DIR_WS_IMAGES . 'icon_status_yellow.gif'); ?>&nbsp;&nbsp;<?php echo (!empty($user_array[1]) ? $user_array[1] : ''); ?></div>
+          <div class="col"><?php echo zen_image(DIR_WS_IMAGES . 'icon_status_red.gif'); ?>&nbsp;&nbsp;<?php echo (!empty($user_array[2]) ? $user_array[2] : ''); ?></div>
+          <div class="col"><?php echo zen_image(DIR_WS_IMAGES . 'icon_status_red_light.gif'); ?>&nbsp;&nbsp;<?php echo (!empty($user_array[3]) ? $user_array[3] : ''); ?></div>&nbsp;
```
drbyte: Since these arrays are initialized earlier, could all these be shortened using the `?:` syntax?

`echo $user_array[0] ?: '';`

mc12345678:
Unfortunately, while there is a possibility that each of the four sub-keys have been initialized, there is no guarantee that all or any more than one will be initialized, but the four states of each of the variables will be attempted. Perhaps on a busy store all of them will be defined, but not so on one that is somewhat idle.

Further, unfortunately using the "Elvis operator" as suggested would still result in a strict notification in the case(s) where the status was not detected. And the null coalesce operator while it would resolve the issue isn't available until php 7.x. Unless there was some sort of ditch out process to not cause an error for someone using what is reported to be an allowable php version (especially on the first page of the admin) then to "simplify" the code something else needs to be done. Either all possible status settings need to purposefully be stepped through to ensure that they are all set to 0 before attempting to add to them and then later display can be more "straight forward" or some level of filtering amenable to the supported software should be incorporated/kept. 

Basically, if all of the arrays are cycled through for each possible status and given a value then yes the "Elvis operator" as suggested would work for php 5.3 and above.

drbyte:
Agreed. Given that there are just 4 status keys (0-3) to initialize for each, let's just initialize them all at the start.
IMO initializing early is preferable, since:
the initialization step is all in one spot, semantic, and obvious to the reader/maintainer of the code, 
and then later output logic is also obvious, sensible and as brief as possible while being as explicit as necessary.

(I agree, null-coalesce would be more desirable, but we can't ... yet)

For extreme brevity for the init, what do you think of this? ...
```php
// initialize
$user_array = $guest_array = $spider_array = [0=>0, 1=>0, 2=>0, 3=>0];
```

mc12345678:
Init looks good: forces a 0 based key, all known keys set to 0 and all arrays initialized. Works.

PR #2155, opened 2019-02-09
use laravel blade as a template language

zcwilt:
Laravel Blade Integration
=========================

Laravel Blade is a template language. This means that templates based on Blade are initially compiled 
into pure php and also cached for future display. At the moment the idea is to only use this for admin, although in the future we may also use it for catalog side code
However regardless of that, templates can always use pure php and templates don't need to use blade structures/formatting.

Why use Blade
-------------

please see https://laravel.com/docs/5.2/blade.
Note we are using the 5.2 version to preserve PHP 5.6 compatibility.

Blade offers some advantages over using pure php templates.

- Template Inheritance

The use of @extends allows us to set a global layout template and then use @section, @yield etc to automate inclusion of sub templates.

- Simpler template language constructs
 @include
 @foreach
 @if
etc.

- mustache output and escaping

Something like 
    ``<?php echo TEXT_DASHBOARD_ADD_WIDGETS; ?>`` becomes 
    ``{{ TEXT_DASHBOARD_ADD_WIDGETS }}``
Mustache output also automatically escapes all output as well, although this can be overriden when unescaped output is needed.

- template cacheing

Legacy Templates
----------------

This integration is meant for admin templates and not currently for catalog templates.
Even given this, there is no requirement to use blade operators in any template. You can still write purely php based templates.

Examples
--------

``resources/views/admin/layouts/master.blade.php``

    <?php require_once('includes/template/common/tplHtmlHead.php'); ?>
    </head>
    <body class="skin-blue-light">
    <div>
        <?php require_once('includes/template/common/tplHeader.php'); ?>
        <div class="container-fluid">
            @yield('content')
        </div>
        <?php require('includes/template/common/tplFooter.php'); ?>
    </div>
    </body>
    </html>

note: this is a layout file and is meant to be extended by a template file.
e.g.

``app/resources/views/admin/index.blade.php``

    @extends('layouts/master')

    @section('content')
        <div class="row">
            <h1><a href="#" class="widget-add">{{ TEXT_DASHBOARD_ADD_WIDGETS }}</a></h1>
         </div>
        <div id="main-widget-container">
            @include('partials/dashboardWidgets')
        </div>

        <div id="add-widget" class="modal fade " tabindex="-1" role="dialog">
             <div class="modal-dialog">
                <div class="modal-content">
                     <div class="modal-header">
                        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span
                                    aria-hidden="true">&times;</span></button>
                         <h4 class="modal-title">{{ TITLE_MODAL_DASHBOARD_ADD_WIDGETS }}</h4>
                    </div>
                    <div class="modal-body add-widget-container">
                    </div>
                </div>
            </div>
        </div>

        <div id="widget-settings" class="modal fade " tabindex="-1" role="dialog">
            <div class="modal-dialog">
                <div class="modal-content widget-settings-container" >
                </div>
            </div>
        </div>

    @endsection

note that in the above we create a section called ``content`` and that it is this that is yielded in the layout.


zcwilt:
note: There are further changes required to make the use of blade easier and these will come in separate pr's.

PR #2159, opened 2019-02-12
156: Admin specials - Using pre add results in error

Zen4All:
Funny this never was reported, but when using the pre add feature, the $specials_price variable is never set. This results in a database error because $specials_price is not a proper decimal.
Removing the variable will result in sql using the default value for the field (0.0000)., and thus continuing to the next page.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2159)

### DIFF
--- a/admin/specials.php
+++ b/admin/specials.php
@@ -165,7 +165,7 @@ if (zen_not_null($action)) {
       $expires_date = ((zen_db_prepare_input($_POST['end']) == '') ? '0001-01-01' : zen_date_raw($_POST['end']));
 
       $products_id = zen_db_prepare_input($_POST['pre_add_products_id']);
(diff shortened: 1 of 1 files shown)


admin/specials.php
```diff
@@ -165,9 +165,8 @@
       $expires_date = ((zen_db_prepare_input($_POST['end']) == '') ? '0001-01-01' : zen_date_raw($_POST['end']));
 
       $products_id = zen_db_prepare_input($_POST['pre_add_products_id']);
-      $db->Execute("INSERT INTO " . TABLE_SPECIALS . " (products_id, specials_new_products_price, specials_date_added, expires_date, status, specials_date_available)
```
drbyte: I think removing both commas here is going to create an error.

PR #2161, opened 2019-02-13
Use loop instead of repetition to include files

scottcwilson:
This makes it easier to have multiple templates, and tightens up the
code.


drbyte:
> This makes it easier to have multiple templates, and tightens up the code.

I get that the loop makes the code tidier. 
But I'm trying to understand the "makes it easier to have multiple templates" part.

scottcwilson:
That's just a detail - some customers are asking to use responsive_classic on mobile and their own template on desktop.  This change reduces the number of mods required to do that.  But it's valuable aside from that as a code cleanup.

PR #2166, opened 2019-02-18
Accidental assignment in a condition $new_price =

mesnitu:
Most of the main_template_vars.php files are like this. I believe it's a mistake ?


drbyte:
No, it's intentional, because the next line uses (needs) that value defined, else it will fail.

PR #2172, opened 2019-02-25
Substitute a constant for hard code in dealing with dates.

mc12345678:
Can't seem to find the thread; however, there was a report that
when dates were modified to support a different "location" this
caused data to be incorrectly displayed on the index_dashboard.

This modification ensures alignment between the dates used to
compare and determine what to display.

### DIFF
--- a/admin/index_dashboard.php
+++ b/admin/index_dashboard.php
@@ -412,8 +412,8 @@ foreach ($whos_online as $session) {
 
         $ds = $dsc = $ys = $ysc = $msc = 0;
 
-        $now = date("d/m/Y");
-        $yesterday = date("d/m/Y", strtotime('-1 days'));
+        $now = date(DATE_FORMAT);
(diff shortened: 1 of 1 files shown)


drbyte:
Do I need to test this myself, or have you already tested this with multiple combinations?

mc12345678:
I'm still trying to find the forum thread; however, in trying to as provided arrangement with the default English layout and a test purchase on the same date as tested, a value of $0 showed for "today".  When I substituted `DATE_FORMAT`, then the value of the order that had been placed was confirmed.  I then modified the value of `DATE_FORMAT` in `admin/includes/languages/english.php` to `d/m/Y` and using the as provided code, the value showed for today as expected.  Again replacing the hard-coded text with DATE_FORMAT, the appropriate value remained.

Also this was identified/verified by review of the admin's `zen_date_short` function results which uses `DATE_FORMAT` as the date format for the result. 

So, yes, did test this and also understand the need to ask.

PR #2191, opened 2019-03-02
Addresses strict notifications identified on base page load

mc12345678:


admin/option_values.php
```diff
@@ -225,7 +229,7 @@ function init() {
         </div>
         <div class="row">
             <?php echo zen_draw_form('update_product_attributes', FILENAME_PRODUCTS_OPTIONS_VALUES, 'action=update_product', 'post', 'class="form-horizontal"'); ?>
-            <?php echo zen_draw_hidden_field('products_update_id', $_GET['products_update_id']); ?>
+            <?php echo zen_draw_hidden_field('products_update_id', (isset($_GET['products_update_id']) ? (int)$_GET['products_update_id'] : '0')); ?>
```
drbyte: `$_GET['products_update_id']` where is this ever set as a $_GET?

admin/option_values.php
```diff
@@ -244,7 +248,7 @@ function init() {
         </div>
         <div class="row">
             <?php echo zen_draw_form('update_categories_attributes', FILENAME_PRODUCTS_OPTIONS_VALUES, 'action=update_categories_attributes', 'post', 'class="form-horizontal"'); ?>
-            <?php echo zen_draw_hidden_field('categories_update_id', $_GET['categories_update_id']); ?>
+            <?php echo zen_draw_hidden_field('categories_update_id', (isset($_GET['categories_update_id']) ? $_GET['categories_update_id'] : '0')); ?>
```
drbyte: `$_GET['categories_update_id']` - where is this ever set as a $_GET?

admin/option_values.php
mc12345678: None found in a current version of Zen Cart.  There is reference to setting `$_GET['products_update_id'] = '';` in a few places where a product is being copied, (attributes_controller and category_product_listing), otherwise it seems that all instances are transferred/referenced by posting values.  Is it possibly for some historical plugin or other operation to remain backwards compatible? 

I've only looked back as far as ZC 1.5.5 for comparison/review as of this writing.

drbyte: ya, I can't see where they're relevant.
Removing.

PR #2192, opened 2019-03-02
Vertically center the jAlert popup window.

mc12345678:


drbyte:
Please redo this by adding new CSS that doesn't touch the original vendor-supplied file.

PR #2193, opened 2019-03-02
Add a notifier to the footer of the admin.

mc12345678:


drbyte:
What's the use-case for this?
What's the business problem this is solving?

mc12345678:
Has been a direction to be added for the plugin User Tracking to identify flow within the admin area.  Acts similar to the notifier on the catalog side that was added to the main_page.  Admin side does not appear to have such a similar/consistent "trigger point".  

Would simplify installation if no file needed to be modified for full usage on a default/version fully compatible installation.

lat9:
I could use it also for Super Globals, although I'd prefer the notification to be after the `<footer></footer>`.

mc12345678:
I'm impartial to the location within the file, looking for relative consistent inclusion and expectation that the notifier would fire (even if observer load points have to be adjusted).

drbyte:
Alright, ya, move it below the footer.

PR #2194, opened 2019-03-03
Undefined index: configuration_value

### LINKED ISSUE #2184: PHP Notice thrown when Modules->Payment->Free Order is not installed
As identified in the One-Page Checkout [support-thread](https://www.zen-cart.com/showthread.php?220781-One-Page-Checkout-Support-Thread&p=1355395#post1355395), a PHP Notice is issued by the `zen_get_configuration_key_value` function, present in `/includes/functions/functions_lookups.php`:

```
[27-Feb-2019 21:22:25 UTC] PHP Notice:  Undefined index:  configuration_value in  /home/pc/public_html/156a/includes/functions/functions_lookups.php on  line 558
```

... when called by `payment::__construct` via the following statement and the Free Charger payment-method is not installed (so the value is not present in the database):

```
        if (zen_get_configuration_key_value('MODULE_PAYMENT_FREECHARGER_STATUS') and ($_SESSION['cart']->show_total()==0 and (!isset($_SESSION['shipping']['cost']) || $_SESSION['shipping']['cost'] == 0))) {
```

That payment-class processing presumes that a null-type value will be returned by the function if the `freecharger` is enabled, but:

1. The `freecharger` status can be either **'True'** or **'False'**, neither of which are a null-type value.
1. When the function (see below) finds a null-type value, it returns a non-null HTML `span`.

```
/*
 *  configuration key value lookup
 *  TABLE: configuration
 */
  function zen_get_configuration_key_value($lookup) {
    global $db;
    $configuration_query= $db->Execute("select configuration_value from " . TABLE_CONFIGURATION . " where configuration_key='" . $lookup . "'");
    $lookup_value= $configuration_query->fields['configuration_value'];  //<-- Line 558
    if ( !($lookup_value) ) {
      $lookup_value='<span class="lookupAttention">' . $lookup . '</span>';
    }
    return $lookup_value;
  }
```

The issue here is, in my mind, two-fold:

1. The `payment` class constructor should just be checking for whether `MODULE_PAYMENT_FREECHARGER_STATUS` is defined and, if so, set to `True` rather than using the `zen_get_configuration_key_value`'s return.
1. The `zen_get_configuration_key_value` function should be checking to see that the `$lookup` configuration-key was found in the database prior to making its return.

There are a number of storefront and admin uses of the `zen_get_configuration_key_value` and I'm hesitant to make the correction to that function's processing since I'm not sure what the function's users expect to see if the associated key is not present in the database.

lat9:
Fixes #2184

### DIFF
--- a/includes/classes/payment.php
+++ b/includes/classes/payment.php
@@ -37,7 +37,8 @@ class payment extends base {
       } else {
 
         // Free Payment Only shows
-        if (zen_get_configuration_key_value('MODULE_PAYMENT_FREECHARGER_STATUS') and ($_SESSION['cart']->show_total()==0 and (!isset($_SESSION['shipping']['cost']) || $_SESSION['shipping']['cost'] == 0))) {
+        $freecharger_enabled = (defined('MODULE_PAYMENT_FREECHARGER_STATUS') && MODULE_PAYMENT_FREECHARGER_STATUS == 'True');
+        if ($freecharger_enabled && $_SESSION['cart']->show_total() == 0 && (!isset($_SESSION['shipping']['cost']) || $_SESSION['shipping']['cost'] == 0)) {
           $this->selected_module = $module;
           if (file_exists(DIR_FS_CATALOG . DIR_WS_MODULES . '/payment/' . 'freecharger.php')) {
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks

PR #2195, opened 2019-03-03
Ensure that $var_linksList is countable.

mc12345678:


mc12345678:
To also go "on record" from [What are the Server requirements to run Zen Cart?](https://www.zen-cart.com/content.php?48-what-are-the-server-requirements-to-run-zen-cart): 
>PHP Version
Using the latest version of Zen Cart is always recommended for maximum compatibility.
>
>Zen Cart v1.5.6 is designed for PHP 5.5 through PHP 7.3

A similar issue would currently be found in:
- includes/classes/currencies.php
- includes/modules/pages/checkout_payment/jscript_payeezy.php
- includes/modules/pages/contact_us/header_php.php

and multiple locations in the admin.

Are you advocating that code should be written and/or maintained for unsupported versions?

lat9:
I'm suggesting that code be written in a _downward-compatible_ manner.  It hurts **nothing** to use `array()` instead of `[]` and, instead, allows store-owners who have been putting off their Zen Cart upgrades (and associated PHP version upgrades, too) to have a less troublesome path for the upgrade process.

mc12345678:
Looking also at what the OP had suggested: https://www.zen-cart.com/showthread.php?225151-Myquote-Logs-quot-folder-keeps-filling-up-with-log-files&p=1355501#post155501 The solution places the assignment of the variable before the use of the module file.  Essentially if the operations of the module are not enabled (various database settings), then `$var_linksList` is neer set in the original code.  In trying to consider further how best to address this situation (after having proposed the above change), trying to step back and figure out what is the best/most appropriate way to handle this condition.

The thoughts/philosophies that come to mind are:
- Manipulate the variable (array, unset) regardless of any previous assignment. (if so, before or after the notifier?)
- Manipulate the variable (array, unset) if previously unset or empty. (if so, before or after the notifier?)
- Before the call to the module to manipulate the variable (assign as array or unset) with some level of assignment per the above or below.
- After the call to the module to verify that the variable can be used in the loop and if not, then assign to support.

I'm somewhat partial to the last option as a "last check".  For some reason I thought that PSR would answer the question, but I think I'm too deep in code at the moment so open to direction.

drbyte:
Agreed. Less is more. 
There's no need here for the isset/unset stuff. All that was really required was a proper init of the variable.

PR #2200, opened 2019-03-04
156: remove extra enter

Zen4All:
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2200)


drbyte:
In this case that linebreak was intentional.

Zen4All:
Why?

drbyte:
> Why?

To make the output html more readable. 
I suppose it could have been done with a bunch more code that wraps \n in double quotes or PHP_EOL constant etc. Line break was simpler.

Zen4All:
Okay, that's clear. thanks. Maybe we should add a comment in the code to alert others the break is intentional?

PR #2202, opened 2019-03-05
Store HTML email without de-fanging it

scottcwilson:
Reduces the number of core file changes for EMail Archive Manager.

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -784,7 +784,6 @@ use PHPMailer\PHPMailer\SMTP;
     return $customers_values->fields['customers_email_address'];
   }
 
-
   function zen_db_prepare_input_html_safe($string) {
     if (is_string($string)) {
       return trim(stripslashes($string));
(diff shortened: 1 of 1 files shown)


lat9:
I'll note that this change should also help in debugging email-formatting issues, since you'd be able to copy from the email_archive into an .html file for viewing in the browser.

includes/functions/functions_email.php
```diff
@@ -783,3 +783,19 @@ function zen_get_email_from_customers_id($customers_id) {
...
     return $customers_values->fields['customers_email_address'];
   }
+
+
+  function zen_db_prepare_input_html_safe($string) {
+    if (is_string($string)) {
+      return trim(stripslashes($string));
+    } elseif (is_array($string)) {
+      reset($string);
+      while (list($key, $value) = each($string)) {
```
drbyte: Note: This `reset/while/list/each` syntax is deprecated since PHP 7.2

Related: https://github.com/zencart/zencart/pull/1377

scottcwilson: Good catch.  Thanks!

PR #2209, opened 2019-03-11
Update languages.php

mesnitu:
Fix PHP Fatal error: 1054:Unknown column 'language_id' in 'where clause' :: DELETE FROM ezpages_content WHERE language_id

Or update the install file to use language_id (?)


drbyte:
Thanks!

PR #2215, opened 2019-03-17
ZC156: Optimize featured queries

Zen4All:
- Removing the extra clauses that are already set in the on clause.
- Moving the language clause to the LEFT JOIN
- DISTINCT is not needed

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2215)

### DIFF
--- a/includes/modules/featured_products.php
+++ b/includes/modules/featured_products.php
@@ -19,14 +19,13 @@ $featured_products_query = '';
 $display_limit = '';
 
 if ( (($manufacturers_id > 0 && empty($_GET['filter_id'])) || !empty($_GET['music_genre_id']) || !empty($_GET['record_company_id'])) || empty($new_products_category_id) ) {
-  $featured_products_query = "SELECT DISTINCT p.products_id, p.products_image, pd.products_name, p.master_categories_id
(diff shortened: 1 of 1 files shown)


drbyte:
I'm trying to recall why DISTINCT was added. 
I remember a long time back there were a few places where undesired duplicates were appearing. Maybe that's when DISTINCT got added ... instead of refactoring to a more explicit JOIN.

What are your thoughts? Is there any edge case where these changes could bring back unexpected results?
(I realize that these changes may actually bring back "more accurate" results, which is of course the intended result! Or at least the efficiency of the better query call when loading this page.)

Zen4All:
The only reason I can think of for the distinct is to suppress the same product_id for multiple languages, but that is where the $_SESSION['languages_id'] is for.
I tested on a few different shops (some live for the sake of creating random data), and could not find any obvious flaw in the new query.
But is if others can show I am wrong please do 😉 .

drbyte:
Update: The `distinct` has been there since this feature (and this file) was added in v1.2.0.
I don't think the use of `distinct` here is related to my earlier thought.

I agree with your proposed change. I've pinged @zcwilt for his thoughts.

PR #2218, opened 2019-03-18
Correct checkout_new_address notices

lat9:
If, during the checkout process, a customer chooses to change their shipping or billing address, selecting one of the addresses currently in their address book, a PHP Notice is generated since the new-address form has not been filled out:

```
[18-Mar-2019 08:14:05 America/New_York] PHP Notice:  Undefined index: firstname in C:\xampp\htdocs\zc156posm\includes\modules\checkout_new_address.php on line 30
```

### DIFF
--- a/includes/modules/checkout_new_address.php
+++ b/includes/modules/checkout_new_address.php
@@ -27,7 +27,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 
 if (isset($_POST['action']) && ($_POST['action'] == 'submit')) {
   // process a new address
-  if (zen_not_null($_POST['firstname']) && zen_not_null($_POST['lastname']) && zen_not_null($_POST['street_address'])) {
+  if (!empty($_POST['firstname']) && !empty($_POST['lastname']) && !empty($_POST['street_address'])) {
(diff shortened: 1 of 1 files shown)


includes/modules/checkout_new_address.php
```diff
@@ -27,7 +27,7 @@
 
 if (isset($_POST['action']) && ($_POST['action'] == 'submit')) {
   // process a new address
-  if (zen_not_null($_POST['firstname']) && zen_not_null($_POST['lastname']) && zen_not_null($_POST['street_address'])) {
+  if (!empty($_POST['firstname']) && !empty($_POST['lastname']) && !empty($_POST['street_address'])) {
```
mc12345678: May result in the same issue as with someone having the last name of 'Null' if they have the last name of 'false'.  Shouldn't an `isset` be incorporated instead of just `!empty`?

lat9: `!empty` encompasses `isset`.  If the customer's last name is either 'Null' (a string) or 'false' (also a string), the result of !empty will be `(bool)true`.

PR #2223, opened 2019-03-21
Ensure from name is set in zen_mail, especially for cron

scottcwilson:
If the email address in EMAIL_FROM isn't preceded by a name, in some cases the default name "Root User" will be used.

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -78,8 +78,8 @@ use PHPMailer\PHPMailer\SMTP;
       $from_email_name = trim($regs[1]);
       $from_email_address = $regs[2];
     }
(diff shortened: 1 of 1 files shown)


drbyte:
Are you seeing this on production sites? Or mostly just on dev sites?

I haven't seen that happen in years, and only on minimally-configured dev environments. Never saw it in production.

scottcwilson:
Live sites when scripts are called from cron.

drbyte:
Interesting. That's most likely because PHP-for-CLI (command-line) (used by cron) is often configured separately from PHP-for-web (apache/nginx/FPM), and some hosts neglect setting the CLI configs similar to the FPM configs.

PR #2237, opened 2019-03-31
ZC 1.5.6: Give the index dashboard a modern look

Zen4All:
In line with the other changes I gave the Dashboard a more modern look.
This matches the other pages and makes the dashboard easier to read. It is an in between Zen Cart 1.5.5 and 2.0

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2237)

### DIFF
--- a/admin/includes/stylesheet.css
+++ b/admin/includes/stylesheet.css
@@ -468,46 +468,14 @@ div.copyrightrow {
     padding: 25px;
     text-align: center;
 }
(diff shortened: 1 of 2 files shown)


drbyte:
@Zen4All Were you intending that the "Total" line shows fully right-aligned?
Screen Shot 2019-04-04 at 2 04 28 PM

PR #2248, opened 2019-04-06
Allow override of zen_has_product_attributes_values

lat9:


lat9:
Another notification; I didn't see that you'd previously accepted the prior PR.

PR #2260, opened 2019-04-08
Change admin-modules email message response

mc12345678:
Addressed strict notifications

### DIFF
--- a/admin/modules.php
+++ b/admin/modules.php
@@ -79,10 +79,10 @@ if (zen_not_null($action)) {
                         set configuration_value = '" . zen_db_input($value) . "'
                         where configuration_key = '" . zen_db_input($key) . "'");
       }
-      $msg = sprintf(TEXT_EMAIL_MESSAGE_ADMIN_SETTINGS_CHANGED, preg_replace('/[^\w]/', '*', $_GET['module']), $admname);
+      $msg = sprintf(TEXT_EMAIL_MESSAGE_ADMIN_SETTINGS_CHANGED, preg_replace('/[^\w]/', '*', (!empty($_GET['module']) ? $_GET['module'] : (!empty($_GET['set']) ? $_GET['set'] : 'UNKNOWN'))), $admname);
       zen_record_admin_activity($msg, 'warning');
       zen_mail(STORE_OWNER_EMAIL_ADDRESS, STORE_OWNER_EMAIL_ADDRESS, TEXT_EMAIL_SUBJECT_ADMIN_SETTINGS_CHANGED, $msg, STORE_NAME, EMAIL_FROM, array('EMAIL_MESSAGE_HTML' => $msg), 'admin_settings_changed');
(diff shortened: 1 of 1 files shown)


admin/modules.php
```diff
@@ -79,10 +79,10 @@
                         set configuration_value = '" . zen_db_input($value) . "'
                         where configuration_key = '" . zen_db_input($key) . "'");
       }
-      $msg = sprintf(TEXT_EMAIL_MESSAGE_ADMIN_SETTINGS_CHANGED, preg_replace('/[^\w]/', '*', $_GET['module']), $admname);
+      $msg = sprintf(TEXT_EMAIL_MESSAGE_ADMIN_SETTINGS_CHANGED, preg_replace('/[^\w]/', '*', (!empty($_GET['module']) ? $_GET['module'] : (!empty($_GET['set']) ? $_GET['set'] : 'UNKNOWN'))), $admname);
```
drbyte: Why the addition of `$_GET['set']` here, when it's not used in this `case`?

drbyte: @mc12345678 I'm still not seeing where the `$_GET['set']` got added from...

mc12345678: I've been trying to chase down more specific details; however, this is what I have.

On the day in question, I received an email from my test store saying that `NOTE: Admin settings have been CHANGED for the [] module, by your Zen Cart admin user (my_user_id_info)` with no indication as to what specifically had been modified. So it seemed either nothing had been modified and this was an errant message or something had been modified but insufficient information had been passed to the code to identify specifically what/where had been modified.

So, I also look in my User Tracking log: note that this reports based on the final page being loaded, not the intermediate transfers/redirects that might occur.  At the time in question (recent to oldest) I found:

16:22:55: | Shipping Modules | /directory/admin/modules.php?set=shipping
-- | -- | --
16:22:45: | Shipping Modules | /directory/admin/modules.php?set=shipping&action=edit
16:22:42: | Shipping Modules | /directory/admin/modules.php?set=shipping

with my next action some 20ish minutes later to go to the orders page.

The 16:22:42 entry was the first record of an admin being logged in at/near the time frame of the email being sent.  Reviewing/comparing the message content with the messages expected to be sent, it seemed that this area was the "source" of the message and seemed to show that it was possible it would provide some form of notification, but in this case it gave no unique information.  Basically sometihng somewhere changed...

Now I also know that in a previous issue I had been evaluating system response of returning to operation with a `$_GET` parameter action of `set` or similar and it is remotely possible that I attempted to reactivate a link on a browser tab to see what would happen, but I no longer have ready access to the server log of when this event occurred to know what was on the uri to land me at the above location.

Maybe it was a shotgun approach, but from the data that I gathered, it seemed like it was the source of my "mysterious" message of something changing somewhere but with no helpful information.  I believe I have received such similar messages in the past as well.

mc12345678: To add to the above, the text: `NOTE: Admin settings have been CHANGED for the [] module, by your Zen Cart admin user (my_user_id_info)` comes from the constant `TEXT_EMAIL_MESSAGE_ADMIN_SETTINGS_CHANGED` which is only used/presented in the edit section, line 82 of this file.

PR #2262, opened 2019-04-08
Admin/category_product_listing lost class parameter

mc12345678:


admin/category_product_listing.php
```diff
@@ -1247,7 +1247,7 @@ function init() {
         }
         ?>
       </div>
-    </div>
+    <!--</div>-->
```
drbyte: I understand the `class` thing, but why remove the `div`?

mc12345678: Receiving an unmatched end tag when that additional </div> is present.  Noticed that there also is an "offset" in formatting that may make one think that a tag is necessary, but this happens earlier on with the open `<h1` tag also closing on the same line and the next html tag is offset to the right and therefore appears to need some sort of closure to return to the same closing point as `<div class="container-fluid">`

Sorry that I didn't discuss it in my submission.

Zen4All: I do not think it is an error., but please prove me wrong.
According to my file, your "extra closing" div belogs to line
https://github.com/zencart/zencart/blob/4c0d5e45dc596c69a3c4ddb7a4b2882127a9be6e/admin/category_product_listing.php#L566

drbyte:
Merging for the class fix.

If the `</div>` issue is legitimate, please submit a PR for it, with details.

PR #2267, opened 2019-04-09
Corrects code for processing the val_function of configuration

### LINKED ISSUE #2263: val_function - is it and the supporting code, ready?
The other day a user of new ZC 1.5.6a ran into a problem with setting the email address of one of the configuration email settings where an email address like this `name <email@domain>` turned into `name &lt;email@domain&gt;`. So I began looking into what it might take to clean up the response to be acceptable.

Thing is, because of the admin sanitizer, the data that is fed to the configuration value check is already strictly sanitized.

So question is, is that really the desired and expected condition? Before I take the already cleaned up content to make it usable for final result, I'd like to be sure that there isn't some other quick flip expected. Especially since I was looking at offering it as a class for the base code.

mc12345678:
Adds class to which additional callback type functions could
be added to supplement standard filter_var style filters and
validators.
Adds an email filter to support filtering email addresses such as
CONTACT_US_LIST.

The new class function can be access by the 'options' key
of the val_function being set to array('configurationValidation', 'sanitizeEmail')
and the 'id' set to FILTER_CALLBACK or integer equivalent.

This also modifies the admin sanitizer to bypass the
configuration_value sanitization if there is a val_function for
the associated configuration_key, and sanitizes as normal if
there is no val_function.

Changed the return on failure to support editing the value again.

Corrected call to link generator from `zen_admin_href_link` to
`zen_href_link`.

Lastly refactored the query in admin/configuration.php to remove
the additional quotes around the casting to integer.

Added a language define for the configuration page to support
the response text that might be expected for the default installed
val_function call.

Fixes #2263.

Do think that perhaps the return message could/should include
the content that was originally entered, even if it is strictly
sanitized to help the data enterer.  Haven't implemented this yet.

As far as I can tell, have tested all potential conditions of using a 
valid email address and an invalid email setup, though the "name"
may be sanitized in a way that may over sanitize as well. Hopefully
though an improvement above what currently exists.


mc12345678:
Seems like perhaps a language file should be "set aside" to help with the error message definitions instead of trying to group them all in `admin/includes/languages/english/configuration.php`. Ideas or suggestions?

I was considering a separate file such as `admin/includes/languages/english/configuration_errors.php` and was momentarily thinking of a "template" override, but these are in the admin...

Also need to go ahead and update the zc_install/upgrade files to at least address the few items that are covered directly by the additional code of this PR.

And a note for anyone trying to further develop/use the val_function on boolean values (`0`/`1` or similar).  The boolean validation (`FILTER_VALIDATE_BOOLEAN`) will return `false` if the value is boolean `false`, but also will return `false` if it is neither a `1` nor a `0` such as an unrelated word `run`...  This php document addresses this in more detail: https://www.php.net/manual/en/filter.filters.validate.php. 

 There are a few other potential responses, but even if a `default` is set, the `default` is what is returned for those/that situation.  Basically the boolean test is only good at validating that the value tested is in fact `true`, all other responses are `false`. So beware blindly using that validation.  Would suggest creating a separate "user" defined function to discern between an original value of `run` or `0` as part of validating that the value is a boolean.

drbyte:
I guess another language file would best be named similar to the class it's related to, so something like `configuration_validation.php` is probably good.

drbyte:
The changes here are not trivial. Is this best in 1.5.6x? or for 1.5.7?

mc12345678:
> The changes here are not trivial. Is this best in 1.5.6x? or for 1.5.7?

Well, changes made in 1.5.6 perform strict sanitization on `configuration_value`.  This causes entries such as email addresses in the format of `name <email address>` to no longer be enterable in the configuration window and the default `BREAD_CRUMBS_SEPARATOR` of `&nbsp;::&nbsp;` to not be able to be entered. There certainly are other ways to make it possible to positively use the configuration menu such as not sanitizing `configuration_value`, to execute SQL statements to populate the database with such information, or add-on/develop another data entry tool that would allow population of the associated information.

Not expecting open discussion of "when" 1.5.7 would be released, it seems that something needs to be done to restore the functionality while continuing to maintain the desired data security.

Yes, there are likely a number of configuration items that would need to be touched to incorporate this beyond strict sanitization; however, it seems to me that the ones of concern really are those where data may be manually entered (text boxes). The "things" I'm worried about are where plugins have modified say the key value of a dropdown to use text instead of a numerical index.  On the one hand it might be expected that the val_function would check for integers, but on the other a program depends on non-numerical keys to operate. Is either right or wrong?

drbyte:
> The "things" I'm worried about are where plugins have modified say the key value of a dropdown to use text instead of a numerical index. On the one hand it might be expected that the val_function would check for integers, but on the other a program depends on non-numerical keys to operate. Is either right or wrong?

Without an example, I'm not seeing an issue here, unless of course you're wanting to use specialchars in the dropdown keys, which seems like something we should break so devs don't do it in the first place.

mc12345678:
>> The "things" I'm worried about are where plugins have modified say the key value of a dropdown to use text instead of a numerical index. On the one hand it might be expected that the val_function would check for integers, but on the other a program depends on non-numerical keys to operate. Is either right or wrong?

> Without an example, I'm not seeing an issue here, unless of course you're wanting to use specialchars in the dropdown keys, which seems like something we should break so devs don't do it in the first place.

@lat9 would have to identify an example.  I recall her saying that in I believe it was the use of dropdowns (`zen_draw_pull_down_menu`) that in some cases she uses a meaningful word as a key instead of an anonymous number.  I know I have wanted to do that for a configuration setting for one reason to be able to display the human readable meaning of the choice made, but have instead resolved on adding more functions to display that information instead of modifying the ZC code.  The effect though is that if the verification tag is to identify that the `configuration_value` is an integer (key to the dropdown), then storing text would not work.

drbyte:
> The effect though is that if the verification tag is to identify that the `configuration_value` is an integer (key to the dropdown), then storing text would not work.

Correct. I'm not sure why it needs to be an integer though.

mc12345678:
> Correct. I'm not sure why it needs to be an integer though.

Good to hear, even if it was an expression of doubt. Sanitize/validate that doesn't cause an outward problem or cause a compatibility issue is what I understand.

lat9:
> @lat9 would have to identify an example. I recall her saying that in I believe it was the use of dropdowns (`zen_draw_pull_down_menu`) that in some cases she uses a meaningful word as a key instead of an anonymous number. I know I have wanted to do that for a configuration setting for one reason to be able to display the human readable meaning of the choice made, but have instead resolved on adding more functions to display that information instead of modifying the ZC code. The effect though is that if the verification tag is to identify that the `configuration_value` is an integer (key to the dropdown), then storing text would not work.

FWIW, I don't remember saying such a thing.  If you 'refresh' my memory with a posting, I'll be happy to comment.

mc12345678:
On one of my access points I have an answer/statement mostly written up in response to lat9; however, not on this device and there is too much detail and research done to recreate at the moment. Short story, don't know where the post was as it was a side topic off of another relatively unrelated discussion, but it stuck with me because I had to research the code and how to get around the issue discussed.

So, a couple of other thoughts on this modification.  It was pulled from the development version as can be seen by the use of zen_admin_href_link which has not yet made it to this code set. So either it should be brought in or the above code modified as right now, this area of code will fail without that function.  Trying to continue to work through/help with this aspect of the code and correct the observed issues.

chindraba-work:
Glad to see this issue has been raised. I hope that it is addressed in the next release (current is v1.5.6b) as I am loath to fiddle with security related things in something this large, and critical. The CONTACT_US_LIST issue completely breaks the Contact Us email function if the default From email is a `no-reply` type address, as the email never gets received. I've been doing direct database edits to "correct" the sanitized data stored by the admin settings.

mc12345678:
@drbyte,
On my test server I have this functional for the email addresses. This PR needs to get updated to reflect the additional changes. A couple of things though.

1. The changes affect the database, which means that an upgrade path is needed which seems to imply that it can not be incorporated into ZC 1.5.6 because of the database changes that would be needed for someone to upgrade within 1.5.6. They could be incorporated for those upgrading to 1.5.6.
2. Because I initially applied this to 1.5.6 and in trying to adhere to the direction of pushing only to 1.5.7 and not adjusting the merge point, this branch needs to be against 1.5.7. I can move it myself, but direction has been not to.
3. If/when moved to 1.5.7, I'll need a moment to ensure I create the proper upgrade code so that the installer works properly, though I think there are enough examples to be able to do it relatively easily.
4. The changes that need to be incorporated into this are to put the language incorporation into the class' `__constructor()` method and to adjust the `val_function` code in the database just a little. I plan to make each of those a separate commit to correct my earlier commit. I could force push those changes, but have a feeling that others may have already pulled this and I thought by applying a separate commit it would clarify what was modified.

Thoughts?

zcwilt:
Just a note.
I'm, currently testing this.
once done, I'll rebase it against v157 and after that you can then address the upgrade path.
It would definitely be nice to have this as part of any 156(c) release.

zcwilt:
Given this is now rebased to v157 and the merge conflicts fixed, are you happy for it to be merged as is.

mc12345678:
I was satisfied with its functionality and coding before the rebase. I have not attempted comparing this final result (rebased) with what I have on my development server.  In principle I am satisfied, I just do not know if the "move" carried it over completely.  (Of course do still need to address the upgrade path portion as well.)

mc12345678:
Something for discussion/pondering though. In review of the php manual on use of the the associated filter, it does not accept dotless domains as an accepted email. The impact of that is installation of the software and attempt to use an email address such as `thissite@localhost` would not be accepted.

I do not know if such a configuration is "routinely" used or desired to be incorporated. Such an email address would have to be independently treated within the new class code to evaluate if it is dotless and then go from there...

chindraba-work:
I consider the PHP filter to be defective. [RFC 822](https://www.ietf.org/rfc/rfc822.txt) is anything but new. Any filter claiming to apply the syntax of RFC 822 should _not_ have exceptions to the basic rules. Comments and whitespace folding are rarely needed in most applications, and are difficult, at best, to validate. Their difficulty makes their inclusion in a library that much more important. Dotless domains, on the other hand, are simple to validate and very common in enterprise environments. The filter not working with them is inexcusable.

As to application in Zen-Cart, I don't think `user@localhost` is likely to be very common outside of test-bed installations. In such cases the user can do direct database edits to make it work. Mail sent to the localhost on a publicly accessible webserver isn't likely to be very effective anyway. Dotless domains other than `localhost` could be seen in the wild, however. A company could do self-hosting, and the webserver could be part of the local network as well. Order confirmation emails could be sent to `office@fulfillment` and the Contact Us page could include emails for `dave@sales` and `returns@accounting`. All would be valid within the network, which the webserver is part of, and would result in properly delivered emails if the hosts named are part of that network.

How "routinely" either setup is used I cannot guess. I almost always use `something@localhost` when doing test installs on my local machine. Such as `orders@localhost`, `new_accounts@localhost` `admin@localhost`, `reviews@localhost`, and etc. That way I know which link triggered the email, it arrives in my system mail box, and can verify that what I intended to send, or trigger, got the results I expected. I wouldn't dream of such a setup on a production level server, and wouldn't even have the ability for that machine to receive email at all.

For my use case, I can replace `localhost` with `127.0.0.2` and make it work the same. For the posited scenario above, that would be decidedly inconvenient. I would expect that anyone able to set up such a network probably has the skills to place the dotless domains into the database directly anyway, so their exclusion from Zen-Cart's general setup might not be worth the extra work, in developing it, or in executing it every time an email is validated.

mc12345678:
> I consider the PHP filter to be defective. RFC 822 is anything but new. Any filter claiming to apply the syntax of RFC 822 should not have exceptions to the basic rules. Comments and whitespace folding are rarely needed in most applications, and are difficult, at best, to validate. Their difficulty makes their inclusion in a library that much more important. Dotless domains, on the other hand, are simple to validate and very common in enterprise environments. The filter not working with them is inexcusable.

I would agree with the "defective" part from the perspective of counting on the filter to truly handle all types of email addresses. I can see (perhaps an excuse) why it does not by default allow/recognize at least dotless domains. If this filter were applied on the customer input side (think about the contact us page) and there was no unique identifier to allow/recognize that a dotless domain address had been entered, then such user could enter something like `root@localhost` as the email address and it would be accepted, how a system then uses that email address (again think about global internet not just Zen Cart development) could cause an issue.  Yeah it would be nice if there was an option added to the filter to support such evaluation; however, it also seems like it could be something relatively easily addressed by inspection of the entered email address after found to fail the filter.

> As to application in Zen-Cart, I don't think user@localhost is likely to be very common outside of test-bed installations. In such cases the user can do direct database edits to make it work. Mail sent to the localhost on a publicly accessible webserver isn't likely to be very effective anyway. Dotless domains other than localhost could be seen in the wild, however. A company could do self-hosting, and the webserver could be part of the local network as well. Order confirmation emails could be sent to office@fulfillment and the Contact Us page could include emails for dave@sales and returns@accounting. All would be valid within the network, which the webserver is part of, and would result in properly delivered emails if the hosts named are part of that network.

This to me indicates that although *I* do not personally use such email addresses that there is an identified case where it would make sense to support at least on the admin side of entry.  As  indicated above, it seems that incorporating some sort of flag/identifier to allow/prevent the use of dotless domain allowance would be of benefit to users of the software (developers are users so *I* think that they should be considered, I just didn't want to target if there was not an identified need).  

> How "routinely" either setup is used I cannot guess. I almost always use something@localhost when doing test installs on my local machine. Such as orders@localhost, new_accounts@localhost admin@localhost, reviews@localhost, and etc. That way I know which link triggered the email, it arrives in my system mail box, and can verify that what I intended to send, or trigger, got the results I expected. I wouldn't dream of such a setup on a production level server, and wouldn't even have the ability for that machine to receive email at all.

Well this is simply a further indication that having the ability out-of-the-box to handle at least dotless domains as discussed above.

> For my use case, I can replace localhost with 127.0.0.2 and make it work the same. For the posited scenario above, that would be decidedly inconvenient. I would expect that anyone able to set up such a network probably has the skills to place the dotless domains into the database directly anyway, so their exclusion from Zen-Cart's general setup might not be worth the extra work, in developing it, or in executing it every time an email is validated.

And this is why I at least documented the "constraints" of the method, so that others could either see that it had been considered even if nothing further comes of it; however, I think it is important to address such things when they are identified rather than to just leave them for someone else to work around or address at least if there seems to be a standard use case.  I mean I have other email addresses that I use to do the testing of above and yes that requires my test computer to be on the internet at that time and not in a bottle; however, that is my use and I have added this code to try to address the use expected by others.  So in answer to @zcwilt, perhaps good as is, but expect additional functionality to be added.

Now as to comments and whitespace folding, I expect that I may be able to identify more by a deeper review of the RFC; however, I would welcome one or more examples of each so that I could get an idea of what would be needed to validate that they are/were properly incorporated.

chindraba-work:
I think the limits imposed by the PHP filter could be acceptable for here. As-is seems ok for this PR, and, as you said, expect additional functionality, or in the minimum, don't box out the possibility of changes.

Comments and whitespace folding are esoteric, at best, in the current network environment. I think they can be safely ignored here. Any user with such an email address, customer or admin, has already become accustomed to that being rejected and has developed an alternative. I have developed alternatives to those sites which still reject 4-char + tld addresses, such as .space, .blog, .museum, etc. It becomes annoying when it's an important use-case, such as my pharmacy, yet I have accommodated their out-dated filters. (Not without a note, sometimes friendly, to the webmaster about it, of course.)

Dotless domains, as noted above, can be problematic on the store-front side - outside of testing. I think that allowing them in the admin side would be nice, without being a deal-breaker if they are not available. Some kind of notice about the dotless-ness being the reason for the rejection could be helpful to the admin user. Doing so, however, could be just as involved as allowing them to begin with.

From my practice, the biggest issue was the "sanitizing" of what I entered, resulting in html entities rather than valid email characters in the database, and the breaking of the Contact Us page. Things that need to be done for a testing environment might be inconvenient, as long as it "works" for production it's acceptable.

drbyte:
@mc12345678 
While I previously suggested a separate language file, looking at it now I feel like it's not needed.

Do you have any objections to just moving it into the `configuration.php` language file?

mc12345678:
> While I previously suggested a separate language file, looking at it now I feel like it's not needed.
>
> Do you have any objections to just moving it into the configuration.php language file?

Being open: yes and no on objections.  My one feeling will be hurt :) then on the no side, if to be considered usable "only" for the configuration menu as compared to say the admin user enter area or the customer review process (to validate the email address meets the criteria "monitored" by the associated filters) then absolutely am not worried about the placement of the constant(s), just mostly about them existing and not causing some sort of system meltdown with newer PHP requirements, etc...

Biggest "concern" I was trying to address was the ability to add language defines to support one's potentially additional sanitizer(s) and/or conditions.  Just felt like such changes may be better made in an independent file, but then again would that be appropriate for ever "thing" to have its own file filled with a bunch of other information that mostly just takes up file space? I do also see that the class itself doesn't return "words" about the issue, that seems to come from something else that processes the result(s) making the load of the language file on each page unnecessary except where used.

Assuming the above internal/external debate has been at least briefly reviewed and it still seems better to draw the language into configuration.php, then cool. I've no problem doing it. :)

drbyte:
Thanks.

While discussing with @zcwilt about plugin system updates and language-file-loading improvements, this case came up as an anomaly where this file was getting loaded outside the way most language-loading happens, and that creates a complicated wrinkle in making new changes easier to implement and to be used by mod maintainers/developers etc. 

Your point about using the validation external to the current intended workflow is noted.

We'll mull it over some more ;)   Thanks!

zcwilt:
Just to add some context. Have been working on a hybrid language loader for admin based on the installer system of using php arrays. It's a hybrid in the sense that it needs to support both legacy style define files and new array based files. 

To do this it needs access to the language loader class to run a method to actually `require` the language file. 

The problem arises in the configurationValidation class contructor, that requires the language file.
There are a number of ways of addressing that, one being moving the language defines somewhere that's already auto-loaded. I could just for now global the language loader concrete object, or inject it in the constructor from where it is called in the configuration page.

zc_install/sql/install/mysql_zencart.sql
```diff
@@ -2569,28 +2569,28 @@ INSERT INTO configuration (configuration_title, configuration_key, configuration
...
+INSERT INTO configuration (configuration_title, configuration_key, configuration_value, val_function, configuration_description, configuration_group_id, sort_order, date_added) VALUES ('Send Copy of Order Confirmation Emails To', 'SEND_EXTRA_ORDER_EMAILS_TO', '', '{"error":"TEXT_EMAIL_ADDRESS_VALIDATE","id":"FILTER_CALLBACK","options":{"options":["configurationValidation","sanitizeEmail"]}}', 'Send COPIES of order confirmation emails to the following email addresses, in this format: Name 1 &lt;email@address1&gt;, Name 2 &lt;email@address2&gt;', '12', '12', now());
+INSERT INTO configuration (configuration_title, configuration_key, configuration_value, val_function, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Send Copy of Create Account Emails To - Status', 'SEND_EXTRA_CREATE_ACCOUNT_EMAILS_TO_STATUS', '0', '{"error":"TEXT_EMAIL_ADDRESS_VALIDATE","id":"FILTER_CALLBACK","options":{"options":["configurationValidation","sanitizeEmail"]}}', 'Send copy of Create Account Status<br />0= off 1= on', '12', '13', 'zen_cfg_select_option(array(\'0\', \'1\'),', now());
```
drbyte, line 2577: This STATUS entry was reverted in #3346

PR #2273, opened 2019-04-11
Remove language files from responsive_classic.

scottcwilson:
Having these replicated files just makes it harder for people to
figure out what they need to update or migrate.


drbyte:
While I understand the sentiment, I don't think it's helpful.
The text in all of these files is different. Deleting would be unfruitful.

PR #2278, opened 2019-04-17
product_reviews_write header_php

mc12345678:
Initially began with correction to the check of a customer being logged in (strict processing) that led to other review considerations:
require of file not to use parentheses.
Potential of the products_id not being fed to access the page.
Refactoring for code alignment.
Remove else statements where the else action is simple. (assign a default value then modify if conditions warrant an alternate result).
Add notifiers to leverage options available.

### DIFF
--- a/includes/modules/pages/product_reviews_write/header_php.php
+++ b/includes/modules/pages/product_reviews_write/header_php.php
@@ -6,7 +6,7 @@
  * @copyright Copyright 2003-2016 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: Author: DrByte  Sun Oct 18 23:03:39 2015 -0400 Modified in v1.5.5 $
+ * @version $Id: Author: DrByte modifier: mc12345678 Sun Oct 18 23:03:39 2015 -0400 Modified in v1.5.6b $
  */
 /**
  * Header code file for product reviews "write" page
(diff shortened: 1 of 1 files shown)


lat9:
The change on line 18 will allow guest-customers to write reviews; it that behavioral change intended?

drbyte:
> The change on line 18 will allow guest-customers to write reviews; it that behavioral change intended?

Hmmm, probably not.

mc12345678:
Wouldn't the notifier/observer allow control of that aspect where the plugin providing guest checkout would control the situation?

lat9:
Nope, at least for One-Page-Checkout, the `zen_in_guest_checkout` simply identifies whether or not the customer is currently checking out as a guest.

mc12345678:
So then what's the issue? The change offered the possibility of guest like software to permit writing reviews, but if there was no such software then the user would have to be logged in to write a review. Afterall, a review could be inserted into the checkout process following the use of One-Page-Checkout...

lat9:
The OPC plugin offers two controls, one for pages allowed post-checkout and another for pages that are disallowed during guest-checkout.  Your addition of having the `product_reviews_write` page check for `in_guest_checkout` will make it so that current users of OPC that have allowed reviews for guests will have that functionality broken (since in-guest-checkout will be active on entry).

lat9:
In short, I'm asking that the `zen_in_guest_checkout()` portion of the comparison be removed, to provide downward-compatible operation.

drbyte:
That makes sense. Reviews aren't written "during checkout" anyway (er, reputable ones anyway).

mc12345678:
Matter of perspective, *product* being reviewed, and what aspect of review is being sought. Besides that all could be done with just `zen_is_logged_in` instead of having/using two separate functions, so still not an issue for someone that wants to offer the additional capability.

PR #2285, opened 2019-04-22
ZC 1.5.6 bug fix: Change to fix warning when $var_linklist is empty

Zen4All:
as reported on the forum
https://www.zen-cart.com/showthread.php?225343-myDEBUG-file-created-from-responsive-classic-template-(stock)&p=1356969#post1356969

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2285)


mc12345678:
Note that this locally resolves the same issue that was corrected at: https://github.com/zencart/zencart/commit/87046999a9d4cb666a016a951953f92f34d03e8b

PR #2286, opened 2019-04-23
v156 Shipping Estimator retain ship-to selection after cart edit

### LINKED ISSUE #2234: Shipping estimator "blows" currently-set shipping address
I've been chasing an issue for days trying to get One-Page Checkout readied.  Essentially, if a customer starts the checkout process (supplying a different shipping than billing address) and then (via the `shopping_cart` page) edits their order's contents and re-enters the checkout, their previously-set shipping address has been reset by the shipping-estimator's processing.

FWIW, I've verified that the shipping-estimator's processing also affects the 3-page checkout.  If you've chosen a different shipping-address than the customer's default address and click back to edit the shopping cart, the ship-to address has been (silently) reverted to the customer's default address.  This could lead to customer confusion, since I'm not sure that a customer will recognize that the address has been changed.

mc12345678:
As identified in issue #2234, if:
- a user is logged in, 
- has gone beyond the `shopping_cart` page, 
- has changed the destination (`sendto`) address,
- has returned to the `shopping_cart` page, and
- the store is setup to display the potential shipping charges on the `shopping_cart` page.

Then when they arrive back at the shopping cart page, the selection determined to be active is the customer's default address that is then carried through the remainder of the checkout process until the customer again changes the selection.  This commit chooses the selection that was previously made/available.

Also, for a customer that is not logged in and because of PR #1952, if the shipping estimator is presented on the shopping cart page, the zip code is not provided and a country has been selected, then upon refresh of the page, the country selection is restored back to the store's default address.  This commit moves the strict notification handling specifically to the zip code within this section of code using the variable `$zip_code` that is set earlier in processing to be set to the posted zip code if provided, otherwise the existing session's zip code and if that is not present to set it to an empty string (`''`). This area is expected to be executed after the shopping_cart has been visited, but before logging in for ordering allowing the posted zip code to take the place of the existing session zip code if it exists.
Fixes #2234.

### DIFF
--- a/includes/modules/shipping_estimator.php
+++ b/includes/modules/shipping_estimator.php
@@ -55,6 +55,9 @@ if ($_SESSION['cart']->count_contents() > 0) {
     if (isset($_POST['address_id'])){
       // user changed address
       $sendto = $_POST['address_id'];
+    } else if (!empty($_SESSION['sendto'])) {
+      // user has previously selected a destination address
+      $sendto = $_SESSION['sendto'];
     }elseif (!empty($_SESSION['cart_address_id'])){
       // user once changed address
       $sendto = $_SESSION['cart_address_id'];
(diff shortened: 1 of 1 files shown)


ajeh:
:+1:

PR #2289, opened 2019-04-24
unmatched end tag

mc12345678:
in a default install of ZC 1.5.6a+ on this page.

I navigate to other admin pages and do not have an unclosed tag
(meaning I don't appear to have an extra tag somewhere above the
main body code).

I was asked to "prove" it, though don't know exactly how I can
do that other than possibly show the source code possibly with
my private information obscured.

 I have again installed the code to a new site location and applied demo products.  When looking at this page using Firefox Developer, I receive the following message: Source map error: request failed with status 404
Resource URL: https://template7.mc12345678.com/v156/sWeet-jmg-atTic/includes/css/bootstrap.min.css
Source Map URL: bootstrap.min.css.map'

On Microsoft Edge I receive an unexpected end tag for the </div> at "line" 1363.  I've repeated this on multiple computers and browsers, but it remains an issue.  I've even reloaded with the changes incorporated today.  Some browsers "brush it off" and ignore it, while others will point out that there is in fact an extra </div> at the end of the code.  I was previously asked in #2262 to "prove" it. I can provide the entirety of the source page code that could then be compared/evaluated by an html validator if that will work for those enquiring.  I had started to sit down and manually parse the tags of the source code to identify what closed out what but then after a few hundred lines realized that it would only prove to me that there was an issue, not to anyone else.

So I still stand by the removal of the flagged </div> tag, though it also looks like the code needs to be refactored so that tags line up better.

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -564,17 +564,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
       <?php } ?>
       <div class="row"><?php echo zen_draw_separator('pixel_black.gif', '100%', '1px'); ?></div>
       <div class="row">
-          <?php
-          if ($action != '') {
-            ?>
-          <div class="col-xs-12 col-sm-12 col-md-9 col-lg-9 configurationColumnLeft">
-              <?php
-            } else {
-              ?>
-            <div>
-                <?php
-              }
-              ?>
+        <div class="<?php echo (empty($action)) ? '' : 'col-xs-12 col-sm-12 col-md-9 col-lg-9 configurationColumnLeft'; ?>">
             <table class="table table-striped table-hover">
               <thead>
                 <tr>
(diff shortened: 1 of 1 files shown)


drbyte:
I think the problem is elsewhere. 
More likely here:
https://github.com/zencart/zencart/blob/24993c302335b92b8775e0971503b95ba49139ca/admin/category_product_listing.php#L570-L579

@Zen4All @mc12345678

Zen4All:
Which mean that it was a false positive? Right?

mc12345678:
@drbyte
Not sure which aspect of the issue is discussed by that section of code of lines 570 to 579.  Either side of that if generates an opening `div`. One is shown "offset" from the other (further to the right on the second one), but in either option the additional closing `div` of this commit causes "the right browser" to identify that there is an additional unmatched end tag.

@Zen4All
I think that the code section identified by DrByte is only showing how trying to identify the equality of opening and closing tags could be incorrectly identified if using vertical alignment of tags alone. So, not sure about the "false positive" perspective.

drbyte:
I meant I do see the mismatch in w3c validator when viewing the categories list in admin. 

Not a false positive. 

I was trying to identify where the mismatch started. Requires more investigation.

Zen4All:
Yes that's what I meant. The code drbyte mentioned is causing the mismatch. I checked the file multiple times, and can not find another explanation.

drbyte:
Found it. Yes, it was mismatched. I chose to remove the other one since it matches the indentation of the opening tag. 
I removed the if/else so that IDE helpers are less confused by it.

PR #2295, opened 2019-04-24
zen_update_orders_history: Correct PHP Notice

lat9:
... due to misspelled variable name.


drbyte:
Aren't typos fun!?!!

drbyte:
@lat9 what do you think about whether we should touch `NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE` in orders.php class? 
ie: the extra 'i'.

lat9:
Yes, they are ... not!

For that notifier, it is (in my mind) a legacy value and should not be changed as the change would undercut any observer that's currently "watching".

drbyte:
> For that notifier, it is (in my mind) a legacy value and should not be changed as the change would undercut any observer that's currently "watching".

Agreed. And adding a corrected spelling one beside it could make it double-fire if a plugin chooses to listen on both!  :(

PR #2293, opened 2019-04-24
156 index filters

Zen4All:
Reworked filters

- consolidated 4 queries in to 1, and only put the difference in an if statement.
- modernised the rest of the code, like while to each conversion

see #2282

Issue #2282, opened 2019-04-22
ZC 1.5.6: Reduce complexity in default filter queries


Zen4All:
This has been bugging me for some time.
The queries below can easily be boiled down into one query with some simple php if statements. This will reduce the four queries in to one, which is much easier to maintain.

current:
https://github.com/zencart/zencart/blob/5ee60ab898edc25e234e64f89516529b0d70af0b/includes/index_filters/default_filter.php#L31-L90

proposed (but untested):
```
$listing_sql = "SELECT " . $select_column_list . " p.products_id, p.products_type, p.master_categories_id, p.manufacturers_id, p.products_price, p.products_tax_class_id, pd.products_description,
                       IF(s.status = 1, s.specials_new_products_price, NULL) AS specials_new_products_price,
                       IF(s.status = 1, s.specials_new_products_price, p.products_price) AS final_price,
                       p.products_sort_order, p.product_is_call, p.product_is_always_free_shipping, p.products_qty_box_status
                FROM " . TABLE_PRODUCTS . " p
                LEFT JOIN " . TABLE_SPECIALS . " s ON s.products_id = p.products_id
                LEFT JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON pd.products_id = p.products_id
                  AND pd.language_id = " . (int)$_SESSION['languages_id'] . "
                LEFT JOIN " . TABLE_MANUFACTURERS . " m ON m.manufacturers_id = p.manufacturers_id
                LEFT JOIN " . TABLE_PRODUCTS_TO_CATEGORIES . " p2c ON p2c.products_id = p.products_id
                WHERE p.products_status = 1
                " . (isset($_GET['manufacturers_id']) && $_GET['manufacturers_id'] != '' ? " AND m.manufacturers_id = " . (int)$_GET['manufacturers_id'] : "") . "
                " . (isset($_GET['filter_id']) && zen_not_null($_GET['filter_id']) ? " AND p2c.categories_id = " . (int)$_GET['filter_id'] : "") . "
                " . $alpha_sort;
```
If interested, I will make a PR for this, After I have tested this single query solution.

drbyte:
Great idea!

Keep in mind that if people have made customizations to these queries, we don't want to break those. Granted, their changes "should" be in an override subdir, but we still need to honour the same $_GET and other variables (and CONSTANT_NAMEs if there were any) for consistency, and retain the same functionality/features/flexibility.

Zen4All:
[...]

PR #2298, opened 2019-04-24
Reworked index_filters

Zen4All:
consolidated 4 queries in to 1, and only put the difference in an if statement.
modernised the rest of the code, like while to each conversion
see #2282, and #2293

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2298)


scottcwilson:
This change should be held until after the 1.5.6b release to prevent further delay.

PR #2305, opened 2019-04-26
Cleanup json_encode calls in VersionServer.php

dennisns7d:
In VersionServer.php there are two calls to json_encode that pass a boolean value of true as the options parameter. The documentation for json_encode on php.net specify that the options parameter is of type integer that specifies a bitmask. It's not clear to me the effects, if any, of passing a boolean value.

Removing the options parameter from these two calls provides compatibility with PHP 5.2 since this parameter was added in PHP 5.3. I found this necessary when upgrading Zen Cart from 1.3.9h (modified to run on PHP 5.2) to 1.5.5f.

### DIFF
--- a/admin/includes/classes/VersionServer.php
+++ b/admin/includes/classes/VersionServer.php
@@ -98,9 +98,9 @@ class VersionServer
 
     protected function buildCurrentInfo()
     {
-        $systemInfo = json_encode(zen_get_system_information(true), true);
+        $systemInfo = json_encode(zen_get_system_information(true));
 
-        $moduleInfo = json_encode($this->getModuleInfo(), true);
+        $moduleInfo = json_encode($this->getModuleInfo());
 
         $country_iso = $this->getCountryIso();
 


drbyte:
> I found this necessary when upgrading Zen Cart from 1.3.9h (modified to run on PHP 5.2) to 1.5.5f.

Do you mean to say "in my temporary staging area, running PHP 5.2, the versionserver checks were failing"? 
If so, was it a showstopper because of fatal errors? Or just a minor inconvenience that the version-check response was unintelligible ... but is moot because it was a temporary staging server, and after the upgrading the PHP version would be upgraded to a newer version and then all the inconvenience symptoms would go away?

dennisns7d:
I encountered fatal errors in my PHP 5.2 staging area concerning the options parameter that PHP 5.2 didn't know about.

It also bothers me that a boolean value is being passed to an integer parameter.

I saw no difference in the versionserver results with PHP 5.6 whether the options parameter was present or not.

zcwilt:
Hi @dennisns7d 
Going to guess that the json_encode calls were based on ( or copy/pasted ) from a json_decode call, where the use of the `true` parameter is acceptable. Assuming the `true` value is converted to an integer would mean it would set the JSON_HEX_TAG option. As such  I can't see any reason not to remove the option, not so much to allow backward compatibility, but just that the use as you have pointed out, is incorrect.

PR #2306, opened 2019-04-26
Make logo for admin login use the one defined for rest of admin

dennisns7d:
Instead of using a hardwired image for admin login and admin password forgotten pages, use the logo configured for the regular admin pages.


drbyte:
Thanks!

PR #2329, opened 2019-05-06
Admin orders validation fixes

torvista:
orders.php is a page I heavily customise.
I rely on the browser and W3C validator to spit out my errors (why doesn't everyone do that???)...but when there are 365 warnings and 556 errors in the vanilla file, I can't see the wood for the trees.
eg:
Clipboard05
These fixes reduce things to one warning. 
Sorry if my timetable does not suit the release schedule, but better late than never.

I committed each file separately to add more detailed information per file.


admin/orders.php
```diff
@@ -352,16 +352,16 @@ function couponpopupWindow(url) {
             <div class="form-group col-xs-4 col-sm-3 col-md-3 col-lg-3">
                 <?php
                 echo zen_draw_form('search', FILENAME_ORDERS, '', 'get', '', true);
-                echo zen_draw_label(HEADING_TITLE_SEARCH_ALL, 'search', 'class="sr-only"');
+                echo zen_draw_label(HEADING_TITLE_SEARCH_ALL, 'searchAll', 'class="sr-only"');
```
Zen4All: Shouldn't this stay "search"? as it refers to the name of the input field?

admin/orders.php
```diff
@@ -721,43 +721,37 @@ function couponpopupWindow(url) {
...
             <div class="form-group">
                 <?php echo zen_draw_label(ENTRY_STATUS, 'status', 'class="col-sm-3 control-label"'); ?>
               <div class="col-sm-9">
-                  <?php echo zen_draw_pull_down_menu('status', $orders_statuses, $order->info['orders_status'], 'class="form-control"'); ?>
+                  <?php echo zen_draw_pull_down_menu('status', $orders_statuses, $order->info['orders_status'], 'id="status" class="form-control"'); ?>
               </div>
             </div>
             <div class="form-group">
-                <?php echo zen_draw_label(ENTRY_NOTIFY_CUSTOMER, 'notify', 'class="col-sm-3 control-label"'); ?>
+                <div class="col-sm-3 control-label" style="font-weight: 700;"><?php echo ENTRY_NOTIFY_CUSTOMER; ?></div>
```
Zen4All: Why remove the label, and introduce a style="" ?

admin/orders.php
```diff
@@ -968,13 +962,13 @@ function couponpopupWindow(url) {
 
                     $show_payment_type = $orders->fields['payment_module_code'] . '<br>' . $orders->fields['shipping_module_code'];
                     ?>
-                <td class="dataTableContent" align="right"><?php echo $show_difference . $orders->fields['orders_id']; ?></td>
-                <td class="dataTableContent" align="left" width="50"><?php echo $show_payment_type; ?></td>
+                <td class="dataTableContent right"><?php echo $show_difference . $orders->fields['orders_id']; ?></td>
```
Zen4All: this should be
```
<td class="dataTableContent text-right">
```

admin/includes/header.php
```diff
@@ -130,7 +130,7 @@
     // display the "check for updated version" button.  The button link should be the current page and all params
     $url = zen_href_link(basename($PHP_SELF), zen_get_all_get_params(array('vcheck'), 'SSL'));
     $url .= (strpos($url, '?') > 5 ? '&' : '?') . 'vcheck=yes';
-    if ($zv_db_patch_ok == true || $version_check_sysinfo == true) $new_version .= '<a href="' . $url . '">' . '<input type="button" class="btn btn-link" value="' . TEXT_VERSION_CHECK_BUTTON . '"/></a>';
+    if ($zv_db_patch_ok == true || $version_check_sysinfo == true) $new_version .= '<a href="' . $url . '">' . TEXT_VERSION_CHECK_BUTTON . '</a>';
```
Zen4All: Why not leave it as a button
```
<a href="' . $url . '" role="button" class="btn btn-link">' . TEXT_VERSION_CHECK_BUTTON . '</a>
```

admin/includes/functions/html_output.php
```diff
@@ -113,13 +113,14 @@ function zen_image_submit($image, $alt = '', $parameters = '') {
...
-    return zen_image(DIR_WS_IMAGES . 'pixel_black.gif', '', '100%', '1');
+    return zen_image(DIR_WS_IMAGES . 'pixel_black.gif', '', '', '1', 'style="width:100%;"');
   }
 
 ////
 // Output a separator either through whitespace, or with an image
   function zen_draw_separator($image = 'pixel_black.gif', $width = '100%', $height = '1') {
-    return zen_image(DIR_WS_IMAGES . $image, '', $width, $height);
+      if (substr(rtrim($width), -1) != "%") $width = $width . 'px';
+    return zen_image(DIR_WS_IMAGES . $image, '', '', $height, 'style="width:' . $width . ';"');
```
Zen4All: I think when using style, width should also say px after the value

admin/orders.php
torvista: zen_draw_label make a **Label** with a **For** pointing at an input with id=search.
That does not (currently) exist so the For is invalid.
If you make the corresponding input have a corresponding id=search, that duplicates the name of the form, so not valid.
Hence I used a new value searchAll for the Label-For input and used that as the id in the corresponding input.
When the form is submitted, the value of the input is still passed using the existing name value of **search** as before, so it still works.
Having a changed id would affect only css and javascript things, not the form.

admin/orders.php
torvista: This is defined as a label, but a label can be linked to only one input.
In this case it is covering three radio buttons so it cannot be a label for three inputs, it is really only a title for the group. Adding the style to make it bold as all the rest is messy I know, but the bold attribute is applied to the other in that section through the label tag, so I didn't see another way to make it bold apart from copying that particular style.

admin/orders.php
torvista: This is in the admin stylesheet
.right {
  text-align: right;
}

text-right is part of bootstrap. Is that preferred?

admin/includes/header.php
torvista: you cannot nest an input inside an <a tag
And it is only a text link.

admin/includes/functions/html_output.php
torvista: for HTML5 only pixels are allowed, so no units are needed.
Other pages are calling this function with either a percentage or with no unit at all, so making the change in the function means the calling pages do not need to change.

admin/orders.php
Zen4All: That's what I used in the rest of the pages

torvista: ok, I've modified them.

admin/includes/functions/html_output.php
Zen4All: ok, wasn't sure so that's why I asked

admin/includes/header.php
Zen4All: You are right about the nesting, but look at the line I posted above, The class makes a button with css.

admin/orders.php
Zen4All: That sounds sensible 😃

admin/includes/header.php
torvista: That validates. 
I didn't bother to maintain it as a button as it is not formatted as a button, just as a link that underlines, didn't think about the accessibility aspect.
Corrected.

admin/includes/header_navigation.php
```diff
@@ -11,14 +11,14 @@
...
   <!-- Brand and toggle get grouped for better mobile display -->
   <div class="navbar-header">
     <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-adm1-collapse">
       <span class="sr-only">Toggle navigation</span>
-      <span class="icon-bar"></span>
-      <span class="icon-bar"></span>
-      <span class="icon-bar"></span>
+      <span class="icon-bar">&nbsp;</span>
+      <span class="icon-bar">&nbsp;</span>
+      <span class="icon-bar">&nbsp;</span>
```
drbyte: Why all the `&nbsp;` in several files of this PR?
Is it because the markup is invalid without them?
Or is it personal preference against having nothing inside the `span` or `i`/`icon` elements?

torvista:
Purely because the validator complained, I added those in the interests of having nothing at all in the list of "errors".
If it had made a difference to the aspect, I would left them empty as obiously they are not errors, but the spaces don't appear to make any visible difference so I put them in.

drbyte:
> Purely because the validator complained, I added those in the interests of having nothing at all in the list of "errors".
> If it had made a difference to the aspect, I would left them empty as obiously they are not errors, but the spaces don't appear to make any visible difference so I put them in.

Meh. I agree the display impact is minimal. But I disagree with the stupid validator in this case.
I'll allow it this time, but that's not permission to go and add a zillion `&nbsp` all over the code in other files just for the sake of the validator (which again is wrong IMO).

Thanks for your work on this PR.

PR #1518, opened 2017-09-24
Support ISO-8601 style date-of-birth date (merge #1515)

drbyte:
Thanks @mc12345678

PR #2337, opened 2019-05-13
v156 account_edit recognize ISO-8601 DOB

mc12345678:
Similar to #1518, but for account edit.  If using a mobile device, the date is presented with a date selector (though still haven't been able to get the date selector to populate with the known/existing date) that stores the date locally/internally using ISO-8601 format which in the previous format would prevent updating the record.  This at least allows updating the record, though it seems that the user would have to re-enter the birthdate.

It appears that in jscript_responsive_framework.php the line associated with the dob:
`$('input#dob').clone().attr('type','date').insertAfter('input#dob').prev().remove();`

causes the attribute of `value` to get cleared once the type is modified to date. In briefly trying to address this in light of upcoming distribution, I haven't come up with a "one-liner" to resolve this.  May require more than one line in that file. That though is basically a separate issue. :/

### DIFF
--- a/includes/languages/english.php
+++ b/includes/languages/english.php
@@ -189,8 +189,8 @@
   define('ENTRY_LAST_NAME_ERROR', 'Is your last name correct? Our system requires a minimum of ' . ENTRY_LAST_NAME_MIN_LENGTH . ' characters. Please try again.');
   define('ENTRY_LAST_NAME_TEXT', '*');
   define('ENTRY_DATE_OF_BIRTH', 'Date of Birth:');
-  define('ENTRY_DATE_OF_BIRTH_ERROR', 'Is your birth date correct? Our system requires the date in this format: MM/DD/YYYY (eg 05/21/1970)');
-  define('ENTRY_DATE_OF_BIRTH_TEXT', '* (eg. 05/21/1970)');
+  define('ENTRY_DATE_OF_BIRTH_ERROR', 'Is your birth date correct? Our system requires the date in this format: MM/DD/YYYY (eg 05/21/1970) or this format: YYYY-MM-DD (eg 1970-05-21)');
+  define('ENTRY_DATE_OF_BIRTH_TEXT', '* (eg. 05/21/1970 or 1970-05-21)');
   define('ENTRY_EMAIL_ADDRESS', 'Email Address:');
   define('ENTRY_EMAIL_ADDRESS_ERROR', 'Is your email address correct? It should contain at least ' . ENTRY_EMAIL_ADDRESS_MIN_LENGTH . ' characters. Please try again.');
   define('ENTRY_EMAIL_ADDRESS_CHECK_ERROR', 'Sorry, my system does not understand your email address. Please try again.');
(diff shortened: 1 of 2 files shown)


mc12345678:
Figured it out. That is, I figured out why on mobile devices or if the `dob` field is of type `date` that the date selector is not pre-populated. The question becomes, what direction(s) should the solution(s) take. I'll try to explain.

It appears that a date selector which is javascript/jquery based expects the date to be in ISO-8601 format. As discussed in the chain of #1518 (see also #1515) there are four potential formats that such a date can be formatted. YYYYMMDD, YYYY-MM-DD, YYYY/MM/DD, and I believe YYMMDD (this fourth one I have not attempted to address in my previous incorporation of ISO-8601 formatting because of the potential confusion in what the three sets of values truly mean). So for the date selector to be populated, the value in the field should appear in one of those formats at the onset.

OK, so requirement one, field's value should be in what might be considered a raw format, or at least formatted slightly from the raw format.

Next, and this is the "weird" thing. Regardless a value provided to `zen_draw_input_field` if a variable exists with the same "name" as the field being generated, the variable is a string and perhaps a few other things I have forgotten at the moment, then that variable's value will be used as the variable for the input field.

In this case the variable is created/used in `includes/modules/pages/account_edit/header_php.php` with final assignment at the bottom. For initial page load, `$dob` isn't really set, so the value is blank and basically results in "prepopulating" the field as empty text. If, however, a value is submitted, but there is an issue with some other aspect of saving the record, then `$dob` will be populated with the entry that was made. 

So, to prevent the field from being empty because of the use/assignment of `$dob`, the date from the user's record needs to be assigned.

OK, no big deal, right? Just determine if `$dob` needs to be set to `customers_dob`, but what format should that date take? Should it be determined/set using the format currently expected by the form? (forgot to mention, the current expectation is MM/DD/YYYY or so one is alerted prior to the modification of accepting most of the ISO-8601 formats).

If the default displayed format as requested is presented, then the date is not in the format for the browser's handling of a `date` field. If it is shown in ISO-8601 format, then if/when a date selector tries to display the value it will work, but if such is not implemented (standard desktop or tablet format) then the date is presented differently than "required".

Now a general user, likely won't realize the discrepancy because it is only when the field is blank that there is any "help" text, but if they decide to change the value and enter it incorrectly, then a message is introduced which may then cause confusion. The message expects the date in one format, but the data was presented in a different format.

So.... whew, made it to here. Thoughts are these:
1 Date of birth gets removed, no further issues, but doesn't really solve any overall date "type" usage.
2 Message is modified to expand on the allowable date formats.
3 the date is prepopulated using ISO-8601 format.
4 the date is prepopulated using the existing "required" format, but in the jquery process of creating the date type field, the value is converted from the existing short date, to an ISO-8601 format. Problem I see here, may be that where those that use DD/MM/YYYY format, this "subtle" change/calculation may be missed when modifying the store files. This jquery script doesn't use php, but instead references the existing page/form data.
5 the assignment `$dob` depends on the current layout invoked, to set it to ISO-8601 as necessary to support the jquery involved. Problem with this is maintaining the relationship(s) between the various locations/data. If one updates the jquery to apply the `date` field to other display types, then must also modify the output format of `$dob`. As far as customer experience, this seems to be the best, but is one of the more complex solutions and makes management/maintenance difficult.

I'm thinking at least item 3, possibly with 2 as the solution that is easiest for those maintaining a Zen Cart install with recommendation of 4 possibly 5 being implemented by those wishing to have a more consistent solution, though with noting at least the issues identified with each of those solutions.

PR #2375, opened 2019-05-30
Ensure inner loop variable names are unique

scottcwilson:


drbyte:
This isn't necessary. The foreach resets those variables anyway. There's no clash or mis-use.

scottcwilson:
Line 85 creates `$key` and `$value`.  If you use these variables accidentally after line 277, you could introduce a bug.  I know it's unlikely but I have actually been bitten by this.   It's more to make the code more self-documenting than to truly fix an issue.

PR #2376, opened 2019-05-30
Remove some commented code

scottcwilson:
Remove old code or put bracket on new line if old code needed for
reference.  This way, bracket matching will work.

### DIFF
--- a/includes/modules/order_total/ot_coupon.php
+++ b/includes/modules/order_total/ot_coupon.php
@@ -494,15 +494,16 @@ class ot_coupon {
       }
 //echo 'ot_coupon coupon_total: ' . $coupon->fields['coupon_calc_base'] . '<br>$orderTotalDetails[orderTotal]: ' . $orderTotalDetails['orderTotal'] . '<br>$orderTotalDetails[totalFull]: ' . $orderTotalDetails['totalFull'] . '<br>$coupon_total: ' . $coupon_total . '<br><br>$coupon->fields[coupon_minimum_order]: ' . $coupon->fields['coupon_minimum_order'] . '<br>$coupon_total_minimum: ' . $coupon_total_minimum . '<br>';
 // @@TODO - adjust all Totals to use $coupon_total but strong review for what total applies where for Percentage, Amount, etc.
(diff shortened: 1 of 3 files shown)


includes/modules/order_total/ot_coupon.php
```diff
@@ -494,15 +494,16 @@ function calculate_deductions() {
       }
 //echo 'ot_coupon coupon_total: ' . $coupon->fields['coupon_calc_base'] . '<br>$orderTotalDetails[orderTotal]: ' . $orderTotalDetails['orderTotal'] . '<br>$orderTotalDetails[totalFull]: ' . $orderTotalDetails['totalFull'] . '<br>$coupon_total: ' . $coupon_total . '<br><br>$coupon->fields[coupon_minimum_order]: ' . $coupon->fields['coupon_minimum_order'] . '<br>$coupon_total_minimum: ' . $coupon_total_minimum . '<br>';
 // @@TODO - adjust all Totals to use $coupon_total but strong review for what total applies where for Percentage, Amount, etc.
-//      if ($coupon->RecordCount() > 0 && $orderTotalDetails['totalFull'] != 0) {
-      if ($coupon->RecordCount() > 0 && $orderTotalDetails['orderTotal'] != 0 ) {
+//      if ($coupon->RecordCount() > 0 && $orderTotalDetails['totalFull'] != 0)
+      if ($coupon->RecordCount() > 0 && $orderTotalDetails['orderTotal'] != 0 ) 
+      {
```
zcwilt: Our current coding standards (psr2) state that opening braces for if statements should be on the same line. Given you didn't move the brace in `attributes.php` was it intentional here.

scottcwilson: My thinking was to move it down in places where there are commented out lines that may be preferred to the current uncommented out line.  But I could move it back to the current live line if preferred; people would just have to remember to add it  to the commented out line if they uncomment it.

zcwilt: Yes, would move it back. I guess as well I'm not a great fan of putting alternate code into comments any way.

Issue #2379, opened 2019-05-30
Defines in main language file should not be redefined in includes/languages/english/<page>.php


scottcwilson:
./languages/english.php:  define('TABLE_HEADING_QUANTITY', 'Qty.');
./languages/english/advanced_search.php:  define('TABLE_HEADING_QUANTITY', 'Quantity');
./languages/english/advanced_search_result.php:define('TABLE_HEADING_QUANTITY', 'Quantity');

The second and third defines will fail because the first one is done so early (when application_top is included.) 

This define needs to be removed from files in the english folder and only defined once in the main language file.

mc12345678:
This issue makes me wonder why the language system is currently designed to not offer true override capability? Currently, if it is desired for a page to use a different definition than what is provided in the main language file, then the main language file must be modified along with the language file for all other pages that use/might use the define that was to be overridden or the template file for the associated page and language file for that file needs to be modified to use a different define that is then added to the page's language file(s).

Basically, as is anything in `includes/languages/LANGUAGE.php` can only be "overridden" by `includes/languages/YOUR_TEMPLATE/LANGUAGE.php` 

but not by 
`includes/languages/LANGUAGE/YOUR_PAGE.php` nor `includes/languages/LANGUAGE/YOUR_TEMPLATE/YOUR_PAGE.php`

Is there a technical reason for this? Is it expected to change in a future significant release (I can't say that I saw anything in the develop branch that seemed to modify that approach)?

scottcwilson:
Fair question - I am guessing this is a legacy constraint.  Right now the main language file is loaded first, so it takes priority.  Changing this could break things for upgraders - not sure it's wise.  I have been arguing for just consolidating these defines.  Do you REALLY need different values for a string like Quantity?

drbyte:
> Do you REALLY need different values for a string like Quantity?

Well, when a column heading is being shortened to just `Qty` to not waste horizontal space, yes.

torvista:
Why not just have:
> ./languages/english.php: define('TABLE_HEADING_QTY', 'Qty.');
> ./languages/english.php: define('TABLE_HEADING_QUANTITY', 'Quantity');

and remove both defines from the page-specific lang file.

If a busy table uses the abbreviation out of the box, there will be no need to change it.
If a sparse table using the complete heading gets modifed such that it needs the abbreviation...well so what, it gets modified in the script anyway.

lankeeyankee:
[...]

Issue #2008, opened 2019-01-01
v2: reduce duplication in language defines


scottcwilson:
For consideration in V2 (per suggestion in #2003).

Which things should we retain duplicates for (and why)
Which things should be de-duplicated, etc.

scottcwilson:
Did you truly intend to close?  That was only a one-issue fix - much more remains.

PR #2380, opened 2019-05-30
Remove TABLE_HEADING_QUANTITY override

scottcwilson:
This value is defined in english.php so the override won't work

### DIFF
--- a/includes/languages/english/account.php
+++ b/includes/languages/english/account.php
@@ -17,7 +17,6 @@ define('TABLE_HEADING_DATE', 'Date');
 define('TABLE_HEADING_ORDER_NUMBER', 'No.');
 define('TABLE_HEADING_SHIPPED_TO', 'Ship To');
 define('TABLE_HEADING_STATUS', 'Status');
-define('TABLE_HEADING_TOTAL', 'Total');
 define('TABLE_HEADING_VIEW', 'View');
 
 define('MY_ACCOUNT_TITLE', 'My Account');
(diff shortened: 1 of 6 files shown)


scottcwilson:
Fixes #2379

includes/languages/english/account.php
```diff
@@ -17,7 +17,6 @@
 define('TABLE_HEADING_ORDER_NUMBER', 'No.');
 define('TABLE_HEADING_SHIPPED_TO', 'Ship To');
 define('TABLE_HEADING_STATUS', 'Status');
-define('TABLE_HEADING_TOTAL', 'Total');
```
zcwilt: Was this intentional

scottcwilson: Yes, same issue with that define.  But I can break it out if you prefer.

zcwilt:
looking at your issue #2379 I thought the intention was to remove the defines from the main language file.

scottcwilson:
I thought about doing it that way but realized the other way around was a better path.

zcwilt:
However it does mean that there is now no way to override those defines on other pages.

scottcwilson:
In the spirit of  issue #2008 I would prefer to move this define to the top and reduce duplication.  If a different string is truly required, people can create a new define. If we were to move this define out of the top level, it would necessitate creating it at a number of new places which require the top level define.

scottcwilson:
I think this approach - consolidating string definitions and assuming they only have one value throughout the site - will make upgrades easier and faster for people.

PR #2382, opened 2019-05-31
Don't specify countries_id on INSERT

scottcwilson:
This guy's point is well taken: 
https://www.zen-cart.com/showthread.php?224820-Database-upgrade-fails-when-upgrading-my-1-5-3-to-1-5-6&p=1355473#post1355473

Suggest merging this into 156 (in case there's a 156c patch) as well as 1.5.7.

### DIFF
--- a/zc_install/sql/updates/mysql_upgrade_zencart_153.sql
+++ b/zc_install/sql/updates/mysql_upgrade_zencart_153.sql
@@ -43,7 +43,7 @@ UPDATE configuration set configuration_group_id = 6 where configuration_key in (
 #ISO Updates:
 UPDATE countries SET countries_name = 'Libya' WHERE countries_iso_code_3 = 'LBY';
 UPDATE countries SET countries_name = 'Palestine, State of' WHERE countries_iso_code_3 = 'PSE';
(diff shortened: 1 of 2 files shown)


mc12345678:
If `countries_id` is not specified, then there will not be a matchup in other areas of the `countries_id` to the expected value(s). Although I believe I've seen discussion of "relaxing" the direct assignment of a country to an id, it appears that it would involve significant code revision.

Additionally, this "duplication" of attempting to store the same value back to the countries list has not proven to actually stop the install or prevent it from completing.

As discussed further in the referenced thread, the issue is data in a field in the `orders` table that was incorrectly populated at one point in time and needed to be updated as discussed in the forum thread for the plugin that added the field and in a previous version of the software incorrectly populated the database with a date value of '0000-00-00'. The current version of that software no longer does that.

scottcwilson:
There is no matchup required at this point since there are no states/provinces for the newly added countries.  And even if there were, the id could be deduced by a query on the ISO value.

The orders table issue you are pointing out is a different problem.  Multiple issues were reported by the OP.

zcwilt:
As this only relates to upgrading, I can't see any problem.

PR #2383, opened 2019-05-31
Product Copy: add metatags, linked-categories etc.

torvista:
1) remove hardcoded language texts from function zen_copy_products_attributes to english.php
2) remove duplicated constants from attributes_controller and category_product_listing to english.php
3) Product Copy infobox format changes, remove invalid labels for two radio buttons and invalid widths for lines.
4) Product Copy: added metatags (if defined), lined categories (if defined) to already-existing Discounts (if defined) and Attributes (if defined).
5) Simplification of copy_product_confirm.

Info Box:
copy_to_new
Messages:
copy_to_new-messages


torvista:
Sigh....no there are not four commits...I still get frustrated by the inner-workings of Git and SmartGit and how to get things synchronised before making modifications....

drbyte:
Is there a particular reason why the other product-type files weren't included?

torvista:
I am trying to make each upgrade less tedious, with more of my personal mods in the core where appropriate! That's the only way I can even try and justify the stupid amount of time it takes me to prepare this stuff for public consumption.
Since I don't use any other product types and I assume 
a) this can be duplicated easily if someone needs it, 
b) that duplication will be removed in the future anyway,
so, not done.

drbyte:
> Since I don't use any other product types

Fair enough. That's why I was asking.

> That's the only way I can even try and justify the stupid amount of time it takes me to prepare this stuff for public consumption.

Understood. And people wonder why releases aren't done monthly!

torvista:
...I have left the formatting consistent with the rest as it is for core inclusion, but I think it takes up too much vertical space.
In fact I think the whole admin template wastes a lot of space: I have an override template to get it to **look like** an admin screen (info and as much detail as possible in the screen) and not infantile-size text with big spaces for slow-reading customers.

drbyte:
Files added, and merged.

PR #2385, opened 2019-05-31
Simplify admin tab name logic

scottcwilson:


drbyte:
In english, what's the new logic here?

scottcwilson:
Remove the need to check TEXT_ADMIN_TAB_PREFIX before setting the title string.

drbyte:
Okay.
What's the problem/symptom that this is solving from the end-user perspective?

scottcwilson:
The issue is that if they switch to the second define of TEXT_ADMIN_TAB_PREFIX (which people with multiple stores would want), you'd need to add code because of the additional check I removed. with my update, you can use either define for TEXT_ADMIN_TAB_PREFIX with no additional changes.

drbyte:
Okay, I found the old PR where the code you're removing was initially added:
https://github.com/zencart/zencart/pull/2061/files

PR #2389, opened 2019-05-31
Fix delete music genre entry in MySQL 5.7

scottcwilson:
Same fix as PR #2386


drbyte:
Yes, would prefer to leave out the quotes. In all of those PRs.

PR #2398, opened 2019-06-02
Remove references to missing images in banner_tools.css

scottcwilson:
The img/ folder doesn't exist, nor do the referenced images


drbyte:
Elsewhere you're championing removing comments. Why did you choose to comment these instead of removing them?

scottcwilson:
I am open to removing them but I thought perhaps some templates included these images.

PR #2401, opened 2019-06-03
Remove old-style javascript comments

Zen4All:
Most javascripts contained very old style comments to hide javascript from ancient browsers like ie6, because they could not handle javascript.
Modern browsers all know how to handle javascript. This will also make coding and debugging easier using an IDE.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2401)

### DIFF
--- a/admin/admin_account.php
+++ b/admin/admin_account.php
@@ -88,10 +88,9 @@ $userDetails = $userList[0];
 <link rel="stylesheet" type="text/css" href="includes/stylesheet.css">
 <link rel="stylesheet" type="text/css" href="includes/cssjsmenuhover.css" media="all" id="hoverJS">
 <link rel="stylesheet" type="text/css" href="includes/admin_access.css" />
-<script language="javascript" src="includes/menu.js"></script>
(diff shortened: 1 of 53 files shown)


torvista:
The w3C validator points out that  type="text/javascript" is not required anymore, why not get rid of those too?

Zen4All:
@torvista 
If I remember correctly, that only is valid for html5, but I will look it up to be sure.
Also this PR is purely meant to get rid of the comments. There is more things that can be updated, but I do not want to mix to much different changes if it can be avoided (this also makes it easier for other people to understand/review).

torvista:
> only is valid for html5
Is not everything now for html5?

Zen4All:
Nope,
- template default => ```<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">```
- responsive classic => html5
- A lot of custom templates are still on html4

So we can not assume html 5.

PR #2402, opened 2019-06-03
auto update line endings for edited files

Zen4All:
This one line will auto-update all edited files to use a single type of  ending (LF), making comparing files easier

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2402)


drbyte:
I hope this doesn't bite us by making upgrades harder, not easier ... since lots of peoples' files are using the (wrong) CRLF ending.

PR #2405, opened 2019-06-03
zen_debug_error_handler return value

scottcwilson:


drbyte:
While I agree that if a function is going to return a value anywhere, then it should always return a value, isn't this change going to alter how the error handler works? Returning `false` causes a fallback to PHP doing its own handling. Doesn't `returning no value` intercept and terminate processing?

scottcwilson:
Example 1 looks like this very case:

https://www.php.net/manual/en/function.set-error-handler.php

PR #2408, opened 2019-06-04
Supports required file uploads.

### LINKED ISSUE #2407: v157 (and others in past) No way to require a file to be uploaded
Discussed at https://www.zen-cart.com/showthread.php?225489-Setting-up-a-product-with-a-single-required-file-upload-attribute as originated by @lat9.

mc12345678:
As identified in the Zen Cart forum:
https://www.zen-cart.com/showthread.php?225489-Setting-up-a-product-with-a-single-required-file-upload-attribute
This offers a way to honor marking an File attribute as required.

<strike>This does not yet address the option name being identified as
Text when it is a file attribute.</strike> <- Addressed by this [commit](https://github.com/zencart/zencart/pull/2408/commits/0bc618a3a2a34de95545f03a9c71200e5413bcf1).

Note, if two or more files are marked as required that this
will "upload" all files identified; however, if any required
file is not uploaded, then none of the uploaded files get
attached to a store product and would need to be uploaded again.
Not sure yet if/how to address this, but maybe another will be
able to help or "accept" that issue.

Partially fixes #2407

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -1959,7 +1959,7 @@ class shoppingCart extends base {
         } else {
           // process normally
           // iii 030813 added: File uploading: save uploaded files with unique file names
-          $real_ids = isset($_POST['id']) ? $_POST['id'] : "";
+          $real_ids = isset($_POST['id']) ? $_POST['id'] : array();
           if (isset($_GET['number_of_uploads']) && $_GET['number_of_uploads'] > 0) {
             /**
              * Need the upload class for attribute type that allows user uploads.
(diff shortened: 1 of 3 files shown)


includes/classes/shopping_cart.php
```diff
@@ -1959,15 +1959,15 @@ function actionAddProduct($goto, $parameters) {
...
+          $real_ids = isset($_POST['id']) ? $_POST['id'] : array();
           if (isset($_GET['number_of_uploads']) && $_GET['number_of_uploads'] > 0) {
             /**
              * Need the upload class for attribute type that allows user uploads.
              *
              */
             include(DIR_WS_CLASSES . 'upload.php');
             for ($i = 1, $n = $_GET['number_of_uploads']; $i <= $n; $i++) {
-              if (zen_not_null($_FILES['id']['tmp_name'][TEXT_PREFIX . $_POST[UPLOAD_PREFIX . $i]]) and ($_FILES['id']['tmp_name'][TEXT_PREFIX . $_POST[UPLOAD_PREFIX . $i]] != 'none')) {
+              if (isset($_POST[UPLOAD_PREFIX . $i]) && isset($_FILES['id']['tmp_name'][TEXT_PREFIX . $_POST[UPLOAD_PREFIX . $i]]) && zen_not_null($_FILES['id']['tmp_name'][TEXT_PREFIX . $_POST[UPLOAD_PREFIX . $i]]) && (!isset($_POST[UPLOAD_PREFIX . $i]) || !isset($_FILES['id']['tmp_name'][TEXT_PREFIX . $_POST[UPLOAD_PREFIX . $i]]) || ($_FILES['id']['tmp_name'][TEXT_PREFIX . $_POST[UPLOAD_PREFIX . $i]] != 'none'))) {
```
drbyte: Can `isset($_FILES['id']['tmp_name'][TEXT_PREFIX . $_POST[UPLOAD_PREFIX . $i]]) && zen_not_null($_FILES['id']['tmp_name'][TEXT_PREFIX . $_POST[UPLOAD_PREFIX . $i]])` 
be shortened with just `!empty` instead of `isset && zen_not_null`?

mc12345678: If the simple assumption is that all files will have some sort of non-empty extension even for a file named 0, then yes could substitute with `!empty`.  At the time of writing I had briefly considered it as done in other similar rewrites, but opted to leave as seen, awaiting further discussion, potential testing for that case and evaluation.

By the way, at the time of writing, I was unaware of a plugin that performs a similar function with additional "notifications".  I have not fully reviewed that code, but would not think that it should be discounted potentially as a more complete solution.

PR #2435, opened 2019-06-11
Remove un-needed language files from responsive_classic

scottcwilson:


drbyte:
But ... the files are different, specifically pointing to different image assets.

That's why this wasn't merged the last time you submitted it.

scottcwilson:
They don't *have* to be different is my point. Some of this can be simplified out.

drbyte:
Correct: no overrides "have to" be used.

But if you delete these defined overrides then you force the default install (which does use responsive_classic) to use the wrong image assets (the gifs from template_default, instead of the pngs from responsive_classic).

Is that what you're wanting? Why?

scottcwilson:
Which file are you specifically talking about?

drbyte:
The images I'm talking about are the ones referenced in the files you're deleting.

Here's a visual comparison of the files you're proposing to remove. I kinda assumed you had done the compare before submitting the PR, so you knew what the impact was.

Screen Shot 2019-06-11 at 15 01 14
Screen Shot 2019-06-11 at 15 00 41
Screen Shot 2019-06-11 at 14 59 46
Screen Shot 2019-06-11 at 14 58 42

scottcwilson:
How about consolidating the specific changes so multiple language files don't have to be updated and merged just because a template changes some images?  See new file `includes/languages/english/<template_name>.php`.

scottcwilson:
Got carried away in cleanup.

drbyte:
Q: Will this approach now trigger a lot more `Constant already defined` notices?

scottcwilson:
*Fewer* because the number of duplicated defines is reduced.

scottcwilson:
With this change, you have three *less* files to maintain, because the ones from template_default are used but the key changes are in a single override file for that template.

drbyte:
Closing in favor of @zcwilt pending plugin updates which augments language loading slightly. De-duplication can be re-addressed later.

PR #2442, opened 2019-06-15
Fatal Error at ipn_postcfg.php

lat9:
As reported in this (https://www.zen-cart.com/showthread.php?220781-One-Page-Checkout-Support-Thread&p=1358543#post1358543) Zen Cart forum posting.

Added that missing **/** to correct the error.

### DIFF
--- a/includes/init_includes/init_ipn_postcfg.php
+++ b/includes/init_includes/init_ipn_postcfg.php
@@ -19,6 +19,6 @@ if (!defined('IS_ADMIN_FLAG')) {
  */
 if (!isset($_SESSION['language'])) $_SESSION['language'] = 'english';
 
-require(zen_get_file_directory(DIR_FS_CATALOG . DIR_WS_LANGUAGES . $_SESSION['language'], 'checkout_process.php', 'false'));
+require(zen_get_file_directory(DIR_FS_CATALOG . DIR_WS_LANGUAGES . $_SESSION['language'], '/checkout_process.php', 'false'));
 
 ?>


scottcwilson:
@drbyte This one should be backported to 1.5.6.

scottcwilson:
I think it warrants a mention in https://www.zen-cart.com/showthread.php?224706-Known-bugs-(and-fixes)-for-v1-5-6-v1-5-6a as well. 

Thanks for fixing @lat9 !

includes/init_includes/init_ipn_postcfg.php
```diff
@@ -19,6 +19,6 @@
  */
 if (!isset($_SESSION['language'])) $_SESSION['language'] = 'english';
 
-require(zen_get_file_directory(DIR_FS_CATALOG . DIR_WS_LANGUAGES . $_SESSION['language'], 'checkout_process.php', 'false'));
+require(zen_get_file_directory(DIR_FS_CATALOG . DIR_WS_LANGUAGES . $_SESSION['language'], '/checkout_process.php', 'false'));
```
mc12345678: The slash really should be after `$_SESSION['language']` instead of before `checkout_process.php` unless the use of a template override is purposefully necessary.

All other properly formatted and operational instances of `zen_get_file_directory` present the folder path with a slash at the end rather than before the filename and the function when so presented supports a template override.

scottcwilson:
See #2467, which should be backported instead of  this change .

PR #2444, opened 2019-06-16
Admin: removal of constants already defined in main language file

torvista:
These all get spat out by Report All Errors.

These are the only ones not fixed.

TEXT_GV_REDEEM
If this is removed from gv_name, then 
admin->Discounts->Mail Gift Certificate gives an error as it is undefined.
admin->Discounts->Gift Certificates sent gives an error as it is undefined.

On the Admin-Payment modules page, there are 100 duplicated paypal constants errors!

The Catalog side needs a lot of TLC in this respect that someone else will need to do!


scottcwilson:
issue #2379 is tracking the catalog side of this issue.

PR #2465, opened 2019-06-19
v157 PHP 7.3 notice of non-variable passed by reference

### LINKED ISSUE #2464: Passing function result by reference in notifier
````php
Request URI: /v157/index.php?main_page=account_edit, IP address: xxx.xxx.xxx.xxx
#1 require(/home/v157/includes/modules/pages/account_edit/header_php.php) called at [/home/v157/index.php:36]
--> PHP Notice: Only variables should be passed by reference in /home/v157/includes/modules/pages/account_edit/header_php.php on line 107.
````

Line 107 of the affected file is a notifier (line after the comment):
````php
    //update external bb system with new email address
    $zco_notifier->notify('NOTIFY_NICK_UPDATE_EMAIL_ADDRESS', $nick, $db->prepareInput($email_address));
````

It appears that the desire was to be able to feed back into a function any changes that may have been made to the prepared input of `$email_address` which wouldn't be possible.

Seeing that the data is to be sanitized "again" in a line of code that follows, it appears that another variable is needed here to receive the assignment, make the call to the notifier, then possibly destroy the variable afterwards? At least in this way users of the notifier would not have to make any changes instead of just having fed that information into an array for the first argument.

Environment of discovery was PHP 7.3.6, strict error reporting on. Was seen in ZC 1.5.6 at one point with PHP 7.2.x, but was not formally identified by PR or otherwise.

mc12345678:
Fixed #2464.

Created a variable specifically to pass the sanitized information to the observer and then deleted the variable.  This maintains the notifier as previously provided while still providing sanitized information to the second parameter (editable by design of the observer class beginning in Zen Cart 1.5.3) but is then unset to not affect the subsequent storage of the original provided `$email_address`.

### DIFF
--- a/includes/modules/pages/account_edit/header_php.php
+++ b/includes/modules/pages/account_edit/header_php.php
@@ -104,10 +104,8 @@ if (isset($_POST['action']) && ($_POST['action'] == 'process')) {
   $zco_notifier->notify('NOTIFY_HEADER_ACCOUNT_EDIT_VERIFY_COMPLETE');
 
   if ($error == false) {
-    //update external bb system with new email address
-    $new_email_address = $db->prepareInput($email_address);
-    $zco_notifier->notify('NOTIFY_NICK_UPDATE_EMAIL_ADDRESS', $nick, $new_email_address);
-    unset($new_email_address);
+    //update external bb system with submitted email address
+    $zco_notifier->notify('NOTIFY_NICK_UPDATE_EMAIL_ADDRESS', $nick, $email_address);
 
     // build array of data to store the requested changes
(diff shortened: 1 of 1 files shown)


drbyte:
Looking at the [original commit](https://github.com/zencart/zencart/pull/360/files#diff-75443be36a052db4f0f14aed5557794dR103), I think the right solution here is just to pass `$email_address` directly.

mc12345678:
Seemed to me that there was a desire to offer the data in as sanitized condition as might be expected and that it was undesirable to offer the ability to change the value within the observer. (Yes, it could still be changed by code in the observer if the file remains in the global space, but I thought that was besides the point.)

PR #2467, opened 2019-06-19
Language inclusion requires slash before filename

scottcwilson:
Recommended for backport to 1.5.6

### DIFF
--- a/includes/init_includes/init_ipn_postcfg.php
+++ b/includes/init_includes/init_ipn_postcfg.php
@@ -19,6 +19,6 @@ if (!defined('IS_ADMIN_FLAG')) {
  */
 if (!isset($_SESSION['language'])) $_SESSION['language'] = 'english';
 
-require(zen_get_file_directory(DIR_FS_CATALOG . DIR_WS_LANGUAGES . $_SESSION['language'], '/checkout_process.php', 'false'));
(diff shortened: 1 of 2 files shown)


scottcwilson:
Right now, order processing stops cold because of this issue.  My fix resolves that.  Yes I agree it would be even better to use proper template handling to fix this but that's not what 1.5.5-1.5.6 do - they just pull the file from the default language location.   Making this inclusion template sensitive would be a great 1.5.7 enhancement, but let's fix the main problem first right now.

mc12345678:
Ummm, it wouldn't be a 1.5.7 "enhancement", it would be a restoration to operation as provided in 1.5.6 and 1.5.6a that was removed in 1.5.6b.
Zen Cart 1.5.6 original code:
````php
          // require the language file
          global $language_page_directory, $template_dir;
          if (!isset($language_page_directory)) $language_page_directory = DIR_WS_LANGUAGES . $_SESSION['language'] . '/';
          if (file_exists($language_page_directory . $template_dir . '/create_account.php')) {
            $template_dir_select = $template_dir . '/';
          } else {
            $template_dir_select = '';
          }
          require($language_page_directory . $template_dir_select . '/create_account.php');
````

scottcwilson:
Yep you're right. Updated to include / in directory specifier.

scottcwilson:
Recommended for backport INSTEAD OF #2442.

PR #2470, opened 2019-06-20
v157 modify pull down menu function

Zen4All:
The current `zen_draw_pull_down_menu()` function only accepts `$values arrays`, starting with an index of 0. In not all cases `$values` will start with 0, resulting in a wrong out put.
By replacing the `for ($i=0, $n=sizeof($values); $i<$n; $i++) {` with a `foreach` the `$values` array can have any value.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2470)

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -300,14 +300,14 @@
   }
 
   /**
-   * Output a form pull down menu
-   * @param string $name name
-   * @param array $values values
-   * @param string $default default value
-   * @param string $parameters parameters
-   * @param boolean $required required
-   * @return string
-   */
+ * Output a form pull down menu
+ * @param string $name name
+ * @param array $values values
+ * @param string $default default value
+ * @param string $parameters parameters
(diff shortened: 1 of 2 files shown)


lat9:
Could you identify a case where the values-array doesn't have numerical indexes?  FWIW, I've used the PHP `array_values` function to re-index numerical arrays when I'm fussing with (i.e. removing) some of the indices.

Zen4All:
It mot about not numerical, but not starting at zero.
like
```
$array
[1]
   ['id'] = 1
   ['text'] = 'first'
[2]
   ['id'] = 2
   ['text'] = 'second'
```

lat9:
Right, and you can use `array_values` to re-index that array so that it then contains indexes of 0 and 1.

Zen4All:
Yes I know, but I do not want to start at 0, I want to be able to start at any number I like

lat9:
OK, then be sure to apply the change to the storefront version of the function, too!

PR #2471, opened 2019-06-20
Ensure MIN  and MAX values from groups 2 and 3 are integer

scottcwilson:
Blocks script injection into fields used in Javascript validation.

### DIFF
--- a/includes/init_includes/init_db_config_read.php
+++ b/includes/init_includes/init_db_config_read.php
@@ -15,13 +15,18 @@ if (!defined('IS_ADMIN_FLAG')) {
 }
 
 $use_cache = (isset($_GET['nocache']) ? false : true ) ;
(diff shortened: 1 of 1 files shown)


scottcwilson:
Recommended for backport to 1.5.6c.

mc12345678:
If this is to become a thing, then would suggest (at some point) using/populating the `val_function` for these constants to have them be integers as well possibly by the validation filter `FILTER_VALIDATE_INT`.

scottcwilson:
I like that idea as a future enhancement, but this specific PR is an "anti-bad-guy" change. They're not using the admin panel - they're modifying the db directly.

wiztechinc:
The fix is reasonable. I would like to know how in the world this is happening to begin with. That to me is the larger issue.

PR #2472, opened 2019-06-20
Remind on folder location

scottcwilson:
Remind people to place backup and download folders above public_html.

### DIFF
--- a/admin/includes/dist-configure.php
+++ b/admin/includes/dist-configure.php
@@ -87,9 +87,13 @@ define('SESSION_STORAGE', 'temporary value added by zc_install');
  * Advanced use only:
  * The following are OPTIONAL, and should NOT be set unless you intend to change their normal use. Most sites will leave these untouched.
  * To use them, uncomment AND add a proper defined value to them.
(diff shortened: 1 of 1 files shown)


drbyte:
Note that the DIR_FS_DOWNLOAD folder needs to be inside the webroot if they're not using symlink redirects, because with symlinks off (or not supported by the server) then the download is served directly from its folder. Unless download-by-streaming is used, in which case it is just chunked in RAM but served over PHP instead of direct from Apache.

scottcwilson:
Fair - I could split up the remarks and give guidance on DIR_FS_DOWNLOAD when default are used, then perhaps link to a page - would that work?

drbyte:
A few things:

1. The desire is to keep the configure. php files as lean as possible. Less room for confusion.
2. The dist-configure.php file is merely decorative. It's never used by the core code. Ever.
3. The zc_install template for configure.php is what is used to actually generate the populated file from the user-interview during installation.
4. Sites that don't do downloads don't need to do anything.

Perhaps "for your security, `if you are using the backups or downloads folders,` ...move them..." might suit?

PR #2473, opened 2019-06-20
PHP warning: Creating default object from empty value

scottcwilson:
Creating default object from empty value


admin/includes/modules/copy_product.php
```diff
@@ -20,6 +20,7 @@
 $heading[] = array('text' => '<h4>' . TEXT_INFO_HEADING_COPY_TO . '</h4>');
 // WebMakers.com Added: Split Page
 if (empty($pInfo->products_id)) {
+  if (!is_object($pInfo)) $pInfo = new StdClass();
```
drbyte: Perhaps this should be `new objectInfo(array())` instead of `StdClass`?

drbyte:
Looks good. Can you include the music product-type too, since it also has the same code segment.

PR #2476, opened 2019-06-21
Simplify and DRY up extra definitions inclusions

scottcwilson:


includes/modules/extra_definitions.php
```diff
@@ -1,74 +1,48 @@
...
+$lang_extra_defs_dir = DIR_WS_LANGUAGES . $_SESSION['language'] . '/extra_definitions/';
+$lang_extra_defs_dir_template = DIR_WS_LANGUAGES . $_SESSION['language'] . '/extra_definitions/' . $template_dir . '/';
+
+$file_array = array(); 
+$folderlist = array($lang_extra_defs_dir_template, $lang_extra_defs_dir);
+
+foreach ($folderlist as $folder) { 
+  $this_folder = DIR_FS_CATALOG . $folder; 
+  if ($dir = @dir($this_folder)) {
+    while ($file = $dir->read()) {
```
mc12345678: Is there a reason that directories with the name of 0 for example are purposely not allowed to be used?

includes/modules/extra_definitions.php
```diff
@@ -1,74 +1,48 @@
...
+      if (!is_dir($this_folder. $file)) {
+        if (array_key_exists($file, $file_array)) {
+          // skip name exists
+        } else {
+          $file_array[$file] = $folder . $file;
```
mc12345678: Without any additional test of `$file`, this will add all files that are in the directory including `.htaccess`, any hidden file and those that are not related to `.php` including files "renamed" to potentially prevent being processed.  Shouldn't an additional filter as provided in the removed code be included to eliminate files without an extension of `php`, files that begin with a period (`.`) and perhaps files that begin with an underscore (`_`)? (I assume the underscore option was added at some point to make it easy to "hide" a file from being loaded.)

Another thing, the use of the filename as the array key, doesn't this use more memory than a numerical index? The filename is or seems to already exist in the value and could be identified if needed, couldn't it?

includes/modules/extra_definitions.php
```diff
@@ -1,74 +1,48 @@
...
+}
+if (sizeof($file_array)) {
+    ksort($file_array);
+}
+
+
+$file_cnt=0;
+foreach ($file_array as $file => $include_file) { 
+  $file_cnt++;
+  require($include_file);
```
mc12345678: Why are these files being `require`d (hard stop in operation) instead of simply `include`d (operation continues even if there is a file issue but a `warning` is produced)?

scottcwilson:
- if ($dir = @dir was from before; I just didn't change it.
- you are correct about the .php check - fixed.
- I think require was safe because you've already checked file existence (the prior version did not in the case of non-overridden files) but I went back to include just in case.
- filename as array key - this is a tiny array :)

mc12345678:
> - if ($dir = @dir was from before; I just didn't change it.

It's actually the `$file = $dir->read()` that rejects the directory name that is falsey because there is not a strict comparison to `false`.  `if  ($dir = @dir)` sets up a handler `$dir` to be able to work the directory. The php faq shows a recommended coding style for `dir->read` or `readdir()` as seen here:  https://www.php.net/manual/en/function.readdir.php

> - I think require was safe because you've already checked file existence (the prior version did not in the case of non-overridden files) but I went back to include just in case.

While I hadn't gone through to fully test the concept it seems that it could be possible that a file could be seen in the path, but not accessible by include/require, though I personally don't know why such would purposefully been done.

scottcwilson:
Updated per example from https://www.php.net/manual/en/function.dir.php

includes/modules/extra_definitions.php
```diff
@@ -1,74 +1,50 @@
...
+
+$file_array = array(); 
+$folderlist = array($lang_extra_defs_dir_template, $lang_extra_defs_dir);
+
+foreach ($folderlist as $folder) { 
+  $this_folder = DIR_FS_CATALOG . $folder; 
+  if ($dir = @dir($this_folder)) {
+    while (false !== ($file = $dir->read())) {
+      if (!is_dir($this_folder. $file)) {
+        if (array_key_exists($file, $file_array)) {
```
mc12345678: Is there any case where `$folder . $file` would be purposefully and expectedly null?

includes/modules/extra_definitions.php
```diff
@@ -1,74 +1,50 @@
...
+    $dir->close();
+  }
+}
+if (sizeof($file_array)) {
+    ksort($file_array);
+}
+
+
+$file_cnt=0;
+foreach ($file_array as $file => $include_file) {
```
zcwilt: $file_cnt no longer required

includes/modules/extra_definitions.php
```diff
@@ -1,74 +1,48 @@
...
-
-// Check for new definitions in extra_definitions directory
-$dir_check = $directory_array;
-
-if ($dir = @dir($languages_extra_definitions_directory)) {
-  while ($file = $dir->read()) {
-    if (!is_dir($languages_extra_definitions_directory . $file)) {
-      if (preg_match('~^[^\._].*\.php$~i', $file) > 0) {
-        if (in_array($file, $dir_check, TRUE)) {
-          // skip name exists
```
zcwilt: Can we remove the empty if here

includes/modules/extra_definitions.php
scottcwilson: Added $file check - good catch.

includes/modules/extra_definitions.php
scottcwilson: Fixed.

zcwilt:
May not have been explicit in my last comment.
What I meant was to remove the empty clause in 

        if (array_key_exists($file, $file_array)) {
          // skip name exists
        } else {
          $file_array[$file] = $folder . $file;
        }

and reduce it from an if else to just an if, by reversing the test

scottcwilson:
sorry - thought I had pushed that change but I needed to merge first.

PR #2477, opened 2019-06-23
responsive_classic folder does not need sample html_includes files

scottcwilson:
It's confusing that these example files exist here; they are not needed.  Examples are in the parent directory.


dbltoe:
Agreed that it could be if template_default were to remain.  However, with template_default being non-responsive, wouldn't it be better to let responsive_classic become the template_default?  That may well have been the original reasoning for leaving the files there in the first place and, with previously the previously discussed addition of the color changes, it would give users a basic template with many possibilities at the onset.  I still have thoughts on this but, I've got to get a lot better in programming first.

scottcwilson:
> Agreed that it could be if template_default were to remain. However, with template_default being non-responsive, wouldn't it be better to let responsive_classic become the template_default?

No.  Dozens of files are not included in responsive_classic and it would be disruptive to add them.

mc12345678:
But the `define_main_page.php` file contains information that would only be (expected to be) available if using the `responsive_classic` template or a clone of it.

In this commit that file has been removed without incorporating a similar change to the default `define_main_page.php` file that would work in absence of `$detect` not being defined.

I can see why these files were incorporated so that they would be easier to maintain/edit such that they would not need to be copied from the base directory (via ftp/ssh) to further be edited to support similar customization as is done in `define_main_page.php`.  At the same time I could see where the other files that do not incorporate responsive characteristics may not serve as valuable of a reference or action.

I would say that if deleting files that `includes/languages/english/html_includes/responsive_classic/define_main_page.php` should remain at least until a suitable replacement is provided in `includes/languages/english/html_includes/define_main_page.php` provides the same responsive functionality without requiring `responsive_classic` or any other responsive template be the current template.

scottcwilson:
You are correct - that file should have been left in.

drbyte:
The problem with removing these files is that unless the template-specific file exists on the server, the define-pages-editor will edit the default one, not the template-specific one.

Also, I think a lot of the responsive_classic content is better than the defaults.

scottcwilson:
It's true that the default one is updated but that's just fine - these are not code files that need to be merged on  an update.   

I'd be happy to update the default ones with the content from responsive_classic if you'd like.

drbyte:
I'm concerned that there's a huge push here and in other issues that's based solely on minimizing pain for upgrading (which in itself is excellent) but is blindly ignoring the conveniences that are offered for those getting started.
Should we just go ahead and delete ALL examples of overrides, and assume people will figure it out themselves?

scottcwilson:
Actually, deleting overrides for **this** folder is an excellent idea.  They're not code files, you don't need to get updates and do merges, and you want the same content no matter what template is in use.

barco57:
> you want the same content no matter what template is in use.
Not always true, I have seen some sites where they dupe templates for different holidays and the define_main_page.php, and additional custom define pages added are different between templates

scottcwilson:
Got carried away in cleanup.

PR #2478, opened 2019-06-23
Don't trigger error if MODULE_PAYMENT_PAYPAL_IPN_DEBUG is not defined

scottcwilson:


scottcwilson:
Suggested for backport since this creates a log in PHP 7.2.

PR #2482, opened 2019-06-25
Column files should use zen_get_file_directory()

scottcwilson:
No need for if (template file exists) include template file; else include base file.  This is what zen_get_file_directory() does.


scottcwilson:
Got carried away in cleanup.

PR #2483, opened 2019-06-27
Ensure $column_width is an integer

scottcwilson:
Required because includes/templates/YOUR_TEMPLATE/sideboxes/tpl_search.php does math on this value.


scottcwilson:
Got carried away in cleanup.

PR #2486, opened 2019-06-27
Don't report success if admin email send failed

scottcwilson:


drbyte:
What are the scenarios motivating this change?  Since you haven't stated such, it's unclear whether this is "only" that the count of recipients intended to be mailed to is incorrect? Or is there something more that's trying to be captured here? And why only on this page?

While I agree it can be improved, and this approach is probably fine, given that complaints about it have been extremely sparse, what drove you to submit this PR? The "blank description" means context is completely missing.

scottcwilson:
Sorry should have been more descriptive.  The intention was to prevent misleading double messages such as the image below indicates. 

email_errors

Issue #2490, opened 2019-06-29
Table 'database.ezpages_content' doesn't exist


lat9:
I'm in the process of _re-building_ a zc154 store, using the zc156b distribution as the base.  

Simple enough, right?  Start by cloning the current site's database, copying the zc156 distribution's files to a fresh file-system and modifying the zc154-era `configure.php` files to point to reference the upgrade's URL and file-system directory.

When I then pointed the installation to the `/zc_install` processing, I was met with a partial whitescreen.  Turns out that the site's current configuration uses EZ-Pages in the header, sidebox and footer locations and the zc156b EZ-pages' processing presumes the presence of the `ez_pages_content` table which, unfortunately, isn't there until the installation/upgrade completes:

```
--> PHP Fatal error: 1146:Table 'database.ezpages_content' doesn't exist :: SELECT e.*, ec.*
                                FROM ezpages e,
                                     ezpages_content ec
                                WHERE e.pages_id = ec.pages_id
                                AND ec.languages_id = 1
                                AND e.status_sidebox = 1
                                AND e.sidebox_sort_order > 0
                                ORDER BY e.sidebox_sort_order, ec.pages_title ==> (as called by) /home/store/includes/modules/sideboxes/ezpages.php on line 23 <== in /home/store/includes/classes/db/mysql/query_factory.php on line 171.
```

I was able to work-around the issue by creating a temporary file in the site's `/includes/extra_configures` directory, containing:

```php
<?php
define('EZPAGES_STATUS_HEADER', '0');
define('EZPAGES_STATUS_SIDEBOX', '0');
define('EZPAGES_STATUS_FOOTER', '0');
```

to cause the page-handling to not bring in those EZ-pages' settings until the upgrade was completed, but there's surely a better way to handle this chicken/egg situation!

scottcwilson:
@lat9 Can you post the log you got?

lat9:
Look above.

scottcwilson:
Ah LOL it got cut off and I didn't see it.

lat9:
Noting that the log I posted was for the sidebox version; there were similar logs for the header and footer.

drbyte:
> When I then pointed the installation to the `/zc_install` processing, I was met with a partial whitescreen. Turns out that the site's current configuration uses EZ-Pages in the header, sidebox and footer locations

Puzzling ... because `zc_install` doesn't fire the catalog sideboxes/header/footer.
[...]

PR #2491, opened 2019-06-29
Ezpages abort for old databases

scottcwilson:
Prevent issue identified in #2490


includes/modules/ezpages_bar_footer.php
```diff
@@ -17,14 +17,20 @@
 
 // test if bar should display:
 if (EZPAGES_STATUS_FOOTER == '1' || (EZPAGES_STATUS_FOOTER == '2' && (strstr(EXCLUDE_ADMIN_IP_FOR_MAINTENANCE, $_SERVER['REMOTE_ADDR'])))) {
-  $pages_query = $db->Execute("SELECT e.pages_id, e.page_open_new_window, e.page_is_ssl, e.alt_url, e.alt_url_external, e.toc_chapter, ec.pages_title
+
+  if (defined('TABLE_EZPAGES_CONTENT')) {
```
drbyte: My concern here is that the code begins to assume that the normally-expected dependencies do NOT exist, which feels backwards.

includes/modules/ezpages_bar_footer.php
```diff
@@ -17,14 +17,20 @@
...
                               FROM  " . TABLE_EZPAGES . " e,
                                     " . TABLE_EZPAGES_CONTENT . " ec
                               WHERE e.pages_id = ec.pages_id 
                               AND ec.languages_id = " . (int)$_SESSION['languages_id'] . "
                               AND e.status_footer = 1
                               AND e.footer_sort_order > 0
                               ORDER BY e.footer_sort_order, ec.pages_title");
+  } else {
+    // Failing query 
+    $pages_query = $db->Execute("SELECT * FROM " . TABLE_EZPAGES . " WHERE pages_id < 0");
```
drbyte: The test for `pages_id < 0` is going to just give empty results ... which in itself is an okay way to "obtain a database response, and pass the empty response on to the rest of the code.

But I'd rather see us do something like an "early return" if things are "in an old unupgraded-and-therefore-unusable-state", since then that part of the code can be easily identified and removed later when someone comes along wanting to remove deprecated code/comments.

scottcwilson:
Actually my original logic would not have worked properly because the define could be set if the files were staged but the db not updated.  This seems better.

PR #2495, opened 2019-07-04
Retrieve ezpages_title from TABLE_EZPAGES_CONTENT in query

scottcwilson:


zcwilt:
Surely this will fail with multiple languages as there is no where for the languages_id.

Also, while we may have examples in old code, not a fan of implicit joins.

scottcwilson:
Good catch on multilingual.  As for query structure, trying to be consistent with other TABLE_EZPAGES_CONTENT queries.

PR #2499, opened 2019-07-05
Fix PHP 7.2 notices when dealing with GV (admin)

scottcwilson:


admin/gv_queue.php
```diff
@@ -162,7 +163,7 @@ function init()
   $gv_split = new splitPageResults($_GET['page'], MAX_DISPLAY_SEARCH_RESULTS, $gv_query_raw, $gv_query_numrows);
   $gv_list = $db->Execute($gv_query_raw);
   while (!$gv_list->EOF) {
-    if (((!$_GET['gid']) || (@$_GET['gid'] == $gv_list->fields['unique_id'])) && (!$gInfo)) {
+    if (((!isset($_GET['gid'])) || (@$_GET['gid'] == $gv_list->fields['unique_id'])) && (!isset($gInfo))) {
```
mc12345678: Won't this get bypassed if `$_GET['gid'] == 0`, `$_GET['order']` is not provided even if `$gInfo` isn't yet set?

In other words shouldn't the first `isset` be `empty`?

admin/gv_queue.php
```diff
@@ -200,6 +201,9 @@ function init()
 //      $contents[] = array('align' => 'center', 'text' => '<a href="' . zen_href_link('gv_queue.php', 'action=confirmrelease&gid=' . $gInfo->unique_id . '&page=' . $_GET['page'],'NONSSL') . '">' . zen_image_button('button_confirm_red.gif', IMAGE_CONFIRM) . '</a> <a href="' . zen_href_link('gv_queue.php', 'action=cancel&gid=' . $gInfo->unique_id . '&page=' . $_GET['page'],'NONSSL') . '">' . zen_image_button('button_cancel.gif', IMAGE_CANCEL) . '</a>');
       break;
     default:
+      if (!isset($gInfo) || !is_object($gInfo)) { 
+        $gInfo = new objectInfo(array());
```
mc12345678: Thought this was supposed to rid notices, won't a notice be generated 2 lines down if this is executed? (no `unique_id`, no `date_created`, etc...)

PR #2500, opened 2019-07-05
Fix PHP7.2 notices during buy gift certificate

scottcwilson:


includes/classes/order.php
```diff
@@ -335,7 +335,7 @@ function cart() {
                                 from " . TABLE_ADDRESS_BOOK . " ab
                                 left join " . TABLE_ZONES . " z on (ab.entry_zone_id = z.zone_id)
                                 where ab.customers_id = " . (int)$_SESSION['customer_id'] . "
-                                and ab.address_book_id = " . ($this->content_type == 'virtual' ? (int)$_SESSION['billto'] : (int)$_SESSION['sendto']);
+                                and ab.address_book_id = " . ($this->content_type == 'virtual' ? (!empty($_SESSION['billto']) ? (int)$_SESSION['billto'] : 0) : (int)$_SESSION['sendto']);
```
drbyte: It seems to me that `billto` is more likely to be set than `sendto`, so why not handle `sendto` the same ?

scottcwilson: I could - I have just never seen a sendto log, that's why I only changed the thing that was logging.

drbyte: It's particularly strange that there'd be no billto address on a virtual order.

Especially given the other sanitizations recently made (lat9 I think) to the order class to prevent such "notices" (or "logs" as you call them).

includes/classes/order.php
```diff
@@ -325,6 +325,8 @@ function cart() {
     // set default tax calculation for not-logged-in visitors
       $taxCountryId = $taxZoneId = 0;
 
+    $billto = (!empty($_SESSION['billto']) ? (int)$_SESSION['billto'] : 0);
+    $sendto = (!empty($_SESSION['sendto']) ? (int)$_SESSION['sendto'] : 0);
```
drbyte: Should these then be moved to the top of the function, and used in the various queries that are run between there and here?

PR #2506, opened 2019-07-06
Remove unnecessary tests if action performed.

mc12345678:
The way that the code had been revised makes subsequent
operations initially unnecessary but necessary if the first
is not executed.

### DIFF
--- a/includes/classes/payment.php
+++ b/includes/classes/payment.php
@@ -218,9 +218,12 @@ class payment extends base {
     if (is_array($this->modules)) {
       if (is_object($GLOBALS[$this->selected_module]) && ($GLOBALS[$this->selected_module]->enabled) ) {
         $confirmation = $GLOBALS[$this->selected_module]->confirmation();
-        if (!is_array($confirmation)) $confirmation = array('title' => '', 'fields' => array());
-        if (!isset($confirmation['fields']) || !is_array($confirmation['fields'])) $confirmation['fields'] = array();
-        if (!isset($confirmation['title'])) $confirmation['title'] = '';
+        if (!is_array($confirmation)) {
+          $confirmation = array('title' => '', 'fields' => array());
+        } else {
+          if (!isset($confirmation['fields']) || !is_array($confirmation['fields'])) $confirmation['fields'] = array();
+          if (!isset($confirmation['title'])) $confirmation['title'] = '';
+        }
       }
     }
     return $confirmation;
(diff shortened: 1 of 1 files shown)


scottcwilson:
Technically correct, and it does save two memory-to-memory comparisons, but this seems just a little harder for a human to read.

mc12345678:
I found the latest revision to this specific area hard to understand. Before it seemed to make sense where the desired results were built in pieces allowing almost effectively one line of code to need revision if it were desired for a different default. Here now the default is reflected in two lines one for a non-array and one where it is an array, but not set. Perhaps it would make more sense to test if it is an array first rather than not an array.

There was no problem with the previous internal structure, the problem was that there was no specific assignment returned. This revision makes it clearer about which set of code (array or not an array) is making the assignment of the array keys. It also strengthens the need of the two lines that follow the array check showing that they are not superfluous. At least not until the returned result always contains the necessary data.

zcwilt:
How about this. Haven't tested fully but would be cleaner.

    function confirmation()` {
      $default = ['title' => '', 'fields' => []];
      if (!is_array($this->modules)) {
        return $default;
      }
      if (!is_object($GLOBALS[$this->selected_module]) || (!$GLOBALS[$this->selected_module]->enabled) ) {
       return $default;
      }
      $confirmation = $GLOBALS[$this->selected_module]->confirmation();
      $confirmation = (!is_array($confirmation)) ? $default : array_merge($default, $confirmation);
      return $confirmation;
    }

mc12345678:
> How about this. Haven't tested fully but would be cleaner.
>
> ````php
> function confirmation()` {
>   $default = ['title' => '', 'fields' => []];
>   if (!is_array($this->modules)) {
>     return $default;
>   }
>   if (!is_object($GLOBALS[$this->selected_module]) || (!$GLOBALS[$this->selected_module]->enabled) ) {
>    return $default;
>   }
>   $confirmation = $GLOBALS[$this->selected_module]->confirmation();
>   $confirmation = (!is_array($confirmation)) ? $default : array_merge($default, $confirmation);
>   return $confirmation;
> }
> ````


Looks like a good rewrite, certainly cleaner and if anything I like the array_merge use. :) Course looks like the first two ifs could be combined to a single if statement all `||`'d together, but whatever.

zcwilt:
Yes, the two if's could be combined, but I was mindful of creating long if statements. 
Sometimes when doing these `guard` type tests I'll move the test to a separate `can` type function, but that's probably a bit overblown for this case.

drbyte:
Updated

PR #2510, opened 2019-07-06
Perform products_id validate on first load/session not started.

mc12345678:
A portion of includes/init_includes/init_sanitize.php is not
executed on the first load to a site, leaving follow on code to
respond instead of this change.  Not sure if this is to in some
way improve first page load time or if wasn't something simply
expected to be handled later but was offerred an opportunity
during routine navigation.

### DIFF
--- a/includes/init_includes/init_sanitize.php
+++ b/includes/init_includes/init_sanitize.php
@@ -129,7 +129,7 @@
 /**
  * validate products_id for search engines and bookmarks, etc.
  */
-  if (isset($_GET['products_id']) && isset($_SESSION['check_valid']) &&  $_SESSION['check_valid'] != 'false') {
+  if (isset($_GET['products_id']) && (!isset($_SESSION['check_valid']) || $_SESSION['check_valid'] != 'false')) {
     $check_valid = zen_products_id_valid($_GET['products_id']);
     if (!$check_valid) {
       $_GET['main_page'] = zen_get_info_page($_GET['products_id']);
@@ -139,9 +139,9 @@
       $_SESSION['check_valid'] = 'false';
       zen_redirect(zen_href_link($_GET['main_page'], 'products_id=' . $_GET['products_id']));
     }
-  } else {
-    $_SESSION['check_valid'] = 'true';
   }
+ 
+  $_SESSION['check_valid'] = 'true';
 /**
  * We do some checks here to ensure $_GET['main_page'] has a sane value
  */


zcwilt:
So, the intention of this code is to test if a product id is valid. If not the code redirects to the product info page for the invalid product id, which in essence will produce a 'Sorry, the product was not found.' 
and was intended to handle 'bookmarked'  product pages where the product was later removed.

The use of the $_SESSION['check_valid'] has nothing to do with page load time or hits on first load, but to stop an infinite loop occurring.

mc12345678:
> So, the intention of this code is to test if a product id is valid. If not the code redirects to the product info page for the invalid product id, which in essence will produce a 'Sorry, the product was not found.'
and was intended to handle 'bookmarked' product pages where the product was later removed.
> 
> The use of the $_SESSION['check_valid'] has nothing to do with page load time or hits on first load, but to stop an infinite loop occurring.

I thank you for that expectation explanation.  I believe that is/was the intent as well.  In review of the original code, the inner check (ideal goal of the code snippet) is not performed when the site is first visited as `$_SESSION['check_valid']` is not yet set by this code section.

The revision proposed will attempt to execute as described above where on first load the provided information will attempt to be validated and if that is truly unsuccessful to then pass the handling on to whatever `main_page` result is provided by `zen_get_info_page` (expected to be `product_info` if no other good product type is discovered).

Thinking about it some more, assuming that there is no other expected use of `$_SESSION['check_valid']`, then it seems that it could be set only once in the code section but to be unset otherwise:

````php
/**
 * validate products_id for search engines and bookmarks, etc.
 */

 if (isset($_GET['products_id']) && !isset($_SESSION['check_valid'])) {
    $check_valid = zen_products_id_valid($_GET['products_id']);
    if (!$check_valid) {
      $_GET['main_page'] = zen_get_info_page($_GET['products_id']);
      /**
       * do not recheck redirect
       */
      $_SESSION['check_valid'] = 'false';
      zen_redirect(zen_href_link($_GET['main_page'], 'products_id=' . $_GET['products_id']));
    }
  }
 
  unset($_SESSION['check_valid']);
````

In here, `$_SESSION['check_valid']` is only set so long as necessary to redirect to the next location that is supposed to be *appropriate* and to then be unset.

Something noticed as well though is that it seems that database reads might possibly be reduced if the query in this function matched the query used in the later investigation. There likely though is an issue of chicken/egg based on what is available during the load sequence.  Just was an additional thought.

Issue #2504, opened 2019-07-05
Enable configuration default for 'Notify Customer'?


lat9:
On the admin's Customers->Orders, I've got a bunch of clients that require a change to the /admin/orders.php to change the default from 'Email'.

What's everyone think about adding a configuration setting to `My Store` to define the default value to be used?

drbyte:
You can already change the default "email from" email address under Admin->Configuration->Email Options.

But I assume you already know that, so clearly I'm misunderstanding whatever you meant to say.

lat9:
Nope, I'm talking about the customer-visibility of the orders-update:

![image](https://user-images.githubusercontent.com/2685585/60739274-53d10f80-9f2f-11e9-9a25-e754c758df60.png)

drbyte:
What's the default they prefer to have? Is there a pattern?

barco57:
yes, it should probably be a configuration option. I have had to change it for a couple of clients that normally change order status for internal tracking while processing an order that they don't need/ want an email to be sent and sometimes they forget to switch it to no email manually and end up getting question about what the email means when it has internal company stuff in it...

scottcwilson:
Very interesting suggestion. One change I have done is NO default, with javascript to ensure that one of the buttons is picked.

barco57:
interesting, and would also be a good idea

lat9:
> 
> 
> What's the default they prefer to have? Is there a pattern?
No real 'pattern', I've got some clients that want the default to 'Hide' and others that want 'No Email'.

lat9:
I'll be happy to submit a PR, but I'll need some guidance on the `zc_install`-related changes.

lankeeyankee:
yes, this would be a nice feature to have

scottcwilson:
@lat9 I went ahead and did the db stuff because it's a pain the first time for a release - when PRs #2508 and #2509 are accepted, you can make the changes to admin/orders.php (and perhaps edit orders) using the DB configuration key NOTIFY_CUSTOMER_DEFAULT.

lat9:
Thanks, @scottcwilson.  I'll get that going.  Note that v4.4.0 of EO already has its own setting for this!

PR #2511, opened 2019-07-06
#2504: Admin orders, default customer notify

lat9:
... based on the new configuration setting (thanks @scottcwilson).


scottcwilson:
Requires PRs #2508 and #2509.

PR #2512, opened 2019-07-07
Improve upgrade message

scottcwilson:
When I went to test the upgrade, the new version had not yet been defined in the `$updateList`, so the message was confusing (since one of the strings which were embedded in the error message were blank). 

Instead of saying:
[we] must perform the updates to get to version 1.5.6 first.
it said, 
[we] must perform the updates to get to version first.


scottcwilson:
Why you no like, @drbyte ?

drbyte:
That was a development-only error. 
This would be the only place where we add error messages to say that there is missing code.
Plus, why are we adding more language defines for something that doesn't need to be translatable since it's info-only.

drbyte:
I've reworked the approach to something leaner.

PR #2513, opened 2019-07-07
admin - remove old language="javascript" tags

Zen4All:
These tags are old and obsolete, and should be removed.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2513)


scottcwilson:
Love getting rid of the <!-- around JS too.  Thanks for doing this.

PR #2515, opened 2019-07-07
Remove the admin demo functionallity

Zen4All:
Officially the zen_admin_demo function has been deprecated years ago, and was planned to be removed in 2.0
This PR removes it for 1.5.7

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2515)


scottcwilson:
I wish we weren't doing this.  I use this functionality to build admin demos. :(

Zen4All:
@scottcwilson Maybe continue it as a separate add on? I think the you and I ,and a very small group of developers are actually using this.

drbyte:
Unfortunately this is hard to move to an addon.

It begs of a full per-action-permission authorization infrastructure.

I too would like to see the demo-specific code removed (and I wonder if it's thorough enough as it is right now), but it may be better to morph it to an `action`-based permissions-control system.

zcwilt:
Agree that removing this does tidy up code and given it is a very niche requirement.
@scottcwilson  have a really simple idea about how we can do this as an addon. will ping you with details later.

drbyte:
Note: for a true demo-only site, setting a cron job to replace the database daily is probably an even better option. And probably hard-code a few feature overrides like config settings, as constants instead of loading them from the db.

PR #2516, opened 2019-07-07
Remove CDATA XML comments around JS

scottcwilson:


drbyte:
Have you tested this thoroughly?

Is it truly safe to remove the `<![CDATA[` wrapper? (even though it "appears" to be in comments)

scottcwilson:
Didn't see errors in Chrome Inspect while browsing the site.

zcwilt:
99% sure this is alright to do now. CDATA tags were originally needed for validators as XML So removing them won't affect the js from running. Also we have plenty of other inline scripts that aren't using CDATA tags.

PR #2517, opened 2019-07-07
Fix PHP Notice when no db checks defined yet in zc_install

scottcwilson:
Resolves 

[07-Jul-2019 03:33:16 America/New_York] PHP Notice:  Undefined index: methods in /Users/scott/Sites/gh_demo_157/zc_install/includes/classes/class.systemChecker.php on line 163


drbyte:
This may be a consequence of not having a proper list of db checks in the yml file. 
I haven't put any in there yet.

PR #2523, opened 2019-07-07
Fix PHP7.3 notices when non-existent coupon entered

scottcwilson:


includes/modules/order_total/ot_coupon.php
```diff
@@ -170,17 +170,25 @@ function collect_posts() {
...
+    }
+
     if ((isset($_POST['dc_redeem_code']) && $_POST['dc_redeem_code'] != '') || (isset($discount_coupon->fields['coupon_code']) && $discount_coupon->fields['coupon_code'] != '')) {
       // set current Discount Coupon based on current or existing
-      if (isset($_POST['dc_redeem_code']) && $discount_coupon->fields['coupon_code'] == '') {
-        $dc_check = $_POST['dc_redeem_code'];
-      } else {
-        $dc_check = $discount_coupon->fields['coupon_code'];
-      }
+      if ($discount_coupon != null) {
```
drbyte: Is the parent `if` needed anymore if we add these nested if/else stuff?

includes/modules/order_total/ot_coupon.php
```diff
@@ -170,17 +170,25 @@ function collect_posts() {
...
+        if (isset($_POST['dc_redeem_code']) && $discount_coupon->fields['coupon_code'] == '') {
```
drbyte: `$_POST['dc_redeem_code']` is always set, per line 160. Maybe this should be checking for `empty`?

drbyte: Oh wait, maybe not if "removed".

includes/modules/order_total/ot_coupon.php
scottcwilson: I believe so.  This is pretty hairy logic that took a while to debug.

PR #2538, opened 2019-07-11
Indicate in backend if orders placed on mobile

scottcwilson:
This is a proof of concept implementation; once accepted I'd like to do something nicer with the indicator on the orders list (maybe an image for M and D or something).

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -890,6 +890,7 @@ if (zen_not_null($action) && $order_exists == true) {
                     }
                     ?>
                   <td class="dataTableHeadingContent text-center"><?php echo TABLE_HEADING_ORDERS_ID; ?></td>
+                  <td class="dataTableHeadingContent text-center">&nbsp;</td>
                   <td class="dataTableHeadingContent"><?php echo TABLE_HEADING_PAYMENT_METHOD; ?></td>
                   <td class="dataTableHeadingContent"><?php echo TABLE_HEADING_CUSTOMERS; ?></td>
                   <td class="dataTableHeadingContent text-right"><?php echo TABLE_HEADING_ORDER_TOTAL; ?></td>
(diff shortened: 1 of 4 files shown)


torvista:
I have this:
Clipboard01

scottcwilson:
You have added a column Info.  Check your integration.  
Admin_Orders

scottcwilson:
oh sorry I thought you were reporting a bug.  Yes it's the same idea as what you're showing.

Zen4All:
Just out of curiosity, but why would you need that info?

scottcwilson:
It's useful to know how your customers are interacting with your site.  If you see that most of your orders are coming from mobile, you might want to invest more in optimizing that experience.

Zen4All:
OK, I use Google Analytics for that. 😃

drbyte:
I think an analytics solution is better than adding a single extra field to the orders table which is 1% of the data analytics can give.

scottcwilson:
A number of my clients complain that GA is impossible to figure out.  This gives them an easy perspective on how customers are buying.

drbyte:
Is it better to create a `metadata` field into which a bunch of info-only data is stored (json perhaps)? Rather than starting a pattern of adding a bunch more fields that need to be maintained? Particularly where the data is display-only and doesn't really need to be searchable.

@torvista in your screenshot, how are you capturing that an order was placed on mobile? Is it a field in your database? Or another approach?

torvista:
That column data comes from a old swguy mod which, without bothering to check, I imagine uses the same method as this commit.

At the time, I wanted an idea of how urgent it was to sort out a responsive template....but in this day and age I think it is pretty pointless to know how many use a mobile to buy when the answer now is at least "lots" and is only going to change to "most"...so you need to cater for them now. Period.
This column only serves to annoy me that I still haven't upgraded and done a responsive template yet. It's of slight interest but I don't know if worthy of inclusion as an improvement: there are bigger fish to fry.

scottcwilson:
It acts as a convincer to people using old desktop-only templates who insist "my template is awesome!"

drbyte:
This appears to be the wrong (incomplete) solution to the identified concern.

Maybe there's an opportunity to have the UserTracking plugin link data to a specific order.
Or if GA is too complicated, Matamo (formerly Piwik) is probably ideal.

PR #2541, opened 2019-07-11
Respect overrides in includes/modules/sideboxes

scottcwilson:
Using `require(DIR_WS_MODULES . 'sideboxes/search_header.php');` means a sidebox template override will be ignored.

### DIFF
--- a/includes/functions/functions_general.php
+++ b/includes/functions/functions_general.php
@@ -1322,7 +1322,7 @@
     }
   }
 
-  function zen_get_module_sidebox_directory($check_file, $dir_only = 'false') {
+  function zen_get_module_sidebox_directory($check_file) { 
     global $template_dir;
 
     $zv_filename = $check_file;
@@ -1334,11 +1334,7 @@
       $template_dir_select = 'sideboxes/';
     }
 
-    if ($dir_only == 'true') {
-      return $template_dir_select;
-    } else {
-      return $template_dir_select . $zv_filename;
-    }
+    return $template_dir_select . $zv_filename;
   }
 
 
--- a/includes/functions/functions_general.php
+++ b/includes/functions/functions_general.php
@@ -1322,6 +1322,25 @@
     }
   }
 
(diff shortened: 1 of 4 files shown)


includes/functions/functions_general.php
```diff
@@ -1322,6 +1322,25 @@ function zen_get_module_directory($check_file, $dir_only = 'false') {
...
+    $zv_filename = $check_file;
+    if (!strstr($zv_filename, '.php')) $zv_filename .= '.php';
+
+    if (file_exists(DIR_WS_MODULES . 'sideboxes/' . $template_dir . '/' . $zv_filename)) {
+      $template_dir_select = 'sideboxes/' . $template_dir . '/';
+    } else {
+      $template_dir_select = 'sideboxes/';
+    }
+
+    if ($dir_only == 'true') {
```
drbyte: Given this is a brand new function, let's treat this `'true'` string as a boolean instead.

Even better, is this `$dir_only` conditional even needed? If it's never gonna be used, let's leave it out.

drbyte:
Should the template_default file be included?

scottcwilson:
Oh I was going to do that one once this was accepted.

drbyte:
Concept accepted.
Let's put all the affected files into a single PR. Makes it a lot easier to track changes down the road.

scottcwilson:
Coming right up!

drbyte:
Maybe these are already on your radar?

- Are the `includes/modules/column_left.php` files affected by this?
- Should `responsive_classic/templates/tpl_modules_mobile_menu.php` be included?

scottcwilson:
Second was on my radar.   First one needs additional attention (should use zen_get_file_directory, etc.).

scottcwilson:
... and by "the first one" I mean all three of those files (includes/modules/column_left.php, includes/modules/column_single.php, includes/modules/column_right.php) which I suspect was what you meant.  Can we merge this one and I'll do those next?

PR #2542, opened 2019-07-11
Ensure submitonce() is available on payment page

### LINKED ISSUE #2536: Using PADSS_AJAX_CHECKOUT means some JS not loaded
Specifically, from includes/modules/pages/checkout_confirmation/jscript_main.php (with the submitonce check) is not loaded.

scottcwilson:
Fixes #2536 PADSS_AJAX_CHECKOUT means checkout confirmation javascript not loaded so submit once protection is missing.

### DIFF
--- a/includes/modules/pages/checkout_payment/jscript_main.php
+++ b/includes/modules/pages/checkout_payment/jscript_main.php
@@ -75,3 +75,4 @@ function doCollectsCardDataOnsite()
     });
 
 </script>
+<?php require(DIR_WS_MODULES . '/pages/checkout_confirmation/jscript_double_submit.php'); ?>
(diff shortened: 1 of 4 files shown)


includes/modules/pages/checkout_payment/jscript_main.php
```diff
@@ -75,3 +75,4 @@ function doCollectsCardDataOnsite()
     });
 
 </script>
+<?php require(DIR_WS_MODULES . '/pages/checkout_confirmation/jscript_double_submit.php'); ?>
```
drbyte: Can we stick this require into a separate file? 
ie: since all jscript_*.php files are autoloaded, putting `jscript_double_submit.php` (or any name) into the checkout_payment folder, but still doing the require, makes code maintenance simpler.

drbyte:
Note: I removed the closing `?>` tag. Don't put those at the end of PHP files, especially when blank lines follow!

PR #2544, opened 2019-07-11
Remove debug messages

scottcwilson:


drbyte:
let's leave this one alone

PR #2545, opened 2019-07-11
Remove debug code

scottcwilson:


drbyte:
1. For all these "cleanups", one PR would be preferred, not hundreds.

2. Please don't treat the fact that some cleanups have been merged as a free-for-all to do a blanket removal of any commented code.

scottcwilson:
Reason 1 wouldn't allow you to easily decline one.  I did them separately because of reason 2.

PR #2550, opened 2019-07-11
Updating password should reset locked_out time, failures

scottcwilson:


drbyte:
I haven't checked: If you're locked-out, can you reset your pwd? If so, then this would defeat the purpose of a lockout (and in such case, this PR would be fine, but the reset should be prevented).

scottcwilson:
Good catch.

PR #2551, opened 2019-07-12
Provide the option of not going to 404 when product status is 0

### LINKED ISSUE #1699: Provide the option of not going to 404 when product status is 0
Right now, due to checks in includes/modules/pages/product_info/header_php.php  and includes/modules/pages/product_info/main_template_vars.php, products which are in the catalog but have status = 0 return a 404. 

Should we (optionally) show the product info page with the add to cart button grayed out instead?  This way vendors could show back in stock notification, related items, etc.

drbyte:
Split out the code for handling 404 Header responses so it can be intercepted with an Observer if needed.
Added ability to define a constant PRODUCT_THROWS_200_WHEN_DISABLED to boolean `true` without needing an Observer, if a Store wishes to allow a product to display even when disabled.
A 404 will still trigger if the product doesn't actually exist in the db.
A 410 can be triggered by setting `products_status` to `-1` in custom code, without needing an observer for the 410 response.

This change should also reduce queries on product page.

Fixes #1699

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -1,0 +1,101 @@
+<?php
+/**
+ * functions_products.php
+ * Functions related to products
+ * Note: Several product-related lookup functions are located in functions_lookups.php
+ *
+ * @package functions
+ * @copyright Copyright 2019 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: New in v1.5.7 $
+ */
+
+/**
+ * Query product details
+ * @param int $product_id
+ * @param int $language_id (optional)
+ * @return queryFactoryResult
+ */
+function zen_get_product_details($product_id, $language_id = null)
+{
+    global $db;
+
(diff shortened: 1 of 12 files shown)


drbyte:
Testing/feedback welcome

scottcwilson:
@drbyte do you want me to put PRODUCT_THROWS_200_WHEN_DISABLED in the database?

drbyte:
> @drbyte do you want me to put PRODUCT_THROWS_200_WHEN_DISABLED in the database?

No, I wasn't planning on adding more config entries. The related discussion talked about this somewhat.

scottcwilson:
Wondering if we should change the 404 to a 410 still.  I am not strong on this - just wondering.

drbyte:
In this PR the 404 is triggered if "product does not exist in database".
This situation would happen for basically 2 cases:
1. product never existed. This is likely only to occur with bots or uncouth visitors. 404 is ideal.
2. product "did" exist, but has been actually deleted. A hit of this sort is likely from search engines or bookmarks or old newsletters. A 410 would be ideal.

Thinking about it, I guess the latter is more likely to occur.

PR #2555, opened 2019-07-12
Add condition and doc for submitonce check on payment page

scottcwilson:


scottcwilson:
Suggested as a backport candidate for 1.5.6.

PR #2559, opened 2019-07-13
Extra check for non super user access to banner statistics page

scottcwilson:
Non superusers who are allowed to use the Tools->Banner Manager page should also have access to the banners statistics page, but there's no checkbox for this page, so we'll use Banner Manager as a proxy.

### DIFF
--- a/admin/includes/functions/admin_access.php
+++ b/admin/includes/functions/admin_access.php
@@ -59,6 +59,13 @@ function check_page($page, $params) {
   return $retVal;
 }
 
+function check_related_page($page, $params) {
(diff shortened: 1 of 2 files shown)


scottcwilson:
Haven't heard this need expressed by anyone else.

zcwilt:
yet :)

scottcwilson:
Easy to add when needed.  Just augment `check_related_page` with 

`
$checkfunction = "check_" . $page;
if (function_exists($checkfunction)) {
  return $checkfunction();
}
`

PR #2560, opened 2019-07-13
Add utility sql script to remove demo data from db

scottcwilson:
Allows iterative testing of demo data.

### DIFF
--- a/zc_install/sql/demo/README.txt
+++ b/zc_install/sql/demo/README.txt
@@ -1,0 +1,40 @@
+The demo_cleanup.sql script deletes ALL of records in the 
+tables listed below.  It is intended for developers of demo data only.
+In particular, it should not be used if you have begun adding real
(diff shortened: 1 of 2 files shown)


zcwilt:
Have tested and seems ok. 
However would add some possible gotchas

There is the possible intention to remove music type stuff from core to a plugin. 
If we do move to innodb and especially the use of foreign keys, then the order of truncation may need 
adjusting.

Given that, I can merge this but would it be better to wait for other stuff to happen first.

torvista:
What about the demo folders and images? I would have thought the idea would be to get things the same as if the demo data had not been installed at all.

scottcwilson:
Steve, I will add the demo data separation as a different PR; it's on my list.

Wilt, I'd rather you merged this now so that it's done- I will need it for other tests I'm doing, and the other changes you're mentioning are a ways off.

PR #2565, opened 2019-07-13
Consolidate duplicated functions

scottcwilson:
This change pulls out common functions between admin/functions/general.php and includes/functions/functions_general.php.  Only functions which were identical or identical except for a small tweak were included; other opportunities for consolidation exist but require additional catalog/admin checks (or at least testing to determine if the differences are still needed).


drbyte:
How does this affect all the admin plugins that touch functions_general and therefore replace it upon install, and now will trigger function-already-exists errors?

Or all all those plugins no longer doing such replacement?

scottcwilson:
I don't see many such plugins in my archive.  SBA, SEO URL rewriters.  Bottom line, people have to update plugins as the core changes.

scottcwilson:
Not all that different from any other change to admin/functions/general.php, which would also have to be merged.

mc12345678:
> I don't see many such plugins in my archive. SBA, SEO URL rewriters. Bottom line, people have to update plugins as the core changes.

As of ZC 1.5.5, SBA no longer revised the `admin/includes/functions/general.php` file and as of ZC 1.5.6 there no longer is any function code revised on either admin or catalog side based on core code changes making it unnecessary to touch them.

That said, a couple of years ago when attempting to assist with the [removal of zen_db_output](https://github.com/zencart/zencart/pull/831), the differences in the "shared" function `zen_output_string` were seen as a significant issue.  On the catalog side, the text is double-encoded, but on the admin side it is not.  Further then the `zen_output_string_protected` function basically works in the same two different ways where on the catalog side the content is double encoded, but on the admin side it is not.  So, the same content posted on the admin side and the catalog side potentially has two different results of outputting the same content for an undisclosed reason...

For example, on the admin side when viewing the preview of a product, the products_name is sent through `zen_output_string_protected` which will effectively execute `htmlspecialchars($string, ENT_COMPAT, CHARSET, false)`, while on the catalog side, when viewing a product, the raw `$string` will be displayed instead of going through such a 'protection'.  

On a slightly different but similar topic: Why is the function `zen_parse_input_field_data` not also consolidated? (Noticed because it is called by both the admin and catalog side when processing `zen_output_string` and not trying to provide "protected" encoding but is not in the combined code.)

The difference in condition between the catalog and the admin of the output from `zen_output_string` when called with `$protected  == true` is still puzzling and in a way misleading to anyone attempting to use these functions from the standpoint of what they appear to mean...

scottcwilson:
> Why is the function zen_parse_input_field_data not also consolidated?
I just did the most simple ones that required little or no extra code.  As noted in the PR submit comments, "other opportunities for consolidation exist."   This is a start.  Once this is approved, I'd like to do other files.  Most of this copy & pasted code is logic we inherited from osCommerce, and it would improve the quality of the code base to get rid of this needless replication.

drbyte:
@scottcwilson  
Before final release I'd like to see the filename changed from "functions_general_shared" to something "better".

If you look at v2 where a bunch of consolidation was done, things were broken into files based on context. Would like to continue that approach, even if it means breaking up this "functions_general_shared" into separates.

And probably drop the "functions_" prefix.

scottcwilson:
No problem but can you please pick your preferred name(s)?  How about just shared.php?

Issue #2564, opened 2019-07-13
Admin, new document_general, missing constants


lat9:
zc156b.  Navigate to the admin's _Catalog::Categories/Products_ and create a new product of type _Document - General_.  A log similar to the following will be generated for missing constants:
```
[13-Jul-2019 08:48:00 America/New_York] Request URI: /test_store/adminzc156/product.php?action=new_product, IP address: ::1
#1  require(C:\xampp\htdocs\test_store\adminzc156\includes\modules\document_general\collect_info.php) called at [C:\xampp\htdocs\test_store\adminzc156\product.php:101]
--> PHP Notice: Use of undefined constant DEFAULT_DOCUMENT_GENERAL_PRODUCTS_VIRTUAL - assumed 'DEFAULT_DOCUMENT_GENERAL_PRODUCTS_VIRTUAL' in C:\xampp\htdocs\test_store\adminzc156\includes\modules\document_general\collect_info.php on line 21.

[13-Jul-2019 08:48:00 America/New_York] Request URI: /test_store/adminzc156/product.php?action=new_product, IP address: ::1
#1  require(C:\xampp\htdocs\test_store\adminzc156\includes\modules\document_general\collect_info.php) called at [C:\xampp\htdocs\test_store\adminzc156\product.php:101]
--> PHP Notice: Use of undefined constant DEFAULT_DOCUMENT_GENERAL_TAX_CLASS_ID - assumed 'DEFAULT_DOCUMENT_GENERAL_TAX_CLASS_ID' in C:\xampp\htdocs\test_store\adminzc156\includes\modules\document_general\collect_info.php on line 27.

[13-Jul-2019 08:48:00 America/New_York] Request URI: /test_store/adminzc156/product.php?action=new_product, IP address: ::1
#1  require(C:\xampp\htdocs\test_store\adminzc156\includes\modules\document_general\collect_info.php) called at [C:\xampp\htdocs\test_store\adminzc156\product.php:101]
--> PHP Notice: Use of undefined constant DEFAULT_DOCUMENT_GENERAL_PRODUCTS_IS_ALWAYS_FREE_SHIPPING - assumed 'DEFAULT_DOCUMENT_GENERAL_PRODUCTS_IS_ALWAYS_FREE_SHIPPING' in C:\xampp\htdocs\test_store\adminzc156\includes\modules\document_general\collect_info.php on line 35.
```

lat9:
Noting that this occurred on a site being upgraded from zc154.

[...]

PR #2566, opened 2019-07-13
Fix constants missing notices

scottcwilson:
As reported in #2564.


admin/includes/modules/document_general/collect_info.php
```diff
@@ -18,21 +18,21 @@
   'products_model' => '',
   'products_image' => '',
   'products_price' => '0.0000',
-  'products_virtual' => DEFAULT_DOCUMENT_GENERAL_PRODUCTS_VIRTUAL,
+  'products_virtual' => DEFAULT_DOCUMENT_PRODUCT_PRODUCTS_VIRTUAL,
```
drbyte: Should Document-General be using Document-Product defaults?

scottcwilson:
It's a much bigger fix otherwise.  Are you sure it's worth it to change the DB and add the missing settings?

drbyte:
> It's a much bigger fix otherwise. Are you sure it's worth it to change the DB and add the missing settings?

I'd prefer not to.

How much of an issue is it to share defaults between the 2? @lat9 ?

(If I could get away with it I'd delete all these "other" product-types. I never use them myself, but I know some people do.)

scottcwilson:
@zcwilt mentioned he wanted to get rid of the music stuff.  Me too.  

I will do it if requested, just wanted to use this shortcut for now.

scottcwilson:
BTW this has been a bug since 1.3.0 (if not before).  We're only seeing this now because of PHP 7.2+.

scottcwilson:
I could just hardcode these to 0 if you prefer.

lat9:
My feeling is that the best approach is to do as @scottcwilson indicated above.  Given that the processing has essentially been using empty values _forever_, just hardcode those settings to their 'off' positions.

lat9:
> (If I could get away with it I'd delete all these "other" product-types. I never use them myself, but I know some people do.)
Oh, please don't!  I've got clients that have re-purposed those product types and the built-in alternates provide some basic guidance for plugin authors that are generating new product types.

drbyte:
> I could just hardcode these to 0 if you prefer.

> just hardcode those settings to their 'off' positions.

Okay, yes, let's do that. It's actually the most sensible given what those entries are for.

PR #2574, opened 2019-07-14
Update Admin keepalive module

drbyte:
Now stops the flashing on the title tab after session expires and forced logout.
To disable flashing altogether, set `flashTitle` to false.
To disable the keepalive feature altogether, delete the keepalive.php file from the server.


admin/includes/keepalive_module.php
```diff
@@ -27,6 +27,7 @@
 $(function(){
    $.jTimeout(
     {
+    'flashTitle': true, //whether or not to flash the tab/title bar when about to timeout, or after timing out
```
scottcwilson: I was hoping you would use 

        'flashTitle' : false,

which prevents the flashing.

drbyte: Setting it to `true` flashes it while attention is being drawn to the fact that the session is about to expire.
If it is ignored (ie: they're "away" and therefore it should expire, or they're ignoring it because they're willing to let it expire) then when the session expires, the `destroy()` call will stop the flashing, since the session expiry no longer warrants any attention.

scottcwilson:
Works like a champ!  Thanks.

PR #2579, opened 2019-07-15
Correct duplicate, linked products in mfgrs listing

lat9:
As identified in [this](https://www.zen-cart.com/showthread.php?225669-Item-showing-twice-(or-more)&p=1359616#post1359616) Zen Cart forum posting.

### DIFF
--- a/includes/index_filters/default_filter.php
+++ b/includes/index_filters/default_filter.php
@@ -38,6 +38,8 @@ if (isset($_GET['manufacturers_id']) && $_GET['manufacturers_id'] != '') {
   if (isset($_GET['filter_id']) && zen_not_null($_GET['filter_id'])) {
 // We are asked to show only a specific category
     $and .= "AND p2c.categories_id = " . (int)$_GET['filter_id'] . " ";
+  } else {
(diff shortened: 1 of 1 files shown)


drbyte:
While DISTINCT works, I wonder if there's a tweak to the linking of the `products_to_categories` table that might be even more "correct"?

drbyte:
@Zen4All did you see any of this when you were tweaking the filters?

Zen4All:
Not that I remember, but is people report they do, we should put the DISTINCT back

lat9:
@drbyte is correct; the 'culprit' in the duplication appears to be the clause `LEFT JOIN " . TABLE_PRODUCTS_TO_CATEGORIES . " p2c ON p2c.products_id = p.products_id`.  If a product is in multiple categories, i.e. linked, that will result in multiple copies of the product being displayed.

drbyte:
Does this need backporting?

drbyte:
And does it apply equally to other product-types?

lat9:
Back-porting would be a "good thing" and yes, the change applies in a _similar_ manner to the other index-filters.

PR #2581, opened 2019-07-15
Build in basic version of DaveWest's non-captcha spam detection

drbyte:
This is a reduced version of davewest's non-captcha spam trap catcher, which only handles URL detection and dynamic fieldname for honeypot.

It does NOT handle the slider or other human-detection questions. Those can be added by using the Addon to replace core functionality.


scottcwilson:
Link to original contrib for reference: 
https://www.zen-cart.com/downloads.php?do=file&id=2207

PR #2584, opened 2019-07-16
Fix PHP 7.3 notices during account creation

scottcwilson:


includes/modules/create_account.php
```diff
@@ -55,7 +55,11 @@
   if (ACCOUNT_COMPANY == 'true') $company = zen_db_prepare_input($_POST['company']);
   $firstname = zen_db_prepare_input(zen_sanitize_string($_POST['firstname']));
   $lastname = zen_db_prepare_input(zen_sanitize_string($_POST['lastname']));
-  $nick = zen_db_prepare_input($_POST['nick']);
+  if (isset($_POST['nick'])) { 
+     $nick = zen_db_prepare_input($_POST['nick']);
+  } else {
+    $nick = '';
+  }
```
drbyte: Can all these be 1-line statements instead of if/else-equals-empty-string?
ie:
`$antiSpam = isset($_POST['should_be_empty']) ? zen_db_prepare_input($_POST['should_be_empty']) : '';`

drbyte:
I've removed the extra spaces your editor added.

Issue #2578, opened 2019-07-15
admin, zen_draw_input_field: parameter required not implemented


torvista:
> function zen_draw_input_field($name, $value = '~*~*#', $parameters = '', $required = false, $type = 'text', $reinsert_value = true) {

The "required" parameter does not get used/added.

I did this
`      $field .= ($required ? ' required' : '') . ' />';`

scottcwilson:
Looks like this got broken in 1.3.9.  In 1.3.8, it was used.

drbyte:
Is there any forum discussion about "why" that was changed(removed)?

scottcwilson:
I suspect this is lost in the mists of time.

scottcwilson:
Changed in 1.3.9a:
 * @version $Id: html_output.php 14139 2009-08-10 13:46:02Z wilt $
@zcwilt you don't by chance remember a ten year old change, do you? :)

scottcwilson:
LOL it gets worse. The `required` parameter *is* used in `zen_draw_pull_down_menu`, but if your template sets it to true, in PHP 7.3 you get a notice about the undefined string TEXT_FIELD_REQUIRED.

drbyte:
> LOL it gets worse. The `required` parameter _is_ used in `zen_draw_pull_down_menu`, but if your template sets it to true, in PHP 7.3 you get a notice about the undefined string TEXT_FIELD_REQUIRED.
> 
> (In admin, this define is set to `'&nbsp;<span class="fieldRequired">* Required</span>'`)

I can't find it right now, but I think this has been discussed before.

scottcwilson:
Yikes.  Mucho hacking required to get this to work.  Admin fields were too wide to show asterisk for required fields.  Submitting two PRs - one for catalog side and one for admin side.

torvista:
My "fix" looks pretty puny after all that!!

scottcwilson:
All fixes matter, Steve. :)

scottcwilson:
Please close; this has been resolved.

drbyte:
Merged #2589 and #2590

PR #2573, opened 2019-07-14
Correct document products' editing

lat9:
Fixes #2572

Issue #2572, opened 2019-07-14
document_general/_product: Invalid HTML and PHP Notices


lat9:
Each of the `document_general` and `document_product` types' `preview_info.php` is generating invalid HTML due to the placement of the `</form>` tag.

In addition, saving a `document_general` product throws a slew of PHP Notices (see below) due to database fields that aren't supported by the product type:

```
[14-Jul-2019 09:05:47 America/New_York] Request URI: /zc156posm/gRoan-Dsw-Habit/product.php?cPath=17&action=insert_product, IP address: ::1
#1  require(C:\xampp\htdocs\zc156posm\gRoan-Dsw-Habit\includes\modules\update_product.php) called at [C:\xampp\htdocs\zc156posm\gRoan-Dsw-Habit\product.php:34]
--> PHP Notice: Undefined index: products_quantity in C:\xampp\htdocs\zc156posm\gRoan-Dsw-Habit\includes\modules\update_product.php on line 24.

[14-Jul-2019 09:05:47 America/New_York] Request URI: /zc156posm/gRoan-Dsw-Habit/product.php?cPath=17&action=insert_product, IP address: ::1
#1  require(C:\xampp\htdocs\zc156posm\gRoan-Dsw-Habit\includes\modules\update_product.php) called at [C:\xampp\htdocs\zc156posm\gRoan-Dsw-Habit\product.php:34]
--> PHP Notice: Undefined index: products_model in C:\xampp\htdocs\zc156posm\gRoan-Dsw-Habit\includes\modules\update_product.php on line 26.

[14-Jul-2019 09:05:47 America/New_York] Request URI: /zc156posm/gRoan-Dsw-Habit/product.php?cPath=17&action=insert_product, IP address: ::1
#1  require(C:\xampp\htdocs\zc156posm\gRoan-Dsw-Habit\includes\modules\update_product.php) called at [C:\xampp\htdocs\zc156posm\gRoan-Dsw-Habit\product.php:34]
--> PHP Notice: Undefined index: products_price in C:\xampp\htdocs\zc156posm\gRoan-Dsw-Habit\includes\modules\update_product.php on line 27.

[14-Jul-2019 09:05:47 America/New_York] Request URI: /zc156posm/gRoan-Dsw-Habit/product.php?cPath=17&action=insert_product, IP address: ::1
[...]

PR #2590, opened 2019-07-16
Properly handle form-input $required parameter on admin side

scottcwilson:
Parameter ignored in some cases; fixed display of fields so that when used it would appear to the right rather than below the field. 

Part of the fix for #2578

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -206,7 +206,7 @@
 
     $field .= ' />';
 
-    if ($required) {
+    if ($required && !empty(TEXT_FIELD_REQUIRED)) {
       $field .= '&nbsp;<span class="alert">' . TEXT_FIELD_REQUIRED . '</span>';
     }
     return $field;
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -206,6 +206,9 @@
 
     $field .= ' />';
 
+    if ($required) {
+      $field .= '&nbsp;<span class="alert">' . TEXT_FIELD_REQUIRED . '</span>';
+    }
     return $field;
   }
 
@@ -222,7 +225,7 @@
   }
 
 ////
-// Output a form filefield
+// Output a form file field
   function zen_draw_file_field($name, $required = false, $parameters = '') {
     $field = zen_draw_input_field($name, '', ' size="50" ' . $parameters, $required, 'file');
 
--- a/admin/includes/stylesheet.css
+++ b/admin/includes/stylesheet.css
@@ -8,6 +8,10 @@
 
 @import url("css/bootstrap.min.css");
 /* @import url("https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css"); */
+.form-control {
+  width: 90%;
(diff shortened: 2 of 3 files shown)


admin/includes/stylesheet.css
```diff
@@ -8,6 +8,10 @@
 
 @import url("css/bootstrap.min.css");
 /* @import url("https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css"); */
+.form-control {
```
Zen4All: Why do we need this?

scottcwilson: Otherwise the * would appear on the next line.  You can try it yourself by modifying a simple form like banner manager to add the required flag.

Zen4All: ok. that's valid, and easier than adding a new div or span. 👍

admin/includes/functions/html_output.php
```diff
@@ -206,6 +206,9 @@ function zen_draw_input_field($name, $value = '~*~*#', $parameters = '', $requir
 
     $field .= ' />';
 
+    if ($required) {
+      $field .= '&nbsp;<span class="alert">' . TEXT_FIELD_REQUIRED . '</span>';
```
mc12345678: To maintain html validation, shouldn't `TEXT_FIELD_REQUIRED` be tested to have/contain something before adding the `<span>`?

scottcwilson: An empty span is not a validation error.

admin/includes/stylesheet.css
scottcwilson: Thanks @Zen4All  and BTW, thanks for the huge improvement to how admin pages look in 1.5.6.  Really makes the product look more up to date.

admin/includes/functions/html_output.php
mc12345678: No not an error, but an unnecessary warning.

scottcwilson: Why would you validate the admin side?  And why would you blank out this define?  This seems excessive.

mc12345678: A non-validated admin side page may have problems executing javascript/jQuery.  Validation nowadays can be done within an individual's browser and not require reaching out to provide information/content from the page. 

May be blanked out on either the admin or catalog side (I chose not to duplicate the comments at PR #2589 after first having seen and commenting on this) to instead support some form of such javascript/jQuery or perhaps that all fields have been made required and the excess identifier is found unnecessary, etc...

A few recent examples of checking validation (notably on the admin) include PRs #2572, #2573, and #2513.

scottcwilson:
With the working required parameter, you can see how now required fields are more clearly indicated. 

Before change: 
regular_156

With this change: 
156_plus_required

admin/includes/stylesheet.css
scottcwilson: Would up having to move this stuff to the bottom - it was messing with the menu.

lankeeyankee:
This breaks some page layouts. Perhaps going back to the * would be better than "Required" as it is more compact.
required2
required

PR #2600, opened 2019-07-18
Don't hardcode count in text string

scottcwilson:


admin/index_dashboard.php
```diff
@@ -457,7 +457,7 @@ function drawCharts() {
 
       <!--// Counters graph = populated by javascript  //-->
       <div class="panel panel-default reportBox">
-        <div class="panel-heading header"><?php echo sprintf(TEXT_COUNTER_HISTORY_TITLE, $number_of_visitors_history); ?></div>
+        <div class="panel-heading header"><?php echo sprintf(TEXT_COUNTER_HISTORY_TITLE, $number_of_visitors_history-1); ?></div>
```
drbyte: It's not actually off-by-one. The `-1` isn't needed.
It really does pull the last X number of records defined by that variable, which is set at the top of the file to 15 by default.

PR #2605, opened 2019-07-19
Sanitize IP addresses via filter_var

lat9:
As identified in [this](https://www.zen-cart.com/showthread.php?225679-PHP-Fatal-error-1406-Data-too-long-for-column-ip_address-at-row-1&p=1359685#post1359685) Zen Cart forum thread.

Use the built-in PHP `filter_var` function to fully vet the incoming IP address.  The change specifies the IP filter types to make the processing more obvious.

If an invalid IP is found, set (as current) the IP value to a single dot and issue a notification allowing follow-on code to catch and possibly record the invalid access.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -3683,12 +3683,16 @@ function zen_get_ip_address() {
   /**
    * sanitize for validity as an IPv4 or IPv6 address
    */
-  $ip = preg_replace('~[^a-fA-F0-9.:%/]~', '', $ip);
+    $original_ip = $ip;
+    $ip = filter_var((string)$ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6 | FILTER_FLAG_IPV4);
 
   /**
-   *  if it's still blank, set to a single dot
+   *  If it's an invalid IP, set the value to a single dot and issue a notification.
    */
-  if (trim($ip) == '') $ip = '.';
+  if ($ip === false) {
+      $ip = '.';
+      $GLOBALS['zco_notifier']->notify('NOTIFY_ZEN_ADMIN_INVALID_IP_DETECTED', $original_ip);
+  }
 
   return $ip;
 }
--- a/includes/functions/functions_general.php
+++ b/includes/functions/functions_general.php
@@ -781,12 +781,16 @@
     /**
      * sanitize for validity as an IPv4 or IPv6 address
      */
-    $ip = preg_replace('~[^a-fA-F0-9.:%/,]~', '', $ip);
+    $original_ip = $ip;
+    $ip = filter_var((string)$ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6 | FILTER_FLAG_IPV4);
 
     /**
-     *  if it's still blank, set to a single dot
+     *  If it's an invalid IP, set the value to a single dot and issue a notification.
      */
-    if (trim($ip) == '') $ip = '.';
+    if ($ip === false) {
+        $ip = '.';
+        $GLOBALS['zco_notifier']->notify('NOTIFY_ZEN_INVALID_IP_DETECTED', $original_ip);
+    }
 
     return $ip;
   }


admin/includes/functions/general.php
```diff
@@ -3683,12 +3683,16 @@ function zen_get_ip_address() {
   /**
    * sanitize for validity as an IPv4 or IPv6 address
    */
-  $ip = preg_replace('~[^a-fA-F0-9.:%/]~', '', $ip);
+    $original_ip = $ip;
+    $ip = filter_var((string)$ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6 | FILTER_FLAG_IPV4);
```
mc12345678: Why limit this to only `FILTER_FLAG_IPV6` and `FILTER_FLAG_IPV4` such that when an additional IPV designation is generated this code would need to be modified?

admin/includes/functions/general.php
```diff
@@ -3683,12 +3683,16 @@ function zen_get_ip_address() {
...
 
   /**
-   *  if it's still blank, set to a single dot
+   *  If it's an invalid IP, set the value to a single dot and issue a notification.
    */
-  if (trim($ip) == '') $ip = '.';
+  if ($ip === false) {
+      $ip = '.';
+      $GLOBALS['zco_notifier']->notify('NOTIFY_ZEN_ADMIN_INVALID_IP_DETECTED', $original_ip);
```
mc12345678: Why isn't the notifier being allowed to modify the `$ip` variable by sending it as an editable value?

lat9: Because the function's doing the work; no need to obfuscate debugging with some unknown alterations.

admin/includes/functions/general.php
lat9: Because then it's a known set of flags (remembering the fuss when the defaults for `htmlspecialchars` changed.

drbyte: Is there another IPV spec nearing release?

mc12345678: > Is there another IPV spec nearing release?

Not in a position to know and do not know of one. Looking at the ways to maximize the use and accessibility of this new method without having to touch the code again. Ie. use of an `$options` variable that can be overridden to serve the desire/need of the code section.

For example to prevent a private or reserved range ip address from being used as a capability not as a code change, or to use a reverse logic to only allow a private or reserved range ip address to be returned/used.

> Because then it's a known set of flags (remembering the fuss when the defaults for `htmlspecialchars` changed. 

So then why not assign the value to be returned if the value does not pass the filter?  Doesn't leaving this to the current default also lead to the same concern/confusion in that if the default changes from `false` to perhaps `0`, then it again won't work.

admin/includes/functions/general.php
mc12345678: By offering the value to be modified then the function's response can be honed to support the environment *without* touching the code which would seem desirable.

admin/includes/functions/general.php
lat9: Because the point of filtering the IP address to a known type is to ensure that the space reserved in the database to hold the value is sufficient.  You are overthinking the requirements here, IMO.

admin/includes/functions/general.php
lat9: Not to me; I'd consider a call to this function as the _single-source-of-truth_.  Noting that, upon receipt of the notification, you can do whatever is needed to "hone" the environment.

drbyte:
Is there a need to lengthen the db fields too?
(Most ip-related fields are 45 chars. One is 96.)

lat9:
Assuming that the IP address is properly sanitized, the max-length of an IPv4 address is 15 characters and that for an IPv6 address 39 characters.  I'm not seeing the need to change.

admin/includes/functions/general.php
mc12345678: However, this change has also revised the operation of the function.  Formerly, it ensured that the provided response met the criteria of characters acceptable in an ip address without regard of the type of ip address that was being detected.  

The use of `filter_var` has changed the response to a type of ip address that matches a set of hard-coded criteria or a dot if not acceptable.  That criteria has become IPV4 or IPV6 without the ability to provide back from the function the value that was actually collected, to refine what is considered acceptable or to take action on the value determined even if it met the criteria.  The response no longer provides the detected ip address (made of ip address permissible values) but instead only the ip address that has been allowed through the original filter. Any other desired filter requires code modification affecting upgrade instead of just adding to the system outside of the core code.

Part of what led me to think/consider this was discussion about providing filterable information about "visitors" to the site.  Specifically the discussion was related to spiders; however, this area could provide an additional operator on which to act.

lat9: > However, this change has also revised the operation of the function. Formerly, it ensured that the provided response met the criteria of characters acceptable in an ip address without regard of the type of ip address that was being detected.

Erm, the comment prior to the current clause reads:
```
    /**
     * sanitize for validity as an IPv4 or IPv6 address
     */
```
I take that, as do the built-in users of the function, as ensuring that the IP address returned is going to be either a valid IPv4/IPb6 address or an (uncommented) `.` otherwise.  An interested observer has the opportunity to grab the invalid value to do something with it if needed.

drbyte:
In your forum discussion you went back and forth about including `FILTER_FLAG_IPV6 | FILTER_FLAG_IPV4` ... How relevant is this?

lat9:
IMO, since we've got database storage defined _specifically_ for IPv4 and IPv6 address lengths, we should keep the filter restrictions so that if/when another IP protocol emerges we're not scrambling to increase database storage if needed.

PR #2606, opened 2019-07-19
Check both shipping and shipping id

scottcwilson:
I have seen cases where the first is set but not the second, leading to a notice in PHP 7.3.


includes/modules/pages/checkout_shipping/header_php.php
```diff
@@ -185,7 +185,7 @@
   $quotes = $shipping_modules->quote();
 
   // check that the currently selected shipping method is still valid (in case a zone restriction has disabled it, etc)
-  if (isset($_SESSION['shipping'])) {
+  if (isset($_SESSION['shipping']) && isset($_SESSION['shipping']['id'])) {
```
drbyte: You can skip the first check in this case. The isset() test will start with the outermost, and work its way in.
Plus, this is only checked again in line 199, where only `id` is used.

So just `if (isset($_SESSION['shipping']['id']))` would suffice.

PR #2608, opened 2019-07-20
Handle error string not defined in val_function check

scottcwilson:
The database creation specifies a set of error strings for min max checks.  None of the strings (such as TEXT_MAX_ADMIN_ADDRESS_BOOK_ENTRIES_LENGTH) have been defined, so the check just returns successfully even though the test would fail.  Creating a default check will fix this for the time being.


scottcwilson:
I'd kind of rather change ALL these strings to just a default string - not sure there's a ton of value in a specific error message for each one.  But let's leave that for another PR.  Once done, it also needs to be added to the upgrade.  See issue #2610.

PR #2609, opened 2019-07-20
Set and use length for previews

scottcwilson:
Instead of just using magic numbers 60 and 100, specific constants for storefront and admin and use them for previews.


drbyte:
Should we be more specific that "previews" actually refers to "product review previews"?

scottcwilson:
I'd like to use it for other previews.  Anywhere ellipses are used.

drbyte:
okay, fair.

Perhaps there should be clarity in the config setting that the number refers to words vs characters?

How do you envision the need for a different Admin-specific length? Why not just one setting so the experience is consistent between admin and catalog?

scottcwilson:
We could ... the only reason I didn't is because the hardcoded numbers specified in the current code differ.  In admin/ezpages.php truncates at 100 and shows the ellipses at this length.  In reviews, the cutoff is 60 but ellipses are 100.  I think the latter is confusing - it should be 60 for cutoff and ellipses or 100 for both.  But I'm open to going to a single MAX if you prefer.

drbyte:
Yes, let's use a single max setting. 

> admin/ezpages.php truncates at 100

I don't see this file in this PR. Was that intentional? Why not move to the constant in there too?

scottcwilson:
It hasn't been added because there's no zen_trunc_string in admin.  Waiting for PR #2565 to be merged.

drbyte:
> Waiting for PR #2565 to be merged.

Okay, merged.

PR #2615, opened 2019-07-20
Show currency code rather than "Dollars"

### LINKED ISSUE #2613: Fix hardcoded "DASHBOARD_DOLLARS" define
There is a hardcoded "DASHBOARD_DOLLARS" define in "admin/includes/languages/english/index.php" that assumes all stores would use "Dollars".

Needs to be either changed to something generic or some logic written to  accept other currency names.

scottcwilson:
Fixes #2613


drbyte:
Should the Dollars constant be removed? Is it needed anymore?

scottcwilson:
Darn it, I thought I had done that.  Hang on ...

PR #2616, opened 2019-07-21
V157 install upgrade val function

mc12345678:
Addresses the upgrade aspect of PR #2267.

### DIFF
--- a/zc_install/sql/updates/mysql_upgrade_zencart_157.sql
+++ b/zc_install/sql/updates/mysql_upgrade_zencart_157.sql
@@ -46,6 +46,29 @@ UPDATE countries set address_format_id = 5 where countries_iso_code_3 in ('ITA')
 
 DELETE FROM configuration WHERE configuration_key = 'ADMIN_DEMO';
 
+#val_function update for MIN values
+UPDATE configuration SET val_function = '{"error":"TEXT_MIN_ADMIN_FIRST_NAME_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}' WHERE configuration_key ='ENTRY_FIRST_NAME_MIN_LENGTH';
+UPDATE configuration SET val_function = '{"error":"TEXT_MIN_ADMIN_LAST_NAME_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}' WHERE configuration_key ='ENTRY_LAST_NAME_MIN_LENGTH';
+UPDATE configuration SET val_function = '{"error":"TEXT_MIN_ADMIN_DOB_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}' WHERE configuration_key ='ENTRY_DOB_MIN_LENGTH';
+UPDATE configuration SET val_function = '{"error":"TEXT_MIN_ADMIN_EMAIL_ADDRESS_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}' WHERE configuration_key ='ENTRY_EMAIL_ADDRESS_MIN_LENGTH';
+UPDATE configuration SET val_function = '{"error":"TEXT_MIN_ADMIN_STREET_ADDRESS_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}' WHERE configuration_key ='ENTRY_STREET_ADDRESS_MIN_LENGTH';
+UPDATE configuration SET val_function = '{"error":"TEXT_MIN_ADMIN_COMPANY_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}' WHERE configuration_key ='ENTRY_COMPANY_MIN_LENGTH';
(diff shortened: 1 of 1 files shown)


scottcwilson:
Questions: 
a) Are we sure we want separate error messages for each of these checks?   This seems unnecessary to me; one generic message would be easier to maintain.  (I added TEXT_DATA_OUT_OF_RANGE in #2608.)
b) If we really do, these strings all need to be defined - TEXT_MIN_ADMIN_FIRST_NAME_LENGTH, etc. are currently not defined.

drbyte:
I agree that it would be best to not create actual defines for "every" one of these, since they're going to be mostly the same as each other, and that's just busywork to maintain.

However, there may be a few legit cases where specific help text would be beneficial, in which case those language defines could(should) be added. Probably to just the admin configuration lang file, so easily self-contained.

mc12345678:
Perhaps a little back story from my perspective which hopefully can be flushed out by others more in the know for those reviewing this now and in the future.  The val_function field was added to the configuration table in ZC 1.5.6.  From what I recall seeing it was considered necessary as a feature developed in the development version of ZC.

At the time of inclusion to ZC 1.5.6, only one val_function field was populated and it was populated using the `configuration_key` as a part of the error message.  So, it is that initial incorporation that seemed to lay a groundwork of which direction to go.  One sample to potentially indicate how all the others would be populated.

Ok, so one example.  Something that possibly hadn't been thought out for "mass" application.  I think that's where we are with this at the moment.

Move forward a little and in addressing issues in the forum a separate class was developed to support use of the `filter_var` function in not so straightforward cases such as `FILTER_VALIDATE_INT`.  To keep the definitions associated with that usage/class packaged, a separate language file was created which freed up the usage a little. In there, a more consolidated "one-liner" was created and it is/was unrelated to existing code/database entries.

Now we are at the point of implementation on a broader scope.  Do I think that the message could be identified with more of a generic statement? (Exceeds maximum, is below minimum, is not an integer/whole number, etc...) Yes. Do I think that it could be handled with "all" variables defined? Sure, though would suggest a more consistent naming convention to possibly help out such that something like the content up to the first three underscores is reviewed and if it matches some predefined filter then such generic statement is used and the remainder of the error provides some sort of useful information to help with the remainder of the message. Or anything in between.

Let me just say that I was concerned about introducing almost anything especially since the existing error message definition would likely need to be updated/modified to follow this "pattern" which will affect all such systems and whatever may have been done to benefit from the process that wasn't fully implemented.  But, that's perhaps the small stuff. :)

I think it would be nice if the error message process would better support the use of variable substitution.  Not sure if that could be better offered by use of a notifier or some other method. This way the message could be generic but have some piece of information provided back to the user.

Note, that one thing I had considered was to try sanitize anything coming out of the database/entry in attempts to not have something executed just because an improper entry was made if that makes sense.  If a problem with that is/has been found, I'd be glad to make adjustments/changes if they aren't independently offered up.

I am for a more generic/action specific type message rather than a slew of defines likely having the same error message content.

scottcwilson:
Note issue #2632 has been entered to be sure these are added (however it's decided to do it).

mc12345678:
> However, there may be a few legit cases where specific help text would be beneficial, in which case those language defines could(should) be added. Probably to just the admin configuration lang file, so easily self-contained.

Now that there is more of a one message for all errors if an error message was at least referenced but not defined, this perhaps is easier. The original implementation was a sort of, no defined error message then not going to use the filter process...

zc_install/sql/updates/mysql_upgrade_zencart_157.sql
```diff
@@ -46,6 +46,87 @@ UPDATE countries set address_format_id = 5 where countries_iso_code_3 in ('ITA')
...
+UPDATE configuration SET val_function = '{"error":"TEXT_MAX_ADMIN_DISPLAY_SEARCH_RESULTS_EZPAGE_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}' WHERE configuration_key ='MAX_DISPLAY_SEARCH_RESULTS_EZPAGE';
+
+#val_function update for email addresses
+UPDATE configuration SET val_function = '{"error":"TEXT_EMAIL_ADDRESS_VALIDATE","id":"FILTER_CALLBACK","options":{"options":["configurationValidation","sanitizeEmail"]}}' WHERE configuration_key ='STORE_OWNER_EMAIL_ADDRESS';
+UPDATE configuration SET val_function = '{"error":"TEXT_EMAIL_ADDRESS_VALIDATE","id":"FILTER_CALLBACK","options":{"options":["configurationValidation","sanitizeEmail"]}}' WHERE configuration_key ='EMAIL_FROM';
+UPDATE configuration SET val_function = '{"error":"TEXT_EMAIL_ADDRESS_VALIDATE","id":"FILTER_CALLBACK","options":{"options":["configurationValidation","sanitizeEmail"]}}' WHERE configuration_key ='SEND_EXTRA_ORDER_EMAILS_TO';
+UPDATE configuration SET val_function = '{"error":"TEXT_EMAIL_ADDRESS_VALIDATE","id":"FILTER_CALLBACK","options":{"options":["configurationValidation","sanitizeEmail"]}}' WHERE configuration_key ='SEND_EXTRA_CREATE_ACCOUNT_EMAILS_TO_STATUS';
```
drbyte, line 120: change to this STATUS entry was reverted in #3346

Issue #2217, opened 2019-03-17
Customer Interface Features needed in core


scottcwilson:
- Ask A Product Question should be part of the Zen Cart core.
- Contact Us and Ask A Question should include a phone number in addition to a name and email address. 

These are pretty basic features and Zen Cart users shouldn't have to do work to add them.

torvista:
Agreed. Long since implemented in my shop.

scottcwilson:
This is (mostly) done.  Closing.

PR #2618, opened 2019-07-21
Add telephone to contact us

scottcwilson:
Telephone is indicated as a required field, but it's really not because older templates might not provide it, and I didn't want to break them.  

This is the first part of #2217.


drbyte:
> Telephone is indicated as a required field, but it's really not because older templates might not provide it, and I didn't want to break them.

Note to my future self: the HTML in the default template requires it, in all modern browsers, but for the sake of older templates that haven't added the field, the PHP isn't forcing the field as a requirement.

drbyte:
Actually, while it's convenient to default it to required, I think the storeowner should be forced to decide to make it required ... so we should default it to not-required in the template HTML.

PR #2619, opened 2019-07-22
Restructure body creation for readability

scottcwilson:


admin/salemaker.php
```diff
@@ -258,16 +258,13 @@ function SetCategories() {
...
-      <?php
-    } else {
-      ?>
-    </head>
-  <body onload="SetFocus(); init()">
-            <?php
-          }
-          ?>
+<?php } ?>
+<?php if (($action == 'new') || ($action == 'edit')) { ?>
```
drbyte: Let's get rid of the unnecessary parentheses.

PR #2628, opened 2019-07-23
Restructure payment page to make it easier to modify

scottcwilson:
Multiline statements with embedded terniary operators and switching PHP on and off makes this file incredibly hard to read or modify.  Relaxing the code standard about having embedded HTML elements  leads to a much simpler structure which will be easier to customize.

### DIFF
--- a/includes/templates/template_default/templates/tpl_checkout_payment_default.php
+++ b/includes/templates/template_default/templates/tpl_checkout_payment_default.php
@@ -14,198 +14,160 @@
 ?>
 <?php echo $payment_modules->javascript_validation(); ?>
 <div class="centerColumn" id="checkoutPayment">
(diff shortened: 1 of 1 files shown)


drbyte:
What about the scenario where one wants to insert extra HTML in between a bunch of these outputs? That's the reason why the PHP on/off stuff was implemented.
Changing it means that it's harder for templaters to style the page to suit display requirements.

scottcwilson:
To me, this is a much easier starting point than the existing one.  It's much easier to follow what's going on with this smaller file.

drbyte:
After talking with @zcwilt, we're gonna close this.

zcwilt:
my downvote here is based on editors handling the splitting of html/php better.

scottcwilson:
OK but can you submit a PR to reformat the file?  The current one is quite hideous.

Issue #2480, opened 2019-06-24
"classic" folder should be deprecated


scottcwilson:
The "classic" folder exists to indicate to the user which folders can be overridden by template files. 

We could allow "responsive_classic" to serve this same purpose, and remove the existing "classic" folders.  

**Remove**: 
./includes/languages/classic
./includes/languages/english/classic
./includes/languages/english/extra_definitions/classic
./includes/languages/english/html_includes/classic
./includes/languages/english/modules/order_total/classic
./includes/languages/english/modules/payment/classic
./includes/languages/english/modules/shipping/classic
./includes/modules/classic
./includes/modules/sideboxes/classic
./includes/templates/classic
./includes/index_filters/classic  (assuming #2629 is merged)

**Create**: 
./includes/languages/responsive_classic
./includes/languages/english/modules/order_total/responsive_classic
./includes/languages/english/modules/payment/responsive_classic
./includes/languages/english/modules/shipping/responsive_classic
./includes/modules/sideboxes/responsive_classic
./includes/index_filters/responsive_classic

**Already Present**: 
./includes/languages/english/extra_definitions/responsive_classic
./includes/languages/english/html_includes/responsive_classic
./includes/languages/english/responsive_classic
./includes/modules/responsive_classic
./includes/templates/responsive_classic

torvista:
My view is that the example template folders should never be touched....they should be copied.
So how about naming them appropriately, like "example_responsive1", "example_classic".
Maybe retire classic once and for all...what purpose does it serve that responsive does not?

lat9:
Just remember, too, that there's _guidance_ sprinkled in various wikis and readme.txt files that identify that template-override locations are identified by the inclusion of `/classic` sub-directories in various spots.

wiztechinc:
No! Not everyone wants to use that dang responsive code. Most existing templates out there don't. I don't and having a default normal template base to use to prove problems or lack of problems is definitely muddied by the complicated responsive.  Just keep it.

dbltoe:
IMHO, that's what I used to call "clinging tenaciously to the trailing edge of technology."  With SE ranking being affected by a lack of responsive sites, I would prefer that the respsonsive_classic become just classic.  Maybe use the ZCA or Color change the responsive_classic.
Keep template_default as is and, if wanted, add the ZCA Bootstrap.
We will provide a far better out-of-the-box product.

scottcwilson:
[...]

PR #2629, opened 2019-07-23
Indicate that includes/index_filters folder can be templated

scottcwilson:
Creating the `classic` folder is the hint we are using currently.  Note that this will change if PR #2480 is implemented. 

Using templates in the `includes/index_filters` is done via the `zen_get_index_filters_directory` function.


drbyte:
Is this really necessary?

scottcwilson:
Just wanted to be consistent with the standard of identifying folders which can be templated.  Even after >10 years of working with Zen Cart, I didn't realize you could do this, so I skipped this folder on an upgrade.

drbyte:
Point taken. 

But not really wanting to add another "classic" folder to maintain, or even just delete later.

PR #2301, opened 2019-04-25
Update status categories products

torvista:
when changing of status of a category, offer choices for status changes of subcategories and products contained therein.
as discussed
https://github.com/zencart/zencart/issues/2280

PR #2630, opened 2019-07-23
Display product options when enabling disabled category

mc12345678:
See bug report: https://www.zen-cart.com/showthread.php?225704-v156c-cannot-reenable-all-product-when-enabling-category&p=1359899#post1359899

Ref: #2301


mc12345678:
Note that affects ZC 1.5.6c also.

PR #2633, opened 2019-07-23
Consolidate common database functions

scottcwilson:
This continues the consolidation started in #2565 so that shared functions have a single definition. 

If you like this name I can rename `functions_general_shared.php` to `general_common.php` in this same PR.


drbyte:
Can the admin database functions file be moved completely to catalog side? and just continue calling it `database.php`, but admin and catalog both call the same file?

scottcwilson:
We could, but isn't it more self documenting if they stay where they're used?

drbyte:
> We could, but isn't it more self documenting if they stay where they're used?

I hear your point, but if it's for the sake of just 1 or 2 functions, and they're very similar to others (ie the lang-related one) then I'd probably find it more efficient to look to a single file rather than wonder if I need to look at the admin side vs the catalog side to find the file.

scottcwilson:
Will do.

PR #2636, opened 2019-07-23
Refactor and de-duplicate vars in attributes module

drbyte:
There's essentially no core functionality changes here.

Granted, it does convert a string-based boolean check to actual boolean for 2 vars, affecting the template slightly.


drbyte:
There's essentially no core functionality changes here.

Granted, it does convert a string-based boolean check to actual boolean for 2 vars, affecting the template slightly.

Issue #2586, opened 2019-07-16
Get rid of old PHP_VERSION checks for 5.5 and below


scottcwilson:
Candidates for cleanup
./admin/includes/functions/general.php
./admin/includes/header.php
./admin/sqlpatch.php
./includes/application_top.php
./includes/classes/class.zcPassword.php
./includes/extra_configures/enable_error_logging.php
./includes/functions/password_funcs.php
./includes/modules/payment/square.php
./ipn_main_handler.php

Instead we can do this in application_top for catalog and admin: 
    if (version_compare($phpVersion, '5.6.0', '<')) die();

drbyte:
How about the scenario where someone is upgrading from an older ZC version, which requires an older PHP version, and staging it on the same server? If you force a die() (or remove critical functionality that keeps it secure like password hashing support for older versions), then "staging" becomes very complicated.

scottcwilson:
They use php.ini to set the PHP version for the updated directory.

drbyte:
Do you envision just changing the code, or are you planning to also do thorough testing that nothing is broken by the removal of logic that is nested around/amongst the version-checking code?

scottcwilson:
I really try to test changes, Chris.  It's a best practice, after all.

drbyte:
Sorry, that came out wrong. I was in the middle of another conversation while typing. I forgot to write the important part: there are some complex things in there, inside some complex parts of the code, and some of them are difficult to test (and I don't have time to do that kind of testing right now), particularly if it involves payment modules. (Although I don't think the Square stuff is relevant.)

It's a good idea. Just gotta handle-with-care.

Would also need to make sure zc_install is very aggressive about making it clear that nothing (in catalog/admin) will run if the PHP version is too old.

scottcwilson:
I could do the easiest and most obvious ones first so people could carefully check them in bite sized pieces.  It's true that some of them are "controversial."  :)

scottcwilson:
Files that need attention: under includes: 
includes/functions/password_funcs.php
includes/extra_configures/enable_error_logging.php
.includes/classes/class.zcPassword.php
includes/modules/payment/square.php
includes/application_top.php
includes/extra_configures/enable_error_logging.php

under admin:
admin/sqlpatch.php
admin/includes/header.php
admin/includes/functions/general.php

drbyte:
Is this completed?

scottcwilson:
No there are a few more I believe.  I will check.

PR #2637, opened 2019-07-23
Removing old PHP Version checks

scottcwilson:
Part 2 of #2586.  

Note that some of the removed checks were actually wrong; they used OR where they should have used AND.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -907,11 +907,8 @@ function zen_get_uprid($prid, $params)
 
 ////
 // Sets timeout for the current script.
-// Cant be used in safe mode.
   function zen_set_time_limit($limit) {
-    if (!get_cfg_var('safe_mode')) {
-      @set_time_limit($limit);
-    }
+     @set_time_limit($limit);
   }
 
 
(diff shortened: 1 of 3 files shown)


admin/includes/functions/general.php
```diff
@@ -909,7 +909,7 @@ function zen_set_product_status($products_id, $status) {
 // Sets timeout for the current script.
 // Cant be used in safe mode.
   function zen_set_time_limit($limit) {
-    if (version_compare(PHP_VERSION, 5.4, '>=') || !get_cfg_var('safe_mode')) {
+    if (!get_cfg_var('safe_mode')) {
```
drbyte: do we even need the `safe_mode` check anymore?

scottcwilson: I do see it once in a while.  (Of course I tell the client to switch, but I do see it.)

drbyte: But, in the sqlpatch update you removed it. 
Let's be consistent and remove it in both places. It was deprecated in PHP 5.3 and removed in PHP 5.4

scottcwilson: LOL you're right.  So much for keeping track of time.  OK removed.

PR #2647, opened 2019-07-25
Consolidate zen_get_uprid

mc12345678:
Now that the admin and catalog finally provide the same
uprid on the admin as it does on the catalog and there is
a consolidation available, this is the simplified (refactored)
version of the function used on both catalog and admin side.

### DIFF
--- a/includes/functions/functions_general_shared.php
+++ b/includes/functions/functions_general_shared.php
@@ -82,6 +82,31 @@
 
 
 /**
+ * Return a product ID with attributes hash
+ * @param string|int $prid
+ * @param array|string $params
+ * @return string
+ */
+  function zen_get_uprid($prid, $params) {
+    $uprid = $prid;
+    if (!is_array($params) || strstr($prid, ':')) return $prid;
+
+    foreach($params as $option => $value) {
+      if (is_array($value)) {
+        foreach($value as $opt => $val) {
+          $uprid .= '{' . $option . '}' . trim($opt);
+        }
+      } else {
+        $uprid .= '{' . $option . '}' . trim($value);
+      }
+    }
+
+    $md_uprid = md5($uprid);
+    return $prid . ':' . $md_uprid;
+  }
+
+
+/**
  * Return a product ID from a product ID with attributes
  * Alternate: simply (int) the product id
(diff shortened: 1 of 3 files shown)


drbyte:
I notice the generated output no longer includes `(` and `)` parentheses on the admin side. Thoughts?

includes/functions/functions_general_shared.php
```diff
@@ -81,6 +81,31 @@ function zen_round($value, $precision) {
...
+/**
+ * Return a product ID with attributes hash
+ * @param string|int $prid
+ * @param array|string $params
+ * @return string
+ */
+  function zen_get_uprid($prid, $params) {
+    $uprid = $prid;
+    if (!is_array($params) || strstr($prid, ':')) return $prid;
+
```
scottcwilson: $lb = ''; $rb = ''; 
if (IS_ADMIN_FLAG) { 
   $lb = '('; 
   $rb = ')'; 
}

scottcwilson:
Could just do a check for  IS_ADMIN_FLAG create left_bracket and right bracket characters (empty string on catalog side, brackets on admin side).   Then use them in setting $uprid.

scottcwilson:
Still a consolidation win!  Thanks for doing this @mc12345678.

drbyte:
Hmmm ... looks like the function is never used admin-side anyway.

lat9:
> 
> Hmmm ... looks like the function is never used admin-side anyway.

I was just going to post that.  I'll note, though, that I've got a couple of plugins that _do_ use that function on the admin side.

drbyte:
> I was just going to post that. I'll note, though, that I've got a couple of plugins that _do_ use that function on the admin side.

👍 
Yes, that's why I decided to merge, since this makes it still available in both places.

mc12345678:
I would have used the function years ago for an application, but because some other existing plugins happily used it causing elimination of the hash portion of the record I couldn't tell if the uprid was on its way in and not fully incorporated or on its way out and being phased. 

As far as the single quotes that were around the content in the admin, I have yet to see how it had any functional effect though it seemed to make it easier to read when following changes from ZC version to version.

PR #2649, opened 2019-07-25
Correct messaging for admin uploads

lat9:
1.  The `ERROR_FILE_TOO_BIG` message incorrectly points the admin to the image settings (should be Maximum Values)
2.  Update the `ERROR_FILETYPE_NOT_ALLOWED` to point the admin to the same set of settings (you'd never find them, otherwise!).


scottcwilson:
Good catch, Cindy!

PR #2650, opened 2019-07-25
Allow items with a single valued attribute to be added from listing page

scottcwilson:
Examples would be products with PDF documents or Zip files.


includes/functions/functions_lookups.php
```diff
@@ -270,6 +270,63 @@ function zen_has_product_attributes($products_id, $not_readonly = 'true') {
...
+      }
+      // if there's any type from $noSingles, we fail
+      if (in_array($field['options_type'], $noSingles)) {
+        $fail = true;
+        break;
+      }
+    }
+
+    // return 1 to indicate selections must be made, so a more-info button needs to be presented
+    if ($fail) return 1;
```
drbyte: Looking at this code now, the `$fail = true; break;` can both be replaced with just `return 1;`, and then this `if($fail)return 1` can be removed.
Maybe incorporate the comment by changing ", we fail" to "we return 1 so selections can be made".

drbyte:
Oops - somehow you added credit card installments files 👀

scottcwilson:
ack sorry that was for debugging.

mc12345678:
No handling for other product types in `zen_requires_attribute_selection`?

scottcwilson:
Your PR is welcome if you can see a miss.

drbyte:
> No handling for other product types in `zen_requires_attribute_selection`?

It's not product-type specific anyway.

PR #2651, opened 2019-07-25
Add observer hooks, and field-data-properties to Attributes module

drbyte:
Observer hooks can be used to set `data-x="y"` properties on html fields, which can later be leveraged by front-end clients to control display/selection dependencies.


drbyte:
Note: some hooks similar to SBA mod are included, making it possible to leverage some of that functionality if using the supplied data-properties doesn't work better.

PR #2653, opened 2019-07-25
First round of consolidating price functions

scottcwilson:
More opportunities exist but these are the most straightforward ones. 

Note: use the gear and check "hide whitespace changes."

### DIFF
--- a/admin/includes/functions/functions_prices.php
+++ b/admin/includes/functions/functions_prices.php
@@ -7,148 +7,7 @@
  * @version $Id: lat9 2019 Apr 05 Modified in v1.5.6b $
  */
 ////
-//get specials price or sale price
-  function zen_get_products_special_price($product_id, $specials_price_only=false) {
-    global $db;
-    $product = $db->Execute("select products_price, products_model, products_priced_by_attribute from " . TABLE_PRODUCTS . " where products_id = '" . (int)$product_id . "'");
-
-    if ($product->RecordCount() > 0) {
-//  	  $product_price = $product->fields['products_price'];
-      $product_price = zen_get_products_base_price($product_id);
-    } else {
-      return false;
-    }
-
-    $specials = $db->Execute("select specials_new_products_price from " . TABLE_SPECIALS . " where products_id = '" . (int)$product_id . "' and status='1'");
-    if ($specials->RecordCount() > 0) {
-//      if ($product->fields['products_priced_by_attribute'] == 1) {
-        $special_price = $specials->fields['specials_new_products_price'];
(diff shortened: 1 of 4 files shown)


drbyte:
I'm concerned about this.

Why all the if-function-doesn't-exist checks? That's a recipe for trouble.
Why allow "the same function to exist in two places but do something different in each"? Also a recipe for pain.

scottcwilson:
Each of those consolidations is going to be a lot of work and testing.  And we *currently* have many functions in 2 places that do something different in each - no change there.

scottcwilson:
Also we could avoid those checks by having a separate shared file, but you didn't want that.

drbyte:
Right. I don't want duplicates. If we're going to consolidate, then the functions should yield the same outcome.

The only time it should do something "different" is perhaps in the case of checking something in the shopping-cart, which of course doesn't exist admin-side.

scottcwilson:
This gets us part of the way there.  If you prefer, I can restructure so that there's a shared file and no if defined checks; that would also get us part of the way there (with lower risk).  Do you want me to do that?   I can't do the whole thing you're looking for in one PR.

drbyte:
I almost want this done function-by-function, because code-reviewing requires manually comparing the original Admin code against whatever has been added/changed/adjusted to move it to catalog-side.

But I'm also mindful that I already did all this in v2.

scottcwilson:
Before we do the "duplicated but different" why don't we start with the "duplicated and identical" and just put them in a shared file like I did for general?  That would save a ton of time.  Then we can go function by function.

scottcwilson:
Current v2 dups:
zen_draw_checkbox_field
zen_draw_file_field
zen_draw_form
zen_draw_hidden_field
zen_draw_input_field
zen_draw_label
zen_draw_password_field
zen_draw_pull_down_menu
zen_draw_radio_field
zen_draw_selection_field
zen_draw_separator
zen_draw_textarea_field
zen_hide_session_id
zen_href_link
zen_image
zen_image_OLD
zen_image_button
zen_image_submit

A pretty short list.  Do you want to just pull in v2 functions, and apply the changes since 1.5.5?

scottcwilson:
To be clear, the above is something I'm volunteering to do if you will approve the PR.

scottcwilson:
If we decide not to move ahead with this, perhaps we should back out #2565 and #2633 so that the job isn't half done.  In the future we could evaluated backporting the function consolidation from 1.6.0 in a single PR well in advance of a release to allow sufficient testing.

zcwilt:

PR #2659, opened 2019-07-26
Fix the Github links in changed files doc

scottcwilson:


drbyte:
Argh - I thought I did that before releasing `c`! Doh.

scottcwilson:
It's AMAZINGLY hard to see bugs like this no matter how long you stare at it.  Maybe mark for backport just in case there's a 156d.

PR #2664, opened 2019-07-27
Consolidate language-defines for Product Model

scottcwilson:
This is PR #623 from 2.0.  Make it easier to change the string "Product Model", which clients do ask for.

### DIFF
--- a/admin/includes/languages/english/coupon_restrict.php
+++ b/admin/includes/languages/english/coupon_restrict.php
@@ -41,7 +41,7 @@ define('TEXT_ALL_MANUFACTURERS_ADD', 'Add All Manufacturer Products');
 define('TEXT_ALL_MANUFACTURERS_REMOVE', 'Remove All Manufacturer Products');
 
 define('HEADER_PRODUCT_STATUS', 'Status');
-define('HEADER_PRODUCT_MODEL', TABLE_HEADING_MODEL);
+define('HEADER_PRODUCT_MODEL', 'Model');
 
 define('ERROR_RESET_CATEGORY_MANUFACTURER', 'Category and Manufacturer filters reset. Use filters individually.');
 
--- a/admin/includes/languages/english/coupon_restrict.php
+++ b/admin/includes/languages/english/coupon_restrict.php
@@ -41,7 +41,7 @@ define('TEXT_ALL_MANUFACTURERS_ADD', 'Add All Manufacturer Products');
 define('TEXT_ALL_MANUFACTURERS_REMOVE', 'Remove All Manufacturer Products');
 
 define('HEADER_PRODUCT_STATUS', 'Status');
-define('HEADER_PRODUCT_MODEL', 'Model');
+define('HEADER_PRODUCT_MODEL', TABLE_HEADING_MODEL);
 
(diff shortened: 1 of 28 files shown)


admin/includes/languages/english/coupon_restrict.php
```diff
@@ -41,7 +41,7 @@
 define('TEXT_ALL_MANUFACTURERS_REMOVE', 'Remove All Manufacturer Products');
 
 define('HEADER_PRODUCT_STATUS', 'Status');
-define('HEADER_PRODUCT_MODEL', 'Model');
+define('HEADER_PRODUCT_MODEL', TABLE_HEADING_MODEL);
```
drbyte: Does this actually work correctly right now? Doesn't the Admin currently load this file before english.php where that constant is defined?

admin/includes/languages/english/specials.php
```diff
@@ -58,4 +57,4 @@
 define('TEXT_INFO_PRE_ADD_INTRO', 'On large databases, you may Manually Add a Special by the Product ID<br /><br />This is best used when the page takes too long to render and trying to select a Product from the dropdown becomes difficult due to too many Products from which to choose.');
 define('TEXT_PRE_ADD_PRODUCTS_ID', 'Please enter the Product ID to be Pre-Added: ');
 define('TEXT_INFO_MANUAL', 'Product ID to be Manually Added as a Special');
-?>
\ No newline at end of file
+?>
```
drbyte: Shouldn't add the `?>` at the end of the file.

scottcwilson: I can fix that, but this file wasn't included in the PR.

admin/includes/languages/english/coupon_restrict.php
scottcwilson: Ugh you're right.  I'll put the string back in here, and then do a new PR to make these defines more like all the rest (TABLE_HEADING_PRODUCTS_MODEL instead of HEADER_PRODUCT_MODEL, etc.)

admin/includes/languages/english/specials.php
drbyte: ... if the file wasn't in the PR, how does it show in the diff?

drbyte: I've removed it.

dbltoe:
Apparently, this commit removed a define for TEXT_PRODUCTS_MODEL which is used on the all products page.

define{'TEXT_PRODUCTS_MODEL', 'Model:'};

Should be added to line 17 to match the 156c version.

It is confusing that both TEXT_PRODUCT_MODEL and TEXT_PRODUCTS_MODEL are both the same but used in different areas of the cart.

scottcwilson:
`TEXT_PRODUCTS_MODEL` is no longer used in the built-in templates.  If you still see it, your branch is not in sync.

dbltoe:
I download from zencart/zencart.  And,, I show nearly twenty files outside of any override that use TEXT_PRODUCTS_MODEL
Case in point - admin/includes/languages/english.php line 469

scottcwilson:
Admin side is different. You were asking about catalog side / templates. 

All uses of TEXT_PRODUCTS_MODEL are on the admin side.

PR #2665, opened 2019-07-27
Address notices on main_page=product_info if product is missing

mc12345678:


drbyte:
Blech. All this `isset()` stuff is ugly.

Your commit title is `if product is missing`. What exactly does "missing" mean in your context?


Is there no way to early-return and just abort when the "product is missing"? 

Again I go back to ... why is the product "missing"?

mc12345678:
> Blech. All this `isset()` stuff is ugly.

Agree. Wasn't sure what would be constituted as an exceptable "bypass" for the "missing" product.

By missing, I had come across the notices, when I first tried a `products_id=0`, then tried with nothing following the equals sign, then tried to access `main_page=product_info` with no `products_id` parameter, then lastly tried with a `products_id` that doesn't exist in the demo data such as 1111111. 

Only one change that was different among the cases was that if `products_id` was not on the Web address, there was the one "assignment" of `$_GET['products_id']` that helped eliminate notifications.

Seems basically, the assignments could be made if there were records returned, otherwise bypass the assignment and all of those `isset()` statements or php 7 null coalesce type statements.

includes/modules/pages/product_info/main_template_vars.php
```diff
@@ -22,7 +22,7 @@
   $product_not_found = $product_info->EOF;
 
   if (!defined('PRODUCT_THROWS_200_WHEN_DISABLED') || PRODUCT_THROWS_200_WHEN_DISABLED !== true) {
-      if ($product_info->fields['products_status'] != 1) {
+      if (!$product_not_found && $product_info->fields['products_status'] != 1) {
```
drbyte: Why this addition?

mc12345678:
Sorry, browser won't let me respond directly to the question:

> Why this addition?

If the product does not exist in the database, then using any specific field from the query results in a notice. Therefore, to omit the notice some sort of test against data identifying the absence of the product is needed. Seeing that the next line was intended to cause the variable `$product_not_found` to go true where it's default is false offered an opportunity to reuse that variable that already seems to have identified that the product does not exist at all in the database.

If there is something more expected to occur in that section regardless of the existence of the product in the database, then perhaps the previous test can be reincorporated. i.e. `!isset($product_info->fields['products_status']) ||`.

There seems to be a lot of code consolidation going on though where reused "long" variables are being "shortened" or assigned just one time instead. In the current state, that's what seemed to be possible here.

includes/modules/pages/product_info/main_template_vars.php
mc12345678: > Why this addition?

If the product does not exist in the database, then using any specific field from the query provides a notice. Therefore, to omit the notice some sort of test against data identifying the absence of the product is needed. 

Seeing that the next line was intended to cause the variable `$product_not_found` to go true where it's default is false offered an opportunity to reuse that variable. At this point in the code, it seems that it has already been identified that the product does not exist at all in the database.

If there is something additional expected to occur in that section (regardless of the existence of the product in the database) then perhaps something similar to the test in the previous commit can be reincorporated. i.e. `!isset($product_info->fields['products_status']) ||`.

There seems to be a lot of code consolidation going on though where reused "long" variables are being "shortened" or assigned just one time instead. In the current state, that's what seemed to be possible here.

drbyte:
I'm guessing these big diffs are gonna need to be done to the other product types too?

mc12345678:
Unfortunately, yes. May still not (yet) need the other `functions_` modifications. I haven't had an opportunity to revisit operation without making those internal checks.

But yes, the `header_php.php` files as well as applicable `main_template_vars.php` files would need to be touched.

Another thought on how to incorporate is that all of the assignments could be within the new grouping of the `else` side of the `$product_not_found` test.  It would mean that there would be no data available for the subsequent template file(s).  I assume though the goal of the recent change was to not totally obscure information, but keep a certain amount "freely" available.

Just offering alternative concepts even though this seems to have been reviewed as acceptable.

scottcwilson:
> I'm guessing these big diffs are gonna need to be done to the other product types too?

ONLY IF WE DON'T DELETE THEM! :)

drbyte:
> could be within the new grouping of the `else` side of the `$product_not_found` test.

I dunno. If the product doesn't exist, then maybe we should just be redirecting, instead of initializing a bunch of empty fields that have no point being passed through to a template that should never be displayed.

includes/modules/meta_tags.php
```diff
@@ -191,6 +191,9 @@
   $review_on = META_TAGS_REVIEW;
   //  case 'product_info':
   case (strstr($_GET['main_page'], 'product_') or strstr($_GET['main_page'], 'document_')):
+  if (empty($_GET['products_id'])) {
+    $_GET['products_id'] = 0;
+  }
```
drbyte: This check/set is done pretty late in the file. Is there any value in doing it earlier?

drbyte: (I'm not bothered if we don't need to do it earlier. Just checking what your thoughts are.)

Otherwise, I'm generally satisfied that we could squash these commits and then merge.

mc12345678: The initial placement was based on where it was known it would be needed (viewing a product's `_info` page).  Having run the "test" that gave all of the notices, I would think that other similar tests should be run with other conditions.  

Unfortunately, making that an exhaustive test seems a bit (unless there is some utility that can be used to simulate all of the potential combinations one at a time. Something I guess I could put together as some code to be fed all of the parameters and some level of conditions to each parameter having both good and bad conditions that would access the index.php repetitively until completion). 

Further, it seemed that if one of the other `case` statements expected to specifically handle `$_GET['products_id']`, that I would have come across some issue at some point and/or the case would have something within it.  The issue there instead being to test what problems occur when it is provided with values in the database and then when not in the database.

Basically, I expect that I will tinker with parameters on the URL regarding known/expected parameters and associated values, though doubt that I will get as creative as has been seen in the wild. At the moment I can't say that the statement would hurt being earlier in processing/evaluation (ie. init_sanitize) or be acceptable to other code.

I can squash the commits and force push if so desired.

PR #2671, opened 2019-07-29
Additional notifiers and valid HTML on previews

lat9:
1.  Add notifiers to each product type's `collect_info` enabling observers to seamlessly add more product-related fields.
2.  Add notifiers to the various `copy_to_confirm` and `update_product` modules, to work with the above additions
3.  Update various `preview_info` modules to product valid HTML

### DIFF
--- a/admin/includes/modules/document_general/collect_info.php
+++ b/admin/includes/modules/document_general/collect_info.php
@@ -235,6 +235,39 @@ for ($i = 0, $n = sizeof($tax_class_array); $i < $n; $i++) {
       ?>
     </div>
   </div>
+  
+<?php
+//-bof-zen_tags-lat9  *** 1 of 1 ***
+    // -----
+    // Give an observer the chance to supply some additional product-related inputs.  Each
+    // entry in the $extra_product_inputs returned contains:
+    //
+    // array(
+    //    'label' => array(
+    //        'text' => 'The label text',   (required)
+    //        'field_name' => 'The name of the field associated with the label', (required)
+    //        'addl_class' => {Any additional class to be applied to the label} (optional)
+    //        'parms' => {Any additional parameters for the label, e.g. 'style="font-weight: 700;"} (optional)
+    //    ),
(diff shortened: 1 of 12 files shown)


admin/includes/modules/document_general/collect_info.php
```diff
@@ -235,6 +235,37 @@
...
+    //    'label' => array(
+    //        'text' => 'The label text',   (required)
+    //        'field_name' => 'The name of the field associated with the label', (required)
+    //        'addl_class' => {Any additional class to be applied to the label} (optional)
+    //        'parms' => {Any additional parameters for the label, e.g. 'style="font-weight: 700;"} (optional)
+    //    ),
+    //    'input' => 'The HTML to be inserted' (required)
+    // )
+    $extra_product_inputs = array();
+    $zco_notifier->notify('NOTIFY_ADMIN_DOCUMENT_GENERAL_COLLECT_INFO_EXTRA_INPUTS', $pInfo, $extra_product_inputs);
```
drbyte: re: `NOTIFY_ADMIN_DOCUMENT_GENERAL_COLLECT_INFO_EXTRA_INPUTS` etc vs `NOTIFY_ADMIN_PRODUCT_COLLECT_INFO_EXTRA_INPUTS`

Would it make sense to use one common notifier name, and read `$product_type` either from a parameter or from the environment?

(My thinking here is that there's likely gonna be a lot of duplication in Observer class code, for each product type. Why not let that code be mostly type-agnostic and use a simple if/else to do things specific to a certain product-type?)

I realize some product types may have a ton of customization, but in those cases these notifiers may be moot because a completely separate product type would make sense.

Thoughts?

lat9: Yes, I was being overly pedantic.  It looks like the `product_type` (**note** not the `products_type`) is available either via posted variables or the $pInfo array passed.

lat9: Consolidation of those notifications now available for review.

PR #2681, opened 2019-07-31
Add dynamic error messages for val_function validations

### LINKED ISSUE #2632: New MINMAX defines missing from language files
Recent PRs have added filter_var checks for some of the integer fields in admin.  However, the error messages use defined text strings which are still missing.  For example: `TEXT_MIN_ADMIN_EMAIL_ADDRESS_LENGTH`, `TEXT_MAX_ADMIN_ADDRESS_BOOK_ENTRIES_LENGTH`, etc.

Note that `TEXT_EMAIL_ADDRESS_VALIDATE` *is* present.

mc12345678:
This adds a generic message to handle the values that are in the
minimum configuration group and the values that are in the
maximum configuration group if a specific error message is not
already defined.

Two methods of processing are presented, my less preferred is
commented out. Less preferred because it seems to involve more
code though perhaps is easier to "read".

Partially fixes #2632 by at least providing a generic define for each, still think
need to address overrides and developer incorporation.

### DIFF
--- a/admin/includes/functions/configuration_checks.php
+++ b/admin/includes/functions/configuration_checks.php
@@ -21,26 +21,12 @@
      // check inputs - error should be a defined constant in the language files
      if (empty($data['error'])) return; 
      if (!defined($data['error'])) {
-        $msg_array = explode('_', $data['error']);
-
-        $str_msg = array();
-
-        foreach ($msg_array as $key => $word) {
-          $str_msg[$key] = implode('_', array_slice($msg_array, 0, $key + 1));
-/*          if ($key == 0) {
-            $str_msg[$key] = $word;
-          } else {
-            $str_msg[$key] = $str_msg[$key - 1] . '_' . $word;
-          }*/
-        }
-
-        $str_count = count($str_msg);
 
         switch (true) {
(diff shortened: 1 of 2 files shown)


admin/includes/functions/configuration_checks.php
```diff
@@ -21,7 +21,31 @@ function zen_validate_configuration_entry($variable, $check_string) {
...
+            $str_msg[$key] = $word;
+          } else {
+            $str_msg[$key] = $str_msg[$key - 1] . '_' . $word;
+          }*/
+        }
+
+        $str_count = count($str_msg);
+
+        switch (true) {
+          case ($str_count >= 3 && $str_msg[2] == 'TEXT_MIN_ADMIN'):
```
mc12345678: I'm a dolt.

This is simply able to be performed by:
````php
case (strpos($data['error'], 'TEXT_MIN_ADMIN') === 0):
````

No need for `explode`, array walking and/or `implode`.

I believe I had some other potential reason for such additional capacity such as finding content within the string at other specified opportunities, but looking at this today, I'm not seeing the need for the extra work/code.

PR #2683, opened 2019-07-31
add sort order to insert

zcwilt:


lat9:
If you're going to add sort orders to the initial install, shouldn't they also be added on an upgrade?

scottcwilson:
No.  They may have already changed the sort orders and we should not modify them.  The database hasn't changed, just the initialization - if it was a new field, you'd be absolutely right.

lat9:
Erm, the `sort_order` _is_ a new field.

lat9:
Updating: I now see the point of not initializing the `sort_order`, in case a store's upgrade already includes the sort-orders ... don't want to overwrite any existing settings.

scottcwilson:
oh sorry - misunderstood.  The db change *is* in the upgrade, but the values are all set to 0.  We should probably let people set their own sorts if they have an existing db.

PR #2685, opened 2019-07-31
Min check needs to be done on every call

scottcwilson:
min error (quantity below config specified min) was not blocking checkout since it was not being rechecked.

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -69,7 +69,7 @@ class shoppingCart extends base {
    * @var string
    */
   var $display_debug_messages = FALSE;
-  var $flag_duplicate_msgs_set = FALSE;
+  var $flag_duplicate_msgs_set = array(); 
   var $flag_duplicate_quantity_msgs_set = array();
   /**
    * constructor method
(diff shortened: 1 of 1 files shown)


includes/classes/shopping_cart.php
```diff
@@ -69,7 +69,7 @@ class shoppingCart extends base {
    * @var string
    */
   var $display_debug_messages = FALSE;
-  var $flag_duplicate_msgs_set = FALSE;
+  var $flag_duplicate_msgs_set = array();
```
mc12345678: Where is this value used as an array? Current processing uses TRUE/FALSE other than these two changes.

mc12345678:
How to reproduce this issue? Not seeing where attempting to checkout with a minimum quantity set and the actual in cart being less than that minimum is allowed.

I tried with both products_id=165 and products_id=53.

Testing was done without incorporation of this suggested change which also would affect how the existing msgs work.

Of course, for some reason I found it unusual that when at the shopping cart with a quantity issue that selecting checkout initially moved forward to the login screen, but then again, a not logged in individual may have something in the stored cart that could complete the quantities. Anyways, once logged in, selecting checkout from the shopping cart did returned to the shopping cart with an error message.

includes/classes/shopping_cart.php
scottcwilson: You're right - this was carried over from 1.5.6 where the logic is different.

scottcwilson:
This might be a 156 only bug.  Let me retest in 1.5.7.

scottcwilson:
Ugh I think this was only a bug in 1.5.6a.  :( Closing.

Issue #2596, opened 2019-07-18
Undefined variable: aInfo in admin1\record_artists.php when no artists


torvista:
>  PHP Notice:  Undefined variable: aInfo in ....zencart-157\admin1\record_artists.php on line 292
> PHP Notice:  Trying to get property 'artists_id' of non-object in ....\zencart-157\admin1\record_artists.php on line 292

Install with no demo data/delete Artists and browse to Admin->Extras->Record Artists

scottcwilson:
@zcwilt How 'bout we get rid of the music stuff in 1.5.7?

zcwilt:
Have no problem with removing, as per v2.

PR #2691, opened 2019-08-01
Remove music

scottcwilson:
Concept approved in #2596.


mc12345678:
@lat9, thought you might be interested in this.

lat9:
Thanks, @mc12345678!  I'll note that I've got a couple of clients that _do_ use `product_music` (what's to become of them?) and that this product-type kind of gives guidance to anyone who's considering creating a new product-type.

scottcwilson:
@lat9 Add back what they use as a mod, and customize their upgrade SQL one time.

zc_install/sql/install/zero_dates_cleanup.sql
```diff
@@ -80,15 +80,3 @@ UPDATE upgrade_exceptions SET errordate = NULL WHERE errordate < '0001-01-01';
 UPDATE zones_to_geo_zones SET last_modified = NULL WHERE last_modified < '0001-01-01';
 UPDATE zones_to_geo_zones SET date_added = '0001-01-01 00:00:00' WHERE date_added < '0001-01-01' and date_added is not null;
 
-UPDATE media_clips SET date_added = '0001-01-01 00:00:00' WHERE date_added < '0001-01-01' and date_added is not null;
```
mc12345678: Shouldn't these remain as part of an upgrade at least? Lest the tables need this fix and get repurposed/revisited (ideally) in transition to this removal? Likely though this will be something to be addressed after the fact considering a void is being created without an alternative substitution first existing.

zc_install/sql/updates/mysql_upgrade_zencart_157.sql
```diff
@@ -35,6 +35,65 @@
 TRUNCATE TABLE whos_online;
 TRUNCATE TABLE db_cache;
 
+# Remove music
+DROP TABLE IF EXISTS `media_clips`;
```
mc12345678: WOW!!!! So, going to just wipe everyone's listings and data??? Isn't this up to the store owner/user? Is this encouragement ***NOT*** to upgrade?

Adding an option to prevent this removal???

scottcwilson: In ten years of doing this, I know exactly ZERO people who are using this stuff.  The small handful of people who actually do use it can tweak this upgrade to not do the removal; it's all in one place and can be removed.    I don't want to lug this stuff around anymore; I want to make room for new useful capabilities that will actually benefit the majority of stores.

lat9: Hold your breath ... I agree with @mc12345678 on this!  It's one thing to not include the tables on an initial install, but to remove them on an upgrade?

Since the majority (I'm just guessing) of Zen Cart store-owners follow the forums more closely than the GitHub, how about posting something on the forums that indicates that the removal is "under consideration" to see how much blow-back is received?

torvista:
> Since the majority (I'm just guessing) of Zen Cart store-owners follow the forums more closely than the GitHub

I would say the vast majority of ZC users never set mouse in Github whatsoever...they are users not developers, so why would they?
Change/modification is one thing, removal is quite another and yes should be well discussed with the _**users**_.

mesnitu:
The music type was a good base to show how to make other product types.

drbyte:
Someone needs to extract all this functionality into a plugin before this can be merged.

scottcwilson:
> Someone needs to extract all this functionality into a plugin before this can be merged.

How about we timebox that?  Merge it in a month. By then, someone who purports to care about this stuff can create a plugin, or not.  You can always go back to Zen Cart 1.5.6 (or any prior version) to find this logic; it's not lost forever.

drbyte:
No.
In the interest of supporting those who create their own product-types, packaging all this into a plugin would serve them better than having it in core anyway.

zcwilt:
I will `commit` to providing the plugin, given that I'm working on the the new plugin infrastructure.
I do agree though that aggressive removal of tables etc. in the update sql is a bad move.

scottcwilson:
It's easy peasy to remove that logic if that's what people want.  But it was exactly what was in V2, so ...

barco57:
is there a problem with checking to see if there are any products assigned to the music product type and if there are skip the removal  when upgrading?

drbyte:
> is there a problem with checking to see if there are any products assigned to the music product type and if there are skip the removal when upgrading? ..and if there are products assigned to the music product type Probably a good idea to let people know in the upgrade process that they will need to install the new plugin for the product type

Both are good ideas. Maybe check that all the related tables are empty before purging.

scottcwilson:
I just moved the clean up logic out to a separate script.  It's not critical to run; people can do this if they wish.

JSWebSteve:
Removal of product type music just because YOU don't know anyone who uses it is rather dictatorial, don't you think? And, to be quite frank, your comment of " I don't want to lug this stuff around anymore" is a rather petty one. The few additional files and database tables this involves is hardly adding a lot of bloat to the core.
What I find most shocking is that you will write in an SQL command do dump entire tables during an upgrade process! That is some real low level skullduggery Scott. 
At the very least you should be checking to see if they have any content first, and maybe offering a checkbox during upgrade that the UNCHECK if they want to drop those tables.
This is not a decision that should be taken in the back rooms of github where the minority of ZC users lurk, but a question that should be openly asked on the main Zen Cart forum. Run a poll on it.

zcwilt:
Just a couple of points here.
Scott has already moved the removal of tables to a separate sql file, so upgrades will NOT remove them. 

As mentioned earlier, there will be a plugin that reproduces the music type functionality and it's highly likely that this will be in the form of a 'core' plugin i.e distributed with the core download.
Then if people want to enable it they can, and if they want to remove the plugin completely they can do that as well. As mentioned elsewhere the plugin system will have options to delete plugins (and old plugin versions) within the plugin admin.
Lets please try and not go down the path of personal attacks when discussing issues on github.

scottcwilson:
Besides,  look on the bright side: once this logic is removed from the core, it will be another piece of code that JSWeb can ionCube encode, sell, and take credit for!   Win-win-win!

barco57:
> Lets please try and not go down the path of personal attacks when discussing issues on github.

...and fail

JSWebSteve:
Indeed!

And my comments and opinions are my own as a freelance dev, not those of JSWeb.

I’ll say no more on the issue as Mr Wilson seems to have taken umbridge.

JSWebSteve:
One last point I will make, JSWeb does not use ionCube encoding. I’ve never installed, written, or modified anything that is ionCube encoded when doing work for them.

zcwilt:

scottcwilson:
Reopen?

PR #2694, opened 2019-08-01
Ensure shipping_module_code value is "sane".

### LINKED ISSUE #2690: Order construction, shipping_module_code set to array?
In [this](https://github.com/zencart/zencart/commit/4bb9ea8d142fdc13c8b83376bf89ee4d7840b138#diff-0c591cf6bf5e115dc78721e77d60d9ea) commit, going back to August 2018, the initial value for the order's `info['shipping_module_code']` is now set to an _array_.  What's the reasoning there?  A shipping-module-code is a string, normally including an `_` and the array value is causing `explode` to throw a hissy-fit on one of my plugins.

lat9:
Fixes #2690.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -383,6 +383,24 @@ class order extends base {
 
     }
 
+    // -----
+    // A shipping-module's 'code', if present in the session, **must** contain a '_' character, separating
+    // the shipping module's name from the selected method, e.g. 'module_method'.  That '_' cannot be the first
+    // character of the 'code' value.
+    //
+    // If that's not the case, issue a PHP Notice and reset the shipping to its unselected state.
+    //
+    if (empty($_SESSION['shipping']['id'])) {
+        $shipping_module_code = '';
+    } else {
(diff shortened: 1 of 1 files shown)


includes/classes/order.php
```diff
@@ -383,6 +383,24 @@ function cart() {
...
+        $shipping_module_code = '';
+    } else {
+        if (!strpos((string)$_SESSION['shipping']['id'], '_')) {
+            trigger_error('Malformed value for session-based shipping module; customer will need to re-select: ' . json_encode($_SESSION['shipping']), E_USER_NOTICE);
+            unset($_SESSION['shipping']);
+            $shipping_module_code = '';
+        } else {
+            $shipping_module_code = $_SESSION['shipping']['id'];
+        }
+    }
```
drbyte: This nested if/else/if/else could be made much simpler and shorter by just setting the string to blank beforehand (replace the outer if/else) and then only keep the inner if/else and remove the then-redundant `= ''` inside it.

PR #2697, opened 2019-08-02
Correct sanitization of UPLOAD_FILENAME_EXTENSIONS

lat9:
... as introduced in #2471 and reported [here](https://www.zen-cart.com/showthread.php?225748-File-Upload-Attribute-not-working-in-v1-5-6c&p=1360358#post1360358) on the Zen Cart forums.

### DIFF
--- a/includes/init_includes/init_db_config_read.php
+++ b/includes/init_includes/init_db_config_read.php
@@ -22,7 +22,7 @@ while (!$configuration->EOF) {
  * dynamic define based on info read from DB
  */
   if ($configuration->fields['configuration_group_id'] == 2 || 
-    $configuration->fields['configuration_group_id'] == 3) { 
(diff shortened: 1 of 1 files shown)


lat9:
Backport suggested.

mc12345678:
What about the associated val_function and error message that now also is wrong and will try to force the entry/change of that option to be an integer?

lat9:
Nice catch!  Are you going to submit a follow-on PR?

mc12345678:
Can, though thought it would be more complete by being a part of this pull (understanding that it is now incorporated and closed).

lat9:
... and I've deleted my branch, so go for it!

mc12345678:
Looks like some "logic" was considered in setting up the val_functions for items in min/max as provided in PR #2267.  Ie. Non-numeric values were skipped, at least this one was.  So, there is no correction needed as a result of the change in this PR.  If anything a final resolution would be needed to provide some level of sanitization other than the default.

Thanks for entertaining. :)

PR #2702, opened 2019-08-04
Fix incorrect last order reporting in customers screen

scottcwilson:
See forum report for this issue at https://www.zen-cart.com/showthread.php?225753-Customer-Summary-Box-showing-first-order-instead-of-last-order

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -1396,16 +1396,17 @@ if (zen_not_null($action)) {
                     $contents[] = array('text' => '<br>' . TEXT_INFO_GV_AMOUNT . ' ' . $currencies->format($customer_gv_balance));
 
                     $contents[] = array('text' => '<br>' . TEXT_INFO_NUMBER_OF_ORDERS . ' ' . $customers_orders->RecordCount());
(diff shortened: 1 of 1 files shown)


admin/customers.php
```diff
@@ -1399,11 +1399,16 @@ function init() {
                     if ($customers_orders->RecordCount() != 0) {
 
                       $lifetime_value = 0;
+                      $last_order = array(
+                        'date_purchased' => $customers_orders->fields['date_purchased'],
+                        'order_total' => $customers_orders->fields['order_total'], 
+                        'currency' => $customers_orders->fields['currency'], 
+                        'currency_value' => $customers_orders->fields['currency_value']);
```
drbyte: Our standard formatting here is to end the array with a comma and new line, putting the closing bracket on the new line. This makes it easier to read and makes future diffs easier to parse since adding another element to the array wouldn't change any existing rows just to add the required comma.

scottcwilson: Noted.  Thanks.

scottcwilson:
suggested for backport

PR #2707, opened 2019-08-05
Enable customer order-confirmation email to be skipped/bypassed

lat9:
This could be desired, for instance, if an admin is placing an order on behalf of a customer.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -1162,8 +1162,7 @@ class order extends base {
     $html_msg['EXTRA_INFO'] = '';
     
     // -----
-    // Enable an observer to indicate that the customer email should not be sent.
-    //
+    // Send customer confirmation email unless observer overrides it.
     $send_customer_email = true;
     $this->notify('NOTIFY_ORDER_INVOICE_CONTENT_READY_TO_SEND', array('zf_insert_id' => $zf_insert_id, 'text_email' => $email_order, 'html_email' => $html_msg), $email_order, $html_msg, $send_customer_email);
     if ($send_customer_email === true) {
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -815,9 +815,9 @@ class order extends base {
           } else {
             $stock_left = $stock_values->fields['products_quantity'];
           }
-          
+
           $products_status_update = ($stock_left <= 0 && SHOW_PRODUCTS_SOLD_OUT == '0') ? ', products_status = 0' : '';
-		
+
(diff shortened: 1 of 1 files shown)


lat9:
... and why the down-vote?

scottcwilson:
I don't like it. I don't think there's a use case, and I think it's astonishing behavior.

drbyte:
I think there's a use-case. There are stores that handle confirmations differently, including letting the payment gateway's email be the confirmation in order to reduce the number of emails sent for a transaction.

lat9:
> I think there's a use-case. There are stores that handle confirmations differently, including letting the payment gateway's email be the confirmation in order to reduce the number of emails sent for a transaction.

Exactly.

drbyte:
@lat9 while it's slightly less intuitive, the current observer can be leveraged for this already, by setting `$email_order = ''` and `$html_msg = array();` before returning.

https://github.com/zencart/zencart/blob/f6bf6bd284c5728eb60efad8c270894684989f16/includes/functions/functions_email.php#L74

lat9:
> 
> 
> @lat9 while it's slightly less intuitive, the current observer can be leveraged for this already, by setting `$email_order = ''` and `$html_msg = array();` before returning.
> 
> https://github.com/zencart/zencart/blob/f6bf6bd284c5728eb60efad8c270894684989f16/includes/functions/functions_email.php#L74

Unfortunately, that wipes the admin-copy of the email, too.

drbyte:
Hmm, ya that'd require some fancy remembering and reinstating of things in NOTIFY_ORDER_INVOICE_CONTENT_FOR_ADDITIONAL_EMAILS which could be ugly.

torvista:
My gateway confirms the order (mostly), but I don't imagine a reason not to send the admin or customer email.
We use the admin copy of the order email exclusively (i.e. we never go into the admin to look for orders, only to update statuses), and I  think the email should be sent whether the customer wants it or not: it's useful to them and us to have in writing the items that are ordered (which may not be what they thought they ordered) , and especially items on backorder for example, when they start moaning that it has not arrived the next day. It's really a confirmation of the "contract" to buy and sell and both parties should have a copy. IMO.

drbyte:
I've worked with storeowners who take a phone order or enter a recurring item, but the customer has requested not to get emails over and over.
While I think the override should be case-by-case and not just a "blanket" thing, this PR allows for no edits to this important core file in order to allow for the override.

torvista:
Always useful to be informed of other workflows.

PR #2710, opened 2019-08-05
missing parameter large_link expected by colorbox and image handler plugins

torvista:
Between this and a Colorbox plugin observer, caused a php notice with this plugin.
Image Handler observer also expects it.


scottcwilson:
Should this be left in the realm of plugins?

torvista:
From what I have chased around the code, I think it should be considered incomplete implementation of the associated notifier.

lat9:
> 
> From what I have chased around the code, I think it should be considered incomplete implementation of the associated notifier.
Yes, I apparently foobar'd that when I originally added the notification.

Issue #2680, opened 2019-07-30
[Suggestion] Admin log in as customer


judygunderson:
This is a mod that is requested A LOT by my customers and I think would be great to be included in future versions of Zen Cart.

wiztechinc:
Since you can now change the customer's password, it's not as necessary but so many issues my clients ask about could be taken care of with this mod. Sure does give a cart owner control when buyers are clueless about  - well, just about everything it seems like. However, I don't install it very often cause I'm all about keeping the number of mods down and we figure out other ways to deal with issues. Cart owners might really like this as a default part of zen cart.

scottcwilson:
Definitely a step saver over using Master Password and logging in that way.

torvista:
Regularly, the crap bank payment gateway I need to use does not complete the order post-payment so I need to do this to complete the order with another payment option, just to record it. 
So, yes I need this feature.

A great deal of discussion and effort could be saved by just importing lat9's entire repository!
There is little/nothing there that is not of use and the code quality is core-quality+.

judygunderson:
I am guessing my customers use it to take credit card phone orders - you know how customers can be - too lazy to do it themselves I guess!  giggle  
Delia - I'm with you on keeping the mods down.  I am constantly discouraging customers from loading their carts with mods they may never or seldom ever use!

lat9:
> 
> 
> Definitely a step saver over using Master Password and logging in that way.

If you're referring to ]Encrypted Master Password](https://github.com/lat9/encrypted_master_password), I don't understand the step-saving.

scottcwilson:
You don't have to type the customer's email or the password.  Just press the place order button.  (It's super minor I know but these little things add up for a busy shop that's processing a lot or orders.)

lat9:
Oh, the functionality that was present in EMP until DrByte requested its removal?

scottcwilson:
Not sure of the history - it might have had the same security issue that ALAC used to have.  At any rate, let's stay on ALAC for this thread please.

drbyte:
I haven't looked at ALAC in awhile, but last time I did I was concerned that it created security issues.
Plus IIRC it was very invasive code (ie: changed a ton of core file logic for the little value it offered).
Maybe it's improved since then.

scottcwilson:
It has!  The security issue was fixed.  https://www.zen-cart.com/downloads.php?do=file&id=583

lat9:
[...]

PR #2711, opened 2019-08-06
#2680: EMP initial integration

lat9:
Initial full-on incorporation into core.


admin/customers.php
```diff
@@ -1370,8 +1370,9 @@ function init() {
                   $contents[] = array('align' => 'text-center', 'text' => '<br><button type="submit" class="btn btn-warning">' . IMAGE_RESET_PWD . '</button> <a href="' . zen_href_link(FILENAME_CUSTOMERS, zen_get_all_get_params(array('cID', 'action')) . 'cID=' . $cInfo->customers_id) . '" class="btn btn-default" role="button">' . IMAGE_CANCEL . '</a>');
                   break;
                 default:
-                  if (isset($_GET['search']))
+                  if (isset($_GET['search'])) {
                     $_GET['search'] = zen_output_string_protected($_GET['search']);
+                  }
```
scottcwilson: one liner maybe?

admin/customers.php
```diff
@@ -1396,17 +1414,17 @@ function init() {
...
                       $last_order = array(
                         'date_purchased' => $customers_orders->fields['date_purchased'],
                         'order_total' => $customers_orders->fields['order_total'], 
                         'currency' => $customers_orders->fields['currency'], 
                         'currency_value' => $customers_orders->fields['currency_value'],
                       );
-                      foreach ($customers_orders as $result) {
-                        $lifetime_value += ($result['order_total'] * $result['currency_value']);
+                      foreach($customers_orders as $result) {
+                         $lifetime_value+= ($result['order_total'] * $result['currency_value']);
```
scottcwilson: seems like your editor added this.

admin/customers.php
lat9: FWIW, except for the case of checking for undefined constants, my preference for processing conditionals is to use the braces to help identify the full clause.  I can't count the number of times where I've missed one of those one-liners when trying to reformat a module.

admin/customers.php
lat9: It might have, but what's the issue with those additional spaces?

lat9: Argh, that must have been Beyond Compare "helping".

PR #2712, opened 2019-08-06
Shopping cart define

wiztechinc:


scottcwilson:
Good job.  Note that your editor has introduced some whitespace changes, so you'll want to tighten that up.

wiztechinc:
Yay! Dang white spaces!

wiztechinc:
I took a look and don't see the white spaces when doing a compare. Can you be specific? I've definitely seen these issues before and if it's due to my default editor, then I need to use another.

scottcwilson:
Your compare tool likely is configured to ignore whitespace changes.
Look at the Files Changed tab.  Look at includes/filenames.php for example.  Lines 32 and 63 should not be changed.

drbyte:
I think the lines Scott pointed out were probably using a wrong line-ending already.

I've updated other files by changing tabs to spaces, to conform with standard.

drbyte:
Thanks Delia!

wiztechinc:
You are welcome and yes, that was the problem. Once I told beyond compare to show the whitespaces, I see what he was talking about. I feel like dancing...............
image

lat9:
@wiztechinc, I'm having the same issue with Beyond Compare.  Where did you find that setting?

wiztechinc:
Kinda funny as it turns out I haven't updated it in years. Still on version 3 but the latest build of that version. 

image

lat9:
Thanks, Delia!

PR #2716, opened 2019-08-07
Add lazy loading (new feature in Chrome)

scottcwilson:


drbyte:
Given that I've seen some glitches with this done elsewhere (in apps even simpler than the extra markup handling that Zen Cart offers), I'd like to ensure some thorough testing has been done across various browsers, to ensure that in a Zen Cart setup this doesn't cause any unexpected outcomes.

scottcwilson:
Sounds good.  We can also sit tight for a while - this is brand new.

barco57:
I can say it does seem premature as even with chrome this is hidden behind a flag
https://caniuse.com/#feat=loading-lazy-attr

drbyte:
Closing for now. We can reopen it when the feature is more widely embraced.

PR #2720, opened 2019-08-09
Hide product from cart until available date..

mc12345678:
This additional/removable code offers an opportunity for a product
to be fully defined in the store, but not shown as available,
upcoming, new, etc (because it is disabled) until the product
available date at which point it appears in the store as an
enabled product ready for the customer to see and/or purchase.


includes/functions/disabled_upcoming.php
```diff
@@ -0,0 +1,38 @@
...
+
+    $date_range = time();
+
+    $zc_disabled_upcoming_date = date('Ymd', $date_range);
+
+    $disabled_upcoming_query = "SELECT products_id
+                                            FROM " . TABLE_PRODUCTS . "
+                                            WHERE products_status = 0
+                                            AND ((products_date_available <= " . $zc_disabled_upcoming_date . "
+                                            AND products_date_available != '0001-01-01'))
```
drbyte: Why the `((`double parentheses`))`?

drbyte:
This PR reads more like a plugin than a core-code component.

mc12345678:
Seeing as just about everything is in question of needing to be in core code versus being an addition, I hesitated on incorporating directly into "harder to edit" core code. (the edit isn't really difficult, just one of those things where it is unknown how others have possibly modified the related code to suit their needs.)

It could be incorporated more as part of the core, the existing could possibly be modified a little to be more adaptable or this could remain outside of the core code and instead be a plugin. As usual, open to thoughtful recommendations.

includes/functions/disabled_upcoming.php
mc12345678: Held over from copy/paste from other areas where an `or` clause was present and needed to have double parentheses.  Not necessary to be present and serves no value.

lat9: Shouldn't there be single-quotes around the first `products_date_available` AND-condition?

mc12345678: Maybe I'm not folllowing.  `products_date_available` is a `datetime` database field.  Presumably such a field does not require quotes in order to reference the database content.  That field is being used as comparison against the formatted string date `$zc_disabled_upcoming_date`.  This is similar in construction as in includes/functions/specials.php, featured.php and salemaker.php (with below code taken from featured):
````php
    $featured_query = "select featured_id
                       from " . TABLE_FEATURED . "
                       where status = '0'
                       and (((featured_date_available <= " . $zc_featured_date . " and featured_date_available != '0001-01-01') and (expires_date > " . $zc_featured_date . "))
                       or ((featured_date_available <= " . $zc_featured_date . " and featured_date_available != '0001-01-01') and (expires_date = '0001-01-01'))
                       or (featured_date_available = '0001-01-01' and expires_date > " . $zc_featured_date . "))
                       ";
````

I did remove the single quotes around the value of `0` being compared against the field `products_status` as it is defined as a `tinyint` instead of a string. Is that perhaps the single quotes that appear to be missing? 

I *have* been pondering the MySQL 8.x issue(s) and whether a check against `IS NOT NULL` should be added in addition to the check against `!= '0001-01-01'`, but then also wondered if I was getting ahead of things with that.

Comment (lat9) [reply includes/functions/disabled_upcoming.php]: My bad (again); it just looked "funny" to have the quotes around the value when checking for equality but not when comparing.

**Comment by drbyte** [comment]
Suggestions:
- Can we move these function definitions into functions_products.php ?
- And then remove the config autoload file by merging these additions into the core autoloader ... perhaps as 150 alongside the sales/specials init. (You've got 149 now, so let's just collapse it into 150 with the other.)
- And then add a config switch for enable/disable it. Maybe group 9-Stock or 24-IndexListing? Or is someplace else better? Open to discussion.

**Comment by mc12345678** [comment]
> Suggestions:
> - Can we move these function definitions into functions_products.php ?
> - And then remove the config autoload file by merging these additions into the core autoloader ... perhaps as 150 alongside the sales/specials init. (You've got 149 now, so let's just collapse it into 150 with the other.)
> - And then add a config switch for enable/disable it. Maybe group 9-Stock or 24-IndexListing? Or is someplace else better? Open to discussion.

Sorry for the delay in response. Good, so bring this into the "routine" code, be glad to. Answers: yes, yes, initially was thinking 9-Stock, but then also 1-My Store (there is product control there relating to out-of-stock action), but still looking/considering.

Is it ok if `$_SESSION['today_is']` is taken from date to datetime aware (perhaps an additional variable could/should be used)? It would knowingly affect sales, specials, and featured product if `$_SESSION['today_is']` is modified as these look to verify that the date has changed. They do not evaluate the current datetime to see if it is greater than a database value. Or is the `datetime` aspect something that should be addressed as part of #2397?

**Comment by drbyte** [comment]
I think it makes sense to have consistency with sales/specials/featured date parsing. No point introducing new variables for accomplishing the same task.

While I hate the use of `0001-01-01` dates/times, I think this isolated feature should be treated on its own, assuming as little variation from core code as possible. Keep it atomic.  Other much bigger issues should be handled separately, otherwise it's impossible to follow the bouncing ball when looking back into history to figure out why/when something was changed or understand something why something might operate differently, etc.

**Comment by mc12345678** [comment]
Completed the move of the files into existing Zen Cart core files, added an admin controllable switch located in group 9 just after the handling of out-of-stock product with the default set to disabled upon installation.  Did *not* adjust the products_available_date check to being `datetime` specifically and updated the `functions_products.php` file to PSR-2.

There are multiple commits, some made to correct mistakes I made while incorporating the new code, but generally as a change to each file and in an add before delete sequence.  If desired to keep more than one commit, then could combine some of them such as simultaneous add and delete together, later edits of new code to be combined with the first time the new code is provided, etc...

I believe this is ready for review at leisure.

A not incorporated consideration was the ability to activate this feature during the install/update process.  I don't yet see the necessity though it may be nice.

File: zc_install/sql/updates/mysql_upgrade_zencart_157.sql [hunk 2]
```diff
@@ -62,6 +62,9 @@ INSERT INTO product_type_layout (configuration_title, configuration_key, configu
...
 INSERT INTO product_type_layout (configuration_title, configuration_key, configuration_value, configuration_description, product_type_id, sort_order, set_function, date_added) VALUES ('Show \"Ask a Question\" button?', 'SHOW_PRODUCT_FREE_SHIPPING_INFO_ASK_A_QUESTION', '1', 'Display the \"Ask a Question\" button on product Info pages? (0 = False, 1 = True)', 5, 14, 'zen_cfg_select_drop_down(array(array(\'id\'=>\'1\', \'text\'=>\'True\'), array(\'id\'=>\'0\', \'text\'=>\'False\')), ', now());
 
+# Add control for enabling product that were disabled and have a products_available_date that has passed in time and is not set to the Zen Cart equivalent of an ignored/empty date.
+INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Enable disabled product by available date', 'ENABLE_DISABLED_UPCOMING_PRODUCT', '0', 'Enable product that had an available date entered and the current date has come to exist or pass.<br /><br />0= off<br />1= on', '9', '12', 'zen_cfg_select_option(array(\'0\', \'1\'), ', now());
```
drbyte: I think the wording for `configuration_title` and `configuration_description` needs adjustment to be more clear about what this is for.

mc12345678: > I think the wording for `configuration_title` and `configuration_description` needs adjustment to be more clear about what this is for.

No disagreement there.

Please review the updated version and if further clarification is desired, welcome discussion.

drbyte: Better, yes. 
Still slightly verbose, and maybe confusing.

drbyte:
Idea, not fully explored yet: would making the config switch options be "automatic" or "manual" allow the description/title to be expressed simpler?

scottcwilson:
Suggested title/description:
"Hide upcoming products until available date" 
"Treat upcoming products with available dates in the future as disabled, so customers cannot see them." 

The default for this should be OFF - most people want to start marketing and get excitement building BEFORE launch day.

torvista:
Late to the party here..so this is a global setting, affecting all products with a future availability date?

mc12345678:
@scottcwilson, 
> The default for this should be OFF - most people want to start marketing and get excitement building BEFORE launch day.

I am thinking that the default for a fresh install would be to have this enabled for automatic, but for an upgrade to be off.  Offers full capability for new user (no need to ask how to make it work, rather would ask how to turn it off if found to be an "issue") and then for those performing a database upgrade would not affect how data was maintained. Albeit upon turning it on from an off status, changes would be made as applicable to the data.

@torvista,
> so this is a global setting, affecting all products with a future availability date?

Yes, when this feature is "on" (recently suggested to be "automatic") then all products that have a future date that are disabled would remain "inaccessible" to the front end until the date has come to be.  At that point the product would become enabled and included in all places that the active product was attached. For those that wish to use "future planning" or teasers of the product while this is *on*/*automatic*, leaving the product enabled with a future date would allow the product to be presented for advertising and as otherwise allowed/setup pre-orders.  This feature may apply more to dated "advertising" or sales allowance that the product (e.g., physical, virtual, printed) becomes immediately available at the point in time designated.

scottcwilson:
> I am thinking that the default for a fresh install would be to have this enabled for automatic, but for an upgrade to be off. 

This is confusing.  Just keep it off for both upgraders and new installers - people who want it will turn it on. 

This is a nice feature, I'm glad you added this capability, but it should be off unless someone wants it.

zcwilt:
Subject to further review we should move this to `'on hold`

mc12345678:
Anything specific that could be addressed/corrected for moving forwards? E.g. Updates to other date related functions to set dates to `null` instead of `0001-01-01`, the use of `null` as the final state, the install VS upgrade paths, anything I can do...

zcwilt:
To be more specific, I had approved this, but want to wait for another approval, but if it's going to be merged it needs to be soon to allow for live testing before release.

mc12345678:
Understand. Thank you very much for the feedback and clarification.

includes/init_includes/init_special_funcs.php
```diff
@@ -7,56 +7,66 @@
...
  * require the banner functions, auto-activate and auto-expire
  */
-require(DIR_WS_FUNCTIONS . 'banner.php');
+require DIR_WS_FUNCTIONS . 'banner.php';
 zen_activate_banners();
 zen_expire_banners();
+/**
+ * require product functions one time such that if previously loaded will not cause an error here.
+ */
+require_once DIR_WS_FUNCTIONS . 'functions_products.php';
```
drbyte, line 29: @mc12345678 
Given that `init_general_funcs.php` already has a `require` for `functions_products.php` which loads at breakpoint `60`, why are we requiring it again here again in `init_speical_funcs.php` which loads at breakpoint `150`?

Ref: https://github.com/zencart/zencart/pull/2551

mc12345678: In essence, to allow such a comment/question/tracking to identify the dependency of these changes on `functions_products.php`... Possibly could have just added a note to indicate that dependency; however, also wanted to be sure that it was actually included at or before the subsequent code in the event the file's load point was otherwise changed.  I recall there being some "adjustments" being made to its (`functions_products.php`) load point and/or inclusion.

As a "final product" I would say that identifying the dependency as a code comment should be sufficient and that the line could otherwise be removed.

PR #2729, opened 2019-08-11
Add images/uploads/ to .gitignore

scottcwilson:


lat9:
Er, isn't there a special .htaccess in /images/uploads?  Won't that .gitignore prohibit its inclusion in the release?

scottcwilson:
Files which are already checked in are ok.

lat9:
... and there will never be a reason to update that .htaccess?

scottcwilson:
Files which are checked in can be updated.  There's no problem.

drbyte:
Hmmm. I was a bit fast on the trigger when merging this.

Generally the best practice is to put a .gitignore file inside the directory you want to exclude files from, and inside that .gitignore simply tell it to keep itself but exclude everything else.
```text
*
!.gitignore
```
I'll move it over.

PR #2733, opened 2019-08-13
Add edit link to product name

### LINKED ISSUE #2724: Click table-row to jump into edit-mode for product details
I noticed in 1.5.6c that you can no longer just click on the product name int the admin to open it to edit it.  You have to go to the right side and click the edit icon.

After all these years of being able to just click the product name... this is mildly irritating, so thought I would bring it up just in case anyone else is mildly irritated by it. :)

I only tested this in Mozilla Firefox... so maybe it is a browser issue????

Zen4All:
On The products listing page a link to edit the product is added to the product name
Fixes #2724

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2733)

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -841,7 +841,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                 ?>
                 <tr>
                   <td class="text-right"><?php echo $product['products_id']; ?></td>
-                  <td><a href="<?php echo zen_catalog_href_link($type_handler . '_info', 'cPath=' . $cPath . '&products_id=' . $product['products_id'] . '&language=' . $_SESSION['languages_code'] . '&product_type=' . $product['products_type']); ?>" target="_blank"><?php echo zen_image(DIR_WS_ICONS . 'preview.gif', ICON_PREVIEW); ?></a>&nbsp;<?php echo $product['products_name']; ?></td>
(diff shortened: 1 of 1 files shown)


drbyte:
I think this is a reasonable compromise. 👍 

The "old" behavior allowed clicking (almost) anywhere on the table-row to invoke the edit.
Then it was changed to just the edit button/icon.
This now allows also clicking on the product name to invoke editing. 
While it's no longer a "click anywhere on the row", I think intuitively most people will aim to click on the product name text anyway, and will easily retrain any muscle memory to find the text to click, and will be grateful that it's no longer just the tiny "e" icon.

scottcwilson:
Thanks for doing this, guys.  This is a big improvement!

mc12345678:
While I like(d) the ability to click anywhere that wasn't a specific icon, should such a change be done to categories for consistency? Or (and preferable though time associated) is there something specific about a product row that is fouling the behavior that could be addressed and the `onclick` event for product be restored?

Page seems to HTML validate (Internet Explorer, Microsoft Edge, Firefox all tested), haven't tried some similar javascript validation and dislike how this comment is likely to be considered some sort of displeasure. 

I don't seem to have an environment that reproduces the problem that was corrected, though was aware of the problem through other forum discussion and that it basically prevented working with product. 

Locally I can take the associated `zen_href_link` code and apply it to the `<tr>` above the current line (using the format of line 637 `<tr onclick="document.location.href = '<?php echo zen_href_link(FILENAME_PRODUCT, 'cPath=' . $cPath . '&product_type=' . $product['products_type'] . '&pID=' . $product['products_id'] . '&action=new_product' . (isset($_GET['search']) ? '&search=' . $_GET['search'] : '')); ?>'" role="button">`) and it works, but it always did so not a good test... :(

Zen4All:
@mc12345678 The problem with the onclick is/was, if I remember correctly, mainly with safari.
The categories work fine, so we should leave that as is. If you think you can get the onclick working again, be my guest 😉 .

mc12345678:
Would love to; however, as said I do not see that I can do so on my own equipment.  Safari can not be installed on a windows computer, I do not have a Macintosh/apple device.  As implied, the commit is a solution that works, while inconsistent with similar other options it provides at least a middle ground/point. Maybe someone that can reproduce the problem can get involved in one way or another even if it is to just identify that they can and a path forward can be made.

Basically, don't wait on me for a solution, at this time not able to reproduce the original problem.

PR #2734, opened 2019-08-13
Contact Us page in 1.5.6+

judygunderson:
I'm not sure if I am doing this right, so I will plunge in and maybe someone can help me if I am doing it wrong.
Can you compare version 1.5.5F to 1.5.6C and 1.5.7 on includes/modules/pages/contact_us/header.php.php around line 54 (1.5.6c) where it says:

// declare variable
            $send_to_array = [];

I am wondering if it should be like it is in 1.5.5F (because I see an error in Dreamweaver) as follows:

// declare variable
$send_to_array = array();

When I change my file to that, the error goes away.


drbyte:
> includes/modules/pages/contact_us/header.php.php around line 54 (1.5.6c) where it says:
> 
> // declare variable
> $send_to_array = [];
> 
> I am wondering if it should be like it is in 1.5.5F (because I see an error in Dreamweaver) as follows:
> 
> // declare variable
> $send_to_array = array();
> 
> When I change my file to that, the error goes away.

You must be using a VERY old PHP version. That short-array syntax was added in PHP 5.4 ... so that tells me you're using PHP 5.3, which expired 5 years ago. https://www.php.net/eol.php

scottcwilson:
Also, you're going to need to abandon this branch, since it has > 900 changes on it. :(

PR #2738, opened 2019-08-17
Display order number in heading

lat9:
Currently, when displaying an order in the admin, the order number _is_ present but in relatively small text within the order's details.

Update the order's detailed display to include the order number in the level-1 heading so that it's more obvious to admin staff.


lat9:
Is there a problem with this change?  I've got a couple of other updates for the orders tool in progress.

lat9:
Thanks, @zcwilt!

PR #2741, opened 2019-08-19
Additional notifications for admin/orders.php

lat9:
- "Simplify" the order-listing's navigation, splitting into 3 equal sections for medium/large display.
- Add a notifier to enable additional 'nav' buttons to be inserted to the left or right of the listing navigation.
- Additional notifiers for the order-status-history block, primarily in support of "Ty Package Tracker"

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -825,7 +825,7 @@ if (zen_not_null($action) && $order_exists == true) {
           //     ),
           // );
           //
-          // Observer note:  Be sure to check that the $p3/$extra_data value is specifically (bool)false before initializing, since
+          // Observer note:  Be sure to check that the $p2/$extra_data value is specifically (bool)false before initializing, since
           // multiple observers might be injecting content!
           //
           $extra_data = false;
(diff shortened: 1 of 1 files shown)


admin/orders.php
```diff
@@ -747,6 +761,31 @@ function couponpopupWindow(url) {
...
+  // A watching observer can provide an associative array in the form:
+  //
+  // $extra_headings = array(
+  //     array(
+  //       'align' => $alignment,    // One of 'center', 'right', or 'left' (optional)
+  //       'text' => $value
+  //     ),
+  // );
+  //
+  // Observer note:  Be sure to check that the $p2/$extra_headings value is specifically (bool)false before initializing, since
```
mc12345678: Based on the note of line 828, should this be `$p3/…` instead of `$p2/…` or some other value as these `p` values are not defined anywhere? Also wouldn't below a test of `!empty($extra_headings)` be more encompassing ($extra_headings gets `unset` for example)?

lat9: Good point; need to remember to review after copy&paste!

PR #2746, opened 2019-08-21
Add automatic login to EMP/Admin login as customer.

zcwilt:
Also see issue #2680


admin/includes/init_includes/init_db_config_read.php
```diff
@@ -9,6 +9,15 @@
 if (!defined('IS_ADMIN_FLAG')) {
   die('Illegal Access');
 }
+
+  $sql = "SELECT configuration_value FROM " . TABLE_CONFIGURATION . " WHERE configuration_key = 'GLOBAL_AUTH_KEY'";
+  $authkey = $db->Execute($sql);
+  if ($authkey->fields['configuration_value'] == '') {
```
mc12345678: Part of what led me to submit #2871 was that I received a notice at this point in execution both before and after incorporating the update sql.  As such while I would like to submit a PR for this change, GitHub is currently trying to calculate the jump-to-point and therefore is not allowing me to do so.

Recommend this line to read:
````php
if (!$authkey->EOF && $authkey->fields['configuration_value'] == '') {
````

mc12345678: Disregard, I was on the wrong repository when attempting to locate the code.

PR #2748, opened 2019-08-21
start to add psr4 class autoloading

zcwilt:
note. Not doing anything with this just yet, but will use it later for some stuff related to plugins etc.


lat9:
Will this update enable a plugin (or site-specific customization) to override the built-in Zen Cart class definitions?

zcwilt:
Well sort of, customisations can be done with in the init_includes system as it stands. This will be come more apparent once I've done some more pr's to cover other areas prior to doing the main plugin overrides PR.

PR #2755, opened 2019-08-22
Refactor initsystem

zcwilt:
refactored to support new plugin system.

### DIFF
--- a/includes/classes/InitSystem.php
+++ b/includes/classes/InitSystem.php
@@ -1,0 +1,284 @@
+<?php
+/**
+ *
+ * @package classes
+ * @copyright Copyright 2003-2019 Zen Cart Development Team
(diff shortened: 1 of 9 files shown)


includes/classes/InitSystem.php
```diff
@@ -0,0 +1,284 @@
...
+        if (file_exists($overrideDir . '/' . $file)) return true;
+        return false;
+    }
+
+    protected function getLoadersFromFilelist($fileList)
+    {
+        foreach ($fileList as $file) {
+            require($file);
+        }
+        return $autoLoadConfig;
```
drbyte: @zcwilt 
This `$autoLoadConfig` seems to be the wrong variable. 

But returning `$fileList` instead is not a correct solution.

zcwilt: $autoLoadConfig is defined in the auto_loaders file that is being required

PR #2766, opened 2019-08-25
Add new tables for plugin management and initial load of installed plugins

zcwilt:
…ugins for admin


lat9:
@zcwilt, I'd received comments from @drbyte on another PR requesting that the defaults for integer-type database fields, if provided, be specified as non-quoted (i.e. integer) values.

drbyte:
@lat9 is referring to things like:

```sql
  status tinyint(1) NOT NULL default '0',
```

PR #2775, opened 2019-08-28
Allow a filename attribute to be added with a missing filename

scottcwilson:
You may want to prepare a download product before the zip file is completely ready.  With this change, it will show up as a red entry in Downloads Manager and can be fixed at a later time.  I have used this step-saver in a few releases and wanted it to be considered as a core change.


scottcwilson:
That change is not required in the current release.  You can set a filename to a file that doesn't exist and everything is fine (red icon appears, which can be fixed later by uploading the file). The testing I did worked because in my config, option 1 *was* a downloadable file, but that's generally not true.

PR #2785, opened 2019-08-30
Revision of Admin gv_mail.php

torvista:
A complete revision of Admin gv_mail.php

admin added tt to css
Modification of gv_mail language constants.
Change of "Dear" from email template to a language constant

gv_mail completely revised, including:
Fixed Back button on Preview page: submitted email
Form values made sticky when Back button used on preview page.
JS amount check expanded to allow decimals only.
PHP error checking corrected to prevent use of preview page with bad values.
Single email address: added optional name field.
Consolidation of duplicated email content build for lists and single email
Hard-coded "Dear" in email template replaced with EMAIL_SALUTATION in /english.php: to be corrected in other emails shortly.
Consolidation and modification of text/HTML email format content.
Simplification of table layout.
CSS buttons.
HTML5 and fixed HTML errors.


torvista:
I see there is more to do regarding simplifying the email content / $block['GV_BLOCK'] in functions_email.php, so back to work.

torvista:
consolidation of text/html fragments for construction of GV email message block.
removal of references to TEXT_GV_REDEEM from gv_mail for chicken-egg php warning
Change to email_template_gv_mail to use same block as welcome email.
Change to email css for gv block overlapping to the right.
Change to email function for GV_BLOCK construction

torvista:
When this PR gets cleared, I can submit the mods I have done to finally replace "Dear" from all of the email templates.

admin/gv_mail.php
```diff
@@ -4,421 +4,382 @@
...
+                    error = true;
+                }
+            }
+        }
+
+        function check_amount(field_name, field_size, message) {
+            if (form.elements[field_name] && (form.elements[field_name].type !== "hidden")) {
+                let field_value = form.elements[field_name].value;
+
+                let decimal_check = /^[-+]?[0-9]+\.[0-9]+$/;
```
mc12345678: Since `$currencies` is defined and in play, is there a reason not to use something like `$currencies->currencies[$_SESSION['currency']]['decimal_point']` as the divider instead of hard-coding a period? There are currencies that use a period for the thousands separator and a comma for the "decimal": 7.000,50 to be seven thousand and fifty parts of the currency. Is there some other constraint for which only a decimal is being permitted?

torvista:
That's ironic...we use commas in Spain, ha, ha.
I started looking at using currencies but decided why bother. I often use web forms that are spanish or mock-english on a spanish site and 9/10 times it is unclear how you are supposed to enter decimals until you try, and it spits out an error or fails.
But usually it is still with a point as they are too lazy to code the form to deal with commas…in a country that uses commas!
I have that field pre-checked by the browser as an input  type **number** which only accepts decimal points. The supporting text says enter it as a decimal. I think that is clear enough.
You can jump through hoops to make it match the default currency separator and decimal limiter, but is it really worth the effort?...I tried for a while and decided not.

admin/gv_mail.php
```diff
@@ -4,421 +4,382 @@
...
+                $audience_select = get_audience_sql_query($_POST['customers_email_address'], 'email');
+                $mail = $db->Execute($audience_select['query_string']);
+                $mail_sent_to = $audience_select['query_name'];
+                // set time-limit for processing to 5 minutes... if allowed by PHP configuration
+                zen_set_time_limit(600);
+            }
+
+            foreach ($mail as $row) {
+                $id1 = zen_create_coupon_code($row['customers_email_address']);
+                echo '$row[customers_email_address]=' . $row['customers_email_address'] . '<br>';
```
drbyte: Is this debug code?

admin/gv_mail.php
```diff
@@ -4,421 +4,382 @@
...
-      // Now create the coupon main entry
-      $insert_query = $db->Execute("insert into " . TABLE_COUPONS . "
-                                    (coupon_code, coupon_type, coupon_amount, date_created)
-                                    values ('" . zen_db_input($id1) . "', 'G', '" . zen_db_input($_POST['amount']) . "', now())");
+                $message = EMAIL_SALUTATION . ' ' . $row['customers_firstname'] . ' ' . $row['customers_lastname'] . ',' . "\n\n";
+                $html_msg['EMAIL_SALUTATION'] = EMAIL_SALUTATION;
+                $html_msg['EMAIL_FIRST_NAME'] = $row['customers_firstname'];
+                $html_msg['EMAIL_LAST_NAME'] = $row['customers_lastname'];
+
+                $message .= zen_db_prepare_input($_POST['message']);//steve
```
drbyte: Is this comment suggesting it's a temporary line, or just left as an artifact from prior code?

admin/gv_mail.php
```diff
@@ -4,421 +4,382 @@
...
+                $mail_sent_to = zen_db_prepare_input($_POST['email_to']);
+                if (!empty($_POST['email_to_name'])) {
+                    $mail_sent_to_names = explode(' ', zen_db_prepare_input($_POST['email_to_name']), 2);
+                    $customers_firstname = $mail_sent_to_names[0];
+                    $customers_lastname = $mail_sent_to_names[1];
+                } else {
+                    $customers_firstname = '';
+                    $customers_lastname = TEXT_CUSTOMER;
+                }
+                $mail = (object)array('fields' => array('customers_firstname' => $customers_firstname, 'customers_lastname' => $customers_lastname, 'customers_email_address' => $mail_sent_to));
```
drbyte: Does this `(object)` typecasting still need to be here? I don't see where `$mail` is used as an object later, since it's being used in `foreach` instead of the older `while...MoveNext ... ->fields` syntax.

admin/gv_mail.php
```diff
@@ -4,421 +4,382 @@
...
+                }
+            }
+        }
+
+        function check_amount(field_name, field_size, message) {
+            if (form.elements[field_name] && (form.elements[field_name].type !== "hidden")) {
+                let field_value = form.elements[field_name].value;
+
+                let decimal_check = /^[-+]?[0-9]+\.[0-9]+$/;
+                decimal_check = /^(\d*\.)?\d+$/;
```
drbyte: Why declare it on line 185, and then change it again on 186?

admin/gv_mail.php
```diff
@@ -4,421 +4,382 @@
...
+        <?php echo '</form>';
+    }
+    ?>
+    <!-- body_text_eof //-->
+</div>
+<!-- body_eof //--> <!-- footer //-->
 <?php require(DIR_WS_INCLUDES . 'footer.php'); ?>
-<!-- footer_eof //-->
-<br />
+<!-- footer_eof //--> <br/>
```
drbyte: Not sure why this got moved up to be on the same line.
Also, no `/` is needed in `<br>` tags. So, if you're going to reformat this to remove the space, may as well remove the slash.

admin/gv_mail.php
```diff
@@ -4,421 +4,382 @@
...
-              </tr>
-              <tr>
+        ?>
+        <?php echo zen_draw_form('mail', FILENAME_GV_MAIL, 'action=send_email_to_user'); ?>
+        <table>
+            <tr>
+                <td><b><?php echo TEXT_FROM; ?></b> <?php echo htmlspecialchars(stripslashes($_POST['from']), ENT_COMPAT, CHARSET, true); ?></td>
+            </tr>
+            <tr>
+                <td><b><?php echo TEXT_TO; ?></b> <?php echo $mail_sent_to . (!empty($_POST['email_to']) && $mail_sent_to_name != '' ? ' - ' . $mail_sent_to_name : ''); ?></td>
```
drbyte: You've added this new POST field, but haven't added it to the Sanitizer.
c/f Lines 246, 336

So, either all the `echo` statements for this (including line 259 and 336) need to use `zen_output_string_protected` or `htmlspecialchars` ... or you need to add the new POST field to the `init_sanitize.php` ... in this case since it's a person's name, probably the `WORDS_AND_SYMBOLS_REGEX` group makes sense (akin to `customers_first_name`).

admin/gv_mail.php
torvista: Doh, yes.

admin/gv_mail.php
torvista: No, it's meant to tell me I've added that bit...should have removed it.

admin/gv_mail.php
torvista: agreed (now I've just looked it up..)

admin/gv_mail.php
torvista: debris from trial and errors..

admin/gv_mail.php
torvista: ok

admin/includes/init_includes/init_sanitize.php
```diff
@@ -209,7 +209,7 @@
                'group_name', 'geo_zone_name', 'geo_zone_description',
                'tax_class_description', 'tax_class_title', 'tax_description', 'entry_company', 'customers_firstname',
                'customers_lastname', 'entry_street_address', 'entry_suburb', 'entry_city', 'entry_state', 'customers_referral',
-               'symbol_left', 'symbol_right', 'products_model', 'alt_url');
+               'symbol_left', 'symbol_right', 'products_model', 'alt_url', 'email_to', 'email_to_name');
```
drbyte: Is `email_to` supposed to be an email address? If so, it needs to be moved to the group for emails.

Issue #35, opened 2013-12-17
Bring default country-address-format assignments up to date


drbyte:
Many built-in country records have a default address-format assignment of '1' which is not always correct. 

We welcome input for updates, but with your update recommendation, please provide a URL to a website which offers official confirmation of the correct format you're proposing. 

We'll happily incorporate updates which can be officially confirmed. 

Thanks!

jtwebsitedesign:
If we're good with this list, I can start updating the records in ZenCart?
http://bitboost.com/ref/international-address-formats.html#Formats

drbyte:
Sure. 
If your findings suggest the need to create "new" formats, double-check other references to be certain one of the existing formats can't be used.

jtwebsitedesign:
Ok, will do.

jtwebsitedesign:
Since some countries were missing from the other link, I'm using this reference instead:
https://www.informatica.com/products/data-quality/data-as-a-service/address-verification/address-formats.html

A few questions as I'm running through this list:

1) Is this accurate formatting for ID 1-5 in v1.6?
https://www.zen-cart.com/wiki/index.php/Table_address_format#address_format_id

2) I see an ID 6 (UK) and ID 7 (Australia) in v1.6, but not sure of the actual address format for these ID's. Can you confirm?

3) I assume ID 3 is meant to have the country on the next line, as follows. Is that correct?
$firstname $lastname
$streets
$city
$postcode - $state
$country

4) Do the parenthesis in ID 4 indicate that the postcode will be hidden if not present? For example, Angola has a Locality but no Postcode. Not sure if we need a new format, or if ID 4 will suffice.

5) What to do for countries with "no standard address format" (for example, Afghanistan)?

drbyte:
> 2) I see an ID 6 (UK) and ID 7 (Australia) in v1.6, but not sure of the actual address format for these ID's. Can you confirm?

Yes.

> 5) What to do for countries with "no standard address format" (for example, Afghanistan)?

I believe it just uses id:1 for that

jtwebsitedesign:
Sorry, maybe my questions weren't clear. I see these formats in the following Wiki. But 6 and 7 are missing. Is this Wiki correct and up-to-date for ZC v1.6? What are the formats for 6 & 7?

https://www.zen-cart.com/wiki/index.php/Table_address_format#address_format_id

Default Entries

address_format_id:	1
[...]

PR #2789, opened 2019-09-03
Multi Language Country Names

Zen4All:
This was already implemented into ZC1.6, so I thought it would be a good idea to put it in ZC1.5.7 too.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2789)


lat9:
I'll note that this change, like the Multi-lingual EZ-Pages change to zc156, has the possibility of wrecking havoc with various plugins that (currently) pull the `countries::countries_name` directly from the database &mdash; One-Page Checkout and Edit Orders as examples.

I'm not suggesting that the change not be made, but I believe that changes like this need to be posted on the Zen Cart forums so that plugin 'owners' have a change to make modifications prior to zc157's release.

lat9:
Are the `zones` going to be multi-lingual as well?

Zen4All:
Was planning on doing the zones in a separate PR

lat9:
Thanks, @Zen4All, that makes sense.

drbyte:
The SQL for installing country names needs refactoring too, since it now involves 2 tables.

Zen4All:
@drbyte The Zen Cart 1.6.0 installation contains three extra countries, What to do with those? (Saint-Barthélemy, Congo - Kinshasa, St. Martin)

drbyte:
> @drbyte The Zen Cart 1.6.0 installation contains three extra countries, What to do with those? (Saint-Barthélemy, Congo - Kinshasa, St. Martin)

They should be included (assuming they're still relevant).

drbyte:
I'm not loving the complexity this adds to everything. Hence the delay in merging.

Zen4All:
No rush.
It is basically the same thing as added to 1.6.0, and the separate mod released for previous ZC versions.

drbyte:
Given that "country" names, and their translations, are a "relatively" static set of data, would it make sense to have all that data housed in a text/json file (or PHP array), and skip all the db stuff altogether?

The master list of Zones would be covered too. (Store-specific geo-zones, which l link to shipping/taxes/payment-modules/etc would still need to be in DB, but don't require multilang support.)

Kinda goes back to #35 ... but fed from a static source (perhaps managed by a composer package).

Zen4All:
That's fine with me

torvista:
> Given that "country" names, and their translations, are a "relatively" static set of data, would it make sense to have all that data housed in a text/json file (or PHP array), and skip all the db stuff altogether?

Yes. Given my experience with Dolibarr. Or there will constant tinkering required. Having it external pushes it to the language pack.
Countries should be referenced to their ISO codes. IMHO.
Their provinces/districts whether geographical/political/historical/postal are a world of pain....not possible to always use iso codes.

zcwilt:

PR #2793, opened 2019-09-04
Allow search to jump directly to product if only 1 result found

dbltoe:
Created over 10 years ago and may have been in one of the earlier releases but not found in 1.5.5 series.
Allows system to take the customer straight to the product's page when only a single product is the result of their search.


scottcwilson:
I like how it reuses the existing SKIP_SINGLE_PRODUCT_CATEGORIES flag.  Nice.

dbltoe:
drbyte' s code from long ago.  I just thought it should be part of the core.  Like mod list.

PR #2797, opened 2019-09-06
Exclude "hidden" order-status updates from account-history

drbyte:
While a "hidden" admin-only order-status comment is excluded from display catalog-side, the "Order Status" name was being displayed, which might not be appropriate for customer to see.
This PR changes it to show the last non-hidden order-status-name in the My Account page and the My Account History page (which is accessed by clicking the Show All Orders) from the My Account page).


drbyte:
the split-pages code reformats the query in order to run a "count" on the query first, so it can help with pagination indicators and grabbing the first "page". In doing that it chops off things in the query that aren't needed for the counting step. In this case the newly added (but required) subselect's ORDER BY was getting chopped off, instead of chopping the "last" one. So using strrpos looks for the "last" one, and chops that instead.

PR #2814, opened 2019-09-12
Hard-coded Image Size in includes/modules/category_row.php

dbltoe:
Still learning only wanted category_row.php to be looked at.
Hard-coding the product image height and width was old school to combat product images of varying sizes.
Removing the hard-coding allows better stylesheet control over all display environments.


drbyte:
@dbltoe since you're doing this PR from your `master` branch, it's incorporating "everything" you're doing in there. Separate branches are advised for each PR. They can be deleted after a PR is submitted.
It's always advisable to make new branches off of the latest master, so if you always keep master as just an unaltered copy of the ZC github master, then you can always pull from that zc github master into your local master (so that it's current), before branching for a new PR.

To fix this one, you can probably pull `master` from the Zen Cart repo, down into your local `master`. Then push your local master up to your dbltoe github master. That should fix this PR.
Looks like you did "some" of that, but it's not complete. Or you could just close this, revert all your changes in your local master, pull fresh down from zc github master, then force-push your local master to your dbltoe github master. Then branch locally for a new PR as mentioned above.

dbltoe:
Maybe fifth time is a charm,  I'll try again later.

PR #2816, opened 2019-09-14
Fix category_product_listing.php

Zen4All:
- remove comma (bug)
- move  AND statements so it is even more easier to add joins

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2816)


drbyte:
Oops. Thanks for catching the comma typo. 

I put the AND back on the same line so it stays contextually with the JOIN

PR #1550, opened 2017-11-09
V156 admin facelift

Zen4All:
This update may be a good intermediate between the 1.5.x series and 1.6

These that are included right now are the first. I am working on the rest.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/1550)

PR #2818, opened 2019-09-16
add missing $check_status query on update_order action

boomer196:
I just upgraded to 1.5.6 and noticed an issue introduced with #1550.  
```
Request URI: /admin/orders.php?oID=12345&action=update_order
#1  mktime() called at [/admin/includes/functions/general.php:3303]
#2  zen_date_diff() called at [/admin/orders.php:203]
--> PHP Warning: mktime() expects parameter 6 to be int, string given in /admin/includes/functions/general.php on line 3303.
```

## Steps to recreate
* Update an order with downloadable products to "UPDATE" status
* Get the warning in the logs and date does not update correctly.

It happens because the `$check_status->fields['date_purchased']` is undefined in the call to
`zen_date_diff` on `line 203`

**NOTE**: would also happen inside `if ($order->info['payment_module_code'])` block... on `line 169`

**NOTE**: the `customers_name` and `customers_email_address` from the query is not used but was present in 1.5.5, so I left it.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -158,6 +158,10 @@ if (zen_not_null($action) && $order_exists == true) {
       $status_updated = zen_update_orders_history($oID, $comments, null, $status, $customer_notified, $email_include_message);
       $order_updated = ($status_updated > 0);
 
+      $check_status = $db->Execute("select customers_name, customers_email_address, orders_status,
+                                      date_purchased from " . TABLE_ORDERS . "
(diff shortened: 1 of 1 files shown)


drbyte:
@boomer196 You are exactly correct. Thanks!

I've adjusted it slightly to match the same query used in the `edit` block a few dozen lines earlier, for consistency.

Thanks again!

PR #2822, opened 2019-09-16
Admin users.php, minor fixes

torvista:
1) After editing currentuser and update: Php Notice Undefined index: profile
Fixed with
`<td class="profile"><?php echo $userDetails['profileName'] . zen_draw_hidden_field('profile', $userDetails['profile']) ?></td>`

2)  missing table cell when Add User was used: missing the last (actions) empty table cell to match header row.

3) Was missing closing form tag for overall form, for all actions except Delete action, which was a nested form.
Needed to hide main form and closing tag when Delete in use.

4) tfoot was empty on Edit: validation error. Removed it. Only used for Add button.

I found this code very difficult to read and modify with so many ifs on the actions!

### DIFF
--- a/admin/users.php
+++ b/admin/users.php
@@ -131,7 +131,7 @@ $userList = zen_get_users();
     <div class="container-fluid" id="pageWrapper">
 
       <h1><?php echo HEADING_TITLE ?></h1>
-      <?php if (($action != '') && $action != 'delete') { //steve hide this form when delete in use ?>
+      <?php if (($action != '') && $action != 'delete') { // Hide this form when delete in use ?>
         <?php echo zen_draw_form('users', FILENAME_USERS); ?>
         <?php if (isset($formAction)) echo zen_draw_hidden_field('action', $formAction) ?>
       <?php } ?>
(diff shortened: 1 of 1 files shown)


admin/users.php
```diff
@@ -131,7 +131,7 @@ function init() {
     <div class="container-fluid" id="pageWrapper">
 
       <h1><?php echo HEADING_TITLE ?></h1>
-      <?php if ($action == 'edit' || $action == 'add' || $action == 'password') { ?>
+      <?php if (($action != '') && $action != 'delete') { //steve hide this form when delete in use ?>
```
drbyte: It'd be preferable to not have your name in the comment if it is merged.

admin/users.php
```diff
@@ -189,18 +189,18 @@ function init() {
...
                     </td>
                   <?php } else { ?>
                     <td class="actions">&nbsp;</td>
                   <?php } ?>
                 <?php } elseif ($action != 'add') { ?>
                   <td class="actions">
                       <?php if ($action != 'delete') { ?>
-                      <a href="<?php echo zen_href_link(FILENAME_USERS, 'action=edit&user=' . $userDetails['id']) ?>" class="btn btn-primary" role="button"><?php echo IMAGE_EDIT; ?></a> <a href="<?php echo zen_href_link(FILENAME_USERS, 'action=password&user=' . $userDetails['id']) ?>" class="btn btn-primary"><?php echo IMAGE_RESET_PWD; ?></a>
+                            <a href="<?php echo zen_href_link(FILENAME_USERS, 'action=edit&user=' . $userDetails['id']) ?>" class="btn btn-primary" role="button"><?php echo IMAGE_EDIT; ?></a> <a href="<?php echo zen_href_link(FILENAME_USERS, 'action=password&user=' . $userDetails['id']) ?>" class="btn btn-primary"><?php echo IMAGE_RESET_PWD; ?></a>
```
drbyte: I'm confused by this double-length indentation. And above on `195`

torvista: dunno, wasn't me, phpstorm did it then ran off

PR #2823, opened 2019-09-17
Admin category listing: enable/disable subcategories and products

torvista:
add missing icon alt/title text used in category_listing.php

fix required for php notice when disabling categories if no subcategories, due to $_POST['set_subcategories_status'] not being set.

Reworked the code so
a) the evaluation for subcategories/products status is done once: it was mistakenly inside the category tree loop.
b) status updates are only done when a subcategories/products status **change** has been selected.

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -64,44 +64,46 @@ if (zen_not_null($action)) {
 
         // change the status of categories and products
         zen_set_time_limit(600);
-        for ($i = 0, $n = sizeof($categories); $i < $n; $i++) {
           if ($_POST['categories_status'] == '1') {//form is coming from an Enabled category which is to be changed to Disabled
-            $categories_status = '0';//Disable this category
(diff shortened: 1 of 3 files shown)


admin/category_product_listing.php
```diff
@@ -64,44 +64,46 @@
...
-            $sql = "UPDATE " . TABLE_CATEGORIES . "
-                    SET categories_status = " . (int)$categories_status . "
+        for ($i = 0, $n = sizeof($categories); $i < $n; $i++) {
+
+          //set categories_status
+            if ($categories[$i]['id'] == $categories_id) {//always update THIS category
+                $sql = "UPDATE " . TABLE_CATEGORIES . "
+                    SET categories_status = " . $category_status . "
+                    WHERE categories_id = " . $categories[$i]['id'];
+                $db->Execute($sql);
```
drbyte: Why have you removed the typecasting here?

torvista: Since it was set to '1' a few lines above I thought it didn't need it. I'll put it back.

PR #2834, opened 2019-09-20
Admin: Product move, improved text and formatting

torvista:
Clipboard01move


drbyte:
As a result of merging one of your other PRs, the language file now has conflicts, because the same lines are being changed.

PR #2840, opened 2019-09-25
Add Laravel database/orm

zcwilt:
The intention, initially, is to use this for the plugin installer only
To allow for easier/cleaner code to manage schema creation/alteration and seeding
Some more infrastructure may be added later to allow more general use.


zcwilt:
Closing for now. In v2 we used vendor/composer autoloading, and going forward this would probably be a better solution.

PR #2843, opened 2019-09-26
admin: make products select drop-down/select field ordered as prev-cat-next order

torvista:
admin: make products select drop-down/select field ordered as prev-cat-next order navigation buttons when used on the same page (attribute controller, products to categories and products price manager)
: sort order defined by admin switch PRODUCT_INFO_PREVIOUS_NEXT_SORT

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -246,8 +246,10 @@
 
 ////
 // products with name, model and price pulldown
-  function zen_draw_products_pull_down($name, $parameters = '', $exclude = '', $show_id = false, $set_selected = false, $show_model = false, $show_current_category = false) {
-    global $currencies, $db, $current_category_id;
+function zen_draw_products_pull_down($name, $parameters = '', $exclude = '', $show_id = false, $set_selected = false, $show_model = false, $show_current_category = false) {
+    global $currencies, $db, $current_category_id, $prev_next_order;
+
+    $order_by = $prev_next_order ?? ' ORDER BY products_name';
 
     if ($exclude == '') {
       $exclude = array();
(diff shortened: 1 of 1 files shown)


admin/includes/functions/general.php
```diff
@@ -246,8 +246,10 @@ function zen_get_category_tree($parent_id = '0', $spacing = '', $exclude = '', $
 
 ////
 // products with name, model and price pulldown
-  function zen_draw_products_pull_down($name, $parameters = '', $exclude = '', $show_id = false, $set_selected = false, $show_model = false, $show_current_category = false) {
-    global $currencies, $db, $current_category_id;
+function zen_draw_products_pull_down($name, $parameters = '', $exclude = '', $show_id = false, $set_selected = false, $show_model = false, $show_current_category = false) {
+    global $currencies, $db, $current_category_id, $prev_next_order;
+
+    $order_by = $prev_next_order ?? ' ORDER BY products_name';
```
drbyte: 1. While I'd love to use `??` all over the place, at this point we're still trying to keep the code compatible with PHP 5.5/5.6. So, you could rework this to use `?:` or a full on `isset/isnull ? use : else` ternary.

2. It's not save to trust `$prev_next_order` here, without sanitizing it first. Else you risk introducing SQL Injections. You could wrap the whole assignment with `$db->prepare_input()`.

torvista: ok.
regarding the sanitation by $db->prepare_input.

For this clause
`$prev_next_order = ' ORDER BY LPAD(p.products_id,11,"0")';`
it produces
`
ORDER BY LPAD(p.products_id,11,\"0\")`

which chokes the query.

Is that LPAD necessary? It can't be just p.products_id?

drbyte: I believe the padding is there to force product ID sorting to be truly numeric, like 1,2,3,4,10,11,100,1000 instead of 1,10,11,100,1000,2,3,4

torvista: Either way seems to work as a natural sort
orderby1

orderby2

and in my shop, for the admin category-product listing I have long had a sort by product_id and that also gives a natural sort.

torvista: So the LPAD can be removed...?

drbyte: I've removed the LPAD in `/admin/includes/modules/products_prev_next.php`.

You still need to sanitize the variable, since you're accepting it in from global scope without any validation.

torvista: ok

PR #2863, opened 2019-10-09
Address mysql notification received when order by a field not in select.

mc12345678:


drbyte:
You've added `sort_order` as another field, but it's never used in this page.

I'm guessing you had some external need for the field. Care to share?

mc12345678:
I guess I was a little quick on the review/change of code.  I was thinking about some problems that I have come across where `ORDER BY`, `GROUP BY`, and/or `DISTINCT` were used on a field that was not included in the `SELECT` portion of the query.  This particular query doesn't appear to present a problem in current execution (although I'm still trying to follow some of the execution path), but I was in the file and for some reason the absence of the `sort_order` field combined with sorting on it caught my eye as a potential future issue.

I'm okay with closing this PR without incorporation and apologize for taking the review/code time that could go to something better.

drbyte:
For clarification:
- `GROUP BY` requires the grouped fields to be included in the `SELECT`, since it aggregates data based on those fields.
- However, `ORDER BY` is used only inside the MySQL engine, and doesn't need to be included in the `SELECT`.

PR #2865, opened 2019-10-09
Add error logging that was sought in previous versions if incorrect i…

mc12345678:
…nformation fed.

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -12,6 +12,12 @@
 function zen_href_link($page = '', $parameters = '', $connection = 'SSL', $add_session_id = true) {
     global $zco_notifier, $session_started;
 
+    if ($page == '') {
+      trigger_error("zen_href_link($page, $parameters, $connection), unable to determine the page link.",
+            E_USER_ERROR);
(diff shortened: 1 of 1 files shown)


mc12345678:
Was being a little pro-active from what others had requested. I haven't needed such additional error checking and found it a little unusual to be included considering the effect(s) of logging the issue (repetitively).

drbyte:
> from what others had requested

Where?

zcwilt:
Before the updates to use the new url structure for admin, a missing page parameter would cause a hard error.
With the new structure a missing page parameter will just produce a url pointing to the home page and no hard error.
Given that the logging may still be useful.

admin/includes/functions/html_output.php
```diff
@@ -12,6 +12,12 @@
 function zen_href_link($page = '', $parameters = '', $connection = 'SSL', $add_session_id = true) {
     global $zco_notifier, $session_started;
 
+    if ($page == '') {
+      trigger_error("zen_href_link($page, $parameters, $connection), unable to determine the page link.",
+            E_USER_ERROR);
+      die('</td></tr></table></td></tr></table><br><br><font color="#ff0000"><b>Error!</b></font><br><br><b>Unable to determine the page link!<br><br>Function used:<br><br>zen_href_link(\'' . $page . '\', \'' . $parameters . '\', \'' . $connection . '\')</b>');
```
drbyte: Is the `</td></tr></table></td></tr></table>` markup necessary?

zcwilt:
So I've changed my mind here. I definitely don't think we should be `dying` if `$page` is not set.
While in the past this may have needed some kind of fatal error, in the future not passing `$page` could be seen as a legitimate way of addressing the home page for admin.

Issue #2860, opened 2019-10-09
v1.5.7 Admin Request Sanitizer class not available at admin/includes/extra_datafiles


mc12345678:
Not sure yet of the true cause, though this problem didn't exist until I incorporated changes to config.core.php, the new bootstrap index page and the plugin modifications, but just updated the files surrounding the admin system that control the base loading and am finding that plugin usage of the admin sanitizer is not accessible at the point/previous point where a sanitizer file is to be loaded.  As described in [previous sanitization direction](https://docs.zen-cart.com/Developer_Documentation/1.5.5/code_docs/admin_sanitization#adding-parameters/groups) a sanitizer files was to be placed at: `admin/includes/extra_datafiles`.

If I understand correctly, `AdminRequestSanitizer::getInstance()` represents a call to the class `AdminRequestSanitizer` requesting to use the method `getInstance`.  Therefore, a check for `class_exists('AdminRequestSanitizer')` should be true if the class is present.  Previously the class was present when loading such a file, but now in ZC 1.5.7 it is not and if it is loaded at this point in the processing an error occurs when it is finally loaded because it will has been attempted to be loaded twice...

So what is the new plan or method for code to try to sanitize posted admin data without directly modifying `admin\includes\init_includes\init_sanitize.php`?

zcwilt:
Hi.
You are correct. The loading order has changed somewhat due to the new plugin structure being introduced in v157.
Previously `extra_datafiles` were loaded as part of the `init_files_db_names.php` file which was loaded using the `initSystem` and hence after the `AdminRequestSanitizer` class was loaded at breakpoint 0. 

Adding `AdminRequestSanitizer` to the psr4Autoloader would solve the problem, but would mean using namespacing to access the class. A simpler solution may be to just hard code the loading of the class in `application_bootstrap.php`

Will explore these options and hope to do a PR tonight.

mc12345678:
This appears to have corrected the issue.  Question though, shouldn't/couldn't constants be incorporated into `admin/includes/application_bootstrap.php` instead of the hard coding? Sorry for asking without testing.

zcwilt:
no reason to use constants for these simple requires.

PR #2872, opened 2019-10-09
Hardcode loading of AdminRequestSanitizer class.

zcwilt:
Fixes issue #2860
and allows using extra_datafiles directory for loading extra sanitizers
as originally intended.


lat9:
Is there a way that this can be back-ported to zc156?

drbyte:
> Is there a way that this can be back-ported to zc156?

Why?

zcwilt:
think @lat9  may have realized the same thing we did for #2873

PR #2875, opened 2019-10-10
Fix js check for onsite payment when only 1 module available

scottcwilson:
If there is only one payment method (say Authorize AIM), there isn't a group of radio buttons with name "payment" that you can inspect to see if one is checked - there's only  one hidden field named "payment".  So an additional test is required in the onsite payment check.


scottcwilson:
Suggested for backport to 1.5.6d if there is such a release.

PR #2877, opened 2019-10-12
Cast $_SESSION[''languages_id] to an integer

Zen4All:
Since $_SESSION[''languages_id] can only be an integer, why not type cast it as one on creation.

---
This change is [Reviewable](https://reviewable.io/reviews/zencart/zencart/2877)

### DIFF
--- a/admin/includes/functions/languages.php
+++ b/admin/includes/functions/languages.php
@@ -1,12 +1,13 @@
 <?php
+
 //
 // +----------------------------------------------------------------------+
 // |zen-cart Open Source E-commerce                                       |
 // +----------------------------------------------------------------------+
 // | Copyright (c) 2003 The zen-cart developers                           |
-// |                                                                      |   
-// | http://www.zen-cart.com/index.php                                    |   
-// |                                                                      |   
+// |                                                                      |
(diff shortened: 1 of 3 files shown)


drbyte:
In theory it should already be of the `integer` type because that's how it's stored in the db.
However, being explicit with this, particularly in the context of setting other related variables, reduces cognitive load and even helps IDEs. So I'm good with it.

drbyte:
@Zen4All we stopped using Reviewable awhile back. Are you actively using it? Or is it just firing on your PRs because it's "leftover"?
If you're actively using it, that's fine. If not, feel free to unlink it from your repos.

Zen4All:
> In theory it should already be of the `integer` type because that's how it's stored in the db.
> However, being explicit with this, particularly in the context of setting other related variables, reduces cognitive load and even helps IDEs. So I'm good with it.

Looking in the session data, it was set as a string,

Zen4All:
> @Zen4All we stopped using Reviewable awhile back. Are you actively using it? Or is it just firing on your PRs because it's "leftover"?
> If you're actively using it, that's fine. If not, feel free to unlink it from your repos.

I have no idea, I do not use it. I'll see if I am linked somehow

PR #2883, opened 2019-10-18
V156

judygunderson:
I am getting several customers that get confused by "old password" box when they click the forgotten password link.  I am constantly having to explain that "old password" is the "new password" that was just sent to them in an email.
I am wondering if we could change the verbiage to maybe say something like "new password from email" or "password from email" or something similar?  I don't know if the space is limited there too much to do this or not.

This would affect 

admin/includes/languages/english/admin_account.php
--
Line #13 : define('TEXT_ADMIN_OLD_PASSWORD', 'Old Password:');
and

admin/includes/languages/english/login.php
--
Line #17 : define('TEXT_ADMIN_OLD_PASSWORD', 'Old Password');

When I explain it to customers its like a light goes off in their heads.... ohhhhhhh, that makes sense!  lol
Hoping making a change like this will make it easier for store owners who are already in a "stressful" situation by forgetting their passwords.


mc12345678:
Although I haven't confirmed this, isn't the provided text used on a screen that is shared between "forgotten password" and just when it's time to change the password? As such, would it be better for customer experience to split these two features out or at least use some indicator to provide different logic within the same file?

At the point of password expiration there is no email. The same could be true if the email attempted doesn't exist when attempting to reset the password, though it would be an indicator to the individual that they are to look in their email or investigate an email issue...

Otherwise, @judygunderson, please do something with the pull request to bring it to a single commit rather than the history of commits and conflicts shown above. Whether it is to rebase, create a new branch off of the current branch or something. Also, note that generic changes as I might call them seem to not really be accepted for ZC 1.5.6 and should be made against 1.5.7 with possibility of backporting to earlier versions.

PR #2895, opened 2019-10-28
equalise functions zen_get_categories catalog/admin

torvista:
replacement of admin while/movenext, add of missing sort_order to admin select, removal of unnecessary quotes around integers (as suggested by a real developer).
Shortly I shall be posting a PR with expansion to this function in conjunction with products to categories, so I thought I should get them the same prior to that.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -2700,33 +2700,40 @@ function zen_copy_products_attributes($products_id_from, $products_id_to) {
   }
 
 ////
-  function zen_get_categories($categories_array = array(), $parent_id = '0', $indent = '') {
+function zen_get_categories($categories_array = array(), $parent_id = '0', $indent = '', $status_setting = '')
+{
     global $db;
 
-    if (!is_array($categories_array)) $categories_array = array();
-
-    $categories_query = "select c.categories_id, cd.categories_name
-                         from " . TABLE_CATEGORIES . " c, " . TABLE_CATEGORIES_DESCRIPTION . " cd
-                         where parent_id = '" . (int)$parent_id . "'
-                         and c.categories_id = cd.categories_id
-                         and cd.language_id = '" . (int)$_SESSION['languages_id'] . "'
-                         order by sort_order, cd.categories_name";
-
-    $categories = $db->Execute($categories_query);
-
-    while (!$categories->EOF) {
-      $categories_array[] = array('id' => $categories->fields['categories_id'],
-                                  'text' => $indent . $categories->fields['categories_name']);
-
-      if ($categories->fields['categories_id'] != $parent_id) {
-        $categories_array = zen_get_categories($categories_array, $categories->fields['categories_id'], $indent . '&nbsp;&nbsp;');
-      }
-      $categories->MoveNext();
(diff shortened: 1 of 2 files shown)


torvista:
I'm looking at zen_get_category_tree and it appears to do exactly the same as zen_get_categories but with a few more options...so zen_get_categories could be removed completely I think??

Zen4All:
It's bedtime for me, I'll take a look tomorrow.

Zen4All:
Good-morning, I had a quick look, and I think it is not feasible. The 2 functions use different input variables, which would require some converting. This can break existing uses of a function.
It may be possible to something like this
```php
function zen_get_categories ($var1, $var2 $var3) {
// convert variables
$new_var1 = $x;
$new_var2 = $var1;
$new_var3 = $var3;
$new_var4 = $var2;

 zen_get_category_tree($new_var1, $new_var2,$new_va3,$new_var4);
}
```

torvista:
This has arisen as part of my change to products to categories, which requires the array to have the category name as a concatenated string so it all becomes ordered alphabetically.
0 category
1 category > subcat
2 category > subcat >subsubcat

instead of 
0 category
1   subcat
2     subsubcat

So I was planning to add another $format variable to zen_get_categories to get that output when necessary and so not add another similar function / not break existing uses.

I would have thought it would be preferable to have a single function with various (six?) parameters to optionally modify it..?
So, any support for a global revision of this and its useages to reduce to one function...?

torvista:
If there is no interest in consolidation, I'll withdraw this PR and submit a new function with P2Categories....

scottcwilson:
Should we sit tight on the function merges and just pull in the work from 2.0 when @drbyte is ready?

torvista:
Ok. I'll leave it.
If I have a function that is only used (at present) in that one admin file, should I put it only in that file or add it to functions general so it is more "in the public view" for future consolidation?
I guess the admin overhead for having unnecessary functions loaded is negligible, so better in general?

drbyte:
> If I have a function that is only used (at present) in that one admin file, should I put it only in that file or add it to functions general so it is more "in the public view" for future consolidation?

Depends what its purpose is. And how you're using it.  Of course if it's only for your own use then putting it into the `extra_functions` directory is ideal. If it's for a PR, then use general unless there's a clear more obviously named file for it.

admin/includes/functions/general.php
```diff
@@ -2700,33 +2700,40 @@ function zen_has_category_subcategories($category_id) {
...
+                         ORDER BY c.sort_order, cd.categories_name";
+    $results = $db->Execute($categories_query);
+
+    foreach ($results as $result) {
+        $categories_array[] = [
+            'id' => $result['categories_id'],
+            'text' => $indent . $result['categories_name'],
+        ];
+        if ($result['categories_id'] != $parent_id) {
+            $categories_array = zen_get_categories($categories_array, $result['categories_id'], $indent . '&nbsp;&nbsp;', '1');
```
drbyte: Did you really mean to pass `1` to this new parameter? Doesn't that change the original functionality, thus making it a breaking change?

torvista: corrected to use the variable.

PR #2896, opened 2019-10-29
advanced_search_result: Correcting PHP notice

lat9:
... when `$_GET['keyword']` isn't supplied.  Also simplifies the initial "any variables set" check.


torvista:
https://github.com/zencart/zencart/pull/2893
but if no keyword...should it not abort the rest of the page?

lat9:
Nope, you can search by price _only_ (no keywords required).

PR #2906, opened 2019-11-03
add message stack texts for developer email testing notifications

torvista:
I was deliberating if the developer disable email switch was simply a duplication of the admin switch, but it has its use, placed in a /local/configure.php.

Clipboard01

### DIFF
--- a/admin/includes/init_includes/init_errors.php
+++ b/admin/includes/init_includes/init_errors.php
@@ -31,9 +31,8 @@ if (!defined('IS_ADMIN_FLAG')) {
 // check if email sending has been disabled by developer switch
   if (defined('DEVELOPER_OVERRIDE_EMAIL_STATUS') && DEVELOPER_OVERRIDE_EMAIL_STATUS == 'false') {
       $messageStack->add(WARNING_EMAIL_SYSTEM_DEVELOPER_OVERRIDE, 'info');
-  }
 // check if email destinations have been diverted by developer switch
-  if (defined('DEVELOPER_OVERRIDE_EMAIL_ADDRESS') && DEVELOPER_OVERRIDE_EMAIL_ADDRESS != '') {
+  } elseif (defined('DEVELOPER_OVERRIDE_EMAIL_ADDRESS') && DEVELOPER_OVERRIDE_EMAIL_ADDRESS != '') {
       $messageStack->add(sprintf(WARNING_EMAIL_SYSTEM_DEVELOPER_EMAIL, DEVELOPER_OVERRIDE_EMAIL_ADDRESS), 'info');
   }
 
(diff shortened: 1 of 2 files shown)


torvista:
I realised the second message should be dependent on the first, then realised that:
DEVELOPER_OVERRIDE_EMAIL_STATUS
This is checked for, by function zen_mail in the catalog functions....so it needs to exist on the catalog side (config.php) I would expect.
However, for a notice in the admin messagestack, as I have it at the moment, it requires the constant on the admin side.
Same for DEVELOPER_OVERRIDE_EMAIL_ADDRESS.
1) So, how best to approach this?
2) Why bother?
Well I certainly want a way to test functionality that sends emails without creating debug errors about unable to instantiate the email (because it's on a local server and I can't be bothered to sort out the server).
a) Test functionality up to email sending, without sending any email
b) Sending all emails to a defined developer email.

lat9:
FWIW, I use that feature on various test sites so that I can keep the email configuration the same as the associated live site, but to ensure that any emails sent (like orders-status updates) are sent _only_ to myself and the site owner.  That allows the test site's 'clone' of the live database to be used, but ensures that customers won't be getting those test-site messages.

torvista:
Agreed, it would be better to trigger this external to the database.
How to get catalog file-based constants loaded in the admin?

lat9:
I use the storefront /includes/extra_configures/developer_override_email_address.php to contain those overrides and accompany it with an /admin/includes/extra_configures/developer_override_email_address_admin.php that pulls in the storefront version of the file:
```
<?php
require DIR_FS_CATALOG . DIR_WS_INCLUDES . 'extra_configures/developer_override_email_address.php';
```

torvista:
ok. 
I made an extra file in catalog/includes/local and included it 
in the catalog/includes/local/configure.php
and admin/includes/local/configure.php

drbyte:
Is there any objection to wrapping the email address in `zen_output_string_protected()`?

As for email-testing, mailtrap.io has an adequate free tier for most email testing/bypassing use. Alternatively, setting up MailHog locally may simplify all that email-testing/bypassing stuff.

torvista:
Locally, I've been using the Mercury Mail that comes with Xampp which seems to do the same as those alternatives...if I actually do want to see the emails.
Remotely, I still think the in-built options have a use, to easily redirect all emails to an admin or simply stop email sending without creating irrelevant debug logs, for when you are testing something earlier in the processing chain.

drbyte:
> Locally, I've been using the Mercury Mail that comes with Xampp which seems to do the same as those alternatives...if I actually do want to see the emails.

Ya, there's a lotta times in testing when I don't care about the emails. I just don't want them going anywhere at all, nor to an undeliverable domain due to the dev environment.

> Remotely, I still think the in-built options have a use, to easily redirect all emails to an admin or simply stop email sending without creating irrelevant debug logs, for when you are testing something earlier in the processing chain.

Agreed

PR #2912, opened 2019-11-09
Ensure GET parameters do not begin with amp;

mc12345678:
Finding that there are cases where $_GET keys begin with 'amp;'
and that as a result testing against the key without the
prefix of 'amp;' basically prevents finding it... In ZC 1.5.6
and 1.5.7 this appears to cause a product to be considered
non-existent when it is in the database, active, and in stock.

For example: index.php?main_page=product_info&amp;cPath=1_4&amp;products_id=1

causes the store to present the category breadcrumb but then
the product is identified as not present.


mc12345678:
I do/did have a thought about going through all of the $_GET parameters evaluating the first four characters and processing in a similar way as posed here; however, kept it to just the parameters specifically identified by ZC.

Ideas?

What about the admin side? There still are some issues with that as well if `&amp;` is used in the web-page address.

drbyte:
But ... why are the _GET keys containing `amp;` in the first place?

The reason the `&` is encoded is so that it is properly rendered to and by the browser, but when the browser "goes" to the page (as a result of a click, or other invocation), the browser and webserver convert it back to `&`. Or, at least, it's done that for the last decade.

So, why is that now operating differently?  Or ... if the problem has always existed, what new "change" is exposing it "now" when it hasn't been visible before?

mc12345678:
> But ... why are the _GET keys containing amp; in the first place?
The reason the & is encoded is so that it is properly rendered to and by the browser, but when the browser "goes" to the page (as a result of a click, or other invocation), the browser and webserver convert it back to &. Or, at least, it's done that for the last decade.
So, why is that now operating differently? Or ... if the problem has always existed, what new "change" is exposing it "now" when it hasn't been visible before?

This is a good question. It appears to be something changed between 1.5.5 and 1.5.6.

I know that in the admin I have noticed this as an issue where I have tried to use a link that referenced the `categories.php` file and it seemed like the file did nothing until I came to realize that it had something to do with the action having `&amp;` instead of just `&`.

As for at least the product page, when I added in a check and processing for `$_GET['amp;products_id']` in `includes/modules/pages/product_info/main_template_vars.php` then the issue with that web address went away...

mc12345678:
I just remembered that I had access to a site that was using ZC 1.5.3 and I tested the following:
`index.php?main_page=product_info&amp;products_id=1' where I was presented with a product not found notification. When I revised the uri to replace `&amp;` with just `&` the product was presented.

I also, to be sure that it wasn't session related, I then entered the Web address again as previously provided and was given a product not found response.

So, not so sure that it is/has been a change in the code that has affected response, but the way it has been...

drbyte:
Typing/pasting the link containing `amp;` will understandably give grief. Is that how you got it into the browser?
**Clicking** the link on a page should be translated automatically.

mc12345678:
> Typing/pasting the link containing amp; will understandably give grief. Is that how you got it into the browser?
Clicking the link on a page should be translated automatically.

My original discovery of the issue (a "single" admin situation) was clicking on a link that was generated to convert all `&` to `&amp;`. When clicking on the version that went to edit a category, the page was rendered, but no category information was provided. When I substituted `&` for `&amp;` just before the action parameter then the associated page information was presented.

The latest issue was described as coming from a php notice type response where the page requested contained all `&amp;` between the second and following parameters. I was able to repeat the issue and correct it by ZC adjusting for the parameter(s) having the initial `amp;` removed if present but the parameter not already existing... thought was that if `&parameterx` was provided and `&amp;parameterx` was also provided, that the one fully added would be the one to stay and not be replaced by the other.

PR #2916, opened 2019-11-12
Ensure delivery_address_format_id is an integer

scottcwilson:
In some cases this value can be '' because of no shipping address (e.g. downloads).

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -667,7 +667,7 @@ class order extends base {
                             'delivery_postcode' => $this->delivery['postcode'],
                             'delivery_state' => $this->delivery['state'],
                             'delivery_country' => $this->delivery['country']['title'],
-                            'delivery_address_format_id' => $this->delivery['format_id'],
+                            'delivery_address_format_id' => (int)$this->delivery['format_id'],
                             'billing_name' => $this->billing['firstname'] . ' ' . $this->billing['lastname'],
                             'billing_company' => $this->billing['company'],
                             'billing_street_address' => $this->billing['street_address'],


drbyte:
Are you sure this isn't already covered by the init and the assignment in the top and bottom of the highlighted section here? 
https://github.com/zencart/zencart/blob/753c6b7091ee455f8ab114032a16111c4dca5ab8/includes/classes/order.php#L435-L449

scottcwilson:
Yes I'm sure.  The user's installation had a vanilla 1.5.6c copy of includes/classes/order.php.

drbyte:
Hmmm ... that shouldn't be happening.

lat9:
Perhaps the leading conditional should be modified from
```
    if ($this->content_type == 'virtual') {
      $this->delivery = array(
        'firstname' => '',
        'lastname' => '',
        'company' => '',
        'street_address' => '',
        'suburb' => '',
        'city' => '',
        'postcode' => '',
        'state' => '',
        'zone_id' => 0,
        'country' => array(
            'id' => 0, 
            'title' => '', 
            'iso_code_2' => '', 
            'iso_code_3' => ''
        ),
        'country_id' => 0,
        'format_id' => 0
      );
    } elseif ($shipping_address->RecordCount() > 0) {
```
to
```
    if ($shipping_address->EOF) {
      $this->delivery = array(
        'firstname' => '',
        'lastname' => '',
        'company' => '',
        'street_address' => '',
        'suburb' => '',
        'city' => '',
        'postcode' => '',
        'state' => '',
        'zone_id' => 0,
        'country' => array(
            'id' => 0, 
            'title' => '', 
            'iso_code_2' => '', 
            'iso_code_3' => ''
        ),
        'country_id' => 0,
        'format_id' => 0
      );
    } else {
```
I can't envision what combination of settings caused that initial condition to pass, either.

lat9:
Hmm, I wonder if the issue is due to that change regarding the shipping-address' recording when the order is a store-pickup.

drbyte:
I'm going to merge this. But I think the real root cause should be further investigated.
Perhaps even initialize the delivery address (to empty stuff) before running all the conditionals that set real values into it, instead of the empty assignment happening "as a last resort".

PR #2921, opened 2019-11-14
Update comments

mc12345678:


drbyte:
I'd rather not change the comment contents or date of a very old file whose functionality isn't changing, particularly when it's called "security patch".

zcwilt:
Agree with @drbyte

mc12345678:
On board and understand.

PR #2942, opened 2019-11-27
Refactor attribute_controller for layout improvements

Zen4All:
I made some updates, to:
- reduce code duplication
- reduce code complexity
- lay-out improvements
  - switched from images to font-awesome icons
  - radio buttons in bootstrap style (selecting is now more easy)
  - put the prefixes for price and weight in a drop down. This will improve correct input, and make it more clear where to prefixes belong to.

Ultimately I would like to merge the "add attribute" and "edit attribute" code, because that is where there is still a lot of duplicate code,


admin/attributes_controller.php
```diff
@@ -146,113 +146,64 @@
...
-                        SET product_attribute_is_free = 1
-                        WHERE products_id = " . (int)$_GET['products_filter'] . "
-                        AND products_attributes_id = " . (int)$_GET['attributes_id']);
-        } else {
-          $db->Execute("UPDATE " . TABLE_PRODUCTS_ATTRIBUTES . "
-                        SET product_attribute_is_free = 0
-                        WHERE products_id = " . (int)$_GET['products_filter'] . "
-                        AND products_attributes_id = " . (int)$_GET['attributes_id']);
-        }
-        zen_redirect(zen_href_link(FILENAME_ATTRIBUTES_CONTROLLER, $_SESSION['page_info'] . '&products_filter=' . $_GET['products_filter'] . '&current_category_id=' . $_GET['current_category_id']));
```
drbyte: Did you really intend to remove all these `zen_redirect` calls?

admin/attributes_controller.php
```diff
@@ -524,22 +477,22 @@
 // turned off until working
           $db->Execute("UPDATE " . TABLE_PRODUCTS_ATTRIBUTES . "
                         SET attributes_image = '" . zen_db_input($attributes_image_name) . "'
-                        WHERE products_attributes_id = " . (int)$attribute_id);
+                        WHERE products_attributes_id = " . $attribute_id);
 
           $db->Execute("UPDATE " . TABLE_PRODUCTS_ATTRIBUTES . "
-                        SET products_id = " . (int)$products_id . ",
-                            options_id = " . (int)$options_id . ",
-                            options_values_id = " . (int)$values_id . ",
```
drbyte: I'm slightly reluctant to remove these typecasts from the various sql queries, because if someone installs a 3rd party modification to earlier code that doesn't properly sanitize the query fields then it could leave things vulnerable. Plus it's faster to visually scan the code and be sure that the query is safe instead of having to go check all over to be sure.

That said, I definitely prefer `(int)`/`(float)` where possible since they're specific. And I prefer `zen_db_input` over `zen_db_prepare_input` when handling strings.

admin/attributes_controller.php
Zen4All: Yes it is.
The first case was already missing a redirect. The others had a redirect with minimum get values. Since the the cases are entered by clicking a link, all the info is already passed to url, and not really needed.
You could use the redirect to empty the action value.

admin/attributes_controller.php
Zen4All: ok I understand. I'll put the typecasts.

admin/attributes_controller.php
Zen4All: After testing on a live site, I concluded that te redirect should be put back. It is necessary for the action Get value to empty

Zen4All:
Do we really need this piece of code blow, including the accompanied code in the rest of the file?? I have removed it on a test site, and can see no difference in operation.
```
    <form id="divertClickProto" action="#" method="post">
      <input type="hidden" name="divertClickProto" value="" />
      <input type="hidden" name="securityToken" value="<?php echo $_SESSION['securityToken']; ?>" />
    </form>
    <script>
      function divertClick(href)
      {
        document.getElementById('divertClickProto').action = href;
        document.getElementById('divertClickProto').submit();
        return false;
      }

    </script>
```

mc12345678:
Apologies for not yet looking at how the below topic has been implemented:
> put the prefixes for price and weight in a drop down. This will improve correct input, and make it more clear where to prefixes belong to.

Previously, the only real limitation has been if it is negative then subtract, otherwise add...

What consideration has been made for the use or addition of other symbols in this list?

Zen4All:
> Apologies for not yet looking at how the below topic has been implemented:
> 
> > put the prefixes for price and weight in a drop down. This will improve correct input, and make it more clear where to prefixes belong to.
> 
> Previously, the only real limitation has been if it is negative then subtract, otherwise add...
> 
> What consideration has been made for the use or addition of other symbols in this list?

I did not know there were other possibilities. The code only saves +, - or nothing. This is how it always has been.
If you need other symbols, this wil also mean the logic for saving to the database has to be altered
This is how the the symbols are limited currently:
```  $price_prefix = ($price_prefix == 1 ? '+' : ($price_prefix == 2 ? '-' : ''));
And these are also the only selectable values in the configuration for the product layout see: DEFAULT_PRODUCT_PRICE_PREFIX

**Comment by mc12345678** [comment]
Maybe my screen is too small, but in view of that aspect all I see/saw was that when initially selecting to display an attribute that there are defaults to be displayed, but I don't see in `admin/attributes_controller.php` anything that prevents entering some other character. It appears to take whatever is posted, sanitize it and then store in the database.

As far as other options, besides seeing users post in the forum that they wanted a specific character or text, I personally also have interest in the ability to use other mathematical related characters such as `*`,`/` or instead of `*` to use `x` or `X`. While it is not a default feature of Zen Cart, it is something that can be incorporated to support multiplying/dividing an attribute against some form of previous calculated values. This is in addition and beyond the use of price factor type values which are calculated from the base price of the product not from the cost of selections made in attribute selections before the current attribute.

So yes, if one wanted an alternate default prefix, then that section of code would need to be changed, but as far as limiting what can be entered after the attribute is shown, I don't see any existing restrictions.

**Comment by Zen4All** [comment]
Yeah I agree, visibility needs some updating on smaller screens
I am all for being able to use multiplying/dividing, but since it currently is not supported, I was trying to minimize user input error. (I used to maintain a site selling products which needed multiplication of attribute prices.)
I have the values for the drop-down hard-coded in to the code, but it would be easy to make use of a configuration value. This can then be edited from the admin, so users can add/remove symbols.

**Comment by drbyte** [comment]
> make use of a configuration value. This can then be edited from the admin, so users can add/remove symbols

Wouldn't those values also need corresponding php code to actually "use" them? If so, then I'd suggest putting those values in code instead of adding a new config setting.

**Comment by Zen4All** [comment]
@drbyte Currently only adding and subtracting is supported, so IMHO there is no need for extra database entries, but as @mc12345678 was mentioning there are users that put in other symbols (even though they are not used?). But it may be possible i the future?

**Comment by drbyte** [comment]
> @drbyte Currently only adding and subtracting is supported, so IMHO there is no need for extra database entries, but as @mc12345678 was mentioning there are users that put in other symbols (even though they are not used?). But it may be possible i the future?

Yes, I understand.

My point was: Perhaps instead of hard-coding the entries in the dropdown, and without creating a configuration setting, you can simply create an array earlier in the code, which the UI uses for drawing the dropdown entries. 
Then an addon or future code change could simply update that array, without having to make UI changes, since UI/HTML changes are often more fragile.

**Comment by drbyte** [comment]
Granted, maybe making such changes should be a separate PR.

**Comment by drbyte** [comment]
> Do we really need this piece of code below, including the accompanied code in the rest of the file?? I have removed it on a test site, and can see no difference in operation.
> 
> ```
>     <form id="divertClickProto" action="#" method="post">
>       <input type="hidden" name="divertClickProto" value="" />
>       <input type="hidden" name="securityToken" value="<?php echo $_SESSION['securityToken']; ?>" />
>     </form>
>     <script>
>       function divertClick(href)
>       {
>         document.getElementById('divertClickProto').action = href;
>         document.getElementById('divertClickProto').submit();
>         return false;
>       }
> 
>     </script>
> ```

I've no recollection of why that's there. Probably something for the old UI.

Feel free to remove it.

**Comment by Zen4All** [comment]
Ok, I am going to leave my drop-down as is for now. Only change i will make is some visual improvements for smaller screens.

Going to remove the divertClick

File: admin/attributes_controller.php [hunk 3]
```diff
@@ -753,25 +754,14 @@ function zen_js_option_values_list($selectedName, $fieldName)
...
-        cssjsmenu('navbar');
-        if (document.getElementById) {
-          var kill = document.getElementById('hoverJS');
-          kill.disabled = true;
-        }
-      }
-    </script>
   </head>
   <!-- <body onload="init()"> -->
-  <body onload="init()">
```
drbyte: Is this truly safe to remove?

Zen4All: Yes, the init function sets the viewport in the head, if it is not already there. 
Since we now use bootstrap, this viewport should always be present, 
```html
<body onload="init()">
```
calls
```javascript
    <script>
      function init() {
          cssjsmenu('navbar');
          if (document.getElementById) {
              var kill = document.getElementById('hoverJS');
              kill.disabled = true;
          }
      }
    </script>
```
cssjsmenu() is set in the file admin/includes/menu.js
menu.js contains
```javascript
/* this is only for backward compatibility - will be removed in future version */
function cssjsmenu() {
  viewport = document.querySelector("meta[name=viewport]");
  if (viewport != undefined) {
    viewport.setAttribute('content', 'width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=1');
  } else {
    var metaTag=document.createElement('meta');
  metaTag.name = "viewport"
    metaTag.content = "width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=1"
    document.getElementsByTagName('head')[0].appendChild(metaTag);
  }
}
function init() {}
```
So if we hard-code the required viewport in the <head>, the above code can safely be removed.
This what I do in all my admin module pages for some time now.

Issue #2904, opened 2019-11-01
admin/categories.php->edit_category_metatags: php notices


torvista:
ZC157, php 7.3.10, Report All Errors
Admin->Catalog->Categories/Products->Click on the metags icon:

> [Request URI: /zencart-157/admin1/index.php?cmd=categories&cPath=&cID=1&action=edit_category_meta_tags, IP address: 127.0.0.1
#1  require(..\zencart-157\admin1\categories.php) called at [..\zencart-157\admin1\index.php:11]
--> PHP Notice: Undefined offset: 1 in ..\zencart-157\admin1\categories.php on line 490.
[Request URI: /zencart-157/admin1/index.php?cmd=categories&cPath=&cID=1&action=edit_category_meta_tags, IP address: 127.0.0.1
#1  require(..\zencart-157\admin1\categories.php) called at [..\zencart-157\admin1\index.php:11]
--> PHP Notice: Undefined index: metatags_title in ..\zencart-157\admin1\categories.php on line 497.
[Request URI: /zencart-157/admin1/index.php?cmd=categories&cPath=&cID=1&action=edit_category_meta_tags, IP address: 127.0.0.1
#1  require(..\zencart-157\admin1\categories.php) called at [..\zencart-157\admin1\index.php:11]
--> PHP Notice: Undefined offset: 1 in ..\zencart-157\admin1\categories.php on line 509.
[Request URI: /zencart-157/admin1/index.php?cmd=categories&cPath=&cID=1&action=edit_category_meta_tags, IP address: 127.0.0.1
#1  require(..\zencart-157\admin1\categories.php) called at [..\zencart-157\admin1\index.php:11]
--> PHP Notice: Undefined index: metatags_keywords in ..\zencart-157\admin1\categories.php on line 516.
[Request URI: /zencart-157/admin1/index.php?cmd=categories&cPath=&cID=1&action=edit_category_meta_tags, IP address: 127.0.0.1
#1  require(..\zencart-157\admin1\categories.php) called at [..\zencart-157\admin1\index.php:11]
--> PHP Notice: Undefined offset: 1 in ..\zencart-157\admin1\categories.php on line 528.

torvista:
490
`            <?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_TITLE, 'metatags_title[' . $languages[$i]['id'] . ']', 'class="col-sm-3 control-label"'); ?>`

$languages has not been created.
On first look, this is missing code in 
```
[...]

PR #2946, opened 2019-12-05
php warnings: edit product and category metatags

torvista:
Admin->Categories->--->Edit a product metatag and proceed to save it:
--> PHP Notice: Undefined index: products_date_available in
\admin1\includes\modules\update_product_meta_tags.php on line 20.

### DIFF
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -487,14 +487,14 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
           <?php echo TEXT_EDIT_CATEGORIES_META_TAGS_INTRO; ?> - <strong><?php echo $cInfo->categories_id ?> <?php echo $cInfo->categories_name; ?></strong>
         </div>
         <div class="form-group">
-            <?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_TITLE, 'metatags_title[' . $languages[$i]['id'] . ']', 'class="col-sm-3 control-label"'); ?>
+            <?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_TITLE, '', 'class="col-sm-3 control-label"'); ?>
           <div class="col-sm-9 col-md-6">
               <?php
               for ($i = 0, $n = sizeof($languages); $i < $n; $i++) {
                 ?>
               <div class="input-group">
                 <span class="input-group-addon">
-                    <?php echo zen_image(DIR_WS_CATALOG_LANGUAGES . $languages[$i]['directory'] . '/images/' . $languages[$i]['image'], $languages[$i]['metatags_title']); ?>
+                    <?php echo zen_image(DIR_WS_CATALOG_LANGUAGES . $languages[$i]['directory'] . '/images/' . $languages[$i]['image'], $languages[$i]['name']); ?>
                 </span>
                 <?php echo zen_draw_input_field('metatags_title[' . $languages[$i]['id'] . ']', htmlspecialchars(zen_get_category_metatags_title($cInfo->categories_id, $languages[$i]['id']), ENT_COMPAT, CHARSET, TRUE), zen_set_field_length(TABLE_METATAGS_CATEGORIES_DESCRIPTION, 'metatags_title') . ' class="form-control"');
                 ?>
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -525,7 +525,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
           </div>
         </div>
         <div class="form-group">
-            <?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_DESCRIPTION, 'metatags_description[' . $languages[$i]['id'] . ']', 'class="col-sm-3 control-label"'); ?>
+            <?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_DESCRIPTION, '', 'class="col-sm-3 control-label"'); ?>
           <div class="col-sm-9 col-md-6">
               <?php
               for ($i = 0, $n = sizeof($languages); $i < $n; $i++) {
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -513,7 +513,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                 ?>
               <div class="input-group">
                 <span class="input-group-addon">
-                    <?php echo zen_image(DIR_WS_CATALOG_LANGUAGES . $languages[$i]['directory'] . '/images/' . $languages[$i]['image'], $languages[$i]['metatags_keywords']); ?>
+                    <?php echo zen_image(DIR_WS_CATALOG_LANGUAGES . $languages[$i]['directory'] . '/images/' . $languages[$i]['image'], $languages[$i]['name']); ?>
                 </span>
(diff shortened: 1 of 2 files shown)


Zen4All:
This probably has the same origin ass your other issue #2904  ?

I can see no mention of an index at or around line 20
https://github.com/zencart/zencart/blob/dce11b3435ce6f55b579578f20f40894f2c38f37/admin/includes/modules/update_product_meta_tags.php#L14-L23

admin/categories.php
```diff
@@ -487,7 +487,7 @@ function init() {
           <?php echo TEXT_EDIT_CATEGORIES_META_TAGS_INTRO; ?> - <strong><?php echo $cInfo->categories_id ?> <?php echo $cInfo->categories_name; ?></strong>
         </div>
         <div class="form-group">
-            <?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_TITLE, 'metatags_title[' . $languages[$i]['id'] . ']', 'class="col-sm-3 control-label"'); ?>
```
Zen4All: Instead of ripping out the complete tag, make it like this:
```<?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_TITLE, 'metatags_title[]', 'class="col-sm-3 control-label"'); ?>
this way it stays "linked" to the following inputs

Comment (Zen4All) [reply admin/categories.php]: By the way, good catch on the bug 👍

**Comment by torvista** [comment]
think I am done for now

File: admin/categories.php [hunk 2]
```diff
@@ -487,16 +487,16 @@ function init() {
...
-            <?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_TITLE, '', 'class="col-sm-3 control-label"'); ?>
+            <h3><?php echo TEXT_EDIT_CATEGORIES_META_TAGS_TITLE; ?></h3>
           <div class="col-sm-9 col-md-6">
               <?php
               for ($i = 0, $n = sizeof($languages); $i < $n; $i++) {
                 ?>
               <div class="input-group">
                 <span class="input-group-addon">
-                    <?php echo zen_image(DIR_WS_CATALOG_LANGUAGES . $languages[$i]['directory'] . '/images/' . $languages[$i]['image'], $languages[$i]['name']); ?>
+                    <?php echo zen_draw_label(zen_image(DIR_WS_CATALOG_LANGUAGES . $languages[$i]['directory'] . '/images/' . $languages[$i]['image'], $languages[$i]['name']), 'metatags_title[' . $languages[$i]['id'] . ']'); ?>
```
Zen4All: Why add a label?

admin/categories.php
```diff
@@ -525,16 +525,16 @@ function init() {
...
-            <?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_DESCRIPTION, '', 'class="col-sm-3 control-label"'); ?>
+            <h3><?php echo TEXT_EDIT_CATEGORIES_META_TAGS_DESCRIPTION; ?></h3>
           <div class="col-sm-9 col-md-6">
               <?php
               for ($i = 0, $n = sizeof($languages); $i < $n; $i++) {
                 ?>
               <div class="input-group">
                 <span class="input-group-addon">
-                  <?php echo zen_image(DIR_WS_CATALOG_LANGUAGES . $languages[$i]['directory'] . '/images/' . $languages[$i]['image'], $languages[$i]['name']); ?>
+                  <?php echo zen_draw_label(zen_image(DIR_WS_CATALOG_LANGUAGES . $languages[$i]['directory'] . '/images/' . $languages[$i]['image'], $languages[$i]['name']), 'metatags_description[' . $languages[$i]['id'] . ']'); ?>
```
Zen4All: Same as above why a label?

admin/categories.php
```diff
@@ -506,16 +506,16 @@ function init() {
           </div>
         </div>
         <div class="form-group">
-            <?php echo zen_draw_label(TEXT_EDIT_CATEGORIES_META_TAGS_KEYWORDS, '', 'class="col-sm-3 control-label"'); ?>
```
Zen4All: Why remove the label, it is perfectly fine to use one, and adding the h3 changes the whole layout. of the page.

torvista: The original label(s) was for multiple inputs: invalid html. There didn't appear to be a way to make it valid as labels should be per input and so I thought that probably the inputs should have a label each for accessibility?
Either the text/old label should be around the inputs as a legend/fieldset or be just text as simply as possible, as I have done with simple formatting as h3. It moves the text boxes to the left, but I didn't think that mattered really.

mc12345678:
I am unable to see all of the changes, but @Zen4All you mentioned not seeing an index at line 20. There is this code though:
`$products_date_available = zen_db_prepare_input($_POST['products_date_available']` where the index in question relates to the posted data. If that particular data is not posted, then in strict mode the notice is generated.

Zen4All:
@mc12345678 thanks

torvista:
This started with a simple change to
$products_date_available = zen_db_prepare_input($_POST['products_date_available'] 
as I thought the category errors were more complicated....not really, so they are both in this PR.

torvista:
Is there something else to resolve with this?

Zen4All:
I don't think so, just wait until @drbyte has time to review. This can sometimes take a long time, but he's a busy man.

admin/includes/modules/update_product_meta_tags.php
```diff
@@ -17,7 +17,7 @@
   if (isset($_GET['pID'])) {
     $products_id = zen_db_prepare_input($_GET['pID']);
   }
-  $products_date_available = zen_db_prepare_input($_POST['products_date_available']);
+  $products_date_available = !empty($_POST['products_date_available']) ? zen_db_prepare_input($_POST['products_date_available']) : '';
```
drbyte: I see this PR started with just fixing an undefined array index. 
But, is this `$products_date_available` variable actually even used anywhere in this workflow? 
Can't this line (and the one below it) be safely removed?

torvista: I don't see any use of it, I assume it is some debris originally copied from update_product?
I'll remove it.

Zen4All:
Can we please revert these changes for categories.ph?
It ruins the page lay out, and the labels are in the wrong place. If making the html correct on the page this is IMHO not the way to go.

torvista:
To make a smaller screenshots, I've set keywords rows=2, and description rows=5, which is probably how it should be anyway.
This is how the previous version looked:
categories old 1
The "Title", "Keywords" and "Descriptions" are tagged as labels but they are not:
errors

This is the current version:
categories current

It needs some tlc on styling, but I guess I ran out of (....something....) when I was doing it.
I understand you have done a massive amount of work to make the admin look better/responsive and so understandably get peeved when someone takes a hammer to it. So sorry for that.

I think it mainly looks daft due to the stupid multi-flag EN icon that has been added:
https://github.com/zencart/zencart/commit/db5394c06a277509eb2abb93ef2c56d1f84aa8f7#commitcomment-33136228

torvista:
I'll have a go at improving it.

Zen4All:
Thanks Steve. Sorry for the rand. Looks much better now. I'll have a better look at it tomorrow.
I also do not understand why the English icon was swaped with the multi flag icon, but that decision was made some time ago.

torvista:
I'm still fiddling with it...
https://github.com/torvista/zencart/tree/admin_categories_php

torvista:
ok, enough already.
https://github.com/zencart/zencart/pull/3001

Issue #2950, opened 2019-12-06
options values manager have a "go to" or search feature


lankeeyankee:
When you get to over 100 pages of options values, it can be quite tedious to locate the one you want to edit or just to see what is there for a certain option name. Being able to search by option name to return just those results would be amazing. Or a go-to dropdown list of all option names would work, too, giving another way to see all option names from the values page. 

Using attributes for filters and comparisons on many thousands of products means a huge amount of options values. It's gotten unwieldy and we waste a lot of time guessing which page number to click to get to where we need to be.

wiztechinc:
Agree this needs to be looked at. Major issue when trying to edit any of it. Not as bad when setting things up at first.

lankeeyankee:
A bonus would be to add a "display per page" dropdown to change how many results are displayed per page including a "display all" option. I know this is in the config>maximum values, but to have the flexibility to change it on the fly would be a real time saver. Leaving it at the default 10 is best when you are adding since the add section is at the bottom of the display and you don't have to scroll too much. If you are searching for an option name and there are 30 results, it is usually desirable to be able to see all 30 at once, to help when changing sort orders or editing the values.

Zen4All:
Funny how brilliant minds have the same idea at the same time.
I am almost ready to make a PR for this page. Also containing a drop-down to filter to one option.
I have no time to do it this weekend, but will push the PR on Monday

Zen4All:
> A bonus would be to add a "display per page" dropdown to change how many results are displayed per page including a "display all" option. I know this is in the config>maximum values, but to have the flexibility to change it on the fly would be a real time saver. Leaving it at the default 10 is best when you are adding since the add section is at the bottom of the display and you don't have to scroll too much. If you are searching for an option name and there are 30 results, it is usually desirable to be able to see all 30 at once, to help when changing sort orders or editing the values.

I like this suggestion. I'll see if I can incorporate it in the upcoming PR

torvista:
> Being able to search by option name to return just those results would be amazing.
I'm finishing off an overhaul for option_name_manager (I have a branch open). I've got the global tools showing only matching products/categories, maybe filter code could be adapted for this purpose.

Zen4All:
a screen shot of what i have coded
![Screenshot from 2019-12-07 00-49-37](https://user-images.githubusercontent.com/2766593/70364394-b175d900-188c-11ea-801c-a20d0311f873.png)
 so far

lankeeyankee:
> 
[...]

PR #2955, opened 2019-12-08
Update option values manager

Zen4All:
-update html
- typecasting
- added a filter for shops with numerous option values (see #2950)

### DIFF
--- a/admin/includes/languages/english/options_values_manager.php
+++ b/admin/includes/languages/english/options_values_manager.php
@@ -117,4 +117,4 @@ define('TABLE_TEXT_MAX_COUNT_SHORT', 'Max:');
 
   define('SUCCESS_OPTION_VALUES_DELETE', 'Successful: Deletion of: ');
   define('LABEL_FILTER', 'Select Option Vaulue to filter');
-  define('TEXT_DISPLAY_NUMBER_OF_OPTION_VALUES', 'Displaying <b>%d</b> to <b>%d</b> (of <b>%d</b> OptionValues)');
+  define('TEXT_DISPLAY_NUMBER_OF_OPTION_VALUES', 'Displaying <b>%d</b> to <b>%d</b> (of <b>%d</b> Option Values)');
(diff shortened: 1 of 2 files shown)


lankeeyankee:
small typo: 
define('LABEL_FILTER', 'Select Option Vaulue to filter'); 
define('LABEL_FILTER', 'Select Option Value to filter');

lankeeyankee:
Also, would it be difficult to add a dropdown option "All" to the display amount?

Zen4All:
> Also, would it be difficult to add a dropdown option "All" to the display amount?

I don't know, will take a look at it

lankeeyankee:
When using the new filter dropdown, after adding a new value it reloads back to page 1. Would it be difficult to keep all of the parameters so it stays on the same filter and page?

Zen4All:
I have not tested this code on 1.5.6.

Zen4All:
> Is this change only for 1.5.7? Running it on 1.5.6c I get a fatal error
> `1062:Duplicate entry '1-1' for key 'PRIMARY' :: INSERT INTO products_options_values (products_options_values_id, language_id, products_options_values_name, products_options_values_sort_order) VALUES (`
> 
> I only get this error from this code change. Has 1.5.7 removed next id?

Found the problem, I accidentally removed the query for $next_id.
Will make a commit to fix that shortly

Zen4All:
As far as I am concerned I am finished with this PR.

 I think I tested all cenarios, but @lankeeyankee could you also test this file?

lankeeyankee:
yes, it's working properly in 1.5.6c. Thanks for adding the feature to stay on the same page/filter selection!

Zen4All:
Thanks for testing, and you're welcome.

drbyte:
Merged. Not sure why Github is saying it's still open and that it has merge conflicts.

I'm gonna click Close. Before you delete your branch, you might wanna double-check things ;)

Zen4All:
thanks, will check tomorrow

Zen4All:
Looks like it merged correct. Iĺl keep the branch for some time

lankeeyankee:
I have found a bug but can't figure out how to fix it. How to recreate: add a new option name, then go to the option value manager. Use the new select box to select that new name, then try to add a value. You will notice the ID that is shown next to the boxes to fill in with data is 1 instead of the next available value ID number. And if you try to insert the new value it throws a fatal error since primary key (1) is already in use.

Zen4All:
@lankeeyankee 
I thought I already fixed that in the PR, because I found that mistake during testing. I'll check it today.

scottcwilson:
@Zen4All Do we need an issue on the bug noted by you and @lankeeyankee ?

PR #2964, opened 2019-12-15
Move uploadable filename extensions list out of db

### LINKED ISSUE #2717: Need to move UPLOAD_FILENAME_EXTENSIONS out of configuration group 2
This is not an integer value.  See forum thread
https://www.zen-cart.com/showthread.php?225774-1-5-6c-bug-for-file-upload-attributes&p=1360580#post1360580

Probably a good location would be group 13 (attributes).

scottcwilson:
Fixes #2717


scottcwilson:
My archive of plugins goes back many years, and I don't see any that reference this constant.

lat9:
OK, that was the first point; how about stores that have customizations?  This change could be 'disruptive'.

scottcwilson:
I don't think it's worth the effort to create a log (which most upgraders would ignore).  The number of people who actually change this value is miniscule and they will see the change as part of the upgrade.

lat9:
They'll see it _if-and-only-if_ the person performing the upgrade is aware of the value's change for the store.

drbyte:
The constant is only used catalog-side.
The only place in core code that uses it is inside the upload class, and this PR retains that consistency.

Like all other constants, if someone defines this new constant in their catalog configure.php file then that will take precedence over the default. That way if they have need to customize it then authorized persons can do it quite easily.

scottcwilson:
It's not information that needs to be in a database; it's not subject to frequent change.  Putting it in a code file is fine.

drbyte:
Security.

drbyte:
I already said elsewhere that the change was for security reasons.
Allowing uploads of arbitrary end-user-entered filename extensions is a recipe for security nightmares. Hence the move to hardcoded requirements.

drbyte:
I don't see the value in keeping unused/orphaned/deprecated/obsolete fragments in the database. I definitely don't want any code to accidentally use this old setting; it's better to remove it.

drbyte:
For those of you advocating for not deleting the old value, I'm interested in what legitimate changes you've made to this db entry and why. If those changes reveal a legitimate change to core code defaults, then a PR with updates to the allowed values would make great sense.

If you don't have any actual changes and you're just arguing "for the common good", that's fair. The 1.5.7 upgrade docs could use brief instructions and a specific example for identifying the old setting and copying it to their catalog-side configure.php file as a new constant.

lat9:
I, for one, have changed the default setting to `pdf,tif,tiff,txt` since any uploads I receive are expected to be scope-of-work descriptions.  I don't want pictures or zip files.

> If you don't have any actual changes and you're just arguing "for the common good", that's fair. The 1.5.7 upgrade docs could use brief instructions and a specific example for identifying the old setting and copying it to their catalog-side configure.php file as a new constant.

Why catalog-side only; isn't the upload class used in the admin as well?  

Since you're suggesting that this constant be added to a site's `configure.php` file(s), how about the `zc_install` process (a) includes the default setting in the `dist-configure.php` files for use on initial installs and (b) includes a process in its upgrade 'path' where the site's current database setting for the constant gets migrated into a like-named constant in the upgraded `configure.php` files?

drbyte:
It's only actually used catalog-side (specifically in the shopping_cart class), because all the admin code specifically sets the extensions allowed whenever the upload class is used.

lat9:
OK, so how about the `zc_install` upgrade processing creates that constant in the already-present `/includes/configure.php`?

drbyte:
Clearly you want to preserve your own customization. Understood.
But you're asking for the change to be automated, and I'm not convinced that it should be automated:
- if it's never been changed, then there's no need to add an override for it, and the defaults that are already in code will suffice.
- if it's been changed from the original, then how do we know it's "safe"? That should involve human intervention to review it. We shouldn't just blindly transfer bad content into a new location.

Also, zc_install doesn't have any custom logic processing for specific versions or specific situations. It primarily just runs SQL scripts needed to keep the db schema in sync.

In hindsight configure.php was a bad suggestion: extra_configures should be used instead.
Ideally the configure.php files are as lean as possible. The only things in there should be the core requirements, or things the end-user has added for their own unique needs (and I'd argue that extra_configures is a better place for their own customizations). 
In my prior post I mentioned maybe using the catalog configure.php because it's one of many places where an end-user can set custom defines that override system defaults, and I was specifically thinking of how one can override the defaults found in defined_paths.php by placing them in configure.php. But for location it's not as semantically logical as a user creating something like `/includes/extra_configures/custom_allowed_upload_filetypes.php` for their override.

scottcwilson:
@lat9 as a low tech solution we could just write a log if the DB value has been changed from the default.

lat9:
@scottcwilson, that's better than nothing!

PR #2966, opened 2019-12-16
Use zen_is_logged_in() instead of checking $_SESSION['customer_id']

lat9:
Use zen_is_logged_in instead of checking for presence of $_SESSION['customer_id']

### DIFF
--- a/includes/functions/functions_lookups.php
+++ b/includes/functions/functions_lookups.php
@@ -863,11 +863,11 @@ function zen_get_configuration_key_value($lookup)
       // showcase no prices
         $zc_run = false;
         break;
-      case (CUSTOMERS_APPROVAL == '1' and $_SESSION['customer_id'] == ''):
+      case (CUSTOMERS_APPROVAL == '1' && !zen_is_logged_in()):
       // customer must be logged in to browse
         $zc_run = false;
         break;
-      case (CUSTOMERS_APPROVAL == '2' and $_SESSION['customer_id'] == ''):
+      case (CUSTOMERS_APPROVAL == '2' && !zen_is_logged_in()):
       // show room only
       // customer may browse but no prices
         $zc_run = false;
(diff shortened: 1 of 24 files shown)


includes/functions/functions_lookups.php
```diff
@@ -876,11 +876,11 @@ function zen_run_normal() {
       // show room only
         $zc_run = false;
         break;
-      case (CUSTOMERS_APPROVAL_AUTHORIZATION != '0' and $_SESSION['customer_id'] == ''):
+      case (CUSTOMERS_APPROVAL_AUTHORIZATION != '0' && !zen_is_logged_in()):
       // customer must be logged in to browse
         $zc_run = false;
         break;
-      case (CUSTOMERS_APPROVAL_AUTHORIZATION != '0' and $_SESSION['customers_authorization'] > '0'):
+      case (CUSTOMERS_APPROVAL_AUTHORIZATION != '0' && isset($_SESSION['customers_authorization']) && (int)$_SESSION['customers_authorization'] > 0):
```
drbyte: Should this `isset && (int) > 0` just be `empty()` ?

lat9: That depends.  The current/previous implementation _kind of_ envisions negative (e.g. -1) values in the 'customers_authorization' session value.  If that's not what's desired, i.e. values will **always** be numeric and greater than or equal to 0, then just using `empty` would suffice.

mc12345678:
Why do only some of the changes test for guest checkout and not all of them?

What pitfalls should one using the associated notifier(s) consider since these functions are not page specific as has generally been applied elsewhere in the code?

I had previously considered incorporating these changes but seemed like there were too many pitfalls of having code attached to the notifier(s) at each instance of them being called with no easy/quick way to perhaps evaluate whether to be true or false at the point that the function(s) were called.

lat9:
Some of the changes check for guest checkout (or not) because they're features that shouldn't be provided to guests.

The notifier present in the `zen_in_guest_checkout` should be used by checkout-modifying plugins, e.g. FEC, COWOA, COWAA, OPC, to identify that the checkout is currently being processed for a guest.

I am unsure where you're "going" or what issues you're trying to solve with that last comment.

PR #2968, opened 2019-12-18
Integrate Square module to latest API spec 3.20200226.0

drbyte:
NOTE: Upgraders will be required to re-authorize to your Square account manually within 30 days after installing this module. You can force it by Removing and Re-Installing the module in your Admin.


lat9:
How 'vetted' are these changes to Square's v2 API; is it ready to be deployed on a live store?

drbyte:
> How 'vetted' are these changes to Square's v2 API; is it ready to be deployed on a live store?

I just posted a zip on the forum inviting testing on live stores.

My original hesitation holding back the commit was uncertainty about currency exchange, but when I re-reviewed the changes there were no substantive changes in that functionality. Basic testing of it seemed fine, but my test environment doesn't have all the complexities that some production stores have, so feedback is welcome.

drbyte:
> Hello,
> Since there's 521 changed files in 1 commit, maybe someone can comment on if any of the new Square library files required edits to work in ZC - or are all Square files untouched copies of the latest release?
> Perhaps Square library files, their alterations (if any), and ZC files could each go in their own separate commits.

The square.php module file is dependent on the specific version of the Square SDK files in the vendor directory. So they must go together.
While there are no "alterations" to the actual Square SDK files, there are a few things that interact specifically with the exact SDK version released with the files.

drbyte:
Updated to include products/delivery details in "note", for reference (not a full integration of the customers/orders API at this time).

PR #2971, opened 2019-12-26
status message for attribute duplication

torvista:


torvista:
Hmmm, not so fast....
This message of a successful duplication is placed after the function call to zen_copy_products_attributes.
This function produces its own error and caution messages, but not a success message.
So the situation arises of a failure to copy message (eg when a source product has no attributes), followed by a success message from this PR.
I think the existing function should continue as is but return a true/false for use by the final status message in this PR.

PR #2977, opened 2019-12-30
Refactor zen_copy_products_attributes() to return a status

drbyte:
Returns false if unacceptable conditions occur.
Returns true if copy is successful, even if a duplicate causes a skip

Addresses concern raised in https://github.com/zencart/zencart/pull/2971#issuecomment-569550040


drbyte:
@torvista does this do what you were hoping for?

torvista:
That works as advertised, thanks!

PR #2978, opened 2019-12-30
If ACCOUNT_SUBURB is not true, don't show suburb

scottcwilson:
Neither admin nor customer can fix a bad entry in this field once ACCOUNT_SUBURB is set to false, so don't show it in the order data.

Scenario: 
- ACCOUNT_SUBURB is set to true
- Customer enters a bad value in address line 2
- ACCOUNT_SUBURB is set to false
- Bad value cannot be fixed by customer, and appears when their address is shown in order data.


drbyte:
> Neither admin nor customer can fix a bad entry in this field

Should that be changed?

scottcwilson:
Yes.  Scenario: 
- ACCOUNT_SUBURB is set to true
- Customer enters a bad value in address line 2
- ACCOUNT_SUBURB is set to false
- Bad value cannot be fixed by customer, and appears when their address is shown in order data.

PR #2981, opened 2020-01-06
Additional attribute-management notifications

lat9:
Enabling a watching observer to keep track of additional option-related fields when attributes are copied or removed.

 Also adds a notification to indicate that a product is being copied, allowing an observer to copy additional product-related fields, too.


lat9:
Thanks, @drbyte!

PR #2983, opened 2020-01-07
Allow admin to enter HTML order updates

scottcwilson:


scottcwilson:
The entry with no updated by should only be the first one, which was entered by the customer (and thus should be treated with maximum care).

lat9:
Shouldn't the order-class be responsible for that initial sanitization?  It seems to me that these changes are sweeping the issue under the rug.

lat9:
Note, too, that an initial order placed by an EMP admin will not have an empty 'updated_by' field.

scottcwilson:
> Note, too, that an initial order placed by an EMP admin will not have an empty 'updated_by' field.

That's fine - we trust this entry more than one from a regular customer.

scottcwilson:
> Shouldn't the order-class be responsible for that initial sanitization? It seems to me that these changes are sweeping the issue under the rug.

No one is sweeping anything under the rug.  I'm not using `zen_db_output` so we can see HTML rendered properly.  There are a couple of extra checks to be sure nothing untoward is happening.  That's all.

lat9:
I don't understand why the changes are being made "post-order".  Shouldn't the order-class' processing have properly sanitized those entries?

scottcwilson:
Defense in depth.  ALL THE TIME I see people who have DIY upgrades where the order.php class is out of date.

lat9:
OK, point taken.  I should have started the discussion with "What's the problem you're trying to solve?", which would have brought the underlying issue out sooner.

drbyte:
> What's the problem you're trying to solve?

I'm still wondering this about this whole PR. The only reason I've seen for using HTML in updates is to post a URL to a shipment tracking site ... but every email client auto-converts those to clickable anyway, so the customer already has a clickable link when the update is emailed to them.

scottcwilson:
I have a client who likes to use **bold**.  A lot. :)   With long blocks of text, separated by line breaks.

lat9:
Rather than using the `updated_by` field as an indicator (which it's not meant to be), why doesn't the change simply enable the HTML display for all status-history records ... other than the first?

scottcwilson:
Because the first might be something entered by the EMP user.   

Honestly, I had no idea this would be so controversial.  It's really a pretty small change.

lankeeyankee:
I like the idea of adding HTML to comments

lat9:
I do too, I just don't like the use of the `updated_by` field as the means of determining whether/not HTML is allowed.  That's not what that field is intended to indicate.

drbyte:
If we do it, shouldn't we make the comments field a rich-text-editor field too?

lankeeyankee:
yes, I think that would be great. A config switch to turn the editor on/off would be even better

scottcwilson:
> If we do it, shouldn't we make the comments field a rich-text-editor field too?

People will think they *have* to enter HTML.  Most won't want to - this just allows the ones who do to be able to.  Why not stop here and see if there are further requests after people start using this?

lat9:
How's this going to play with pre-existing templates?  Since they don't know these new "rules", won't those status history records be displayed like

image

scottcwilson:
Yep.  And then they'll say, oh, gee, I guess I should update my template since I'm using this new feature.

drbyte:
> I just don't like the use of the `updated_by` field as the means of determining whether/not HTML is allowed. That's not what that field is intended to indicate.

Agreed. Let's find a way to not use this field for that purpose.

scottcwilson:
I could just remove this check.  It was just a tweak; it's not critical.

includes/templates/template_default/templates/tpl_account_history_info_default.php
```diff
@@ -111,7 +111,19 @@
     <tr>
         <td><?php echo zen_date_short($statuses['date_added']); ?></td>
         <td><?php echo $statuses['orders_status_name']; ?></td>
-        <td><?php echo (empty($statuses['comments']) ? '&nbsp;' : nl2br(zen_output_string_protected($statuses['comments']))); ?></td>
+        <td>
+<?php 
+    if (empty($statuses['comments'])) {
+echo '&nbsp;';
```
drbyte: I know we often stick a non-break-space into an empty `<td>` in order for it to display "properly" because of html4 challenges. 

Is this still needed in more modern browsers and html5? If not, let's simplify this by removing it here.

drbyte:
Can you handle the updated_by in the other file too?

And then squash this PR into a single commit?

drbyte:
Thanks

PR #2984, opened 2020-01-07
admin instantiate message stack earlier

mc12345678:
Creates a method for adding data from the session as that code was removed from the `__construct` and loads the class earlier in processing to support using `$messageStack` and its `add` method earlier in code processing to include providing messages when using early loaded functions.

Doesn't completely fix the referenced issue, but provides a method/process to be able to do so.

### DIFF
--- a/admin/includes/auto_loaders/config.core.php
+++ b/admin/includes/auto_loaders/config.core.php
@@ -111,6 +111,15 @@ if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
                                 'className'=>'sniffer',
                                 'objectName'=>'sniffer');
 /**
+ * Breakpoint 32.
+ *
+ * $messageStack = new messageStack();
+ *
+ */
+  $autoLoadConfig[32][] = array('autoType'=>'classInstantiate',
+                                 'className'=>'messageStack',
+                                 'objectName'=>'messageStack');
+/**
  * Breakpoint 35.
  *
  * require(DIR_WS_FUNCTIONS . 'admin_access.php');
@@ -186,12 +195,12 @@ if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
 /**
  * Breakpoint 100.
  *
- * $messageStack = new messageStack();
+ * $messageStack->add_from_session();
  *
  */
-  $autoLoadConfig[100][] = array('autoType'=>'classInstantiate',
-                                 'className'=>'messageStack',
-                                 'objectName'=>'messageStack');
+  $autoLoadConfig[100][] = array('autoType'=>'objectMethod',
+                                 'object'=>'messageStack',
+                                 'method'=>'add_from_session');
 /**
  * Breakpoint 120.
  *
--- a/admin/includes/classes/message_stack.php
+++ b/admin/includes/classes/message_stack.php
@@ -26,12 +26,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 
       $this->errors = array();
 
-      if (isset($_SESSION['messageToStack']) && is_array($_SESSION['messageToStack'])) {
-        for ($i = 0, $n = sizeof($_SESSION['messageToStack']); $i < $n; $i++) {
-          $this->add($_SESSION['messageToStack'][$i]['text'], $_SESSION['messageToStack'][$i]['type']);
-        }
-        $_SESSION['messageToStack'] = '';
-      }
+      $this->add_from_session();
     }
 
     function add($message, $type = 'error') {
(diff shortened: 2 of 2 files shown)


admin/includes/auto_loaders/config.core.php
```diff
@@ -110,6 +110,15 @@
   $autoLoadConfig[30][] = array('autoType'=>'classInstantiate',
                                 'className'=>'sniffer',
                                 'objectName'=>'sniffer');
+/**
+ * Breakpoint 32.
+ *
+ * $messageStack = new messageStack();
+ *
+ */
+  $autoLoadConfig[32][] = array('autoType'=>'classInstantiate',
```
drbyte: @mc12345678 any recollection about why you chose 32 here?

mc12345678: Certainly.

First there was some level of evaluation about how early the `message_stack` should be made available.  In review of the code loaded in `sniffer` and before, it did not appear that there was much cause/need for the `message_stack` to yet be available. So it seemed as if any earlier than LP 30 wasn't necessary.  Then in review of the code that followed, there seemed like there may be a desire/opportunity to use the stack for subsequent files, there was an expressed need for the files at LP 35, so it seemed that somewhere between the two seemed reasonable.

Then there was the thought of how close to each should it be placed.  If I could have chosen 32.5 I would have used it, but that was not considered available.  I didn't see a need for it to be *immediately* after 30 and didn't see why it should be immediately before 35 nor 35 itself without also knowing how others might have loaded files after/before (respectfully) each of these two LPs.  So, I went with the slightly further away from 35 as an option.  It almost seems reasonable that the class could be loaded at the end of LP 0 and before LP 10, but I didn't see a specific need nor was I sure of potential plans of how to use it now that additional options have been made available.

So, in short it fit the described need, didn't seem to get in the way of anything else, but appears that it could be used for more of the load system than currently targeted.  The separation was already somewhat into the next direction and possibly hard to accept so didn't want to be too drastic with this one. :) It certainly seems that it could be loaded sooner.

drbyte: Thanks.

admin/includes/auto_loaders/config.core.php
```diff
@@ -186,12 +195,12 @@
...
- * $messageStack = new messageStack();
+ * $messageStack->add_from_session();
  *
  */
-  $autoLoadConfig[100][] = array('autoType'=>'classInstantiate',
-                                 'className'=>'messageStack',
-                                 'objectName'=>'messageStack');
+  $autoLoadConfig[100][] = array('autoType'=>'objectMethod',
+                                 'object'=>'messageStack',
+                                 'method'=>'add_from_session');
```
drbyte: Turns out there was a bug in this previously-unused feature of the init system, where it wasn't listening properly for these properties.

Fixed in https://github.com/zencart/zencart/commit/2ebd1a1a35940cd483d68daa5a4c775187f006b9

mc12345678: If I am following correctly though, then this is a "change" (not necessarily a bad thing) from: https://github.com/zencart/zencart/blob/68b372f5cf95c5cbf666fa6d893897ac606a3c97/includes/autoload_func.php#L49-L54

Won't this mean that others that perhaps have used this method of activating an object's method will have to modify their code to support execution on 1.5.7? (I guess I could see where they just include both sets of keys `'object'` and `'objectName'` as well as `'method'` and `'methodName'` and shouldn't have a problem?!...

drbyte: Do you have an example of where this was ever used that way?

I wouldn't recommend including both, unless you wrap two separate sets of entries in an `if` conditional based on ZC version. That way you can clearly identify dead code when it's time to stop supporting the old.

mc12345678: Acckkk.. With a little further historical code review, it looks like the access method is/will be the same, it's just that the "example" has changed...

In ZC 1.5.6 and below the `includes/autoload_func.php` was as below:
https://github.com/zencart/zencart/blob/c80db1cb9b46d08de18c03606713c3050a06d165/includes/autoload_func.php#L83-L93

Messy and all, but clearly identified the need to use `objectName` and `methodName`.  In modifying the LP 100 information, I used the same "process" of determining what to include which was to review the contents of `includes/autoload_func.php` and be sure that the array being used contained the key sought.  Woops, not the case any more.

> Do you have an example of where this was ever used that way?

Not used in *that* way, but the `objectMethod` aspect is/was used:
https://github.com/mc12345678/Stock_By_Attributes_Combined/blob/22d8151bd75fb82961658a70681bfc907d628469/includes/auto_loaders/config.products_with_attributes_stock.php#L38-L42

drbyte: Apologies for the confusion. That temporary refactor in v157 (which uses `object` instead of `objectName`) is (as far as I can tell) an error. And is deprecated.
Now the InitSystem class is used instead, as it also supports the (emerging) plugin stuff:
https://github.com/zencart/zencart/blob/05116c9ffdcd9cb5fd8918b970bd12a19e795885/includes/classes/InitSystem.php#L115-L129

PR #2992, opened 2020-01-13
Allow admins to switch templates for dev/testing

drbyte:
In this context, "Admins" are defined as "IP addresses allowed to do maintenance".

Pass `&t=template_name` on the URL and that template will be used for the duration of that Session.
Open multiple private-browsing (aka incognito) browser tabs to use multiple sessions simultaneously with different templates, for comparison.

### DIFF
--- a/includes/init_includes/init_templates.php
+++ b/includes/init_includes/init_templates.php
@@ -28,6 +28,22 @@ $sql = "SELECT template_dir, template_language, template_language=" . (int)$_SES
 $result = $db->Execute($sql);
 $template_dir = $result->fields['template_dir'];
 
+/**
+ * Allow admins to switch templates using &t= URL parameter
+ */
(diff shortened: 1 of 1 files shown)


torvista:
So I have EXCLUDE_ADMIN_IP_FOR_MAINTENANCE = 127.0.0.1
and template classic selected, 
and this url:
https://www.dev.test/zencart-157/index.php?main_page=index&t=responsive_classic
but I am still seeing classic green.

drbyte:
It only lets you select templates that have been "added" to the database for the current (or all) language. So, you must use Admin->Tools->Template Select, Insert.

torvista:
Whatever the correct process is, I'm afraid I cannot figure it out, and in the process of adding/removing templates associated with languages, a slew of error messages are generated and a corrupt record:
Clipboard01
So, I would stick my neck out and say something is broken in this area and needs some tlc (my spanish language pack is up to date enough to use for testing).

torvista:
Hopefully this
https://github.com/zencart/zencart/pull/3484

PR #2994, opened 2020-01-13
escape forward slashes in javascript for html validation

torvista:


drbyte:
I don't understand: why do forward-slashes require escaping, when they're not actually ever translated as something else? Is this just another unnecessary validator tool's overzealous feedback?

torvista:
After investigating this (for once), these seem to be the most relevant infos I can find,
https://stackoverflow.com/questions/6117886/javascript-and-forward-slashes-in-strings
https://github.com/htacg/tidy-html5/issues/475

and I conclude its relevant for html4 but not html5, and it's the Tidy algorithm in the Chrome HTML validator spitting this out. I think only a complete "/script" tag in the text needs to be escaped.
But, it's hard to make solid conclusion from this. It seems it does not matter to javascript if it is escaped or not, but it may matter to the browser parser.

PR #2996, opened 2020-01-14
match names and ids for login and create account forms

torvista:
put closing form tags in php
reformat


drbyte:
Does this require also updating the javascript that handles field auto-focus?
https://github.com/zencart/zencart/blob/5a6032654118208b153f247cedf57fe075bb50b9/includes/modules/pages/login/on_load_main.js#L1

torvista:
...not anymore.

PR #2997, opened 2020-01-14
Remove unused GET parameter

Zen4All:
As far as can see this GET parameter is not used, and can be removed:
&mode=NONSSL


drbyte:
It appears to have been added in v1.2.0 -- I have no recollection why.
And no reason to believe it's needed anywhere. Esp as "non"

PR #3003, opened 2020-01-19
Specials cosmetic

torvista:
code changes/fixes to follow when this is approved.

### DIFF
--- a/admin/specials.php
+++ b/admin/specials.php
@@ -571,7 +571,7 @@ if (zen_not_null($action)) {
               if (empty($action)) {
                 ?>
                 <tr>
-                  <td colspan="2" align="right">
+                  <td colspan="2" class="text-right">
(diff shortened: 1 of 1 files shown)


drbyte:
What's your long-term plan here? Are you targeting 120 admin pages like this? Or just 2?

torvista:
Díos mío, no! I just look at whatever causes me errors on my site, initially....then do some more.
This needed a fix for a php warning only, but while I'm at it, I see if there is anything I want to add, if the html structure (td/divs etc) are unbalanced/html errors and what the IDE may spit out...a bit of future-proofing. There is not much to add to this in the next PR, but I split it anyway.

drbyte:
Let's skip the conversion to short-array syntax here.

Which basically leaves just the td align tag and the pending php warning fix you mentioned.

PR #3004, opened 2020-01-20
V157 admin functions general

mc12345678:
Simple refactor and address strict response for parent_id that could be null/not-set. Especially an issue if the current_category_id is zero, but may also occur if the parent_id has been set to null.


drbyte:
I'm going to rewrite the zen_get_path function, as there seems to be a lot of duplication going on here, and adding these extra conditions makes it even harder to follow.

PR #3007, opened 2020-01-20
Correct strict notification for non-existent variable.

mc12345678:


mc12345678:
Based on review of the conflicts, this may be a non-issue/already incorporated in some way, though wasn't when I was last aligning available code with local changes.  Likely can be closed, I would not be upset.

PR #3011, opened 2020-01-20
Prevent strict notification when no discounts

mc12345678:


mc12345678:
Another case of correction was already addressed at some point after aligning local code with GitHub at last occurrence.  Closing as no action really present to be performed.

PR #3023, opened 2020-01-21
Alert when using PHP 5.3 or older due to short-array-syntax requirement

### LINKED ISSUE #2999: array syntax: traditional or short
IDE likes to suggest short array syntax is the path to happiness.
Does the team have a preference?

1) $contents[] = array('text' => '<br>' . TEXT_INFO_DATE_ADDED . ' ' . zen_date_short($sInfo->specials_date_added));
or
2) $contents[] = ['text' => '<br>' . TEXT_INFO_DATE_ADDED . ' ' . zen_date_short($sInfo->specials_date_added)];

drbyte:
Closes #2999

### DIFF
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -66,6 +66,16 @@ define('PAGE_PARSE_START_TIME', microtime());
 @ini_set("arg_separator.output","&");
 @ini_set("html_errors","0");
 /**
+ * Ensure minimum PHP version.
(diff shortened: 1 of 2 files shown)


torvista:
I assume this still means the null coalescing operator (??) is not allowed?

drbyte:
> I assume ... the null coalescing operator (??) is not allowed?

Unfortunately that is true.

Issue #2979, opened 2020-01-01
Admin page registration GUI feedback, instantiate messagestack earlier, return values


torvista:
While fiddling with auto-adding a new menu page, I see that zen_register_admin_page adds an entry to the admin log, but there is no on-screen indication of anything having happened.
messageStack is instantiated after the admin_access functions so is not available.
a) Could this be moved earlier to provide this?
b) On a successful registration, zen_register_admin_page could return the values it was called with, to provide variables for use in the user confirmation message.

mc12345678:
Perhaps some sort of intermediate handler would be necessary.

This is what I see with trying to move the as provided `message_stack` class to an earlier load point:
The `message_stack` class (loaded at load point 100) depends on session data being available.
Session data is made available at load point 60 which also is well after the load point for the associated function.  
Further, the message_stack class does not provide a method to possibly move the session data to the `$this->errors` array at a later point and if the `__construct` method were again called where that session data is moved, then any data previously stored in `$this->errors` is removed by the first line of code in the method.

Perhaps a global variable could be set that is then "converted" or referenced at least after the session has been started to incorporate/produce the message that is to be shown.

I think that something like this was previously discussed back in Zen Cart 1.5.3 or earlier where the request was to modify or remove the `$this->errors = array();` statement from the `__construct()` method, though I'm thinking that actually perhaps a better solution is to move the code:
````php
      if (isset($_SESSION['messageToStack']) && is_array($_SESSION['messageToStack'])) {
        for ($i = 0, $n = sizeof($_SESSION['messageToStack']); $i < $n; $i++) {
          $this->add($_SESSION['messageToStack'][$i]['text'], $_SESSION['messageToStack'][$i]['type']);
        }
        $_SESSION['messageToStack'] = '';
      }
````
to a separate method within the class may be better and to have that method called from the `__construct`.  This way, if the class is instantiated prior to the session, at least the `add` method could be used and a later trigger point could be incorporated to transition `$_SESSION['messageToStack']` into `$this->errors` and therefore as applicable that method could be executed.



(Some time later)
And after all of that above, I created a PR to support making the above "possible". See https://github.com/zencart/zencart/pull/2984

torvista:
...had/have no time to look at this yet, thanks for the effort.

mc12345678:
[...]

PR #3033, opened 2020-01-23
Incorrect test causing fatal error

zcwilt:


mc12345678:
Doesn't this "invalidate" the solution of: https://github.com/zencart/zencart/pull/2984#discussion_r369643046?

Is the issue that there is not a test for existence before the is_object test?

This change will if I understand correctly not execute the object method identified here: https://github.com/zencart/zencart/pull/2984/files#diff-39234f367c971f12fac815675944ea4fR201-R203

Or are objects such as `$messageStack` becoming session variables?

lat9:
What was the motivation for the instantiation of the `$messageStack` before the session's and languages' initialization?  Shouldn't messages posted to the stack be based on the currently-active language?

includes/classes/InitSystem.php
```diff
@@ -121,7 +121,7 @@ protected function processAutoTypeObjectMethod($entry)
             $this->actionList[] = ['type' => 'sessionObjectMethod', 'object' => $objectName, 'method' => $methodName];
             $this->debugList[] = 'run session object method - ' . $objectName . ' => ' . $methodName;
         }
-        if (!is_object($_SESSION[$objectName])) {
+        if (is_object($_SESSION[$objectName])) {
```
drbyte: @zcwilt how does this differ from line 120 above?

zcwilt:
Yes, the fix is incorrect, even though it did `fix' the fatal error.
The problem is the tests themselves will never work correctly. 
This is because the parsing of the autoloaders happens in two phases. 
In phase 1 the autoloaders are reduced to a set of actions in the InitSystem class that are then actually actioned within the autoload_func.php script.
Therefore the $_SESSION array is not actually populated within the InitSystem class, so testing for it there will never work. 
The test should be done within autoload_func.php.
Preparing PR now.

mc12345678:
> What was the motivation for the instantiation of the $messageStack before the session's and languages' initialization? Shouldn't messages posted to the stack be based on the currently-active language?

To offer the ability to push information to the `$messageStack` at a point before all of the information/data necessary to display the message in the session's language has been defined.  At the (brief) moment it would likely be English only; however, I have an idea on how to support incorporating the session's language and will prepare a PR to support.

drbyte:
> What was the motivation for the instantiation of the `$messageStack` before the session's and languages' initialization? Shouldn't messages posted to the stack be based on the currently-active language?

It came from this issue https://github.com/zencart/zencart/issues/2979 by @torvista

lat9:
Just because `$messageStack` is instantiated after the load of the `admin_access` functions doesn't mean that one of those functions doesn't have the message stack available ... it's a _runtime_ thing.

Having not followed the above-posted issue, I'll say that I believe that whatever 'process' is adding a menu item during the auto-load should 'take responsibility' for the to-admin messaging, rather than overloading the admin-access functions.

PR #3037, opened 2020-01-24
Add ability to translate a string provided constant to session's language

mc12345678:
Was identified in #3033 that by moving the loading of `$messageStack` to before the `$_SESSION` has been started and before the language has been identified, that messages pushed to `$messageStack` would not be language dependent.  This adds the ability to push the expected language constant as a string that would be translated to the current session language when the session was available and the additional message(s) were then presented.


drbyte:
This doesn't feel thorough enough. It's kinda rare to only pass a constant name without additional suffix text, or a sprintf which does replacements.

I guess for very early-in-the-bootup-process messages it might be just a static constant, but I worry that people will expect it to work completely, and I don't think this would offer that.

Thoughts?

lat9:
What process is it that loads before the session and language initialization that _has to_ message the (possibly not present) admin?

If this is to support a plugin that perform their installation steps during the `extra_functions` load point, why not update the plugin to auto-load an `init_script` that runs _after_ the session/language initialization?

PR #3040, opened 2020-01-28
Return autoload object init to autoload_func

zcwilt:
Related: #3033 and #2984

### DIFF
--- a/includes/autoload_func.php
+++ b/includes/autoload_func.php
@@ -40,17 +40,16 @@ foreach ($initSystemList as $entry) {
                 $_SESSION[$objectName] = new $className();
             }
             break;
-        case 'sessionObjectMethod':
-            //echo 'sessionObjectMethod ' . $entry['class'] . "\n";
-            $objectName = $entry['object'];
-            $methodName = $entry['method'];
-            $_SESSION[$objectName]->$methodName();
-            break;
         case 'objectMethod':
             //echo 'objectMethod ' . $entry['class'] . "\n";
             $objectName = $entry['object'];
             $methodName = $entry['method'];
-            ${$objectName}->$methodName();
+            if (is_object($_SESSION[$objectName])) {
+                $_SESSION[$objectName]->$methodName();
+            }
+            if (!is_object($_SESSION[$objectName])) {
+                ${$objectName}->$methodName();
+            }
             break;
 
     }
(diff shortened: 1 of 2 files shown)


drbyte:
@torvista @lat9 @mc12345678 
does this resolve the messageStack concerns you had?

nb: Also want to test that add-to-cart and navigation-history continue to operate normally.

lat9:
I'm still trying to understand why the `$messageStack` instantiation 'wants' to be pre-session and pre-language initializations.

drbyte:
> I'm still trying to understand why the `$messageStack` instantiation 'wants' to be pre-session and pre-language initializations.

I'm not bothered by instantiating messageStack very early, much like other things are. 
I do agree that it's ideal not to stuff things onto the stack prior to language init. But that should be entirely up to whatever wants to stuff something onto the stack, shouldn't it?

torvista:
I seem to have generated a lot of work for other people...my original concern was not seeing a confirmatory message of an admin page registration post-login, if it happened pre-login. That is working now.

mc12345678:
Although would like to "properly" address at the "right" location and with the appropriate supporting information, but looks like just need to go with it.

Couple of things, there was a conversation a couple of years ago I think it was now about a redesign of the $messageStack class and how one or more of the lines should be commented out so that it didn't "self-destroy" content captured/held by the class which seems to have led to the repeated discussion that a plugin looking to do something different with the $messageStack should do what it will with it and as necessary comment the changes.

I've seen two ways that such behavior has been handled, one is to directly modify the class files (a core file) another has been to capture the data before processing in hopes that nothing else has been inserted between the two and then to restore/amplify that content a moment later just to insert a piece of information.

Then along comes a use case (with no back story of course) but a simple, how can this be done as it would be nice to have the ability to provide some level of information at other points in the load of the application. So, even though there is all the potential backstory above, the simple fact is/was that there didn't exist a way to capture information to be carried through a normal, existing process without the development of some other holder or addition of the same holder loaded at some other point or by some other process... in other words why reinvent the wheel? So I started looking at what existed, considering ways that the class could be used and ways that it could not be, and if changes were made to improve usability of the class then what would still fit in line with expressed coding layout/expectations. Then with the consideration of the message being generated so early, I figured that the message would likely be something very basicm. Somewhat of a true/false type scenario but needing to be reported. But then the discussion about the narrow scope made me think about how building something versatile may generate new ideas for usage. I've thought of an array type syntax to support some sort of variable substitution although there still runs the issue of where that fits in with data validation and permitting continued operation.

Kind of thinking that if a array is provided, but the necessary keys are not included to then implode the array for "display" in some fashion.

Anyways after an unknown number of interruptions, that's what I have/had to say about that. Basically an opportunity to solve a problem even if the problem didn't directly apply it still seems/seemed valid.

PR #3045, opened 2020-02-01
Lookup customer by id, not email

scottcwilson:
> What's the problem that we're trying to solve with this change?

It's a faster operation on a large database.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -584,7 +584,7 @@ if (zen_not_null($action) && $order_exists == true) {
               </tr>
               <tr>
                 <td class="noprint"><strong><?php echo ENTRY_CUSTOMER; ?></strong></td>
-                <td class="noprint"><?php echo '<a href="' . zen_href_link(FILENAME_CUSTOMERS, 'search=' . $order->customer['email_address'], 'SSL') . '">' . TEXT_CUSTOMER_LOOKUP . '</a>'; ?></td>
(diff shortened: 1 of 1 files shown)


drbyte:
This would break lookups for guest shoppers, right?

How about a hybrid by passing both and the lookup does a fallback?

scottcwilson:
Don't guests also have their own customer id?   Or are they customer id 0?   (Different mods may have handled this in different ways over the years I suppose).

drbyte:
No guarantees. Especially for older mods. Or for newer ones that don’t create a pseudo account

lat9:
FWIW, **_One-Page Checkout_** re-uses a fixed `customer_id` for all guest orders, so searching on the customers_id will bring up **all** guest-checkout coupons.

What's the problem that we're trying to solve with this change?

scottcwilson:
> What's the problem that we're trying to solve with this change?
It's a faster operation on a large database.

wiztechinc:
Original COWOA just had a extra field for guest versus signed up buyer so that it can be determined in admin which is which. No difference otherwise.

lat9:
> > It's a faster operation on a large database.

That makes sense; noting that for _OPC_, that click will present the common guest-customer's record.

scottcwilson:
I'm not convinced that there is ever *not* a customer table record for an order.  I think this change is OK as-is.

lat9:
Agreed, unless it's been removed.

drbyte:
> FWIW, **_One-Page Checkout_** re-uses a fixed `customer_id` for all guest orders, so searching on the customers_id will bring up **all** guest-checkout coupons.

Ya, this is going to be a problem. Until changes to core can allow that to not be an issue, we shouldn't make this proposed change.  After that I'm all for a change that allows speed improvements.

PR #3046, opened 2020-02-01
zen_get_master_categories_pulldown: allow show full path

torvista:
default: shows the single category name
Clipboard01

added optional parameter to show the full category path
Clipboard02

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -2959,29 +2959,31 @@ function zen_get_categories($categories_array = array(), $parent_id = '0', $inde
 
 /**
  * master category selection
+ * @param int $product_id
+ * @param bool $fullpath
+ * @return array
  */
-  function zen_get_master_categories_pulldown($product_id) {
+function zen_get_master_categories_pulldown(int $product_id, $fullpath = false)
+{
     global $db;
-
-    $master_category_array = array();
-
+    $master_category_array = [];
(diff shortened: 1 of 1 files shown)


lankeeyankee:
Where would you put the switch to turn fullpath on and off? As a checkbox on each page where the master category list is displayed or as a config setting? I like this idea.

torvista:
you could do both....what does the panel think?

drbyte:
Do we really need any switches? ie: How about just using the long form all the time?

Will that slow the pages down unnecessarily?
Will it clutter the display painfully?

lankeeyankee:
I personally would like a per-page checkbox to enable it. Usually I would not want to use this, I tried the changes and when there are deep subcats the dropdown is extremely wide, almost full page width on a 1920 display. So I would want it as a way of clarifying/resolving questions of where the product is located when it's not immediately clear, like in Steve's example. I see it as a helpful tool to use when needed, not on by default. I did not notice any performance issues but I'm on a ripping dedicated server.

drbyte:
Well, if we make the checkbox "in-page", then it could/would be ajax-driven, meaning the dropdown could be empty until needed, meaning the initial page-load would be faster (particularly for sites with lots of categories, meaning it has to dump a ton of HTML into the page on load). So, if one wanted to change from the default of short to long they could check the box and then the dropdown would update.

torvista:
...that's exactly what I would like to have, but I'm not up to speed on ajax...so need an example/someone else do do it.

Issue #3047, opened 2020-02-02
Category Product Listing lay out updates


Zen4All:
After the last major update of this page I think it is time to do the refinements :smile: .
The changes I propose are:
- Update Html head to remove old deprecated files
- fix lay out of the 3 columns above the categories table. This fixes spacing between items, and a more conform code use.
- change image icons for font-awesome icons
- category hover, shows folder opening. clicking the folder or category name opens the next level (as before). I removed the javascript for clicking the entire row, as this in some cases prevents clicking other objects in the row.
- changing the action buttons to proper buttons giving a cleaner look.
![Screenshot from 2020-02-02 13-43-57](https://user-images.githubusercontent.com/2766593/73608446-48c48780-45c3-11ea-815f-0d5f252c72f3.png)
![Screenshot from 2020-02-02 13-53-55](https://user-images.githubusercontent.com/2766593/73608482-c6889300-45c3-11ea-8992-b6fc79d60a91.png)

drbyte:
Closed by #3048

PR #3048, opened 2020-02-02
Layout updates for category product listing

Zen4All:
- Update Html head to remove old deprecated files
- fix lay out of the 3 columns above the categories table. This fixes spacing between items, and a more conform code use.
- change image icons for font-awesome icons
- category hover, shows folder opening. clicking the folder or category name opens the next level (as before). I removed the javascript for clicking the entire row, as this in some cases prevents clicking other objects in the row.
- changing the action buttons to proper buttons giving a cleaner look.
- Remove parentheses from requires

also see #3047

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -844,10 +844,9 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                           <i class="fa fa-dollar fa-lg" aria-hidden="true"></i>
                         </a>
                       <?php } else { ?>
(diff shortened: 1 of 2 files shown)


admin/category_product_listing.php
```diff
@@ -81,8 +80,9 @@
           }
 
           //set products_status
-          if ($products_status == '') continue;
-            
+          if ($products_status == '')
```
lat9: Suggest including braces here, for clarity.

PR #3049, opened 2020-02-03
Third argument for sprintf call should be max setting

scottcwilson:
See includes/modules/also_purchased_products.php


includes/classes/db/mysql/define_queries.php
```diff
@@ -31,5 +31,5 @@
                      AND p.products_status = 1
                      GROUP BY p.products_id, p.products_image
                      ORDER BY date_purchased desc, p.products_id
-                     LIMIT 50");
+                     LIMIT %s");
```
drbyte: For safety, this should probably be `%u` ('integer, treated as unsigned number')

PR #3051, opened 2020-02-04
Only display the "No discounts defined" msg if it is true

scottcwilson:


admin/products_price_manager.php
```diff
@@ -1088,7 +1088,9 @@ function updateSpecialsNet() {
               </div>
             </div>
           <?php } else { ?>
+            <?php if (sizeof($discount_name) == 0) { ?>
```
lat9: Shouldn't we be using `count` instead of `sizeof`?

proseLA: we should...

scottcwilson:
`count` is an alias for `sizeof`.  `sizeof` is used throughout this file; `count` is not.

lat9:
Of course it's an alias, but why propagate that older, less desirable usage in new code?

Zen4All:
I agree with @lat9, there has been a move from ```sizeof``` , towards ```count``` lateley

drbyte:
In this case `empty()` is more appropriate.

PR #3052, opened 2020-02-05
Ensure multiple-qty min/max is enforced in increments of multiples

mc12345678:
When the product is not in the cart, the default quantity displayed (buy now quantity) that has a minimum greater than the number of units is adjusted to the quantity identified as a multiple of the units that is greater than or equal to the minimum.  

This also adds two notifications when adding/editing product to identify that the minimum is greater than the number of units and is not a multiple of the units and the other is if the minimum is below the number of units.

The "ideal" situation for a standard cart seems to be for the minimum to be a multiple of the units, either equal to it or a factor of it.  This may not be the case if two different products_id (e.g. 1 and 2) are used to meet the number of units.  Such use would likely affect the admin product entry, but would still likely try to get the same result from `zen_get_buy_now_qty` when neither product is in the cart.

### DIFF
--- a/admin/includes/languages/english/product.php
+++ b/admin/includes/languages/english/product.php
@@ -59,6 +59,8 @@ define('TEXT_PRODUCTS_TAX_CLASS', 'Tax Class:');
 define('TEXT_PRODUCTS_QUANTITY_MIN_RETAIL', 'Product Qty Minimum:');
 define('TEXT_PRODUCTS_QUANTITY_UNITS_RETAIL', 'Product Qty Units:');
 define('TEXT_PRODUCTS_QUANTITY_MAX_RETAIL', 'Product Qty Maximum:');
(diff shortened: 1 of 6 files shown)


drbyte:
Unless there's a review-specific reason for not combining this all into just 1 commit instead of 11, please squash it down to 1.

Best-practice is to keep commits atomic so the full context of what was done makes sense in both the micro and macro scope.  Adding the same change to numerous files in separate commits breaks that atomic context.

mc12345678:
> Unless there's a review-specific reason for not combining this all into just 1 commit instead of 11, please squash it down to 1.

Combination of proper software access (committing directly on GitHub website) and "laziness" of not creating a separate branch, PR my local repo, assign the branch to the single commit and then try to PR request against this... Lots easier to then get to proper software and squash the commits after making the branch "visible".

torvista:
this introduced bugs in preview_info.php
https://github.com/zencart/zencart/pull/3064/commits/9af5bbe6bd6623cb6ff9161c1367ccd6b2a6daec

PR #3053, opened 2020-02-06
Adminn Copy Product - add edit duplicate checkbox

torvista:
Allows to open a duplicate product for editing, after the copy
Clipboard01

Someone asked for this in the forum, can't find the thread, seemed a good idea.

### DIFF
--- a/admin/includes/modules/copy_product_confirm.php
+++ b/admin/includes/modules/copy_product_confirm.php
@@ -111,7 +111,7 @@ if (isset($_POST['products_id'], $_POST['categories_id'])) {
 
         $db->Execute("INSERT INTO " . TABLE_PRODUCTS_TO_CATEGORIES . " (products_id, categories_id)
                       VALUES (" . $dup_products_id . ", " . $categories_id . ")");
-                      
+
         // -----
         // Notify that a copy of a "base" product has just been created, enabling an observer to duplicate
         // additional product-related fields.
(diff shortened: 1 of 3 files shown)


admin/includes/modules/copy_product_confirm.php
```diff
@@ -145,11 +145,11 @@
...
-                metatags_products_name_status = '" . zen_db_input($metatags_status->fields['metatags_products_name_status']). "',
-                metatags_model_status = '" . zen_db_input($metatags_status->fields['metatags_model_status']). "',
-                metatags_price_status= '" . zen_db_input($metatags_status->fields['metatags_price_status']). "',
-                metatags_title_tagline_status = '" . zen_db_input($metatags_status->fields['metatags_title_tagline_status']). "'
+                metatags_title_status = '" . zen_db_input($metatags_status->fields['metatags_title_status']) . "',
+                metatags_products_name_status = '" . zen_db_input($metatags_status->fields['metatags_products_name_status']) . "',
+                metatags_model_status = '" . zen_db_input($metatags_status->fields['metatags_model_status']) . "',
+                metatags_price_status= '" . zen_db_input($metatags_status->fields['metatags_price_status']) . "',
+                metatags_title_tagline_status = '" . zen_db_input($metatags_status->fields['metatags_title_tagline_status']) . "'
```
mc12345678: Aren't all of these integer values? Shouldn't `zen_db_input` basically be replaced with `(int)` and the single quotes removed from around the value (in addition to standardizing the text to have the additional space before the period and the extra parentheses removed from around the variable associated with `zen_db_input`)?

drbyte: Sure. But immaterial to this PR.

PR #3055, opened 2020-02-08
Ensure IP is not too long for field

scottcwilson:
"Field too long" error can happen in MySQL 5.7 if this value contains a string that is too long.

### DIFF
--- a/includes/functions/whos_online.php
+++ b/includes/functions/whos_online.php
@@ -31,6 +31,9 @@ function zen_update_whos_online() {
 
   $wo_session_id = zen_session_id();
   $wo_ip_address = (isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : 'Unknown');
+  if (strlen($wo_ip_address) > zen_field_length(TABLE_WHOS_ONLINE, 'ip_address') ) {
+     $wo_ip_address = "Too Long"; 
+  }
   $wo_user_agent = substr(zen_db_prepare_input(isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : ''), 0, 254);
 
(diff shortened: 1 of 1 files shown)


proseLA:
i don't like it.  it looks like this error is caused by strict mode as opposed to mySql version number.  i am guessing that it is possible a server is reporting a ipV6 address where the whos_online table has not been updated to address the increased length of that field.  or the server is configured to somehow not correctly assign the value to ```$_SERVER['REMOTE_ADDR']```.  in either case, i think this PR just obfuscates what the actual problem is.

scottcwilson:
> it looks like this error is caused by strict mode 
Yes - part of 5.7. 
> or the server is configured to somehow not correctly assign the value to $_SERVER['REMOTE_ADDR'].
... or the value is being tampered with.  In any event, it's outside our control.

lat9:
At a minimum, an `E_USER_WARNING` log should be generated to identify that the issue was detected and a work-around provided.

drbyte:
I'd like to know a little more context here. 

Was the origin problem discovered on a site using newer than ZC v1.5.1 where the field was increased to 45 chars?

In this same database, are all the other tables where an ip address is stored also using fields that are too short? Or is it isolated to just this table? 

(ie: several factors at play: a) is this PR really a complete "fix"? or is this really just a localized issue; b) if we're already going to query the db for its field length then why don't we just substr truncate the value like we do with other fields, instead of a generic string; etc)

scottcwilson:
I believe it was a 1.5.6 site, but incredibly, I seem to have no records on this fix. :( 
I don't think it makes sense to truncate because you're still getting garbage - this field is either not correctly set or tampered with. 
Please let this PR sit until it happens again and I can give more details.

scottcwilson:
> At a minimum, an `E_USER_WARNING` log should be generated to identify that the issue was detected and a work-around provided.

The point is to avoid generating logs when conditions happen that are out of our control and we can't fix.

lat9:
> The point is to avoid generating logs when conditions happen that are out of our control and we can't fix.

I _totally_ disagree.  You've modified data, placing it into the database, with no record of what actually happened, losing possibly pertinent information in determining an issue's root-cause.

drbyte:
Closing this. Feel free to re-open if more details are found.

PR #3057, opened 2020-02-08
Address book pages: Ensure form start and end tags match

scottcwilson:
In some scenarios, there could be an extra open form.


scottcwilson:
Note: This change is also required for the checkout billing address form.

drbyte:
> Note: This change is also required for the checkout billing address form.

Can you add that to the PR

scottcwilson:
Done.

PR #3062, opened 2020-02-12
Updates to Specials page changes in prior PR

torvista:
Found two bugs and made strict changes,


drbyte:
My reluctance on merging is related to concerns about all the strict-comparison changes. I "think" the edge cases are covered here, but I'd rather not be doing them "just because the IDE suggests them".

torvista:
ok, I won't bother in future.

PR #3064, opened 2020-02-13
Preview info - Product

torvista:
Fixes bug (white screen) introduced in last PR to this file

https://github.com/zencart/zencart/commit/30f9bfc9ee3940556a8eebf99328c5a9c457d862#diff-2d0663a226d3074aec93b3f41745b5cd

and other minor changes...PRs to similar files affected by this bug to follow.

### DIFF
--- a/admin/includes/languages/english/product.php
+++ b/admin/includes/languages/english/product.php
@@ -73,15 +73,15 @@ define('TEXT_PRODUCT_DATE_ADDED', 'This product was added to our catalog on %s.'
 define('TEXT_PRODUCT_DATE_AVAILABLE', 'This product will be in stock on %s.');
 
 // meta tags
-define('TEXT_META_TAG_TITLE_INCLUDES', '<strong>Mark What the Product\'s Meta Tag Title Should Include:</strong><br><span class="alert">NOTE: These switches have NO effect if no Keywords or Description text are entered below.</span>');
+define('TEXT_META_TAG_TITLE_INCLUDES', '<strong>Select items to show in the page &lt;title&gt; tag (shown in this order):</strong><br><span class="alert">NOTE: If the Keywords and Description meta tag fields are both empty, all items (apart from the Title Additional Text) will be set to "yes". However, in this case the display of the Product Model and Product Price may be overriden (disabled) in Admin page Configuration->Product Info.</span>');
 define('TEXT_PRODUCTS_METATAGS_PRODUCTS_NAME_STATUS', '<strong>Product Name:</strong>');
-define('TEXT_PRODUCTS_METATAGS_TITLE_STATUS', '<strong>Title:</strong>');
-define('TEXT_PRODUCTS_METATAGS_MODEL_STATUS', '<strong>Model:</strong>');
-define('TEXT_PRODUCTS_METATAGS_PRICE_STATUS', '<strong>Price:</strong>');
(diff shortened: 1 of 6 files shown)


torvista:
So, these are identical:
/document_product/preview_info.php
/product/preview_info.php
/product_free_shipping/preview_info.php
/product_music/preview_info.php

and this almost:
/document_general/preview_info.php

Is this redundancy-removal planned?

drbyte:
First, thanks for the bugfixes.

I want to comment on the rest, as I see a growing trend here lately which is contrary to our usual workflow ...

Bugfixes are one thing. And it's always preferable for those to be done independent of other changes in a PR. Otherwise it's really hard to identify the exact change when researching later.
While I appreciate that at least you've done those as separate commits, it would be preferred to keep bugfixes in their own PR, and leave other "code improvements" or "code reformatting" as separate PRs.

Question:
IMO a whole lot of the changes in here are completely unnecessary if you allow the loose comparisons.
I don't see why strictly-typed comparisons add any value in most of these cases. They just make the code more complicated. 
Q: Other than to satisfy an IDE is there a purpose behind the change?

torvista:
I was under the impression that with each release php gets more "strict".
So, from a shop owner point of view, I should use the latest php with strict reporting to get an early warning on things that will subsequently become a problem/error.

I assumed the inspection warnings fall under that remit, early warnings that while not a problem now, will be at some point. Otherwise, why do they have inspections at all, if not to help/give pointers to  improve the code?

Also, in a similar vein to the much-loved html validator, when there are a mass of inspections/"errors" for minor non-issues like this: 
specialsold
it's much easier to spot the new ones I've added when it's like this:
Clipboard03specialsnew
Ideally a file would have none to begin with, and then every mistake would be easier to see.

Apart from the actual syntax error in this particular file (which was suprising to see uncaught by an IDE), there was another (post variable comparison as integer), which showed up after making all the strict comparisons. 

I don't think any of the files are going away any time soon,  so I thought the strict thing would be some measure of future-proofing. I didn't think it made the code any more complicated, just more correct/less open to errors and hence more secure.
Isn't that the end goal?

barco57:
> the much-loved html validator

lol

drbyte:
Can we simplify this to just the bugfix, and the clarification made to delineate the section on multiple languages?

torvista:
I see that comparing files/commits in Github online is awful even when it's only formatting...but very easy to see the no-real-changes with Beyond Compare.
Anyway, simplified as requested.

drbyte:
Thanks. This is good.  👍 

Are you willing to apply the same to the other product types?

torvista:
...are there no plans to sort out the duplication instead?

drbyte:
> ...are there no plans to sort out the duplication instead?

No, the duplication is intentional, as each product-type is intended to be able to stand on its own.

I agree: frustrating. I hate it equally when doing upgrades and when merging the updates that plugins make. But the "right" solution is a more complex overhaul, which is separate from this PR. (Plus, I hate the "preview" step in the first place: I never "use" it, just sigh at it and click the next submit button.)

torvista:
> I hate the "preview" step in the first place: I never "use" it, just sigh at it and click the next submit button.
I put autosubmit javascript in it, so I don't even have to click it.

torvista:
that should be it

PR #3068, opened 2020-02-15
fatal error in Also Purchased: too few arguments in sprintf

torvista:
shown up by this change
https://github.com/zencart/zencart/commit/e048dd76a4cc2c54a4fe72b1f9b43df048e4574f#diff-2b7f16a8829d3825c2a7bc6a345428f3


includes/modules/also_purchased_products.php
```diff
@@ -13,13 +13,13 @@
...
 
-  $also_purchased_products = $db->ExecuteRandomMulti(sprintf(SQL_ALSO_PURCHASED, (int)$_GET['products_id'], (int)$_GET['products_id']), (int)MAX_DISPLAY_ALSO_PURCHASED);
+  $also_purchased_products = $db->ExecuteRandomMulti(sprintf(SQL_ALSO_PURCHASED, (int)$_GET['products_id'], (int)$_GET['products_id'], (int)MAX_DISPLAY_ALSO_PURCHASED));
 
   $num_products_ordered = $also_purchased_products->RecordCount();
 
   $row = 0;
   $col = 0;
-  $list_box_contents = array();
+  $list_box_contents = [];
```
scottcwilson: Why is this needed?

includes/modules/also_purchased_products.php
```diff
@@ -32,8 +32,10 @@
...
+      $list_box_contents[$row][$col] = [
+          'params' => 'class="centerBoxContentsAlsoPurch"' . ' ' . 'style="width:' . $col_width . '%;"',
+          'text' => (($also_purchased_products->fields['products_image'] === '' and PRODUCTS_IMAGE_NO_IMAGE_STATUS === 0) ? '' : '<a href="' . zen_href_link(zen_get_info_page($also_purchased_products->fields['products_id']), 'products_id=' . $also_purchased_products->fields['products_id']) . '">' . zen_image(DIR_WS_IMAGES . $also_purchased_products->fields['products_image'], $also_purchased_products->fields['products_name'], SMALL_IMAGE_WIDTH, SMALL_IMAGE_HEIGHT) . '</a><br />') . '<a href="' . zen_href_link(zen_get_info_page($also_purchased_products->fields['products_id']), 'products_id=' . $also_purchased_products->fields['products_id']) . '">' . $also_purchased_products->fields['products_name'] . '</a>'
```
drbyte: @torvista in this "wafer-thin strict" you're introducing a problem, because `PRODUCTS_IMAGE_NO_IMAGE_STATUS` is going to be a string since it comes in as a config setting stored in a string field in the db. 
While there's something eye-irritating about wrapping a digit in quotes, for strictness to work here you'll need to do that. Or `(int)` it. Or check it using `empty()`.

Also, when adding strictness for a string value like `$also_purchased_products->fields['products_image'] === ''` my preference would be to use `empty()` instead of `===''`. (The only very rare case where I'd avoid that is if it could legitimately be a `0` or `false` value, which is not the case here.)

torvista: > PRODUCTS_IMAGE_NO_IMAGE_STATUS is going to be a string since it comes in as a config setting stored in a string field in the db.
Yes, I knew that....but wasn't paying attention.

> $also_purchased_products->fields['products_image'] === '' my preference would be to use empty() instead of ===''

These occur in many places....I had assumed it was safe to treat these as strict without checking, same as a defined string or boolean. I sit corrected, thanks.

Still, while I fix more than I break, it's still an advance, no!

includes/modules/also_purchased_products.php
torvista: That's a trick question, it isn't needed!

PR #3072, opened 2020-02-16
Fix messageStack notices from autoload_func.php

scottcwilson:
[16-Feb-2020 01:57:37 America/New_York] PHP Notice:  Undefined index: messageStack in /Users/scott/Sites/gh_demo_157/includes/autoload_func.php on line 48
Same notice lin 51.

### DIFF
--- a/includes/autoload_func.php
+++ b/includes/autoload_func.php
@@ -44,11 +44,13 @@ foreach ($initSystemList as $entry) {
             //echo 'objectMethod ' . $entry['class'] . "\n";
             $objectName = $entry['object'];
             $methodName = $entry['method'];
-            if (is_object($_SESSION[$objectName])) {
-                $_SESSION[$objectName]->$methodName();
-            }
-            if (!is_object($_SESSION[$objectName])) {
-                ${$objectName}->$methodName();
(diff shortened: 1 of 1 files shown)


includes/autoload_func.php
```diff
@@ -44,13 +44,15 @@
             //echo 'objectMethod ' . $entry['class'] . "\n";
             $objectName = $entry['object'];
             $methodName = $entry['method'];
-            if (is_object($_SESSION[$objectName])) {
-                $_SESSION[$objectName]->$methodName();
-            }
-            if (!is_object($_SESSION[$objectName])) {
-                ${$objectName}->$methodName();
+            if (isset($_SESSION[$objectName])) {
+              if (is_object($_SESSION[$objectName])) {
```
lat9: Couldn't lines 47-53 be reduced to a simple if-then-else rather than an if (condition) and an if (!condition)?

scottcwilson: Maybe but I only wanted to fix the one bug so that people could start using Report All Errors (Admin) again.

torvista: Much appreciated.

PR #3077, opened 2020-02-20
changes to meta tags texts

torvista:
I've always found the description around the construction of the custom meta tag title confusing.
This is an attempt to make the options self explanatory in the product edit page and the layout options.


torvista:
I thought I would wait until this was cleared before making the changes for the other product types. Same for the update script.

drbyte:
Approved. Please proceed with adding the upgrade portion.

torvista:
I've done this:
Clipboard01

This was such a pain, I decided to build the queries in a spreadsheet, so these texts can be changed _relatively_ easily, if necessary.

PR #3079, opened 2020-02-23
allows multi-word customer search

proseLA:
the separation of first name and last name in the customers table has always made searching difficult.  if you had a customer names ```matt king```  and yet, you had many matts and matthews as well as many kings, how can one find the customer?

by separating the keywords into an array, and then searching each of the search fields for each element of that array makes finding that customer way easier IMO.


drbyte:
Thanks.

Please change your code editor to use spaces for indenting, and remove all the tabs it added in your PR.

proseLA:
sorry about that.  i think the format is proper now.

drbyte:
Thanks. I've often been annoyed by the old customer-search-result too. This gives results that are much more usable. 👍

PR #3080, opened 2020-02-27
Remove $rows variable not in use in admin files

ScotteSolutionsLtd:
It was (un)used in a few more places than just the stats_* files; all gone in this branch.


scottcwilson:
Thanks for pitching in, @ScotteSolutionsLtd !

drbyte:
Thanks!

PR #3082, opened 2020-02-27
Admin, dashboard, visitors chart, axis with multibyte day name abbreviation: fix for display on Windows

torvista:
Problem:
Clipboard01

### DIFF
--- a/admin/index_dashboard.php
+++ b/admin/index_dashboard.php
@@ -55,6 +55,9 @@ foreach ($visits as $data) {
     $counterData = "," . $counterData;
   }
   $date = strftime('%a %d', mktime(0, 0, 0, substr($data['startdate'], 4, 2), substr($data['startdate'], -2)));
+if (stripos(strtoupper(PHP_OS), 'WIN') === 0) { // Windows: for correct display of multibyte chars on visitors chart axis
+    $date = iconv('ISO-8859-1', 'UTF-8', $date);
+}
   $counterData = "['$date'," . $data['session_counter'] . "," . $data['counter'] . "]" . $counterData;
   $i++;
 }


drbyte:
There are other places where you've added the same `iconv(ISO-8859-1, UTF-8)` specifically for Windows environment.

Is this a generic problem to all Windows environments? Or is it specific to yours which is rendering all date strings as ISO-8859-1 when perhaps you can tell it to simply use UTF-8 in order to be consistent with your website's configuration?

torvista:
> Is this a generic problem to all Windows environments? Or is it specific to yours which is rendering all date strings as ISO-8859-1

Don't know...need others input on this. This is Windows 10, Xampp. I don't have access to a real Windows server.

> perhaps you can tell it to simply use UTF-8 in order to be consistent with your website's configuration?
Not sure what "it" may be in this context...I know I did a lot of digging in the past for the other time I added iconv and came to the conclusion then that Windows is just rubbish in this aspect/ it doesn't "do" utf-8.

proseLA:
@torvista i would encourage you to get virtualbox (or something similar) and then configure a virtualized linux web server to run your development server.  the number of windows web servers is rather small compared to linux web hosts.  and i would venture to guess that the number of ZC installs out in the wild, running on windows hosts is even smaller.  i think cluttering code for a minimal number of non-standard installs should be avoided.  just my opinion.

torvista:
Thanks. I would agree with that.
While well off-topic I have long thought a detailed guide as to the recommended development environment in Windows by people who do this professionally would be very welcome.
I'm still manually installing Xampp every five minutes for ten years+ and know there are better ways these days, but get dizzy with the acronyms and stacked technologies and the many ways there seems to skin this cat. So if someone can point to a page that sums up their approach, that would be helpful.

PR #3087, opened 2020-03-01
Fix display of long config settings when truncated

scottcwilson:
A cart I was working on had the contact us dropdown set to 

Sales <sales@theXXXXXstore.com>, GeneralMailbox <XXXXX@theXXXXXstore.com>

the value shown in the admin configuration page is truncated, which meant it showed up as 

Sales <sales@theABCDEstore.com&amp;g...

Performing the truncation on the **original** string, then running `htmlspecialchars` resolves this issue: 

Sales <sales@theABCDEstore.com>, Ge...


scottcwilson:
Note that this change cannot be backported to 1.5.6 by itself - it requires @mc12345678 's 1.5.7 sanitization fixes.

PR #3090, opened 2020-03-02
Allow override of sort order for admin menus

scottcwilson:
This change allows people to sort specific menus (tools and reports by default) using an alpha sort rather than the numeric sort which is used by default.  Functionality can be disabled by removing entries from the NAME_SORTED_MENUS define in the main language file. 

The motivation for this change is that these two menus are the most often added to by modders, and the menu lists can quickly become long and unwieldy.

### DIFF
--- a/admin/includes/functions/admin_access.php
+++ b/admin/includes/functions/admin_access.php
@@ -677,6 +677,16 @@ function zen_get_admin_pages($menu_only)
     }
     $result->MoveNext();
   }
+  if ($menu_only) {
+    if (defined('NAME_SORTED_MENUS') && !empty(NAME_SORTED_MENUS)) {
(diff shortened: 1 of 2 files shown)


drbyte:
Do you mind changing `NAME_SORTED_MENUS` to something slightly more declarative and less geek-speak? ;)

scottcwilson:
> Do you mind changing `NAME_SORTED_MENUS` to something slightly more declarative and less geek-speak? ;)
Happy to - what's your preferred name?

drbyte:
What do you think of something like `MENU_CATEGORIES_TO_SORT_BY_NAME` ?

scottcwilson:
Done.

torvista:
This "feature" seems to be undocumented.

PR #3094, opened 2020-03-03
Fix Responsive Classic styling of shipping options + OT lines

scottcwilson:
Without these fixes, price and title are displayed on two separate lines for shipping options and for order totals.

### DIFF
--- a/includes/templates/responsive_classic/css/responsive_mobile.css
+++ b/includes/templates/responsive_classic/css/responsive_mobile.css
@@ -282,7 +282,7 @@ span.cssButton.normal_button.button.button_continue_shopping, .button_continue_s
 .cartRemoveItemDisplay input[type=checkbox]{display:none;}
 
 /*bof checkout*/
-#checkoutPayment .forward, #checkoutShipping .forward{float:none;width:95%;}
+#checkoutConfirmDefault .forward, #checkoutPayment .forward, #checkoutShipping .forward{float:right;width:auto !important;}
 #checkoutConfirmDefault input.submit_button, #checkoutPayment input.submit_button, #checkoutShipping input.submit_button, #checkoutShipping input.submit_button:hover{display:block !important;width:100% !important;}
 #checkoutConfirmDefault input.cssButtonHover, #checkoutPayment input.cssButtonHover, #checkoutShipping input.cssButtonHover{display:block !important;width:100% !important;}
 #checkoutConfirmDefaultHeadingComments{text-align:center;}
--- a/includes/templates/responsive_classic/css/responsive_tablet.css
+++ b/includes/templates/responsive_classic/css/responsive_tablet.css
@@ -278,4 +278,6 @@ fieldset p{width:90%;margin:0 auto;}
 .gvBal{float:none;}
 #sendSpendWrapper{width:auto;}
 #checkoutSuccess fieldset .buttonRow.forward{margin:0 !important;}
+
+#checkoutConfirmDefault .forward, #checkoutPayment .forward, #checkoutShipping .forward{float:right;width:auto !important;}
 }


scottcwilson:
Screen Shot 2020-03-04 at 7 17 51 AM
Screen Shot 2020-03-04 at 7 18 53 AM
Screen Shot 2020-03-04 at 7 19 26 AM

proseLA:
is there any way to accomplish these goals without the use of ```!important```?

scottcwilson:
It's required.

proseLA:
scott, can you be more specific?  or should i just take your word for it?

pretty well established that the use of ```!important``` should be limited and only in specific cases.  i get it that it is 'easier', but i would think the base ZC should not make use of the ```!important``` declaration.  unless we have a javascript file that could be overriding the declaration on the fly.  which i'm not seeing.

css can be a pain, but at this point in its development, i think it would behoove us to clean up the base ZC as opposed to adding more code that makes it more difficult for users to override and for coders to maintain.

scottcwilson:
If you take off the !important, it doesn't work. I don't know what to tell you. 
This markup applies to only three specific blocks (it's based on an ID), so it doesn't seem that egregious.  !important is used 141 times in the css in responsive classic. 
If your css-fu is strong, please feel free to submit an alternate solution.  I just want the basic template to actually work.

proseLA:
i would suggest if your css-fu is not strong, to ask for help prior to submitting a PR.  plenty of people out there that have strong css skills (i am not one of them).

given that my css-fu is not strong, i would keep your changes, but eliminate the ```!important``` in both places on both scripts.  i would them eliminate the ```!important``` in the width declaration in following places:

https://github.com/zencart/zencart/blob/4fabb54d77be415e8c73618e9c9fe19b3edf60e7/includes/templates/responsive_classic/css/responsive_mobile.css#L437

https://github.com/zencart/zencart/blob/4fabb54d77be415e8c73618e9c9fe19b3edf60e7/includes/templates/responsive_classic/css/responsive_tablet.css#L275

does that accomplish what you want?

scottcwilson:
I don't think making a global change to the CSS which will affect every page and necessitate extensive retesting this close to a release is a great idea.  I'd rather make a small targeted change that fixes a real problem.   But let's sit tight and wait for the core committers to weigh in.

proseLA:
@scottcwilson  given some recent conversations, and the lack of weighing in by the core ZC people, i am STILL a fan of removing the !important as i discussed in the 4 places.  these affects a very small section of the template.  i have tested it.  it works.  and i would argue that just because !important is used in x number of other places, does not make it right.  my opinion is that we should look to remove !important as opposed to adding more.  you just want it to work.  it does.

if it does not, given the ZC history of releases, i see a v157a, v157b, v157c....  in our future... :grin: :rofl:

wiztechinc:
FYI I don't use it ever. I did for some crossbrowser support for IE years ago but I do feel it's inappropriate. That said it's really not majorly important.

drbyte:
> i am STILL a fan of removing the !important

> FYI I don't use it ever. I did for some crossbrowser support for IE years ago but I do feel it's inappropriate. That said it's really not majorly **important**.

I love the pun!

wiztechinc:
LOL, that was definitely not deliberate!

proseLA:
> FYI I don't use it ever. I did for some crossbrowser support for IE years ago but I do feel it's inappropriate. That said it's really not majorly important.

except for those of us trying to figure out a bug in the CSS and after an hour and and half realize its because of an `!important`  oy!

wiztechinc:
> > FYI I don't use it ever. I did for some crossbrowser support for IE years ago but I do feel it's inappropriate. That said it's really not majorly important.
> 
> except for those of us trying to figure out a bug in the CSS and after an hour and and half realize its because of an `!important` oy!

Ow! I see what you mean.

scottcwilson:
There is no cascading risk with this fix.  So no danger that you'll waste 1.5 hours. 
It's a change to a single id which exists only on a single page. 

However, it is risky to make CSS changes that affect multiple pages.  Because that's how you get bugs like the one this PR fixes. 

At any rate, you have repeatedly made your point.  @drbyte will either merge this or not.  I really just want the bug to be fixed.

drbyte:
I agree we should avoid using `!important` wherever possible.
In fact, this is only needed because of a later use of `!important` on the input label.

That said, since the scope of this is limited and fixes this specific situation suitably, and is unlikely to break other things in peoples' templates, I'll merge it. 
Fixing the broader issues is a bigger task.

PR #3095, opened 2020-03-04
Sample image needs to be present

scottcwilson:
Without this image, you'll get 404 errors when using the sample data configuration. 

Screen Shot 2020-03-04 at 7 23 21 AM


drbyte:
I'm pretty sure you've got old data in your banners table if you're getting requests for that image.
The `pan` server serves all its images from AWS or other external places, not from localhost.

PR #3104, opened 2020-03-06
Changes to error messages for adding invalid coupon in checkout payment

torvista:
I actually found one undefined constant (TEXT_INVALID_COUPON_ORDER_LIMIT) on the way!

Examples (note this is a frigged output to see all the error messages at once):
coupon errors

order of constants in language file now matches useage order.

### DIFF
--- a/includes/modules/order_total/ot_coupon.php
+++ b/includes/modules/order_total/ot_coupon.php
@@ -220,7 +220,7 @@ class ot_coupon {
       if (!$coupon_result->EOF) {
 
         if ($coupon_result->RecordCount() < 1 ) {
-          $messageStack->add_session('redemptions', TEXT_INVALID_REDEEM_COUPON . ' ' . $dc_check,'caution');
+          $messageStack->add_session('redemptions', sprintf(TEXT_INVALID_REDEEM_COUPON, $dc_check),'caution');
           $this->clear_posts();
           zen_redirect(zen_href_link(FILENAME_CHECKOUT_PAYMENT, '', 'SSL',true, false));
         }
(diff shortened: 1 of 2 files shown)


includes/modules/order_total/ot_coupon.php
```diff
@@ -313,23 +313,21 @@ function collect_posts() {
         // JTD - end of handling coupon product restrictions
 
         $date_query=$db->Execute("select coupon_start_date from " . TABLE_COUPONS . "
-                                  where coupon_start_date <= now() and
-                                  coupon_code='" . zen_db_prepare_input($dc_check) . "'");
+                                  where coupon_code='" . zen_db_prepare_input($dc_check) . "' LIMIT 1");
 
-        if ($date_query->RecordCount() < 1 ) {
-          $messageStack->add_session('redemptions', TEXT_INVALID_STARTDATE_COUPON . ' ' . ($dc_link_count ==0 ? $dc_link : ''),'caution');
+        if (date("Y-m-d H:i:s") < $date_query->fields['coupon_start_date']) {
```
drbyte: I'm not sure I understand why you're doing another date-comparison here when it was just done 2 lines earlier in the database query so the query results are only those within a suitable date range.

torvista: not sure what you mean. I just substituted the sql checks that were there already for the start and expire dates. You mean you want to merge to a single query with a single error message for an invalid date?

drbyte: No, I mean, I don't see why you made a change at all. There's no need to add a PHP date comparison when it's already done in the sql.

torvista: ah, it's to provide date information in the message. Previously it would just return no result, now it returns the date for use in the message.

includes/modules/order_total/ot_coupon.php
```diff
@@ -313,23 +313,21 @@ function collect_posts() {
...
           $error_issues ++;
           $dc_link_count ++;
 //          $this->clear_posts();
 //          zen_redirect(zen_href_link(FILENAME_CHECKOUT_PAYMENT, '', 'SSL', true, false));
         }
 
         $date_query=$db->Execute("select coupon_expire_date from " . TABLE_COUPONS . "
-                                  where coupon_expire_date >= now() and
-                                  coupon_code='" . zen_db_prepare_input($dc_check) . "'");
+                                  where coupon_code='" . zen_db_prepare_input($dc_check) . "'");
```
drbyte: In your other query change for dates you added LIMIT 1, but you didn't do that here. Intentional?

torvista: no, fixed.

PR #3105, opened 2020-03-06
add language_code to orders table

torvista:


scottcwilson:
What's the PR for setting the language code in the orders table at checkout time?  Otherwise this isn't done.

torvista:
Steady on, only just got the wheels on the cart, have to shoe the horse now.

I intend to PR this with my mod based on your Quick Status Update to send the emails in the order language without switching Admin language.

scottcwilson:
OK -to set the language in the orders table at order time, just modify includes/classes/order.php after line 655 where the update happens.

torvista:
Yes, I need to update my production with this field_name first, then port my supporting code to a PR....

Issue #2783, opened 2019-08-30
Bug: admin, use of parameter 'back_x' with Back button incorrect


torvista:
This parameter occurs in various admin files and is used to check if the Back button image has been used in a form, based on this spec:

> The submitted data includes name.x=x-value and name.y=y-value where "name" is the value of the name attribute, and x-value and y-value are the x and y coordinate values, respectively.

Note the submitted parameters are back.x and back.y NOT back_x or back_y

Certainly in gv_mail, the Back button submits the form, and the subsequent action check should not send the email but since it checks for back_x being set instead of back-x, it DOES send the email same as the Submit button!
I am working on this file, and have removed the image buttons anyway (so the x and y does not apply anymore), but any other files that use this method need to be corrected/modified to use css buttons.

torvista:
currently only in coupon_admin...needs review.

PR #3111, opened 2020-03-10
Admin coupon admin: bugfixes etc.

torvista:
Additionally this needs some html fixes but I'm not doing that (now/at all).


torvista:
Out of time today: on sending an email, gives a php notice for undefined index of back_x.
This use of $_POST['back_x'] needs a review:
https://github.com/zencart/zencart/issues/2783
as appears twice in this file. It's not so clear as was fixed in gv_mail (compare 156 and 157 for further info).

Issue #2993, opened 2020-01-13
layout settings, box widths reduced to integers: width has no units


torvista:
eg:
Admin->Configuration->Layout Settings->Column width - Left Boxes
default is 150px and text states px may be used.
In fact this is stripped by this:
`$column_width = (int)BOX_WIDTH_LEFT;`
leading to this:
`<div class="leftBoxContainer" id="<?php echo str_replace('_', '-', $box_id ); ?>" style="width: <?php echo $column_width; ?>">`
resulting in this:
`<div class="leftBoxContainer" id="documentcategories" style="width: 150">`
and this warning

> CSS: “width”: only “0” can be a “unit”. You must put a unit after your number.

So, while this can be easily fixed to be consistent and remove the error for px, does it not need to be re-thought to handle any units?

drbyte:
Units have been forced to `px` directly within the template via https://github.com/zencart/zencart/pull/3442

PR #3116, opened 2020-03-12
remove leading and trailing spaces in title tags

torvista:
to lay to rest a long burning issue...
https://www.zen-cart.com/showthread.php?217229-zen_image-Why-the-extra-spaces-in-the-title-attribute

### DIFF
--- a/admin/includes/classes/split_page_results_new.php
+++ b/admin/includes/classes/split_page_results_new.php
@@ -107,7 +107,7 @@ class splitPageResultsNew{
     if (zen_not_null($parameters) && (substr($parameters, -1) != '&')) $parameters .= '&';
 
     // previous button - not displayed on first page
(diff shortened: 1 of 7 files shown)


drbyte:
I still don't understand why people (or "validators" even) are so bothered about this. What exactly is the technical "reason" for it being a "problem"?

Ya, we'll merge this. But will somebody please answer the "why" question with a legit practical reason?

wiztechinc:
The main reason for me is so that when template writers check their own work, they don't have to wade through errors and warnings they didn't create. That said, that particular problem is not highlighted with the checker I use. There are plenty of other errors and warnings that appear. You can see the result on the 1.5.7 cart (as of a few days ago) here: https://validator.w3.org/nu/?doc=http%3A%2F%2Fdeliasdomains.com%2F1_5_7%2F 

I do think we need to be aware of these issues and correct them as we go along. But I do not consider this something to stress over.

torvista:
Of course it is not a real problem, it's just not supposed to be there and serves no real purpose apart from prompting a few nit pickers to ask why it is there. So why not get rid of it.

torvista:
>other errors and warnings that appear
already reported:
https://github.com/zencart/zencart/issues/2993

PR #3117, opened 2020-03-12
Remove these old files

scottcwilson:


drbyte:
They're part of the 3D-secure code. Can't remove them until/unless that code no longer needs them.

scottcwilson:
These are admin side references only - what if we update the links to point to zen-cart.com (and install these files under docs).

drbyte:
Admin side? 
No, they're part of checkout, catalog-side:
https://github.com/zencart/zencart/blob/98fb59484b4ad1bfb1572b11f9e29cfd86b7992f/includes/modules/payment/paypaldp.php#L358-L366

scottcwilson:
ack you're right - for some reason I thought this was shown on the admin install screen. 
I wish we could reference these files from PayPal or something rather than including them in the build.

drbyte:
I agree with you. 
A future approach would probably best implement a modal that's part of a sub-template that loads only on checkout_payment. But even that has its complexities when it comes as part of a payment module.

PR #3124, opened 2020-03-14
bugfix: constant used for category, used for product with sprintf

torvista:
Not needed: removed from Move Category.
Renamed constant for clarity.


torvista:
This morphed into changes for move product too, due to finding another error there, followed by some formatting.

PR #3125, opened 2020-03-15
Common html head

Zen4All:
This has been a long standing issue in Zen Cart, and a we all now it is a pain to update all the duplicate code if a change is required.
This PR is my attempt to create a common html head, that can be used in all admin files.
I tried to keep backward compatibility for legacy pages (like ad-dons), that do not yet make use of the new common html head.
The new head also provides a simple css and js auto-loader. It will search the css an javascript foleder for a file named similar as the page you try to open.
So if there is a configuration.css file in the css folder it will automatically load on the configuration.php page
I add three pages as an example, developers_tool_kit.php auto-loads its own css file.


torvista:
Great work.
Something I appreciate is having the name of the admin page in the page title so it shows on the browser tab (in the case of category listing I use the current category path or product name), as I often have multiple tabs open.
Is that contemplated?

Zen4All:
@torvista Yes

proseLA:
:clap: :clap:

big fan. suggestion, instead of title, what are your thoughts about something like:
```php
echo 'Admin:  ' . ucwords(str_replace('_', ' ', basename($_SERVER["SCRIPT_FILENAME"], '.php')));
```
i find tiresome the maintenance of titles for every page, especially as i create new admin pages.  especially as it is mostly the same thing that this code will generate.

Zen4All:
@proseLA For now I just want to push this as is, and after that we can propose other changes in a separate PR. This way we do not introduce to much changes in one PR, making it easier to track possible bugs.

drbyte:
You're all familiar with a bunch of plugins that affect a bunch of admin files, way more than I am.

Please comment on the degree of pain this introduces for people installing/using/maintaining/upgrading-to those plugins.

lat9:
Is the intent here that the `admin_html_head.php` is _available for use_ as opposed to **always loaded**?

Zen4All:
My idea is to gradually replace the current head with this file.
It is still possible, to add you own CSS and JavaScript files to the head, by adding the links below the require for the admin_html_head.php file

admin/includes/admin_html_head.php
```diff
@@ -0,0 +1,25 @@
...
+<meta http-equiv="X-UA-Compatible" content="IE=edge">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title><?php echo TITLE; ?></title>
+<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css" integrity="sha384-HSMxcRTRxnN+Bdg0JdbxYKrThecOKuH5zCYotlSAcp1+c8xmyTe9GYg1l9a69psu" crossorigin="anonymous">
+<link rel="stylesheet" href="includes/css/font-awesome.min.css">
+<link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
+<link rel="stylesheet" href="includes/css/jAlert.css">
+<link rel="stylesheet" href="includes/css/menu.css">
+<link rel="stylesheet" href="includes/css/stylesheet.css">
+<?php if (file_exists('includes/css/' . $_GET['cmd'] . '.css')) { ?>
```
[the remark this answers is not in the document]
Zen4All: Yes, I will change that

admin/includes/admin_html_head.php
```diff
@@ -0,0 +1,25 @@
...
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title><?php echo TITLE; ?></title>
+<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css" integrity="sha384-HSMxcRTRxnN+Bdg0JdbxYKrThecOKuH5zCYotlSAcp1+c8xmyTe9GYg1l9a69psu" crossorigin="anonymous">
+<link rel="stylesheet" href="includes/css/font-awesome.min.css">
+<link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
+<link rel="stylesheet" href="includes/css/jAlert.css">
+<link rel="stylesheet" href="includes/css/menu.css">
+<link rel="stylesheet" href="includes/css/stylesheet.css">
+<?php if (isset($_GET['cmd']) && $_GET['cmd'] != '' && file_exists('includes/css/' . $_GET['cmd'] . '.css')) { ?>
+  <link rel="stylesheet" href="includes/css/<?php echo $_GET['cmd']; ?>.css">
```
drbyte, line 21: Let's avoid using `$_GET['cmd']` anywhere.

`$PHP_SELF` should be sufficient. Or maybe `basename($PHP_SELF)`

(also in the js require)

admin/includes/admin_html_head.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/**
+ * @package admin
+ * @copyright Copyright 2003-2016 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: admin_html_head.php  New in v1.5.7 $
+ */
+?>
+<meta charset="<?php echo CHARSET; ?>">
```
drbyte, line 10: Can we add a check to abort if `IS_ADMIN_FLAG` isn't defined. That prevents malicious actors from reaching in and outputting stuff that either reveals things it shouldn't or at least avoids filling error logs with notices about undefined vars etc.

Zen4All: done

admin/includes/admin_html_head.php
```diff
@@ -9,6 +9,7 @@
 if (!defined('IS_ADMIN_FLAG')) {
   die('Illegal Access');
 }
+$cmd = str_replace('.php', '', basename($PHP_SELF));
```
drbyte, line 12: I don't think you need this.

admin/includes/admin_html_head.php
```diff
@@ -20,8 +21,8 @@
 <link rel="stylesheet" href="includes/css/jAlert.css">
 <link rel="stylesheet" href="includes/css/menu.css">
 <link rel="stylesheet" href="includes/css/stylesheet.css">
-<?php if (isset($_GET['cmd']) && $_GET['cmd'] != '' && file_exists('includes/css/' . $_GET['cmd'] . '.css')) { ?>
-  <link rel="stylesheet" href="includes/css/<?php echo $_GET['cmd']; ?>.css">
+<?php if (file_exists('includes/css/' . $cmd . '.css')) { ?>
+  <link rel="stylesheet" href="includes/css/<?php echo $cmd; ?>.css">
```
drbyte, line 25: How about:
```php
<?php if (file_exists($file = 'includes/css/' . basename($PHP_SELF, 'php') . '.css')) { ?>
   <link rel="stylesheet" href="<?php echo $file; ?>">
```

drbyte:
I'm not sure if my resolution of the merge conflict was correct. I suspect there were some CSS changes from recent commits that probably need to be moved to the new file .

proseLA:
> My idea is to gradually replace the current head with this file.
> It is still possible, to add you own CSS and JavaScript files to the head, by adding the links below the require for the admin_html_head.php file

this gradual replacement makes it much more difficult for those of us attempting to upgrade.

in the future can we not do this gradual replacement?  lets do **WHOLESALE** changes.  i'm a **BIG** fan of this ```admin_html_head``` file.  i had previously done my own quick and dirty version of it.  now it seems that i have to screw around with headers in multiple places.  again.

i think there are enough people testing new versions that gradual replacement, while it may seem easier at the time, just makes it that much harder for other members of the community.

PR #3128, opened 2020-03-16
Remove debug comments

scottcwilson:


drbyte:
No thanks. These are still useful when debugging.

PR #3130, opened 2020-03-18
Code reformatting

scottcwilson:
This is such a nice practice -let's use it everywhere.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -922,7 +922,7 @@ class order extends base {
                                   'attributes_price_letters_free' => $attributes_values->fields['attributes_price_letters_free'],
                                   'products_options_id' => (int)$this->products[$i]['attributes'][$j]['option_id'],
                                   'products_options_values_id' => (int)$this->products[$i]['attributes'][$j]['value_id'],
-                                  'products_prid' => $this->products[$i]['id']
+                                  'products_prid' => $this->products[$i]['id'],
                                   );
 
           zen_db_perform(TABLE_ORDERS_PRODUCTS_ATTRIBUTES, $sql_data_array);
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -1045,9 +1045,9 @@ class order extends base {
     $html_msg['EMAIL_ORDER_DATE']      = date(ORDER_EMAIL_DATE_FORMAT);
     $html_msg['EMAIL_TEXT_TELEPHONE']  = EMAIL_TEXT_TELEPHONE;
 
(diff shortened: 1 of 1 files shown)


proseLA:
1. why do you feel this is a “nice” practice.  it seems there are arguments are both sides.
2. if this is going to be the new practice, lets all agree to it and then add it to the developer docs.
3. we already have tons of legacy coding standards.  2 spaces for legacy scripts, PSR-2 for new ones going forward.  if we want to make it a new standard, lets do it going forward.  if there are no other differences in the code, lets leave well enough alone.
4. as someone who maintains a few live ZC carts, this is the sort of change that infuriates me whenever I do an update.  there is no FUNCTIONAL difference in the code, and yet it is something that I have to look at using diff because someone decided this is a better practice.  I am reminded of times when someone decided that parts of the sql statement should be in caps and other parts not in caps.  make a standard going forward.  and lets not touch legacy code for the sake of the standard.
5. PLEASE, PLEASE, PLEASE, lets NOT make innocuous changes to legacy code just because it is “nice”.  lets agree on a policy going forward for coding standards, and not make these changes just for the sake of making a change.

scottcwilson:
Look carefully at the 1.5.6 code for  this file.  It is done everywhere except in this one place.

scottcwilson:
The rationale behind doing this is that if you have to add items to the array, the last entry doesn't have to change (with a new ",").

wiztechinc:
If I understand what you are talking about, I truly prefer this way instead of without the comma. But then again, I'm no longer creating or maintaining mods anymore for others (just for me now). I might feel differently otherwise.

drbyte:
While "yes", this "keep a trailing comma after the last array entry, when entries span multiple lines" is preferable, and is even the PSR standard, I also agree that generally we ought to avoid creating unnecessary upgrade pain wherever possible.
(I'm aware of the irony that making this specific sort of change actually directly reduces upgrade pain once it's been done.)

So, yes, it should be in the coding standards (it is insomuch as PSR applies)

But ... let's **not** go looking for files to apply this format to.

proseLA:
i will say one last thing (in order to clarify my point) and then move on.  i stand by what drB says...

with regards to the specific changes in this PR, there is no functional difference so i am inherently against that.

HOWEVER, between the current released version of ZC (v156c) and the new v157 version prior to this PR, it looks like there are changes already made to this file.  which necessitates someone already reviewing the file when doing an upgrade.  so i am not so vehemently opposed to the changes.

nothing infuriates me more (ok that's an overstatement), when reviewing changes on a file for an upgrade, and the only thing different is the added comma to the end of the array, and a new stamp at the top of the code.

drbyte:
Merging because we've already touched this file aggressively for the next release.

PR #3146, opened 2020-03-27
If comments are removed after being added, clear SESSION data

scottcwilson:


drbyte:
What's the scenario that this is fixing?

scottcwilson:
Repro steps: 
- on checkout shipping, enter comments, press continue. 
- on payment page, click checkout (for example, you realize the shipping you have chosen is too expensive) and go back to checkout.
- now you're back on the checkout shipping page.  change your shipping method. 
- decide you don't want the comments after all, and remove them from the checkout shipping page.
- click continue to go to payment.  Oops - comments are still there.

drbyte:
Does the same problem exist with comments that get blanked out on checkout-payment page?

scottcwilson:
No - if you remove comments on the payment page and submit, the confirmation page correctly shows no comments.

drbyte:
How about instead of adding an `else` clause, why don't we just change things like this:

```diff
-    if (isset($_POST['comments'])) {
+    if (zen_not_null($_POST['comments'])) {
```

scottcwilson:
I'm not following your suggestion.

drbyte:
You did the following:

```diff
    if (zen_not_null($_POST['comments'])) {
       $_SESSION['comments'] = zen_output_string_protected($_POST['comments']);
+     } else {
+       $_SESSION['comments'] = ''; 
     }
```

I'd rather you did this instead:
```diff
-    if (zen_not_null($_POST['comments'])) {
+    if (isset($_POST['comments'])) {
       $_SESSION['comments'] = zen_output_string_protected($_POST['comments']);
     }
```

PR #3149, opened 2020-03-29
for a duplicated product, add the suffix DUPLICATE to the name

torvista:
my IT team (all 1 of him) liked this....


drbyte:
I thought we already added '[COPY]' to the product `name`.
Apparently we don't. We should.

While I understand `your` business is heavily centered on `model number`, that's a lot more specific than most stores get. 
Thus I'd prefer to see it added to the product name.

I guess it should support multiple languages as well.

Whether it's `[COPY]` or `[DUPLICATE]` I have no strong opinion.

drbyte:
I guess this should be carried over to other product types too?

torvista:
I don't think so, where?

drbyte:
Looks like just `products_music` has an override in this case.

drbyte:
Oy. I didn't realize that one was so far behind.
Thanks.

PR #3153, opened 2020-03-30
Change admin home page number_of_ variables to constants

scottcwilson:


scottcwilson:
This is an EVEN BETTER idea but let me get initial approval for this change from @drbyte before moving forward.

drbyte:
Why do these need to be constants in the first place? 
How often are these default values being changed? 
Is the "real" need just to change the defaults to something more "common"?

lat9:
FWIW, I've got a couple of stores that I manage that change those values.

scottcwilson:
It was good to get these values out of the SELECT statement because if you change those values (and many of my clients do), you get this HUGE diff.
It is better to make them constants so it's obvious when you're looking at the code what is going on.
Cindy's suggestion is even better because then you could change them more easily - with the caveat that admin doesn't have overrides so this is a bit of a pain, but still worthwhile.

drbyte:
As soon as we make them constants someone else is going to come along and want to add even more Admin config switches for them.

Can't we just set some reasonable values that will work for the majority? What are those values that you're usually changing to?

I'm not really bothered by a tiny diff where the changes are extremely clear, as they are if one or all three of these lines are edited for a given store. It's the least of complexities in any upgrade.

scottcwilson:
Is the compromise just to merge my PR as is to reduce complexity?  It would be a very small diff if all you have done is change these values.

drbyte:
> Is the compromise just to merge my PR as is to reduce complexity? It would be a very small diff if all you have done is change these values.

How does as-is reduce any complexity? It changes nothing other than adding more memory overhead with more defined constants if people start to try to override them.

Again, what are the values you're changing these to?

scottcwilson:
I don't think there's a single set of values you can use - but I do think these should be defined constants and not variables.   I was saying "not adopt Cindy's solution" to reduce complexity.   Just take the PR as is.

lat9:
> 
> 
> It was good to get these values out of the SELECT statement because if you change those values (and many of my clients do), you get this HUGE diff.
> It is better to make them constants so it's obvious when you're looking at the code what is going on.
> Cindy's suggestion is even better because then you could change them more easily - with the caveat that admin doesn't have overrides so this is a bit of a pain, but still worthwhile.
Actually, if you create an /admin/includes/extra_datafiles/something.php that contains
```
define('NUMBER_OF_CUSTOMERS', '30');
```
that definition will be the overriding value.

drbyte:
Still, nobody's proposing better defaults or saying what they're changing them to.
If the defaults aren't wrong, why are you changing them?

lat9:
> 
> 
> Still, nobody's proposing better defaults or saying what they're changing them to.
> If the defaults aren't wrong, why are you changing them?

At least one site shows 5 new customers, 5 new orders and 0 history records, choosing to display "more important" information on that main page.

scottcwilson:
The customer that I'm working on now keeps the default customers and orders but increases the history to 40.  There's no single answer.

drbyte:
Okay, let's go ahead, but 2 changes:

- as lat9 suggested add the `if-not-defined`
- Since you're using constants that can be defined externally, then you MUST sanitize (`(int)`) them in the db query, else you risk creating vulnerabilities.

And squash the commits into one (git amend)

PR #3154, opened 2020-03-30
set 1 config to do 2 things

proseLA:
i'm not sure if this a fight worth having.  it would be nice to combine these constants.  what i have done here is by setting `ADMIN_BLOCK_WARNING_OVERRIDE` you no longer see the zc_install message.  ideally i would prefer to remove the other constant, but i suppose i can see the value in it.  as someone who debugs ZC constantly, it would be nice to not have so many config switches.  i'm sure most store owners never use them.

### DIFF
--- a/admin/includes/init_includes/init_admin_auth.php
+++ b/admin/includes/init_includes/init_admin_auth.php
@@ -29,6 +29,7 @@ if (basename($PHP_SELF) == FILENAME_AJAX . '.php') {
 // admin folder rename required
 if (!defined('ADMIN_BLOCK_WARNING_OVERRIDE') || ADMIN_BLOCK_WARNING_OVERRIDE == '')
 {
+  define('ADMIN_BLOCK_WARNING_OVERRIDE', false);
   if (basename($PHP_SELF) != FILENAME_ALERT_PAGE . '.php')
   {
     if (substr(DIR_WS_ADMIN, -7) == '/admin/' || substr(DIR_WS_HTTPS_ADMIN, -7) == '/admin/')
--- a/admin/includes/init_includes/init_errors.php
+++ b/admin/includes/init_includes/init_errors.php
@@ -69,7 +69,7 @@ if (!defined('WARN_DOWNLOAD_DIRECTORY_NOT_READABLE')) define('WARN_DOWNLOAD_DIRE
  */
 if (!defined('WARN_DATABASE_VERSION_PROBLEM')) define('WARN_DATABASE_VERSION_PROBLEM','true');
 // check if the installer directory exists, and warn of its existence
-if (WARN_INSTALL_EXISTENCE == 'true') {
+if (WARN_INSTALL_EXISTENCE == 'true' && ADMIN_BLOCK_WARNING_OVERRIDE !== 'true') {
   $check_path = realpath(DIR_FS_CATALOG . '/zc_install');
   if (is_dir($check_path)) {
     $messageStack->add(sprintf(WARNING_INSTALL_DIRECTORY_EXISTS, ($check_path == '' ? '..../zc_install' : $check_path)), 'warning');


admin/includes/init_includes/init_errors.php
```diff
@@ -69,7 +69,7 @@
  */
 if (!defined('WARN_DATABASE_VERSION_PROBLEM')) define('WARN_DATABASE_VERSION_PROBLEM','true');
 // check if the installer directory exists, and warn of its existence
-if (WARN_INSTALL_EXISTENCE == 'true') {
```
scottcwilson: Noooooooo!  This is not needed and will be a pain for people who are working on the next version out of Github.

proseLA: i don't follow why it would be a pain.  can you elaborate your thoughts, please.

scottcwilson: If you are working in the zen-cart.com Github repo you need both of these flags on.

drbyte: @scottcwilson I'm not sure I understand.

@proseLA I'm undecided. Is this just to avoid defining "2" constants in dev?
Also, is using `false` boolean vs `'true'` string a good idea?

admin/includes/init_includes/init_admin_auth.php
```diff
@@ -29,6 +29,7 @@
 // admin folder rename required
 if (!defined('ADMIN_BLOCK_WARNING_OVERRIDE') || ADMIN_BLOCK_WARNING_OVERRIDE == '')
 {
+  define('ADMIN_BLOCK_WARNING_OVERRIDE', false);
```
drbyte: Hmmm ... if we get here via the `!defined()` part, this is okay. But if we get here by the `== ''` part then this will throw an error about already being defined, won't it?

proseLA: that does sound correct.

admin/includes/init_includes/init_errors.php
proseLA: i'm a fan of the boolean.  and i'm not opposed to changing it as well as the define that you mentioned above.  the point is to avoid the constants, yes.  i'm not sure why these 2 need to be separated.  IMO, there are so many constants that one needs to figure out which constant to change prior to actually changing it.  when one is doing ZC 24/7; perhaps one remembers all of these things.  but it would be nice to combine so much of that.  especially when some of us look away from ZC and then come back.
@scottcwilson this PR (except for the points raised by @drbyte) should not affect turning both of these flags on in development or elsewhere.  unless i am missing something...

drbyte: > perhaps one remembers all of these things

nah, I just copy in my override files and they take care of it. No thinking or remembering involved! ☺️

scottcwilson:
Here's my complaint: 
ADMIN_BLOCK_WARNING_OVERRIDE is innocuous in a deployed config file as long as you have renamed the admin, which you would as part of your deploy and certainly after the upgrade when you go back in to take the site out of maintenance, etc.

WARN_INSTALL_EXISTENCE should only be needed by people working on the next version of Zen Cart who need to have a zc_install folder in their cart directory.   Everyone else should rename or delete that folder and should be warned if they haven't.

Lots of my deployed carts still have ADMIN_BLOCK_WARNING_OVERRIDE in their configure files, because it doesn't affect anything so I never removed it when updating the config files for deployment.  But I still want to be warned to remove the zc_install folder.  I could easily forget after running the process and going to admin to update things to do so.  Having two separate flags ensures that I don't forget.

proseLA:
just so i'm clear @scottcwilson, you have overridden the ADMIN warning on all your sites, because you always remember to change the admin.  but you might forget to remove the zc_install directory, so you want to be reminded about that?

if this is correct, i am not sure that these factors are what should be the driver for development.  but that is just my opinion.

without getting into a bigger discussion on this PR, i just want to point out the arguments about security through obscurity (which we have had before).  and maybe we can potentially add 2 factor authentication for both of these functions.  and then make this whole point moot.

torvista:
I put these switches (and strict mode) only in /local/configure.php. Works for me, never gets out into the wild.

scottcwilson:
I rename zc_install if I need to keep it around and don't want to get warnings.  No code changes needed!

proseLA:
last attempt:

right now, your workflow -> determines code.
i think code -> determine workflow.

what is best for the code?  
i think all of these constants are ridiculous.  and there should be a concerted effort to simplify, not to create complexity.

legacy code is always difficult to re-engineer.  this gets to a similar discussion we had on the use of `!important` in another PR which i do not remember if it got resolved.  back then, you seemed to imply we were close to a new release...   i suppose we can have different opinions on close.

scottcwilson:
We have two constants for two different things. Not sure why you think this is ridiculous.

drbyte:
> you have overridden the ADMIN warning on all your sites, because you always remember to change the admin. but you might forget to remove the zc_install directory, so you want to be reminded about that?

That's exactly the purpose of having two separate constants. 

Sorry if this is unpopular, but I'm going to decline this proposed change.

PR #3155, opened 2020-03-30
Update zen_mail, enabling the $mail object to be updated via observer.

lat9:
Enables the setting of additional cc:/bcc: addresses.  Note that observers need to check that `$p2` isn't null (i.e. previous release) when attaching to the notification.


drbyte:
@lat9 good idea. I've also added the `$attachments_list` as another subsequent param so it can be modified directly instead of needing to use the `global`. I left the other wrappings for backward compatibility.

PR #3156, opened 2020-03-30
Changes to Product collect info

torvista:
I intended to just rework the image section the same as I just did for the category edit, to make it more intuitive.
But, there was more to do...


torvista:
The labels that referred to a single input I linked to an id on the input
The labels that referred to multiple inputs (radio buttons) I changed: as far as I know, a label has to refer to one input, not multiple.

PR #3161, opened 2020-03-31
Copy * instead of specific fields on product copy

scottcwilson:
The current copy logic will miss additional fields added to the products table.  Copying all fields (except for the ones requiring special handling) means this fill will need less frequent customization.

### DIFF
--- a/admin/includes/modules/copy_product_confirm.php
+++ b/admin/includes/modules/copy_product_confirm.php
@@ -31,13 +31,7 @@ if (isset($_POST['products_id'], $_POST['categories_id'])) {
         }
     } elseif ($_POST['copy_as'] == 'duplicate') {
 
-        $product = $db->Execute("SELECT products_type, products_quantity, products_model, products_image,
-                                    products_price, products_virtual, products_date_available, products_weight,
-                                    products_tax_class_id, manufacturers_id,
-                                    products_quantity_order_min, products_quantity_order_units, products_priced_by_attribute,
-                                    product_is_free, product_is_call, products_quantity_mixed,
-                                    product_is_always_free_shipping, products_qty_box_status, products_quantity_order_max, products_sort_order,
-                                    products_price_sorter, master_categories_id
+        $product = $db->Execute("SELECT * 
                              FROM " . TABLE_PRODUCTS . "
                              WHERE products_id = " . $products_id);
 
@@ -47,6 +41,7 @@ if (isset($_POST['products_id'], $_POST['categories_id'])) {
                     SET products_quantity_order_units = 1
                     WHERE products_id = " . $products_id;
             $results = $db->Execute($sql);
+            $product->fields['products_quantity_order_units'] = 1; 
         }
         // fix Product copy from if Minimum is 0
         if ($product->fields['products_quantity_order_min'] == 0) {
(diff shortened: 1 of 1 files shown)


torvista:
I understand that all the Select query results come out as strings.
There is no need/is it not good practice to cast the types on Insert (like the integers)?

scottcwilson:
This logic works even with DB_MYSQL_MODE = STRICT.  The casting is not required.

torvista:
ok.

admin/includes/modules/copy_product_confirm.php
```diff
@@ -47,52 +41,33 @@
...
+        $sql_data_array = array(); 
+        foreach ($product->fields as $key => $value) {
+          switch ($key) {
+          case products_id: 
+          case products_last_modified: 
+          case products_date_added: 
+          case products_date_available: 
+            continue 2;
+          }
+          $sql_data_array[$key] = $value;
```
drbyte, line 64: NOOOOOOO! This would create a huge security hole.

`$value` here needs to be `zen_db_input($value)`

admin/includes/modules/copy_product_confirm.php
```diff
@@ -47,52 +41,33 @@
...
-                          '" . zen_db_input($categories_id) . "')");
+        $sql_data_array = array(); 
+        foreach ($product->fields as $key => $value) {
+          switch ($key) {
+          case products_id: 
+          case products_last_modified: 
+          case products_date_added: 
+          case products_date_available: 
+            continue 2;
+          }
```
drbyte, line 63: While this `switch/case/continue` logic works, the cognitive load of figuring out what this is doing is pretty high.

I'd prefer to see it changed to something clearer like:
```php
if (in_array($key, [
    'foo',
    'bar',
    'baz',
]) {
    continue;
}
```
While the difference is small, this is a bit clearer, and still easy to customize if someone wants to add more fields to it.

admin/includes/modules/copy_product_confirm.php
```diff
@@ -47,52 +41,33 @@
...
             $results = $db->Execute($sql);
+            $product->fields['products_quantity_order_min'] = 1; 
         }
 
-        $tmp_value = zen_db_input($product->fields['products_quantity']);
-        $products_quantity = (!zen_not_null($tmp_value) || $tmp_value == '' || $tmp_value == 0) ? 0 : $tmp_value;
-        $tmp_value = zen_db_input($product->fields['products_price']);
-        $products_price = (!zen_not_null($tmp_value) || $tmp_value == '' || $tmp_value == 0) ? 0 : $tmp_value;
-        $tmp_value = zen_db_input($product->fields['products_weight']);
-        $products_weight = (!zen_not_null($tmp_value) || $tmp_value == '' || $tmp_value == 0) ? 0 : $tmp_value;
```
drbyte, line 64: I'd like to see this change still accommodate specific handling for `float`/`decimal` fields like these (and also for `integer` fields), particularly when the incoming input is left blank; otherwise you'll have SQL Strict errors.

drbyte:
> This logic works even with DB_MYSQL_MODE = STRICT. The casting is not required.

Yes and no. Incoming blank strings for decimal/integer fields *do* still need to be cast.

admin/includes/modules/copy_product_confirm.php
scottcwilson: Fixed -but this data is already in the database - it's not new user input.

admin/includes/modules/copy_product_confirm.php
scottcwilson: Roger that.  Will do shortly.

admin/includes/modules/copy_product_confirm.php
scottcwilson: Done.

admin/includes/modules/copy_product_confirm.php
drbyte: > Fixed -but this data is already in the database - it's not new user input.

Good point. I'd forgotten this was for "copy", and not collect/preview like other current open PRs are.

scottcwilson: It's not a bad change just in case bad data snuck in many moons ago.  Also it's good in case folks copy and paste this naively.

scottcwilson:
Love the concept - would you do that as a PR after this is merged?

lat9:
Sure, if I remember!

scottcwilson:
LOL I will ping you.  It will take me a day or so to do the other requested changes.

scottcwilson:
> Yes and no. Incoming blank strings for decimal/integer fields _do_ still need to be cast.

Done

admin/includes/modules/copy_product_confirm.php
```diff
@@ -47,52 +41,56 @@
...
+              $sql_data_array[$key] = (!zen_not_null($value) || $value == '' || $value == 0) ? 0 : $value; 
+            }
+          } else { 
+            $sql_data_array[$key] = $value; 
+          }
+        }
+
+        // separately_updated_fields - last_modified and products_id are skipped
+        $sql_data_array['products_date_added'] = 'now()';
+        $sql_data_array['products_date_available'] = (!empty(zen_db_input($product->fields['products_date_available'])) ? zen_db_input($product->fields['products_date_available']) : 'null');
```
drbyte, line 90: This could be simplified slightly by removing the first call to zen_db_input during the conditional check:
```diff
-  $sql_data_array['products_date_available'] = (!empty(zen_db_input($product->fields['products_date_available'])) ? zen_db_input($product->fields['products_date_available']) : 'null'); 
+  $sql_data_array['products_date_available'] = (!empty($product->fields['products_date_available']) ? zen_db_input($product->fields['products_date_available']) : 'null'); 
```

admin/includes/modules/copy_product_confirm.php
```diff
@@ -47,52 +41,56 @@
...
-                                                      products_sort_order, products_price_sorter, master_categories_id)
-                  VALUES ('" . zen_db_input($product->fields['products_type']) . "',
-                          '" . $products_quantity . "',
-                          '" . zen_db_input($product->fields['products_model']) . "',
-                          '" . zen_db_input($product->fields['products_image']) . "',
-                          '" . $products_price . "',
-                          '" . zen_db_input($product->fields['products_virtual']) . "',
-                          now(),
-                          " . (zen_not_null(zen_db_input($product->fields['products_date_available'])) ? "'" . zen_db_input($product->fields['products_date_available']) . "'" : 'null') . ",
-                          '" . $products_weight . "', '0',
```
drbyte, line 81: The `0` here refers to `products_status`, which we hard-code to 0 on copy.
But I don't see that accommodated in this PR. Was that intentional?

scottcwilson: - Nope - accidental miss.

drbyte:
> @lat9 wrote: 
> How about a notification between the base $separately_updated_fields and that foreach, just in case a store's customizations have other fields to be 'separately updated'?

Does simply using the existing notifier `NOTIFY_MODULES_COPY_TO_CONFIRM_DUPLICATE` work for this? 
NOTE: the logic has been changed such that individual fields are all being copied forward. While it may be sensible to intercept each of those in real time, doing special handling afterward may be sufficient. Just something to consider when you're looking at your proposed change.

PR #3164, opened 2020-03-31
Select p.* for collect, not specific fields

scottcwilson:


drbyte:
Hmmm ... 

A downside to this is that one no longer has a reference to what the actual field names are without going out to the actual database to investigate.

scottcwilson:
> A downside to this is that one no longer has a reference to what the actual field names are without going out to the actual database to investigate.
There's documentation on that! :) 
https://docs.zen-cart.com/dev/schema/schema_156/#products

drbyte:
> There's documentation on that! :)

Perhaps, but I'm undecided about keeping that in the docs, because it's volatile.

scottcwilson:
It's not volatile once a release is done, and it's done "per version" so you can go to "your" release.   It's volatile for 1.5.7, but I haven't put 1.5.7 out there for that very reason.

drbyte:
Note to self: the field names are (mostly) listed in the array at the top of the file.

PR #3165, opened 2020-04-01
Bugfix: Messagestack add_session not working

torvista:
In
https://github.com/zencart/zencart/pull/3072
A check clause was added to stop a php notice, but it was added around **all** of the case 'objectMethod':
Consequently this never fired the class->method add_session as that item is never in the SESSION array for stacked messages: it's called  "messageToStack".

Also, the previous PR from Wilt added 
if (!is_object($_SESSION[$objectName]))
which causes an php warning as again, SESSION messageToStack does not exist.
In my limited understanding, I don't think that check is needed, as the method messagestack->add_from_session is then used, which **does** check for the correct item (messageToStack).

Works for me, but I am not very aware of the big picture so....

### DIFF
--- a/includes/autoload_func.php
+++ b/includes/autoload_func.php
@@ -44,14 +44,11 @@ foreach ($initSystemList as $entry) {
             //echo 'objectMethod ' . $entry['class'] . "\n";
             $objectName = $entry['object'];
             $methodName = $entry['method'];
-            if (isset($_SESSION[$objectName])) {
-              if (is_object($_SESSION[$objectName])) {
+              if (isset($_SESSION[$objectName]) && is_object($_SESSION[$objectName])) {
                   $_SESSION[$objectName]->$methodName();
-              }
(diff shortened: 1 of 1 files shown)


lat9:
@torvista, that looks like the correct change, since the admin-level `$messageStack` is not instantiated in the session so that the previous version wouldn't be running the `add_from_session` method.

proseLA:
the changing of all of these commented out echo statements looks like a good opportunity to make use of one of the numerous debug constants already out there.  would it be nice to turn on a debug constant and then have all of these things echo?   if not, then why do we have these commented lines in here?  and why are there changes to commented out code?

torvista:
Purists would argue to remove them completely, but they helped me. I thought about making them dependent on another constant, but decided that it would have produced too much screen output when, at least in this case, I was only looking at one aspect and was not even sure if this "fix" was the best course of action anyway. I changed the echo's to make them more readable.

PR #3166, opened 2020-04-01
Admin: category image changes

torvista:
Bugfix: from prior PR. When no image is defined, no option is presented to delete it so the POST img_delete is not created/passed and a warning is triggered on the preview page.
After reviewing the three options in the image section (delete /upoad new / use existing), I decided to simplify it, rather than just added isset.

The manual filename entry uses the selected subdirectory, which was not evident (not to me anyway), and no check was made on the filename existing: added check and corresponding messagestack texts.

The image shown on the page was at full size: added parameters to zen_info_image to allow use of responsive class.
Added file info (path, sizes, weight) to image info.
Clipboard01

### DIFF
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -405,7 +405,11 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                     <div class="col-sm-offset-3 col-sm-9 col-md-6">
                         <div><?php echo zen_info_image($cInfo->categories_image, $cInfo->categories_name, '', '', 'class="table-bordered img-responsive"'); ?></div>
                         <br>
-                        <?php echo $cInfo->categories_image; ?>
+                        <?php
+                        [$width, $height] = getimagesize(DIR_FS_CATALOG_IMAGES . $cInfo->categories_image);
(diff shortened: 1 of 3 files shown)


drbyte:
This appears to remove the ability to enter `none` in order to delete the image. 
Is that intentional?

torvista:
I was not sure what the none was about, is that documented somewhere / is it necessary when there is a radio button to remove the image?

drbyte:
I've never thought of it as an undocumented feature, but we've always supported this `none` option (pretty much) everywhere we support image edits.
You raise a fair point. Do you think we should remove the `none` feature for everyone?

torvista:
Well, obviously I won't miss what I never knew about...but for other users??

wiztechinc:
I certainly didn't know about it. Using the radio button only deletes it from the database - not actually deletes the image itself. So when you enter none, it's the same function as the radio button and does not actually delete the image? If so, why in the world was there even a reason to be able to also enter none?  I did find this documented the estart manual but I never used that thing to learn from in the first place. It also states it doesn't work in IE which is obviously outdated.  So streamline that code!

barco57:
> but we've always supported this none option (pretty much) everywhere we support image edits.

I had no idea

drbyte:
Interesting.

PR #3172, opened 2020-04-03
Messagestack losing session messages (catalog)

### LINKED ISSUE #3171: (storefront) $_SESSION['messageToStack'] messages 'lost' on redirect
The `message_stack` is instantiated during the autoload-function processing and gathers the session-based messages during that instantiation, clearing the session-based messages.

If an observer (like that in One Page Checkout) redirects a page's access based on the page-header processing, those messages are lost.  The session-based message gathering should be deferred until the `output` of the messages is requested.

lat9:
Provides refactoring of code, then moves the `$_SESSION['messageStack']` 'removal' to the class `size` method so that session-based messages aren't lost on a redirect.

Fixes #3171

### DIFF
--- a/includes/classes/message_stack.php
+++ b/includes/classes/message_stack.php
@@ -9,7 +9,7 @@
  * @version $Id: mc12345678 2019 Apr 30 Modified in v1.5.6b $
  */
 if (!defined('IS_ADMIN_FLAG')) {
-  die('Illegal Access');
+    die('Illegal Access');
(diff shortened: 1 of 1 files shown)


torvista:
Is this independent of this?
https://github.com/zencart/zencart/pull/3165

lat9:
Yes, the admin-side message_stack is a totally different 'animal'.

PR #3176, opened 2020-04-04
Refactor for early return

drbyte:
Easier to see the identical patterns when debugging.
In the case of the payment class it makes the code much more like an interface, which could be integrated later particularly by de-duplicating etc.

### DIFF
--- a/includes/classes/payment.php
+++ b/includes/classes/payment.php
@@ -2,11 +2,10 @@
 /**
  * Payment Class.
  *
- * @package classes
- * @copyright Copyright 2003-2019 Zen Cart Development Team
+ * @copyright Copyright 2003-2020 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: lat9 2019 Mar 03 Modified in v1.5.6b $
+ * @version $Id:  Modified in v1.5.7 $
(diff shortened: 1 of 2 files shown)


includes/classes/payment.php
```diff
@@ -92,30 +92,28 @@ function __construct($module = '') {
...
   function update_status() {
-    if (empty($this->selected_module)) return; 
-    if (is_array($this->modules)) {
-      if (is_object($GLOBALS[$this->selected_module])) {
-        if (method_exists($GLOBALS[$this->selected_module], 'update_status')) {
-          $GLOBALS[$this->selected_module]->update_status();
-        }
-      }
-    }
+    if (empty($this->selected_module)) return;
```
lat9, line 111: Any reason for multiple `if` statements instead of aggregating on a single line?

lat9, line 111: Any reason for the multi-if-statements instead of aggregating on a single `if`?

drbyte: Largely easier readability.

drbyte: To clarify: the "early return" concept allows one to quickly see all the criteria that do not apply to the outcome, and then see only the actual real "actions" once all those disqualifiers are out of the way.
Several short `if` statements that just directly return/abort makes human reading of it very quick.

lat9: Different strokes for different folks.

PR #3180, opened 2020-04-06
Admin users - bugfix for new user set profile

### LINKED ISSUE #3179: Saving to new user profile only saves to super_user
@torvista , @drbyte 
https://github.com/zencart/zencart/blob/34fd4324ac76dbfec1cfa9196a846ec8545880c8/admin/users.php#L183

since the above line was added all users are saved as super_users when trying to set another user profile to a user

torvista:
and other minor changes

Fixes #3179


admin/users.php
```diff
@@ -66,7 +66,7 @@
       $profilesList = array_merge([['id' => 0, 'text' => TEXT_CHOOSE_PROFILE]], zen_get_profiles());
     } else {
       $action = '';
-      $messageStack->add(SUCCESS_NEW_USER_ADDED, 'success');
+      $messageStack->add(sprintf(SUCCESS_NEW_USER_ADDED, zen_db_prepare_input($_POST['name'])), 'success');
```
drbyte, line 69: Since we're doing output here, let's use `zen_output_string_protected` instead of `zen_db_prepare_input`.

PR #3186, opened 2020-04-07
Add notifier to admin product-copy-duplicate processing

### LINKED ISSUE #3185: Add notifier to product copy
Per @lat9 
"How about a notification between the base $separately_updated_fields and that foreach, just in case a store's customizations have other fields to be 'separately updated'?" - see #3161

lat9:
Fixes #3185

### DIFF
--- a/admin/includes/modules/copy_product_confirm.php
+++ b/admin/includes/modules/copy_product_confirm.php
@@ -70,6 +70,12 @@ if (isset($_POST['products_id'], $_POST['categories_id'])) {
           'product_is_call' =>  'int', 
           'products_quantity_mixed' =>  'int', 
         ); 
+        
+        // -----
+        // Give an observer the chance to add any customized fields to the two arrays above.
+        //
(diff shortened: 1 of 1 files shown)


admin/includes/modules/copy_product_confirm.php
```diff
@@ -70,6 +70,12 @@
           'product_is_call' =>  'int', 
           'products_quantity_mixed' =>  'int', 
         ); 
+        
+        // -----
+        // Give an observer the chance to add any customized fields to the two arrays above.
+        //
+        $zco_notifier->notify('NOTIFY_MODULES_COPY_PRODUCT_CONFIRM_DUPLICATE_FIELDS', $products_id, $separately_updated_fields, $casted_fields);
```
drbyte, line 77: Would it make sense to use `$product` instead of `$products_id` here, since there's probably a desire to inspect multiple fields, such as product-type?

PR #3169, opened 2020-04-02
Misc cleanup: uppercase SQL, remove obsolete comment and end array on new line

scottcwilson:
These files have already been changed in 1.5.7 - may as well finish the cleanup.

There are more files that have already been changed but haven't had these fixes: 
**in storefront:** 
column_left.php
column_right.php
column_single.php
create_account.php
meta_tags.php
order_total/ot_coupon.php
order_total/ot_gv.php
payment/authorizenet.php
payment/authorizenet_aim.php
payment/firstdata_hco.php
payment/paypal.php
payment/paypal/paypal_functions.php
payment/paypaldp.php
payment/paypalwpp.php
payment/square.php
product_prev_next.php
products_quantity_discounts.php
shipping/freeoptions.php
shipping_estimator.php
sideboxes/best_sellers.php
sideboxes/order_history.php
sideboxes/whos_online.php

**in admin:** 
featured.php
currencies.php
geo_zones.php
products_price_manager.php
gv_queue.php
template_select.php
includes/init_includes/init_templates.php
includes/init_includes/init_db_config_read.php
includes/init_includes/init_cache_key_check.php
includes/init_includes/init_errors.php
includes/header.php
includes/functions/html_output.php
includes/functions/localization.php
includes/functions/admin_access.php
includes/functions/general.php
product_types.php
gv_sent.php
store_manager.php
specials.php
orders_status.php
developers_tool_kit.php
coupon_admin_export.php
manufacturers.php
sqlpatch.php
media_manager.php
reviews.php
coupon_admin.php
popup_image.php
categories.php
customers.php
paypal.php
tax_classes.php
orders.php

PR #3188, opened 2020-04-08
Updating SQL syntax

scottcwilson:
In this sprit of #3169, this set of updates is limited to ensuring the following:
* SQL statements are in upper case
* The last entry in an array has a comma, and the closing bracket is on a new line
* File does not end with ?>
In addition, meaningless comments may be removed.

### DIFF
--- a/includes/modules/column_left.php
+++ b/includes/modules/column_left.php
@@ -13,7 +13,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 }
 $column_box_default='tpl_box_default_left.php';
 // Check if there are boxes for the column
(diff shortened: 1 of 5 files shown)


Zen4All:
When updating these queries, why not remove the single quotes from integer values at the same time?

drbyte:
> When updating these queries, why not remove the single quotes from integer values at the same time?

Yes, but be sure to ONLY do it for values that are clearly cast as integers. 
(eg: Sometimes the products_id is a string, such as the customers_basket_attributes table, etc.)

scottcwilson:
This is a lot more work and a lot more chance of error.  Are you SURE we want to do this?

drbyte:
hmmm

PR #3196, opened 2020-04-09
built in help for admin

scottcwilson:
Works on Admin > Localization > Currencies and Admin > Catalog > Categories/Products.  

Help function could be overridden to open a new page with locally created content as well as default content.

### DIFF
--- a/admin/includes/functions/functions_help.php
+++ b/admin/includes/functions/functions_help.php
@@ -1,0 +1,89 @@
+<?php
+/**
+ * @package admin
+ * @copyright Copyright 2020-2020 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: DrByte 2019 May 26 Modified in v1.5.6b $
+ */
+
+ function page_has_help($help_page) {
+   global $PHP_SELF;
+
+   $page = basename($PHP_SELF, ".php");
+   // This list will be built out in the final version
+
+   $pagelist = array(
+
+     FILENAME_CONFIGURATION => "https://docs.zen-cart.com/user/admin_pages/configuration/",
+     FILENAME_CATEGORIES => "https://docs.zen-cart.com/user/admin_pages/catalog/categories/", 
+     FILENAME_CATEGORY_PRODUCT_LISTING => "https://docs.zen-cart.com/user/admin_pages/catalog/categories_products/", 
+     FILENAME_PRODUCT_TYPES => "https://docs.zen-cart.com/user/admin_pages/catalog/product_types/",  
+     FILENAME_PRODUCTS_PRICE_MANAGER => "https://docs.zen-cart.com/user/admin_pages/catalog/products_price_manager/",  
+     FILENAME_OPTIONS_NAME_MANAGER => "https://docs.zen-cart.com/user/admin_pages/catalog/option_name_manager/",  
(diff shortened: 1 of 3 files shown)


scottcwilson:
image

wiztechinc:
Good start. So does use the new docs for the tips.

drbyte:
Suggestions (related to code, not UI):
- instead of functions/general, I suggest functions/functions_help.php
- instead of database table, a PHP array would be sufficient (I prefer to avoid db tables that are strictly static data, unless they're useful in query joins with other relational data)

scottcwilson:
Roger that.

admin/includes/functions/functions_help.php
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * @package admin
+ * @copyright Copyright 2020-2020 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: DrByte 2019 May 26 Modified in v1.5.6b $
+ */
+
+ function page_has_help(&$help_page) {
+   global $db;
```
drbyte, line 11: $db not needed.

admin/includes/functions/functions_help.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/**
+ * @package admin
+ * @copyright Copyright 2020-2020 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: DrByte 2019 May 26 Modified in v1.5.6b $
+ */
+
+ function page_has_help(&$help_page) {
```
drbyte, line 10: Let's avoid the `&` pass-by-reference here.

admin/includes/functions/functions_help.php
```diff
@@ -0,0 +1,92 @@
...
+     FILENAME_MEDIA_MANAGER => "https://docs.zen-cart.com/user/admin_pages/extras/media_manager/",  
+     FILENAME_MEDIA_TYPES => "https://docs.zen-cart.com/user/admin_pages/extras/media_types/",  
+     FILENAME_STATS_SALES_REPORT_GRAPHS => "https://docs.zen-cart.com/user/admin_pages/reports/graphical_sales_report/",  
+   ); 
+
+   if (isset($pagelist[$page])) {
+     $help_page = $pagelist[$page]; 
+     return true; 
+   }
+   return false;
```
drbyte, line 91: Returning the URL or `false` will be enough for detecting valid response.

admin/includes/header_navigation.php
```diff
@@ -43,3 +43,12 @@
     </ul>
   </div><!-- /.navbar-collapse -->
 </nav>
+<?php if (page_has_help($help_page)) { ?>
```
drbyte, line 46:
```diff
-<?php if (page_has_help($help_page)) { ?>
+<?php if ($url = page_has_help($help_page)) { ?>
```
and then `echo $url` later

admin/includes/functions/functions_help.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/**
+ * @package admin
+ * @copyright Copyright 2020-2020 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
```
drbyte, line 5: None of the logic here came from or had inspiration from osc

drbyte:
Thanks for pulling all this together @scottcwilson

PR #3198, opened 2020-04-09
Resolves oversanitization in admin, fixes #2794

### LINKED ISSUE #2794: Oversanitization of configuration entries
If you go to admin-configuration-layout settings, and set Bread Crumbs Navigation Separator to >> it converts these symbols to &gt;&gt;.  This actually works the first time, but if you try to edit it again, it converts the & to &amp; and everything falls apart.  Verified bad in 1.5.6c.

scottcwilson:


scottcwilson:
Maybe once this gets merged, @lat9 could throw a notifier in so folks could add their own fields (such as `CHECKOUT_ONE_OTTOTAL_SELECTOR` for example.

lat9:
> 
> Maybe once this gets merged, @lat9 could throw a notifier in so folks could add their own fields (such as `CHECKOUT_ONE_OTTOTAL_SELECTOR` for example.

Thanks for thinking of me, @scottcwilson;  I've got this PR tagged and will add that notification once it's accepted.

admin/includes/init_includes/init_sanitize.php
```diff
@@ -249,11 +274,17 @@
...
+    $checks = $db->Execute("SELECT configuration_key, val_function FROM " . TABLE_CONFIGURATION . " WHERE configuration_id = " . (int)$cID);
 
-if (!(!$checks->EOF && $checks->fields['val_function'] != NULL)) {
-    $group = array_merge($group, array('configuration_value'));
+    if (!$checks->EOF && $checks->fields['val_function'] != NULL) {
+      $skip_group = array('configuration_value' => array('sanitizerType' => 'NULL_ACTION', 'method' => 'post'));
+      $sanitizer->addComplexSanitization($skip_group);
+    } else if (!$checks->EOF && in_array($checks->fields['configuration_key'], $configs_with_special_characters)) {
+      $skip_group = array('configuration_value' => array('sanitizerType' => 'NULL_ACTION', 'method' => 'post'));
+      $sanitizer->addComplexSanitization($skip_group);
```
drbyte, line 284: Since the payload of both the `if` and `elseif` here are identical, does it make sense to combine them?

scottcwilson: OK.

scottcwilson:
I am going to close this PR so a record is preserved of a good fix for this issue in 1.5.6. (This issue only affects 1.5.6. and 1.5.7-alpha). 

The final 1.5.7 fix will look different, in part because of the hard work of @mc12345678  who has already properly validated emails in 1.5.7, so emails are not needed on any skip list.  

Stay tuned ... same Zen time ... same Zen channel.

admin/includes/init_includes/init_sanitize.php
```diff
@@ -238,8 +238,33 @@
...
+   'CATEGORIES_SEPARATOR_SUBS',
+   'CATEGORIES_COUNT_PREFIX',
+   'CATEGORIES_SUBCATEGORIES_INDENT',
+   'EZPAGES_SEPARATOR_HEADER',
+   'EZPAGES_SEPARATOR_FOOTER',
+   'CURRENCIES_TRANSLATIONS',
+   'STORE_OWNER_EMAIL_ADDRESS',
+   'EMAIL_FROM',
+   'SEND_EXTRA_ORDER_EMAILS_TO',
+   'SEND_EXTRA_CREATE_ACCOUNT_EMAILS_TO_STATUS',
```
drbyte, line 254: As mentioned elsewhere, this _TO_STATUS entry should not be in this list.
Just pointing it out here in case someone tries to use this code. NOTE: This code is NOT what Zen Cart has chosen to use to address this issue.

PR #3207, opened 2020-04-10
Extend order message to handle admin order updates

scottcwilson:
Extends #1670 so that a message may be added to admin order update emails as well.

### DIFF
--- a/includes/functions/functions_osh_update.php
+++ b/includes/functions/functions_osh_update.php
@@ -122,7 +122,15 @@ function zen_update_orders_history($orders_id, $message = '', $updated_by = null
                     strip_tags($email_message) .
                     $status_text . $status_value_text .
                     OSH_EMAIL_TEXT_STATUS_PLEASE_REPLY;
-           
+
+                // Add in store specific order message
+                $email_order_message = defined('EMAIL_ORDER_UPDATE_MESSAGE') ? constant('EMAIL_ORDER_UPDATE_MESSAGE') : '';
+                $GLOBALS['zco_notifier']->notify('ZEN_UPDATE_ORDERS_HISTORY_SET_ORDER_UPDATE_MESSAGE'); 
+                if (!empty($email_order_message)) {
+                 $email_text .= "\n\n" . $email_order_message . "\n\n";
+                }
+                $html_msg['EMAIL_ORDER_UPDATE_MESSAGE'] = $email_order_message;
+
                 $html_msg['EMAIL_SALUTATION'] = EMAIL_SALUTATION;
                 $html_msg['EMAIL_CUSTOMERS_NAME']    = $osh_info->fields['customers_name'];
                 $html_msg['EMAIL_TEXT_ORDER_NUMBER'] = OSH_EMAIL_TEXT_ORDER_NUMBER . ' ' . $orders_id;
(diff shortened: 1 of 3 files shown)


includes/functions/functions_osh_update.php
```diff
@@ -122,7 +122,15 @@ function zen_update_orders_history($orders_id, $message = '', $updated_by = null
                     strip_tags($email_message) .
                     $status_text . $status_value_text .
                     OSH_EMAIL_TEXT_STATUS_PLEASE_REPLY;
-           
+
+                // Add in store specific order message
+                $email_order_message = defined('EMAIL_ORDER_UPDATE_MESSAGE') ? constant('EMAIL_ORDER_UPDATE_MESSAGE') : '';
+                $GLOBALS['zco_notifier']->notify('NOTIFY_ORDER_SET_ORDER_UPDATE_MESSAGE');
```
drbyte, line 128: Shouldn't the `$email_order_message` variable be passed to the notifier here? Otherwise modifying it is more complicated than it needs to be.

scottcwilson: The other notifier also has no parms - see includes/classes/order.php line 1123.

scottcwilson: @lat9 can you weigh in here?  I have no opinion.

drbyte: Ya, the other one was inside a class though.

drbyte: Granted, I guess since it isn't a class property there then it's gotta be modified by reaching in as a global .... and reaching across things as globals is always a bad practice. While we already do that a bit in Zen Cart I'm trying to avoid proliferating that problem as much as possible. Think: never use globals unless there's no other way.

Perhaps we need to modify that notifier call in both places. A separate PR is fine for that.

scottcwilson: But wouldn't that break the code of people using the notifier now?

scottcwilson: the notifiers do have 2 different names, so if this one had a parm and the original one didn't that would be ok.

lat9: It's a brand-new notifier, so I'm guessing that its use at this point is _low_.   I agree with @drbyte that the `$email_order_message` should be passed as notification parameter 2 (so it can be modified) with the `$orders_id` as the first parameter to give an observer a bit of context.

I'll note, too, that all other notifications issued by `zen_update_orders_history` use events named `ZEN_UPDATE_ORDERS_HISTORY_*`.

> Perhaps we need to modify that notifier call in both places. A separate PR is fine for that.

_Both_ places?  Where's the other one?

scottcwilson: These two notifiers are sisters: NOTIFY_ORDER_SET_ORDER_UPDATE_MESSAGE, NOTIFY_ORDER_SET_ORDER_MESSAGE.

lat9: Could we change the 'name' of the one issued by `zen_update_orders_history` so it's more indicative of where it's 'coming from'?

scottcwilson: Sure!  What is your preferred name?

lat9: `ZEN_UPDATE_ORDERS_HISTORY_SET_ORDER_UPDATE_MESSAGE`

drbyte:
Missing parameters added in https://github.com/drbyte/zencart/commit/956448d87280888ec046e5dfe4af878b9366b1aa

PR #3187, opened 2020-04-08
Remove unused admin defines

scottcwilson:

PR #3209, opened 2020-04-10
Remove unused defines

scottcwilson:
Note that the scope of this PR is to remove *unused defines only* on the catalog side.  This is a follow on PR from the admin side PR #3187


includes/languages/english/extra_definitions/ask_a_question_defines.php
```diff
@@ -7,5 +7,3 @@
  * @TODO consider merge into core files 
  */
 define('ASK_A_QUESTION', 'Ask a question about this product');
```
drbyte, line 9: I guess this single remaining constant should be moved into `english.php` and this file deleted.

neekfenwick:
This seems to have removed `DATE_TIME_FORMAT` from `includes/languages/english.php`, it never got added back into `lang.english.php`, and now `zen_datetime_short()` tries to use `DATE_TIME_FORMAT` and crashes.  It was moved into `english.php.removed` which was then deleted by @zcwilt in a1d9b862.  Is this an Issue I should raise and PR for?

PR #3210, opened 2020-04-10
fixes countable error

proseLA:
if the transaction completes successfully, the `$result->getErrors()` on line 1100 returns something other than an array.  probably empty....  this will address that situation.

### DIFF
--- a/includes/modules/payment/square.php
+++ b/includes/modules/payment/square.php
@@ -1107,7 +1107,7 @@ class square extends base
             $messageStack->add_session(MODULE_PAYMENT_SQUARE_TEXT_COMM_ERROR, 'error');
         }
 
-        if (count($errors_object)) {
+        if (is_array($errors_object) && count($errors_object)) {
             $error = $this->parse_error_response($errors_object);
             $messageStack->add_session(MODULE_PAYMENT_SQUARE_TEXT_UPDATE_FAILED . ' [' . $error['detail'] . ']', 'error');
 


drbyte:
I'm contemplating a broader improvement.  Work in progress...

proseLA:
look forward to seeing it....

zcwilt:

includes/modules/payment/square.php
```diff
@@ -1107,7 +1107,7 @@ public function _doRefund($oID, $amount = null, $currency_code = null)
             $messageStack->add_session(MODULE_PAYMENT_SQUARE_TEXT_COMM_ERROR, 'error');
         }
 
-        if (count($errors_object)) {
+        if (is_array($errors_object) && count($errors_object)) {
```
mc12345678, line 1110: Can't tell from the discussion part if this is the current suggested revision or if it is being reworked.

Would recommend the use of `!empty($errors_object)` as the single replacement for count unless there is some reason to suspect that `{$api_instance-refundPayment($request_body)}->getErrors()` can provide a non-false result that is not an array?! I already tested the use of `is_array` where the variable is actually set to `null` and even with PHP 7.4.x it worked without issuing an error or warning, so don't see a real need to test `isset` in this use because the `try/catch` assigns a value in both cases.

drbyte: I've been thinking: surely there's a way to make us never have to check is_array on it. Feels dumb to be checking that repeatedly.

proseLA: > I've been thinking: surely there's a way to make us never have to check is_array on it. Feels dumb to be checking that repeatedly.

i concur!

drbyte:
Will merge for now. Still reviewing for more thorough solution.

PR #3212, opened 2020-04-11
Use mnemonics instead of magic numbers

scottcwilson:


admin/includes/extra_configures/stats_sales_report_graphs.php
```diff
@@ -9,3 +9,11 @@
...
 // adjust and uncomment if a different value is desired
 //define('SALES_REPORT_GRAPHS_FILTER_DEFAULT', '00000000110000000000');
+
+// Define constants
+define('HOURLY_VIEW', '1');
+define('DAILY_VIEW', '2');
+define('WEEKLY_VIEW', '3');
+define('MONTHLY_VIEW', '4');
+define('QUARTERLY_VIEW', '6');
+define('YEARLY_VIEW', '5');
```
drbyte, line 19: Shouldn't these constants be inside the class? There's no need for them to ever be customized.

https://www.php.net/manual/en/language.oop5.constants.php

scottcwilson: I could make them defines in the file `admin/includes/classes/stats_sales_report_graph.php` if you prefer.  Using `const` wouldn't be good since these are referenced in the constructor, so if you accidentally use `MONTHLY_VIEW` instead of `self::MONTHLY_VIEW`, your code *doesn't work*.

lat9: Why would using class constants _not_ be good?  It consolidates the functionality into that one file.

If you accidentally used `MONTHLY_VIEW` instead of `self::MONTHLY_VIEW`, you've made a coding error!

scottcwilson: A coding error that fails completely silently, with no clues.  You think, "Oh, I guess there was no data during that interval."  Unfortunately, in PHP, you can't make `const`s static the way you can in Java or other languages, and at that point, they may as well be variables since they are associated with a class instance.

scottcwilson: ... and you have to remember this, which makes your code really brittle.

drbyte: I guess moving logic outside the constructor would be a good idea.

That way if someone wants to use the class constant when passing in `$mode`, then they could just refer to it as `classname::constantname`.

scottcwilson: How about we just merge this PR, which is a genuine readability improvement that removes the magic number use, and make the other improvements in a future PR?

drbyte: I've switched it to class constants, and fixed a few other spots that needed the numbers changed over to match.

scottcwilson: If you use `const` outside the class, you don't have to type that ugly `self::` every time you reference a constant.  Seems like such a hideous syntax to me.  But in any event, it's easier to read MONTHLY_VIEW than to have to remember what '4' means so thanks for getting this merged.

PR #3216, opened 2020-04-12
Admin:header_navigation: php notice for $help_page undefined

torvista:
fixes php notice for $help_page undefined

### DIFF
--- a/admin/includes/header_navigation.php
+++ b/admin/includes/header_navigation.php
@@ -43,7 +43,7 @@ $menuTitles = zen_get_menu_titles();
     </ul>
   </div><!-- /.navbar-collapse -->
 </nav>
-<?php if ($url = page_has_help($help_page)) { ?>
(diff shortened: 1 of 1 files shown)


scottcwilson:
Thanks - sorry about that.

scottcwilson:
The better fix though is to remove $help_page from admin/includes/header_navigation.php and includes/functions/functions_help.php.  The parameter is no longer needed.

drbyte:
If it's not needed, then let's not merge this PR.

scottcwilson:
Sorry I didn't explain that very well.  The parameter is not needed in both places and should be removed.  It's an artifact of when it was passed to be updated; now the  link is returned.

torvista:
@scott So, you'll PR the changes to stop this error...

scottcwilson:
Just did!  Thanks for reporting.

PR #3219, opened 2020-04-12
Improve layout and appearance of help button

scottcwilson:


admin/includes/stylesheet.css
```diff
@@ -835,6 +835,16 @@ input.faint {
...
 /* End of css buttons */
 
+/* Additions for help */
+.navbar {
+  margin-bottom: 8px;
+}
+.btn-help {
+  margin-right: 4px;
+  margin-bottom: 8px;
+  background-color: #5ce400;
```
drbyte, line 845: I think this bright green is too strong. Something that matches existing colors is better IMO. Perhaps `#E7E6E0`

admin/includes/stylesheet.css
```diff
@@ -835,6 +835,16 @@ input.faint {
 
 /* End of css buttons */
 
+/* Additions for help */
+.navbar {
+  margin-bottom: 8px;
```
drbyte, line 840: The addition of this margin has pointed out that adding the help button has forced a vertical loss of space for every page on which the button appears. 
Perhaps the HTML where it is rendered can be reworked to prevent forcing all the page content to be shoved down just so the button can be shown?

admin/includes/stylesheet.css
scottcwilson: OK.

admin/includes/stylesheet.css
scottcwilson: You have to push things down otherwise the button will overwrite on page content.

scottcwilson: If this is a big concern we could make it a link next to support, but I think making it more visible is better.   Let's try it with the blue color from the details button on the orders screen so it stands out from the gray in the navbar above it.

drbyte:
I do think this still needs work, both for placement, size, and color.
I'll merge it because it's a slight improvement.

PR #3220, opened 2020-04-12
Blacken docs text for old codgers like me

scottcwilson:
The gray color that is used currently is almost unreadable.

### DIFF
--- a/docs/zen.css
+++ b/docs/zen.css
@@ -4,7 +4,7 @@
 /*! Zen Cart Overrides */
 body {
   background: #fff;
-  color: #777;
+  color: #303030; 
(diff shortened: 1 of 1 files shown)


docs/zen.css
```diff
@@ -4,7 +4,6 @@
 /*! Zen Cart Overrides */
 body {
   background: #fff;
-  color: #777;
```
drbyte, line 7: I agree that the 777 grey is weak contrast.

And yet flipping the grey all the way to black negatively affects some of the aesthetics.

How about a slighter darkening by changing `#777` to `#606060`?

scottcwilson: #303030 looks better.  Old people need strong contrast.

PR #3224, opened 2020-04-12
Admin: admin_account matched to users

torvista:
Made the files match to easily compare.
To avoid php notice for duplicated constant,
I rationalised the constants between users and admin_account and  included users.php constants into admin_account. Admin account only has its own HEADING_TITLE, so I put a is_defined clause on the user version.
Othewise just cosmetic changes so it looks like admin users.

### DIFF
--- a/admin/admin_account.php
+++ b/admin/admin_account.php
@@ -104,10 +104,6 @@ $userDetails = $userList[0];
         echo zen_draw_form('users', FILENAME_ADMIN_ACCOUNT);
         if (isset($formAction)) {
             echo zen_draw_hidden_field('action', $formAction);
-        }
-
-    if ($action == 'edit' || $action == 'password') {
-        echo zen_draw_hidden_field('user', $user);
     } ?>
     <table class="table">
         <thead>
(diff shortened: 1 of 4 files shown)


admin/admin_account.php
```diff
@@ -107,9 +107,15 @@ function init()
...
+    <?php
+        echo zen_draw_form('users', FILENAME_ADMIN_ACCOUNT);
+        if (isset($formAction)) {
+            echo zen_draw_hidden_field('action', $formAction);
+        }
 
-<form action="<?php echo zen_href_link(FILENAME_ADMIN_ACCOUNT) ?>" method="post">
-<?php if (isset($formAction)) echo zen_draw_hidden_field('action',$formAction) . zen_draw_hidden_field('securityToken', $_SESSION['securityToken']); ?>
+    if ($action == 'edit' || $action == 'password') {
+        echo zen_draw_hidden_field('user', $user);
```
drbyte, line 110: Why was this unused field added?

torvista: Sorry.

drbyte:
Note that I fixed some spacing issues and changed tabs to spaces for correct formatting.

PR #3227, opened 2020-04-12
Remove payment module related unused defines

scottcwilson:


drbyte:
Thanks, but most of these are there as alternatives or for consistency between modules.

scottcwilson:
Fair enough - that's why I separated it.

PR #3236, opened 2020-04-13
Properly sanitize configs which could contain meta characters

scottcwilson:
Fixes ##2794.  Note that this fix does not include email addresses, which are fixed with val_functions in 1.5.7. 

Please note that this fix is targeted to 1.5.7 only because it assumes the val_functions will fix the email address sanitization.

### DIFF
--- a/admin/includes/init_includes/init_sanitize.php
+++ b/admin/includes/init_includes/init_sanitize.php
@@ -204,12 +204,14 @@ $group = array(
 );
 $sanitizer->addSimpleSanitization('ALPHANUM_DASH_UNDERSCORE', $group);
 
-$group = array('pages_title', 'page_params', 'music_genre_name', 'artists_name', 'record_company_name', 'countries_name', 'name', 'type_name', 'manufacturers_name',
-               'title', 'coupon_name', 'coupon_copy_to_dup_name', 'banners_title', 'coupon_code', 'coupon_delete_duplicate_code', 'coupon_type',
-               'group_name', 'geo_zone_name', 'geo_zone_description',
-               'tax_class_description', 'tax_class_title', 'tax_description', 'entry_company', 'customers_firstname',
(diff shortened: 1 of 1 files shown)


admin/includes/init_includes/init_sanitize.php
```diff
@@ -237,9 +237,27 @@
...
+   'STORE_NAME_ADDRESS',
+   'PRODUCT_LIST_SORT_ORDER_ASCENDING',
+   'PRODUCT_LIST_SORT_ORDER_DESCENDING',
+); 
 
-$checks = new queryFactoryResult($db->link);
-// if current page is configuration, the configuration option does not have a val_function and saving the setting, then sanitize configuration_value to array.
+// if current page is configuration, handle separately
 if (defined('FILENAME_CONFIGURATION')
       && $_SERVER['SCRIPT_NAME'] == DIR_WS_ADMIN . (!strstr(FILENAME_CONFIGURATION, '.php') ? FILENAME_CONFIGURATION . '.php' : FILENAME_CONFIGURATION)
```
drbyte, line 262: Can this be simplified so that we only check for `!empty($_GET['cID'])` and pass page=configuration into the $group array?

ref comment at https://github.com/zencart/zencart/commit/ee3e8eb53344ac402fd4af378db441a5d7f37c84#r38459481

scottcwilson: Can we merge this now and improve it later?  I'm worried about breaking something by making a quick change.  This fixes a real bug in the system and should be taken.

drbyte: Other fixes have been proposed for v156. 
Are you planning to tell people "this" is the "new" fix for v156? 
If so, I'd like it to be more atomic than "add these three, in this specific order"

scottcwilson: This is not a fix for 1.5.6 - 1.5.6 needs email addresses to be handled as well.  Sorry, I could have sworn I said that this fix was targeted to 1.5.7 only in the description.  Fixing now.

PR #3240, opened 2020-04-13
Turn off autocomplete for this field

scottcwilson:
Having autocomplete enabled can block the widget, as shown in this image: 

datepicker


Zen4All:
Thought I had already pushed that :blush: 
thx

drbyte:
Should this be applied to other places the datepicker is used? 
And the other collect_info files?

scottcwilson:
Re: other datepickers: it's a bit tricky because it's not required for spiffycal, which we still use. 

I'd like to get rid of the other collect_info files so we don't have to do this kind of maintenance. 
What if 1.5.8 was the simplification release?

torvista:
Can spiffycal not be culled now?

scottcwilson:
1.5.8!

PR #3247, opened 2020-04-13
Add product-details popover for orders

drbyte:
Modernized implementation of order-product-Info-At-A-Glance plugin

Screen Shot 2020-04-13 at 10 03 38 PM
Screen Shot 2020-04-13 at 10 04 03 PM
Screen Shot 2020-04-13 at 10 04 41 PM


drbyte:
Screen Shot 2020-04-13 at 10 03 38 PM
Screen Shot 2020-04-13 at 10 04 03 PM
Screen Shot 2020-04-13 at 10 04 41 PM

drbyte:
@lankeeyankee is this what you had in mind when you mentioned the Info At A Glance plugin?

lankeeyankee:
Yes, this is what I was suggesting, it's extremely handy and we all use it every day.

torvista:
1) I find that now I cannot select orders...clicking on any row does not change the infobox: I have to open the order.
2) The popup looks good, but in my opinion wastes space in its own column.
I implented it as a title tag on the price so I don't need to even click on it:

Clipboard01

scottcwilson:
Confirmed finding #1.  @torvista will you write a defect report?

scottcwilson:
Issue #3304 created.

torvista:
in Admin/orders.php
I note that the ubiquitous (for me) HTML validator spits out some errors,  for the data-content: "missing quote mark for attribute value" for specific orders.
There isn't a quote problem, but it IS related to an opening < tag...sometimes, as br and hr are used in this attribute. Removing or encoding them stops the errors.
I could not find a pattern to explain when it occurs although it is on orders or two or more products, but all of those.
I suspect it is related to the parsing by the validator. 
To remove those errors, and so see any warnings (which are not shown when errors exist), adding true to the call works.
Actual
`data-content="<?php echo zen_output_string($product_details, array('"' => '&quot;', "'" => '&#39;', '<br />' => '<br>')); ?>"`

Modified
`data-content="<?php echo zen_output_string($product_details, array('"' => '&quot;', "'" => '&#39;', '<br />' => '<br>'), true); ?>"`

I note that removing the breaks altogether has no visual effect as there are hr in between the products which makes a line break:
`data-content="<?php echo zen_output_string($product_details, array('"' => '&quot;', "'" => '&#39;', '<br />' => ''), true); ?>"`

I'm not suggesting a change, it's just for information.

torvista:
I should have illustrated this:

Clipboard03

PR #3249, opened 2020-04-14
Possible fix for htaccess test in zc_install

### LINKED ISSUE #3248: htaccess check issues
As discussed in https://www.zen-cart.com/showthread.php?225304-zc_install-Support-for-quot-htaccess-quot-files-is-not-enabled&p=1367659#post1367659

zc_install/includes/classes/class.systemChecker.php line 347

does not consider that some systems have .htacess files that force https so it will fail.

We should check $_SERVER (or other data) to see if http or https should be used in the check.

scottcwilson:
Fixes #3248 (not yet tested live).

### DIFF
--- a/zc_install/includes/classes/class.systemChecker.php
+++ b/zc_install/includes/classes/class.systemChecker.php
@@ -344,7 +344,14 @@ class systemChecker
   public function checkHtaccessSupport($parameters)
   {
     $testPath = $_SERVER['HTTP_HOST'] . $_SERVER['SCRIPT_NAME'];
-    $testPath = 'http://' . substr($testPath, 0, strpos($testPath, '/zc_install')) . '/includes/filenames.php';
+    $proto = 'http://'; 
+
+    if (isset($_SERVER['REQUEST_SCHEME']) && $_SERVER['REQUEST_SCHEME'] == 'https') { 
+      $proto = "https://"; 
+    } else if (isset($_SERVER['SCRIPT_URI']) && strncmp($_SERVER['SCRIPT_URI'], 'https', 5) == 0) { 
+      $proto = "https://";
+    }
+    $testPath = $proto . substr($testPath, 0, strpos($testPath, '/zc_install')) . '/includes/filenames.php';
     if (function_exists('curl_init'))
     {
       $resultCurl = self::curlGetUrl($testPath);
--- a/zc_install/includes/classes/class.systemChecker.php
+++ b/zc_install/includes/classes/class.systemChecker.php
@@ -341,42 +341,6 @@ class systemChecker
     return $result;
   }
 
-  public function checkHtaccessSupport($parameters)
-  {
-    $testPath = $_SERVER['HTTP_HOST'] . $_SERVER['SCRIPT_NAME'];
-    $proto = 'http://'; 
-
-    if (isset($_SERVER['REQUEST_SCHEME']) && $_SERVER['REQUEST_SCHEME'] == 'https') { 
-      $proto = "https://"; 
-    } else if (isset($_SERVER['SCRIPT_URI']) && strncmp($_SERVER['SCRIPT_URI'], 'https', 5) == 0) { 
-      $proto = "https://";
-    }
-    $testPath = $proto . substr($testPath, 0, strpos($testPath, '/zc_install')) . '/includes/filenames.php';
-    if (function_exists('curl_init'))
-    {
-      $resultCurl = self::curlGetUrl($testPath);
-      if (isset($resultCurl['http_code']) && $resultCurl['http_code'] == '403')
-      {
-        $result = TRUE;
-      } else
-      { // test again with redirects enabled
-	      $resultCurl = self::curlGetUrl($testPath, true);
-	      if (isset($resultCurl['http_code']) && $resultCurl['http_code'] == '403')
-	      {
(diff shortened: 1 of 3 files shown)


zc_install/includes/classes/class.systemChecker.php
```diff
@@ -344,7 +344,14 @@ public function checkPhpVersion($parameters)
   public function checkHtaccessSupport($parameters)
   {
     $testPath = $_SERVER['HTTP_HOST'] . $_SERVER['SCRIPT_NAME'];
-    $testPath = 'http://' . substr($testPath, 0, strpos($testPath, '/zc_install')) . '/includes/filenames.php';
+    $proto = 'http://'; 
+
+    if (isset($_SERVER['REQUEST_SCHEME']) && $_SERVER['REQUEST_SCHEME'] == 'https') { 
+      $proto = "https://"; 
+    } else if (isset($_SERVER['SCRIPT_URI']) && strncmp($_SERVER['SCRIPT_URI'], 'https', 5) == 0) {
```
drbyte, line 351: Let's take a different approach:
- since `$request_type` is already set to `SSL` when all the usual detections pass, let's use that instead of all this new code
- use that to determine whether we test https first
- then if that first test fails, fallback and re-test with the other

Reason: If the server can't be trusted to do its own 301 redirect to a proper 403 response (and not a 404 because the 403 isn't fully configured), we probably can't trust it a binary either/or situation either, so a fallback is the only option. 

Or maybe we just remove this check altogether and stop taking responsibility for exposing the host's misconfiguration.

scottcwilson: YAY!  I like door number three.

mc12345678: What impact will this "removal" have on PCI of the product (ZC) overall? Its one thing for a server to be misconfigured and prone to its own issues and another for performing the check and identifying the strengths/weaknesses of performing the check with additional information about how to resolve...

drbyte: > > Or maybe we just remove this check altogether and stop taking responsibility for exposing the host's misconfiguration.

> What impact will this "removal" have on PCI of the product (ZC) overall? Its one thing for a server to be misconfigured and prone to its own issues and another for performing the check and identifying the strengths/weaknesses of performing the check with additional information about how to resolve...

None. Zen Cart isn't responsible for the server's actual configuration. That's up to the hosting company. Just like a properly functioning SSL certificate chain isn't something Zen Cart can fix or adjust or otherwise do anything about.

scottcwilson: Ditto.  Zen Cart also doesn't check for modified RPMs.  Out of scope.

mc12345678: Well, I had come up with this based on the earlier discussion/advice of @drbyte.

````php
  public function checkHtaccessSupport($parameters)
  {
    global $request_type;
    $testPath = $_SERVER['HTTP_HOST'] . $_SERVER['SCRIPT_NAME'];
    $testPath2 = substr($testPath, 0, strpos($testPath, '/zc_install')) . '/includes/filenames.php';
    $testPath_array = [1 => 'http://', 2 => 'https://'];
    $testPath = ($request_type != 'SSL' ? $testPath_array[1] : $testPath_array[2]) . $testPath2;
    
    if (!function_exists('curl_init')) {
      return TRUE;
    }

    $result = TRUE;
    $repeat = 2;
    while ($repeat) {
      $repeat--;
      $resultCurl = self::curlGetUrl($testPath);
      if (!(isset($resultCurl['http_code']) && $resultCurl['http_code'] == '403'))
      { // test again with redirects enabled
        $resultCurl = self::curlGetUrl($testPath, true);
        if (!(isset($resultCurl['http_code']) && $resultCurl['http_code'] == '403'))
        {
          if ($request_type == 'SSL' && $repeat) {
            $testPath = $testPath_array[$repeat] . $testPath2;
            continue;
          }
          $result = FALSE;
        }
      }
      break;
    }

    return $result;
  }
````

drbyte: Hmmm ... ya, maybe something like that is fine.
I'm inclined to drop the 2 test URLs into an array, with https first, then http, then early-return as soon as one passes. (still doing the redirect fallback).  Then can drop the $repeat var and the $result boolean

scottcwilson: This would be a bunch more testing.  Do you have the time and resources (webserver that can be switched in and out of forced https) to do that?

drbyte: The test would need to be done by those on the forum complaining that it's "broken" currently. Granted, I still think the real issue is a cPanel problem. I can't test cPanel though.

drbyte: It's never been a "hard fail". This check only sets up a warning/alert for consideration.

If an update prevents "false positives", great. But if the server has a misconfiguration we can't detect that without making this code overly complicated. And we can never "fix" it because it's out of scope.

wiztechinc:
I do have cpanel. If you can easily give me a couple of scenarios, I can give it a try. But both servers I use do not force https, only default http,  and I have auto cpanel certs. I'm uncertain if then any testing I do can duplicate the issue?

Issue #2973, opened 2019-12-27
When to release?


torvista:
Is there a milestone that will determine the next release?
a) more people on the forum are starting to report notices and warnings that are aleady fixed in 157 (so time/effort wasted there)
b) an idea of the schedule would help to allocate/focus my meagre resources in completing multiple almost-finished PRs for inclusion. I know I should do them all immediately, but it's just not possible in the real world.

scottcwilson:
We have a December release tradition, don't we? :)

torvista:
NOOO!:scream: 

How about when PRs are approaching zero!!

drbyte:
Much as I'd like to put a release out before end of 2019, it's not going to happen this time.

lat9:
> 
> 
> NOOO!😱
> 
> How about when PRs are approaching zero!!
... and the issues' count.

scottcwilson:
> ... and the issues' count.

Only for critical must-fix issues.  Low priority issues and feature requests will always be there.   Right now there is tremendous value in 1.5.7 that is not available to any user.

judygunderson:
> 
> 
> > ... and the issues' count.
> 
> Only for critical must-fix issues. Low priority issues and feature requests will always be there. Right now there is tremendous value in 1.5.7 that is not available to any user.

I agree Scott!

torvista:
Just running through all the pages with "Report All Errors" on still generates errrors, but I believe it is now required, to set a "clean" baseline of no php notices/warnings from core code on php7.3.
I have achieved that with my 156 but it has taken months of my (non) spare time. chipping away...

lat9:
The last thing that we want to do with zc157 is release it too soon.  Other than the squishing of some of the PHP 7.3/7.4 interoperability issues, I'm not seeing any _major_ functional/feature changes that would warrant a zc157 vs. zc156d.

judygunderson:
> 
> 
> The last thing that we want to do with zc157 is release it too soon. Other than the squishing of some of the PHP 7.3/7.4 interoperability issues, I'm not seeing any _major_ functional/feature changes that would warrant a zc157 vs. zc156d.

Definitely agree, but I am like a kid at a candy store and my Mama just saying no no no.  giggle

What I don't want to see is a 1.5.6d come out and then 2 months later 1.5.7 because that ticks my customers off! :(

wiztechinc:
Please just no.

drbyte:
[...]

PR #3251, opened 2020-04-14
Bug fix to prior PR: wrong value_id value in options_values_manager.php

Zen4All:
As reported in #2973 This fixes a bug in the options_values_manager.php

This happen on creating a new value when selecting an option from the filter dropdown that contains no values first.

The problem is that the ```$value_id``` sits inside a ```foreach``` that is never fired on empty options. 
Moving the query that sets ```$value_id``` outside the ```foreach``` fixed the error on pressing insert
@torvista : This is the fix we talked about through email

### DIFF
--- a/admin/options_values_manager.php
+++ b/admin/options_values_manager.php
@@ -812,11 +812,11 @@ if (zen_not_null($action)) {
                     <?php } ?>
                     <?php
                   }
-                  $max_values_id_values = $db->Execute("SELECT MAX(products_options_values_id) + 1 AS next_id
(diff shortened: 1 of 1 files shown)


torvista:
Are you not having a go at the unbalanced tags?

Clipboard01

Since you touched it last, it must be your job..:laughing:

Zen4All:
> Are you not having a go at the unbalanced tags?
> 
> Clipboard01
> 
> Since you touched it last, it must be your job..

I think I just fixed most of that in my two latest PR'S

PR #3260, opened 2020-04-14
remove manual from main page, tighten up shopping cart page

scottcwilson:


scottcwilson:
Current image is for a book which is not available, using a link that doesn't work.

zcwilt:
Think this may also be better done by changing the graphic to something served from our banner server.

PR #3264, opened 2020-04-15
Change from document.write of JQuery to forced CDN load

dbltoe:
There are so many versions of JQuery flloating around and this one causes...
***************************
index.php?main_page=product_info&cPath=81_1_3&products_id=1:22 A parser-blocking, cross site (i.e. different eTLD+1) script, https://code.jquery.com/jquery-3.3.1.min.js, is invoked via document.write. The network request for this script MAY be blocked by the browser in this or a future page load due to poor network connectivity. If blocked in this page load, it will be confirmed in a subsequent console message. See https://www.chromestatus.com/feature/5718547946799104 for more details.
*************************
removing  original line 140 solves this problem for our sites.

### DIFF
--- a/includes/templates/responsive_classic/common/html_header.php
+++ b/includes/templates/responsive_classic/common/html_header.php
@@ -136,7 +136,7 @@ if (count($lng->catalog_languages) > 1) {
 
 /** CDN for jQuery core **/
 ?>
-
+<script src="https://code.jquery.com/jquery-3.5.0.min.js" integrity="sha256-xNzN2a4ltkB44Mc/Jz3pT4iU1cmeR0FkXs4pru/JxaQ=" crossorigin="anonymous"></script>
 <?php if (file_exists(DIR_WS_TEMPLATE . "jscript/jquery.min.js")) { ?>
 <script type="text/javascript">window.jQuery || document.write(unescape('%3Cscript type="text/javascript" src="<?php echo $template->get_template_dir('.js',DIR_WS_TEMPLATE, $current_page_base,'jscript'); ?>/jquery.min.js"%3E%3C/script%3E'));</script>
 <?php } ?>
--- a/includes/templates/responsive_classic/common/html_header.php
+++ b/includes/templates/responsive_classic/common/html_header.php
@@ -137,7 +137,6 @@ if (count($lng->catalog_languages) > 1) {
 /** CDN for jQuery core **/
 ?>
 
-<script type="text/javascript">window.jQuery || document.write(unescape('%3Cscript type="text/javascript" src="https://code.jquery.com/jquery-3.4.1.min.js" integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo=" crossorigin="anonymous"%3E%3C/script%3E'));</script>
 <?php if (file_exists(DIR_WS_TEMPLATE . "jscript/jquery.min.js")) { ?>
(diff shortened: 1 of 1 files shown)


scottcwilson:
Many of those versions are obsolete and have security issues.  This message is a warning, not a problem.

drbyte:
I'm not willing to remove it "just so a browser warning goes away" when the browser is ignorant about "why" it's there.

I *am* willing to remove it if we're certain that we can trust that the templates will actually load it properly all the time. 
It's there as a safeguard for those who delete the local file or otherwise break their template.
Can we finally trust the users and template authors nowadays?

scottcwilson:
> Can we finally trust the users and template authors nowadays?
NO.  Every day I'm seeing commercial templates (that people are paying for!) load obsolete versions of jQuery.  Current ZC approach of loading current jQuery is totally appropriate.

drbyte:
Just to be clear: dbltoe is NOT saying to stop loading jQuery from a CDN

He's pointing out that Chrome doesn't like it when we use JAVASCRIPT to do the CDN load.

What I'm saying is that the use of javascript is merely a fallback.

Scott, are you talking about the same thing? Specifically the document.write call?

scottcwilson:
I may have misunderstood - I thought he was saying it was sufficient to load whatever jQuery was supplied by the template.  I am neutral on the mechanism, as long as the core loads the latest jQuery early.

dbltoe:
https://developers.google.com/web/tools/lighthouse/audits/document-write was what got me started.
With all the time spent loading images that are too large, do we need to add to the problem with a slow document.write?

drbyte:
> <script>window.jQuery || document.write(unescape('%3Cscript type="text/javascript" src="https://code.jquery.com/jquery-3.4.1.min.js"%3E%3C/script%3E'));</script>


Since PHP can't tell whether the browser has already had jQuery loaded, we have to give it javascript to execute to do the detecting:

Here's how you read that code:

1. `window.jQuery`
This means "check to see if window.jQuery is an object. Is it declared? If yes, ignore and go to the next command. In this case there's no next command. Done.
The reason for this is to prevent loading multiple versions of jQuery if it's already been loaded somewhere before this part of the template HTML gets processed.
(Granted, upon further inspection, I see that there's almost no entry point before this line that would have already loaded jQuery. See proposal later below.)

2. `||` 
If the previous statement failed (ie: no jQuery present), then continue along the row ...

3. `document.write(echo a command that injects a CDN call to load jQuery)`
We only get here if jQuery wasn't found already loaded by the template. And if we get here, then we tell the browser to inject a call to load jQuery from the CDN.
Sure, Chrome won't actually do it because of its own attempt to "protect you from yourself".


So, I'll change my question to a proposal:
Should we just assume nobody's loading another jQuery version before this line, and change this document.write to a full-on CDN load of jQuery?

... and then let the subsequent checks on the next lines (which try to load a LOCAL jQuery file if jQuery isn't loaded in the browser successfully yet) do their thing?  Or will Chrome complain about those too? 

I'm thinking we can just switch it to a normal CDN load. Instead of fully removing the document.write.

barco57:
html5-boilerplate still uses a document.write if the CDN fails. Chrome currently is the only browser that whines about this as a warning and it mostly causes a problem for mobile users using chrome on 2g with slow connections, then chrome may refuse to load it.
`  <script src="https://code.jquery.com/jquery-3.5.0.min.js" integrity="sha256-xNzN2a4ltkB44Mc/Jz3pT4iU1cmeR0FkXs4pru/JxaQ=" crossorigin="anonymous"></script>
  <script>window.jQuery || document.write('<script src="js/vendor/jquery-3.5.0.min.js"><\/script>')</script>`

torvista:
>Should we just assume nobody's loading another jQuery version before this line

I'll take that to mean "should we assume that someone modifying code is taking care and checking the results of their actions"..well yes, absolutely.

drbyte:
Okay, let's switch it to skip checking for existing, and load from CDN.
Then if that fails (ie: offline or whatever) it'll document.write to load the local one.

PR #3266, opened 2020-04-15
Do not display linked category text if no linked cats

### LINKED ISSUE #2839: Delete of category with no linked product displays linked product text
Found today when testing the deletion of a category that had no linked product that there is a warning presented regarding linked product and ensuring that the link(s) are properly modified; however, this doesn't make sense to display if the category doesn't have linked product.

Good to know the process, but this seems like the wrong point at which to discuss it.  Why?
- Always showing this text regardless the status of the contained product will reduce the importance of what the message is conveying.
- Displaying this message when there are no linked product would cause the concerned store owner to break away from the current action of deleting the category to go check if there is in fact a linked product and send them down some unnecessary path using time that could be better spent managing the store.
- Displaying this text for all conditions may also drive the concerned individual to seek out if there is a problem with the code/database when it is simply useful information provided for the wrong condition.

This was initially a part of: https://github.com/zencart/zencart/blame/f89a3b8269c7b513693ab5a0dce4469fd403cff9/admin/category_product_listing.php#L1041 in: https://github.com/zencart/zencart/pull/1604/files which became:  https://github.com/zencart/zencart/blame/v157/admin/category_product_listing.php#L1045

Perhaps a new function should be added to check if a category has linked product? There is already `zen_get_product_is_linked($products_id)` which either could be used within the function or it could have its own code to determine if the category has linked product.  Seems that if this concept is to be included that it not just be some stand alone code but something that could be otherwise used.

scottcwilson:
Fixes #2839


torvista:
I've reviewed this and added the preselect delete master categories and changed a couple of things...how do I modify/add to this PR  (git process) ?

scottcwilson:
Just make this same change in your PR so it's covered, and if it gets merged first, then we're done.  
If your PR does not get merged, this one can be merged to fix part of the issue and then we'll do something else for the rest.

drbyte:
@torvista @scottcwilson 
Steve you could also do a PR to Scott's branch in his own fork. Then if he merges your PR into that branch it will become part of the PR from that branch to ZC core.

drbyte:
This appears to be resolved. Is the origin issue still outstanding?

scottcwilson:
Steve fixed it in a separate PR along with other items.  I will close.

PR #3270, opened 2020-04-15
Default to https URLs in installer

### LINKED ISSUE #3269: installer: SSL all site?
I note that while the installer is run from https: and "Enable SSL for storefront" is marked by default, it defaults to the storefront as http, which is what is used in HTTP_CATALOG_SERVER
Clipboard01
Is that still a valid default for 2020 when all sites should be https now?
I am aware of the use of HTTP_CATALOG_SERVER/HTTPS_CATALOG_SERVER and how the subsequent discussions of the settings required for SSL are confusing to users, due to legacy code.
But maybe its better to set it all https in the installer.

scottcwilson:
Fixes #3269

Also preserves ability for people testing in a local environment to use http without having to edit URLs.

### DIFF
--- a/zc_install/includes/functions/general.php
+++ b/zc_install/includes/functions/general.php
@@ -191,14 +191,14 @@ function zen_get_select_options($optionList, $setDefault)
     global $request_type;
     if (isset($_POST['adminDir'])) $adminDir = zen_output_string_protected($_POST['adminDir']);
     $documentRoot = zen_get_document_root();
-    $url = ($request_type == 'SSL' ? 'https://' : 'http://') . $_SERVER['HTTP_HOST'] . str_replace('/zc_install/index.php', '', $_SERVER['SCRIPT_NAME']);
+    $proto = ($request_type == 'SSL' ? 'https://' : 'http://'); 
+    $url = $proto . $_SERVER['HTTP_HOST'] . str_replace('/zc_install/index.php', '', $_SERVER['SCRIPT_NAME']);
     $httpServer = zen_parse_url($url, 'host', true);
-    $adminServer = ($request_type == 'SSL') ? 'https://' : 'http://';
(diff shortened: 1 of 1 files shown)


drbyte:
> Also preserves ability for people testing in a local environment to use http without having to edit URLs.

Please explain further...

scottcwilson:
I run this procedure in MAMP all the time.  Without this change I have to edit the URLs to remove the "s" in "https" so all my URLs are http.  (It's a localhost test - no SSL)

drbyte:
You mean when you set ENABLE_SSL to false, you still have to change the HTTPS_SERVER etc URLs to be http?  
Why?

scottcwilson:
because otherwise it would write the https URLs to the configure file, which wouldn't work in my localhost environment.

drbyte:
That's not what I asked. And that's not how zc_install works.

When you uncheck the "Enable SSL?" box, it sets ENABLE_SSL to false for catalog and admin, and puts http:// where it should, and https:// where it should. It does NOT put https in wrong places in the configure file.

Then when you run catalog/admin, it honors the ENABLE_SSL to determine whether to use http or https.

Why is your environment not switching properly?

torvista:
You can have SLL on a local environment, with your own certificate. I would have thought it to be essential for realistic testing. If I can do it, it can't be so difficult!

scottcwilson:
> Why is your environment not switching properly?

let me do more testing.  this may have been fixed many moons ago.

scottcwilson:
LOL so I have been wasting my time removing the "s" from "https" on this screen forever. 
OK.  Do you want me to tweak this PR to fix Steve's issue?

drbyte:
> Do you want me to tweak this PR to fix Steve's issue?

I still don't understand Steve's issue.

PR #3271, opened 2020-04-15
Add status to zen_get_countries response array

### LINKED ISSUE #2991: admin: zen_get_countries includes disabled countries
The storefront version of the function properly returns a country-list including _only_ (under function parameter control) the countries currently enabled for a store.  The same cannot be said for the admin version of that function (present in `/admin/includes/functions/general.php`).

**Update:** Noting that the function is used by the admin's `customers.php`, `geo_zones.php`, `index_setup_wizard.php1 and `zones.php` (as well as a couple of plugins).

scottcwilson:
Fixes #2991, no extra cleanup .

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -656,13 +656,15 @@
       $countries_array[] = array('id' => '',
                                  'text' => $default);
     }
-    $countries = $db->Execute("SELECT countries_id, countries_name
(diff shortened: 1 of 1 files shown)


drbyte:
While I'm not certain this fully "fixes" the original issue, at least it provides data that could be used to display a status indicator.

PR #3272, opened 2020-04-15
Updating SQL syntax

scottcwilson:


admin/includes/functions/general.php
```diff
@@ -575,7 +575,7 @@ function zen_get_manufacturer_url($manufacturer_id, $language_id) {
     $manufacturer = $db->Execute("SELECT manufacturers_url
                                   FROM " . TABLE_MANUFACTURERS_INFO . "
                                   WHERE manufacturers_id = '" . (int)$manufacturer_id . "'
-                                  and languages_id = '" . (int)$language_id . "'");
+                                  AND languages_id = '" . (int)$language_id . "'");
```
mc12345678, line 578: Should remove single quotes from values cast as integers..

proseLA:
please don't.  other than what numbers said with regards to removal of single quotes for values cast as integers, what does this PR get us?

its a ROYAL pain for those of us that update existing carts and have to do a diff on all files.  i've seen these kind of changes in ZC updates previously and have hated them then, and i hate them now.

i get it; we want the code to be consistent.  but look at the styling docs.  we have modified PSR1 for existing code, and then PSR2 for new code (or something like that).  what is the point of this "proper" style for mysql?  if we want to enforce this style, why don't we go back and change all of the PSR1 code to PSR2?  in both cases, there is no functional difference.

these are pedantic changes that for new code, i am all in.  but not here.  

as we approach the release of a new version, i'm more curious if the header stamping line ending is more consistent with the line endings on the scripts.  from what i remember all of the code has LF (unix) line endings, but the stamper does CRLF (windows) line endings.  it would be nice to be consistent there.

as i have a tendency to be repetitive, i will attempt to refrain from commenting again on this PR.

best.

scottcwilson:
> its a ROYAL pain for those of us that update existing carts and have to do a diff on all files.

You might be making things harder than they need to be.  Start with a the NEW Zen Cart files, then apply only changes you made to your current store vs. the initial fileset for that version.  DO NOT compare your live store to the new version of Zen Cart;  this would be way more work and much more error prone.

torvista:
The horse is well out of the stable on this issue, but I thought that files should not be touched with what are basically formatting changes unless they are functionally modified.
I've see files changed where it's only the licence header and thought why?

drbyte:
I'm holding off on this one because it's so massive, and this file is frequently touched by plugins, and I'm undecided yet on final choice related to consolidating functions files to catalog-side in this release.
Hence the delay.

scottcwilson:
Sorry, I didn't realize the change would be so huge when I started it.  Yes, we do need to figure out the final set of changes to functions.  Are you tilting towards just importing the v2 change?

drbyte:
> Yes, we do need to figure out the final set of changes to functions. Are you tilting towards just importing the v2 change?

Undecided

scottcwilson:
Suggest third party intervention: https://en.wikipedia.org/wiki/Miss_Cleo

PR #3281, opened 2020-04-15
Reverts #2775 - wrongly specifying "missing_file" when saving attributes

### LINKED ISSUE #2927: Admin: attribute_controller, missing filename warning
after adding an attribute to a test product, this warning appears stuck in the middle.
Clipboard01
Seems to be using the empty field for the filename when adding the attribute.
I have no experience with this, but I would have thought this field should not even appear if the product is not virtual?

scottcwilson:
Fixes #2927

### DIFF
--- a/admin/attributes_controller.php
+++ b/admin/attributes_controller.php
@@ -337,12 +337,6 @@ if (zen_not_null($action)) {
             $products_attributes_maxdays = (int)$_POST['products_attributes_maxdays'];
             $products_attributes_maxcount = (int)$_POST['products_attributes_maxcount'];
 
-            // check to see if it's a file missing a name
-            if ($options_id === 1) {
-              if (empty($products_attributes_filename)) {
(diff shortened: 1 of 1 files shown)


drbyte:
I'm confused.

This was introduced in https://github.com/zencart/zencart/pull/2775
and now you're removing it.

What about the original problem you were trying to solve back then?

scottcwilson:
That change is not required in the current release. You can set a filename to a file that doesn't exist and everything is fine (red icon appears, which can be fixed later by uploading the file). The testing I did worked because in my config, option 1 was a downloadable file, but that's generally not true.

scottcwilson:
Yep, this was a blatant bug.

drbyte:
The commit description and PR title don't make any sense to me. Seems to be talking about a side-effect and not what the code change actually is.

So instead of trying to come up with another one, I'll just use the original PR title that's being reversed.

PR #3285, opened 2020-04-16
zc_install language update: fix regex and code for embedded defines

zcwilt:


zcwilt:
Note: requires embedded defines to be wrapped in %% characters

Issue #3278, opened 2020-04-15
Upgrade issue: drop final_price not firing


scottcwilson:
mysql_upgrade_zencart_156.sql contains the statement

ALTER TABLE customers_basket DROP final_price;

This is not firing, for some reason I can't figure out.   I fixed a couple of upgrade issues this morning and AFAIK this is the last one, but I can't seem to crack it.  

Repro steps:
- Install Zen Cart 1.5.5, build database through installer. 
- On Zen Cart 1.5.7, attempt to upgrade the 1.5.5 database you have just created.

After upgrade completes, you will see the table customers_basket still has the field final_price.

This can also be reproduced with an older database; I first saw this issue testing an upgrade from 1.3.9.

mc12345678:
Basically, because the default operation within parseraltertable is executed for that specific SQL statement, the presence of the field to be dropped causes the parser to ignore the line and therefore not get executed...  Note that this is to some extent buried within an existing `switch` statement intended to support other "verbiage" in the SQL statement, such as: `ALTER TABLE xxxx DROP COLUMN yyyy` Although that too would have failed in this case as well, because `$exists` would evaluate to true within the `COLUMN` section and therefore would cause the SQL line to be skipped...

Added a second commit to provide a single line evaluation instead of using a second line.  More complicated logic though...  Action may still be desired to be added just before the check to see if `$exists == true` to set `$this->ignoreLine = true`.

scottcwilson:
Thank you - this is a potentially huge breakthrough.  Once this is confirmed and merged I will retest with adding the DROP to the 1.5.7 upgrade - to be sure it gets done!

mc12345678:
As you may see in the commit history of that PR, I have provided at least two versions, there may be a third which would apply more globally, its just I am not so sure that `DROP` is applicable/appropriate for every "feature" of `ADD` and thus, either the code allows whatever might be entered in the SQL file or it is to be more intelligent and only provide the negation in certain cases... 

I remember seeing an issue with this in ZC 1.5.6; however, that version didn't execute a `DROP` and therefore wasn't researched further... I sort of thought it would get resolved "independently'; however, it popped up again and actually needs to be correct now.

scottcwilson:
Hang on!  It ain't done yet!  I need to modify the 157 update script to do this again because it would have been missed for folks who are on 156.

PR #3286, opened 2020-04-16
zc_install - Sync SQL updates (tested back to 1.5.1)

scottcwilson:
As @mc12345678 figured out, DROP wasn't working in prior versions of the upgrade software, so the Zen Cart 1.5.6 drop of the final_price field wasn't actually working. 

Now that that bug has been fixed (in #3282), we can correctly update the schema.  I have removed the DROP from 1.5.6 so upgraders in the future won't do it twice. 

Fixes (truly fixes!) #3278

### DIFF
--- a/zc_install/sql/install/mysql_zencart.sql
+++ b/zc_install/sql/install/mysql_zencart.sql
@@ -1718,7 +1718,7 @@ CREATE TABLE products_options (
   products_options_sort_order int(11) NOT NULL default '0',
   products_options_type int(5) NOT NULL default '0',
   products_options_length smallint(2) NOT NULL default '32',
-  products_options_comment varchar(255) default NULL,
+  products_options_comment varchar(256) default NULL,
   products_options_size smallint(2) NOT NULL default '32',
   products_options_images_per_row int(2) default '5',
   products_options_images_style int(1) default '0',
--- a/zc_install/sql/updates/mysql_upgrade_zencart_156.sql
+++ b/zc_install/sql/updates/mysql_upgrade_zencart_156.sql
@@ -176,7 +176,8 @@ VALUES ('categoriesProductListing', 'BOX_CATALOG_CATEGORIES_PRODUCTS', 'FILENAME
 
 DELETE FROM admin_pages WHERE page_key = 'linkpointReview';
 
-ALTER TABLE customers_basket DROP final_price;
+# This was moved to the 1.5.7 upgrade; DROP did not work in 1.5.6
+# ALTER TABLE customers_basket DROP final_price;
 
 ## add support for multi lingual ezpages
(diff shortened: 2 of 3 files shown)


mc12345678:
I was thinking that it would be acceptable to have it in both 1.5.6 and 1.5.7...  The software wouldn't have an issue with it and while the upgrade path (looking back seems I missed that the statement existed at the time(s) I had touched the installer) a fresh install would be different than an upgrade for those that installed 1.5.6 as compared to someone that upgraded to 1.5.6.  Further, those individuals that installed 1.5.6 and now upgrade to 1.5.7 are left to have an upgrade "action" that would report an "issue" because the field already didn't exist in 1.5.6....

Sum in total, unfortunately the 1.5.6 upgrade didn't address all SQL statements that were expected, 1.5.7 appears to capture all including correcting an issue generated in 1.5.6.  I'm not in a position to direct either way, just thinking that it should remain in the 1.5.6 upgrade path and to fix the omission in previous upgrades to 1.5.6 be included in the 1.5.7 upgrade...

scottcwilson:
Honestly, it's ok either way -  I removed it primarily as a "documentation" matter since it's a change that wasn't really part of 1.5.6 (because it didn't work).

scottcwilson:
Ugh the same issue exists with the ezpages DROPs from 1.5.6, and it would look bizarre to remove those.  I will restore as @mc12345678 suggested.  Please hold on this PR - more upgrade changes coming.

scottcwilson:
OK this should be the final upgrade related change; schemas look the same except for the authorize table, and I'll let others deal with that since they may be making changes there anyway.  

I commented out the failing drops in 1.5.6 with explanation and did them in 1.5.7.   Tested with mysqldump --no-data.

drbyte:
Is there a better title for this PR now?

scottcwilson:
Sorry I thought I had done that.

drbyte:
Do all of these "fail silently" (no crash) if they've already been done via other means?

scottcwilson:
Yes.  I tested this by doing a run *leaving in* the final_price DROP in 1.5.6 while also doing it in 1.5.7. But truly, that is a pretty unlikely scenario.  

Also tested upgrading a 1.5.5 database (which already has the extra fields in admin_activity_log)  - no issues.

drbyte:
I'm surprised about the ezpages stuff. I thought I saw people talking about it properly dropping those tables, because they were asking whether they could go "back" to look at things and it had been cleaned up already.

Is this mystical missing "drop" unique to MySQL and not a problem with MariaDB?

scottcwilson:
No idea.  I am just running upgrades and checking the resultant schema against a virgin 1.5.7 database - that's how I found all these issues.

scottcwilson:
It would be good to get this merged in soon so people can test with the final upgrade script.

PR #3288, opened 2020-04-16
admin: Sqlpatch - add clause for failed execute from upload, constants for hard-coded texts....

torvista:
prompted by finding a file upload was failing with no message due to a missing semicolon.
added text constants for hard-coded messages, a couple of html fixes...


admin/sqlpatch.php
```diff
@@ -735,22 +735,22 @@ function zen_create_exceptions_table() {
...
-          $messageStack->add($query_results['queries'] . ' statements processed.', 'success');
+          $messageStack->add(sprintf(TEXT_EXECUTE_SUCCESS, (int)$query_results['queries']), 'success');
         } else {
-          $messageStack->add('Failed: ' . $query_results['queries'], 'error');
+          $messageStack->add(sprintf(ERROR_EXECUTE_FAILED, (int)$query_results['queries']), 'error');
         }
         if (zen_not_null($query_results['errors'])) {
           foreach ($query_results['errors'] as $value) {
-            $messageStack->add('ERROR: ' . $value, 'error');
+            $messageStack->add(ICON_ERROR . ': ' . $value, 'error');
```
drbyte, line 744: I'm confused: what does this have to do with icons?

admin/sqlpatch.php
```diff
@@ -735,22 +735,22 @@ function zen_create_exceptions_table() {
...
         }
         if ($query_results['ignored'] != 0) {
-          $messageStack->add('Note: ' . $query_results['ignored'] . ' statements ignored. See "upgrade_exceptions" table for additional details.', 'caution');
+          $messageStack->add(sprintf(ERROR_EXECUTE_IGNORED, (int)$query_results['ignored']), 'caution');
         }
         if (zen_not_null($query_results['output'])) {
           foreach ($query_results['output'] as $value) {
             if (zen_not_null($value)) {
-              $messageStack->add('INFO: ' . $value, 'caution');
+              $messageStack->add(IMAGE_ICON_INFO . ': ' . $value, 'caution');
```
drbyte, line 753: Icon info?

torvista: It was the only constant that says INFO. Ugly I know.
Better to leave it as hard-coded, I think...there are uses of IMAGE_ICON_ ....scattered about since they are in english.php/globally available, due for a spring-clean I think in the future.

drbyte: reverted

PR #3289, opened 2020-04-16
Add CKEditor to Admin - via CDN

drbyte:


admin/includes/ckeditor.php
```diff
@@ -1,40 +1,51 @@
...
-    if ($(this).hasClass('editorHook') || ($(this).attr('name') != 'message' && ! $(this).hasClass('noEditor')))
-    {
-      index = $(this).attr('name').match(/\d+/);
-      if (index == null) index = <?php echo $_SESSION['languages_id'] ?>;
-      CKEDITOR.replace($(this).attr('name'),
-        {
-          coreStyles_underline : { element : 'u' },
-          width : '100%',
-          language: lang[index]
+<script src="https://cdn.ckeditor.com/4.14.0/standard-all/ckeditor.js" title="CKEditorCDN"></script>
```
mc12345678, line 22: Shouldn't this be `//cdn.ckeditor.com`... instead of starting with https: as a src file if it is possible that the admin may be accessed via http: as a result of not having a SSL certificate? This would then prevent a mixed content security issue in the admin of the store...

barco57: No, mixed content issues only happen when the base page is https: and you try to pull http: assets into it. You can pull https: assets into a http: (unsecure) page

PR #3298, opened 2020-04-17
Installer en us

torvista:
correct unclosed i tag


torvista:
Are these missing links deliberate for some magic help?
Clipboard01

zcwilt:
The links are disabled by javascript, although to satisfy the warnings could just be set to #

PR #3303, opened 2020-04-17
Change styling on admin help button

barco57:
Why use an inline style when a Bootstrap helper class is available


Zen4All:
> Why use an inline style when a Bootstrap helper class is available

because I don't know everything :laughing:

PR #3305, opened 2020-04-18
Fix admin keepalive legacy pages

barco57:
I found the keepalive "broken" on legacy pages because jAlert.css was not being loaded.

### DIFF
--- a/admin/includes/header_navigation.php
+++ b/admin/includes/header_navigation.php
@@ -44,7 +44,7 @@ $menuTitles = zen_get_menu_titles();
   </div><!-- /.navbar-collapse -->
 </nav>
 <?php if ($url = page_has_help()) { ?>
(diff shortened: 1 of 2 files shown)


barco57:
arrgggg I am apparently doing something wrong in source tree with branching cause one of these files is in my other pull request.

scottcwilson:
When you create your next branch, be sure you *first* checkout v157, *then* create the new branch. You always want to be branching off v157.

barco57:
Thanks, I figured out that Sourcetree 3.3.7 has an issue where the first tab doesn't refresh to show you where you are at, what branch you're actually on, so when I branched I ended up branching off of another branch instead of v157. I had to manually update Sourcetree to solve the problem so I can checkout the correct branch

Issue #3293, opened 2020-04-16
control of ADMIN_CONFIGURATION_KEY_ON


torvista:
This switch is in group 6, hidden. When set to 1 it causes the configuration_key name to be displayed:
![Clipboard03](https://user-images.githubusercontent.com/4391026/79507794-a00dbc00-8038-11ea-9405-942e953092a5.gif)
I think it is very useful and I think undocumented.
Could the switch for this not be put in the Developers Tool Kit?

scottcwilson:
Note that the keys are also documented in the docs:
https://docs.zen-cart.com/user/admin_pages/configuration/all/

torvista:
If I wanted to know a constant name I'd search in the db for the strings, if I didn't know about ADMIN_CONFIGURATION_KEY_ON, 
which is infinitely faster if you want to build some queries quickly, as I was yesterday.

I understand that list is for building the help on, but has to be maintained.

scottcwilson:
I'm open to this idea - just wanted to clarify on one point:

> I understand that list is for building the help on, but has to be maintained.
The help is automatically generated by a script.  It doesn't need manual maintenance.

scottcwilson:
Also thinking about config page 6 got me inspired to improve that page, so thanks for that.

wiztechinc:
ah, I couldn't find the other discussion about the help where you had asked me a question. The docs appear to have a way to edit them, supposedly expanding the info which is definitely needed. I like this approach using it in admin now but is this temporary for the help buttons?

scottcwilson:
> The docs appear to have a way to edit them, supposedly expanding the info which is definitely needed.

Not sure why you're saying this so tentatively.  If you want to change the documentation, you submit a PR.  

>  I like this approach using it in admin now but is this temporary for the help buttons?

I don't understand what you're asking here.   If you're unhappy about the buttons, please submit a PR.   If you're asking a different question, please rephrase it.

wiztechinc:
You keep saying doc info is generated automatically by a script. Will the script need to be run  again if changes occur? Does the script create the pages? If we edit the docs.zencart, does that make a permanent change that will not be overwritten by running the script again? I can definitely get started adding info into the docs. I just need clarity here. I never saw any discussion about creating this new set of docs so I'm just trying to figure out what's going on.

scottcwilson:
Please read the CONTRIBUTING and README files in the documentation repo. 

If you have more questions, open a separate issue.

[...]

PR #3307, opened 2020-04-18
Improve appearance of configuration group 6

scottcwilson:
These values are really hard to understand out of content - adding prefixes showing where they came from.  Inspired by thinking about #3293.

### DIFF
--- a/admin/configuration.php
+++ b/admin/configuration.php
@@ -7,6 +7,9 @@
  * @version $Id: mc12345678 2019 Apr 30 Modified in v1.5.6b $
  */
 require('includes/application_top.php');
+define('MODULE_SHIPPING','MODULE_SHIPPING');
+define('MODULE_PAYMENT','MODULE_PAYMENT');
+define('MODULE_ORDER_TOTAL','MODULE_ORDER_TOTAL');
 
 $action = (isset($_GET['action']) ? $_GET['action'] : '');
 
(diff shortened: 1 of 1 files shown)


drbyte:
- not sure I like calling this "page 6". Page? Maybe "configuration group"
- not sure I want to do this at all
- got a screenshot of before/after?

scottcwilson:
See screenies: 

before
after

scottcwilson:
I wouldn't say it "encourages" the use of the page - I'd say it makes the page more comprehensible to anyone who wants to use it.   It would definitely be a good idea if you want to move forward on Steve's suggestion in #3293.

torvista:
I think the fact that these constants get displayed at all is just incidental. They are not really for human consumption....if some are/are of use, they should be on another user-accessible page. Otherwise, I don't think anyone would expect to them be prettified at all.
In my opinion, of course!

Zen4All:
I think group 6 should only be available for people who really know what they are doing.
I have not looked at the new code, but I have no clue what you are trying to display/explain in the second image.

scottcwilson:
It just expands the configuration title string to give it some context.  Instead of saying "Tax Class" it shows that this is the Tax Class for the Per Item shipping method, for example.

Zen4All:
OK, thanks for the explanaition :relaxed:

drbyte:
Sentiment appreciated, but I'm not sure I want this. Closing for now.

PR #3308, opened 2020-04-18
ShoppingCart: Skip loop through products if db query is empty.

drbyte:
`shopping_cart::get_products()`:
When the db query is customized to check only for products which match certain criteria, not aborting this loop will cause incorrect empty entries in the resulting array, and will trigger numerous "index not found" notices.
Checking for validity allows skipping the unnecessary processing of this large section of code and avoids false entries.

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -1171,7 +1171,8 @@ class shoppingCart extends base {
                            AND pd.products_id = p.products_id
                            AND pd.language_id = " . (int)$_SESSION['languages_id'] . " LIMIT 1";
 
-      if ($products = $db->Execute($products_query)) {
+      $products = $db->Execute($products_query);
+      if (!$products->EOF) {
         $this->notify('NOTIFY_CART_GET_PRODUCTS_NEXT', $products_id, $products->fields);
 
         $prid = $products->fields['products_id'];


includes/classes/shopping_cart.php
```diff
@@ -1171,7 +1171,8 @@ function get_products($check_for_valid_cart = false) {
                            AND pd.products_id = p.products_id
                            AND pd.language_id = " . (int)$_SESSION['languages_id'] . " LIMIT 1";
 
-      if ($products = $db->Execute($products_query)) {
+      $products = $db->Execute($products_query);
+      if ($products->RecordCount()) {
```
lat9, line 1175: Sorry to be pedantic, but should that be either
```php
    if ($products->RecordCount() !== 0) {
```
or
```php
    if (!$products->EOF) {
```

scottcwilson: `      if ($products->RecordCount()) {`
is the same as 
`    if ($products->RecordCount() !== 0) {`

mc12345678: Probably seen a reason before, but wondering why either alternative is considered necessary? Is it just "forward thinking" in that a "count" would expect a whole number or to be at the end of the listing?  Why though use `===` if the desired goal would be to not execute the internal code if it returned `false` or something `falsy`?

drbyte: I believe @lat9 is simply seeing it through the lens of strict types.
While the test for `falsy` here is perfectly functional, I'll concede that there's an argument to be made for clear indication of intent, so that the future reader doesn't have to wonder what was meant by what was written.
Thus in this case I will switch to testing for `EOF`.

(I don't like using `!== 0` or `=== 0` as I feel they convey the worry that it might ever not be numeric, which actually creates more confusion when reading code later, just like @mc12345678 asked about whether it's for "forward thinking" concerns.)

drbyte: Changed to `->EOF`

PR #3311, opened 2020-04-19
Update old tutorials links to docs site

scottcwilson:
Note: The file docs/important_site_security_recommendations.html should really be removed and replaced with an htaccess redirect to the new location, 

https://docs.zen-cart.com/user/security/security_recommendations/ 

Alternately, the file content could be removed and a simple link could be left in, redirecting the user to this new location.

### DIFF
--- a/docs/0.about_zen_cart.html
+++ b/docs/0.about_zen_cart.html
@@ -196,7 +196,7 @@
     <div>
         <h1>Zen Cart&reg; Requirements</h1>
         <ul>
-            <li>For up-to-date requirements, see: <a href="https://www.zen-cart.com/content.php?48-what-are-the-server-requirements-to-run-zen-cart" target="_blank"> Zen Cart&reg; Server Requirements</a></li>
+            <li>For up-to-date requirements, see: <a href="https://docs.zen-cart.com/user/first_steps/server_requirements/" target="_blank"> Zen Cart&reg; Server Requirements</a></li>
             <li>Apache must be configured with AllowOverride set to either "All" or at least both "Limit" and "Indexes" parameters, and preferably the "Options" parameter as well.</li>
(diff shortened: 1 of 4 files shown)


scottcwilson:
Indeed we could also get rid of 
1.readme_installation.html
2.readme_how_to_upgrade.html

and replace them with links to 
https://docs.zen-cart.com/user/first_steps/how_do_i_install/ 
https://docs.zen-cart.com/user/upgrading/upgrading/

mc12345678:
Wouldn't replacing files in the distribution with outbound links then require a network connection rather than having a relatively self contained functional package of complete information? Ie. no way to use without having an outbound link... Shouldn't the package basically be self contained and functional without having to reach out to somewhere? (Even if the result is that the information is possibly slightly out-of-date...) Wouldn't it be more prudent to capture the current information to provide that full package but then also include a link to obtain the latest if desired/able?

scottcwilson:
Yes.  No.  No.

dbltoe:
MC, if the user did not have a network connection, how did they get the files in the first place?
With the age of mobile, tablet, and desktop I think the lack of connection would be the rarity.
And, no better way to get the latest info than going to a current network location versus an old, out-dated file

drbyte:
There's an argument that if the docs are in the zip then they're the right docs for that zip, and not mismatched.
And working on a laptop on the train doesn't mean I have an internet connection.

scottcwilson:
OK.  Two questions: 

Do you want me to refresh the content on the pages with @dbltoe updates?
Do you want me to add a link at the top of these files pointing to docs.zen-cart.com?

drbyte:
> OK. Two questions:
> 
> Do you want me to refresh the content on the pages with @dbltoe updates?
> Do you want me to add a link at the top of these files pointing to docs.zen-cart.com?

Sure

PR #3313, opened 2020-04-20
Don't double sanitize comments

### LINKED ISSUE #3312: Comments, quotation marks shown as html on order confirmation
quotes_in_attributes_and_comments
related to #2677

scottcwilson:
Fixes #3312

### DIFF
--- a/includes/templates/responsive_classic/templates/tpl_checkout_confirmation_default.php
+++ b/includes/templates/responsive_classic/templates/tpl_checkout_confirmation_default.php
@@ -87,8 +87,7 @@
 
   <h2 id="checkoutConfirmDefaultHeadingComments"><?php echo HEADING_ORDER_COMMENTS; ?></h2>
   <div class="buttonRow forward"><?php echo  '<a href="' . zen_href_link(FILENAME_CHECKOUT_PAYMENT, '', 'SSL') . '">' . zen_image_button(BUTTON_IMAGE_EDIT_SMALL, BUTTON_EDIT_SMALL_ALT) . '</a>'; ?></div>
-  <div><?php echo (empty($order->info['comments']) ? NO_COMMENTS_TEXT : nl2br(zen_output_string_protected($order->info['comments'])) . zen_draw_hidden_field('comments', $order->info['comments'])); ?></div>
-
+  <div><?php echo (empty($order->info['comments']) ? NO_COMMENTS_TEXT : nl2br($order->info['comments']) . zen_draw_hidden_field('comments', $order->info['comments'])); ?></div>
 </div>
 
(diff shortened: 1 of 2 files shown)


drbyte:
Please explain how you arrived at the conclusion that it's safe to entirely skip sanitizing these outputs.

scottcwilson:
They already do get sanitized.  Using 

<script>alert("hi");</script>

as a test did not fire.

scottcwilson:
(This change breaks account history info though...)

scottcwilson:
When I say they already get sanitized, see includes/modules/pages/checkout_shipping/header_php.php line 140. 

      $_SESSION['comments'] = zen_output_string_protected($_POST['comments']);

Similar in includes/modules/pages/checkout_confirmation/header_php.php line 49

zcwilt:
Going to leave this open for a few more days. Haven't had a chance to test.
Will say this though. The original sanitization was there based on PADSS reviews.
The sanitization is also very much flow dependant. e.g. returning to a previous checkout step.

Not saying you are wrong, but I always want to double check where security is concerned

Much like the carpenter, Measure twice, cut once.

scottcwilson:
This PR is by no means urgent - wait until 1.5.8 if you prefer.

zcwilt:
I have duplicated the problem but still not convinced this is the best fix. 

As a general principle html outputs should be sanitized at the point of output, especially where input comes from user entered data. In fact this is a fundamental principle of security to protect against XSS. 

While there is no doubt that somewhere we are double sanitizing, I would prefer that we identify those points. e.g. @scottcwilson  points out the sanitization in 
includes/modules/pages/checkout_shipping/header_php.php
and 
includes/modules/pages/checkout_confirmation/header_php.php

Would it be better to remove the sanitization at those points ?

zcwilt:
One reason we may be doing the extra sanitization in
includes/modules/pages/checkout_shipping/header_php.php etc
is to protect against reflected xss in admin. 
i.e. storing unsanitized comments in the database and then outputting this unsanitized in the admin.
I haven't yet followed that bouncing ball.

zcwilt:
Will still leave this open for a while to see if I can trace all of the workflows.

scottcwilson:
This is by no means hot - we could defer until later and take our time.

zcwilt:
Have tested this pretty thoroughly and the correct fix is to remove the sanitization in the header_php.php files.
Account History doesn't break and admin orders does not reflect any xss either.

scottcwilson:
Thanks - happy to abandon this if a superior solution has been found.

zcwilt:
k, will close this and we can do a new PR

PR #3314, opened 2020-04-20
Fix sizeof test on variable that could be empty

scottcwilson:
Fix issue reported on 
https://www.zen-cart.com/showthread.php?226624-PHP-Warning-Account_notification-header_php


drbyte:
Need to remove the `> 0` part.

includes/modules/pages/account_notifications/header_php.php
```diff
@@ -45,7 +45,7 @@
     $sql = $db->bindVars($sql, ':customersID',$_SESSION['customer_id'], 'integer');
     $db->Execute($sql);
 
-  } elseif (sizeof($products) > 0) {
+  } elseif (!empty($products) > 0) {
```
drbyte, line 48: Wrong syntax. Must remove `> 0`

PR #3315, opened 2020-04-20
fixes count error

proseLA:
turn on global notifications for products and not have any products to be notified on your account_notifications page.  you get a countable error:

`PHP Warning: sizeof(): Parameter must be an array or an object that implements Countable in /var/www/includes/modules/pages/checkout_success/header_php.php on line 114.`

i like properly initializing the array as a solution.

### DIFF
--- a/includes/modules/pages/checkout_success/header_php.php
+++ b/includes/modules/pages/checkout_success/header_php.php
@@ -103,6 +103,7 @@ if ($flag_global_notifications != '1') {
   $products_query = $db->bindVars($products_query, ':ordersID', $orders_id, 'integer');
   $products = $db->Execute($products_query);
 
+  $notificationsArray = [];
   foreach ($products as $product) {
     $notificationsArray[] = array('counter'=>$counter,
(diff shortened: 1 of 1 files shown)


includes/modules/pages/checkout_success/header_php.php
```diff
@@ -103,6 +103,7 @@
   $products_query = $db->bindVars($products_query, ':ordersID', $orders_id, 'integer');
   $products = $db->Execute($products_query);
 
+  $notificationsArray = [];
```
drbyte, line 106: Since the array is already initialized on line 88, let's not add it again.
Perhaps move it up before the loop so that it's available if the loop doesn't get processed.

proseLA: i did not notice that.  not sure why that error would come up.  let me explore.

proseLA: ok. this error was fixed already.  my mistake.  will delete.

PR #3246, opened 2020-04-13
Admin: option_names_manager overhaul

torvista:
This was mostly done in September with a long commit history but was pre-empted by another PR.
Those changes have been mostly incorporated into this but the commit history got hosed in the interim.
So, sorry for yet another big bang.  It is a very complicated file, with a lot to test....

Fixes:
insert of new option: sql had these fields expected:products_options_images_per_row, products_options_images_style, products_options_rows as POST values, but are not set on insert: removed.
Insert of new option: the next option id to use was being defined in the form generation on the previous page, multiple users could obtain the same next_id. Moved it to immediately prior to the INSERT.
check for duplicate name: used COUNT in the sql, then ->recordcount which is always 1, so no alert. Decided to not use count as it does not count nulls.
usual html validations for unclosed tags etc. and label for not pointing at the id of a corresponding input fields.

Changes in no particular order
Additions to two functions, maintaining legacy functionality, to allow ordering and formatting/filtering of results
Improved formatting.
Greatly expanded notification messages for every type of success and fail, with corresponding changes to constants.

Global tools:
removed unecessary show/hide: tools do not run any queries until a Option Name select dropdown is used. Change of an Option name reloads the page with the dropdown results filtered.
Added sort option to dropdown.
Added full path option to dropdown.

PR #3325, opened 2020-04-21
Ensure consistent int casting.

mc12345678:
There are other places in the code where the same variable(s) are
cast to `int`, but were not done so here.  Incorporated for
consistency.

### DIFF
--- a/admin/options_name_manager.php
+++ b/admin/options_name_manager.php
@@ -597,21 +597,21 @@ function translate_type_to_name($opt_type)
                     $inputs2 .= '</div>';
                     $inputs2 .= '<div class="row">';
                     $inputs2 .= '<div class="col-sm-4">';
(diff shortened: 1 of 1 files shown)


torvista:
Are these ints necessary? They come from an integer field in the table and in the function zen_draw_input_field, they go through zen_output_string too.

mc12345678:
A few related posts may at least enlighten if not encourage further considerations:
https://www.zen-cart.com/showthread.php?226168-Using-exactly-equal-comparisons-on-database-fields
https://www.zen-cart.com/showthread.php?226070-PHP-mixed-comparisons-a-cautionary-tale

Suggesting then to remove the other occurrences of casting to integer? Could be made "easier" if earlier on the value(s) were recast without the "need" for subsequent casting... I've seen both methods used recently and not sure which is considered more appropriate/desired.  There are pros and cons for both and structure of other available actions plays a factor as well (e.g. notifiers that could allow the value to be changed from it cast to one thing and potentially reset to be another.)..

torvista:
My opinion is that things should be cast to what they are when created or when they "enter" a script, then no need for casting subsequently, in theory....
When POST and GET are sanitized or checked to a new variable. 
When loading an array from a sql result object.
I guess the problem is the globals that get used.
I'm suprised there is not more impetus to make things more strict as we go along, it all highlights errors.

zcwilt:
Will leave this open for now, but expect it to be in conflict once #3246 is merged.

PR #3326, opened 2020-04-21
Remove additional HTML space.

mc12345678:
There is a space that is incorporated just after each of the
opening lines that can not be modified/adjusted by the associated
string constant and appears unnecessary.  If a space is needed/desired
it can be added into the constant.

### DIFF
--- a/admin/options_name_manager.php
+++ b/admin/options_name_manager.php
@@ -648,12 +648,12 @@ function translate_type_to_name($opt_type)
                     <td colspan="6">
                       <?php echo TEXT_OPTION_IMAGE_STYLE; ?>
                       <ul style="list-style-type: none">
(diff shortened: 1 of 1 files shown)


torvista:
I got rid of the list as it was pointless to attempt to describe different formats: used a link to an image.

mc12345678:
> I got rid of the list as it was pointless to attempt to describe different formats: used a link to an image.

Not sure I follow, but time will tell. Whatever provides some sort of "explanation" to the user whether sighted or not.

barco57:
uh oh, Chad has done it now, He has hinted at Accessibility... I must now go and read the Web Content Accessibility Guidelines (WCAG) before we are in up to our necks with WAI-ARIA tag implementation.


Where is the sarcasm emote?

zcwilt:
Will leave open for now, expect to conflict with #3246

PR #3327, opened 2020-04-21
Restore original code comments

mc12345678:
There has been a push to remove any/all comments in files; however,
these comments provide assistance in matching the code to the
onscreen display.  If they are found to be incorrect then they
should be updated instead of simply removed.


torvista:
Yep, this file really needs comments, it's a lot of code. There's practically a conversation going on in mine.

PR #3328, opened 2020-04-21
Restore footer position to original.

mc12345678:
No reason has been provided for moving the footer to within
the divider.  Without reason it should not be moved as may/will
cause additional work for coders that affect this page through
use of the footer's notifier.

If there is a reason, then it should be made known public for
incorporation/consideration of use.


torvista:
Yes, done that.

mc12345678:
I leave this PR (as always) in the hands of the core team. :)

zcwilt:
Will close this as #3246 changes footer as well. Hope to merge #3246 soon.

PR #3329, opened 2020-04-21
Resequence placement of parameter separator

mc12345678:
While reviewing code associated with the changes to this file,
noticed that there was at least one instance of a url being
generated to where it would/may end with a single `&`. This
places the parameter separator (`&`) with the parameter that
is being added instead of as an intermediate character or as
a final ending.  Such use is not necessary for the first
parameter as functions such as `zen_href_link` will incorporate
the appropriate first separator as necessary.

### DIFF
--- a/admin/options_name_manager.php
+++ b/admin/options_name_manager.php
@@ -675,9 +675,9 @@ function translate_type_to_name($opt_type)
                       <td>&nbsp;</td>
                     <?php } else { ?>
                       <td class="text-right">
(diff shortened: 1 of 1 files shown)


torvista:
Well spotted, I've incorporated this.
https://github.com/zencart/zencart/pull/3246/commits/941e239dfca30f1915d8c58af72c9e712ebe0064

mc12345678:
Sweet.. Just one of the things I'm trying to do as I am taking in everyone's code changes. I doubt I'll see all of them and some things I just accept without question based on time and overall revision.  Anyways, I may not always "sit" on a change awaiting someone else to catch something and while I try to have/maintain conversations, sometimes I see that I either need to say something or halt die Klappe.  

I know I ask a lot of questions and am generally not concerned about expressing my opinion, though I certainly don't want to cause contributors to stop contributing because I might come across as a donkey or seem to make things difficult. :) Like I said elsewhere today might hurt my one feeling, but I'm sure I'll get over it.

torvista:
S'all good to me, the lack of comments I get about my commits worries me...I don't want to be hoist on my own petard next month.

zcwilt:
Will leave open for now, until #3246 is merged.

PR #3330, opened 2020-04-21
Options Name Manager - use function for language icon.

mc12345678:
Although the original code of this function used the language's
`name` for the alternate text and `zen_get_language_icon` uses
the language's `directory`, substituted the associated function
instead of effectively duplicating code.  If it is desired to
change the alternate text to be the language's name, then would
suggest doing so in the associated function rather than at
individual code points throughout the software or create an
alternate function/parameter that target's the data to be desired.


torvista:
Hi, you are doing stuff with this file but there is a PR pending that stamps all over this file, so I fear you may be wasting your effort...

drbyte:
@torvista does your PR incorporate this change?

torvista:
Yes, and shows no icon when only one language:
`$inputs .= zen_draw_label(($n > 1 ? zen_get_language_icon($languages[$i]['id']) . ' ' : ' ') . TABLE_HEADING_OPTION_NAME . ':', 'option_name[' . $languages[$i]['id'] . ']', 'class="control-label"');`

mc12345678:
Hate to ask the question(s) because don't want to lose the development consideration(s) and/or momentum.

Why display the language indicator only for a site that has multiple languages installed? What guidance is present on the page to know what language is expected/supported? (I don't typically look for the overall language of a site, but instead what is expected where text is being entered.)

torvista:
Single language:
Clipboard01
Multiple language:
Clipboard02

You mean you really want that absurd EN icon all the time! Or would be happy with the placeholder?
Or, are you talking about sites that have deleted the english and only have one non-english languages?

Zen4All:
> Single language:
> Clipboard01
> Multiple language:
> Clipboard02
> 
> You mean you really want that absurd EN icon all the time! Or would be happy with the placeholder?
> Or, are you talking about sites that have deleted the english and only have one non-english languages?

The language icons ar not displayed the way you have them in your screen shot. it is like this (and I replaced the new "English" icon with the old one):
Screenshot from 2020-04-22 09-38-51

As you can see the image is inside a input group. Only showing icons on multi language shops will complicate the code IMHO. If we doit on this page it should be done ZC wide :unamused:

torvista:
I showed what I have done in my PR.
I can use your example if preferred.

I've mentioned this before regarding the en icon. It is a mistake.
If you want to be inclusive, include the individual icons separately and those who care can use the one they want. But this composite is just wrong. In my spanish language pack I overwrite it with the old one.

Zen4All:
--- START RANT --
A bit off topic, but what I do not understand, is why the English icon has to include multiple flags.
If that is the way to go we should also add more flags to the Dutch,  the Spanish, French, and Portugese icon. 
This also means that the American flag should be added to the Spanish Icon :rofl: , as a lot of americans speak Spanish.
--- END --

torvista:
Somebody please PR it back to the way it was...it wasn't broken/didn't need fixing.

drbyte:
> --- START RANT --
> A bit off topic, but what I do not understand, is why the English icon has to include multiple flags.
> If that is the way to go we should also add more flags to the Dutch, the Spanish, French, and Portugese icon.
> This also means that the American flag should be added to the Spanish Icon 🤣 , as a lot of americans speak Spanish.
> --- END --

Yes, entirely off-topic, unless the whole reason for these side comments is the dislike of the icon. That's been discussed elsewhere already. As an update I've put out a request for copyright approval for us to distribute an updated icon, but haven't heard back. Maybe I'll have to find another source. Maybe country flags are the wrong indicator. Maybe font-awesome instead. 
But this is completely off topic, and should be discussed in its own thread.

torvista:
ok, back to the burning issue of the day, what format do the cognoscenti prefer: existing, or as I have proposed, or a mixture, or animated gifs of national sterotypes (druken/vomiting brits, smoking french, shouting/shooting yanks etc..)?

Zen4All:
> > --- START RANT --
> > A bit off topic, but what I do not understand, is why the English icon has to include multiple flags.
> > If that is the way to go we should also add more flags to the Dutch, the Spanish, French, and Portugese icon.
> > This also means that the American flag should be added to the Spanish Icon rofl , as a lot of americans speak Spanish.
> > --- END --
> 
> Yes, entirely off-topic, unless the whole reason for these side comments is the dislike of the icon. That's been discussed elsewhere already. As an update I've put out a request for copyright approval for us to distribute an updated icon, but haven't heard back. Maybe I'll have to find another source. Maybe country flags are the wrong indicator. Maybe font-awesome instead.
> But this is completely off topic, and should be discussed in its own thread.

Youŕe right, won't happen again. Thank you for the explanation. Font-awesome might be an option.

torvista:
>dislike of the icon. That's been discussed elsewhere already.
Where?

scottcwilson:
@torvista America is a gun. :(

drbyte:
> ok, back to the burning issue of the day, what format ... preferred?: existing, or as I have proposed, or a mixture, or animated gifs

I don't know how this PR about calling a built-in function has morphed into a fight over icon display styles.

scottcwilson:
I blame it on quarantine.

drbyte:
From the display perspective my preference is what Zen4All proposed: embrace the input group akin to the other bootstrap markup we've been updating:
Screenshot from 2020-04-22 09-38-51

Granted, it may pose challenges with textarea inputs. It's easy for text fields, but textareas can be a bit challenging to format sometimes. I haven't tried it with a bootstrap input group.

I'm okay with hiding the language-icon if only one language is offered; however, it should be consistent across the application, and  .... well ... you tell me: have you already introduced enough upgrade pain for everyone with all the admin changes, that it makes sense to dump it all in at once now? Or do people actually still care how hard upgrades are for end-users and mod-maintainers?

barco57:
> Or do people actually still care how hard upgrades are for end-users and mod-maintainers?
That depends on who you ask

how about drop the icons and go with EN, ES, DE, FR, etc.... just my two cents

wiztechinc:
> > Or do people actually still care how hard upgrades are for end-users and mod-maintainers?
> > That depends on who you ask
> 
> how about drop the icons and go with EN, ES, DE, FR, etc.... just my two cents

Agreed. The flags have caused issues with clients insisting on me finding a usa flag to use before the change, but showing multiple flags just takes up space. They are actually distracting. Upgrading? Whatever. I have one this will impact and it’s had few changes over the years. I expect more changes as time goes on. It’s just part of working on a project like this.

torvista:
language icon ranting here:
https://github.com/zencart/zencart/issues/3335

torvista:
Discussion of format for Add New Option:
https://github.com/zencart/zencart/pull/3246#issuecomment-618276871

zcwilt:
My 2 cents - hate language icons !

PR #3334, opened 2020-04-22
v157 Disallow product copy to category with subcategories

scottcwilson:


mc12345678:
Unfortunate to bring this up in this PR, but what is expected of the new store owner that has first created product without creating sub-categories??? Or more so with this PR, the desire to move product up a category and remove a previously existing sub-category??? Why isn't more done to the front end to prevent "loss" of such display when product is active in a sub-category with sub-categories or perhaps to disable product that are in a category with sub-categories rather than out right disallowance of existence??? Why are these things being so tightly controlled/prevented? 

------- kind of rant? Over?!??? -------

scottcwilson:
On a current branch of 1.5.7 (or any recent vintage actually), try doing what this PR prevents.  For example, on the sample data, copy one of the Mice products to the Software category. 

Here's what happens:

- If the product is disabled, the software category shows No products.  Existing subcats are hidden.
- If the product is enabled, the software category shows ONLY the mouse product.  Existing subcats are hidden. 

Can the average storeowner debug this situation? Nope.

scottcwilson:
> what is expected of the new store owner that has first created product without creating sub-categories??? 
I have never heard of this happening.  But if it does happen, it can be fixed by creating the new category in phpMyAdmin and then doing the move (probably also using phpMyAdmin). 

> Or more so with this PR, the desire to move product up a category and remove a previously existing sub-category?
- Create new category at top level
- move all products there
- delete the old parent category of the products, which is now empty.
- now you can move the products back to their old parent's parent

wiztechinc:
> > what is expected of the new store owner that has first created product without creating sub-categories???
> 
> I have never heard of this happening. But if it does happen, it can be fixed by creating the new category in phpMyAdmin and then doing the move (probably also using phpMyAdmin).
>
Not only do folks create products with no subcategories, some do it deliberately. Very common but the result is the same, usually phpmyadmin work. I really wish there was an easier way to fix/change. Like I said I’ve seen it many times.

torvista:
This file has probably received more modification than any other, a lot from me
I'm a bit shocked that this safeguard was not implemented already!

mc12345678:
I hardly consider it a safeguard... The entire reason for this "restriction" is to support the process in place for items to be given a sale... A store that never puts a category "on sale" (could be done with one or more specials) has no concern for preventing product from being in categories that have sub-categories and can even alter the way the associated header_php.php file operates/returns data and then this "restriction" is nothing. Further, the absence/presence of the other categories at the top of the page can be handled/modified. For one, the side box showing categories still works. Depending on the setting to move to a product in a category when the category has only one product plays a factor in directly moving to the product if there is only the one product in the category (for my testing I moved both mice into the software category).

Ultimately, what I'm getting at is why not just display a notification on the page when this condition exists rather than stopping the movement? If someone has purposefully setup their site to have both product and category in a category and find the message an unnecessary annoyance then they can do something about it.  This change along with another made earlier just makes use of the admin more difficult especially when one of the solutions is to go to a tool such as phpMyAdmin in order to "defeat" a supposed protection...

Sorry for the delay in responding; however, this has taken some collection of data and reobservation of operations to speak to this again.... I've looked at possible ways to incorporate at least the initial issue of product and category existing in a category, though seems that no matter what at least one more set of queries has to be run which really is something I am trying to prevent.  Of course, in a default install of 1.5.7 that shouldn't be possible to see as the logger system has been moved out (to a plugin?) and even with its report of "number of queries" that's a little overzealous as some (many?) are cached but still counted...

scottcwilson:
Did you try the testcase I outlined in my response to your first post?

The reason for this restriction is that it doing this move screws up your store and winds up hiding things.

Zen4All:
I like mc12345678 am for a warning, rather than a restriction. I often need to move products around, removing subcategories. This restriction will make it impossible to do that easily, or without going into the database directly.

drbyte:
If this PR is about preventing "move" of categories, why is the code change in `copy_product_confirm` ... for "copying" products?

scottcwilson:
This PR is not about preventing the move of categories.  It's about products.

drbyte:
Sorry, typo.

I'll re-ask: if this is to prevent *MOVE* of products, why is the code change in the *COPY* section?

scottcwilson:
The title is misleading.  Corrected.

wiztechinc:
I just spent an hour rearranging the categories and subcategories on a site. He didn't even realize subcategories were possible so I had to jump through ridiculous hoops to get there. On a 1.5.6 site, I found it easier to move the category into the existing category and then move the products into that new subcategory and other new ones that I moved there. One reason why it was easier was just the move product dropdown. Once it appeared, then it was much easier to find the category to move it to since it's in the immediate area. What was horrible was the fact that after every product move, it switched to display the category it was moved to and I had to back out and find the original category again. That was the big issue.

I beg of you do not change this and allow moving of the categories.  Now it does a good job of informing that the subcategory doesn't work where you are moving it to. This should be adequate. And then I have an easier time of straightening out category disaster areas. It happens more than you can possibly imagine as shop owners frequently have no clue how to organize their products.

zcwilt:
At this point in time I'm minded to defer this. While @scottcwilson PR does make sense from a safety point of view, and in fact over the history of Zen Cart we have added other safeties to prevent the problem of adding products into top level categories with sub-categories, maybe we need to address the structural problem.

From my `albeit failing` memory, this restriction has been around from pretty much the start of Zen Cart, and may even have been a hold over from original oscommerce code. 
My feeling is that it was put into place to `simplify' code for doing products counts etc and to `shave off` a few recursive queries. 
There was a time in Zen Cart where saving 20 queries on a page was a big thing. Servers in the past 15 years have moved on.

scottcwilson:
Defer is very reasonable, and to address the concerns raised here, maybe there needs to be a "Expert" flag of some sort so that allows you to bypass the normal protections.

zcwilt:
It should be noted that removing the current position of not allowing products in the top level of a category that also has sub-categories would be a huge change to core code, so I'm not necessarily advocating that. Just suggesting we need a wider conversation.

For now will mark as `on hold` till after v157 release.

drbyte:
As discussed, this PR currently does a hard rejection of copying products to a category that has subcategories.

I wonder if we were to enable the "Display * if category has products" might be an aid to help encourage choosing a suitable category? (I acknowledge that a category-with-products is not the same as a category-without-subcategories)
This could be done with extra params to `zen_get_category_tree` in `modules/copy_product.php` line 29 via:

```diff
-$contents[] = array('text' => zen_draw_label(TEXT_CATEGORIES, 'categories_id', 'class="control-label"') . zen_draw_pull_down_menu('categories_id', zen_get_category_tree(), $current_category_id, 'class="form-control" id="categories_id"'));
+$contents[] = array('text' => zen_draw_label(TEXT_CATEGORIES, 'categories_id', 'class="control-label"') . zen_draw_pull_down_menu('categories_id', zen_get_category_tree(0, '', '', array(), false, true), $current_category_id, 'class="form-control" id="categories_id"'));
```

Nothing explains there what the `*` denotes though, so it's not perfect.

drbyte:
As suggested earlier, another option is an "expert mode".
There's no in-page-update option to have an "expert mode toggle" to swap out the list of categories offered.

There's also not currently a function to generate the list of categories which are suitable for putting products into ... which is what should be displayed if expert-mode is not enabled yet.

In lieu of an in-page-update, a simple button could be added that sets a GET param (maybe `expert_mode=1`) and reloads the page, which would now show the unedited list of subcats (as it does presently)

scottcwilson:
guru_mode? in the spirit of all things zen.

scottcwilson:
Closing - we can revisit later or make it a plugin.

PR #3340, opened 2020-04-24
Only display including tax if store runs with DISPLAY_PRICE_WITH_TAX == 'true'

### LINKED ISSUE #3339: Tax display in admin
If I buy something and pay sales tax on it, the (incl) columns in admin are the same as the (excl) columns.  See attached for example.  I has been this way as far back as 1.5.4 (and likely much further) but this makes no sense to me.  

In the attached, why isn't Price (incl) = 534.99?  (499.99 * 1.07) 

taxes

scottcwilson:
Fixed only the order details screen.  If approved, I will also change the invoice screen. 

Fixes #3339 
tax_fixed

### DIFF
--- a/admin/includes/languages/english/invoice.php
+++ b/admin/includes/languages/english/invoice.php
@@ -19,7 +19,8 @@ define('TABLE_HEADING_PRICE_EXCLUDING_TAX', 'Price (ex)');
 define('TABLE_HEADING_PRICE_INCLUDING_TAX', 'Price (inc)');
 define('TABLE_HEADING_TOTAL_EXCLUDING_TAX', 'Total (ex)');
 define('TABLE_HEADING_TOTAL_INCLUDING_TAX', 'Total (inc)');
-
+define('TABLE_HEADING_PRICE', 'Price');
+define('TABLE_HEADING_TOTAL', 'Total');
 define('ENTRY_CUSTOMER', 'CUSTOMER:');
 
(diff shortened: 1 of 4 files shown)


drbyte:
I'm not sure this is the right "fix" if a fix is required.

scottcwilson:
I am open to suggestions.  But how could it be correct that (incl) and (excl) are the same on an order with tax?

drbyte:
> I am open to suggestions. But how could it be correct that (incl) and (excl) are the same on an order with tax?

They're not displayed based on whether the order "has tax", but rather whether the store has "tax-included prices" (EU for example)

drbyte:
> tax_fixed

Why the changed position of the subtotals column?

scottcwilson:
I found it really confusing to have the subtotal be smaller than the line above it.

drbyte:
> > Why the changed position of the subtotals column?
> I found it really confusing to have the subtotal be smaller than the line above it.

Ya, that's the problem with adding tax at the row-level.

Plus, the math in your image looks wrong if your 2 line-items are intended to show 1 as taxable and one as not.

scottcwilson:
Invoice added.  Ready for merge.

drbyte:
I'm gonna be picky again here.
I'd rather see the `_NO_TAX` suffix removed from these constants, because saying "no tax" implies tax-exempt upon initial reading.
(While I "get" the initial intention, when looking at it from a different context the meaning is lost.)

PR #3346, opened 2020-04-25
Revert val_function change to SEND_EXTRA_CREATE_ACCOUNT_EMAILS_TO_STATUS

### LINKED ISSUE #3345: SEND_EXTRA_CREATE_ACCOUNT_EMAILS_TO_STATUS should NOT be sanitized
I did a fresh install of 1.5.7 on XAMPP to do some poking around to familiarize myself with it. Yesterday I encountered a problem in admin->Configuration->E-Mail Options.

I was attempting to change the "Send Copy of Create Account Emails To - Status" and kept encountering the TEXT_EMAIL_ADDRESS_VALIDATE error.

All of the other "Send Copy of * Emails To - Status" worked without error.

After a bit of investigating, I discovered that the configuration for SEND_EXTRA_CREATE_ACCOUNT_EMAILS_TO_STATUS had the val_function field defined while the other SEND_EXTRA_*_EMAILS_TO_STATUS did not.

Using phpMyAdmin, I removed the val_function field and was able to change "Send Copy of Create Account Emails To - Status" without error.

This configuration item used the same value of val_function as SEND_EXTRA_CREATE_ACCOUNT_EMAILS_TO. This doesn't make sense to me since SEND_EXTRA_CREATE_ACCOUNT_EMAILS_TO_STATUS is NOT an email address. It selects between 2 values using radio buttons.

scottcwilson:
Fixes #3345


drbyte:
Scott, can you please also reference here the other PR numbers where this problem was introduced or exacerbated.

scottcwilson:
Initial change for installer: #2267 
Added to upgrade: #2616

PR #3348, opened 2020-04-26
orders - fix php notice for rare situation when all modules are disabled

torvista:
disable payment and shipping modules, import db to ZC157, upgrade, view orders: php notice
PHP Notice: Undefined variable: module in /.../admin/orders.php on line 665.

`        if (is_object($module) && method_exists($module, 'admin_notification')) {`


drbyte:
Will merge, but It is preferable to have done these 2 completely-separate-topic commits as separate PRs so that the history log makes semantic sense when investigating changes in the future.

PR #3351, opened 2020-04-26
Add notifier to change payment/shipping info

scottcwilson:
Adds delivery and billing country to `$orders` variable.  This will enable mods to show a country code, flag, etc. for stores with a lot of international activity.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -1124,7 +1124,7 @@ if (zen_not_null($action) && $order_exists == true) {
                   $order_by = " ORDER BY o.orders_id DESC";
                   $zco_notifier->notify('NOTIFY_ADMIN_ORDERS_SEARCH_PARMS', $keywords, $search, $search_distinct, $new_fields, $new_table, $order_by);
 
-                  $orders_query_raw = "SELECT " . $search_distinct . " o.orders_id, o.customers_id, o.customers_name, o.payment_method, o.shipping_method, o.date_purchased, o.last_modified, o.currency, o.currency_value, s.orders_status_name, ot.text as order_total" .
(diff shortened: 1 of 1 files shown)


scottcwilson:
It's the first parameter ... did I do that wrong?

drbyte:
I was going to ask similarly: Is there intention to update the `$show_payment_type` variable? Or only to update the `$orders` query result?

drbyte:
The first parameter to a Notifier hook is immutable. The rest are mutable.

scottcwilson:
OH!  Sorry - will reverse the order.

scottcwilson:
Note: The intent here is just to provide a way to indicate the billing and shipping countries - if there are other ideas, that would be fine too.

drbyte:
I'm concerned that that payment/shipping column is already pretty busy.
Would it make sense to insert another column somewhere, instead of trying to fudge the current one without it breaking to a 3rd line or widening it without being able to style the country-indicator consistently due to differing line-lengths?
Here's an example from a busy cart that has a few mods installed:
Image 2020-04-26 at 9 22 16 PM

scottcwilson:
Sure!   My screen looks similar - I capped the size of Status so those chatty statuses wrap.

drbyte:
As you hinted in another discussion, adding an icon to display the customer's country may be useful, and would suit right beside the customer's name (feels most semantic there, much like your pay/ship column suggestion is so close to that too).

At the very least, having the country pulled in the query is beneficial.

torvista:
I need and use a country column
Untitled-1

The country flag I added to indicate the language the order was placed in, before I added code to store that info and send the emails in that language automatically (I need to remove it now).

As indicated elsewhere, it seems that most shops are single language and many single country, so display options like extra columns/extra "info" like flags should be ...optional.

Looking at the busy screenshot, it highlights how Customer Comments takes up an unwarranted amount of space.

zcwilt:

scottcwilson:
Should we close this PR and just do it more directly (create a new column and show the flag)?

drbyte:
The 3-char ISO code is unique and specific.

Let's not bother with flags right now. ie: not wishing to include copyrighted icon images in the distro

PR #3355, opened 2020-04-27
Installer: separate files for language info data

zcwilt:
Use separate files for language info data to allow for easier adding of new languages

### DIFF
--- a/zc_install/includes/classes/LanguageManager.php
+++ b/zc_install/includes/classes/LanguageManager.php
@@ -14,8 +14,12 @@ class LanguageManager
 
     public function getLanguagesInstalled()
     {
-        $this->languagesInstalled = require(DIR_FS_INSTALL . $this->langPath . 'languages_installed.php');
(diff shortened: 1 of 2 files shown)


zcwilt:
Note: 

The language info file need only be named `lng_info*.php` The actual file name is not used in code.

torvista:
This isn't working for me (no langauge dropdown).
This is not returning the array.
$languagesInstalled = $languageManager->getLanguagesInstalled();

torvista:
With the array added, the dropdown displays.
```
            $this->languagesInstalled = array_merge($this->languagesInstalled, $infoData);
        }
        return $this->languagesInstalled;
    }

```
But I can't get past the System Setup page (Continue has no effect).

torvista:
In the next couple of days I should have the installer text for spanish back from being reviewed.
Is the intention to include installer texts in the distribution or with the language pack?

PR #3359, opened 2020-04-28
Exit foreach upon result found

mc12345678:
While looking over code, came across this loop where a value is
set and it appears that there is no further need to loop.
This breaks out of the foreach upon result discovery.

### DIFF
--- a/includes/functions/functions_general.php
+++ b/includes/functions/functions_general.php
@@ -1044,14 +1044,13 @@
     global $db;
     $sql = "SELECT * FROM " . TABLE_GET_TERMS_TO_FILTER;
     $query_result = $db->Execute($sql);
-    $retVal = false;
+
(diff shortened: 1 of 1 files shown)


drbyte:
If we're gonna touch this function, perhaps the use of `$retVal` can be deleted, and just return true as soon as a match is found?

mc12345678:
Had considered it, but didn't want to grossly change it if someone had some reason to override/modify...  That said, I'll propose it anyways, there appear to be many that are at least looking over these and providing some sort of comment. :)

drbyte:
Don't need to check EOF. The foreach immediately aborts if it's empty.

mc12345678:
Thought I remembered seeing/finding a situation where `false` would get returned instead of an empty array so went overboard with validation... :) Corrected by force pushing that particular commit.

PR #3363, opened 2020-04-28
empty() should not be used for defined constants

scottcwilson:
empty is for variables, not defined constants.


drbyte:
I don't see what this code change has to do with the explanation.

Calling `empty` on a defined constant checks suitably for falsy values. Why does this code have anything to do with the one-page-checkout mod?

scottcwilson:
argh! Sorry pasted in the wrong log.  The one I had showed a notice in this code.

drbyte:
```
echo empty(FOO);
// PHP Warning:  Use of undefined constant FOO - assumed 'FOO' (this will throw an Error in a future version of PHP)

define('FOO', 0);
echo empty(FOO);
// true

define('FOO', '');
echo empty(FOO);
// true

define('FOO', 'bar);
echo empty(FOO);
// false
```

scottcwilson:
This seems all wrong; please disregard.

PR #3364, opened 2020-04-28
Default for "Save to file on server?" should be true

scottcwilson:
You want to archive these records and then purge them.  This change saves you a click.


drbyte:
They should be downloaded to local, not saved on the server.

barco57:
I would prefer, if anything about admin activity logs management is changed, to have "all logged data;" and "export to CSV" be the default. Rarely have I seen the client ever open them, I suspect most of them just delete after it downloads to them, or they tick the box to save on the server and then forget about them. I find them very useful when I am tracking a possible admin hack but only when they are "all logged data;" and "export to CSV". "export to HTML" never seems to get me everything, the logs seem truncated for some reason. if they are saved on the server I do stand a chance on a busy admin site that there will be some admin activity logs available to me. The default should always be "log more" and in a format, I can easily manipulate. With where they get stored as a secondary issue

PR #3365, opened 2020-04-28
Remove admin "categories" menu entry

dbltoe:
This line creates an unused admin_page setting that is not seen without the MOD List mod installed.


scottcwilson:
Add removal to upgrade script as well?

scottcwilson:
Hang on - @Zen4All says this was intentional.

Zen4All:
Just in time :smiley: . Like explained on the forum, It should stay. The entry is used to control the acces via admin profiles. That is the reason it is normally hidden form the menu, like the product page entry

PR #3369, opened 2020-04-29
Coupon admin page shows empty brackets in sidebar if no coupons

scottcwilson:
coupon_admin_when_empty

### DIFF
--- a/admin/coupon_admin.php
+++ b/admin/coupon_admin.php
@@ -1408,7 +1408,11 @@ function check_form(form_name) {
       $contents[] = array('text' => '<br />' . COUPON_USES_COUPON . '<br />' . zen_draw_input_field('voucher_number_of'));
       break;
     default:
-      $heading[] = array('text'=>'['.$cInfo->coupon_id.']  '.$cInfo->coupon_code . ($cInfo->coupon_id == '' ? ' - (' . (!empty($_GET['cid']) ? $_GET['cid'] : 0) . ')' : ''));
+      if ($cc_list->RecordCount() > 0) {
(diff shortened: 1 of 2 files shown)


admin/coupon_admin.php
```diff
@@ -1408,7 +1408,11 @@ function check_form(form_name) {
       $contents[] = array('text' => '<br />' . COUPON_USES_COUPON . '<br />' . zen_draw_input_field('voucher_number_of'));
       break;
     default:
-      $heading[] = array('text'=>'['.$cInfo->coupon_id.']  '.$cInfo->coupon_code . ($cInfo->coupon_id == '' ? ' - (' . (!empty($_GET['cid']) ? $_GET['cid'] : 0) . ')' : ''));
+      if (!$cc_list->EOF) {
```
mc12345678, line 1411: Shouldn't this check against `$cc_list->RecordCount() != 0` or some other comparison instead?

If followed the code correctly, when there are items in `$cc_list` and are processing through the `default:` action, that the list is stepped through using `while (!$cc_list->EOF) { … $cc_list->MoveNext(); }` and therefore after going through all of the items in the list that `$cc_list->EOF === true`...  I haven't run it, but that's what it looks like on the surface...

mc12345678:
Yup, sure did. Thank you for applying some context.  When I made the previous change, I wasn't sure which way it should next go.  Got rid of the strict notice, but left an unfinished result.  Thank you for picking something that may get incorporated.

admin/coupon_admin.php
scottcwilson: Good catch!  TY.

PR #3372, opened 2020-04-29
Store Manager CSS buttons

proseLA:
bootstrap buttons over the zen_image functions...

### DIFF
--- a/admin/store_manager.php
+++ b/admin/store_manager.php
@@ -259,7 +259,7 @@ if ($processing_message != '') {
         <td colspan="2"><br /><table border="0" cellspacing="0" cellpadding="2">
           <tr>
             <td class="main" align="left" valign="top"><?php echo TEXT_INFO_ATTRIBUTES_FEATURES_UPDATES; ?></td>
-            <td class="main" align="right" valign="middle"><?php echo zen_draw_form('update_all_products_attributes_sort_order', FILENAME_STORE_MANAGER, 'action=update_all_products_attributes_sort_order')?><input type="hidden" name="confirm" value="yes" /><?php echo zen_image_submit('button_update.gif', IMAGE_UPDATE); ?></form></td>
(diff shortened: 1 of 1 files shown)


admin/store_manager.php
```diff
@@ -259,7 +259,7 @@ function init()
         <td colspan="2"><br /><table border="0" cellspacing="0" cellpadding="2">
           <tr>
             <td class="main" align="left" valign="top"><?php echo TEXT_INFO_ATTRIBUTES_FEATURES_UPDATES; ?></td>
-            <td class="main" align="right" valign="middle"><?php echo zen_draw_form('update_all_products_attributes_sort_order', FILENAME_STORE_MANAGER, 'action=update_all_products_attributes_sort_order')?><input type="hidden" name="confirm" value="yes" /><?php echo zen_image_submit('button_update.gif', IMAGE_UPDATE); ?></form></td>
+            <td class="main" align="right" valign="middle"><?php echo zen_draw_form('update_all_products_attributes_sort_order', FILENAME_STORE_MANAGER, 'action=update_all_products_attributes_sort_order')?><input type="hidden" name="confirm" value="yes" /><input class="btn btn-default btn-sm" type="submit" value="Update"></form></td>
```
drbyte, line 262: Probably should use/add language defines for the button text.

scottcwilson:
In 1.5.8, @Zen4All will do a sweep of the remaining old buttons, so we'll get this covered soon.

proseLA:
so, would you prefer we wait, and for me to close this PR?  the issue you reference does not cover this file, as it currently uses the `zen_image_submit` function as opposed to the `zen_image_button`.

drbyte:
I'm fine with this one being implemented in 1.5.7.

proseLA:
ok.  i think this is what we need (and what you were talking about).

i am hopeful that you will soon come around to my way of thinking on `<?= `.  it is NOT a short tag and does not require short_tags on.  it makes readability much better over `<?php echo `. especially when interspersed within html.

drbyte:
> i am hopeful that you will soon come around to my way of thinking on `<?= `. it is NOT a short tag and does not require short_tags on. it makes readability much better over `<?php echo `. especially when interspersed within html.

Point taken. Fully understood. Probably not going to happen.

PR #3373, opened 2020-04-29
Additional developer customization of defaults in zc_install UI

### LINKED ISSUE #3354: Installer - change default db name in developer mode?
While preparing a minor PR regarding DEVELOPER_MODE....
I seem to be doing many repeated installs for testing, and at the moment, if no configure is defined/its a new install, the suggested name is zencart.
While thatmay be fine for a user, that's not informative enough for me, already having multiple installs in a db, and so was thinking that if in developer mode, it would be better to include the version, as shown below?
Clipboard01

drbyte:
Controlled by `/zc_install/includes/localConfig.php`

Closes #3354

### DIFF
--- a/zc_install/includes/application_top.php
+++ b/zc_install/includes/application_top.php
@@ -10,13 +10,16 @@
 @ini_set("arg_separator.output", "&");
 @set_time_limit(250);
 
+// define the project version
+require (DIR_FS_INSTALL . 'includes/version.php');
+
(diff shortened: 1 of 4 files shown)


torvista:
Is there a reason why the remaining user-entered values (admin user, email) could not be also auto-included to make this a completely mindless install process...for repeated install testing?

PR #3374, opened 2020-04-29
add language in use to the order details

torvista:


scottcwilson:
Weird - I wonder why this wasn't already being done, since the field is in the database.

torvista:
Guilty on both counts.

scottcwilson:
If the data does not fit, you must acquit!

PR #3376, opened 2020-04-29
english.php, both

torvista:
Capitalisation of the subjects in the item listing in prev-next
other minor stuff as per commit listing

### DIFF
--- a/admin/includes/languages/english.php
+++ b/admin/includes/languages/english.php
@@ -356,30 +356,30 @@ define('ICON_WARNING', 'Warning');
 
 // constants for use in zen_prev_next_display function
 define('TEXT_RESULT_PAGE', 'Page %s of %d');
-define('TEXT_DISPLAY_NUMBER_OF_BANNERS', 'Displaying <b>%d</b> to <b>%d</b> (of <b>%d</b> banners)');
-define('TEXT_DISPLAY_NUMBER_OF_CATEGORIES', 'Displaying <b>%d</b> to <b>%d</b> (of <b>%d</b> categories)');
-define('TEXT_DISPLAY_NUMBER_OF_COUNTRIES', 'Displaying <b>%d</b> to <b>%d</b> (of <b>%d</b> countries)');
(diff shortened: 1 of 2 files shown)


admin/includes/languages/english.php
```diff
@@ -698,5 +698,4 @@ function zen_date_raw($date, $reverse = false) {
 ///////////////////////////////////////////////////////////
 // include additional files:
   require(DIR_WS_LANGUAGES . $_SESSION['language'] . "/" . FILENAME_EMAIL_EXTRAS);
-  include(zen_get_file_directory(DIR_FS_CATALOG_LANGUAGES . $_SESSION['language'] . '/', FILENAME_OTHER_IMAGES_NAMES, 'false'));
-  
\ No newline at end of file
+  include(zen_get_file_directory(DIR_FS_CATALOG_LANGUAGES . $_SESSION['language'] . '/', FILENAME_OTHER_IMAGES_NAMES));
```
mc12345678, line 701: I thought that somewhere in "here" (GitHub) that I had commented/asked about this use of the default value in the call to the function and had received a response indicating that it was desirable to maintain the `'false'` statement... I can't seem to find that (or perhaps I'm thinking of another similar usage). So I ask what is the background in removing it from here? Further, is a similar change going to be pushed (pulled? :)) to all such uses?

torvista: No concrete reason, just as it is the default.

admin/includes/languages/english.php
```diff
@@ -20,7 +20,7 @@
 define('HEADER_LOGO_IMAGE', 'logo.gif');
 
 // look in your $PATH_LOCALE/locale directory for available locales..
-$locales = array('en_US', 'en_US.utf8', 'en', 'English_United States.1252');
```
scottcwilson, line 23: why are we changing array() to [] ?

torvista: I've been changing to the short form on files I modify, I understand it's preferred.

scottcwilson: ^^^^ @zcwilt @drbyte please confirm and I will update the docs.

zcwilt: definitely prefer the short form.

PR #3377, opened 2020-04-30
Orders screen with clickable number

### LINKED ISSUE #3304: Bug admin, orders. Click on order listing does not update infobox
as per subject.

scottcwilson:
Fixes #3304.  Here's one possible solution 
- eliminate the customer comments title to save space
- make order number a clickable button with a min width so new stores don't have tiny buttons.
- move billing /shipping mismatch under billing/shipping column.

orders


scottcwilson:
Another option would be for the button text to be 

Details
order-number

which would make the button taller but slimmer. 

And of course another option would be to move the shipping-billing difference to the column where the order comments icon is shown.

PR #3378, opened 2020-04-30
Lightweight solution to clickable order

scottcwilson:
Another possible fix to #3304.  Super easy to copy to customers and products listings if desired.


scottcwilson:
Oh I see that this is basically the code that @drbyte wanted to remove.  I would vote for "ugly HTML that works" over what we have now in 1.5.7 though.

PR #3382, opened 2020-04-30
Make Ask a Question a button in Responsive Classic

scottcwilson:


zcwilt:
given long term we want to stop using zen_image_button is this pr necessary.

scottcwilson:
Responsive classic uses CSS buttons so I would say yes.

drbyte:
Is there a reason this wasn't also done in template_default?

scottcwilson:
Just to show two possible variants.

drbyte:
Fair. 

If someone turns off the CSS buttons then they'll get broken-image errors in the HTML because the image file doesn't physically exist, right?

scottcwilson:
No the image does exist.  It's just ugly. :) 

includes/templates/template_default/buttons/english/button_ask_a_question.gif

drbyte:
Hmmm ... Ian and I both looked and came up empty. But I just looked again, and there it is. One of those days :)

PR #3383, opened 2020-04-30
Allow the use of array type defines files in admin

zcwilt:
Using Arrays for Admin Language Files.

We recently changed the way language defines work in the installer.

Rather than have a file that looks like

```
<?php
define('SOME_TEXT', 'Some output text');

```

We now have a file that looks like

 

```
<?php
 return [
     'SOME_TEXT' => 'Some output text',
     ];
 

```

In both cases doing

> echo SOME_TEXT will produce

Some output text

One of the problems of using defines is that once a define is declared it cannot be redefined

So we can't do

> define('SOME_TEXT', 'Some output text');
> define('SOME_TEXT', 'Some NEW output text');

only the original define will work, and further the second define will produce a PHP warning.

With using arrays we can load overrides with each level of override replacing array entries without errors before finally
turning the array entries into actual defines.
This is even more important for contribution/plugins where we want to override core language defines.

In the current implementation we are using a hybrid system. Hybrid in that we allow for the fact that some legacy files may not have been
yet converted to the new array structure. This allow for legacy contributions to continue to work.

The order of operation is as follows

    load array definitions but don't define them
    load legacy files if there is not an array file for those definitions
    define language entries from array files if not defined elsewhere

This should give a system that does not produce php warnings for double defines, but also allows for a staged move from legacy files to new
array files.

Although I'm not convinced this still won't negatively affect contributions.
Lots of contributions will redefine language defines in core code language files simply to get around the double define problem.
So changes could affect those contributions.


zcwilt:
I don't expect this to go into v157, and will leave it open only for a few days before marking as `on hold` and deferring to v158.

torvista:
How can an IDE find these definitions?
https://www.zen-cart.com/showthread.php?224230-Using-PhpStorm&p=1368139#post1368139

zcwilt:
Off the top of my head don't know. However Laravel uses a similar array structure for their language defines. So may have something in their ide-helper. Will investigate.

scottcwilson:
It would be great to accompany this with a global cleanup of duplicated defines in 1.5.8.

zcwilt:
On further investigation the laravel ide helper https://github.com/barryvdh/laravel-ide-helper
doesn't provide any help for hinting language file entries, and is mainly for autocompletion of class/facade aliases. 

There is a laravel plugin for phpstorm that does do hinting for their language/config entries, suggesting that the only way to do this for Zen Cart is to write a similar plugin. 

One thing to note though. The way the admin system works is to prefer the new `lang.filename.php` over the legacy file. This means that at least for some time the legacy file can be left in place and as such would still provide the define hinting. This is of course not a long term solution.

torvista:
Maybe an expansion of this to scan and include these would be viable:
https://github.com/torvista/zen-cart_list-configuration-constants

drbyte:
Note: the same powerful IDE has a powerful search feature that instantly finds all occurrences of the string searched for.

torvista:
Yes, but that is not a replacement for a real-time inspection error that flags up a misspelled/missing  constant or can auto-fill/hint while typing a constant.

zcwilt:
So it would be still be possible to use generated files, like @torvista posted 
https://github.com/torvista/zen-cart_list-configuration-constants
to at least provide some hinting. 
I have explored writing a phpstorm plugin, but while possible is a big separate effort. 

I'm currently working on some command line stuff, initially to help generate new php arrays from define files, which could incorporate creating `idehelper` files. 

See https://github.com/zencart/zencart/compare/v157...zcwilt:zc-command-line?expand=1

I'll post more about this soon.

zcwilt:
Going to close this as I have a new branch that replicates this but also addresses catalog side language loading as well.

PR #3385, opened 2020-05-01
v2 fix fatal error: remove call to missing function

torvista:
fatal error on option values, option names, attributes controller

### DIFF
--- a/admin/attributes_controller.php
+++ b/admin/attributes_controller.php
@@ -1057,7 +1057,7 @@ if ($action == '') {
             </td>
           </tr>
           <tr>
-            <td colspan="10"><?php echo zen_black_line(); ?></td>
+            <td colspan="10"><hr></td>
           </tr>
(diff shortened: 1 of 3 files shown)


lat9:
Was that `zen_black_line` function _really_ removed?  That will wreck havoc with a fair number of plugins, I think.

scottcwilson:
This was supposed to be a V2 change.  

Steve, this change REALLY isn't needed, the current develop branch is going to be orphaned and kept for historical purposes only.  You don't need to pursue this.

lat9:
My bad; I see that the function's still present in the zc157 branch.

torvista:
I understand this branch is to be retired off, and will only be for reference, but this is the only thing that actually stops it from working. I certainly don't intend to do anything else.

drbyte:
Closing.
@zcwilt feel free to reopen and merge if you wish.

PR #3386, opened 2020-05-01
installer language: add return array

torvista:
fixes non-display of language select drop down
https://github.com/zencart/zencart/pull/3355#issuecomment-621050003


zcwilt:
Hmm. Thought I tested with a dummy language. Thanks for the catch.

PR #3390, opened 2020-05-01
admin: orders.php correct casting to integer

torvista:
if post was set, $customer_notified was cast to integer, if not, it was set to '0' as string.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -215,7 +215,7 @@ if (zen_not_null($action) && $order_exists == true) {
       }
 
       $email_include_message = (isset($_POST['notify_comments']) && $_POST['notify_comments'] == 'on');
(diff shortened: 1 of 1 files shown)


admin/orders.php
```diff
@@ -215,7 +215,7 @@
       }
 
       $email_include_message = (isset($_POST['notify_comments']) && $_POST['notify_comments'] == 'on');
-      $customer_notified = (int)(isset($_POST['notify'])) ? $_POST['notify'] : '0';
+      $customer_notified = (int)(isset($_POST['notify']) ? $_POST['notify'] : '0');
```
drbyte, line 218: Thanks for the fix.

Please adjust it to move the `(int)` next to `$_POST`, and change the `'0'` to `0`.

torvista: it was a toss up...

PR #3392, opened 2020-05-01
Ensure whos online widget and page use same timing calculation

scottcwilson:
who's online results in dashboard widget differ from who's online page.

### DIFF
--- a/admin/includes/languages/english/whos_online.php
+++ b/admin/includes/languages/english/whos_online.php
@@ -40,7 +40,8 @@ define('TEXT_WHOS_ONLINE_FILTER_SPIDERS', 'Exclude Spiders?');
 define('TEXT_WHOS_ONLINE_FILTER_ADMINS', 'Exclude Admin IP Addresses?');
 
 define('WHOIS_TIMER_REMOVE', 1200); // seconds when removed from whos_online table - 1200 default = 20 minutes
-define('WHOIS_TIMER_INACTIVE', 180); // seconds when considered inactive - 180 default = 3 minutes
+// moved to english.php 
+// define('WHOIS_TIMER_INACTIVE', 180); // seconds when considered inactive - 180 default = 3 minutes
 define('WHOIS_TIMER_DEAD', 540); // seconds when considered dead - 540 default = 9 minutes
 define('WHOIS_SHOW_HOST', '1'); // show Last Clicked time and host name - 1 default
 define('WHOIS_REPEAT_LEGEND_BOTTOM', '12'); // show legend on bottom when more than how many entries - 12 default
--- a/admin/includes/languages/english.php
+++ b/admin/includes/languages/english.php
@@ -687,6 +687,8 @@ define('BOX_HEADING_PRODUCT_TYPES', 'Product Types');
 
 define('ERROR_DATABASE_MAINTENANCE_NEEDED', '<a href="https://docs.zen-cart.com/user/troubleshooting/error_71_maintenance_required/" target="_blank">ERROR 0071 There appears to be a problem with the database. Maintenance is required.</a>');
 
+// moved from whos online file 
+define('WHOIS_TIMER_INACTIVE', 180); // seconds when considered inactive - 180 default = 3 minutes
 // moved from currencies file:
 define('TEXT_INFO_CURRENCY_UPDATED', 'The exchange rate for %s (%s) was updated successfully to %s via %s.');
 define('ERROR_CURRENCY_INVALID', 'Error: The exchange rate for %s (%s) was not updated via %s. Is it a valid currency code?');
--- a/admin/index_dashboard.php
+++ b/admin/index_dashboard.php
@@ -77,17 +77,16 @@ for ($i = 0, $salesData = ''; $i < $report->size; $i++) {
 }
 
 // Build the whos-online graph
-$whos_online = $db->Execute("SELECT customer_id, full_name, ip_address, time_entry, time_last_click, last_page_url, session_id, host_address, user_agent FROM " . TABLE_WHOS_ONLINE);
+$whos_online = $db->ExecuteNoCache("SELECT customer_id, full_name, ip_address, time_entry, time_last_click, last_page_url, session_id, host_address, user_agent FROM " . TABLE_WHOS_ONLINE);
 // Initialize array variables for display.
 $user_array = $guest_array = $spider_array = [0 => 0, 1 => 0, 2 => 0, 3 => 0];
 $status = 0;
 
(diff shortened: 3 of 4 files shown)


drbyte:
NOTE: it is helpful when both the "commit title" (ie: first line of text in your actual commit message) and your PR title contain context-relevant descriptions.
In this case there's no context that this has something to do with the Admin, nor that it is specific to Whos-online.

More useful (again, for both the commit and the PR title) would be something along the lines of 
`Update whos-online limit display in admin dashboard`

scottcwilson:
PLEASE HOLD on this PR.  More fixes coming in this area.

scottcwilson:
Issue symptom: dashboard would show pink for a customer where the page would show red.

drbyte:
> Issue symptom: dashboard would show pink for a customer where the page would show red.

?

scottcwilson:
Open admin dashboard in one tab, admin whos online in a second, and browser in a third.  Who's online will show browser session as red, but dashboard will show it as pink.  (Assumption is you are not filtering out Admin IP Addresses.)

drbyte:
I'm going to continue to be pedantic, to make my point:

> Open admin dashboard in one tab, admin whos online in a second, and browser in a third. Who's online will show browser session as red, but dashboard will show it as pink.

... and this can be boiled down to a 60-character description which explains the original problem in a meaningful way to someone looking at it 6 months from now, by restating it as ... ________ ?
(both in the PR title and in the commit title (force-pushed to rewrite the git log correctly))

scottcwilson:
These were repro instructions, not a brief summary.  I thought that's what you wanted.  If you can be a bit more explicit than "?" I might be able to get it right faster.

admin/includes/languages/english/whos_online.php
```diff
@@ -40,7 +40,8 @@
 define('TEXT_WHOS_ONLINE_FILTER_ADMINS', 'Exclude Admin IP Addresses?');
 
 define('WHOIS_TIMER_REMOVE', 1200); // seconds when removed from whos_online table - 1200 default = 20 minutes
-define('WHOIS_TIMER_INACTIVE', 180); // seconds when considered inactive - 180 default = 3 minutes
+// moved to english.php 
+// define('WHOIS_TIMER_INACTIVE', 180); // seconds when considered inactive - 180 default = 3 minutes
```
drbyte, line 44: Not loving this change. Breaks up the context.

Sure, to use the same constant elsewhere it must be shared, but that makes me wonder how important this even is ...

scottcwilson: Could move all 3 to english.php Without this, the 2 pages are out of sync - wouldn't you call that a confusing bug?  One of my customers complained as soon as I put 1.5.6 on their site.

drbyte:
Title is better. Thanks.

drbyte:
The changes you've made here are multifold:
- changed queries to avoid any caching, which slows things down
- changed time threshold on dashboard
- changed location of constant to couple the dashboard with the whosonline page

I'm guessing the root cause here is that you'd changed the threshold in the constant but then you wanted the dashboard to use the same number? 

Or was the root cause that the queries were cached and you didn't like that?

admin/includes/languages/english/whos_online.php
drbyte: > Could move all 3 to english.php

Sigh. Yes we certainly could. 

But then everyone's gonna complain that these are not language-related at all, and complain that we're cluttering english.php, and want more configuration switches that are already too cumbersome ...

scottcwilson:
Root cause was dashboard should use same number. 
Noticed the queries were cached and thought this could also be misleading.

admin/includes/languages/english/whos_online.php
scottcwilson: extra_configures?  Put in db?  What would you like to do, Chris?

scottcwilson:
Ugh still seeing wildly different numbers on dashboard vs whos online page.  Have you tested this on a busy store?

drbyte:
Questions:
- in what ways do people commonly customize the admin whos-online page? 
- are there any mods that alter it? In what way?

Or is it largely untouched by everyone?

scottcwilson:
I would guess it is untouched.

wiztechinc:
The only thing I’ve ever done comes with the cowoa mod. Though I actually wish some of it would go away.  So if I could, I would pick and choose what I wanted my clients to see.

drbyte:
> The only thing I’ve ever done comes with the cowoa mod. Though I actually wish some of it would go away. So if I could, I would pick and choose what I wanted my clients to see.

Cool - that's useful: Do you mind elaborating about what you'd like to hide/show etc?

wiztechinc:
The only thing I personally ever look at is new customers and new orders. The ones I would like to be able to hide from my clients are stats, who's online (I actually detest it in general because I've had clients who get hung up on watching it and obsessing over each entry.), visitor history, orders at the bottom left is barely noticeable down there so some might want that to be more prominent. specials - some never run specials. 

In other words each cart owner is different, including me. The idea of being able to move the widgets or turn them off was one of the greatest ideas I've heard and I'm very disappointed this hasn't been implimented.

drbyte:
Okay. Thanks for clarifying that you only meant you prefer to remove it because you think people misuse it.

Unfortunately that's not what I was asking.

drbyte:
Aside: I looked into the various implementations of COWOA and I don't see any ways in which it alters whosonline operation. If I'm missing something in that regard, please show me.

wiztechinc:
> Aside: I looked into the various implementations of COWOA and I don't see any ways in which it alters whosonline operation. If I'm missing something in that regard, please show me.

nope it doesn't - I was talking about changes to the dashboard - and not who's online. Sorry! I guess I wasn't following as closely as I thought I was. 

But I don't feel like it's something that should be on the dashboard at all.

zcwilt:
I guess this is where people could complain about our lack of a `roadmap`
However we did move to using a new system for dashboard widgets in the original v2 branch and 
that would still be what we want to move to. Given that I'd like to defer this to v158 and until we can think about roadmap again.

scottcwilson:
My vote would be to remove this from the default dashboard entirely and close the issue.  The presence of this widget creates a "two sources of truth" problem which drives nervous shopowners crazy as the obsess over the differences between these dashboard and the whos online page.

drbyte:
While I agree that we can improve the dashboard to make it easier to turn things off,
this is the first time I've heard anyone complain about having the widget on the dashboard. Are there a bunch of forum discussions about it?

wiztechinc:
probably not - it's the developers who pay the price

PR #3393, opened 2020-05-02
Hide some fields in admin>orders>order details on phones

scottcwilson:
On the order details screen, when using a phone, hide these fields:
- all tax fields and model from the order table 
- customer notified and updated_by from the history table

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -675,15 +675,15 @@ if (zen_not_null($action) && $order_exists == true) {
           <table class="table">
             <tr class="dataTableHeadingRow">
               <th class="dataTableHeadingContent" colspan="2"><?php echo TABLE_HEADING_PRODUCTS; ?></th>
-              <th class="dataTableHeadingContent"><?php echo TABLE_HEADING_PRODUCTS_MODEL; ?></th>
-              <th class="dataTableHeadingContent text-right"><?php echo TABLE_HEADING_TAX; ?></th>
+              <th class="dataTableHeadingContent hidden-xs"><?php echo TABLE_HEADING_PRODUCTS_MODEL; ?></th>
+              <th class="dataTableHeadingContent text-right hidden-xs"><?php echo TABLE_HEADING_TAX; ?></th>
               <th class="dataTableHeadingContent text-right"><?php echo ($show_including_tax) ? TABLE_HEADING_PRICE_EXCLUDING_TAX : TABLE_HEADING_PRICE; ?></th>
 <?php if ($show_including_tax)  { ?>
-              <th class="dataTableHeadingContent text-right"><?php echo TABLE_HEADING_PRICE_INCLUDING_TAX; ?></th>
+              <th class="dataTableHeadingContent text-right hidden-xs"><?php echo TABLE_HEADING_PRICE_INCLUDING_TAX; ?></th>
 <?php } ?>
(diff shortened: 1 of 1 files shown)


drbyte:
I can see the value of hiding the tax-related columns because those aren't material to the order-fulfillment process. (But they'd be important if one was making live edits to the order.)

However, the product model number is probably relevant for many stores. I'm not sure everyone would be happy with that exclusion.

scottcwilson:
What about this: provide products model in a span underneath the name but with style .hidden-sm-up (i.e. it only displays on xs)?

torvista:
>the product model number is probably relevant for many stores
For me, model number is absolutely non-negotiable with 100's of products with very similar names. So, all I touch has this in mind, from a self-interested point of view.

But, that's just me... I can't say for what proportion of users the model number is just wasted space, but if it is a "significant" %, perhaps it should be a global option "show model number" that would remove all display of it, columns etc. to have some sort of continuity between admin screens.

scottcwilson:
I get it - but those product names probably also won't fit in a field as narrow as the current product name field on the admin orders screen.  So help me figure out a compromise, please. 

Bear in mind that this is on a phone only.  You can see everything if you go to your desktop computer or even your tablet.  To make it workable on a phone, we need to cut some things. 

What did you think of the suggestion of putting the model under the name on a mobile phone?

scottcwilson:
With no changes, here's what it looks like: 
Screen Shot 2020-05-05 at 5 04 23 PM

scottcwilson:
And here's what it would look like with this PR: 
Screen Shot 2020-05-05 at 5 12 40 PM

torvista:
I think if you are looking at the order DETAILS page it's because you want the details. All of them.
Remember this is an Admin screen, so pretty is second to content.
As it is now it compresses well, the customer personal details take up far too much vertical space, but that's Bootstrap for you.
I wouldn't change anything, but then again I wouldn't consider managing my store on a phone either!

torvista:
....I think the layout is wrong anyway, the product identification should come first (model-name) then the quantity etc..

admin/orders.php
```diff
@@ -718,27 +718,31 @@ function couponpopupWindow(url) {
                         echo '</i></small></span>';
                       }
                     }
+                    // Mobile phones only
+                    echo '<span class="hidden-sm-up">';
```
drbyte, line 722: `hidden-sm-up` - Is that official? or accidentally works?  I don't see it in the docs

scottcwilson: Most official. 
https://v4-alpha.getbootstrap.com/layout/responsive-utilities/

drbyte: But ... we're currently using BS 3, not 4
Plus, the v4-alpha docs on that are quite different from the official v4 docs

scottcwilson: Two weeks ago when I did this, I could have sworn I saw this in the CSS.  Was it the whiskey?  Maybe.

barco57: Should be .hidden-sm for BS 3.4.1

barco57: https://getbootstrap.com/docs/3.4/css/#responsive-utilities

scottcwilson: No it's the opposite.  `hidden-sm-up` means only show on xs.

drbyte: I think all BS styles work that way: All apply from "specified size, and larger, until a larger-size rule is provided"

drbyte: Nope, not with .hidden-*

drbyte: Looks like `visible-xs-inline` is what's needed here.

PR #3397, opened 2020-05-02
Make Admin category/products and orders table rows clickable again

### LINKED ISSUE #3304: Bug admin, orders. Click on order listing does not update infobox

drbyte:
Fixes #3304

This uses the on-page action ideas offered by @lat9 in #3304, with some variations to use named classes for excluded columns.

Closes #3377 
Closes #3378

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -572,7 +572,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
 
           $show_prod_labels = ($search_result || $categories->EOF) ? true : false;
           ?>
-          <table class="table table-striped">
+          <table id="categories-products-table" class="table table-striped">
             <thead>
               <tr>
                 <th class="text-right shrink"><?php echo TABLE_HEADING_ID; ?></th>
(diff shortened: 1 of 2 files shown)


scottcwilson:
If approved, could we have the same logic on the products list in admin > catalog > categories/products?

scottcwilson:
Tested. Groovy.

drbyte:
added category/product listing

torvista:
Well, I've been trying to find why the html validator is complaining about an unclosed attribute ", but only when there is more than one product in the order. Something to do with the hr and br, but not of importance.

Then...takes deep breath...I realised I won't be using this popup at all, as it is.
It adds a column/wastes space, it takes a click to show it, a click to close it, the closing click will be on another line causing an unwanted refresh of the infobox and it is enormous (optional I know).



It the risk of repeating myself, I did similar but as popup on hover on the price.
No space wasted, all the info instantly, no obscuring of other useful info,  and no jumping through hoops to make it clickable. Obviously I have never formatted it much.
Compare how much easier this is to read, and how many orders it does not obscure.

mv

(Size is an issue I have with bootstrap in general. It is infantile for customers, but not appropriate for an admin interface, far too much space wasted. I use a css file to squash it all as much as possible.)

scottcwilson:
@torvista The popup is not related to this issue.  Can you move your complaint to a new issue?

torvista:
I understood that the row has become not-clickable due to the additional of this popop, so that appears relevant to me.

scottcwilson:
Chris has been yanking out this functionality because he feels it's ugly HTML.  See https://github.com/zencart/zencart/issues/3304#issuecomment-621409122

That's why it was removed from the categories/products listing in 1.5.6.   

Your concern is valid, but it's separate, so it would help if you moved it from a PR comment to a new issue.

Issue #3375, opened 2020-04-29
Send order status emails in order language


torvista:
I did my own work to implement this ages ago and was revisiting it today to use the new field in the ZC157 database, with the idea of presenting it asap, since things seem to be hotting up.
However, I don't consider my old workaround adequate for core code.

I note that work was done on 1.6/develop to enable this and was intending to look over that to port it over to my site/ZC157. Possibly too late now but since Zen4all did it, perhaps he can orientate me if it's feasible now or requires 1.6 structure/too different to be easily done?

scottcwilson:
I believe this is the relevant PR.  #1058

torvista:
Yes that was it.

scottcwilson:
This should be a 1.5.8 change, so there's no rush.

torvista:
added by 
https://github.com/zencart/zencart/pull/3399

PR #3399, opened 2020-05-02
Admin - order status update - send email in order language

torvista:
I've been using a dirtier version of this for many moons, but based on language id, not code.
So while I've been pre-update modifying it to use language_code, I decided to put it out for persusal, while it's fresh in my mind. Maybe it's not ideal, but it works for me.

Two birds killed with this one, the infobox is modified to allow sending emails directly from it. Stolen from Quick Status Update long ago.

The form action includes the language parameter, and so 
if the admin language is different to the order language, the admin language is changed to the order language for the update/email sending, and restored back to the admin language on the redirect.
Single language sites should see nothing apart from the email option in the infoBox.

I modified the language lookups to allow lookup by code in addition to by id.

Also the email language is indicated by the comments box on the details page to remind the admin user to write in the correct language. This is surprisingly necessary for us!

### DIFF
--- a/admin/includes/languages/english/orders.php
+++ b/admin/includes/languages/english/orders.php
@@ -130,3 +130,5 @@ define('TEXT_MAP_CUSTOMER_ADDRESS', 'Map Customer Address');
 define('TEXT_MAP_SHIPPING_ADDRESS', 'Map Shipping Address');
 define('TEXT_MAP_BILLING_ADDRESS', 'Map Billing Address');
 
+define('TEXT_EMAIL_LANGUAGE', 'order language: %s');
+define('SUCCESS_EMAIL_SENT', 'Email %s sent to customer');
--- a/admin/includes/languages/english/orders.php
+++ b/admin/includes/languages/english/orders.php
@@ -130,5 +130,5 @@ define('TEXT_MAP_CUSTOMER_ADDRESS', 'Map Customer Address');
 define('TEXT_MAP_SHIPPING_ADDRESS', 'Map Shipping Address');
 define('TEXT_MAP_BILLING_ADDRESS', 'Map Billing Address');
 
-define('TEXT_EMAIL_LANGUAGE', 'order language: %s');
+define('TEXT_EMAIL_LANGUAGE', 'Order Language: %s');
 define('SUCCESS_EMAIL_SENT', 'Email %s sent to customer');
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -209,6 +209,7 @@ if (zen_not_null($action) && $order_exists == true) {
     case 'update_order':
       $oID = zen_db_prepare_input($_GET['oID']);
       $comments = zen_db_prepare_input($_POST['comments']);
+      $admin_language = zen_db_prepare_input(isset($_POST['admin_language']) ? $_POST['admin_language'] : $_SESSION['languages_code']);
       $status = (int)$_POST['status'];
       if ($status < 1) {
          break;
@@ -286,12 +287,19 @@ if (zen_not_null($action) && $order_exists == true) {
             $db->Execute($update_downloads_query);
           }
         }
-        $messageStack->add_session(SUCCESS_ORDER_UPDATED, 'success');
-        zen_record_admin_activity('Order ' . $oID . ' updated.', 'info');
+          $messageStack->add_session(SUCCESS_ORDER_UPDATED, 'success');
+          if ($customer_notified === 1) {
+              $messageStack->add_session(sprintf(SUCCESS_EMAIL_SENT, ($admin_language !== $_SESSION['languages_code'] ? '(' . strtoupper($_SESSION['languages_code']) . ') ' : '')), 'success'); // show an email sent confirmation message, with a language indicator if the order/email language was different to the admin user language
+          }
+          zen_record_admin_activity('Order ' . $oID . ' updated.', 'info');
       } else {
-        $messageStack->add_session(WARNING_ORDER_NOT_UPDATED, 'warning');
+          $messageStack->add_session(WARNING_ORDER_NOT_UPDATED, 'warning');
       }
-      zen_redirect(zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(array('action')) . 'action=edit', 'NONSSL'));
+        if (isset($_POST['qsu'])) { // return to the order LISTING page
+            zen_redirect(zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(['action', 'language']) . 'language=' . $admin_language, 'NONSSL'));//remove action, nothing more to do, reset admin language
+        } else {  // return to the order DETAILS page
+            zen_redirect(zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(['action', 'language']) . 'action=edit&language=' . $admin_language, 'NONSSL'));
+        }
       break;
     case 'deleteconfirm':
       $oID = zen_db_prepare_input($_POST['oID']);
@@ -908,11 +916,16 @@ if (zen_not_null($action) && $order_exists == true) {
         <div class="row noprint"><?php echo zen_draw_separator('pixel_trans.gif', '1', '5'); ?></div>
         <div class="row noprint">
           <div class="formArea">
-              <?php echo zen_draw_form('statusUpdate', FILENAME_ORDERS, zen_get_all_get_params(array('action')) . 'action=update_order', 'post', 'class="form-horizontal"', true); ?>
+              <?php echo zen_draw_form('statusUpdate', FILENAME_ORDERS, zen_get_all_get_params(array('action','language')) . 'action=update_order&language=' . $order->info['language_code'], 'post', 'class="form-horizontal"', true); ?>
             <div class="form-group">
                 <?php echo zen_draw_label(TABLE_HEADING_COMMENTS, 'comments', 'class="col-sm-3 control-label"'); ?>
               <div class="col-sm-9">
-                  <?php echo zen_draw_textarea_field('comments', 'soft', '60', '5', '', 'id="comments" class="form-control"'); ?>
+                  <?php echo zen_draw_textarea_field('comments', 'soft', '60', '5', '', 'id="comments" class="form-control"');
+                  // remind admin user of the order/customer language in case of writing a comment.
+                  if (count(zen_get_languages()) > 1) {
+                      echo '<br>' . zen_get_language_icon($order->info['language_code'], true) . ' <strong>' . sprintf(TEXT_EMAIL_LANGUAGE, zen_get_language_name($order->info['language_code'], true)) . '</strong>';
+                     echo zen_draw_hidden_field('admin_language', $_SESSION['languages_code']);
+                  } ?>
               </div>
             </div>
 <?php
@@ -950,30 +963,6 @@ if (zen_not_null($action) && $order_exists == true) {
             </div>
 <?php
             }
-        }
-
-        // -----
-        // Determine which of the 'Notify Customer' radio buttons should be selected initially,
-        // based on configuration setting in 'My Store'.  Set a default, just in case that configuration
-        // setting isn't set!
-        //
-        if (!defined('NOTIFY_CUSTOMER_DEFAULT')) define('NOTIFY_CUSTOMER_DEFAULT', '1');
-        switch (NOTIFY_CUSTOMER_DEFAULT) {
-            case '0':
-                $notify_email = false;
-                $notify_no_email = true;
-                $notify_hidden = false;
-                break;
-            case '-1':
-                $notify_email = false;
-                $notify_no_email = false;
-                $notify_hidden = true;
-                break;
-            default:
-                $notify_email = true;
-                $notify_no_email = false;
-                $notify_hidden = false;
-                break;
         }
 ?>
             <div class="form-group">
@@ -1132,7 +1121,7 @@ if (zen_not_null($action) && $order_exists == true) {
                       $keywords = zen_db_input(zen_db_prepare_input($_GET['search']));
                       $search = " and (o.customers_city like '%" . $keywords . "%' or o.customers_postcode like '%" . $keywords . "%' or o.date_purchased like '%" . $keywords . "%' or o.billing_name like '%" . $keywords . "%' or o.billing_company like '%" . $keywords . "%' or o.billing_street_address like '%" . $keywords . "%' or o.delivery_city like '%" . $keywords . "%' or o.delivery_postcode like '%" . $keywords . "%' or o.delivery_name like '%" . $keywords . "%' or o.delivery_company like '%" . $keywords . "%' or o.delivery_street_address like '%" . $keywords . "%' or o.billing_city like '%" . $keywords . "%' or o.billing_postcode like '%" . $keywords . "%' or o.customers_email_address like '%" . $keywords . "%' or o.customers_name like '%" . $keywords . "%' or o.customers_company like '%" . $keywords . "%' or o.customers_street_address  like '%" . $keywords . "%' or o.customers_telephone like '%" . $keywords . "%' or o.ip_address  like '%" . $keywords . "%')";
                   }
-                  $new_fields .= ", o.customers_company, o.customers_email_address, o.customers_street_address, o.delivery_company, o.delivery_name, o.delivery_street_address, o.billing_company, o.billing_name, o.billing_street_address, o.payment_module_code, o.shipping_module_code, o.ip_address ";
+                  $new_fields .= ", o.customers_company, o.customers_email_address, o.customers_street_address, o.delivery_company, o.delivery_name, o.delivery_street_address, o.billing_company, o.billing_name, o.billing_street_address, o.payment_module_code, o.shipping_module_code, o.orders_status, o.ip_address, o.language_code ";
 
                   $order_by = " ORDER BY o.orders_id DESC";
                   $zco_notifier->notify('NOTIFY_ADMIN_ORDERS_SEARCH_PARMS', $keywords, $search, $search_distinct, $new_fields, $new_table, $order_by);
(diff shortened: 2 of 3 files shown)


torvista:
some red-eye mis-merging in this

scottcwilson:
Fixes #3375 

Thanks for doing this - this is a nice improvement to the product.

torvista:
>I modified the language lookups to allow lookup by code in addition to by id
How about accepting either (id or code), then checking if 
$lookup == (int)$lookup or 
is in-array language code 
i.e. figuring out what the lookup is and treating it as such?

scottcwilson:
My vote on this would be to have two functions.

torvista:
I have already modified two language functions, I would not want to make more just for this.

In fact with the tortoise hindsight I have, I would prefer to merge
zen_get_language_icon
zen_get_language_name
to
zen_get_language_info and pass two parameters to choose the info to be returned.

admin/orders.php
```diff
@@ -1326,7 +1339,19 @@ function couponpopupWindow(url) {
...
+                        '<label class="form-group" style="margin:8px 0">' . BOX_TOOLS_MAIL . zen_draw_checkbox_field('notify', '1', $notify_email, '', 'style="vertical-align: middle;"') . '</label>' . "<br>\n" .
+                        '<label class="form-group" for="status">' . HEADING_TITLE_STATUS . '</label>' . zen_draw_order_status_dropdown('status', $oInfo->orders_status, '', 'onChange="this.form.submit();" id="status"') . "\n" .
+                        '</form>' . "\n"];
+
                     $contents[] = array('text' => '<br>' . TEXT_DATE_ORDER_CREATED . ' ' . zen_date_short($oInfo->date_purchased));
+                    if ($_SESSION['languages_code'] !== $oInfo->language_code) {
+                        $contents[] = array('text' => zen_get_language_icon($oInfo->language_code, true) . ' ' . sprintf(TEXT_EMAIL_LANGUAGE, zen_get_language_name($oInfo->language_code, true)));
+}
```
drbyte, line 1354: I wonder if it makes sense to skip the `if` part here, and just always display the language?
Or, maybe change the `if` to "if the store supports more than one language, then display it"?

Thoughts?

drbyte:
I'm working on updates to the Language class to allow these other helper functions to avoid doing their own queries, but to instead just ask the class what's already been queried and stored in RAM.

In the meantime your query changes could simply keep `$lookup` as the incoming key, and then let your SQL query use an `OR` clause.
If you keep the function call limited to still passing a single lookup value (whether it's numeric or string), it'll be consistent enough with the pending class changes that no further changes should be required in the places where you call the function from.

admin/orders.php
```diff
@@ -908,11 +940,16 @@ function couponpopupWindow(url) {
...
         <div class="row noprint">
           <div class="formArea">
-              <?php echo zen_draw_form('statusUpdate', FILENAME_ORDERS, zen_get_all_get_params(array('action')) . 'action=update_order', 'post', 'class="form-horizontal"', true); ?>
+              <?php echo zen_draw_form('statusUpdate', FILENAME_ORDERS, zen_get_all_get_params(array('action','language')) . 'action=update_order&language=' . $order->info['language_code'], 'post', 'class="form-horizontal"', true); ?>
             <div class="form-group">
                 <?php echo zen_draw_label(TABLE_HEADING_COMMENTS, 'comments', 'class="col-sm-3 control-label"'); ?>
               <div class="col-sm-9">
-                  <?php echo zen_draw_textarea_field('comments', 'soft', '60', '5', '', 'id="comments" class="form-control"'); ?>
+                  <?php echo zen_draw_textarea_field('comments', 'soft', '60', '5', '', 'id="comments" class="form-control"');
+                  // remind admin user of the order/customer language in case of writing a comment.
```
drbyte, line 948: I haven't explored this, and I'm a bit reluctant to encourage rich-text in these fields, but if you add the `.editorHook` class to this textarea and have CKEditor enabled, it's possible we could also link it to the right language so that it offers spellcheck and UI in the corresponding language expected for the field (this already works for product descriptions).

admin/orders.php
```diff
@@ -1326,7 +1339,19 @@ function couponpopupWindow(url) {
...
                     $zco_notifier->notify('NOTIFY_ADMIN_ORDERS_MENU_BUTTONS', $oInfo, $contents);
 
+                    // each contents array is drawn in a div, so this form block must be a single array element.
+                    $contents[] = ['text' =>
+                        zen_draw_form('statusUpdate', FILENAME_ORDERS, zen_get_all_get_params(['action','language']) . 'action=update_order' . (!isset($_GET['oID']) ? '&oID=' . $oInfo->orders_id : '') . '&language=' . $oInfo->language_code, 'post', '', true) . // form action uses the order language to change the session language on the update. On initial page load (from another page), $_GET['oID'] is not set, hence clause in form action
+                        zen_draw_hidden_field('listing', '1') . // identify that this form was submitted from infoBox/listing page and not the order details page (for the redirect to this same page)
```
drbyte, line 1345: For `listing` I wonder if it makes sense to use `cameFrom`. This would retain semantic meaning, and be similar to what's already used with session timeouts.

admin/orders.php
```diff
@@ -1326,7 +1339,19 @@ function couponpopupWindow(url) {
                     $contents[] = array('align' => 'text-center', 'text' => '<a href="' . zen_href_link(FILENAME_ORDERS_INVOICE, 'oID=' . $oInfo->orders_id) . '" target="_blank" class="btn btn-info" role="button">' . IMAGE_ORDERS_INVOICE . '</a> <a href="' . zen_href_link(FILENAME_ORDERS_PACKINGSLIP, 'oID=' . $oInfo->orders_id) . '" target="_blank" class="btn btn-info" role="button">' . IMAGE_ORDERS_PACKINGSLIP . '</a>');
                     $zco_notifier->notify('NOTIFY_ADMIN_ORDERS_MENU_BUTTONS', $oInfo, $contents);
 
+                    // each contents array is drawn in a div, so this form block must be a single array element.
+                    $contents[] = ['text' =>
+                        zen_draw_form('statusUpdate', FILENAME_ORDERS, zen_get_all_get_params(['action','language']) . 'action=update_order' . (!isset($_GET['oID']) ? '&oID=' . $oInfo->orders_id : '') . '&language=' . $oInfo->language_code, 'post', '', true) . // form action uses the order language to change the session language on the update. On initial page load (from another page), $_GET['oID'] is not set, hence clause in form action
```
drbyte, line 1344: This inline "update the status" is a handy option. Perhaps it can be wrapped in a fieldset or some container that makes it obvious that the checkbox and dropdown go together. Maybe add a heading to it too?

I'd initially considered putting the checkbox to the right of the dropdown, but many stores add very verbose order-status names and stuffing a checkbox on the same line is just complicated under those conditions.

admin/orders.php
```diff
@@ -1326,7 +1339,19 @@ function couponpopupWindow(url) {
...
+                        zen_draw_hidden_field('listing', '1') . // identify that this form was submitted from infoBox/listing page and not the order details page (for the redirect to this same page)
+                        ($oInfo->language_code !== $_SESSION['languages_code'] ? zen_draw_hidden_field('admin_language', $_SESSION['languages_code']) : '') . // if the order language is different to the current admin language, record the admin language, to restore it in the redirect after the status update email has been sent
+                        '<label class="form-group" style="margin:8px 0">' . BOX_TOOLS_MAIL . zen_draw_checkbox_field('notify', '1', $notify_email, '', 'style="vertical-align: middle;"') . '</label>' . "<br>\n" .
```
drbyte, line 1347: I'm reluctant to use `BOX_TOOLS_MAIL` here, because that's for the menu system.

admin/orders.php
```diff
@@ -1326,7 +1339,19 @@ function couponpopupWindow(url) {
...
+                        '<label class="form-group" for="status">' . HEADING_TITLE_STATUS . '</label>' . zen_draw_order_status_dropdown('status', $oInfo->orders_status, '', 'onChange="this.form.submit();" id="status"') . "\n" .
```
drbyte, line 1348: Similarly, what's the likelihood of someone customizing the Table Heading texts in a way that might break this re-use of the same constant as an input field label?

drbyte:
@torvista This PR brings some great improvements. Thanks!

I've posted some comments, some which are suggestions for changes, and some which are more just opening dialog by asking questions about decisions used. Responses welcome.

admin/orders.php
torvista: As it is. only the listing page provides this as post, so the redirect check is simple.
This is easily changed to cameFrom (I note all instances of this are camefrom, no camel case), but with what value?
From my vague understanding of camefrom, camefrom is a GET and should be a full filename with parameters?

admin/orders.php
torvista: >checkbox on the same line is just complicated

Yes.

admin/orders.php
torvista: ok

admin/orders.php
torvista: ok

admin/orders.php
drbyte: You're right: it's not camelCase.  I wondered about it, but didn't go check.

If setting the filename and parameters works from this point in the code then it may be good to remain consistent.

One thing I'm trying to avoid is the introduction of new fields that aren't already covered by Admin Sanitization. I'll post about that in a separate comment.

drbyte:
These form additions introduce the need for some sanitization rules. 
There are 2 choices when adding forms/fields/URL params:
- either re-use the same form field names and GET params as what are already covered for the same purpose in other places,
- or add new rules to cover these added fields/params

admin/orders.php
torvista: As they are, in a single language system, it is always true: no icon, no complaining about the flags.
For multilanguage,
in the infoBox, in truth it is unnecessary info as nothing can be changed here. Back in time, before I figured out how to do this, this was a reminder to change the admin language before sending the email!
in details, it's useful for admin user, to add a comment in the correct language:
we are three, and all of us are fluent english/spanish and work in either admin language wihout noticing, so this is relevant.

drbyte: >  in truth it is unnecessary info as nothing can be changed here

Okay, then let's not add it if it's unnecessary.

admin/orders.php
torvista: I'm not getting CKeditor loading on product descriptions...

drbyte: > I'm not getting CKeditor loading on product descriptions...

Requirement: choose the CKEditor from the Editor choice before entering the product-edit
Screen Shot 2020-05-03 at 8 35 58 PM

torvista: I had done that, it didn't display the editor. I left the computer on sleep.
This morning, woke it up, the timeout showed/was logged out.  I logged in, was redirected to order details page. editor was showing.
Disabled the admin switch, editor off.
Enabled the admin switch, editor still off.
Logged off and on, editor still off.

admin/orders.php
torvista: removed

admin/orders.php
torvista: ah, so its not to "use" the value, its just to allow the parameter name since camefrom is already permitted.

torvista: Done.

torvista:
> These form additions introduce the need for some sanitization rules.
> There are 2 choices when adding forms/fields/URL params:
> 
> * either re-use the same form field names and GET params as what are already covered for the same purpose in other places,
> * or add new rules to cover these added fields/params
> 
> (And this reminds me that all the other admin page changes done in the last couple months probably need to be reviewed for that.)

If new parameters need to be added to the sanitation list,  how come they work **without** being added to the list?

admin/orders.php
torvista: as far I can see, this is missing from the head, to allow the editor
```
      <?php
      if ($action === 'edit' && $editor_handler !== '') {
          include ($editor_handler);
      }
      ?>
```

drbyte: Oh, right. Yes, that would need to be added to the Order page for the editor to be usable there.

scottcwilson: This is a nice addition.

torvista: Done. Only spellcheck options for textareas to consider now...best as another PR.

admin/includes/languages/english/orders.php
```diff
@@ -130,3 +130,5 @@
 define('TEXT_MAP_SHIPPING_ADDRESS', 'Map Shipping Address');
 define('TEXT_MAP_BILLING_ADDRESS', 'Map Billing Address');
 
+define('TEXT_EMAIL_LANGUAGE', 'order language: %s');
```
drbyte, line 133: Did you intend this to be all lowercase?

drbyte:
Question re: `admin_language` param: For sending the actual email, you're passing the language id via POST.  Why not lookup the language from the order itself, at send-time, instead of trusting the browser to reliably pass an extra field? Was there a functional reason why it needed to be made interceptable (and thus also indirectly/unintentionally tamperable)?

admin/includes/languages/english/orders.php
torvista: Since it had the icon in front, yes. But I've changed it so they are capitalised.

drbyte: Okay, the first time I'd looked at this I didn't have an icon set for the 2nd language, so I didn't have anything showing beside it. And then looking at the language file which doesn't reflect that context it seemed odd that it was all lowercase, and I figured there'd be people coming along doing translation who would complain about it, so it made sense to ask! 😊

drbyte:
> > These form additions introduce the need for some sanitization rules.
> > There are 2 choices when adding forms/fields/URL params:
> > 
> > * either re-use the same form field names and GET params as what are already covered for the same purpose in other places,
> > * or add new rules to cover these added fields/params
> > 
> > (And this reminds me that all the other admin page changes done in the last couple months probably need to be reviewed for that.)
> 
> If new parameters need to be added to the sanitation list, how come they work **without** being added to the list?

They "work" but they're aggressively sanitized.

But I feel it's best-practice to actually "list" the ones that are "intentionally supported" so that nobody has to wonder later whether it's being sanitized correctly.
Re-using the same param name for something that serves the same purpose seems sensible not only for sanitization reasons but also for general consistency.

torvista:
> Question re: admin_language param: For sending the actual email, you're passing the language id via POST. Why not lookup the language from the order itself, at send-time, instead of trusting the browser to reliably pass an extra field? Was there a functional reason why it needed to be made interceptable (and thus also indirectly/unintentionally tamperable)?

I don't see how to get the correct email language without changing the admin language on a page change: it has to be on the form submit.
Can something be done to the function that sends the email..?

drbyte:
> > Question re: admin_language param: For sending the actual email, you're passing the language id via POST. Why not lookup the language from the order itself, at send-time, instead of trusting the browser to reliably pass an extra field? Was there a functional reason why it needed to be made interceptable (and thus also indirectly/unintentionally tamperable)?
> 
> I don't see how to get the correct email language without changing the admin language on a page change: it has to be on the form submit.
> Can something be done to the function that sends the email..?

I see: it's being used to temporarily flip the session language, and then flip it back again.

No, the way language-files are loaded using `constants`, this is the only way.
Once we fully switch it over to using arrays, then we can load alternate languages anytime we want, so this sort of thing won't be necessary.

admin/orders.php
```diff
@@ -287,12 +312,20 @@
...
         zen_record_admin_activity('Order ' . $oID . ' updated.', 'info');
       } else {
         $messageStack->add_session(WARNING_ORDER_NOT_UPDATED, 'warning');
       }
-      zen_redirect(zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(array('action')) . 'action=edit', 'NONSSL'));
+
+        if (isset($_POST['camefrom']) && $_POST['camefrom'] === 'orderEdit') {
+            zen_redirect(zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(['action', 'language']) . 'action=edit' . ($admin_language !== $_SESSION['languages_code'] ? '&language=' . $admin_language : ''), 'NONSSL'));
+        }
+        zen_redirect(zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(['action', 'language']) . ($admin_language !== $_SESSION['languages_code'] ? 'language=' . $admin_language : ''), 'NONSSL'));
```
drbyte, line 326: What do you think of the following approach, which makes the `if` intention slightly clearer (ie: that the only difference is the `edit` action)?

```php
        $redirect = zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(['action', 'language']) . ($admin_language !== $_SESSION['languages_code'] ? '&language=' . $admin_language : ''), 'NONSSL');
        if (isset($_POST['camefrom']) && $_POST['camefrom'] === 'orderEdit') {
            $redirect .= '&action=edit';
         }
         zen_redirect($redirect);
```

torvista:
So, I think all issues are resolved now?

admin/orders.php
torvista: ok.

drbyte:
> So, I think all issues are resolved now?

Yes. Looks good.

admin/orders.php
```diff
@@ -908,13 +947,19 @@ function couponpopupWindow(url) {
...
+              <div class="form-group">
+                  <?php echo zen_draw_label(TABLE_HEADING_COMMENTS, 'comments', 'class="col-sm-3 control-label"'); ?>
+                  <div class="col-sm-9">
+                      <?php echo zen_draw_textarea_field('comments', 'soft', '60', '5', '', 'id="comments" class="editorHook form-control"');
+                      // remind admin user of the order/customer language in case of writing a comment.
+                      if (count(zen_get_languages()) > 1) {
+                          echo '<br>' . zen_get_language_icon($order->info['language_code']) . ' <strong>' . sprintf(TEXT_EMAIL_LANGUAGE, ucfirst(zen_get_language_name($order->info['language_code']))) . '</strong>';
+                          echo zen_draw_hidden_field('admin_language', $_SESSION['languages_code']);
+                      } ?>
+                  </div>
```
scottcwilson, line 961: This is awesome!  I missed this when I first looked at the PR.

torvista: What exactly?

scottcwilson: You added in Quick Status Update.  Before this you had to open the order to change the status.

PR #3400, opened 2020-05-02
Allow modules to use val_functions

scottcwilson:
- Provide some generic error messages that accept a string
- <strike>Add a parameter to store the field name to the val_function</strike>
- Return false if validation fails and allow caller to determine redirect details
- Provide an example of how to do this in the flat shipping method.


scottcwilson:
I guess I would change it to "parm" or "err_parm" if people prefer that.

drbyte:
- Re: `string` - I agree we should probably use something that has more context.
In the single provided example, `string` is set to the same as the `configuration_title`. How often will that be different? The answer to that question may help determine proper naming.

- should the `zen_redirect` be moved outside of the function, and into the calling code. That way you don't need a 3rd parameter passed solely for looping through a bunch of possible redirect formats.

- since the `product_types` table uses the same format as `configuration` this should be supported there too at some point

Future consideration:
- For future it's probably worth moving validation stuff into its own language file (despite one of my posts to the contrary a few weeks ago) (and probably a class instead of just global functions), so it can be used anywhere. Granted, here we're dealing specifically with `configuration`-style validation, which has some specific requirements.

scottcwilson:
I actually like this much better.

drbyte:
@scottcwilson Is your subsequent plan to go and add this stuff to all core modules? 
Or was this one case the main situation presenting a problem?

scottcwilson:
The intention was simply to provide an example for people to follow.  Over time, sure it would be good to add all this validation, but not immediately.

PR #3401, opened 2020-05-02
Button(s) on Option Values Page should be Edit

dbltoe:
For consistency, the buttons on the Options Values page need to be changed to reflect Edit versus Update.  The system correctly presents an Update button AFTER the option value is edited.
With Update versus Edit, the user is confused as to where to go to edit the value.  Setting the button to Edit will match the same look on the Option Names page.


scottcwilson:
good job!

PR #3406, opened 2020-05-03
Use CDN for admin bs/fa/jqui

drbyte:
Switch to CDN as preferred load source

### DIFF
--- a/admin/includes/admin_html_head.php
+++ b/admin/includes/admin_html_head.php
@@ -15,7 +15,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <title><?php echo TITLE; ?></title>
 <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css" integrity="sha384-HSMxcRTRxnN+Bdg0JdbxYKrThecOKuH5zCYotlSAcp1+c8xmyTe9GYg1l9a69psu" crossorigin="anonymous">
-<link rel="stylesheet" href="includes/css/font-awesome.min.css">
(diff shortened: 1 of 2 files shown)


drbyte:
Question: how important is it to include source files for these in the distribution core?

I'm inclined to delete the self-hosted assets so we don't have to maintain them on an ongoing basis. 

Loading from CDN is simple and straightforward. Updating a version number is simpler maintenance.

Only downside is if you're offline ... but if you're offline then you're not using your live database either.
Or the rare case of a firewall blocking the CDN. But then your storefront will have similar challenges.

barco57:
So no fallback? I am aware that rarely if ever does a CDN fail to load... how about fallback to a different CDN

drbyte:
How do you fallback a css file load?

barco57:
Nevermind, I am an idiot, didn't think it through all of the way

drbyte:
Darn, I was hoping you had a solution! 😊

torvista:
>Only downside is if you're offline ... but if you're offline then you're not using your live database either.

Personally I had for many years had to work/develop offline or with wet-piece-of-string internet connection (petrochemical desert camps/offshore) and had things in place in the code to deal with that/not cause errors/wait for timeouts. 
So while not advocating that all necessary code should be **in** the code, at least it should allow for failure gracefully.

drbyte:
> So while not advocating that all necessary code should be **in** the code, at least it should allow for failure gracefully.

Well, having the files **in** the distro doesn't automatically "do" any sort of failover/fallback.

... unless you know of a non-intrusive way of having the browser identify that the content of a CSS file hasn't been loaded ...

drbyte:
@Zen4All Is the `bootstrap-theme.css` file intentionally not loaded?

Zen4All:
no, didn't know it was there

PR #3411, opened 2020-05-04
admin: orders fix for php notice undefined index

torvista:


drbyte:
As far as I can tell the only reason for this to be "undefined" is from the changes you've made to update a status from the infobox, right?

torvista:
Ah, true!

PR #3414, opened 2020-05-04
Updating docs directory files to match online FAQs

### LINKED ISSUE #3316: PRE-RELEASE refresh install and upgrade guides in docs
Per #3311 for the following files: 
docs/readme_installation.html
docs/readme_how_to_upgrade.html
docs/important_site_security_recommendations.html

do the following in mid May before release: 
- refresh the content on these pages from docs site 
- add a link at the top of these pages pointing to latest content on docs.zen-cart.com

scottcwilson:
Fixes #3316 

I will reiterate my suggestion that these files be replaced by links to the FAQ; it's a bad idea to have two sources of truth.  No one is setting up a Zen Cart on Mars; people *always* have access to the Internet.  The existence of these files is an anachronism.

### DIFF
--- a/docs/1.readme_installation.html
+++ b/docs/1.readme_installation.html
@@ -42,610 +42,373 @@
     <section id="zenTeamMessage">
         <h1>Welcome to Zen Cart&reg;</h1>    
         <div class="alert add-shadow">
-            <em>Dear Zen Cart&reg; User</em>,
+            <em>Dear Zen Cart User</em>,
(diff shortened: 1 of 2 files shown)


torvista:
>people always have access to the Internet.

Not true. But, I agree that it should all only be online, but with the possibility to get a one-click soft-copy/snapshot of the online docs in pdf to take with me to Mars/Qatar when required.

PR #3415, opened 2020-05-05
Use Ask a Question to handle call for price products

scottcwilson:
Instead of redirecting to the contact us page, redirect to the now built-in Ask a Question page, since the product information is available and can be used to save a step in this transaction.


scottcwilson:
Ugh I was on a branch when I opened this.  Closing; will resubmit.

PR #3416, opened 2020-05-05
Use Ask a Question for call for price

### LINKED ISSUE #3341: Feature request: use Ask A Question for Call for price
"Call for price": it makes more sense that the link refers to the "ask a question" page instead of "contact_us"
On the "product_info" page there is the image below the price and the "Call for price" link under the options: it is better to use the image with the links, the text is seen by few users and risks not being used. correctly.
Same thing for "product_listing"

scottcwilson:
Instead of redirecting to the contact us page, redirect to the now built-in Ask a Question page, since the product information is available and can be used to save a step in this transaction.

Fixes #3341

### DIFF
--- a/includes/functions/functions_general.php
+++ b/includes/functions/functions_general.php
@@ -865,7 +865,7 @@
       return $additional_link;
       break;
     case ($button_check->fields['product_is_call'] == '1'):
-      $return_button = '<a href="' . zen_href_link(FILENAME_CONTACT_US, '', 'SSL') . '">' . TEXT_CALL_FOR_PRICE . '</a>';
+      $return_button = '<a href="' . zen_href_link(FILENAME_ASK_A_QUESTION, 'products_id='.(int)$product_id, 'SSL') . '">' . TEXT_CALL_FOR_PRICE . '</a>';
       break;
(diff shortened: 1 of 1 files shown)


includes/functions/functions_general.php
```diff
@@ -865,7 +865,7 @@ function zen_get_buy_now_button($product_id, $link, $additional_link = false) {
       return $additional_link;
       break;
     case ($button_check->fields['product_is_call'] == '1'):
-      $return_button = '<a href="' . zen_href_link(FILENAME_CONTACT_US, '', 'SSL') . '">' . TEXT_CALL_FOR_PRICE . '</a>';
+      $return_button = '<a href="' . zen_href_link(FILENAME_ASK_A_QUESTION, 'products_id='.(int)$product_id, 'SSL') . '">' . TEXT_CALL_FOR_PRICE . '</a>';
```
drbyte, line 868: FYI - this would need to be `pid=` instead of `products_id=` since #3556

drbyte: Please change the parameter and then this can be merged.

drbyte: Actually, I can fix it directly :)

PR #3417, opened 2020-05-05
Admin orders

torvista:
short array syntax, sizeof to count, remove infobox br

### DIFF
--- a/admin/includes/languages/english/orders.php
+++ b/admin/includes/languages/english/orders.php
@@ -62,7 +62,7 @@ define('ENTRY_NOTIFY_COMMENTS', 'Append Comments:');
 
 define('TEXT_INFO_HEADING_DELETE_ORDER', 'Delete Order');
 define('TEXT_INFO_DELETE_INTRO', 'Are you sure you want to delete this order?');
-define('TEXT_INFO_RESTOCK_PRODUCT_QUANTITY', 'Restock product quantity');
+define('TEXT_INFO_RESTOCK_PRODUCT_QUANTITY', 'Restock product quantity?');
 define('TEXT_DATE_ORDER_CREATED', 'Date Created:');
 define('TEXT_DATE_ORDER_LAST_MODIFIED', 'Last Modified:');
 define('TEXT_INFO_PAYMENT_METHOD', 'Payment Method:');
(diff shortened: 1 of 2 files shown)


scottcwilson:
Not certain changing existing code from `array()` to `[]` is a big win.

torvista:
I have more stuff to add, this is a bit of clearing the decks.

drbyte:
> I have more stuff to add, this is a bit of clearing the decks.

Please provide a list. I'm not a fan of "dumping grounds" tossing a ton of stuff into one PR.

torvista:
..I have lots of changes, but on further examination, not much for general consumption.
I was going to do them as individual PRs after this, but possibly not worth it:
make search term sticky/pre-populated search field
set min/max limits on order id search

drbyte:
> I was going to do them as individual PRs after this

I'm open to these ideas: Just want to understand the use-case: 

> make search term sticky/pre-populated search field

Is it to refine the search by adding to it or editing it because of a typo?
Or is it something more complex like needing to re-do the identical search even though the results displayed are based on that same search but the results are stale because of using the browser's "back" button? 

> set min/max limits on order id search

Is this to trap for typos when entering an id number? 
or do you mean to limit the number of results shown?

torvista:
>Is it to refine the search by adding to it or editing it because of a typo?

To pre-load the search term for correction/modification: due to no/bad results from a mis-type.

>Is this to trap for typos when entering an id number?
To set min-max values on the id search field so you can only enter a valid order id number/in the range available.

drbyte:
> > > set min/max limits on order id search

> > Is this to trap for typos when entering an id number?

> To set min-max values on the id search field so you can only enter a valid order id number/in the range available.

I'm still not sure "why" on this. Are you typing ultra-long order id numbers somehow? or negative numbers?

torvista:
The idea was to block invalid numbers in the browser to avoid making a pointless search.
Can't remember why I did it. Like all mods, they all seem necessary at the time, but a few years later...one forgets.
But still, since it gives visual indication of the limits, with no code necessary (for the display), when the user miss-keys out of range.

scottcwilson:
Can this be closed?

torvista:
whatever, too busy

PR #3420, opened 2020-05-05
functions zen_draw_label: remove extra space when no parameters

torvista:


drbyte:
I'm open to the proposed change.

Feedback on 3 points:
- I'm not convinced that adding the docblocks here adds any value, even for an IDE. Let's remove them.

- Instead of `$parameters !== ''` let's use the more expressive `!empty($parameters)`, especially since no parameter will ever be a `0` that could be misinterpreted.

- The first commit you made in this batch is bad. While you corrected it in the next commit, I'd like to see that first commit excluded from the git history. That'll require squashing your commits (which is done by rebasing). (I'll often do an "amend" of a prior commit, but that's only good for altering the most recent one, not multiple behind it.) Followed by a force-push to the github branch.

drbyte:
Update:

- ~I'm not convinced that adding the docblocks here adds any value, even for an IDE. Let's remove them.~
EDIT: If we're going to add them, then it would be appropriate to indicate the field type for all the fields, not just the ones that your IDE's EA tool could detect. 
eg: `@param $text` would become `@param string $text`. Same with $for.

torvista:
That seems to have worked, thanks for the info, another git step learnt!

drbyte:
Can you squash this down to just a single commit?

ie: rebase, squashing all commits into previous ones, so it all looks like "one" commit. Then force-push.

torvista:
can't seem to squash it any more.

torvista:
So, ideally/in general you want a single commit in a PR, but with a text description of the changes,  rather than a series of commits that describe each change?

drbyte:
> So, ideally/in general you want a single commit in a PR, but with a text description of the changes, rather than a series of commits that describe each change?

No, not necessarily. What I don't want is commits that knowingly have a bug in them. Your commit here that wraps the variable name inside single quotes doesn't work. I'd like to see that factored out, which should be doable with an interactive rebase where you tell it which commits to combine into each other.

--

It's been awhile since I wrote about commits and PRs in the git docs that are in the docs repo, so I don't recall how much crossover the following thoughts have with that.

In a perfect world:

- a PR contains just a bugfix or feature addition/improvement, but not also reformatting of code. It can contain multiple commits that are each canonical in terms of their scope, but preferably not commits that are merely iterations of prior commits in the same PR.
- reformatting is done in its own PR. In this case it's fine to have separate commits for different kinds of reformatting (much like you've done separate formatting commits in many of your PRs)
- commits that contain an error are removed/replaced by squashing the fix into the same commit. This way when someone investigates the code in the future they don't see "oh, I see: you do it this way" but accidentally grab the buggy code instead of the fixed code done later.
- PR text descriptions (and commits too really) should explain the problem that exists before the solution was created, why certain design choices were made, what business problem is being solved, and maybe a before/after visual if relevant. PR descriptions should also include a `closes #xyz` mention if it relates to and resolves an open issue, and not "only" say `closes xyz`; provide some context.
- commit titles and PR titles should give the reader the big picture and make sense if you were to look at it 6 months from now without knowing any context

When you have to go back to old code to find out "why" something was done "that way", in order to determine whether someone's new PR which proposes a "different way" or an enhancement, might break the old, you'll appreciate having the commit history be very canonical and well-thought-out.
I spend hours investigating old code whenever someone wants to change things, or is reporting a bug.
Plus, when a new PR contains both code-formatting changes and bugfixes and layout rearrangements and new features .... it takes hours to scan through everything to understand. (To your credit you've done great with at least segmenting each(most) step into separate commits, so at least reviewing things that way is simpler than reviewing the PR as a whole; but it still takes longer to review and approve than if it were 2-3 PRs instead (ie: bugfix or feature in one, and code-reformatting in another.)

I'm fully aware that it's hard to do a 2nd PR of the same file while waiting for a prior one to be reviewed and merged. 
I'll usually do a bugfix commit first, because that's the priority.
If I'm still interested in working with the file further, then I'll do the mass reformatting, get it merged. Maybe multiple commits, each canonical in their own right, but one PR for formatting changes.
After that I'll tackle other changes (feature improvements/additions, etc) in the new-cleaned-up code, and then do a 3rd PR.

I admit: sometimes I break my own rules. But then of course I have only myself to blame.

drbyte:
If you're unable to get it squashed in this case, that's fine. Just say so and I'll do some stuff on my end. It's not a total showstopper. But it is useful as a learning opportunity.

There are multiple times where I've made a complete clone of my local repo directory, rolled back the branch to the beginning of where I forked it, did a git-reset-hard to undo the commits, ran a merge util to copy the changes over from the clone, and then manually did the commits afresh, and force-pushed the branch up to github. And then go find something less frustrating to do for awhile, like eat apple pie or something! :)

torvista:
Well, thanks for taking the time to detail that, I hope you will be putting this in the documentation site.
A couple of weeks ago wilt detailed his Git methodology, which again should be documented...with explanations, as I don't follow all the reasoning.
While everyone may have their own process, it is very helpful to know others and their benefits. There does not have to be a "definitive" recommendation in the docs: having various methods described by real people adds a necessary personal touch to encourage community.

scottcwilson:
This is an easily Googleable topic - not wild about adding it to the Zen Cart FAQ.

drbyte:
After release of 1.5.7 I hope to add more things to the docs

PR #3431, opened 2020-05-07
installer text

torvista:
and used text_replacement variables

### DIFF
--- a/zc_install/includes/languages/en_us/main.php
+++ b/zc_install/includes/languages/en_us/main.php
@@ -169,8 +169,8 @@ return [
 'TEXT_ERROR_MULTIPLE_ADMINS_NONE_SELECTED' => 'Multiple Admin directories seem to exist. Either remove the duplicate admin directories and click Refresh or select the correct admin directory below and click Refresh.',
 'TEXT_ERROR_MULTIPLE_ADMINS_SELECTED' => 'Multiple Admin directories seem to exist. If the selected directory below is incorrect, please choose another and click Refresh.',
 'TEXT_ERROR_SUCCESS_NO_ERRORS' => 'No errors were detected on your system. You may continue with the installation.',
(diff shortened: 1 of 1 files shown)


zcwilt:
Why use the text replacements for such simple pieces of text

torvista:
Since you'd bothered to make it work!

zcwilt:
Haha, guess I deserved that. 
Was more thinking that not having the interpolated defines simplified things anyway, why go back to something more complicated.

zcwilt:
Happy to merge as is.

zc_install/includes/languages/en_us/main.php
```diff
@@ -169,8 +169,8 @@
...
-'TEXT_UPGRADE_INFO' => 'An Upgrade will inspect your database and subsequently offer the steps required to upgrade to the current version (adding new fields/modifying existing fields). This is intended to be a non-destructive process, but as with all modifications you must ensure you have a verified backup of your database available before proceeding.',
-'TEXT_CLEAN_INSTALL_INFO' => 'A Clean Install will revert the database to a new state, deleting all data. Optionally, the demonstration data may be loaded as part of this process.',
+'TEXT_UPGRADE_INFO' => '%%TEXT_UPGRADE%%: will inspect your database and subsequently offer the steps required to upgrade to the current version (adding new fields/modifying existing fields). This is intended to be a non-destructive process, but as with all modifications you must ensure you have a verified backup of your database available before proceeding.',
+'TEXT_CLEAN_INSTALL_INFO' => '%%TEXT_CLEAN_INSTALL%%: will revert the database to a new state, deleting all data. Optionally, the demonstration data may be loaded as part of this process.',
```
drbyte, line 173: Hmmm ... but ... if you look at the defines you're interpolating this will create some confusing output, won't it?

```
'TEXT_UPGRADE' => 'Upgrade ...',
'TEXT_CLEAN_INSTALL' => 'Clean Install',
```

torvista: These are popups on the same buttons, so I don't think so.

PR #3432, opened 2020-05-07
zcgo is a command line interface for Zen Cart

zcwilt:
`zcgo` is a command line interface to run certain actions useful to developers. It is in that sense somewhat similar to `Laravel artisan`.

Currently the only command available is meant to aid in creating the new php array style language files from legacy define files. 

The next thing I would like to do is the creation of `idehelper` files for those array files to provide some `hinting` in `phpStorm`

Going further, I would also like to add helper commands to aid with repeated installations. 
e.g. creating new/updated db schemas without having to run the installer.

Note: There are `huge` security concerns in having  `zcgo` on a live server, and will address this both in documentation and in the building of release zips.


zcwilt:
There will be some documentation for this in the docs website soon.

zcwilt:
Documentation has been added at https://docs.zen-cart.com/dev/zcgo/

Still going to close this and pick up after v157 release.

scottcwilson:
Suggested small demo utility: use zcgo to clean out the admin activity logs on local development sites.

zcgo/MakeDefinesCommand.php
```diff
@@ -0,0 +1,299 @@
...
+        }
+    }
+
+    protected function validateInputOptions($input)
+    {
+        $file = $input->getOption('file');
+        $dir = $input->getOption('dir');
+        $config = $input->getOption('config');
+        if (!isset($file) && !isset($dir) && !isset($config)) {
+            throw new InputOptionsValidationException('Seems you didn\'t pass any options');
```
mc12345678, line 86: Shouldn't this read 'Seems you didn\'t pass all options'? Using `all` instead of `any`?

zcwilt:
note 
The push is mainly to fix problems with the make:defines command.
Other commands are WIP.

scottcwilson:
Will it be possible or practical to backport this to (say) 1.5.6?

zcwilt:
Should be. Currently there is no reliance on core code. 
Even DB stuff will be done thru `illuminate/database'
It would also be easier if I set this up as a separate github/packagist project.
In which case this PR becomes moot.

scottcwilson:
Fantastic!  My sample use case would be: run this sql script via Tools > Install SQL Patches.

drbyte:
> My sample use case would be: run this sql script via Tools > Install SQL Patches.

In lots of cases this would be better handled in code, instead of having people run patch queries.

scottcwilson:
This is not for distributed software.  To be more specific: I create a feature for user X on computer 1 that requires a database change that I do via a .SQL file.  When I go to computer 2, I also have a copy of their store, and I need to remember to run that SQL file.  If I could script this using zcgo, I could automate the process to keep all the databases up to date automatically.

zcwilt:
Going to close this PR as `zcgo` is now installable via composer, so can be used on any version of Zen Cart that supports composer.
So 155 and up 

See https://github.com/zcwilt/zcgo/
and https://docs.zen-cart.com/dev/zcgo

Although at some point I'll move it to zencart/zcgo

PR #3435, opened 2020-05-07
Products-Viewed report - indicate if it's been disabled

scottcwilson:


drbyte:
I see your point on this.
It's certainly a light-weight approach, and thankfully it avoids adding yet another config switch.

But I'm a little resistant to reaching into a catalog-side "extra_datafiles" location to do the determination.  Specifically I don't want people thinking they should follow this pattern when writing other mods.

I'm undecided about best course of action:
- perhaps to explore alternate approaches
- close it
- delete the report altogether (who's been using it anyway?)
- move the report to a plugin

Again, the reason for disabling the tracker instead of refactoring the db schema is because these specific stats would be better tracked via a more feature-rich tracking mechanism, external to the core code. That could be a 3rd party service which provides its own dashboard or has an admin report that reads the external data, or a complete plugin.

scottcwilson:
The day after you did this I got a complaint from a tester saying the report didn't work. 

We have intentionally broken it - let's tell people rather than make them debug it themselves.

drbyte:
> The day after you did this I got a complaint from a tester saying the report didn't work.

What exactly was the complaint? If they were testing it on pre-existing data, the report would still have shown the historical data.

scottcwilson:
Install 1.5.7 from scratch.
Poke at some products
Check the report.

scottcwilson:
My first vote would be to REMOVE this report, since **it doesn't work** anymore. 
Failing that, we should merge this PR. 
The fact that you can see OLD INACCURATE data does not mitigate this.  It doesn't work!

drbyte:
Superceded by #3481

PR #3437, opened 2020-05-08
Refactor admin modules page

torvista:
change Delete for Remove
refactor row styles to show pointer only where there is a clickable action
minor html fixes
no functional changes

### DIFF
--- a/admin/modules.php
+++ b/admin/modules.php
@@ -194,7 +194,7 @@ if (zen_not_null($action)) {
                 }
 
                 $installed_modules = $temp_for_sort = [];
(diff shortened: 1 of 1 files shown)


torvista:
Agree with html.
I didn't find it easy to work out the functionality so I added the comments for my use and thought they do no harm.

drbyte:
Aside: All over the Admin these nested if/else/etc for rendering the `tr`s is something I wish we could get away from. That and the `onclick` attributes.

proseLA:
i was also thinking the same thing:

https://blog.codinghorror.com/flattening-arrow-code/

Issue #3290, opened 2020-04-16
Template bug: black bg if no javascript enabled on responsive_classic


proseLA:
i am operating on a base v156c ZC install.  i have yet to test the v157; but it is on my list.

i was debugging an issue i was having, when it now appears that the issue may be part of the template.  i am using the responsive classic template straight out of the box.  my issue is simple; go to the home page.  then disable javascript and refresh.  do they look the same?  my install does not.  there is a very dark background when javascript is disabled.

is this the intended behavior?  or is there something adrift on my install?  or is this a bug?  if someone could confirm one way or another that would be great. 

thanks in advance.

barco57:
it's not a change in the background color, it is the collapsing of the footer from the bottom up to behind the main content area. The stretching of that footer background to cover a substantial portion of the main content area background is caused by a banner getting stuck under the right column when the footer collapses (Firefox, in chrome the banner disappears to somewhere I haven't determined). I would call it a bug... I don't use Responsive Classic so I have never run into it before and also have never had anyone report it.

proseLA:
@barco57 it is definitely a bug.  amazing how easy it is to get sidetracked.

at easy solution has not appeared to me....  i'll put it on my list...  as i have other things on which i am working.  unless someone else gets to it before me....

thanks for confirming.

proseLA:
bug remains in v157

PR #3442, opened 2020-05-09
Force template column widths to use px units.

drbyte:
This avoids invalid contents due to tampering.
Templates that rely on different units can change the units if needed directly in their own template code that's already being heavily customized.


drbyte:
May also play a part in fixing #3290

PR #3452, opened 2020-05-12
Add count_key for splitpageresults notifier

mc12345678:
There are some queries that require providing a specific field or set of fields when splitting the page results.  While this edit does not add all available parameters of the `splitPageResults` class, it does add the next one.  At this time, the parameters to the class are: `$query, $max_rows, $count_key = '*', $page_holder = 'page', $debug = false, $countQuery = ""` with the first two already considered in the previous code.


proseLA:
if you are adding the count key here, would it not behoove one to add it here as well:

https://github.com/zencart/zencart/blob/0d38ca7ea9389641a44e3eb4e827558c2bed6a38/includes/modules/pages/featured_products/header_php.php#L38

PR #3453, opened 2020-05-12
Ensure products_id parameter has a main_page

mc12345678:


includes/init_includes/init_sanitize.php
```diff
@@ -139,7 +139,7 @@
  * validate products_id for search engines and bookmarks, etc.
  */
   if (isset($_GET['products_id']) && (!isset($_SESSION['check_valid']) || $_SESSION['check_valid'] != 'false')) {
-    $check_valid = zen_products_id_valid($_GET['products_id']);
+    $check_valid = zen_products_id_valid($_GET['products_id']) && isset($_GET['main_page']) && zen_not_null($_GET['main_page']);
```
drbyte, line 142: Probably can just use `!empty()` instead of isset/zen_not_null

mc12345678: So directly exclude `main_page=0`? :)

Just askin'. :)

I had considered it as well, then left it for further decision. Considering adjusting this file, would modify the "equivalent" test a few lines further down to match whatever this becomes (`isset/zen_not_null` vs. `!empty()`).

drbyte: Yes :)  I'm not worried about clashing with `=0`.

and yes, makes sense for line 157 also.

PR #3456, opened 2020-05-14
Fixes currency_cron failures

### LINKED ISSUE #3455: zc157: currency_cron.php results in a fatal error
A log similar to the following is created:
```
[14-May-2020 12:21:03 Europe/Berlin] PHP Fatal error:  Cannot declare class base, because the name is already in use in C:\xampp\htdocs\zc157a\includes\classes\class.base.php on line 17
```

scottcwilson:
Fixes #3455.
Now that admin/includes/application_bootstrap.php does a bunch of includes, the load list for currency cron needs to be slimmed down.

### DIFF
--- a/admin/includes/auto_loaders/currency_cron.core.php
+++ b/admin/includes/auto_loaders/currency_cron.core.php
@@ -12,8 +12,6 @@ if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
   $autoLoadConfig[0][] = array('autoType'=>'require',
                                'loadFile'=> DIR_FS_CATALOG . DIR_WS_INCLUDES .  'version.php');
   $autoLoadConfig[0][] = array('autoType'=>'class',
-                               'loadFile'=>'class.base.php');
-  $autoLoadConfig[0][] = array('autoType'=>'class',
                                'loadFile'=>'class.notifier.php');
   $autoLoadConfig[0][] = array('autoType'=>'classInstantiate',
                                'className'=>'notifier',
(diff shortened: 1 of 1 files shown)


lat9:
Noting that this is going to be a _major pain_ for site-specific customizations that have defined additional cron's based on the pre-zc157 version of `currency_cron.php`.

drbyte:
> Noting that this is going to be a _major pain_ for site-specific customizations that have defined additional cron's based on the pre-zc157 version of `currency_cron.php`.

Noted. Something to add to the upgrade docs for sure. At least it's just a matter of removing a few no-longer-needed lines of code.

drbyte:
That said, @scottcwilson and @lat9 does changing `autoload_func.php` to use `include_once` and `require_once` abate the upgrade pain a bit?
https://github.com/zencart/zencart/blob/0d38ca7ea9389641a44e3eb4e827558c2bed6a38/includes/autoload_func.php#L23-L27

(We'd still merge this PR, but if it means these plugins don't require big changes for different version compability then it'd be an ideal bridge for the shorter term.)

scottcwilson:
@drbyte I considered this but this makes EVERYTHING slower just to save a bit of pain for one file.  Are people really copying this file and customizing it?  I have never seen that.

drbyte:
> I considered this but this makes EVERYTHING slower

I was reading a bit about this the other day, and there are lots of mixed opinions on speed difference. Two common threads from the reading: the difference is extremely small, and if an opcode cache is enabled then the difference is nil.

lat9:
> 
> 
> @drbyte I considered this but this makes EVERYTHING slower just to save a bit of pain for one file. Are people really copying this file and customizing it? I have never seen that.

I'm not sure if I qualify as `people`, but I've used that model for custom cron-scripts for various sites.

lat9:
What would be _really nice_ (albeit not downward compatible) would be to have a version-specificity added to the auto-loading process.  Please note that I'm not suggesting this for zc157.

PR #3460, opened 2020-05-14
Simplify queries

Zen4All:
- Remove unused fields from category, and product queries. Potentially speeding up sites with large numbers of categories, and products


drbyte:
Thanks for this. I've added some additional reduction of duplication.

PR #3468, opened 2020-05-16
Contrast in Responsive Classic Colors Stylesheet

dbltoe:
If a person were to use the existing responsive_classic template out of the box, the accessibility score would be affected by some of the standard colors.

Making the stylesheet_colors.css changes listed here results in a demo store passing all contrast tests.

I will also be addressing this in the Color Change Mod for 1.5.7


drbyte:
Yikes. That's a pretty drastic change in the shade of blue. 
I guess I can get used to the royal-blue for the top horizontal categories-nav.

The hover-color for the blue buttons hasn't been updated to match.
Screen Shot 2020-05-16 at 4 42 47 PM


What I dislike about this blue for all the links and titles is that now the link text looks like it's just the original default blue that browser do by default, making things look like they come from the 80's again.
Image 2020-05-16 at 4 41 40 PM

dbltoe:
I was trying to get as close as possible to the original blue but still meet the min 4.5 to 1 contrast.
Not written in stone if someone has a better choice that meets the 4.5
How about #4865da?

dbltoe:
I'm thinking this can be better and needs some more thought on the hovers.
The best option for me is to close and resubmit as I'm not familiar with changing a PR file.

PR #3480, opened 2020-05-18
Correct grouping of attributes in popover

### LINKED ISSUE #3475: admin/orders.php: Popup hint splits products with attributes
I like that little icon addition to pop-up the products' ordered.  When I placed an order with an attributed product, though, the popup indicated two products, each with their own attribute while the single product in the order had two:
image
image

scottcwilson:
Fixes #3475

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -1251,16 +1251,21 @@ if (zen_not_null($action) && $order_exists == true) {
                 <td class="dataTableContent text-right"><?php echo strip_tags($currencies->format($orders->fields['order_total'], true, $orders->fields['currency'], $orders->fields['currency_value'])); ?></td>
                 <td class="dataTableContent text-right dataTableButtonCell">
                     <?php
-                    $sql = "SELECT op.products_quantity AS qty, op.products_name AS name, op.products_model AS model, opa.products_options AS product_option, opa.products_options_values AS product_value 
+                    $sql = "SELECT op.orders_products_id, op.products_quantity AS qty, op.products_name AS name, op.products_model AS model
                             FROM " . TABLE_ORDERS_PRODUCTS . " op 
-                            LEFT OUTER JOIN " . TABLE_ORDERS_PRODUCTS_ATTRIBUTES . " opa ON op.orders_products_id=opa.orders_products_id
-                            WHERE op.orders_id = " . (int)$orders->fields['orders_id'];
+                            WHERE op.orders_id = " . (int)$orders->fields['orders_id']; 
                     $orderProducts = $db->Execute($sql, false, true, 1800);
                     $product_details = '';
                     foreach($orderProducts as $product) {
                         $product_details .= $product['qty'] . ' x ' . $product['name'] . ' (' . $product['model'] . ')' . "\n";
-                        if (!empty($product['product_option'])) {
-                            $product_details .= '&nbsp;&nbsp;- ' . $product['product_option'] . ': ' . zen_output_string_protected($product['product_value']) . "\n";
+                        $sql = "SELECT products_options, products_options_values 
+                            FROM " .  TABLE_ORDERS_PRODUCTS_ATTRIBUTES . " 
+                            WHERE orders_products_id = " . (int)$product['orders_products_id'] . " ORDER BY orders_products_attributes_id ASC"; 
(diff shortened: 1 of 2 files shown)


admin/orders.php
```diff
@@ -1251,16 +1251,19 @@ function couponpopupWindow(url) {
...
+                            WHERE op.orders_id = " . (int)$orders->fields['orders_id']; 
                     $orderProducts = $db->Execute($sql, false, true, 1800);
                     $product_details = '';
                     foreach($orderProducts as $product) {
                         $product_details .= $product['qty'] . ' x ' . $product['name'] . ' (' . $product['model'] . ')' . "\n";
-                        if (!empty($product['product_option'])) {
-                            $product_details .= '&nbsp;&nbsp;- ' . $product['product_option'] . ': ' . zen_output_string_protected($product['product_value']) . "\n";
+                        $sql2 = $db->Execute("SELECT products_options, products_options_values 
+                            FROM " .  TABLE_ORDERS_PRODUCTS_ATTRIBUTES . " 
+                            WHERE orders_products_id = " . (int)$product['orders_products_id'], false, true, 1800);
```
lat9, line 1263: Shouldn't there be an ORDER BY clause here, e.g. `ORDER BY orders_products_attributes_id ASC`?

drbyte:
> here we have a var called `$sql2` which is not even a sql statement, but the result of a `$db->Execute` function

Agreed. `$sql2` should be something like `$productAttributes` and then `foreach`'d 
Also prefer that the actual query be assigned to a string (`$sql` could be re-used for this) 
and the `Execute()` call be on its own line, especially since it's passing additional parameters.

drbyte:
Also needs repeating in dashboard widget.

drbyte:
> i am not a database efficiency expert by any stretch as well. its entirely possible that we could accomplish the same goal with the original sql statement, and testing for a change in the products_id field to accomplish the same thing. as to whether that approach utilizes less resources and is more efficient, i can not say.

I've been contemplating this. Might require a `GROUP BY` clause or extra `ORDER BY` fields.

I'd kinda rather that all the order queries for this entire page were skipped and just use the `order` class, for consistency of retrieval formulae. Perhaps a 1.5.8 refactor.

scottcwilson:
GROUP_BY would be my first go to but now you have to worry about the full_group_by stuff. :( So this seemed like an approach that would be more likely to work first time.

admin/orders.php
```diff
@@ -1251,16 +1251,19 @@ function couponpopupWindow(url) {
...
+                    $sql = "SELECT op.orders_products_id, op.products_quantity AS qty, op.products_name AS name, op.products_model AS model
                             FROM " . TABLE_ORDERS_PRODUCTS . " op 
-                            LEFT OUTER JOIN " . TABLE_ORDERS_PRODUCTS_ATTRIBUTES . " opa ON op.orders_products_id=opa.orders_products_id
-                            WHERE op.orders_id = " . (int)$orders->fields['orders_id'];
+                            WHERE op.orders_id = " . (int)$orders->fields['orders_id']; 
                     $orderProducts = $db->Execute($sql, false, true, 1800);
                     $product_details = '';
                     foreach($orderProducts as $product) {
                         $product_details .= $product['qty'] . ' x ' . $product['name'] . ' (' . $product['model'] . ')' . "\n";
-                        if (!empty($product['product_option'])) {
```
drbyte, line 1262: What was the reason for removing the `!empty()` check?

scottcwilson: Finger trouble.

proseLA:
> I'd kinda rather that all the order queries for this entire page were skipped and just use the `order` class, for consistency of retrieval formulae. Perhaps a 1.5.8 refactor.

yes. i would agree with that.  bit of a thread drift; but when looking at this script and i see things like `delete_cvv` and `mask_cc`, it makes me cringe.  to be PCI compliant, storing a cvv is never allowed, and i think that field should be removed from the orders table.  masking a cc number implies a stored cc number and i'm not seeing any encryption of that field.  not that i recommend storing that field in any event.  at some point, it would be nice to remove legacy code, and suffer the rantings of the angry mob.

drbyte:
Complete derail of topic, yes. The controls you mention are specifically to manage and remove old data, not to allow storing of it.

PR #3481, opened 2020-05-18
Enhance the product-views tracking, and add date filter to report

drbyte:
This orphans the `products_viewed` field from the `products_description` table,
by relocating the counting to the new `count_product_views` table,
where views are tracked by date.

Also skips tracking hits by Admin and Spider visitors.

This removes the "reset" in Store Manager because the former need for a reset is deprecated because data can now be filtered by date.

The report's default date range is "today for the past 30 days". Editable on-screen to filter to a different date range.

NOTE: Plugins which have cloned the `copy_product_confirm.php` file for custom product-types will want to skip the copy of the `products_viewed` field when copying.


scottcwilson:
Why not just remove the `products_viewed` field entirely, since it is no longer used?

drbyte:
> Why not just remove the `products_viewed` field entirely, since it is no longer used?

I recall being flogged for deleting peoples' data without warning. Is this case different?

scottcwilson:
Do you want me to submit the PR?  Because I have no problem at all ignoring whining. 

It's not used, it should be deleted.  Anyone who disagrees is welcome to mod the upgrade procedure - it's open source!

drbyte:
> Why not just remove the products_viewed field entirely, since it is no longer used?
>
> Do you want me to submit the PR? Because I have no problem at all ignoring whining.
> 
> It's not used, it should be deleted. Anyone who disagrees is welcome to mod the upgrade procedure - it's open source!

It's not "not used". It contains real data. Maybe drop the field in 1.5.8 upgrade.

PR #3484, opened 2020-05-18
Admin-template select: revision for clarifications

torvista:
This revising was instigated by my confusion at not understanding the process of using t=template for admin/development use.
This has been resolved by renaming the infoboxes to better reflect their function and adding extra information.
Also:
Associations can only be added when there are un-associated languages still to be added.
An option to remove the t=override was added.

Usual html cleanups.

### DIFF
--- a/admin/includes/languages/english/template_select.php
+++ b/admin/includes/languages/english/template_select.php
@@ -1,45 +1,41 @@
 <?php
-//
-// +----------------------------------------------------------------------+
-// |zen-cart Open Source E-commerce                                       |
-// +----------------------------------------------------------------------+
-// | Copyright (c) 2003 The zen-cart developers                           |
-// |                                                                      |
-// | http://www.zen-cart.com/index.php                                    |
-// |                                                                      |
-// | Portions Copyright (c) 2003 osCommerce                               |
-// +----------------------------------------------------------------------+
-// | This source file is subject to version 2.0 of the GPL license,       |
-// | that is bundled with this package in the file LICENSE, and is        |
-// | available through the world-wide-web at the following url:           |
-// | http://www.zen-cart.com/license/2_0.txt.                             |
-// | If you did not receive a copy of the zen-cart license and are unable |
-// | to obtain it through the world-wide-web, please send a note to       |
-// | license@zen-cart.com so we can mail you a copy immediately.          |
-// +----------------------------------------------------------------------+
-//  $Id: template_select.php 1105 2005-04-04 22:05:35Z birdbrain $
-//
+/**
+ * @package admin
+ * @copyright Copyright 2003-2020 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id:  Modified in v1.5.7 $
+ */
 
 define('HEADING_TITLE', 'Template Selection');
+define('TEXT_TEMPLATE_SELECT_INFO', '<p>A different template may be assigned to each of the installed languages.<br>The default template is used when there is no template assigned to that language.</p>
+<p>It is possible to view a template in a <em>private</em> session, for example when testing a new template:</p>
+<ol>
+<li>Assign the current/active template to <strong>all</strong> of the installed languages.</li>
+<li>Assign the new template as the Default.</li>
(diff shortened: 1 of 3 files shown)


drbyte:
Undecided on this.

My **not-yet-fully-developed-thoughts**, posted just so there's a response of some sort:

I don't like the large block of on-page documentation: it's like an invitation to use a dev-only feature, which will lead to support questions. I'd rather it was in the dev docs and people can consult that if they need it.

While I empathize with the desire to add clarity around associating templates with languages, I similarly don't want to strongly encourage that activity: it's very rare (ie: never) that one "actually" needs a completely different template for a certain language, and to have both active on the store at the same time.

torvista:
I think private new-template testing is an absolutely essential feature: probably the aspect of development that really demands to be on a real/**public** server for testing with any devices (friends/family) you can get your hands on. So important that I wanted to make sure it did actually work, after failing to understand it on my first attempt!

>I don't like the large block of on-page documentation: 

No, it's ugly. I think a single line mention of the functionality that new templates can be privately tested would be enough, and refer to the help button.

>  it's very rare (ie: never) that one "actually" needs a completely different template for a certain language, and to have both active on the store at the same time.

Yes, which surprised me that the language-association aspect is a necessary part of being able to test-drive a new template.
I had hardly ever ventured into the page prior to this and was unaware of the option.
If this was reworked completely to drop the language aspect I doubt anyone would miss it.

scottcwilson:
> I don't like the large block of on-page documentation.... 

No need anymore. 
Remember you already have the help button which links to this: 

https://docs.zen-cart.com/user/admin_pages/tools/template_selection/

You can then modify this page with a developer section - "Developers, please see <link> for coverage of template creation features ..."

torvista:
>You can then modify this page with a developer section - "Developers, please see for coverage of template creation features ..."

Ok, once this is resolved.

admin/includes/languages/english/template_select.php
```diff
@@ -1,44 +1,32 @@
...
+ * @package admin
+ * @copyright Copyright 2003-2020 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id:  Modified in v1.5.7 $
+ */
 
 define('HEADING_TITLE', 'Template Selection');
-
+define('TEXT_TEMPLATE_SELECT_INFO', '<p>Here you may change the template used to format the Shopfront.</p><p>A different template may be assigned to each of the installed languages. The default template is used when there is no specific template assigned to a language.</p><p>It is possible to view a template in a <em>private</em> session, for example when testing a new template: see the online Help <strong>?</strong>.</p>');
```
drbyte, line 11: A few thoughts:

- While I accept that terminologies are different in various parts of the world, the rest of Zen Cart uses the word "store" instead of "shop".
For consistency can we change "Shopfront" to "storefront"?

- the 2nd sentence, about "a different template per language" and "the default is used...": Can we put the "default" part first, and reword the rest so that it sounds like one will never need to do the language-association of templates?
I don't want to encourage people to start making new template-language associations in case we change that in 1.5.8.

admin/includes/languages/english/template_select.php
```diff
@@ -1,44 +1,32 @@
...
 define('TABLE_HEADING_LANGUAGE', 'Template Language');
 define('TABLE_HEADING_NAME', 'Template Name');
 define('TABLE_HEADING_DIRECTORY', 'Template Directory');
 define('TABLE_HEADING_ACTION', 'Action');
 
+define('TEXT_INFO_HEADING_LANGUAGE', 'Language');
+define('TEXT_INFO_DEFAULT_LANGUAGE', 'Default (any)');
```
drbyte, line 18: Perhaps `Default (any)` might be better reworded as `Default (all languages)` for clarity. This also may help reduce the temptation to create new template-language associations.

torvista: If no languages are in the list, yes "all" is better.
But, when languages are in the list, it is more appropriate to say "any" than all...that was one of the things that I found confusing as it was no longer "all".
Anyway, it has survived long enough as "all" to not matter.

drbyte:
I updated to add clarity about the most common use-case for template-language assignments: supporting RTL+LTR templates in the same store.

Also removed nested define, as it makes things unnecessarily complex.

PR #3488, opened 2020-05-19
Primary Key Isn't Not Null

dbltoe:
If date_viewed is to be a primary key, isn't it required to be NOT NULL?


dbltoe:
Hit enter before adding a default

PR #3499, opened 2020-05-19
Color Contrast in #navMain

dbltoe:
took color call out of  includes/templates/responsive_classic/templates/common/tpl_header.php and added the proper calls in includes/templates/responsive_classic/css/stylesheet_colors.css for the #navMain link hovers.


scottcwilson:
@dbltoe Would you please write a bullet or two about accessibility that I can add to the product features page?  Thanks.

dbltoe:
> @dbltoe Would you please write a bullet or two about accessibility that I can add to the product features page? Thanks.

Can you point to the existing so I can get a better idea of what's needed?

scottcwilson:
Was intending to add to 
https://docs.zen-cart.com/user/about_us/features/

Would also use what you provide to update the 1.5.7 what's new document.

PR #3502, opened 2020-05-20
V157 enable error logging static

mc12345678:
Two items here, one is discussed: https://github.com/zencart/zencart/commit/e87391bbf17368be5f4c0d64518c193d9a35434a#r39297222

The other is a result of seeing how this works and making it possible for the operation to remain as it was for those interested.  Kept the default the same as has been provided (to split out the files), but made it admin controllable to be able to make sense of the data in a logical sequence rather than by some sort of file convention.

### DIFF
--- a/includes/extra_configures/enable_error_logging.php
+++ b/includes/extra_configures/enable_error_logging.php
@@ -21,8 +21,9 @@ function zen_debug_error_handler($errno, $errstr, $errfile, $errline)
         return;
     }
 
+    static $last_log_suffix;
     if (!isset($last_log_suffix)) {
-        static $last_log_suffix = '.log';
+        $last_log_suffix = '.log';
     }
     $ignore_dups = false;
     if (IS_ADMIN_FLAG === true) {


drbyte:
Was the use of `backtrace` intentional in the `REPORT_ALL_ERRORS_BACKTRACE_FILE_ORDER` name?
If not, can you remove it and "amend the last commit and force-push" it?

mc12345678:
Ish.. Not required to be present, was just a spur of the moment naming to assign something. Will do though on all aspects.

drbyte:
Thanks for the update.

Now, back to the first point: I read your comment in the other thread. I follow the PHP logic in the example you posted.

But I'm not clear what your expectation is vs the actual outcome of the implementation in the logging filename extensions.

Are you saying that the generated filenames are wrong? 
Or just that the "use of the static variable is not optimally done, resulting in wasted CPU cycles"?

mc12345678:
> Now, back to your first point: I read your comment in the other thread. I follow the PHP logic in the example you posted.
>
> But I'm not clear what your expectation is vs the actual outcome of the implementation in the logging filename extensions.

>Are you saying that the generated filenames are wrong?
> Or just that the "use of the static variable is not optimally done, resulting in wasted CPU cycles"?

Well, I haven't exactly created a specific test case/example and there is the question of "validity" of it happening, but with the current arrangement, every time the function is entered, `$last_log_suffix` is set to `.log`... Which means that whatever the current `$GLOBALS['debug_logfile_path']` value is, when/if the condition changes to something else (say from a notification to a warning), it appears that the filename will change to have a warning at the end of it, plus the default `.log` extension, so that then the next "issue" that is brought up presuming to be some other issue would arrive with `.log` as the "last" and then would replace that portion with the next new word, so presumably the filename would "extend" to identify each next "issue"...

With the revision that I proposed, there would be only one "filetype" for each issue type, such that if there was a sequence of deprecation, warning, deprecation then there would be two files created, one with -deprecated.log and one with -warning.log... Under the current "usage" I would expect to see -deprecated-warning-deprecated.log.  So not sure which was desired/expected.  If it was the later, then I say forget the use of static and also trying to capture its "final" value.  The former, then suggest the use of static if it is so desired to save the few cycles that might be expected in normal operation (albeit ignoring notices could cause a savings in some environments).

drbyte:
Hmm...
Here are the filenames from mine today:
Screen Shot 2020-05-19 at 11 34 38 PM

mc12345678:
Based on "perusal" of that list, it appears that at most 2 types of log events occurred during any one site load. For files that end in log only, there should be notices only. For the files that end in warning or error, it is possible that notices follow the associated warning or error, but I would expect that no such notice would be logged before the associated warning or error... as to what I was referring and a little beyond. A single page load that generates two or more loggable events under the original code appears would create a file for each event with the name of the file extended for each type of event (except notices which would just get added to the then current file). That is if I understand the error logging system correctly. 

So like in the previous "example", if a single page load were to include a deprecated function followed by a warning and then another deprecated function, then I would expect the following "type" files.

myDebug-time1-deprecated.log
myDebug-time1-deprecated-warning.log
myDebug-time1-deprecated-warning-deprecated.log

With the proposed rewrite I would expect two files:
myDebug-time1-deprecated.log
myDebug-time1-warning.log

Any notices, may cause the addition of a file or be combined with the previous file.

scottcwilson:
Move out to 1.5.8?  This is getting mighty complicated.

lat9:
I was thinking about this as well.  Perhaps instead of plunking each log into its severity-based file, the progression is that once a severity-level has been reached, all future logs are recorded in that severity-level.

- If a PHP notice is originally received, that goes into a myDEBUG*.log file.  
- If a PHP deprecation is received, that log is placed into a myDEBUG*-deprecated.log file ... as are any follow-on PHP notices ... until ...
- If a PHP warning is received, that log is placed into a myDEBUG*-warning.log file ... as are any follow-on PHP notices and deprecations ... until ...
- If a PHP error is received, that log is placed into a myDEBUG*-error.log file.

That would (a) simplify the flip-flopping and (b) provide some ordering to the logs -- given that the sequence of any errors can be indicative.

mc12345678:
> I was thinking about this as well. Perhaps instead of plunking each log into its severity-based file, the progression is that once a severity-level has been reached, all future logs are recorded in that severity-level.
>
> - If a PHP notice is originally received, that goes into a myDEBUG*.log file.
> - If a PHP deprecation is received, that log is placed into a myDEBUG*-deprecated.log file ... as are any follow-on PHP notices ... until ...
> - If a PHP warning is received, that log is placed into a myDEBUG*-warning.log file ... as are any follow-on PHP notices and deprecations ... until ...
> - If a PHP error is received, that log is placed into a myDEBUG*-error.log file.
>
> That would (a) simplify the flip-flopping and (b) provide some ordering to the logs -- given that the sequence of any errors can be indicative.

That would be an improvement.

I ran an additional test that was not addressed in my original discussion of the issue, especially based on what DrByte had said/asked about there only being a difference in the number of process cycles between what was there and what was proposed.

The test involved tracking the status/condition of the `static` variable when it was also modified. So this code:
````php
<?php
include 'includes/application_top.php';

function testthis($data) {
  if (!isset($this_test)) {
    static $this_test = 'test';
    echo 'inside loop: ' . $data . '<br>';
    echo 'this-test: ' . $this_test . '<br>';
    $this_test = $this_test . '_' . $data;
  }
}

function usethis($data) {
  static $thisvar;
  if (!isset($thisvar)) {
    $thisvar = 'used';
    echo 'one time looped: ' . $data . '<br>';
    echo 'thisvar: ' . $thisvar . '<br>';
    $thisvar = $thisvar . '_' . $data;
  }
}

for ($i=1; $i <= 5; $i++) {
  testthis($i);
}
for ($i=1; $i <= 5; $i++) {
  usethis($i);
}

trigger_error('dep', E_USER_DEPRECATED);
trigger_error('war', E_USER_WARNING);
trigger_error('dep', E_USER_DEPRECATED);
trigger_error('not', E_USER_NOTICE);
trigger_error('dep', E_USER_DEPRECATED);
````
In both versions of the `zen_debug_error_handler` resulted in the following output and the same logs (date/time of course being different):
````
inside loop: 1
this-test: test
inside loop: 2
this-test: test_1
inside loop: 3
this-test: test_1_2
inside loop: 4
this-test: test_1_2_3
inside loop: 5
this-test: test_1_2_3_4
one time looped: 1
thisvar: used
````

Note how `this-test` is modified on each loop (doesn't start from scratch).  So the `static` call with a value assigned to it acts like an initiation situation on first call and then retains whatever value was "last" stored.  So in this situation where the only action to occur within the if was to provide an initial assignment, at least in this area of the code the if statement can be removed and the variable declared static with the initial value of `.log`.

As to the log reports generated by the above, well I had three logs, one that ended in `.log`, one that ended in `-deprecated.log` and one that ended in `-warning.log`.  The associated issue was captured in each log only with no real sequencing identified other than one deprecation was encountered before the next, but there was no real "sequence" other than alphabetical by directory sort to identify the "severity".

As to the proposed sequence and separation of the issues into some sort of tiered severity, the only other thing I see that will possibly become a support issue is helping others identify which log(s) are *informative*.  E.g. if a warning occurs before an error, the error log would contain the error itself and anything that follows (assuming that there is any further execution), but the issue that led to the error likely is in one or more of the previous logs...  This somewhat goes against the age old "look at the latest log" discussion.  A change, but provided there is benefit, so be it.

Seems that if the goal were to identify the "worst" issue that occurred during that execution that the logs not be written until full execution has been completed and the logname be identified to the worst case with all information captured into it... Otherwise the proposed "multi-log" situation could identify all the various types of issues that occurred and ideally provide some semblance of sequence of events...

**Comment by lat9** [comment]
IMO, what you're suggesting is way too complicated.  Imagine how long that log file name's going to be if there's a loop that, one each iteration, throws a deprecation followed by a warning.  That would result in a log named myDEBUG*-deprecated-warning-deprecated-warning-deprecated-warning[...].log.

You'd wind up with an _error log since the file's name would be longer than the OS allows.

**Comment by drbyte** [comment]
Lots of discussion here. That's good.

But I'm not clear what the original problem "is": in the day-to-day, what's the "symptom" we're trying to "fix" so that it matches the "desired result"? 
How is the "current day behavior" different from the "desired result", in the storeowner's experience?

**Comment by lat9** [comment]
> But I'm not clear what the original problem "is": in the day-to-day, what's the "symptom" we're trying to "fix" so that it matches the "desired result"?
> How is the "current day behavior" different from the "desired result", in the storeowner's experience?

The issue, IMO, is that in the current swizzling of the log-file's name that you lose the sequencing of errors.  Granted, they're all in a log somewhere, but you might have to open/review multiple log files to understand what happened first.

**Comment by drbyte** [comment]
> > But I'm not clear what the original problem "is": in the day-to-day, what's the "symptom" we're trying to "fix" so that it matches the "desired result"?
> > How is the "current day behavior" different from the "desired result", in the storeowner's experience?
> 
> The issue, IMO, is that in the current swizzling of the log-file's name that you lose the sequencing of errors. Granted, they're all in a log somewhere, but you might have to open/review multiple log files to understand what happened first.

Why don't we just rip out the suffixes then?

**Comment by lat9** [comment]
The suffixes were intended to give a visual hint regarding the severity of the issue(s) present in the log file, allowing site-maintainers to quickly hone in on the most problematic issue.

**Comment by scottcwilson** [comment]
This is not that beneficial and it's a bunch of extra complexity.   Plus we want to get *away* from log files to move to something more modern.

**Comment by lat9** [comment]
> 
> 
> This is not that beneficial and it's a bunch of extra complexity. Plus we want to get _away_ from log files to move to something more modern.

The proposal I posted above actually _simplifies_ the current logging infrastructure.

**Comment by drbyte** [comment]
Given the discussion that's ensued, this PR doesn't seem complete.
Plus, I'd rather avoid adding yet-another config switch. The prior PR added 3, why add a 4th?

Should we just revert the PR that this aims to "fix".

**Comment by scottcwilson** [comment]
YES

**Comment by drbyte** [comment]
> YES

Why? I perceive that your response has nothing to do with the subject matter at hand but rather is focused on shoving 157 out the door in haste.
I would prefer a reasoned response related to the value of reverting and why that's important to you and the end-users.

**Comment by scottcwilson** [comment]
This is fiddling around the margins without a business case.  I process logs EVERY MONTH from dozens of sites and it works fine.  There's no real need to change log processing.

**Comment by lat9** [comment]
FWIW, I find the severity suffix useful as it helps me hone in on the larger issues on high-maintenance sites.  I don't believe that we need yet-another configuration setting, just a teeny update to the error-log handling to sequence the log-file suffix instead of swizzling.

**Comment by mc12345678** [comment]
> FWIW, I find the severity suffix useful as it helps me hone in on the larger issues on high-maintenance sites. I don't believe that we need yet-another configuration setting, just a teeny update to the error-log handling to sequence the log-file suffix instead of swizzling.

I also remember @lat9 that there was a discussed and valid reason to provide some additional level of information at the log listing point, but without making a full effort of search I haven't seen it again in the last couple of days. It was a reason that the plugin associated with this change was modified.  

Btw, I still would prefer to see an attempt to create a single log file for a series/sequence of issues during a page load, but the filename of that log be the *"highest"* level of the issue encountered instead of the generic "yes something was logged" type filename.  Reason for the single filename as discussed earlier is that the issue that led to the actual problem is normally in the entry(ies) just prior to the significant issue. Having the filename switch just *after* the "cause" of the higher level problem then requires review of two or more files. For those in the know it is not a problem, but will add to the support/explanation load.

**Comment by scottcwilson** [comment]
> Btw, I still would prefer to see an attempt to create a single log file for a series/sequence of issues during a page load...

This is not a good idea.  Logs are often repeated many, many times, and keeping them separate means I can fix it once and then delete *all the other* instances of that log with a single command.

**Comment by mc12345678** [comment]
> This is not a good idea. Logs are often repeated many, many times, and keeping them separate means I can fix it once and then delete all the other instances of that log with a single command.

What is it that I am or am not saying that continues to get misunderstood?

What I am saying is to continue to log the same data that is currently being logged but use a filename that has an understandable recognition of what the worst problem encountered was in the issues captured in that log file...

**Comment by drbyte** [comment]
> ... use a filename that has an understandable recognition of what the worst problem encountered was in the issues captured in that log file

Is that actually possible? Given the filename is determined before the actual errors are triggered, I'm not sure you can even do that.

**Comment by torvista** [comment]
>Given the filename is determined before the actual errors are triggered, I'm not sure you can even do that.

Shame.
For the admin debug name, there are 34 characters used only to identify the admin and the timestamp. Seems a lot.

imagine:
a20200522123459-N_undefined_index-FIRSTNAME.php
c20200522123505_N_undefined_variable-codepage.php

I think shop errors are often multiples so a method to group them for subsequent deletion would be good. Since Display Logs can sort, I sort by size and delete those in the same range.

**Comment by mc12345678** [comment]
> Is that actually possible? Given the filename is determined before the actual errors are triggered, I'm not sure you can even do that.

That was where my previous statement:
> … that the logs not be written until full execution has been completed and the logname be identified to the worst case with all information captured into it... 

Played a part... I haven't identified what it would take to recognize that execution was complete to support then dumping the data that was expected to be captured seeing as currently, like said, a filename is identified based on issue to be captured, the data is dumped, execution continues with some level of evaluation of the next issue to be logged and filename identified and the data then dumped... I could see the possibility of everything logged into a single file, the stub (date/time) known, then if the "end" can be identified that the base file get copied to a file that includes the stub and suffix recognizing the worst case problem (or to rename that original file as execution comes along to identify the next highest level error) (This last part seems feasible provided the system to which it is installed allows/supports file renaming).

So execution would be along the lines as lat9 suggested: 
start with a base name.
evaluate the issue, if it is next highest level then
  identify new file extension,
  rename existing file to that new file extension,
in either case (next highest, same or lower)
  log new message.


This would retain the data in a single file (again assuming associated server/server setup supports the file renaming) and the filename itself would identify the highest level of problem.

**Comment by drbyte** [comment]
Aside: This is where sending logs to a log-aggregation service/process makes things so much easier, which I'm exploring for 1.5.8. Building an entire log-management system for tracking PHP errors feels like it's a responsibility that should be outside this framework.

**Comment by lat9** [comment]
Please note that the _last thing_ you want in an error-handler is _complexity_.  That file-renaming suggestion opens the processing to additional file-related errors.

**Comment by mc12345678** [comment]
> That file-renaming suggestion opens the processing to additional file-related errors.

Not denying that. :)

So unasked question, what tracked/observed event is the last thing that execution sees when there is a fatal or non-fatal error to be able to capture/store the issue(s) until that event has occurred, or is this back to the "log-aggregation service/process" issue that is under development?

Note, I don't know what the final product is expected to be for this "something *modern*" system, but I think of log files (written data rather than something `databasy`) to be easy(ier) to handle and access.  Reaching them does not depend on php versions, database versions or other dependencies. Simple to just login to wherever/whatever has read access... There's no updating of anything in order to get to them. Yes, something that might parse them and provide some informative information would be great, but when the site is stuck, the log(s) tend to hold the key(s).

**Comment by drbyte** [comment]
> I don't know what the final product is expected to be for this "something _modern_" system

Envisioning a plugin that integrates sending logs to sentry.io or rollbar.com or honeybadger.io
It wouldn't necessarily do-away-with local file logging, but one would have the option to do so, by simply removing the current error handler.

File: includes/extra_configures/enable_error_logging.php [hunk 1]
```diff
@@ -21,8 +21,9 @@ function zen_debug_error_handler($errno, $errstr, $errfile, $errline)
         return;
     }
 
+    static $last_log_suffix;
     if (!isset($last_log_suffix)) {
-        static $last_log_suffix = '.log';
```
mc12345678, line 25: As otherwise described in the chain of this conversation, an appropriate modification here to address just the use of `static` and there being only one initial action to be taken when this method is called would be to simply remove the `if` statement and leave the line to initially declare the `static` variable to whatever suffix is to initially be used.  Upon each call of this method during the current session, the last known value of the variable will be retained and ***not reassigned*** to the initial variable.

By placing the assignment within the if statement allows for other/additional action to be taken the first time the variable is declared (or if it somehow becomes `null` in iterative processing).

Although at the time of suggesting this edit I apparently did not fully understand that the `static` variable would not be reassigned to the original value, I *am* ok with it still being in the `if` statement as proposed here if this commit remains or is resurrected. This would allow others to "easily" incorporate additional first time actions into this section rather than having to incorporate/add it later... Just saying. :)

drbyte: I'm fine with relocating the `static` declaration to outside the `if` statement.

drbyte:
Still not a fan of the 4th new config switch.

mc12345678:
> So unasked question, what tracked/observed event is the last thing that execution sees when there is a fatal or non-fatal error to be able to capture/store the issue(s) until that event has occurred, or is this back to the "log-aggregation service/process" issue that is under development?

A little research into this file, the implemented methods, and found the answer to my question.

During execution, issues are sent through `zen_debug_error_handler`.  When execution ends (although the [php manual](https://www.php.net/manual/en/function.register-shutdown-function.php) seems to identify one exception associated to the use of SIGKILL even though it seems that ZC doesn't appear to deal with SIGTERM) then `zen_fatal_error_handler` is called. Now, completion of execution doesn't necessarily mean that it is fatal, but it is within this method that fatal errors are currently handled.

Anyways, the data collected by `zen_debug_error_handler` currently is put to file within that method, but it could just as well be collected (database, file, variable of some sort) and then upon call to `zen_fatal_error_handler` which would occur at successful completion, call to `zen_exit()`, `exit()`, or upon occurrence of an actual fatal error, the data previously collected could be written to file (some sort of if statement before the existing if statement) to include the filename that was likely tracked/identified in `zen_debug_error_handler` (and/or as necessary if it is a fatal error as identified during processing of the request in this method) and then further if the associated error/issue is actually a fatal error then the final issue be logged with the "appropriate" name.

Potential issues with this is that instead of the data being logged "immediately", their "value" would be accessible along program execution, whether this is good or bad I am not saying, but it would not be immutable.

> Still not a fan of the 4th new config switch.

Not really either, but was a way to provide/ensure some "normality" of either allowing the logged issues to be split out in whatever determined manner or to keep them all in one place, regardless the filename.

drbyte:
> > I don't know what the final product is expected to be for this "something _modern_" system
> 
> Envisioning (actually, I already have and use) a plugin that integrates sending logs to sentry.io or rollbar.com or honeybadger.io

Or, ideally, anything that Monolog supports: https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

proseLA:
i'm a little late to this party.  i'm not a fan.  i suppose my sites are not "high maintenance" sites; either that, or i'm a little OCD.  i see a log file, i address the problem and delete the log.  for me, i see little value for this additional complexity; and i further agree that it is really outside the scope of this framework.

the original integration of the ZC logging feature was great; because it provided logs that were sometimes rather difficult to get from your host.  but building additional infrastructure around them?

i am not disputing that this change would have value for others; but for me no.

drbyte:
@mc12345678 I'll merge this if you rebase without the commit that adds the extra config switch.

mc12345678:
> @mc12345678 I'll merge this if you rebase without the commit that adds the extra config switch.

Sure. Hope that someone has some other improvement in mind, whether expanding the options of an existing switch or at least modification of how the logging is done in this release. Pulling out the issues without some sort of identified sequence of occurrence reduces the benefit offered by logging the issues. I'm not going to hold up this PR just for that especially when it was the original static issue that really got me digging into operation.

scottcwilson:
If this gets merged, I really need one of you to update the documentation please. 

https://docs.zen-cart.com/user/troubleshooting/debug_logs/

lat9:
> 
> 
> > @mc12345678 I'll merge this if you rebase without the commit that adds the extra config switch.
> 
> Sure. Hope that someone has some other improvement in mind, whether expanding the options of an existing switch or at least modification of how the logging is done in this release. Pulling out the issues without some sort of identified sequence of occurrence reduces the benefit offered by logging the issues. I'm not going to hold up this PR just for that especially when it was the original static issue that really got me digging into operation.

I'm still trying to understand why you think there was an issue with that `static` definition.

mc12345678:
> > > @mc12345678 I'll merge this if you rebase without the commit that adds the extra config switch.
>>>
> > Sure. Hope that someone has some other improvement in mind, whether expanding the options of an existing switch or at least modification of how the logging is done in this release. Pulling out the issues without some sort of identified sequence of occurrence reduces the benefit offered by logging the issues. I'm not going to hold up this PR just for that especially when it was the original static issue that really got me digging into operation.
>>
> I'm still trying to understand why you think there was an issue with that static definition.

During first execution:
````php
function beingcalled() {
  if (empty($var)) { // <--- $var is not declared, defined, set, or even known at this point so is true.
    static $var = 'initial-static-val'; // <---- Establishes the variable $var as static and because this is the first occurrence of this establishes the value as 'initial-static-varl'.
  }
  $var = new_val_function($var); // <---- Assigns some new value to $var, how/what is not really important here.
}
````
Now, on second and subsequent call to this function:
````php
function beingcalled() {
  if (empty($var)) { // <--- $var is not declared, defined, set, or even known at this point so is true. (Executed again and then on every call to this function and therefore unnecessary as always true)
    static $var = 'initial-static-val'; // <---- Recognizes the variable $var is static, retrieves the last value and ignores this assignment.
  }
  $var = new_val_function($var); // <---- Assigns some new value to $var, how/what is not really important here.
}
````

So my thing was the use of the if statement which always evaluated to true before the use of the static statement. I then proposed revision that more appropriately identified that the variable was assigned or not.  I then discovered that if the if statement were simply removed from around the static assignment that would be work as expected as well. So either of the below function rewrites would work and not have some wasted if statement:
````php
function beingcalled() {
  static $var = 'initial-static-val'; // <---- Recognizes the variable $var is static, retrieves the last value and ignores this assignment.

  $var = new_val_function($var); // <---- Assigns some new value to $var, how/what is not really important here.
}
````
or:
````php
function beingcalled() {
  static $var;
  if (empty($var)) { // <--- $var is not declared, defined, set, or even known at this point so is true. (Executed again and then on every call to this function and therefore unnecessary as is always true)
    $var = 'initial-static-val'; // <---- Recognizes the variable $var is static, retrieves the last value and ignores this assignment.
  }
  $var = new_val_function($var); // <---- Assigns some new value to $var, how/what is not really important here.
}
````

Otherwise, what prevents the code from being something like this:
````php
function beingcalled() {
  if (true) {
    if (true) {
      if (true) {
        if (empty($var)) { // <--- $var is not declared, defined, set, or even known at this point so is true. (Executed again and then on every call to this function and therefore unnecessary as always true)
          static $var = 'initial-static-val'; // <---- Recognizes the variable $var is static, retrieves the last value and ignores this assignment.
        }
      }
    }
  }
  $var = new_val_function($var); // <---- Assigns some new value to $var, how/what is not really important here.
}
````

PR #3505, opened 2020-05-20
Display order comments on hover

drbyte:
Screen Shot 2020-05-20 at 12 03 52 AM

Feedback welcome


torvista:
When there is a comment, the yellow icon alt should also use the same text as the title: otherwise the image alt text ("Customer Comments - YES") overrides the containing title.

When there is no comment, the empty title should be suppressed.

PR #3509, opened 2020-05-20
Admin - Load localized CSS overrides if present

drbyte:
This allows one to "drop in" a Bootstrap 3 theme replacement (`bootstrap.min.css` file) from one of the many 3rd party theming sites.

One can also delete the local file if CDN loading is preferred.

### DIFF
--- a/admin/includes/admin_html_head.php
+++ b/admin/includes/admin_html_head.php
@@ -14,9 +14,21 @@ if (!defined('IS_ADMIN_FLAG')) {
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <title><?php echo TITLE; ?></title>
+<?php if (file_exists($file = 'includes/css/bootstrap.min.css')) { ?>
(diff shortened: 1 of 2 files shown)


Zen4All:
Or use an if statement.
If local bootstrap.min.css exists -> USE
else load cdn

drbyte:
If Bootstrap would have embraced their "theme override" stuff this would have been easier.
They want everyone to recompile their own complete bespoke file instead.
I get it, but it makes it harder to accommodate wide varieties of taste.

drbyte:
> Or use an if statement.
> If local bootstrap.min.css exists -> USE
> else load cdn

I dunno.

I've gone back and forth about this several times: which to load first in the if/else?
If we load "local first", then we should DELETE the distributed local files, right?
 - Actually I'd prefer to do that, but what will the reaction be?

Zen4All:
Hmmm. Then leave it as is now in your pr. It's ugly, but apparently the "best" way.

drbyte:
I agree: it's ugly.

Still not decided on the "best way"...

drbyte:
Updated: Now loads local file if present, else loads from CDN.

PR #3515, opened 2020-05-21
Remove Link from Store Name in Footer.

dbltoe:
Not only does the target="blank" set off accessibility alarms, I'm wondering why we would need for a customer to go to our site when they're already on it.   Much less, going where they already are in a new window!

### DIFF
--- a/includes/languages/english.php
+++ b/includes/languages/english.php
@@ -13,7 +13,7 @@
 //define('CUSTOM_KEYWORDS', 'ecommerce, open source, shop, online shopping');
 // END: moved to meta_tags.php
 
-  define('FOOTER_TEXT_BODY', 'Copyright &copy; ' . date('Y') . ' <a href="' . zen_href_link(FILENAME_DEFAULT) . '" rel="noopener" target="_blank">' . STORE_NAME . '</a>. Powered by <a href="https://www.zen-cart.com" rel="noopener" target="_blank">Zen Cart</a>');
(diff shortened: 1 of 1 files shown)


drbyte:
The point of the clickable link is that it ties the URL to the site-name as part of the branding and copyright. Beneficial when the page is indexed in search engines.
It also doubles as a "home" link when lost in obscure pages of the site, although one can of course argue that there are usually other "home" links on the page.

I agree that there's no need to use target-blank on this link. 
But not a fan of removing the link.

dbltoe:
My main goal was to avoid the accessibility hit for the target="_blank".  Been checking today and adding both noopener and noreferrrer to all offsite target="_blank" calls will result in us meeting 508 Accessibility standards for the responsive_classic template.  Not 100% sure that there might be something on an obscure page but, confident enough to say we meet accessibility standards.

I just wondered why I might need a link and didn't pull "lost in space" out of my brain.
Good point.

drbyte:
How is "noreferrrer" an "accessibility" thing?

dbltoe:
The 508 Accessibility test that I use actually recommends a warning that the user is leaving the site but will accept noopener if going to trusted site (i.e. cart to zen-cart.com.  It recommends both in cases of offsite to any site over which the cart owner has no control (i.e, someone can create a phishery on any site we link to without control over the linked site's content.

drbyte:
Hmmm. Maybe I'm missing something about the phishing piece.

By not passing the referrer then SEO benefits are lost. Google's analytics will show the clicks as coming from "unknown" instead of from a specific source.
Also lost is legit ad-based cross-linking tracking (affiliate relationships included).

Hmmm...

dbltoe:
As mentioned, it is not a requirement for compliance.  It's a suggestion for use when going to "A source over which you have no control or might change ownership" (my paraphrase)

PR #3529, opened 2020-05-24
Add Screen Reader info to 3 listing pages for accessibility

### LINKED ISSUE #3528: Help with input labels on all, new, and featured
My original "test" site was all priced by so I missed some labels.  Trying to work up a PR but brain dead on code.

For all, new, and featured all listings.   Where productID = 117, the goal is:

Add: `<label for="products_id[117]">products_id[117]</label><input type="text" name="products_id[117]" value="0" size="4">`

Proposed below is not correct.  Looking for assistance and correction of line 91 of includes/templates/responsive_classic/templates/tpl_modules_products_all_listing.php.
 
```php
 $link = TEXT_PRODUCT_ALL_LISTING_MULTIPLE_ADD_TO_CART . 
"<label for=" . $products_all->fields['products_id'] . "> . $products_all->fields['products_id'] . "</label>
<input type=\"text\" name=\"products_id[" . $products_all->fields['products_id'] . "]\" value=\"0\" size=\"4\" />";
```

scottcwilson:
Fixes #3528

### DIFF
--- a/includes/templates/responsive_classic/templates/tpl_modules_products_all_listing.php
+++ b/includes/templates/responsive_classic/templates/tpl_modules_products_all_listing.php
@@ -88,7 +88,7 @@
 //          $link= '<a href="' . zen_href_link(FILENAME_PRODUCTS_ALL, zen_get_all_get_params(array('action')) . 'action=buy_now&products_id=' . $products_all->fields['products_id']) . '">' . zen_image_button(BUTTON_IMAGE_IN_CART, BUTTON_IN_CART_ALT) . '</a>';
           if (PRODUCT_ALL_LISTING_MULTIPLE_ADD_TO_CART > 0 && $products_all->fields['products_qty_box_status'] != 0) {
 //            $how_many++;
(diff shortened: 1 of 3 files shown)


drbyte:
Why `title` instead of `aria-label`?

scottcwilson:
It was suggested by one of the help pages, but it's easy enough to change, and aria-label works too.

drbyte:
I'm struggling to accept this new function. It feels too highly specialized, especially given it's only on pages that are basically clones of each other, and should be collapsed into one.

I'd rather just see the tags added directly where they're needed, as we've done elsewhere..

drbyte:
... especially given we also have a shared language define for this purpose as well.

scottcwilson:
Why don't we merge this now and eliminate the function when the templates DO get merged.  I wanted to consolidate the function in case someone asked for a tweak.

drbyte:
> I wanted to consolidate the function in case someone asked for a tweak.

The sentiment is respected and appreciated. Thanks for thinking ahead for the end-users, and the fact that they always want to change something! :)

> Why don't we merge this now and eliminate the function when the templates DO get merged. 

Because:
- everyone complains when functions they used to use have disappeared
- removing functions that are called by templates makes those templates incompatible
- all the aria labeling already done in the template has been using direct tags, not a generically-named function that is being used in a very specialized way (eg: expects it to be related to products)

PR #3532, opened 2020-05-25
Fix search form in attributes controller

### LINKED ISSUE #3530: Search function in attributes controller seems to be useless
While working on some last minute UI updates, I was testing the search box.
No matter what i type in, the search always returns empty.
Furthermore, the file name in the form goes to category_product_listing, instead of attributes_controller.
Working on a PR to solve this.

scottcwilson:
Fixes #3530

### DIFF
--- a/admin/attributes_controller.php
+++ b/admin/attributes_controller.php
@@ -775,6 +775,7 @@ function zen_js_option_values_list($selectedName, $fieldName)
           <?php
           echo zen_draw_form('searchForm', FILENAME_CATEGORY_PRODUCT_LISTING, '', 'get');
 // show reset search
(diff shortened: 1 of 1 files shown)


scottcwilson:
It could be that a more comprehensive fix would be to change the admin copy of zen_draw_form from doing 

```     
$form .= '<input type="hidden" name="cmd" value="' . (isset($_GET['cmd']) ? $_GET['cmd'] : 'home') . '">';
```

to

```     
$form .= '<input type="hidden" name="cmd" value="' . $action .  '">';
```

drbyte:
Will keep the former behavior for now.

PR #3535, opened 2020-05-27
"Invoice" to "Order"

torvista:
as per 
https://www.zen-cart.com/showthread.php?226687-Use-of-quot-order-quot-instead-of-quot-invoice-quot&p=1368580#post1368580

### DIFF
--- a/admin/includes/languages/english/invoice.php
+++ b/admin/includes/languages/english/invoice.php
@@ -20,7 +20,6 @@ define('TABLE_HEADING_PRICE_INCLUDING_TAX', 'Price (inc)');
 define('TABLE_HEADING_TOTAL_EXCLUDING_TAX', 'Total (ex)');
 define('TABLE_HEADING_TOTAL_INCLUDING_TAX', 'Total (inc)');
 define('TABLE_HEADING_PRICE', 'Price');
-define('TABLE_HEADING_TOTAL', 'Total');
 define('ENTRY_CUSTOMER', 'CUSTOMER:');
 
 define('ENTRY_SOLD_TO', 'SOLD TO:');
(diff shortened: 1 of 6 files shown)


scottcwilson:
The button says "Invoice" and the labels on the page should be consistent.  This should not be merged.

includes/languages/english/checkout_process.php
```diff
@@ -14,7 +14,7 @@
 define('EMAIL_DETAILS_FOLLOW','The following are the details of your order.');
 define('EMAIL_TEXT_ORDER_NUMBER', 'Order Number:');
 define('EMAIL_TEXT_INVOICE_URL', 'Detailed Invoice:');
```
drbyte, line 16: Was this intentionally left unchanged?

admin/includes/languages/english/invoice.php
```diff
@@ -29,5 +29,5 @@
 define('ENTRY_SHIPPING', 'Shipping:');
 define('ENTRY_DATE_PURCHASED', 'Date Ordered:');
 
-define('ENTRY_ORDER_ID','Invoice No. ');
+define('ENTRY_ORDER_ID','Order No. ');
```
drbyte, line 32: Was this same change in packingslip.php intentionally skipped?

zcwilt:
Guess I should add as well, that strictly speaking an Order id and an Invoice/Receipt number should really be two separate entities and while Zen Cart doesn't support that, it's something we should look to do.

torvista:
As I have mentioned ad nauseum, I think an "invoice" is a legal term, requiring specific items to qualify as an "invoice" legally, per-country definitions. Even "ticket" in spain is classed as a special type of anaonymous invoice with its own specification of what it must include to be legal.

PR #3538, opened 2020-05-27
paypal payment modules admin texts

torvista:
deprecated tags, br / to br, hr / to hr, spaces in spans moved  to after spans


scottcwilson:
These changes are creating a ton of upgrade work for people who have customized these language files for very little benefit.  Can we wait until there is an actual requirement for a change to the file before doing tweaks like this?

torvista:
I don't believe there is any obligation to use these when presented, they can be put on hold as appropriate no?

scottcwilson:
You are imposing work on other people for no clear benefit.  "Putting PRs on hold" (and then monitoring the hold queue and then resolving merge conflicts, etc.) takes *work*.   

The core committers do not have unlimited time.  Wouldn't it be better if they used their time available for making the product and doing a release rather than managing and merging PRs that don't even fix bugs?

torvista:
Can a current version of this go into 158?

PR #3544, opened 2020-05-29
Check needs to be inverted

### LINKED ISSUE #3543: Can't add/update product with new product.php
Was having problems with alpha and thought it might be the DB.  Meanwhile, beta arrives.
Load beta and demo with nothing else.
Go to Categories/Products
Click on hardware, then CDROM Drives.
Click New Product.
When info is added click Preview
Blank form and error.
adminProduct

scottcwilson:
Fixes #3543

### DIFF
--- a/admin/product.php
+++ b/admin/product.php
@@ -36,7 +36,7 @@ if (zen_not_null($action)) {
       break;
     case 'new_product_preview':
       if (!isset($_POST['master_categories_id'])
-          || ((isset($_POST['products_model']) ? $_POST['products_model'] : '') . (isset($_POST['products_url']) ? implode('', $_POST['products_url']) : '') . (isset($_POST['products_name']) ? implode('', $_POST['products_name']) : '') . (isset($_POST['products_description']) ? implode('', $_POST['products_description']) : '') != '')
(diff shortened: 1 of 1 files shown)


zcwilt:
Still not convinced this is correct. 
If I got to the new product page, don't fill in any entries and immediately hit preview, then I will get the warning message which is confusing in what it is warning about.

scottcwilson:
When I edit product 21 in the test data set with this fix, I don't get the issue.
What product are you editing?  Or are you doing a new product?

zcwilt:
New Product

zcwilt:
Just to clarify, there's nothing inherently wrong in warning that say the name entry is empty, although we never did it before. It's just then that the message displayed is ambiguous.

scottcwilson:
It looks ok to me - it says your data is empty.  Which it is!

zcwilt:
Yes, but also says that you may have a problem with bowser or internet.

scottcwilson:
Which you may! :)
What do we need to do to wrap this up - do you want me to remove that language?

zcwilt:
I'm probably being too anal.

PR #3548, opened 2020-05-30
Base Font Size Too Small for Readability

dbltoe:
While there is no specific written standard for font-size, it is considered to be 12pt or 16px.  At the original 72.5% and the fact that most alerts and small text are set to em (some as low as .5em), readability is taking a big hit on the mobiles and things like the site legal info.
Making the setting 80% at least gets us to 10pt and triggers no errors for font too small while still not overpowering the screen.

### DIFF
--- a/includes/templates/responsive_classic/css/stylesheet.css
+++ b/includes/templates/responsive_classic/css/stylesheet.css
@@ -9,7 +9,7 @@
  */
 
 /*bof basic elements*/
-body {margin:0;font-family:verdana, arial, helvetica, sans-serif;font-size:72.5%;line-height:140%;}
+body {margin:0;font-family:verdana, arial, helvetica, sans-serif;font-size:80%;line-height:140%;}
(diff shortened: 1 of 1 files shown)


drbyte:
How much have you tested the impact of this on the large majority of typical screen sizes? 
To what degree is text now going to be too big to fit in its usual places without wrapping and creating completely new usability problems such as chopped-off text and line-wrapping of controls?

dbltoe:
With the exception of the navBreadCrumb, I didn't find any glaring problems  However, on further looking at it today, I have found things are more of a mess than I thought css-wise.  There are no less than four separate css calls to make the mainWrapper 100%.  It's already set for that in the tpl.

I think the better thing to do is just close this and concentrate on 1.5.8 unless a miracle occurs before 1.5.7 release.  I wish I had time to do this earlier in 1.5.7.

It will be a complete workover with decrease in css footprint.

PR #3549, opened 2020-05-31
Prevent Non-Secure Link on Main Page

dbltoe:
The use of // versus http://www. allows the link to be presented in the settings of the remote site.
Stops the browser warnings for mixed content.


mc12345678:
See comment left in PR #3550 as it applies here as well.

PR #3550, opened 2020-05-31
Prevent Non-Secure Link on Main Page

dbltoe:
The use of // versus http://www. allows the link to be presented in the settings of the remote site.
Stops the browser warnings for mixed content.


includes/languages/english/html_includes/define_main_page.php
```diff
@@ -1,4 +1,4 @@
-<a href="http://www.zen-cart.com/book"><img src="images/large/e-start-book.gif" alt="get your manual today" title="Have you got yours yet? Join the 1000's of Zen Cart users that have bought the only comprehensive owners manual !" /></a>
+<a href="//zen-cart.com/book"><img src="images/large/e-start-book.gif" alt="get your manual today" title="Have you got yours yet? Join the 1000's of Zen Cart users that have bought the only comprehensive owners manual !" /></a>
```
mc12345678, line 1: Considering that the destination site permits `https`, this issue is expected to be resolved if `https:` were used rather than the non-assigned protocol.  I've been advised that use of `https:` in the path will not cause a mixed content message even if the page were presented using `http:` as the concern is providing non-secure information on a secure page. Providing secure content on an insecure page is considered good if not some amount of an improvement...

drbyte:
The non-assigned protocol would be based on "my own store's current protocol".
So I wouldn't use just `//` in this case. 
The right fix is `https`, because the destination site's "real" URL is indeed https.

PR #3556, opened 2020-06-02
Shorten GET param for new ask_a_question module

drbyte:
`&products_id=` is now `&pid=`


mc12345678:
Other than shortening the parameter, is there some other reason to do this? 

Doesn't AAQ now need a sanitization similar to what was performed with the `products_id` parameter to validate that the question is to/about a valid product? Changing to `pid` now bypasses that check.

drbyte:
> validate that the question is to/about a valid product?

It already does: includes/modules/pages/ask_a_question/header_php.php

PR #3560, opened 2020-06-03
init() does not exist anymore

Zen4All:


Zen4All:
> As a general comment, that missing javascript function is going to break many existing plugins.

why?

scottcwilson:
Because they're based on an older admin file which *did* have an init() function, and it will take time to get them updated.  Suggest putting a stub in.

Zen4All:
But if an add on uses the skeleton file it is a new or updated addon, which should no longer need init(). Or am I  missing something?

scottcwilson:
Sorry, I misunderstood.   We *do* have a stub for `init()` in `admin/includes/menu.js`. I thought you were saying that this had been removed. 

@lat9 you good too?

lat9:
> 
> 
> Sorry, I misunderstood. We _do_ have a stub for `init()` in `admin/includes/menu.js`. I thought you were saying that this had been removed.
> 
> @lat9 you good too?

Yes, if that skeleton exists.  Otherwise, the older plugins will get a javascript error for that missing function and, if they're using any type of jQuery processing (like [Zen Tags](https://github.com/lat9/zentags)) it'll result in 'pollution' of the javascript environment.

Zen4All:
> > Sorry, I misunderstood. We _do_ have a stub for `init()` in `admin/includes/menu.js`. I thought you were saying that this had been removed.
> > @lat9 you good too?
> 
> Yes, if that skeleton exists. Otherwise, the older plugins will get a javascript error for that missing function and, if they're using any type of jQuery processing (like [Zen Tags](https://github.com/lat9/zentags)) it'll result in 'pollution' of the javascript environment.

Where does this use init() ? Sorry, but I still don't understand your concern.

lat9:
> 
> 
> > > Sorry, I misunderstood. We _do_ have a stub for `init()` in `admin/includes/menu.js`. I thought you were saying that this had been removed.
> > > @lat9 you good too?
> > 
> > 
> > Yes, if that skeleton exists. Otherwise, the older plugins will get a javascript error for that missing function and, if they're using any type of jQuery processing (like [Zen Tags](https://github.com/lat9/zentags)) it'll result in 'pollution' of the javascript environment.
> 
> Where does this use init() ? Sorry, but I still don't understand your concern.

You're right, that plugin doesn't present a page in-and-of-itself, but I've got a bunch of site-specific customizations that (a) use jQuery handlers and (b) use the `onload=init()` construct.

zcwilt:
Just a note here as well .
The skeleton file isn't actually used anywhere. It was meant to provide a file that a plugin author could copy and use for a tableView type page. As such it may well change, with some extra doc comments etc. for plugin authors.

Zen4All:
@zcwilt Off topic question: Does the new manager still check for updated modules on the ZC site?

drbyte:
> I've got a bunch of site-specific customizations that (a) use jQuery handlers and (b) use the `onload=init()` construct.

If you're already using jQuery on the page, could the `onload...init()` be triggered by a page-specific javascript block like:
```js
$(function() {
 // 
});
```
Ref: https://learn.jquery.com/using-jquery-core/document-ready/
or shorter syntaxes like: 
https://www.sitepoint.com/jquery-document-ready-plain-javascript/

or if `init()` is still highly page-specific and you don't want to use jQuery to trigger it, an alternative to `body ... onload=init()` is a js block to trigger the function. You could rename it to something other than `init()` too.
eg: `window.onload = init` or rename it to `foo()`: `window.onload = foo`
Ref: https://stackoverflow.com/questions/4842590/how-to-run-a-function-when-the-page-is-loaded

zcwilt:
@Zen4All 
> @zcwilt Off topic question: Does the new manager still check for updated modules on the ZC site?

Although off topic, will quickly answer.
The short answer is no. The current version checker relies on knowing the plugin id of the contribution which presents a chicken egg situation. @drbyte and I have discussed this and are working on a solution (probably involving using the plugin key/name)

drbyte:
>> Does the new manager still check for updated modules on the ZC site?
> The short answer is no. 

While "no" is the correct answer, the question isn't entirely correct. Plugins currently only check for updates on the ZC site if the plugin itself has code in it to do the checking. It's currently on the plugin to do the checking. It just happens to be that the checking normally takes places in the admin `modules` screen.

We envision adding checker logic to the new Plugins Manager. If the plugin's manifest file contains its plugin ID then (I think, for now anyway, I might change my mind later) it's probably future-proofing itself so it's ready for automatic alerting about new available versions.

> The current version checker relies on knowing the plugin id of the contribution which presents a chicken egg situation

FYI: This has always been the case. We're exploring dealing with this pain point separately.

Zen4All:
Thanks both for the explanation

PR #3561, opened 2020-06-03
Admin: Add ability to handle multiple per-page javascript/css files

Zen4All:
This update adds the ability to load multiple javascript and/or css files. The have to start with the page name you want the file to load on.
For example ```admin/includes/javascript/product_extraJscript.js``` will now load on product.php.
The modification will make it easier for mod makers to include their css or javascript to a page.
Also added the ability to load php javascript files

### DIFF
--- a/admin/includes/admin_html_head.php
+++ b/admin/includes/admin_html_head.php
@@ -33,13 +33,13 @@ if (!defined('IS_ADMIN_FLAG')) {
     <link rel="stylesheet" href="includes/css/stylesheet.css">
 <?php if (file_exists($file = 'includes/css/' . basename($PHP_SELF, '.php') . '.css')) { ?>
     <link rel="stylesheet" href="<?php echo $file; ?>">
-  <?php
-}
+<?php
+  }
 $directory_array = $template->get_template_part('includes/css/', '/^' . basename($PHP_SELF, '.php') . '_/', '.css');
 foreach ($directory_array as $key => $value) {
-  ?>
+?>
     <link rel="stylesheet" href="includes/css/<?php echo $value; ?>">
-  <?php
+<?php
 }
 
 // pull in any necessary JS for the page
--- a/admin/includes/admin_html_head.php
+++ b/admin/includes/admin_html_head.php
@@ -34,7 +34,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 <?php if (file_exists($file = 'includes/css/' . basename($PHP_SELF, '.php') . '.css')) { ?>
     <link rel="stylesheet" href="<?php echo $file; ?>">
 <?php
-  }
+}
 $directory_array = $template->get_template_part('includes/css/', '/^' . basename($PHP_SELF, '.php') . '_/', '.css');
 foreach ($directory_array as $key => $value) {
 ?>
--- a/admin/includes/admin_html_head.php
+++ b/admin/includes/admin_html_head.php
@@ -36,7 +36,6 @@ if (!defined('IS_ADMIN_FLAG')) {
 <?php
 }
 
-$template = new template_func();
 $directory_array = $template->get_template_part('includes/css/', '/^' . basename($PHP_SELF, '.php') . '_/', '.css');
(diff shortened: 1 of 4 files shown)


admin/includes/admin_html_head.php
```diff
@@ -14,26 +14,33 @@
...
+<link rel="stylesheet" href="includes/css/menu.css">
+<link rel="stylesheet" href="includes/css/stylesheet.css">
 <?php if (file_exists($file = 'includes/css/' . basename($PHP_SELF, '.php') . '.css')) { ?>
-    <link rel="stylesheet" href="<?php echo $file; ?>">
-<?php } ?>
+  <link rel="stylesheet" href="<?php echo $file; ?>">
 <?php
+}
+
+$template = new template_func();
```
drbyte, line 39: Isn't `$template` already instantiated by the config loader?

admin/includes/javascript_loader.php
```diff
@@ -31,6 +31,24 @@
 </script>
 <?php if (file_exists($jsFile = 'includes/javascript/' . basename($PHP_SELF, '.php') . '.js')) { ?>
   <script src="<?php echo $jsFile; ?>"></script>
+  <?php
+}
+if (file_exists($jsFile = 'includes/javascript/' . basename($PHP_SELF, '.php') . '.php')) {
+  require 'includes/javascript/' . basename($PHP_SELF, '.php') . '.php';
+}
+$template = new template_func();
```
drbyte, line 39: Since `javascript_loader.php` is called directly from `admin_html_head.php` which is already using `$template`, do we need to re-instantiate it again here?

admin/includes/javascript_loader.php
```diff
@@ -31,6 +31,24 @@
...
+  echo '<script src="includes/javascript/' . $value . '"></script>' . "\n";
+}
+$directory_array = $template->get_template_part('includes/javascript/', '/^' . basename($PHP_SELF, '.php') . '_/', '.php');
+foreach ($directory_array as $key => $value) {
+  require 'includes/javascript/' . $value;
+  echo "\n";
+}
+?>
+<?php if (file_exists($jsFile = 'includes/javascript/' . basename($PHP_SELF, '.php') . '.js')) { ?>
+  <!--  <script src="<?php echo $jsFile; ?>"></script> -->
```
drbyte, line 51: Is there intent behind duplicating lines 32-33 here?

Zen4All: nope, should have removed that

admin/includes/admin_html_head.php
Zen4All: Didn't know that.
 lines are removed

admin/includes/javascript_loader.php
Zen4All: no

drbyte:
I wanna test this a bit, and chat with @zcwilt tomorrow about it.

In the meantime, I notice a couple style-related things I'd like your thoughts on:
- You removed the indenting of prior `<link>` tags
- You kept the indenting for just one inside an `if` statement
- Your new code uses `echo` statements to output all the newly-added HTML, which is a different style than used previously in the file.

I'm curious about the "why" behind these.

Zen4All:
> You removed the indenting of prior <link> tags
> You kept the indenting for just one inside an if statement

My IDE is to blame for that. will restore indenting
> Your new code uses echo statements to output all the newly-added HTML, which is a different style than used previously in the file.

No good reason. I copied the code from the shop side, and was so focused on this, that I didn't even notice the difference.
Will updated code to match the rest.

torvista:
Is this documented?

drbyte:
> Is this documented?

https://docs.zen-cart.com/dev/admin/templating/

torvista:
Hmm, I used the search on the front page for "javascript" and it did not return that page.

drbyte:
> Hmm, I used the search on the front page for "javascript" and it did not return that page.

Good point. I just created the page today. The search index runs overnight, so I think it should show up in search tomorrow.

PR #3585, opened 2020-06-11
Fix PHP Array Warning

jaywilliams:
PHP Warning: Invalid argument supplied for foreach() in /admin/includes/header.php on line 77.

### DIFF
--- a/admin/define_pages_editor.php
+++ b/admin/define_pages_editor.php
@@ -171,8 +171,7 @@ if (!$lng_exists) {
         ?>
         <?php
         if (($_SESSION['language']) && ($_GET['filename'])) {
-          if (file_exists($file)) {
-            $file_array = (array) @file($file);
+          if (file_exists($file) && $file_array = @file($file)) {
             $file_contents = @implode('', $file_array);
 
             $file_writeable = true;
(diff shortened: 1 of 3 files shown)


drbyte:
Odd.

`file()` always returns an array of the file contents. 
Unless the file doesn't exist ... but we're only running that section of code if we first confirm the file exists.

So ... that begs the question: what circumstance caused it to not return an array? ie: how can we re-trigger the problem?

I don't think casting this to an array is the best solution: I'd rather get to the real root cause.

You could remove the `@` from `@file()` to get more specific details, if you're not sure what was originally actually causing the problem. In fact, I'd like to see the output that generates.

jaywilliams:
On my install, the file _did_ exist, so it passed the `file_exists()` check, but the permissions were set which didn't allow reading by the PHP process (600), this caused `file()` to return *FALSE*, which PHP doesn't want to iterate over.

drbyte:
Okay, fair. 

Why are permissions set to `600`? That doesn't seem like a normal workflow. Is this something you've been doing regularly? If so, why? It would seem that doing that to most files would cripple a typical PHP application.
Seeking to understand how this situation came to be, to understand how "common" this is (never saw this reported by other ZC users before, but older PHP versions weren't always so verbose either), in order to ensure the best fix is implemented.

jaywilliams:
I don't think it's very common, I simply extracted the latest release zip file from the ZenCart website, so it would have whatever default permissions were set, but I figured it may be worth correcting. Either way I corrected it on my end, as I don't like seeing regular warning/errors if I can fix them.

PR #3593, opened 2020-06-19
Add comment to announce change of notifier name scheduled for v158

drbyte:
`NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE` existed from v1.5.0 to v1.5.7

In 1.5.8 it will be deleted and replaced with the correctly-spelled `NOTIFY_ORDER_CART_SUBTOTAL_CALCULATE`

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -610,6 +610,8 @@ class order extends base {
       $shown_price = (zen_add_tax($this->products[$index]['final_price'] * $this->products[$index]['qty'], $product_tax_rate))
           + zen_add_tax($this->products[$index]['onetime_charges'], $product_tax_rate);
       $this->info['subtotal'] += $shown_price;
+
+      // NOTE: This misspelled notifier (with an extra "i") will change to NOTIFY_ORDER_CART_SUBTOTAL_CALCULATE in v1.5.8
       $this->notify('NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE', array('shown_price' => $shown_price));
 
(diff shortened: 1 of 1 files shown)


drbyte:
**Or should we just make the change immediately in 1.5.7?** 

Feedback welcome.

What about having "both"? I opted not to, because adding a new one with proper spelling (having both at same time) could result in plugins doing things twice with the data, so I think it should be a hard switch whenever we change it.

Maybe there's not many plugins affected by it?

I'm sure it's been inquired about before, and I probably said to leave it alone. I was wrong. While it's a minor adjustment, it creates problems not only cognitively/visually but also in patterns for code control/flow ... such as improvements planned for notifier/observer code.

lat9:
So, we're going to inject this potentially non-backward-compatible change to correct a misspelling in a notifier event?

If it's got to be done, I'll suggest going for option 1, adding the correctly-spelled notification to the zc157 base with the comment that indicates that the misspelled version won't be carried forward to the next numerical-Zen Cart release.

FWIW, I've searched my plugins and don't see anywhere that they're currently watching for this event.  Can't say for various site-specific customizations.

zcwilt:
Hmm. I'm using it in some custom client stuff.  :(

zcwilt:
I just written a very quick change to the base class to allow for event aliasing, so that the new eventID (notifier) will be changed to the old version. 
This will stop BC breaks. Just testing on some old code before doing a PR.

scottcwilson:
Event aliasing is a fantastic idea to non-disruptively fix issues like this.  Thanks Ian!

PR #3594, opened 2020-06-19
New Observer method lookup pattern for Notifiers

drbyte:
This introduces a new naming convention for Observer class method names which respond to Notifier calls.

This allows IDEs (text editors) to run a (case-insensitive) search for a notifier name and find these methods which relate to those notifiers.

The new pattern/flow is that when a Notifier hook is triggered it will fire an Observer method using the first method name matching the following pattern/order:

1. `snake_cased_name_of_notifier_event()` ... as long as the name starts with "notify_" or "notifier_" (as is the standard convention already)
2. `updateCamelCasedNameOfNotifierEvent()`
3. `update()`

EDIT: 
Function signatures remain the same in all 3 implementations.

### DIFF
--- a/includes/classes/class.base.php
+++ b/includes/classes/class.base.php
@@ -6,161 +6,180 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: DrByte 2020 May 23 Modified in v1.5.7 $
  */
+
 /**
  * abstract class base
  *
  * any class that wants to notify or listen for events must extend this base class
  *
  */
-class base {
-
+class base
+{
+    /**
+     * @var array of aliases
+     */
     private $aliases = ['NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE' => 'NOTIFY_ORDER_CART_SUBTOTAL_CALCULATE'];
 
     /**
-   * method used to an attach an observer to the notifier object
-   *
-   * NB. We have to get a little sneaky here to stop session based classes adding events ad infinitum
-   * To do this we first concatenate the class name with the event id, as a class is only ever going to attach to an
-   * event id once, this provides a unique key. To ensure there are no naming problems with the array key, we md5 the
-   * unique name to provide a unique hashed key.
-   *
-   * @param object Reference to the observer class
-   * @param array An array of eventId's to observe
-   */
-  function attach(&$observer, $eventIDArray) {
-    foreach($eventIDArray as $eventID) {
-      $nameHash = md5(get_class($observer).$eventID);
-      base::setStaticObserver($nameHash, array('obs'=>&$observer, 'eventID'=>$eventID));
-    }
-  }
-  /**
-   * method used to detach an observer from the notifier object
-   * @param object
-   * @param array
-   */
-  function detach($observer, $eventIDArray) {
-    foreach($eventIDArray as $eventID) {
-      $nameHash = md5(get_class($observer).$eventID);
-      base::unsetStaticObserver($nameHash);
-    }
-  }
-  /**
-   * method to notify observers that an event has occurred in the notifier object
-   * Can optionally pass parameters and variables to the observer, useful for passing stuff which is outside of the 'scope' of the observed class.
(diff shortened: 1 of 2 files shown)


mc12345678:
> Unlike the update() and updateCamelCased() methods, the snake_cased_method() will NOT pass the $eventID as the 2nd parameter (it will skip it, because it is in the name)

It may be in the name; however, it is also possible that other/previous notifier events could be used to call a notifier and the omission of that additional information loses the "source" of the call.

1. NOTIFY_EVENT_1 called
2. NOTIFY_EVENT_1 calls NOTIFY_EVENT_2.
3. NOTIFY_EVENT_2 performs action a if called because NOTIFY_EVENT_2 is triggered, but sets an internal variable to something else if called by NOTIFY_EVENT_1. Hmm.. Perhaps there is either another example or have to modify the internal data structure to instead use a class variable within the given class, though crossing classes may need something different, but I've found that it can be beneficial to know the original origin of the notify event... (`$eventID`)

drbyte:
> 3\. I've found that it can be beneficial to know the original origin of the notify event... (`$eventID`)

Fair. Thanks for the example of "why". 
I've usually found it to be redundant data hence leaning towards dropping it; however, this is also why I've posted it as a PR, to fish out whether this is ideal or not.

As you implied, perhaps we could code it to set a class property to track the current eventID, and then your observer wouldn't have to do that.

includes/classes/class.base.php
```diff
@@ -85,21 +85,35 @@ function notify($eventID, $param1 = array(), & $param2 = NULL, & $param3 = NULL,
...
+              if (preg_match('/^notif(y|ier)_/', $snake_case_method) && method_exists($obs['obs'], $snake_case_method)) {
+                  $obs['obs']->{$snake_case_method}($this, $param1, $param2, $param3, $param4, $param5, $param6, $param7, $param8, $param9);
+                  return;
+              }
+              // If the first check failed, we check for a camelCased version starting with "update": updateNotifierNameCamelCased()
+              // If found, we call it. For legacy reasons we include $eventID as the 2nd parameter.
+              // If the camelCased version is not found, we call "update()". If it doesn't exist then a PHP fatal error will occur.
+              $method = 'update';
+              if (method_exists($obs['obs'], $camelCaseMethod = $method . self::camelize(strtolower($eventID), true))) {
+                  $method = $camelCaseMethod;
```
zcwilt, line 113: haven't tested this, but I thought that you could only assign a variable to an argument if it is passed by reference.

scottcwilson: This is just really hard to read.  Could we not just do 

```
$camelCaseMethod = $method . self::camelize(strtolower($eventID), true); 
if (method_exists($obs['obs'], $camelCaseMethod)) {
   $method = $camelCaseMethod;
}


```

drbyte: This part is pre-existing code. The pass-by-reference works when *received* by reference in the observer class.

drbyte:
UPDATE: Function signatures remain the same in all 3 implementations.

includes/classes/traits/EventManager.php
```diff
@@ -13,17 +13,41 @@ trait EventManager
...
-            if ($obs['eventID'] == $eventID || $obs['eventID'] === '*') {
-                $method = 'update';
-                $testMethod = $method . self::camelize(strtolower($eventID), TRUE);
-                if (method_exists($obs['obs'], $testMethod))
-                    $method = $testMethod;
-                $obs['obs']->{$method}($this, $eventID, $param1,$param2,$param3,$param4,$param5,$param6,$param7,$param8,$param9);
+        foreach ($observers as $key => $obs) {
+            // identify the event
+            $actualEventId = $eventID;
+            $matchMap = [$eventID, '*'];
```
mc12345678, line 25: Shouldn't these two be in advance/outside of the `foreach` loop?

drbyte: If it were permanently standing alone, then yes.

But this is a near-clone of code in the `base` class where alias-matching happens right here as well, and in the execution of this outer loop those variables may (I think) get updated by the inner alias-detection loop, so this reset has benefit.

PR #3596, opened 2020-06-19
simple aliasing for notifier eventID

zcwilt:


drbyte:
Should the alias being set up here be put the other way around? Where the misspelling is the thing being "accommodated"?

ie: by default I think core code should have what we "want it to be", and then the aliases can be patches that bridge things to accommodate for past errors

eg: reverse the array, and also fix the misspelling in the order class

zcwilt:
In legacy code the notifier would be NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE
In that case we don't want anything to change
In new code the notifier would be NOTIFY_ORDER_CART_SUBTOTAL_CALCULATE
In that case we want to replace it with NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE so legacy observers will still fire.

zcwilt:
Hmm, but I see what you mean. In this case new code would have to use the legacy notifier name in its `update` method.

zcwilt:
k, going to close this. 
It's probably better to do the aliasing at the observer point.
e.g. if $obs[NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE]  exists, alias it too NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE

and in that case you are right in that would have to reverse the array

scottcwilson:
> e.g. if $obs[NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE] exists, alias it ... 
Wait - you mean if $obs[NOTIFY_ORDER_CART_SUBTOTAL_CALCULATE] (correctly spelled) exists, then alias - right?

zcwilt:
So what is the intent of aliasing.
If legacy code is using an observer based on an old notifier name it will fail.
So the test we currently do is 

`if ($obs['eventID'] == $eventID || $obs['eventID'] === '*')`

In the legacy code the `$obs['eventID'] = NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE`
and `$eventID = NOTIFY_ORDER_CART_SUBTOTAL_CALCULATE`

So we need to check for an alias using $obs['eventID'] and then  `'temporarily` alias $eventID to NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE
and use that to build the update method call.



if

zcwilt:
Testing some new code now.

scottcwilson:
Your update "k, going to close this." has the alias and the observer both misspelled.  Which one should be the correctly spelled one?

zcwilt:
should read 
e.g. if $obs[NOTIFY_ORDER_CART_SUBTOTAL_CALCULATE] exists, alias it too NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE

Although even that doesn't quite capture what needs to happen

Should have something shortly

PR #3597, opened 2020-06-19
Add developer-mode array of configs to override on new installs

drbyte:
When doing local dev sometimes it's helpful to override some default configuration table settings, without having to use custom .sql files.
This offers the ability to set those when developer mode is on. For new installs only.


scottcwilson:
The change to `zc_install/includes/localConfig.php`  requires PHP 7.   See https://www.zen-cart.com/showthread.php?226902-V157-stuck-in-System-Setup&p=1370176#post1370176

PR #3598, opened 2020-06-19
simple aliasing for notifier eventID

zcwilt:
include unittest as well

### DIFF
--- a/includes/classes/class.base.php
+++ b/includes/classes/class.base.php
@@ -13,7 +13,10 @@
  *
  */
 class base {
-  /**
+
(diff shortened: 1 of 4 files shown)


drbyte:
Looks good.

Tiny concern about naming of the `/unittests/support/zcObserverAliasTest.php` file, only because phpunit by standard convention looks for tests in files ending in `Test.php`, but this file is not actually "tests", but is instead an object used while under test.
While it's harmless, and currently the test suite is configured to look in other directories, I did find myself looking to see what tests this file contained, just because of convention. I felt myself wishing the filename ended in `TestObject` instead of just `Test`.

PR #3603, opened 2020-06-23
Fix Non-static method base::getStaticProperty() should not be called statically

drbyte:
```
1) ObserverAliasingTest::testObserverAliasing
Non-static method base::getStaticProperty() should not be called statically
```


zcwilt:
haven't tested your changes yet, but a couple of observations generally
None of the methods in the base class have any visibility set. As such they would all be considered as public/non-static which would have some effect on error messrs ages.

Secondly all of this black magic timey wimey static property stuff is now probably unnecessary. 
Notifiers were introduced in 1.3.0 and we were still supporting php 4. 1.5.8 would be a good time to test some deeper changes here to simplify the code.

Issue #3342, opened 2020-04-24
Product Qty Units and Product Qty Minimum


papanco:
it would be convenient to set the automatic adjustment of
Product Qty Units and Product Qty Minimum in the cart, similar to the Product Qty Maximum: is it possible to insert it?

drbyte:
Hmmm ... I thought it already did that.

papanco:
> 
> 
> Hmmm ... I thought it already did that.

I tried it today on the latest version but only the warning to manually adjust the quantities appears

mc12345678:
Having added the alert message about the discrepancy for units and minimum values, I'm trying to understand the original statement:
> it would be convenient to set the automatic adjustment of
> Product Qty Units and Product Qty Minimum in the cart, similar to the Product Qty Maximum: is it possible to insert it?

I don't see in processing to the preview where an option to auto set/change the maximum is offered.  I do see in the initial page load that there are defaults set:
````php
  'products_quantity_order_min' => '1',
  'products_quantity_order_units' => '1',
 
  'products_quantity_order_max' => '0',
````
and there is a test when displaying at the minimum and units values:
````php
($pInfo->products_quantity_order_min == 0 ? 1 : $pInfo->products_quantity_order_min)
($pInfo->products_quantity_order_units == 0 ? 1 : $pInfo->products_quantity_order_units)
````
What this does is allow a default of some other value to be applied, provided the default is not 0 then the value provided as the default will be populated on applying a new product or loading a screen of an existing product.

Are you asking that in the alert message that if there is a discrepancy between the units and minimum values that there be something like a clickable link to adjust the units to match the minimum and/or the minimum to match the units?  If so, I don't see a similar capability for the maximum entry. 

So again trying to understand the flow process requested.

papanco:
I enclose examples in order to better explain me: if the cart updates itself automatically with the minimum and multiples that can be ordered, everything is simpler
![example-1](https://user-images.githubusercontent.com/41792764/80736979-2b944c00-8b13-11ea-90b2-1321eaf6b625.jpg)
![example-2](https://user-images.githubusercontent.com/41792764/80736996-3222c380-8b13-11ea-817a-e90363ba8f6b.jpg)
![example-3](https://user-images.githubusercontent.com/41792764/80737016-3a7afe80-8b13-11ea-97ef-7bd4bf38c152.jpg)

drbyte:
@papanco You're right: I'd misunderstood your initial question. 

[...]

PR #3605, opened 2020-06-23
Adjust added product to max of max and units

mc12345678:
When adding a product to the cart, where the addition of the item results in more product in the cart than allowed, the resulting quantity was set to the maximum quantity in stock followed by the maximum quantity allowed for the product and/or mix of attributes. The originally calculated maximum may not align with units where this commit corrects the maximum to account for the product's units.  This takes out one factor of a customers concern about what it takes to not exceed the maximum and still meet the units.  It will though likely lead to confusion about why they can not add more especially if there is some sort of indication of what the maximum quantity of product is that is in stock when units are a factor.  

This modification applies only to adding product to the cart, a separate commit is required to address other aspects such as updating the cart or other such `action` sections. Ultimately this prevents adding a quantity that violates the units when attempting to add the maximum allowed/available quantity.

Intended to address one of many aspects of #3342, specifically when initially adding a quantity to the cart that leads to exceeding the maximum quantity supported by the product configuration.

Looking for review/testing of this before merge and also looking for suggestions of "messages" to be presented.

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -5,7 +5,7 @@
  * @copyright Copyright 2003-2020 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: DrByte 2020 Jun 16 Modified in v1.5.7 $
+ * @version $Id: DrByte 2020 May 19 Modified in v1.5.7 $
  */
 
 if (!defined('IS_ADMIN_FLAG')) {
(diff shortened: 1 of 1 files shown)


mc12345678:
Having submitted this PR "moments" before 3607, should I rebase this?

drbyte:
> Having submitted this PR "moments" before 3607, should I rebase this?

Yes. Specifically, this PR is for a new feature (or improvement of sorts) and not a fix for a showstopper bug in v157, so it should be in the new v158 branch.

drbyte:
Since these 3 code blocks are effectively identical, probably best to extract them to a separate method (function in this class), and just pass it the necessary parameters it requires, so it can return the updated $new_qty.

Then again, the request for suggested message responses reminds me that for this situation there's no way to know for certain if any of this is even relevant for a given store. 
This makes me inclined to suggest that these code blocks be replaced with just a single Notifier hook and let the storeowner write their own custom code to handle it (or you could move this logic to an observer and package it as a plugin), instead of having this logic in core code.

scottcwilson:
@mc12345678 Would you please get this updated so we can merge?

PR #3606, opened 2020-06-23
Refactor zen_get_products_quantity_mixed

mc12345678:
Support an early escape if the product doesn't have attributes.
Minimize potential issues associated with providing a non-existent `products_id`.
Use full capitalized MySQL commands.
Remove quote/text conversion for `products_id` value in query.
Limited the database search to a single result.

### DIFF
--- a/includes/functions/functions_prices.php
+++ b/includes/functions/functions_prices.php
@@ -480,18 +480,15 @@
 
 // don't check for mixed if not attributes
     $chk_attrib = zen_has_product_attributes((int)$product_id);
-    if ($chk_attrib == true) {
-      $the_products_quantity_mixed = $db->Execute("select products_id, products_quantity_mixed from " . TABLE_PRODUCTS . " where products_id = '" . (int)$product_id . "'");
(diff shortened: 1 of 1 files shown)


drbyte:
Question: Does this fix a 1.5.7 bug?

drbyte:
Actually, more specifically, I don't see where a string `'none'` response is ever used.

Strongly leaning toward deleting that response.

scottcwilson:
Would this be better for 1.5.8?

mc12345678:
Do not recall the specific test/condition that first raised an issue, but the second entry of the change summary was the primary reason for initial proposal as it caused a myDebug log that led to this code section.

I don't recall if by non-existent it meant a products_id that was outside the scope of product in the database or if it was a 0 value (which also is not expected to be a database record).

Assuming 1.5.7 is still supposed to be compat only up to php 7.3 as highest, then would say yes associated to fixing an issue with 1.5.7 in addition to some other "normal" touchups.

drbyte:
Done in v1.5.8

PR #3613, opened 2020-06-24
Add information on SQL Patches to the admin activity log

scottcwilson:


drbyte:
How do we strip logging of sensitive data?

scottcwilson:
Can you give me an example of when this would occur?

drbyte:
- updating a customer's personal information, whether in customer or order tables
- inserting configuration keys that contain module credentials for external services

scottcwilson:
OK but this data is already in clear text in the database.  Why would it be worse in the log?

drbyte:
> Why would it be worse in the log?

Because log files can be downloaded and stored separately. The rule is to never log sensitive stuff.

drbyte:
You didn't explain your "why" for doing this PR. 

I assume you wanted some sort of audit trail of which patches you'd run?

scottcwilson:
> Because log files can be downloaded and stored separately. The rule is to never log sensitive stuff.

If you *type it into the box* in Install MySQL Patches, it gets logged.  This just gives uploading a file the same behavior.

PR #3614, opened 2020-06-24
Admin template_select: Prevent notice when template dir does not exist

scottcwilson:
This can happen during an upgrade when the custom template has not yet been created.

### DIFF
--- a/admin/template_select.php
+++ b/admin/template_select.php
@@ -107,7 +107,7 @@ if (zen_not_null($action)) {
                 foreach ($templates as $template) {
                   if (!isset($template_info[$template['template_dir']])) {
                      $template_info[$template['template_dir']] = [
-                       'name' => $template['template_dir'],
+                       'name' => '<strong class="errorText"> MISSING DIRECTORY: ' . $template['template_dir'] . '</strong>',
                        'version' => '', 
                        'author' => '', 
                        'description' => '',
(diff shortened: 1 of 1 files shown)


admin/template_select.php
```diff
@@ -104,6 +105,16 @@
...
                 $templates = $db->Execute($template_query_raw);
                 foreach ($templates as $template) {
+                  if (!isset($template_info[$template['template_dir']])) {
+                     $template_info[$template['template_dir']] = [
+                       'name' => $template['template_dir'],
+                       'version' => '', 
+                       'author' => '', 
+                       'description' => '',
+                       'screenshot' => '', 
+                       'missing' => true,
```
drbyte, line 115: Not a fan of adding this `missing` element. Why not just call `continue` here, instead of later? Or unset the `$template` entry from the `$templates` array?

scottcwilson: If you do that then the line on the page showing the current template will be blank.

drbyte:
So what's the goal here? 
- someone does a fresh new install
- imports an old database that points to a template directory that's not in the current filesystem

Instead of triggering a Notice, now we want to do what? 

- say "Sorry, missing template folder"?
- leave the row blank?
- skip the row?
- say "The template you'd previously selected is not present on this server. Please choose another."
- other?

scottcwilson:
Import old db is the use case.

You could give an error message but I like the idea of showing the old name to remind them what they need to copy into their new installation.

PR #3615, opened 2020-06-24

drbyte:
This is telling git to standardize all line-endings in all files.

IIUC, technically this puts all files into git with LF line-endings, and then it's up to your own local git client to adapt locally to CRLF on your own PC if needed, and then translate back to LF again when making commits.
That's the theory anyway.

You can configure your local git to `autocrlf` appropriately using instructions at https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings#global-settings-for-line-endings

Additionally, in the v158 repo an `.editorconfig` file has been added. Most modern code editors honor the values set in such a file if it exists, in order to help enforce standards within the repo.


scottcwilson:
You'd need to have a Windows person test this - I do all my dev on MacOS.  Good concept though - thanks for doing this.

scottcwilson:
Hmmm.... spoke too soon.  includes/classes/site_map.php is still full of ^M characters. :(

admin/includes/modules/products_previous_next.php
```diff
@@ -1,151 +1,151 @@
...
+}
+
+if (!isset($prev_next_list) || $prev_next_list == '') {
+// calculate the previous and next
+
+  $result = $db->Execute("SELECT products_type
+                          FROM " . TABLE_PRODUCTS . "
+                          WHERE products_id = " . (int)$products_filter);
+  $check_type = ($result->EOF) ? 0 : $result->fields['products_type'];
+  if (!defined('PRODUCT_INFO_PREVIOUS_NEXT_SORT')) define('PRODUCT_INFO_PREVIOUS_NEXT_SORT', zen_get_configuration_key_value_layout('PRODUCT_INFO_PREVIOUS_NEXT_SORT', $check_type));
```
mc12345678, line 27: For anyone seeking to further incorporate available Zen Cart created functions into this code, per #5547, see below about doing using the function `zen_define_default` (available as of Zen Cart 1.5.8) for replacement of this line such as:
````php
zen_define_default('PRODUCT_INFO_PREVIOUS_NEXT_SORT', zen_get_configuration_key_value_layout('PRODUCT_INFO_PREVIOUS_NEXT_SORT', $check_type));
```` 
> The creation of a debug log in a case like this seems proper to me. Your database has been damaged (or your code is bad).

There is not yet a function available to test if the `configuration_key` exists in the table `product_type_layout`. So per the above, you have to either test for the existence of that `configuration_key` initially, blindly generate the key, or accept the debug log that will be created on each page load. Why one would do this testing up front and not just carry on collecting the data with that additional code, is beyond me.

PR #3616, opened 2020-06-24
Coupon Popup link on checkout success page does not work

scottcwilson:
Missing javascript.  Checkout with a coupon and you'll see the link on the checkout success page gives a JS error.

### DIFF
--- a/includes/modules/pages/checkout_success/jscript_main.php
+++ b/includes/modules/pages/checkout_success/jscript_main.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * jscript_main
+ *
+ * @copyright Copyright 2003-2020 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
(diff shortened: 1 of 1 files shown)


drbyte:
Should we just be stripping out the link instead?

scottcwilson:
I'd say leave it - could be useful, plus older templates will have this.

drbyte:
Did I miss seeing reports of this being a problem in 1.5.6 ?

scottcwilson:
I don't know if it was an issue in 1.5.6.  I just found this in today's 1.5.7 testing when I updated my store.

drbyte:
Should this invoke the entire popup? Or just provide an empty JS function to avoid the error?
Also, why the commented-out-code?

scottcwilson:
The commented out code should be fixed - I think I may have just pulled this file from a prior release.  
I prefer to do the full popup so people can use this function, which is likely in a number of templates.

scottcwilson:
The commented out code should be fixed - I think I may have just pulled this file from a prior release.  
I prefer to do the full popup so people can use this function, which is likely in a number of templates.

PR #3617, opened 2020-06-24
ot_coupon - Add coupon_code to SELECT statement

scottcwilson:


drbyte:
I can go study the code to guess what problem you're trying to solve, or you could just state it in the PR description. 
Please explain the use-case.

scottcwilson:
I compare the coupon name to a specific string like the email address (for example) and execute some logic on that basis.

drbyte:
> How about using `SELECT *` instead of specific columns as a potential future-proofing step?

NOTE: while this may suit in this case, using `SELECT *` can frequently slow down queries if the number of columns is large and the number of rows retrieved is large ... because it forces all that data to be loaded from disk in the db and then transmitted across the network to be loaded into PHP which then loads it into memory for parsing.
One must balance convenience vs performance.

PR #3618, opened 2020-06-24
Prevent notice on customers_fax field

scottcwilson:


drbyte:
Refactored to avoid the `else` clause.

PR #3619, opened 2020-06-24
Prevent notice during send email during currency char substitution

scottcwilson:


drbyte:
In the scenario where this came up, what was the string that ended up having no substitution available? 
Was this an intentional leaving-out of the replacement? ie: is there a common case for this?

scottcwilson:
Sorry, don't have the specifics, I was just fixing logs and noticed it.

PR #3621, opened 2020-06-24
Fix define pages editor

scottcwilson:
Currently, opening Admin > Tools > Define Pages Editor does not correctly open the define page (instead it shows includes/local/skip_version_check.ini).  First reported here: 

https://www.zen-cart.com/showthread.php?226877-Define-Pages-Editor-only-opens-includes-local-skip_version_check-ini

### DIFF
--- a/admin/includes/header.php
+++ b/admin/includes/header.php
@@ -76,8 +76,8 @@ $version_ini_index_sysinfo = '';
 if (!isset($version_check_sysinfo)) $version_check_sysinfo = false;
 if (!isset($version_check_index)) $version_check_index = false;
 
-$file = DIR_FS_ADMIN . 'includes/local/skip_version_check.ini';
-if (file_exists($file) && $lines = @file($file)) {
+$skip_file = DIR_FS_ADMIN . 'includes/local/skip_version_check.ini';
+if (file_exists($skip_file) && $lines = @file($skip_file)) {
     foreach ($lines as $line) {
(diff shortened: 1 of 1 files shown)


scottcwilson:
Recommended for backport to 1.5.7.

drbyte:
Any idea what caused this break in 157 vs 156 where it wasn't broken?

scottcwilson:
Bear in mind that the `admin/includes/application_bootstrap.php` file is all new.

scottcwilson:
Upon reconsideration, a more future proof fix that would be harder to accidentally break would be:

- Change lines 79-80 of admin/includes/header.php to use a different variable name (say, $skip_file) instead of $file. 
That's where the clobber occurs, and that was where the bug was introduced (the filename was repeated in 1.5.6, but in 1.5.7, a variable to store the name was used). 

As a further enhancement (1.5.8 only)
- Remove lines 96-126 of admin/includes/application_bootstrap.php which are not used.

drbyte:
This is probably better, yes, as far as 157.

I find throwaway variables like $var $file $sql $result etc that are preferable when they are localized in a limited context. I have no issue with re-using them later for another purpose. I prefer not to create new global-scope variables that are used one time and never referenced again (eg $skip_file in this case), so in 158 to prevent clobbering it would be even better to put the skip-ini lookup code into a function so that the variables used aren't in global scope.

scottcwilson:
You're welcome!  Please refactor to taste.

PR #3624, opened 2020-06-29
Include zen_cfg_read_only configuration-helper function

lat9:
... as distributed in various plugins currently, e.g. Sitemap XML.  Credit given to Andrew Berezin (rip).

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -3101,15 +3101,13 @@ function zen_get_master_categories_pulldown($product_id, $fullpath = false)
     $string .= zen_draw_hidden_field($name, '--none--');
     return $string;
   }
-  
+
 /**
- * Function for configuration values that are read-only, e.g. a plugin's version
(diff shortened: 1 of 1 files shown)


zcwilt:
We will need to document this somewhere for plugin authors. 
If they are using it, then it will throw a fatal error.
Us wrapping this in a `!function_exists` probably won't help as it's likely to be loaded before a plugin functions.
Have raise issue https://github.com/zencart/documentation/issues/270 as a reminder.

lat9:
All usage that I've seen (primarily from AndrewBerezin's plugins) wrap the function declaration with a `!function_exists` clause, so I don't think that that's going to be an issue.

Those functions are delivered in the `extra_functions` sub-directory, so they're loaded after the base `general.php` function file.

scottcwilson:
A note will be added to `/user/upgrading/release_specific_upgrade_considerations.md` on the next checkin.

PR #3629, opened 2020-06-30
Fixes default attribute issues

### LINKED ISSUE #3626: Default attribute doesn't work
This is the feature described as follows:
Storefront - Products with single-valued attributes (such as downloads) can now be added to the cart from the listing page.

There is a bug in the test data that led me to believe this was working but it won't work as expected on downloadable products.  Still thinking about the best way to fix this.

scottcwilson:
- Correct test data set to ensure single product single attribute product has an associated file download
- Update documentation on data set
- Update shopping cart: 
  - build attributes array for defaultable attributes products added from listing page.
  - correct in cart quantity in that situation.
  - permit add from listing page when Buy Now button is used (vs Multiple Products Qty Box). 

Fixes #3626

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -236,7 +236,17 @@ class shoppingCart extends base {
   function add_cart($products_id, $qty = '1', $attributes = '', $notify = true) {
     global $db, $messageStack;
     if ($this->display_debug_messages) $messageStack->add_session('header', 'FUNCTION ' . __FUNCTION__, 'caution');
-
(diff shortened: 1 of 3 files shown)


drbyte:
Is the "bug" specific to downloadable products?

scottcwilson:
It's a bug for products where the attribute values *truly* need to be known for the order to be successfully fulfilled - and the most common case of this would be downloadable products.  The orders_products_download table won't be populated and the download button won't be presented if this isn't done.  There might be other cases - not sure.

scottcwilson:
Note: The user friendly name for this fix is "Adding downloadable product from listing page creates an order with a missing download button."

PR #3630, opened 2020-06-30
Add missing message

scottcwilson:
This message is used in the following workflow: 
-  from the product listing screen, click the A icon on a product with attributes
- click the Delete button next to Delete ALL Product Attributes for.


drbyte:
Cherrypicked into v1.5.7

PR #3631, opened 2020-06-30
V157 admin product avail datepicker vs empty or malformed dates

mc12345678:
Support product update if incorrectly formatted date provided 
Addresses: https://www.zen-cart.com/showthread.php?226900-v157-DATE_FORMAT_DATE_PICKER-error-if-date-format-change-and-update-product&p=1370184#post1370184
When storing a product without an available date and using a format other than `yy-mm-dd`, the update action failed because `$dt===false` and `false->format` doesn't exist.

Alternative solutions considered: testing against the existence of the method `format` of the variable `$dt` as part of this resolution, providing admin feedback (likely in the preview area or as a redirect here when such a failure occurs.

The same problem appears to occur if an invalid date format is provided: E.g. if using `dd-mm-yy` for `DATE_FORMAT_DATE_PICKER` and a date is entered as `dd/mm/yy` then `$dt` will be set to false and the same problem would occur as otherwise found for an empty date and a modified format.


proseLA:
if one is serious about localization with regards to dates, should also be on the v157 bugs and fixes list.

PR #3632, opened 2020-06-30
Incorporate $antiSpamFieldName

mc12345678:


mc12345678:
Recommend applying against 1.5.7.

PR #3633, opened 2020-06-30
V158 language loading infrastructure

zcwilt:
This commit adds further infrastructure to the Language loading classes in preparation for converting catalog side code.
Removes legacy admin language files and replaces them with array based defines.
Note. The changes to the gv_faq and index page code on catalog side. This removes the php selection choice of the language defines from the actual language files. 
Note also. Although legacy define files have been removed from admin, code will still allow for legacy type language defines in contributions.


torvista:
I would like to convert language files. 
Mention is made else of zcgo, 
https://github.com/zencart/zencart/pull/3432
but this has been moved to artisan.

The previous request I made concerning my inability to use laravel/artisan seems to have vanished....or I just can't find it.

So, how can it be done?

PR #3637, opened 2020-06-30
Ensure dir is not null before reading

scottcwilson:


drbyte:
More information would be helpful about how this new situation has presented itself.

Also:
- What's the use-case where this was missing but otherwise expected? eg: is this a situation where the images directory has been deleted and therefore all kinds of errors will appear all over?
- Does all code that's calling this function properly handle this new empty array coming back? Or is it expecting at least the default entry to be present?

scottcwilson:
Correct - images folder was missing.  Yes, new code is well handled with no notices.

PR #3651, opened 2020-07-02
Update css for login page

Zen4All:
Because the file stylesheet.css is now loaded on the page, some tweaks needed to be done to make the look the same as in ZC 1.5.7

### DIFF
--- a/admin/includes/css/login.css
+++ b/admin/includes/css/login.css
@@ -7,10 +7,18 @@
 
 body {
   background-color: #feffef;
+  padding : 0;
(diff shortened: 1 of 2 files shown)


scottcwilson:
Thanks for all your hard work in making the UI look better.  When I have to go back to an old version, I'm shocked at how much has been improved!

Zen4All:
> Thanks for all your hard work in making the UI look better. When I have to go back to an old version, I'm shocked at how much has been improved!
You're welcome, and am glad you like it. I had to install a 1.5.0 shop a few days ago, to test an old plugin, and was also amazed of what we have accomplished as a community in the last few years.

PR #3652, opened 2020-07-02
Illuminate support request and routing

zcwilt:


zcwilt:
Support for Arr/String/Collections
Support for Illuminate\Request Illuminate\Routing Illuminate\View

There is no intention to switch to routing for Zen Cart web routes just yet.
The support here will allow some testing for the future, and the possiblity of building
out an API based on routing.

Also updated the config define loader to cast some groups to int
as was done before.

scottcwilson:
Exciting stuff!  Thanks for doing this.

zcwilt:
Hoping to have a very simple example of an api route soon.

scottcwilson:
Use case?

PR #3653, opened 2020-07-03
check page strip of .php suffix done on wrong side

proseLA:
looks like comparison is done on the wrong side.

8 years ago, ```$page``` was defined as without extension:

https://github.com/zencart/zencart/blob/8e14368e080b910a120f6e2ec3d8a8c7dc1b9f11/admin/includes/init_includes/init_admin_auth.php#L55

7 years ago, a fix was implemented to address those `modders` who included the '.php' in their filename:

https://www.zen-cart.com/showthread.php?207650-Antibugging-in-admin-function-check_page%28%29

only problem is that the constant is the one that has the .php at the end.  so now the comparison is for ```filename.php.php == filename```

this PR fixes that.  and allows ```modders``` to continue to mess up with '.php' in their filenames without unintended negative consequences.

### DIFF
--- a/admin/includes/functions/admin_access.php
+++ b/admin/includes/functions/admin_access.php
@@ -36,7 +36,7 @@ function check_page($page, $params) {
   $result = $db->Execute($sql);
   $retVal = FALSE;
   while (!$result->EOF) {
-    if ($result->fields['main_page'] != '' && defined($result->fields['main_page']) && (constant($result->fields['main_page']) == $page || basename(constant($result->fields['main_page']), '.php') == $page) && $result->fields['page_params'] == $page_params) {
+    if ($result->fields['main_page'] != '' && defined($result->fields['main_page']) && basename(constant($result->fields['main_page']), '.php') == $page && $result->fields['page_params'] == $page_params) {
       $retVal = TRUE;
     }
     $result->MoveNext();
--- a/admin/includes/functions/admin_access.php
+++ b/admin/includes/functions/admin_access.php
@@ -36,7 +36,7 @@ function check_page($page, $params) {
   $result = $db->Execute($sql);
   $retVal = FALSE;
   while (!$result->EOF) {
-    if ($result->fields['main_page'] != '' && defined($result->fields['main_page']) && (constant($result->fields['main_page']) == $page || constant($result->fields['main_page']) . '.php' == $page) && $result->fields['page_params'] == $page_params) {
+    if ($result->fields['main_page'] != '' && defined($result->fields['main_page']) && (constant($result->fields['main_page']) == $page || basename(constant($result->fields['main_page']), '.php') == $page) && $result->fields['page_params'] == $page_params) {
       $retVal = TRUE;
     }
     $result->MoveNext();
@@ -966,8 +966,8 @@ function zen_admin_authorized_to_place_order()
         if (count($profile_list) != 0) {
             $profile_clause = ' AND admin_profile IN (' . implode(',', $profile_list) . ')';
             $emp_sql =
-                "SELECT admin_profile, admin_pass 
-                   FROM " . TABLE_ADMIN . " 
+                "SELECT admin_profile, admin_pass
+                   FROM " . TABLE_ADMIN . "
                   WHERE admin_id = :adminId:$profile_clause
                   LIMIT 1";
             $emp_sql = $db->bindVars($emp_sql, ':adminId:', $_SESSION['admin_id'], 'integer');


scottcwilson:
Please be sure this really works.  You said here

https://www.zen-cart.com/showthread.php?218710-Mod-List-Support-Thread/page3

that it didn't.  If that's true, you need to fix your patch so it preserves the behavior.

proseLA:
> Please be sure this really works. You said here
> 
> https://www.zen-cart.com/showthread.php?218710-Mod-List-Support-Thread/page3
> 
> that it didn't. If that's true, you need to fix your patch so it preserves the behavior.

scott,
i apologize.  this PR works **100%.  100%.**  look at the code.

my comment on your mod list forum was that the existing code fix to address this issue does not work.  that is why i did a PR.  i can see how the comment there was very easily mis-construed.

this situation was only brought to my attention via the post here:
https://www.zen-cart.com/showthread.php?226778-Inventory-Updater&p=1370295#post1370295

i looked at this code quite thoroughly.  and if you read above what i wrote with regards to this PR, i think it is pretty clear. 

again, i can see how easily my comments on the mod list forum are easily misconstrued.  and for that i apologize.

i am certainly willing to listen to anything about the code on this PR; but the logic here is more than sound and well tested.

finally, if i actually thought that this PR did not work; i would delete it.  perhaps you can test it for yourself and post back.  its a single line change.  again, if you do not understand my above explanation or the code, i'm more than willing to listen and further explain the bug.

best.

scottcwilson:
If it works, great.  But I thought you said it didn't work. 

Here's the situation: right now, if some goofball (like me) forgets and appends ".php" to a filename definition, the system corrects for this and the code still works as expected for non superuser admins.  I just want this behavior preserved.

mc12345678:
Yeah, instead of removing one of the conditions in place of another, shouldn't an additional `||` be added in to address the case of the database stored `page_name` already having `.php` in it and the value checked against does not...

The current tests appear to address the following:
- 1st condition test (`constant($result->fields['main_page']) == $page`)
  - database doesn't have `.php` and provided `$page` does not have `.php`
  - both database and `$page` have `.php`
- 2nd condition test (`constant($result->fields['main_page']) . '.php' == $page`)
  - database doesn't have `.php`, but provided `$page` has `.php` 

This third condition "described" is that the database has `.php` already in it and `$page` does not...
So shouldn't the "changed" version instead be added as a "third" option?
- 3rd condition (to add presumably addressing the issue of this PR) (`basename(constant($result->fields['main_page']), '.php') == $page`)
  - database does have `.php` and provided `$page` does not have `.php` 

This makes the single line change while maintaining existing operation:
````diff
-    if ($result->fields['main_page'] != '' && defined($result->fields['main_page']) && (constant($result->fields['main_page']) == $page || constant($result->fields['main_page']) . '.php' == $page) && $result->fields['page_params'] == $page_params) {
+    if ($result->fields['main_page'] != '' && defined($result->fields['main_page']) && (constant($result->fields['main_page']) == $page || constant($result->fields['main_page']) . '.php' == $page || basename(constant($result->fields['main_page']), '.php') == $page) && $result->fields['page_params'] == $page_params) {
````

**Comment by scottcwilson** [comment]
> my comment on your mod list forum was that the existing code fix to address this issue does not work.
This is not correct. You can install Mod List (as is, with the bug), create a non super user account, allow them access to Mod List, and voila, they can run Mod List.  A long time ago this was not the case, which was why these and other changes were put in.  But the admin has had a ton of restructuring since then, so if we were doing this now, it might well be different. 

So it works if the language define is in includes/languages/english/extra_definitions/mod_list.php
What if we move it? 
Put it in includes/extra_datafiles/mod_list.php then remove from original location.
It still works. 
Remove mod list from their admin profile, and it does not work - again as expected.

So I'm not convinced your change is needed at all. 

YOUR mod didn't work because you didn't follow the naming conventions.  If you had provided a SQL file or done it with an extra function and used proper names, it would have worked.  

INSERT INTO admin_pages (page_key, language_key, main_page, page_params, menu_key, display_on_menu, sort_order) VALUES ('inventory', 'BOX_REPORTS_INVENTORY', 'FILENAME_INVENTORY', '', 'reports', 'Y', 999);

with corrections to the defines in ./includes/languages/english/extra_definitions/inventory.php and ./includes/extra_datafiles/inventory.php.

**Comment by proseLA** [comment]
> > my comment on your mod list forum was that the existing code fix to address this issue does not work.
> 
> This is not correct. You can install Mod List (as is, with the bug), create a non super user account, allow them access to Mod List, and voila, they can run Mod List. A long time ago this was not the case, which was why these and other changes were put in. But the admin has had a ton of restructuring since then, so if we were doing this now, it might well be different.
> 
> So it works if the filename define is in includes/languages/english/extra_definitions/mod_list.php
> What if we move it?
> Put it in includes/extra_datafiles/mod_list.php then remove from original location.
> It still works.
> Remove mod list from their admin profile, and it does not work - again as expected.
> 
> So I'm not convinced your change is needed at all.
> 

> 
> with corrections to the defines in ./includes/languages/english/extra_definitions/inventory.php and ./includes/extra_datafiles/inventory.php.

scott,
you are wrong.  in addition, you should stay on topic, as opposed to floating all over the place.

the topic on hand is whether someone defines a filename with an additional ```.php``` and whether a non-superuser has access to that page when their security profile allows it.  in that scenario, if it will work or not.  **it does not work.  period.**  that is what this pr addresses.  nothing more, nothing less.

> YOUR mod didn't work because you didn't follow the naming conventions. If you had provided a SQL file or done it with an extra function and used proper names, it would have worked.
> 
> INSERT INTO admin_pages (page_key, language_key, main_page, page_params, menu_key, display_on_menu, sort_order) VALUES ('inventory', 'BOX_REPORTS_INVENTORY', 'FILENAME_INVENTORY', '', 'reports', 'Y', 999);

the above SQL does not follow naming conventions.  i refer you to the documentation:

https://docs.zen-cart.com/dev/code/creating_menu/

the pageKey is supposed to be something like ```toolsInventory```.  you have it listed as inventory.  that's probably how you have done it for all of your mods.  but it is not per the docs, nor is it necessary, nor is it what this PR addresses.  there does seem to be some additional code so that your own way of naming a page key works:

https://github.com/zencart/zencart/blob/f6fe46046a3a465ea13b728beecd6d8bc045a4ae/admin/includes/functions/admin_access.php#L44-L58

**only because you choose a non-standard page key equal to your pagename without the ```.php``` does your mod_list get a return true value and allow access to the page.  nothing to do with this PR at all.** 

i looked at the code.  i've spent a bunch of time studying this.  please stay on topic.  my mod does follow naming conventions and in previous discussions on this board, i expressed the opinion that a ```modder``` should **not** be picking their own page key.  the ZC core should handle it.  but, alas i digress....

@mc12345678 
i agree that your code might be safer.  but i think it unnecessary.    again, i refer you to the definition of ```$page```:

https://github.com/zencart/zencart/blob/8e14368e080b910a120f6e2ec3d8a8c7dc1b9f11/admin/includes/init_includes/init_admin_auth.php#L55

the .php is stripped out.

in looking at the code, prior to v157 the ```check_page``` function is only used in the ```init_admin_auth.php``` script.  with the addition of dashboard widgets, this function is now used for the widgets.  but in the widgets it is comparing a constant to a constant.

in referring to your test 2 above, with the widgets, we are comparing the constant to each other.  and i submit to you, you will never get to that test.  which is why it needs to be removed.

**Comment by scottcwilson** [comment]
OK now. I see your point.  Holy cow this was confusing. 

A super easy way to have explained this, rather than repeatedly saying "your code doesn't work" would have been 

"Your code appears to work but only does so if the page key is precisely chosen to be the same as the filename (without extension) of the file.  This is a more general fix that works for an arbitrary page key."

**Comment by mc12345678** [comment]
> i agree that your code might be safer. but i think it unnecessary. again, i refer you to the definition of $page:
> ````
> zencart/admin/includes/init_includes/init_admin_auth.php
> 
> Line 55 in 8e14368
> 
> $page = basename($PHP_SELF, ".php"); 
> the .php is stripped out.

Ok,. sure that's how the "default" code goes; however, what about the "reusability" of the function where another plugin calls the function `check_page` passing `$page` that has a `.php` extension? Directly modifying the code as suggested in this PR will break that operation. Therefore this modification should support this being a "third" option (whether positioned first, *second*, or third in the list). 

I also can remember that I had a conversation on this topic before, but over the last few days I haven't been able to find it and it related to something like this. Just can't recall where the conversation was held.

**Comment by proseLA** [comment]
> 
> Ok,. sure that's how the "default" code goes; however, what about the "reusability" of the function where another plugin calls the function `check_page` passing `$page` that has a `.php` extension? Directly modifying the code as suggested in this PR will break that operation. 

that situation is captured in your 1st test condition above.  the PR will not break any define that has a .php extension.  

if we were to look at image_handler mod, it makes use of the ```check_page``` function.

```
if (zen_is_superuser() || check_page(FILENAME_IMAGE_HANDLER_VIEW_CONFIG, '')) {
```

the admin_pages table only stores ```FILENAME_*``` in the main_page field.  so if the constant above had a ```.php``` or not, it is irrelevant; it is comparing it to itself, so if the constant is in the table you will get a ```$retVal``` of true.

**Comment by mc12345678** [comment]
> that situation is captured in your 1st test condition above. the PR will not break any define that has a .php extension.
> 
> if we were to look at image_handler mod, it makes use of the check_page function.
> ````
> if (zen_is_superuser() || check_page(FILENAME_IMAGE_HANDLER_VIEW_CONFIG, '')) {
> ````
> the admin_pages table only stores `FILENAME_*` in the main_page field. so if the constant above had a `.php` or not, it is irrelevant; it is comparing it to itself, so if the constant is in the table you will get a `$retVal` of true.

Again, not seeing the opportunity of one *NOT* using a defined type variable to pass/present for review against the database... E.g. `check_page('filename_typed_in.php', '')` Or further still, if the constant placed in the database is different than the constant used for comparison where one has `.php` assigned and the other does not (Database version doesn't have `.php` where `$page` does)...

**Comment by drbyte** [comment]
> > that situation is captured in your 1st test condition above. the PR will not break any define that has a .php extension.
> > if we were to look at image_handler mod, it makes use of the check_page function.
> > ```
> > if (zen_is_superuser() || check_page(FILENAME_IMAGE_HANDLER_VIEW_CONFIG, '')) {
> > ```
> > 
> > 
> > the admin_pages table only stores `FILENAME_*` in the main_page field. so if the constant above had a `.php` or not, it is irrelevant; it is comparing it to itself, so if the constant is in the table you will get a `$retVal` of true.
> 
> Again, not seeing the opportunity of one _NOT_ using a defined type variable to pass/present for review against the database... E.g. `check_page('filename_typed_in.php', '')` Or further still, if the constant placed in the database is different than the constant used for comparison where one has `.php` assigned and the other does not (Database version doesn't have `.php` where `$page` does)...

Valid point. I'm having trouble envisioning why someone would build it like that in the first place, and why we should let them?

File: admin/includes/functions/admin_access.php [hunk 1]
```diff
@@ -36,7 +36,7 @@ function check_page($page, $params) {
   $result = $db->Execute($sql);
   $retVal = FALSE;
   while (!$result->EOF) {
-    if ($result->fields['main_page'] != '' && defined($result->fields['main_page']) && (constant($result->fields['main_page']) == $page || constant($result->fields['main_page']) . '.php' == $page) && $result->fields['page_params'] == $page_params) {
+    if ($result->fields['main_page'] != '' && defined($result->fields['main_page']) && (constant($result->fields['main_page']) == $page || basename(constant($result->fields['main_page']), '.php') == $page) && $result->fields['page_params'] == $page_params) {
```
drbyte, line 39:
```diff
-(constant($result->fields['main_page']) == $page || constant($result->fields['main_page']) . '.php' == $page)
+(constant($result->fields['main_page']) == $page || basename(constant($result->fields['main_page']), '.php') == $page)
```

Doesn't this safely reduce even further by just removing the first half? 
eg:
```diff
-(constant($result->fields['main_page']) == $page || constant($result->fields['main_page']) . '.php' == $page)
+basename(constant($result->fields['main_page']), '.php') == $page
```

proseLA: damn straight. 

absolutely valid point.

PR #3658, opened 2020-07-04
Full laravel integration

zcwilt:
# Laravel Integration

Rather than continuing down the path of just integrating some Laravel
sub-systems, I've decided to go and do a full blown integration of Laravel.

The benefits of using a full integration are :-

 - Easier upgrading of Laravel components
 - Easier integration of 3rd party libraries
 - More Laravel components available
 - Simplified booting of laravel components.

## Dump and Die

Laravel provides some extra debugging tools
`dump()` and `dd()`

`dump($someVar)` is similar to var_dump but is wrapped in some css/js that produces
output where nodes can be expanded/contracted. `dd($someVar)` is just `dump()` followed by `die()`

Examples can be seen in
`includes/illuminate_bootstrap.php` circa line 6 and
`includes/application_bottom.php` circa line 15

## Example Routes

**Note** Routes are disabled by default. To enable them you need to use a couple of defines in
`extra_configures`

define('ALLOW_LARAVEL_API_ROUTES', true);
define('ALLOW_LARAVEL_WEB_ROUTES', true);

I've added a few example routes.

index.php/welcome
index.php/api/adminactivity
index.php/api/adminactivity?where[]=severity:eq:info
index.php/api/adminactivity?where[]=page_accessed:noteq:display_logs.php
index.php/api/adminactivity?where[]=page_accessed:noteq:display_logs.php&sort[]=severity
index.php/api/customer

## Error Reporting

In a standalone Laravel install and given certain settings, Laravel
can display an error page with a parsed stack trace of the error.
Laravel does this by turning PHP errors(warnings/notices etc) into Exceptions.

Generally we don't want this as it would interfere with the normal error display
and logging that Zen Cart does. However as it can be useful in some cases, there are
some new defines to control this.

### `define('STRICT_ERROR_REPORTING', false);`

This works as it has always done in.

### `define('STRICT_ERROR_REPORTING_LEVEL', E_ALL);`

This works in addition to `STRICT_ERROR_REPORTING` as normally Zen Cart would set the error
reporting level to E_ALL. This define just allows a bit more specificity. If not defined
`E_ALL` will be used as the default.

### `define('ILLUMINATE_ERROR_REPORTING_LEVEL', 0);`

If this is set to a non-zero value, the Laravel error/exception system will be used to
display a whoops/error page.

## Laravel Integration Notes.

 - Currently all middleware is disabled.
   See app/Http/Kernel.php

 - artisan is available inside the app/ directory.

 - We are using Laravel 5.8 due to PHP version constraints.

 - Haven't tested webpack etc. Expect they need tweaking.

 - Laravel does some caching of views etc. The directories
   that this is saved to are within the /app folder structure
   and therefore have to be writable. I expect we can set
   these directories to something within the Zen Cart directory
   structure. Still exploring that.

 - There is an .env.example file in the app/ folder. This can
   be copied to .env and is useful for overriding defaults on
   a per server basis. Currently it is not required, but may
   need to be required in the future to store the APP_KEY at a
   minimum.


zcwilt:
screenshot-zc-current test-2020 07 04-21_28_19

zcwilt:
screenshot-zc-current test-2020 07 04-21_23_04

zcwilt:
screenshot-zc-current test-2020 07 04-21_19_40

scottcwilson:
Exciting stuff!  A new era for Zen Cart.

scottcwilson:
It would be great to see a Laravel based plugin as a demonstration of what's available.

zcwilt:
Gimme a chance :)

Actually there is still some chicken/eggery to sort out.

Example :-

Currently the list of installed plugins is built after Laravel bootstraps.
While this does mean I can re-write that code to use eloquent etc, the list of installed plugins is then not available to any
routes used by Laravel. 

I could build the installed plugins before Laravel boots, but then I couldn't use any Laravel code to do that.

What should be done is for Laravel to build the installed plugin list on boot and then make that list available by binding it to the Laravel container. This can then be pulled out later in Zen Cart code.

I would also like Laravel routes to be able to serve back json responses to ajax requests. However this has problems to solve as well, not least session management and the fact that currently Laravel terminates that application on serving it's response.

All of this is solvable, but best done in little increments.

scottcwilson:
LOL didn't mean to rush you!  Just making sure it was on your roadmap.

PR #3662, opened 2020-07-05
Add mention: product feature auto-enable based on available date.

mc12345678:


scottcwilson:
Thanks for catching this (and the other changed files filename issue).

PR #3663, opened 2020-07-05
Saw a potential refactor for this repetitive code.

mc12345678:
When reviewing the code and observering a previous issue that
came from a basic "copy/paste" problem, saw what might be an
improved code setup to support these basic "repetitive"
operations.

### DIFF
--- a/admin/admin_page_registration.php
+++ b/admin/admin_page_registration.php
@@ -51,7 +51,7 @@ if (isset($_POST) && !empty($_POST)) {
             if (empty(${$key})) {
                 $error = TRUE;
                 $messageStack->add(constant('ERROR_' . strtoupper($key) . '_NOT_ENTERED'), 'error');
-            } else if ((!empty($value['elseifFunc']) && $value['elseifFunc'](${$key})) ? empty($value['elseifNot']) /* result was true */ : !empty($value['elseifFunc']) && !empty($value['elseifNot']) /*result was false */) {
+            } else if (!empty($value['elseifFunc']) && ($value['elseifFunc'](${$key})) ? /* result was true */ empty($value['elseifNot']) : /* result was false */ !empty($value['elseifNot'])) {
                 $error = TRUE;
                 $message = 'ERROR_' . strtoupper($key) . '_ALREADY_EXISTS';
                 if ($value['elseifFunc'] == 'defined' && !empty($value['elseifNot'])) {
--- a/admin/admin_page_registration.php
+++ b/admin/admin_page_registration.php
@@ -15,60 +15,57 @@ $sort_order = 0;
 if (isset($_POST) && !empty($_POST)) {
   $error = FALSE;
 
-  if (isset($_POST['page_key'])) {
-    $page_key = zen_db_prepare_input($_POST['page_key']);
-  }
-  if (empty($page_key)) {
-    $error = TRUE;
-    $messageStack->add(ERROR_PAGE_KEY_NOT_ENTERED, 'error');
-  } elseif (zen_page_key_exists($page_key)) {
-    $error = TRUE;
-    $messageStack->add(ERROR_PAGE_KEY_ALREADY_EXISTS, 'error');
+  $keys = [
+            'page_key' => [
+                            'empty' => true,
+                            'elseif' => 'zen_page_key_exists',
+                          ],
+            'language_key' => [
+                                'empty' => true,
+                                'elseif' => 'defined',
+                                'elseifNot' => true,
+                              ],
+            'main_page' => [
+                            'empty' => true,
+                            'elseif' => 'defined',
+                            'elseifNot' => true,
+                           ],
+            'page_params',
+            'menu_key' => [
+                            'empty' => true,
+                          ],
+            'sort_order' => [
+                              'valueType' => 'integer',
+                            ],
+          ];
+
+  foreach ($keys as $key => $value) {
+    if (isset($_POST[$key])) {
+      if (!empty($value['valueType'])) {
+        ${$key} = $db->bindVars(':val:', ':val:', $_POST[$key], $value['valueType']);
+      } else {
+        ${$key} = zen_db_prepare_input($_POST[$key]);
+      }
+    }
+    if (!empty($value['empty']) && !empty($value['elseif'])) {
+      if (empty(${$key})) {
+        $error = TRUE;
+        $messageStack->add(constant('ERROR_' . strtoupper($key) . '_NOT_ENTERED'), 'error');
+      } else if ((!empty($value['elseif']) && $value['elseif'](${$key})) ? empty($value['elseifNot']) /* result was true */ : !empty($value['elseif']) && !empty($value['elseifNot']) /*result was false */) {
+        $error = TRUE;
+        $message = 'ERROR_' . strtoupper($key) . '_ALREADY_EXISTS';
+        if ($value['elseif'] == 'defined' && !empty($value['elseifNot'])) {
+          $message = 'ERROR_' . strtoupper($key) . '_HAS_NOT_BEEN_DEFINED';
+        }
+        $messageStack->add(constant($message), 'error');
+      }
+    }
   }
 
-  if (isset($_POST['language_key'])) {
-    $language_key = zen_db_prepare_input($_POST['language_key']);
-  }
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks for the suggestion.

I'm still strongly leaning toward deleting this entire page, in favor of requiring one to write code to insert menu items.

admin/admin_page_registration.php
```diff
@@ -15,60 +15,57 @@
...
+                            'elseifNot' => true,
+                           ],
+            'page_params',
+            'menu_key' => [
+                            'empty' => true,
+                          ],
+            'sort_order' => [
+                              'valueType' => 'integer',
+                            ],
+          ];
```
drbyte, line 40: Let's stick to PSR-2 for indenting. I find the following much more readable than indenting by "2 chars longer than the last character of the previous line":
```suggestion
    $keys = [
        'page_key' => [
            'empty' => true,
            'elseif' => 'zen_page_key_exists',
        ],
        'language_key' => [
            'empty' => true,
            'elseif' => 'defined',
            'elseifNot' => true,
        ],
        'main_page' => [
            'empty' => true,
            'elseif' => 'defined',
            'elseifNot' => true,
        ],
        'page_params',
        'menu_key' => [
            'empty' => true,
        ],
        'sort_order' => [
            'valueType' => 'integer',
        ],
    ];
```

mc12345678: Glad to stick with that, haven't been sure what the expectation has been for such array "lists".
Looking to apply to the bulk of the code, though did also see an operational change that was needed. Both are being committed separately with the PSR-2 change done all in one commit and after the logic correction.

admin/admin_page_registration.php
```diff
@@ -15,60 +15,57 @@
...
-    $messageStack->add(ERROR_MENU_NOT_CHOSEN, 'error');
+  foreach ($keys as $key => $value) {
+    if (isset($_POST[$key])) {
+      if (!empty($value['valueType'])) {
+        ${$key} = $db->bindVars(':val:', ':val:', $_POST[$key], $value['valueType']);
+      } else {
+        ${$key} = zen_db_prepare_input($_POST[$key]);
+      }
+    }
+    if (!empty($value['empty']) && !empty($value['elseif'])) {
```
mc12345678, line 50: The code inside of this if group can handle `$value['elseif']` being empty (specifically written to handle) so this second "check" isn't applicable.

admin/admin_page_registration.php
```diff
@@ -5,91 +5,88 @@
...
-    $error = TRUE;
-    $messageStack->add(ERROR_LANGUAGE_KEY_NOT_ENTERED, 'error');
-  } elseif (!defined($language_key)) {
-    $error = TRUE;
-    $messageStack->add(ERROR_LANGUAGE_KEY_HAS_NOT_BEEN_DEFINED, 'error');
-  }
+    foreach ($keys as $key => $value) {
+        if (isset($_POST[$key])) {
+            if (!empty($value['valueType'])) {
+                ${$key} = $db->bindVars(':val:', ':val:', $_POST[$key], $value['valueType']);
```
drbyte, line 45: Can't we just cast to integer if $key = sort_order?

admin/admin_page_registration.php
```diff
@@ -5,91 +5,88 @@
...
+            } else {
+                ${$key} = zen_db_prepare_input($_POST[$key]);
+            }
+        }
+        if (!empty($value['empty'])) {
+            if (empty(${$key})) {
+                $error = TRUE;
+                $messageStack->add(constant('ERROR_' . strtoupper($key) . '_NOT_ENTERED'), 'error');
+            } else if ((!empty($value['elseif']) && $value['elseif'](${$key})) ? empty($value['elseifNot']) /* result was true */ : !empty($value['elseif']) && !empty($value['elseifNot']) /*result was false */) {
```
drbyte, line 54: The following is gonna sound highly critical. I hope it doesn't come off snobbish. It's primarily meant to foster discussion to create an understanding of preferred style, as I know you've invited these discussions elsewhere.

This `foreach/if/if/else/if/if/elseif/elseifNot/if(elseif)` loop refactoring makes my head spin.
I "get" it.
It's clever. 
Unfortunately next time I look at this code I'll have to spend 30 minutes figuring out what it's doing, all over again.

I appreciate the refactor.
And still I prefer to avoid the "arrow" anti-pattern, eg: if/if/else/if/if/else-if/if.

I don't follow the purpose of `empty`. Everything's `empty`, other than sort_order. Why not simplify it by just skipping sort_order when going through the loop?

admin/admin_page_registration.php
mc12345678: Sorry for delay in response, other browser wouldn't let me write in the "code review" type box.

Was trying to use/develop code that was independent of some sort of specific "value" or setting.. In a way I would prefer all of the keys to have a `valueType` and just process it/them to the associated value noting that at this point in execution the data is expected to have already gone through sanitization (to whatever level the store has active).

admin/admin_page_registration.php
mc12345678: *I* certainly get it and a reason I put it out there... Most of the refactoring that has been happening causes the code to look nothing like it used to look, but expected to still function as it did... A smidge of "history" on this one, I wanted to literally take it one for one and even use the `elseif` key of `!defined`; however, doesn't work that way... :( So then after quite a lot of thought I finally figured out a logic path that *could* work and trying to consider some realm of "expansion" or reuse... Not all code can be that way, but I wanted to try to start as strong that way as possible.

In most of the cases, the `empty` test is intended to determine if the array's `key` has a value against which to take action and to consider the strict php case(s)... I certainly don't doubt that there are some (a lot?) of readability or other improvements possible. in earnest I don't mind if this stays open for a really long time... It just seems like there is/was an improved way to accomplish this repetitive series of checks and not require every possible array key to be fully filled in.

*Note*: both `sort_order` **and** `page_params` don't use the `empty` check...  The key `empty` implies to test if the variable/key is empty and if it is to take one action, if not to then do/try the second test/check if it is to be performed...

scottcwilson:
This helped people get over the hump during the 1.3->1.5 transition, but it has outlived its usefulness. We should get rid of it to simplify the product.

mc12345678:
I'm surprised that there is anything that provides anything associated with ZC based on some of the "let's just get rid of it" statements.

Other than a minor bug (test to provide an error message) that had existed for 2 years, this code has been a relatively solid example of what is needed to support developing code to actually do the page add... Assuming that it is desired at the time of code load to actually offer the menu option.  I have actually been asked for a plugin to be present on the server, but to take no action until there has been a determination for it to be "installed" at which point the page_key related changes are made. While that is basically done by accessing one of the files in the fileset, it could just as "easily" be incorporated/considered to use the admin page registration tool... And actually, instead of this offering less options, it would be nice if it offered more such as a way to retrieve/modify existing page keys... If for no other reason than to be able to adjust/play with sort orders...

admin/admin_page_registration.php
```diff
@@ -5,91 +5,88 @@
...
+            ${$key} = zen_db_prepare_input(${$key});
+        }
+        if (!empty($value['emptyChk'])) {
+            if (empty(${$key})) {
+                $error = TRUE;
+                $messageStack->add(constant('ERROR_' . strtoupper($key) . '_NOT_ENTERED'), 'error');
+            } else if (!empty($value['elseifFunc']) && ($value['elseifFunc'](${$key})) ? /* result was true */ empty($value['elseifNot']) : /* result was false */ !empty($value['elseifNot'])) {
+                $error = TRUE;
+                $message = 'ERROR_' . strtoupper($key) . '_ALREADY_EXISTS';
+                if ($value['elseifFunc'] == 'defined' && !empty($value['elseifNot'])) {
```
mc12345678, line 57: Almost seems that for proper logic that a separate key should be used here instead of being dependent on `elseifNot` being set... Yes, this was "born" of the original data; however, currently there are two types of test here, that there is either a duplicate or an absence of information... Thing is though, that the current logic of if it is to check that it is `not defined` (perhaps should swap the two tests for a more "logical" sequence) that then these dependencies are correct and an additional key would only make things easier to muck up.

drbyte:
Leaving this on hold.

PR #3668, opened 2020-07-07
Extract template functions from template_select

drbyte:
Removes all non-pagination-related queries into separate functions

### DIFF
--- a/includes/functions/functions_templates.php
+++ b/includes/functions/functions_templates.php
@@ -1,0 +1,115 @@
+<?php
+/**
+ * @copyright Copyright 2003-2020 Zen Cart Development Team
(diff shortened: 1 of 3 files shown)


includes/functions/functions_templates.php
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Get all template directories found in catalog folder structure
+ *
+ * @return array
+ */
+function zen_get_catalog_template_directories()
+{
+    if (!defined('DIR_FS_CATALOG_TEMPLATES')) {
+        die('Fatal error: DIR_FS_CATALOG_TEMPLATES not defined.');
```
lat9, line 19: How about a `trigger_error` here, and below, prior to the `die` so that there's a record of the backtrace associated with the issue?

PR #3685, opened 2020-07-07
Remove pre PHP 5.6 hash method reference

scottcwilson:


drbyte:
Thanks.

Can you PR a removal of the `password_compat` stuff that polyfilled these features for now-unsupported PHP versions? (and of course test that it's not broken as a result)

scottcwilson:
Yes but please create a ticket and assign it to me so I don't lose track of it.

PR #3691, opened 2020-07-08
Refactor repetitive category/product queries into shared functions

drbyte:


includes/functions/functions_products.php
```diff
@@ -127,3 +126,169 @@ function zen_enable_disabled_upcoming() {
...
     }
 }
+
+
+/**
+ * Return a product's category (master_categories_id)
+ * @param int $products_id
+ * @return int|string
+ */
+function zen_get_products_category_id($products_id) {
```
mc12345678, line 136: This line isn't factored as others, recommend:
````diff
- function zen_get_products_category_id($products_id) {
+ function zen_get_products_category_id($products_id)
+{
````

drbyte: Thanks. File updated.

PR #3692, opened 2020-07-08
Force conversion to float

mc12345678:
Using MySQL 5.7, the cast to float was not being applied to the result of the if statement, but instead basically to the if test and a fatal error to be logged (and possibly/eventually a duplicate order number to be generated). This corrects the casting to apply to whatever result is to be used for `settle_amount`.

Applies to ZC 1.5.6 also...

### DIFF
--- a/includes/modules/payment/paypalwpp.php
+++ b/includes/modules/payment/paypalwpp.php
@@ -533,7 +533,7 @@ if (false) { // disabled until clarification is received about coupons in PayPal
                           'mc_gross' => (float)$this->amt,
                           'mc_fee' => (float)urldecode($this->feeamt),
                           'mc_currency' => $this->responsedata['PAYMENTINFO_0_CURRENCYCODE'],
-                          'settle_amount' => (float)(isset($this->responsedata['PAYMENTINFO_0_SETTLEAMT'])) ? $this->urldecode($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) : $this->amt,
+                          'settle_amount' => (float)(isset($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) ? $this->urldecode($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) : $this->amt),
                           'settle_currency' => $this->responsedata['PAYMENTINFO_0_CURRENCYCODE'],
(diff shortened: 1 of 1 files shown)


includes/modules/payment/paypalwpp.php
```diff
@@ -533,7 +533,7 @@ function after_process() {
                           'mc_gross' => (float)$this->amt,
                           'mc_fee' => (float)urldecode($this->feeamt),
                           'mc_currency' => $this->responsedata['PAYMENTINFO_0_CURRENCYCODE'],
-                          'settle_amount' => (float)(isset($this->responsedata['PAYMENTINFO_0_SETTLEAMT'])) ? $this->urldecode($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) : $this->amt,
+                          'settle_amount' => (float)(isset($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) ? $this->urldecode($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) : $this->amt),
```
mc12345678, line 536: As a side note.. Where is the class' function urldecode identified such that `$this->urldecode` actually does something/exists?

mc12345678: Hmmm.. Also thinking about this proposed change. Not sure (don't know) if it "matters", but as revised:
`if isset(responsedata), return the float of the responsedata, if not set then return the float of amt`
Original:
`if the float of responsedata is true (non-zero), return responsedata, if responsedata has a 0 value (or only text) then return amt`

The difference is that it seems that in the original code if the responsedata (withkey) was provided and the numerical portion was 0 or non-existent, then `$this->amt` (whatever it is/was) would be returned.  In the proposed revision, `$this->amt` will be returned only if the associated responsedata is not provided back... 

If the original "logic" was intended, then the original line should be changed to:
````diff
- (float)(isset($this->responsedata['PAYMENTINFO_0_SETTLEAMT'])) ? $this->urldecode($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) : $this->amt,
+ (float)((float)(isset($this->responsedata['PAYMENTINFO_0_SETTLEAMT'])) ? $this->urldecode($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) : $this->amt),
````
So that ultimately `settle_amount` is only the/a numerical value as returned by the casting to `float`...
I do not know at this time which way is correct, but I *do* know that currently a number with associated text is returned by the original code and that MySQL 5.7 (also with php 5.6) doesn't like the result because it is not provided as a number only...

drbyte:
Thanks

PR #3695, opened 2020-07-08
Products_to_categories: fix for current category not being maintained between product selection changes

torvista:


admin/products_to_categories.php
```diff
@@ -20,7 +20,13 @@
...
 
 // current_category: the category the selected product is in
-$current_category_id = (int)(isset($_GET['current_category_id']) ? $_GET['current_category_id'] : $current_category_id);
+if (!empty($_POST['current_category_id'])) { // page load from change of selection in product listing form
+    $current_category_id = (int)$_POST['current_category_id'];
+} elseif (!empty($_GET['current_category_id'])) { // page load from a redirect
+    $current_category_id = (int)$_GET['current_category_id'];
+} else {
+    $current_category_id = 0;
+}
```
drbyte, line 29: In 1.5.8 since we require minimum PHP 7 this would work too:
```php
$current_category_id = (int)($_POST['current_category_id'] ?? $_GET['current_category_id'] ?? 0)
```

scottcwilson: YUCK!

torvista: I just used similar to what was already there: changed them to use the php7 constructs.

scottcwilson: I'm referring to the use of ??.  Ugh.  I greatly prefer what you did.

torvista: 😁

torvista:
I hope you are not all yelling at me...I only followed a suggestion.

I think the original triple block is ugly. I guess those php people did too, hence they introduced the concise form.

Easier to understand the original?..yes of course, for people like me who dip into code when they are forced to.

But,
I see laravel and the accompanying "explanatory" comments piling in thick and fast and feel I'm back to 10 years ago/I haven't got a clue about any of it....so I doubt I'll be contributing much for a long time/I'm well out of my comfort zone. But, I assume the pro developers are lapping it up, so good luck to them and the project.

So, relatively, I think the changes here in terms of losing "understandability" are pretty negligible in comparison to what else is going on/not really something to shout about.

Now you can shout at me.

scottcwilson:
This doesn't improve the code or fix a bug.  There's no need to make this change (and certainly no need to make a bunch more just like it.)

zcwilt:
On the comments regarding Laravel.
This may seem like a huge change, but it really isn't.
Yes, over time we want to move to using more of the Laravel infrastructure, but truth is that will be a slow process with 
lots of time for people to get used to using the new code.

In fact despite what seems like large changes, the actual changes to Zen Cart core code are negligible.

The only place that any Laravel code has been used so far is in reading in the configuration keys.

so this code 

```
$configuration = $db->Execute('SELECT configuration_key AS cfgkey, configuration_value AS cfgvalue, configuration_group_id
                                 FROM ' . TABLE_CONFIGURATION, '', $use_cache, 150);
while (!$configuration->EOF) {
    if ($configuration->fields['configuration_group_id'] == 2 || $configuration->fields['configuration_group_id'] == 3) {
        define(strtoupper($configuration->fields['cfgkey']), (int)$configuration->fields['cfgvalue']);
    } else {
        define(strtoupper($configuration->fields['cfgkey']), $configuration->fields['cfgvalue']);
    }
    $configuration->MoveNext();
```

changed to this code 

```
public function loadConfigSettings()
{
    $configs = self::all();
    foreach ($configs as $config) {
        $configValue = $config['configuration_value'];
        if (in_array($config['configuration_group_id'], [2, 3])) {
            $configValue = (int)$configValue;
        }
        define(strtoupper($config['configuration_key']), $configValue);
    }
}
```

PR #3696, opened 2020-07-09
Remove excess $this->

### LINKED ISSUE #3693: Paypal Express - Duplicate Orders with new Order number with stricter MySQL
Discovered this issue on a ZC 1.5.6 site that is using MySQL 5.7 (php 5.6 currently), and see that the same code exists in ZC 1.5.7.

When a purchase is being processed for immediate claim, the data generated includes the price of the item and the currency.  To store the data, the `settle_amount` is expected to be captured.  The `settle_amount` field is of the data type `decimal(15,4)`... If the text is not "removed", then MySQL throws a fatal error from this assignment: 
ZC 1.5.6:
https://github.com/zencart/zencart/blob/e60c81b6dbac7b1a61fb891397fe20de7b02c623/includes/modules/payment/paypalwpp.php#L544

ZC 1.5.7:
https://github.com/zencart/zencart/blob/07ec9d08de4937aa162edb9f5d767729afd295a0/includes/modules/payment/paypalwpp.php#L536

The result of the fatal error appears to be that an order number is generated which contains all of the paypal information; however, the customer doesn't necessarily seem to be able to fully checkout. After the button times out, the customer can click again, a new order number is generated but the same payment information is captured/collected and it seems this cycle continues...

See #3692 for proposed solution.

Also, I have a feeling that because of another issue in that line `$this->urldecode()` instead of `urldecode()` that information from authorizations is not properly captured back into the database as a result of the code "stopping" at that point... I believe I saw someone posting an issue with authorizations somewhere in the forum and not being able to see the associated data in the ZC admin orders screen... These two issues may be why... I do not (yet) know how the data looks that is returned from: `$this->responsedata['PAYMENTINFO_0_SETTLEAMT']`

mc12345678:
Closes #3693 

Finally figured it out... There were a series of commits made for ZC 1.5.7 to add some "functionality" (processing/storage of `$this->amt` when `$this->responsedata['PAYMENTINFO_0_SETTLEAMT']` is not available). In making the commits, there was one mistake made (absence of `$this->` from `$this->responsedata['PAYMENTINFO_0_SETTLEAMT']`; however, it had been accidentally added to `urldecode` in the original commit (https://github.com/zencart/zencart/commit/03bb26ba50d7fecd472ccda0f1affe55e3ad4f37). When this discrepancy was identified (the absence of `$this->` from `$this->responsedata['PAYMENTINFO_0_SETTLEAMT']`), it was corrected with another commit (https://github.com/zencart/zencart/commit/75bb26f76d75753b7c3d70d5e4bc4b19b5becf2a#diff-1b7e5e4f394173368732a571b5c8a0d5R539); however, the additional `$this->` on `urldecode` was not noticed/corrected.  Then along comes stricter processing/code evaluation where it was desired to cast either of those two results to a float, which seems to happen automatically in "earlier" MySQL default server installations (can be made possible in more recent versions by relaxing some of the default settings).  A separate commit was issued to address the float casting: https://github.com/zencart/zencart/pull/3692/commits/b781a1b92e7f934f8582b58a3595a11e419bd053 where in one of the comments the changes of this commit are discussed... Now having reviewed the history of this line, the change of this commit appears to be the correct "solution" assuming that the value of `$this->responsedata['PAYMENTINFO_0_SETTLEAMT']` still is "available" and of any importance such that it would still go through `urldecode` because it would be set...

Oye…

This is applicable to ZC 1.5.6 as well...

### DIFF
--- a/includes/modules/payment/paypalwpp.php
+++ b/includes/modules/payment/paypalwpp.php
@@ -533,7 +533,7 @@ if (false) { // disabled until clarification is received about coupons in PayPal
                           'mc_gross' => (float)$this->amt,
                           'mc_fee' => (float)urldecode($this->feeamt),
                           'mc_currency' => $this->responsedata['PAYMENTINFO_0_CURRENCYCODE'],
-                          'settle_amount' => (float)(isset($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) ? $this->urldecode($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) : $this->amt),
+                          'settle_amount' => (float)(isset($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) ? urldecode($this->responsedata['PAYMENTINFO_0_SETTLEAMT']) : $this->amt),
                           'settle_currency' => $this->responsedata['PAYMENTINFO_0_CURRENCYCODE'],
                           'exchange_rate' => (isset($this->responsedata['PAYMENTINFO_0_EXCHANGERATE']) && urldecode($this->responsedata['PAYMENTINFO_0_EXCHANGERATE']) > 0) ? urldecode($this->responsedata['PAYMENTINFO_0_EXCHANGERATE']) : 1.0,
                           'notify_version' => '0',


scottcwilson:
What is the symptom one would see because of bug?  I would have expected a blank screen when $this->urldecode was called, but haven't seen any reports of one.

mc12345678:
Well, although like many PayPal "result" values, I've not been able to find `PAYMENTINFO_0_SETTLEAMT` as a value that might be returned; however, the context appears to be as a result of committing an authorization possibly from within paypal, or as a part of a process load by way of an ipn response.  As such (and still making an assumption), I would not expect the operator to "see" a blank screen as far as a white screen of death, but instead likely would *not* see some level of result in the admin orders screen for submitting the total associated with an authorization type purchase.

The above is stated from an attempt to understand what data is being captured in the effected line, that prior to the other update performed that it appears that `PAYMENTINFO_0_SETTLEAMT` was being used (or perhaps never had a value), and that it remained a part of the line's logic with an understanding of how easy it was to generate the coding error that is present (assuming that portion of the code is ever executed).  If the determination is that `PAYMENTINFO_0_SETTLEAMT` should not even be a part of consideration, then this line should be further "reduced" to not include it and then possibly consider even no longer capturing the data (retiring in place).  There does appear to be at least a minor reason to keep it as the value stored is a number (without any units) which makes it possible to do math/conversion on the stored value even if that same "math" could be done from one of the other stored fields...

PR #3704, opened 2020-07-11
Remove unused code

scottcwilson:


admin/coupon_admin.php
```diff
@@ -1373,15 +1373,6 @@ function check_form(form_name) {
...
       break;
-    case 'voucherreport':
-      $heading[] = array('text' => '<b>' . TEXT_HEADING_COUPON_REPORT . '</b>');
-      $contents[] = array('text' => TEXT_NEW_INTRO);
-      break;
-// base code - header for report for basecode
-    case 'voucherreportduplicates':
-      $heading[] = array('text' => '<b>' . TEXT_HEADING_COUPON_REPORT . '</b>');
-      $contents[] = array('text' => TEXT_NEW_INTRO);
-      break;
```
drbyte, line 1384: I'm puzzled how this came to be.

What caused you to stumble upon it?

scottcwilson: Using phpStorm's code inspection feature.  It was reporting a different issue (the fact that these two branches are identical, but then I found that the define TEXT_HEADING_COUPON_REPORT didn't exist, then I realized these blocks weren't used.

PR #3708, opened 2020-07-11
Fix notices created by this file

scottcwilson:
@zcwilt the foreach block at line 71 also needs review - you're looping but overwriting the data each time.


zcwilt:
Fixed by https://github.com/zencart/zencart/pull/3710

PR #3711, opened 2020-07-11
Do basic logging even if Report All Errors is false

scottcwilson:
<strike>Recommended for backport to 1.5.7. </strike>

1.5.7 handles this situation correctly.   No backport needed.

### DIFF
--- a/admin/includes/init_includes/init_report_all_errors_admin.php
+++ b/admin/includes/init_includes/init_report_all_errors_admin.php
@@ -26,6 +26,9 @@ if (!defined('REPORT_ALL_ERRORS_ADMIN')) {
 } elseif (REPORT_ALL_ERRORS_ADMIN != 'No') {
     @ini_set('error_reporting', E_ALL);
     set_error_handler('zen_debug_error_handler', E_ALL);
+} else {
+    @ini_set('error_reporting', E_ALL & ~E_DEPRECATED & ~E_NOTICE & ~E_STRICT); 
+    set_error_handler('zen_debug_error_handler', E_ALL);
 }
 
 if (!defined('REPORT_ALL_ERRORS_NOTICE_BACKTRACE')) {
--- a/includes/init_includes/init_report_all_errors.php
+++ b/includes/init_includes/init_report_all_errors.php
@@ -10,4 +10,7 @@
 if (defined('REPORT_ALL_ERRORS_STORE') && REPORT_ALL_ERRORS_STORE != 'No') {
     @ini_set('error_reporting', E_ALL );
     set_error_handler('zen_debug_error_handler', E_ALL);
+} else {
+    @ini_set('error_reporting', E_ALL & ~E_DEPRECATED & ~E_NOTICE & ~E_STRICT); 
+    set_error_handler('zen_debug_error_handler', E_ALL);
 }


lat9:
Why is this an issue on zc158?  Has something changed in the base `/includes/extra_configures/enable_error_logging.php` that necessitates these changes?

scottcwilson:
You can replicate this by turning off store side logging then breaking a file in includes/functions with a syntax error.  You'll get a 500 but no log.

scottcwilson:
If you back out this fix:
https://github.com/zencart/zencart/commit/3e63d81a252711d9a3ef9cdcc16a766c7c114632

(fix for a syntax error in a SQL query), and have Admin > Configuration > Logging > Report All Errors (Store) = No,  
you'll see the problem that gave rise to this PR.

lat9:
The source of the issue is in `/app/Providers/AppServiceProvider.php`.  Error reporting is 'set' by `/includes/extra_configures/enable_error_logging.php` and gets reset during this call from application_top.php:
```
require DIR_FS_CATALOG . 'includes/illuminate_bootstrap.php';
```

The processing that's squishing the logging is that _AppServiceProvider_'s `boot` method:
```
    public function boot()
    {
        /**
         * set the level of error reporting
         *
         * Note STRICT_ERROR_REPORTING should never be set to true on a production site. <br />
         * It is mainly there to show php warnings during testing/bug fixing phases.<br />
         */
        if (DEBUG_AUTOLOAD || (defined('STRICT_ERROR_REPORTING') && STRICT_ERROR_REPORTING == true)) {
            @ini_set('display_errors', TRUE);
            error_reporting(defined('ILLUMINATE_ERROR_REPORTING_LEVEL') ? ILLUMINATE_ERROR_REPORTING_LEVEL : 0);
        } else {
            error_reporting(0);
        }
        if ($this->app->runningInConsole()) return;

        $pluginManager = new PluginManager(new PluginControl, new PluginControlVersion);
        $installedPlugins = $pluginManager->getInstalledPlugins();
        $this->app->instance('installedPlugins', $installedPlugins);
    }
```

My suggestion is to correct this at the source by removing the error_reporting changes provided there.

zcwilt:
yeah 

The line 

     error_reporting(defined('ILLUMINATE_ERROR_REPORTING_LEVEL') ? ILLUMINATE_ERROR_REPORTING_LEVEL : 0);

is incorrect. The final part of the ternary is currently setting to 0 rather than E_ALL as would normally happen.\

The changes were made to allow Laravel to handle the errors, in which case you get the whoops error pages that Laravel provides.

zcwilt:
Ignore the above, as it's not strictly true.
could have deleted the comment but it shows what my intention was, even if wrong ;)

gimme a minute to get my brain into gear while I roast a chicken.

lat9:
> The changes were made to allow Laravel to handle the errors, in which case you get the whoops error pages that Laravel provides.

Why would Zen Cart defer its error-handling to Laravel?

zcwilt:
@lat9
The changes were meant to allow Laravel to handle errors, but only if you had set defines(in some extra_configures file)  to do that.

The changes were also meant that given what Zen Cart might set as it's error reporting/logging Laravel would never get involved.

There are benefits in allowing Laravel to manage logging, but this is something for the future, and again if/when done **must** be a seamless change, 

The problem we are seeing at the moment is complicated, but fixable and i'm working  on a PR now, which will be up today.

zcwilt:
Think this can be closed based on #3720.
Have tested with sql error and it is logged correctly

PR #3713, opened 2020-07-11
Changed search order for email templates to use $module (not current_page_base)

scottcwilson:
In the file includes/modules/pages/contact_us/header_php.php

if you call

zen_mail($send_to_name, $send_to_email, EMAIL_SUBJECT, $text_message, $name, $email_address, $html_msg, 'contact_us_special');

the template used should be email/email_template_contact_us_special.html (assuming that exists), not
email/email_template_contact_us.html just because zen_mail was called from the contact_us page.

The $module parameter in the call, not the $current_page_base, should be the top priority.


drbyte:
The `EMAIL_TEMPLATE_FILENAME` index was added for that specific purpose, instead of changing the module name.

scottcwilson:
I just find it super surprising that $current_page_base is used in determining which template to use.  Am I all wet?

scottcwilson:
Also, EMAIL_TEMPLATE_FILENAME is still below $current_page_base in the search order.

PR #3714, opened 2020-07-11
Update objectinfo annotations to avoid code-inspect false-positives

scottcwilson:
It would be faster and still probably completely safe to just return a string without checking but just in case I added a check.

Doing this means that we can use the Type compatibility check in phpStorm (and other debuggers).

### DIFF
--- a/admin/includes/classes/object_info.php
+++ b/admin/includes/classes/object_info.php
@@ -55,6 +55,10 @@ class objectInfo
         $this->$field = $value;
     }
 
+    /**
+     * @param $field
+     * @return array|string
+     */
     public function __get($field)
     {
         if (isset($this->$field)) return $this->$field;


drbyte:
What's the scenario where this became needed?

Help me understand the specific workflow.

scottcwilson:
Right now when you run Code Inspect in phpStorm, you get > 100 false positives because the properties in an object created by objectInfo are not cast as strings.  This means when they're used with methods that expect strings, the checker can't verify type compatibility. 

If we did this change, it would fix all these errors, meaning that reports coming from Code Inspect would be more useful. (At least for this category of error.)

zcwilt:
objectInfo is a horrible solution to a problem caused initially by the fact that result sets produced by the queryFactory weren't 
iterable.

zcwilt:
should have added a /rant to the above

zcwilt:
In the code for the updated layout box controller #3723 we use Eloquent to do the main query for displaying row information.
Because Eloquent implements ArrayAccess we can do the follwing 

    foreach ($layoutBoxes as $index => $layoutBox) {

then rather than creating an objectInfo class we can just do

    $bInfo = $layoutBoxes[$index];

Unfortunately  a lot of our code is still using the old style 

    while (!$results->EOF)
    ... 
    $results->moveNext()

however even changing this to a foreach (as say the countries page does) doesnt help as the 
QueryFactoryResults class doesn't implement ArrayAccess

scottcwilson:
But in many cases an objectInfo is created from a single result - a single customer record or special, for example. Ability to iterate won't help in those cases.

scottcwilson:
I think this could be tightened up even further to just always return a cast to string when `isset($this->$field)` is true.

scottcwilson:
OK it looks like just adding a PHPDoc block will suppress the type compatibility warnings, so this change is now comments-only.

PR #3716, opened 2020-07-11
Remove old PHP checks

### LINKED ISSUE #2586: Get rid of old PHP_VERSION checks for 5.5 and below

scottcwilson:
All pre PHP 7 checks can be removed now that we are fully PHP7. 

Fixes #2586.

### DIFF
--- a/includes/classes/observers/auto.non_captcha_observer.php
+++ b/includes/classes/observers/auto.non_captcha_observer.php
@@ -54,11 +54,10 @@ class zcObserverNonCaptchaObserver extends base
 
     protected function generate_random_string($input, $strength = 16)
     {
-        $function = PHP_VERSION_ID >= 70000 ? 'random_int' : 'mt_rand';
         $input_length = strlen($input);
         $random_string = '';
         for ($i = 0; $i < $strength; $i++) {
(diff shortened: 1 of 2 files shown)


drbyte:
I'd prefer to keep the Square module in sync with the official release.

Unfortunately that means it needs to stay compatible with older PHP versions.

scottcwilson:
But Square for Zen Cart will be running in the context of Zen Cart, which means > PHP7.  I think this is still safe.

drbyte:
Here's my point: 

Before releasing new ZC versions I typically copy the latest official Square module from the plugin repo and put it into core.

That will wipe out what you've just changed here. 

So there's no point in changing the Square file in this commit.

scottcwilson:
OK I see your point.

PR #3719, opened 2020-07-12
Admin categories/products listing, display Status ... always

lat9:
... as identified in [this](https://www.zen-cart.com/showthread.php?226934-Status-button-not-showing-on-mobile&p=1370451) Zen Cart posting.

Suggest back-porting to zc157.

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -562,7 +562,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                 <?php if ($show_prod_labels || SHOW_COUNTS_ADMIN == 'true') { ?>
                   <th class="text-right hidden-sm hidden-xs"><?php echo TABLE_HEADING_QUANTITY; ?></th>
                 <?php }; ?>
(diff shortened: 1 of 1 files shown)


scottcwilson:
What if we do this for small (tablets) only and keep the code to hide on xs (phone) devices?

lat9:
> 
> 
> What if we do this for small (tablets) only and keep the code to hide on xs (phone) devices?

How about we ask on that forum-posting?  I believe that the initial report was for a phone.

wiztechinc:
If I was to use my phone , I would want the quickest way to do something. That status button should be visible at all times in my humble opinion.

torvista:
I believe that if someone is in the Admin, its for doing admin things, so all functions should be available.
If it is not pretty/things do not fit well/perfectly...so what.

How many complaints are on the forum about the admin not looking pretty enough, as opposed to "help..the buttons have gone!".

drbyte:
> > What if we do this for small (tablets) only and keep the code to hide on xs (phone) devices?
> 
> How about we ask on that forum-posting? I believe that the initial report was for a phone.

Looks like there's no response to the question posed back on the forum.

I'll merge it, but we'll probably alter the change again later during 1.5.8

PR #3720, opened 2020-07-12
Fix error logging/reporting

zcwilt:
Allow Laravel to take over reporting/logging but only if allowed to
and only during the lifetime of a Laravel request cycle
on return and if we are continuing with Zen Cart, reset the error handler
and reporting levels

### DIFF
--- a/admin/includes/application_bootstrap.php
+++ b/admin/includes/application_bootstrap.php
@@ -68,7 +68,7 @@ if (file_exists('includes/configure.php')) {
  */
 if (defined('STRICT_ERROR_REPORTING') && STRICT_ERROR_REPORTING == true) {
     @ini_set('display_errors', TRUE);
-    error_reporting(defined('STRICT_REPORTING_LEVEL') ? STRICT_REPORTING_LEVEL : E_ALL);
+    error_reporting(defined('STRICT_ERROR_REPORTING_LEVEL') ? STRICT_ERROR_REPORTING_LEVEL : E_ALL);
 } else {
     error_reporting(0);
 }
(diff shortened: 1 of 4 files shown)


zcwilt:
see #3711

scottcwilson:
Just making sure: can you test by backing out 
https://github.com/zencart/zencart/commit/3e63d81a252711d9a3ef9cdcc16a766c7c114632
and be sure a log file is generated in /logs when you go to the product page?

admin/includes/application_bootstrap.php
```diff
@@ -68,7 +68,7 @@
  */
 if (defined('STRICT_ERROR_REPORTING') && STRICT_ERROR_REPORTING == true) {
     @ini_set('display_errors', TRUE);
-    error_reporting(defined('STRICT_REPORTING_LEVEL') ? STRICT_REPORTING_LEVEL : E_ALL);
+    error_reporting(defined('STRICT_ERROR_REPORTING_LEVEL') ? STRICT_ERROR_REPORTING_LEVEL : E_ALL);
 } else {
     error_reporting(0);
```
lat9, line 73: That's still going to reset the error-reporting if that constant's not set.

zcwilt:
So the v157 code is 

```
if (DEBUG_AUTOLOAD || (defined('STRICT_ERROR_REPORTING') && STRICT_ERROR_REPORTING == true)) {
  @ini_set('display_errors', TRUE);
  error_reporting(E_ALL); 
} else {
  error_reporting(0);
}
```

The code you reference above will still set `error_reporting(E_ALL)` if you don't have the `STRICT_ERROR_REPORTING_LEVEL` define

zcwilt:
although, your comment has made me realise that the test doesn't include the `DEBUG_AUTOLOAD` test.

Silver linings etc :)

PR #3726, opened 2020-07-13
allow plugins to load from extra_configures directory

### LINKED ISSUE #3724: Files in ./admin/includes/extra_configures/ under PluginManager
It appears that PluginManager does not load these files.  Is this a bug or should they be moved?  I'm hoping it's the former so plugin authors will not need to change their work.

zcwilt:
fixes #3724

### DIFF
--- a/admin/includes/application_bootstrap.php
+++ b/admin/includes/application_bootstrap.php
@@ -166,6 +166,7 @@ $pageLoader->init($installedPlugins, $PHP_SELF, FileSystem::getInstance());
 
 $fs = FileSystem::getInstance();
 $fs->setInstalledPlugins($installedPlugins);
+$fs->loadFilesFromPluginsDirectory($installedPlugins, 'admin/includes/extra_configures', '~^[^\._].*\.php$~i');
 $fs->loadFilesFromPluginsDirectory($installedPlugins, 'admin/includes/extra_datafiles', '~^[^\._].*\.php$~i');
 
 foreach ($installedPlugins as $plugin) {
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -189,6 +189,7 @@ $installedPlugins = $laravelApp->make('installedPlugins');
 $pluginManager = new PluginManager(new App\Models\PluginControl, new App\Models\PluginControlVersion);
 
 $fs = FileSystem::getInstance();
+$fs->loadFilesFromPluginsDirectory($installedPlugins, 'catalog/includes/extra_configures', '~^[^\._].*\.php$~i');
 $fs->loadFilesFromPluginsDirectory($installedPlugins, 'catalog/includes/extra_datafiles', '~^[^\._].*\.php$~i');
 
 foreach ($installedPlugins as $plugin) {


scottcwilson:
This doesn't quite work.  The defines created in ./admin/includes/extra_configures/mod_list.php are not available when ./admin/mod_list.php is running.

zcwilt:
works for me.
I created an extra_configures entry in Display Logs
on the display logs page I dump the define and it displays fine.

scottcwilson:
I see logs saying  $built_in_boxes is null.

zcwilt:
Oh, so you have code in the extra_configures file.
files like extra_datafiles and in this case extra_configures are loaded in the context off the FileSystem class 
and therefore any variables created will be scoped into that class and not the global scope.
(note: defines automatically get created in the global scope)

scottcwilson:
Can you promote them to global scope?  I'm just worried this is going to break a bunch of things.

zcwilt:
Can't these variables be created in an init file ??
I realise it's extra work for a plugin author.

Dare I say we should be trying to get rid of globally scoped variables. 

and the answer is yes, we could (in application_top/bootstrap) get the list of files (rather than using the FileSystem class to directly load) and iteratively load them there.

drbyte:
It'd definitely be better NOT to load all those custom variables into global scope on EVERY page-load (including unrelated pages). 
If it must be global scope then it should be page-specific.

drbyte:
> I'm just worried this is going to break a bunch of things.

The painful part about upgrades is that everyone wants all old "legacy" approaches to "still work without any maintenance".
If you're reworking a plugin to work in the new architecture, it's the perfect time to houseclean.

IMO it makes sense to stop allowing plugins to do so much autoloading into global scope with all these old autoload directories.

The more we keep supporting the code that expects to run on ZC 1.3 or 1.2 the more painful it will continue to be for upgrades and moving forward.

scottcwilson:
> If it must be global scope then it should be page-specific.

Do we have page specific loading outside the language files?  These are not language variables.  I could stick them in /english/ but it would be great to have a multilanguage solution.

scottcwilson:
Please see https://github.com/zencart/documentation/pull/293 and make suggestions on how to document this.

lat9:
> 
> 
> It'd definitely be better NOT to load all those custom variables into global scope on EVERY page-load (including unrelated pages).
> If it must be global scope then it should be page-specific.

It depends on the plugin.  Some plugins (like my [Favorites Wishlist](https://vinosdefrutastropicales.com/index.php?main_page=product_info&cPath=2_8&products_id=56) span multiple pages, so there's an init_include that 'sets the stage'.

Perhaps there should be a way for a plugin to identify a globally-scoped variable that it needs?

drbyte:
Maybe we need to find a way (maybe a subdir of plugin-associated dependent files) to identify files that only need to be loaded if the plugin is actually being invoked ... instead of always loading everything it "might" use.
As Scott said, right now the only page-specific support for that is with language defines.

drbyte:
> Perhaps there should be a way for a plugin to identify a globally-scoped variable that it needs?

I suppose you could just global it and then declare it.

proseLA:
> Dare I say we should be trying to get rid of globally scoped variables.
> 
yes, please say it.  i am a fan of getting rid of globally scoped variables.


> The painful part about upgrades is that everyone wants all old "legacy" approaches to "still work without any maintenance".
> If you're reworking a plugin to work in the new architecture, it's the perfect time to houseclean.
> 
not everyone.  i am a fan of removing all legacy approaches, and am certainly more than willing to maintain my things to work in the new architecture.  i say out with the old and in with new....

> IMO it makes sense to stop allowing plugins to do so much autoloading into global scope with all these old autoload directories.

concur.

> 
> The more we keep supporting the code that expects to run on ZC 1.3 or 1.2 the more painful it will continue to be for upgrades and moving forward.

i see no reason to keep on providing support for plugins not updated in quite some time.  just my opinion.

for me, upgrades are always painful.  i see some things going on that make it  less painful, but the pain still continues...

Issue #3728, opened 2020-07-13
Plugin upgrade doesn't work yet


scottcwilson:
Tried to upgrade, got the dialog with the blah blah warning, clicked upgrade.

No log but sidebar is blank. 

<img width="1412" alt="blank_sidebar" src="https://user-images.githubusercontent.com/4391638/87336415-cea01380-c50f-11ea-937d-63bbae939b65.png">

scottcwilson:
Still some issues (see new ticket) but this part is fixed.   Closing.

PR #3731, opened 2020-07-13
updates to allow plugin upgrades

zcwilt:
For upgrades we currently only allow a class based scripted Installer class and not a plain sql file to process any changes to the 
database.

This is done for simplicity. 

If we allowed plain sql files then there would need to have a sql file for each upgrade version step, and process these sequentially 
with the possibility of failure at each step. 

Using a ScriptedInstaller class throws the responsibility on to the plugin maintainer to write code that check the version being installed and the currently installed version and intelligently upgrading any tables/alter tables and data seeding.

### DIFF
--- a/admin/includes/languages/english/lang.plugin_manager.php
+++ b/admin/includes/languages/english/lang.plugin_manager.php
@@ -29,6 +29,7 @@ $define = [
     'TEXT_NOT_INSTALLED' => 'Not Installed',
     'TEXT_INSTALL_SUCCESS' => 'Plugin installed successfully',
     'TEXT_UNINSTALL_SUCCESS' => 'Plugin un-installed successfully',
+    'TEXT_UPGRADE_SUCCESS' => 'Plugin upgraded successfully',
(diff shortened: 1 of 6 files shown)


zcwilt:
also 
blah blah

scottcwilson:
Fixes #3728 ?

admin/includes/languages/english/lang.plugin_manager.php
```diff
@@ -39,7 +40,7 @@
     'TEXT_CONFIRM_ENABLE' => 'Are you sure you want to enable this plugin?',
     'TEXT_INFO_UPGRADE' => 'Please select the version you want to upgrade to.',
     'TEXT_INFO_UPGRADE_CONFIRM' => 'Upgrade version %s',
-    'TEXT_INFO_UPGRADE_WARNING' => 'Warning: blah blah',
+    'TEXT_INFO_UPGRADE_WARNING' => '',
```
scottcwilson, line 43: Perhaps in the future, plugins could inject their own warnings?  Could be a powerful feature.

PR #3738, opened 2020-07-16
Allow dropdowns to display a defaulted value.

### LINKED ISSUE #3734: v157 default for dropdown Attributes not respected
@drbyte,
 
In trying to ensure that I was properly incorporating the behavior (duplicating whatever ZC currently does) of selecting one or more default option values within an option name, discovered that in the ZC 1.5.7 version of `includes/modules/attributes.php` that the `$selected_attribute` value is not maintained long enough for it to be usable in a dropdown.

When I override the file with the ZC 1.5.6 version then things work "correctly".

For example, using index.php?main_page=product_info&products_id=1

Modify the Model's option value so that an attribute other than the first one is default.  Expectation would be that when loading the product that the newly selected option value would be the first one seen on the screen.  But this is not the case in ZC 1.5.7. :( And I really like all the work that seemed to go into modifying this version and for it to need a change now... I haven't figured out what it may need, just have identified the issue for now.

mc12345678:
Fixes #3734

Understand that there may be other considerations because the internal `$selected_attribute = false;` is no longer in the inner loop and that this may be placed on hold while still mulling it over.  Either an alternate value/variable for dropdowns, moving code back to elsewhere within the inner loop, acceptance that a value may be retained in the inner loop that may affect an altered attributes interface, or...

### DIFF
--- a/includes/modules/attributes.php
+++ b/includes/modules/attributes.php
@@ -110,6 +110,7 @@ while (!$products_options_names->EOF) {
     $tmp_radio = '';
     $tmp_checkbox = '';
     $tmp_html = '';
+    $selected_attribute = $selected_dropdown_attribute = false; // boolean, used for radio/checkbox/select
 
     $tmp_attributes_image = '';
     $tmp_attributes_image_row = 0;
@@ -540,10 +541,11 @@ while (!$products_options_names->EOF) {
 
 
         // default
-        // find default attribute if set for default dropdown
+        // find default attribute if set. Intended for dropdown's default
         if ($products_options->fields['attributes_default'] == '1') {
-            $selected_attribute = $products_options_value_id;
+            $selected_dropdown_attribute = $products_options_value_id;
         }
+        $selected_attribute = $selected_dropdown_attribute;
 
         $products_options->MoveNext();
         // end of inner while() loop


includes/modules/attributes.php
```diff
@@ -540,7 +542,7 @@
 
 
         // default
-        // find default attribute if set for default dropdown
+        // find default attribute if set. Intended for dropdown's default.
         if ($products_options->fields['attributes_default'] == '1') {
             $selected_attribute = $products_options_value_id;
```
mc12345678, line 547: Maybe could even use a "new variable here" and not have to effect other *outer* code:
````php
$selected_attribute = isset($selected_select_attribute) ? $selected_select_attribute : false; // $selected_select_attribute ?? false; // Possible only in PHP 7.x
if ($products_options->fields['attributes_default'] == '1') {
  $selected_attribute = $selected_select_attribute = $products_options_value_id;
}
````
with `$selected_select_attribute = false` in the outer loop?

(`isset` chosen instead of `!empty` because of the potential for `$products_options_values_id` being 0 as for files, text, and any other option type that has been developed...)

This would for all intensive purposes retain the dropdown's `$selected_attribute` value and not necessarily require a new variable to otherwise be considered outside the inner loop (though those that wish to know/retain the value while in the inner loop would need to pick up on that new variable).

Ughhh, yes, I understand the consternation of deciding/contemplating which way to go... Retain existing functionality with a cleaner set of code or revise how it works and what is needed to potentially take advantage of the original ZC 1.5.7 redesign...

File: includes/modules/attributes.php [hunk 2]
```diff
@@ -540,10 +541,11 @@
 
 
         // default
-        // find default attribute if set for default dropdown
+        // find default attribute if set. Intended for dropdown's default
         if ($products_options->fields['attributes_default'] == '1') {
-            $selected_attribute = $products_options_value_id;
+            $selected_select_attribute = $products_options_value_id;
         }
+        $selected_attribute = $selected_select_attribute;
```
drbyte, line 548: Just to be sure I understand: the outcome of this code is identical to the original, except that another (presently unused) variable is now also at play, right?

drbyte:
@mc12345678 Okay, I've read and re-read the discussion in the original issue and this PR and the individual commits.

I'm in agreement with the most recent changes in terms of logic.

One improvement I think could be made: I find the use of the word "select" a bit confusing. I realize it refers to HTML `select` elements, but `select` is such a generic word. Perhaps something like `dropdown` would help everyone grok the intention very quickly when reading the code afresh, IMO.

Thus I feel like this would make sense:
```diff
-    $selected_attribute = $selected_select_attribute = false; // boolean, used for radio/checkbox/select
+    $selected_attribute = $selected_dropdown_attribute = false; // boolean, used for radio/checkbox/dropdown

...

        // find default attribute if set. Intended for dropdown's default
-            $selected_select_attribute = $products_options_value_id;
+            $selected_dropdown_attribute = $products_options_value_id;

-        $selected_attribute = $selected_select_attribute;
+        $selected_attribute = $selected_dropdown_attribute;
```

Thoughts?

mc12345678:
But... but... I just finally got used to calling dropdowns select boxes. :) MUUHAAHAA...

Absolutely agree, use of dropdown makes SOOO much more sense and would not be as forced as it was to use/reuse the term to be changed. Should it be one word or two though? 

I'll move to push the discussed change shortly.

drbyte:
> Should it be one word or two though

In this case, one. Especially since one-word means the singularity of the word is intentional, not to be confused with "drop" and "down" being possibly separate meanings/entities/references.

drbyte:
While fixing it, perhaps you can squash the commits into just one, as well?

PR #3741, opened 2020-07-16
test data for plugin manager

scottcwilson:
A plugin with two versions can be used for testing upgrades (copy and install 1.4.0 only, then copy in 1.4.1, upgrade) or initial installs (copy 1.4.1 only). 

Also added this to the .gitignore so people can keep test data in place without worrying about accidental checkin.


drbyte:
It's these massive extra_configures files that I would like to see ONLY loaded when the plugin executes, and not load on every page of every request.

Can it be made page-specific instead of global?

@zcwilt

scottcwilson:
LOVE this idea.  The trouble with extra configures is it can easily pollute the global namespace when used by inexperienced coders.

drbyte:
Ya, that's my point. 

Just need to figure out a solid way to load page-specific data/script files.

Actually, given the plugin path lookup function `$path = $pluginManager->getPluginVersionDirectory('myPluginKey', $installedPlugins)` you can do this already. Just stick a file in the plugin's dir and have it load the data script when your plugin code runs.

PR #3742, opened 2020-07-16
Fixes #3740

### LINKED ISSUE #3740: Messagestack $_SESSION content gets wiped out
The 1.5.7 changes to the messagestack class wipe out the data that is stored in $_SESSION['messageToStack'].

This report 

https://www.zen-cart.com/showthread.php?226969-Warning-message-not-showing-for-not-checking-quot-terms-and-conditions-quot-box-on-checkout

came in today.  Restoring the messageStack class constructor and size() functions fixes it, but these changes were presumably done for a reason so I want the folks who did them to weigh in.  @lat9 and @zcwilt ?

lat9:
... deferring the 'reset' of the session-based message-stack until the message-output phase.

The issue reported was a result of 'missed functionality', i.e. that the class' `size` method might be called without a subsequent `output` one.  This change simply moves the session-based messages' reset to the `output` method.  

The assumption is that since the messages are to be output that we've entered the rendering phase of a page-load so any session-based messages will either be output at this time or are no longer pertinent.

### DIFF
--- a/includes/classes/message_stack.php
+++ b/includes/classes/message_stack.php
@@ -99,6 +99,14 @@ class messageStack extends base
     {
         global $template, $current_page_base;
 
+        // -----
+        // Reset the session-based messages, now that message-output has been requested for
+        // at least one $class.  This implies that the 'templating' phase of a page's
+        // rendering is in progress and that all applicable messages will be output at this
+        // time.
+        //
+        $_SESSION['messageToStack'] = array();
+
         if ($this->size($class) === 0) {
             return;
         }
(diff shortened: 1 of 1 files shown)


scottcwilson:
Works like a charm!

scottcwilson:
Recommended for backport to Zen Cart 1.5.7.

scottcwilson:
For situations like this, let's just fix the bug and stop there.  Storeowners are going to have to manually apply this patch and so smaller is better.

scottcwilson:
Suggested re-title: Ensure session messageStack variables are shown when page reloads (or similar).

PR #3743, opened 2020-07-16
Handle Plugin Cleanup operation 

scottcwilson:
Eventually we can move the recursive delete to the File class but please merge this for now so we can get all the Plugin operations working.


scottcwilson:
LOL this has already been done and I was behind the latest.

PR #3744, opened 2020-07-16
Admin orders suppress empty payment details display 

proseLA:
i think these conditionals are better for Card on File transactions; enforcing a standard on other payment modules seems somewhat arbitrary.

furthermore, i left in this masking function.  however, again enforcing a standard as to how a payment module stores a representation of a credit card number seems heavy handed.  i think masking should only be presented for a valid credit card number.

finally, perhaps we can refactor the ```cc_validation``` class at some point and allow direct calling of the ```is_valid``` method for use here?

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -682,29 +682,44 @@ if (zen_not_null($action) && $order_exists == true) {
               <tr>
                 <td colspan="2"><?php echo zen_draw_separator('pixel_trans.gif', '1', '10'); ?></td>
               </tr>
+              <?php if (zen_not_null($order->info['cc_type'])) { ?>
               <tr>
                 <td class="main"><?php echo ENTRY_CREDIT_CARD_TYPE; ?></td>
                 <td class="main"><?php echo $order->info['cc_type']; ?></td>
               </tr>
+              <?php
+              }
+              if (zen_not_null($order->info['cc_owner'])) { ?>
               <tr>
                 <td class="main"><?php echo ENTRY_CREDIT_CARD_OWNER; ?></td>
                 <td class="main"><?php echo $order->info['cc_owner']; ?></td>
               </tr>
+              <?php
+              }
+              if (zen_not_null($order->info['cc_number'])) {
+                      require DIR_FS_CATALOG . DIR_WS_CLASSES . 'cc_validation.php';
+                      $cc_valid = new cc_validation();
+                      $cc_needs_mask = $cc_valid->validate($order->info['cc_number'], date('m'), date('y')+1);
+                  ?>
               <tr>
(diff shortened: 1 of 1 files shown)


scottcwilson:
Is there a downside to always masking?  I'm worried about people running older mods like Ceon Manual Card. (I know they shouldn't, but they do.)

proseLA:
we currently do not always mask. 

i have addressed the conditional so that masking only happens if the credit card is a valid cc number.  the previous conditional presented a masking option if the credit card did not start with a `X` or `********`.

i think it is a better idea to not hard code conditionals like this.

scottcwilson:
OK ... what if we add a method to your payment module, and just do a function_exists() on that method in `admin/orders.php` to determine whether to mask.

proseLA:
> OK ... what if we add a method to your payment module, and just do a function_exists() on that method in `admin/orders.php` to determine whether to mask.

don't like it.

i think this change makes ZC better.  it is not just for my module.

drbyte:
> finally, perhaps we can refactor the `cc_validation` class at some point and allow direct calling of the `is_valid` method for use here?

I'm very strongly leaning towards deleting that class entirely, since it's really up to the gateway whether the card is acceptable or not. The use of that class pre-dates strong real-time card verification by the gateways.

scottcwilson:
Agreed.  It's kind of an anachronism.

proseLA:
> > finally, perhaps we can refactor the `cc_validation` class at some point and allow direct calling of the `is_valid` method for use here?
> 
> I'm very strongly leaning towards deleting that class entirely, since it's really up to the gateway whether the card is acceptable or not. The use of that class pre-dates strong real-time card verification by the gateways.

just so we are clear, you want to submit credit card numbers to the gateway that do not pass Luhn Mod-10 validation?  or do you just want to have a javascript validator for mod-10 validation?  and no fallback in the event that the user does not have javascript enabled?

and users could give an expiration date in the past, that we want the gateway to tell us that the card has expired?

it would be nice to present an easy way to incorporate a modern form entry as part of the base ZC for any credit card module.  losing the mod-10 validation as well as expiration date validation is a step in the wrong direction IMO, as in detracts from the user experience with additional wait time.

drbyte:
The gateways that tokenize the card don't even let our cc-validation run because they take care of it on their end, making ours redundant (and out of date at best).
So for those gateways the cc-validation class is moot.

For modules that collect card data on-site, I'd prefer first that we don't even do that. Ever.
But if it must be done, while I'm okay with validating that a card number matches a heuristic (eg the Luhn test) that makes it worth submitting to the gateway, and that the expiry date is not old, other components of "validation" feel wrong to be maintaining, such as identifying/rejecting the card-type (which is an ever-moving-target).

PR #3749, opened 2020-07-17
Check if directory before proceeding

scottcwilson:
This helps avoid dozens of warnings.


scottcwilson:
You could even take out the @ in the following lines because you've pre-checked it with this change.

Issue #3403, opened 2020-05-03
Should Store Manager include a clear cache capability? 


scottcwilson:

drbyte:
When would this be used?

The cache is already cleared every time an Admin configuration_value is changed.

scottcwilson:
> The cache is already cleared every time an Admin configuration_value is changed.
I'm not seeing this - where is this logic? 

I'm thinking about the situation where you make a bunch of product changes and you don't want to show old data.

drbyte:
> The cache is already cleared every time an Admin configuration_value is changed.

Correction: It's cleared on every Admin page hit:

https://github.com/zencart/zencart/blob/b21f28f541b8bab6d299de9c8417f49f6cab34af/admin/includes/init_includes/init_database.php#L51-L52

scottcwilson:
Sweet!   

I wonder if the sql_cache_flush_cache 'memory' case should clear `$this->cache_array` just in case someone implements 'memory' fully in the future.

drbyte:
> I wonder if the sql_cache_flush_cache 'memory' case should clear `$this->cache_array` just in case someone implements 'memory' fully in the future.

Correct observation; however I'd leave that change for when it becomes context-relevant.

scottcwilson:
oh agreed,  I might just add a comment or something if I ever happen to be in there.  2/5 items done - whoo!

scottcwilson:
I have these changes in test locally and they look good.  I'll push a branch in a few weeks when the 1.5.7 activity settles down.

scottcwilson:
In test now.  My intention is to change the behavior described in this post https://github.com/zencart/zencart/issues/3403#issuecomment-623187248 so that a storeowner can process orders in admin without clearing the cache.

drbyte:
> My intention is to change the behavior described in this post [#3403 (comment)](https://github.com/zencart/zencart/issues/3403#issuecomment-623187248) so that a storeowner can process orders in admin without clearing the cache.

Why is automatically clearing the cache a problem?

PR #3755, opened 2020-07-18
Support for database caching in core MySQL interface

scottcwilson:
This changes caching to default true for storefront $db->Execute statements which are SELECTs. 

To test, after merging this in, you will need to update your configure files to set the caching to database:

define('SQL_CACHE_METHOD', 'database');

A couple of things we'd want to address: 

a) the lack of caching in
includes/init_includes/init_db_config_read.php 
since this has changed to use Eloquent in 1.5.8, we'll need to write code to do this (as is noted in that file). 

b) Any admin page load clears the cache at the current time.  So if a storeowner was processing orders, they would be continually clearing the cache.  We can optimize this to only clear cache when config updates are done, for example, or perhaps from a link on the dashboard.

### DIFF
--- a/includes/classes/cache.php
+++ b/includes/classes/cache.php
@@ -34,7 +34,7 @@ class cache extends base {
       break;
       case 'database':
       $sql = "select * from " . TABLE_DB_CACHE . " where cache_entry_name = '" . $zp_cache_name . "'";
-      $zp_cache_exists = $db->Execute($sql);
+      $zp_cache_exists = $db->Execute($sql, false, false);
       if ($zp_cache_exists->RecordCount() > 0 && !$this->sql_cache_is_expired($zf_query, $zf_cachetime)) {
         return true;
       } else {
@@ -64,7 +64,7 @@ class cache extends base {
       break;
       case 'database':
       $sql = "select * from " . TABLE_DB_CACHE . " where cache_entry_name = '" . $zp_cache_name ."'";
-      $cache_result = $db->Execute($sql);
+      $cache_result = $db->Execute($sql, false, false);
       if ($cache_result->RecordCount() > 0) {
         $start_time = $cache_result->fields['cache_entry_created'];
         if (time() - $start_time > $zf_cachetime) return true;
@@ -93,7 +93,7 @@ class cache extends base {
       break;
       case 'database':
       $sql = "delete from " . TABLE_DB_CACHE . " where cache_entry_name = '" . $zp_cache_name . "'";
-      $db->Execute($sql);
+      $db->Execute($sql); 
       return true;
       break;
       case 'memory':
@@ -120,7 +120,7 @@ class cache extends base {
       break;
       case 'database':
       $sql = "select * from " . TABLE_DB_CACHE . " where cache_entry_name = '" . $zp_cache_name . "'";
-      $zp_cache_exists = $db->Execute($sql);
+      $zp_cache_exists = $db->Execute($sql, false, false);
       if ($zp_cache_exists->RecordCount() > 0) {
         return true;
       }
@@ -128,7 +128,7 @@ class cache extends base {
       $sql = "insert ignore into " . TABLE_DB_CACHE . " (cache_entry_name, cache_data, cache_entry_created) VALUES (:cachename, :cachedata, unix_timestamp() )";
       $sql = $db->bindVars($sql, ':cachename', $zp_cache_name, 'string');
       $sql = $db->bindVars($sql, ':cachedata', $result_serialize, 'string');
-      $db->Execute($sql);
+      $db->Execute($sql, false, false);
       return true;
       break;
       case 'memory':
@@ -152,7 +152,7 @@ class cache extends base {
       break;
       case 'database':
       $sql = "select * from " . TABLE_DB_CACHE . " where cache_entry_name = '" . $zp_cache_name . "'";
-      $zp_cache_result = $db->Execute($sql);
+      $zp_cache_result = $db->Execute($sql, false, false);
       $zp_result_array = unserialize(base64_decode($zp_cache_result->fields['cache_data']));
       return $zp_result_array;
       break;
@@ -179,19 +179,17 @@ class cache extends base {
         $za_dir->close();
       }
       return true;
-      break;
-      case 'database':
-      $sql = "delete from " . TABLE_DB_CACHE;
-      $db->Execute($sql);
-      return true;
-      break;
+
       case 'memory':
       return true;
-      break;
+
+      case 'database':
       case 'none':
       default:
+      // Do it just in case cache on in catalog but off in admin 
+      $sql = "delete from " . TABLE_DB_CACHE;
+      $db->ExecuteNoCache($sql); 
       return true;
-      break;
     }
   }
 
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -172,7 +172,13 @@ class queryFactory extends base {
     echo '</div>';
   }
 
-  function Execute($zf_sql, $zf_limit = false, $zf_cache = false, $zf_cachetime=0, $remove_from_queryCache = false) {
+  function Execute($zf_sql, $zf_limit = false, $zf_cache = true, $zf_cachetime=3600, $remove_from_queryCache = false) {
+     if (strtoupper(substr($zf_sql,0,6)) != 'SELECT') {
+         $zf_cache = false; 
+     }
+     if (defined('IS_ADMIN_FLAG') && IS_ADMIN_FLAG == 'true') {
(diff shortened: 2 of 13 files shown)


scottcwilson:
Fixes #3403

scottcwilson:
The DISPLAY_PAGE_PARSE_TIME setting was removed in 1.5.7, but if you add back this block at the bottom of includes/templates/responsive_classic/common/tpl_main_page.php below the inclusion of tpl_footer.php:

```
<div class="smallText center">Parse Time: <?php echo $parse_time; ?> - Number of Queries: <?php echo $db->queryCount(); ?> - Query Time: <?php echo $db->queryTime(); ?></div>
</div>
```

You'll see that caching reduces the query time by about 50%.

lat9:
I'm happy to see that the change applies _only_ to the storefront; changing the caching on the admin-side might have disastrous effects.

includes/classes/cache.php
```diff
@@ -93,7 +93,7 @@ function sql_cache_expire_now($zf_query) {
       break;
       case 'database':
       $sql = "delete from " . TABLE_DB_CACHE . " where cache_entry_name = '" . $zp_cache_name . "'";
-      $db->Execute($sql);
+      $db->Execute($sql, false, false);
```
proseLA, line 96: if the false, false was not here, what would happen?  it strikes me as odd that for a deletion from the database, one would need to explicitly state to not use the cache as opposed to the framework handling that logic.

scottcwilson: It's arguably overkill.  If you look at Execute, it disables caching for anything that's not a SELECT.

proseLA: i would say its confusing.  if the vars are going to ignored in the method, why send them?  it confused me.

and why should i need to look at the method?  its just more work for me or any developer/coder/modder (call us what you want).

i love the idea of caching.  i'm just hopeful it does not introduce unwanted errors in all the wrong places.  not everyone uses the zen function to retrieve inventory numbers.  and i support sites that have a number of other inventory buckets, ie in stock, on order, committed, shipped, etc.

we already have this method ```ExecuteNoCache```.  would it not be great if we could simplify?  other classes/functions/scripts  could only call the ```Execute``` or  ```ExecuteNoCache``` with only the sql statement and those functions would then call the class protected function that interacts with the database?   

i know more pain for people upgrading...  perhaps an interim step?

i as thinking about this today, and as people upgrade their ZC sites and their PHP versions, PHP frequently logs warnings about soon to be deprecated methods.  why not start doing that with ZC?  warning, next release of ZC will not accept vars in the Execute method in the db class.  please update....

i know i am asking for too much...  perhaps just some food for thought.  

but if the vars are ignored, please don't send them.  if it's not a ```SELECT``` statement, they are then superfluous, and just confusing to the naked eye.  so don't send them.  MO.

includes/classes/db/mysql/query_factory.php
```diff
@@ -172,7 +172,13 @@ function show_error() {
...
   }
 
-  function Execute($zf_sql, $zf_limit = false, $zf_cache = false, $zf_cachetime=0, $remove_from_queryCache = false) {
+  function Execute($zf_sql, $zf_limit = false, $zf_cache = true, $zf_cachetime=3600, $remove_from_queryCache = false) {
+     if (strtoupper(substr($zf_sql,0,6)) != 'SELECT') {
+         $zf_cache = false; 
+     }
+     if (defined('IS_ADMIN_FLAG') && IS_ADMIN_FLAG == 'true') {
+         $zf_cache = false; 
+     }
```
drbyte, line 181: Given this change, how would an admin-side query be executed WITH cache when desired?

scottcwilson: My suggestion would be to cross that bridge in phase 2.  Perhaps another parameter saying force or something?  I'd like to be sure we get the catalog side implementation right before worrying about admin.  Can you think of an admin case where caching would be desirable, given that there could be two admins working in the system?

scottcwilson:
Added in the template change to show SQL metrics.  This can be removed when testing is complete.

zcwilt:
Still looking at the caching of includes/init_includes/init_db_config_read.php

personally I would prefer to use Laravel's caching sytem as it gives us a lot more options, but 
it needs to be done in a way that doesn't produce any BC breaks;

scottcwilson:
Ugh.  I was hoping to backport this for a client to 1.5.7.  :( 
Using Laravel's db handling seems like a huge change to me.  Are you sure?

drbyte:
@scottcwilson 
I notice several `$db->Execute` were changed to `ExecuteNoCache`. Can you outline the logic used to decide which ones were changed vs which weren't? I think I see some patterns, but would like to understand more of your thinking process so I'm on the same page with you!

drbyte:
For backporting, I think adapting for Laravel stuff would be a separate set of code changes.

zcwilt:
@scottcwilson 

not suggesting that using laravel cache should be in 158, just pointing further forward.

scottcwilson:
I was picking and choosing ones I thought *for sure* you'd want the current live data.  Now I could be wrong!  So please see if I missed one (a much bigger worry than pulling one in unnecessarily).

scottcwilson:
@zcwilt No doubt there are many tasty nuggets in Laravel that we'll want to use. :)

drbyte:
Question: set up this way if the Admin is making changes to products/categories but the catalog is always using only cached data, then the catalog-side won't see the results of those edits until the cache expires, right?

scottcwilson:
No - remember the admin *currently* clears cache on every page load.  (Now I'd like to change this but that's how it is now.)

scottcwilson:
Hope that makes sense.  It's kind of why your earlier post (what about Admin + caching?) is a ways off.  

The short term steps are: 
- Make sure this makes sense.
- Figure out a better way to bust cache from admin (not every page load, because that means if you're active in the admin, customers won't benefit from cache).  Possibility: a button on the dashboard, or a button on the dashboard plus save from categories/products, possibly and configuration, possibly and modules.
Then further refinements as needed.

drbyte:
Considering another angle in discussion with @zcwilt:

Approach: **Only cache the things that beg for caching.** (as opposed to caching everything by default)

Most other platforms tend to operate this way, eg: only cache where beneficial.

Kinda like only adding indexes to the db schema where they are actually helpful and not blowing out storage (and performance) with unnecessary indexes by putting indexes on every field in every table, or otherwise painting everything with the same brush.

**Biggest Gains From Caching**
On the catalog-side, the biggest offenders for "lots of queries" is product lookups, and the category-tree. Next is pricing lookups related to sales/specials/featured in conjunction with the product lookups. And everything with attributes augments the problem, sometimes exponentially.

Next is configuration/layout queries, but those (or at least most) are already cached. These are simple queries that are often already cached by the db engine itself anyway, so even if we're not caching them ourselves they're relatively inexpensive queries already.

At other stages the repeated lookups are related to address calculations. These are small queries, but could be temporarily cached. Long-term caching of city-specific queries isn't beneficial. A 15-min TTL would be plenty.

Fortunately since 1.5.7 we've already handled the product-counter breaking the internal db cache. 
Unfortunately the "products_sold" counter and product-stock counters in the products table also break the internal cache. But only on a completed purchase.

**Problem**
The challenge here is that these product/catalog/pricing queries tend to be scattered all over the code instead of centralized (notwithstanding the current consolidation activities in progress to centralize most), so enabling caching of all the affected places requires touching a lot of code. 

In fact, it's this "queries scattered all over the code base" that I'm sure prompted the approach @scottcwilson proposed here by "caching everything by default", since that gives a "quick fix" with touching as little code as possible.

**Plugins**
Unfortunately as well, a lot of plugins that tinker with pricing have aggressively altered the pricing functions code so any caching we add to those will be lost if people accidentally copy-and-replace pricing functions files from un-updated plugins affecting those files.

**Catalog-Side Areas To Optimize**
- categories sidebox (and plugins that replace the sidebox, since they make tons of queries too)
- product listings
- rarely, if ever, use ExecuteNoCache, because it wouldn't be necessary

**Admin-Side Areas To Optimize**
- explore the best places to clear the cache instead of clearing on every click, which may be unnecessarily affecting catalog-side performance if the admin is only processing orders (eg: not updating products/categories/attributes/pricing)
 -- this may be easily addressed using the consolidated functions used for making inserts/updates/deletes, as the cache changes could be triggered there instead

**Nice-to-Have**
Catalog-side template-rendering caching would potentially give huge performance gains, especially since it could bypass many db queries organically.
Unfortunately this is extremely complicated to do given current architecture, and given that sideboxes and centerboxes expect to display randomized data which isn't amenable to display-caching.

scottcwilson:
The search  

grep "db->Execute" `ff`| grep -i "select "  | wc -l

produces 188 results.  If you do the first 90, I'll do the last 98.

PR #3763, opened 2020-07-19
Resolves issues in delete category

### LINKED ISSUE #2839: Delete of category with no linked product displays linked product text

### LINKED ISSUE #3762: Delete Category bugs
Investigating #2839 reveals a number of other bugs: 

- message TEXT_DELETE_WARNING_CHILDS used but not defined.
- delete logic checks variables $cInfo->products_count and $cInfo->childs_count which do not exist.
- message TEXT_DELETE_WARNING_PRODUCTS used but not defined. 

PR coming shortly.

scottcwilson:
Incorrect / missing messages during delete category. 

Fixes #3762 
Fixes #2839


admin/category_product_listing.php
```diff
@@ -870,26 +870,32 @@
...
+            $products_count = zen_products_in_category_count($cInfo->categories_id);
+            $has_linked = false; 
+            $prod_list = $db->Execute("SELECT * FROM " . TABLE_PRODUCTS_TO_CATEGORIES . " WHERE categories_id = " . $cInfo->categories_id); 
+            foreach ($prod_list as $prod) {
+               $other_cat_list = $db->Execute("SELECT * FROM " . TABLE_PRODUCTS_TO_CATEGORIES . " WHERE products_id = " . $prod['products_id'] . " AND  categories_id != " . $prod['categories_id']);
+               if (!$other_cat_list->EOF) {
+                  $has_linked = true; 
+                  break; 
+               }
+            }
```
drbyte, line 883: Isn't there a function which does this foreach loop already?

scottcwilson: It's for a different category.

drbyte: I was thinking of something like this, so that the queries are kept within existing functions that are already doing these lookups:
```suggestion
            $childs_count = zen_childs_in_category_count($cInfo->categories_id);
            $products_count = zen_products_in_category_count($cInfo->categories_id);
            $has_linked = false; 
            $prod_list = zen_get_linked_products_for_category($cInfo->categories_id);
            foreach ($prod_list as $prod) {
               if (zen_get_linked_categories_for_product($prod, $cInfo->categories_id)) {
                  $has_linked = true; 
                  break; 
               }
            }
```

scottcwilson: OK I see.  I misread your earlier comment - I was thinking you were referring to code higher up in the file (which does indeed do a count but for a different category).  I'll get this done shortly.

scottcwilson: Your code doesn't quite work - second parm to zen_get_linked_categories_for_product() is supposed to be an array.  But the problem is this also uncovers a bug in zen_get_linked_categories_for_product().  I'm going to put both in this update, please split if you wish.

drbyte: Good catch. Thanks.

PR #3766, opened 2020-07-19
Prevent Product Qty Maximum instructions from wrapping

scottcwilson:


drbyte:
For things like this could you start providing before-and-after screenshots?

Also, these changes should be applied to all product-types.

scottcwilson:
How about we get rid of the other product types so we don't have to keep maintaining code that almost no one uses?

wiztechinc:
> How about we get rid of the other product types so we don't have to keep maintaining code that almost no one uses?

I’m game, even if you are joking!

drbyte:
> How about we get rid of the other product types so we don't have to keep maintaining code that almost no one uses?

Eventually.

Until then it's best to keep them in sync.

Let's not hijack the topic at hand.

PR #3767, opened 2020-07-19
Permit ampersand in STORE_NAME (already allowed in STORE_NAME_ADDRESS)

scottcwilson:


drbyte:
Unlike the address, the store name value is output in dozens of pages and emails. 
Does this still prevent all malicious actions that could happen in emails?

PR #3769, opened 2020-07-20
$function incorrectly provided for $order_id

mc12345678:
In the refactoring of the payment module, the determined `$function` was incorrectly passed to the associated function instead of the data that was passed into the parent function (`$zf_order_id`) causing issues as identified in: https://www.zen-cart.com/showthread.php?226988-parameter-zf_order_id-is-not-being-transferred-into-the-after_order_create-method.

### DIFF
--- a/includes/classes/payment.php
+++ b/includes/classes/payment.php
@@ -267,7 +267,7 @@ class payment extends base {
     if (!$GLOBALS[$this->selected_module]->enabled) return;
     $function = __FUNCTION__;
     if (!method_exists($GLOBALS[$this->selected_module], $function)) return;
-    return $GLOBALS[$this->selected_module]->after_order_create($function);
+    return $GLOBALS[$this->selected_module]->after_order_create($zf_order_id);
   }
 
   function admin_notification($zf_order_id) {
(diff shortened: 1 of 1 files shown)


mc12345678:
Fixes #3770

includes/classes/payment.php
```diff
@@ -267,7 +267,7 @@ function after_order_create($zf_order_id) {
     if (!$GLOBALS[$this->selected_module]->enabled) return;
     $function = __FUNCTION__;
     if (!method_exists($GLOBALS[$this->selected_module], $function)) return;
-    return $GLOBALS[$this->selected_module]->after_order_create($function);
+    return $GLOBALS[$this->selected_module]->after_order_create($zf_order_id);
```
lat9, line 270: In the spirit of @DrByte's original change, shouldn't this and the following change read
```php
   return $GLOBALS['this->selected_module]->$function($zf_order_id);
```

mc12345678: Was just looking at that as I was closing tabs... Yes... Making second commit.

drbyte:
Thanks!

PR #3773, opened 2020-07-20
Safety code when zc_plugins directory has been removed or never existed

zcwilt:
Currently code would die with a fatal error
Move inspectAndUpdate to service provider.
if zc_plugins doesn't exist the plugin manager page will throw a warning
leaving for now as I have a refactor branch that will address


drbyte:
Backporting would require a different approach (eg: can't cherrypick) given that the `AppServiceProvider` doesn't exist in prior versions.

zcwilt:
yeah, working on backport code now.

PR #3775, opened 2020-07-21
Bump minimum PHP version to 7.3 for ZC v1.5.8

drbyte:
Given that the [previous 3 Zen Cart versions already provide support for up to and including PHP 7.4](https://docs.zen-cart.com/user/first_steps/server_requirements/)
and that [PHP 7.2 is already end-of-life and will be end-of-support by November this year](https://www.php.net/supported-versions.php),
it makes sense that the next Zen Cart release be able to be used on modern PHP versions
and make use of resources that leverage the improved functionality and speed performance of
newer PHP versions.
This will also mean that storeowners who are running on servers that actively make security a priority
will be able to keep using Zen Cart into the future with less compatibility concerns instead of being
held back onto older PHP versions if ZC doesn't support them.

Further, storeowners "stuck" on old servers using entirely obsolete PHP versions
have the choice to upgrade to any of the Zen Cart versions supporting whatever PHP version they're stuck with,
and also to upgrade to whatever alternate PHP versions their server company can offer,
due to the breadth of support of PHP versions already offered by currently-available Zen Cart releases.


zcwilt:
I'm pretty sure we covered off deprecations and BC breaks for 7.0->7.4 in previous releases, but will revisit.
Have been using php 7.4 since release on 1.5.7 and 1.5.8 with no issues.

PR #3779, opened 2020-07-23
remove model from popover when empty

proseLA:
empty parenthesis in order details popover ```()``` were a little confusing.  not everyone has models for all products.

Closes #3781


drbyte:
Does the same apply to the dashboard widget?

proseLA:
it does.

can we move this to the orders class and refactor?

drbyte:
> can we move this to the orders class and refactor?

I was exploring that when first implementing the popover, but the required refactoring for the orders page was massive.
So, yes, we're thinking similarly. 

But that major refactoring is way beyond the tiny scope of this change. I'd treat them separately.

proseLA:
the orders page is massive.  i agree.  i have been in upgrade purgatory as we speak...

adding a new method for the order class, adding the class to the widget, and calling that method in both places, does not seem as massive as redoing the whole orders page.

i'm proposing a minor refactor.

i would prefer to do it that way, if you will entertain/indulge me.  if i run into major headaches, then i will go back to your original suggestion and change it in the widget.

does that work for you?

drbyte:
Go ahead and try it. Do it as a separate branch from this one, so we can decide which one to use when the time suits.

The problem is that the admin orders page doesn't use the order class. Instead it has a hundred (exaggerated)  bespoke queries for each separate data piece it loads. In some ways these are more performant because it avoids loading unnecessary queries where data is not needed. But it's a tougher maintenance situation.

The popover in the orders page did a minor augmenting of the queried data consistent with what was already being done, since it added very little additional overhead. The widget on the other hand invokes the entire order class (IIRC) to populate the data it uses. Doing the same in the admin orders page is a major overhaul. 

I'm open to that overhaul happening in 1.5.8 but it should really be canonical, ie: separate from any feature-changes, in order for code-review and future troubleshooting of any issues to be easily explored.
Do each smaller rearrangement in separate commits to the branch so the changes are easily code-reviewed. Otherwise I'm going to have a headache trying to figure out your logic behind everything you change: I've got to understand and agree with each step as part of the review, else I can't be sure each change is "safe".

Be mindful that function-consolidation is ongoing and may involve moving some functions into the order class. I apologize in advance if these activities step on anything you're doing. You may have some merge-conflicts to deal with along the way.

PR #3781, opened 2020-07-23
consolidate order products popover into order class

proseLA:
i will not comment on the performant aspect.  i do like it more for eliminate duplicate code.


proseLA:
can i ask the reasoning for the use of the other PR as opposed to this one?

drbyte:
In the end, the code is specialized to output logic, and not really relevant to an Order as an Object.
To be fair, it doesn't generate much HTML, but it's very specific to "the widget", which isn't really the ideal sort of responsibility to embed into an Object.

Eventually another sort of abstraction may be practical, after some additional order class refactorings where the data being queried for this widget may be something that can be requested on demand. Or maybe not even necessary at all (eg the orders page could use the object methods for all of its queries (ideally in one response, not multiple)  instead of making dozens of bespoke queries).

Issue #3729, opened 2020-07-13
product search (admin) fails...


proseLA:
https://github.com/zencart/zencart/blob/3acda696b5ff7b8a2ec7d1aad3b7b25c55dc44bf/admin/category_product_listing.php#L689

i have never understood why the master category is required as part of search.

a user duplicated a product to a new category, and as part of the duplication, ZC did not add a record into the products_to_category table for the master category of the product.  now search fails.

pretty easy to duplicate.  see:

https://www.zen-cart.com/showthread.php?226951-Can-t-search-for-duplicate-after-saving-(Numinix-custom-fields)&p=1370670#post1370670

(Numinix has nothing to do with it.  i was able to reproduce on a vanilla v157 install.)

scottcwilson:
> ... ZC did not add a record into the products_to_category table for the master category of the product. 
to me, that's the fix that's needed. (Or at least part of it.) 

Not sure about the SQL change, but creating a P2C record seems like an obvious need.

proseLA:
> Not sure about the SQL change, but creating a P2C record seems like an obvious need.

the more i think about it, the more i think we do need a SQL change.

why are we preventing store admins from finding their product?  because the data is messed up?  is that a reason to not allow admins to find product?

i am all for fixing the creation of a P2C record, but when store owners create new product types and then code their own copy product module, is it not easy for them to miss the code that creates the P2C for the master category?  what about the data that already exists out in the ZC ecosphere?

in addition, with the exception of active status, why is the search on the storefront different from the search on the admin?  on the storefront one can search for `bug collect` and find item 34; but on the admin we find nothing. 

now, to address the multi word issue, we could do something like i did for customer search here:

https://github.com/zencart/zencart/pull/3079

but i think it would behoove us to have 1 consistent place where we do product search.  perhaps it is a lot of work with refactoring and the like, but it makes sense to me.

finally, as someone who has run across this master category problem before (and addressed it on clients sites), can someone give me a reason why we keep the sql matching the master category in?  if the item's master category does not have a record in the products to categories table, we do not allow admins to find it. why?

torvista:
In my test, duplicating a product from catA to catB:

new product has master category of catA as it uses the same as source product - incorrectly.

new product has p2c category of catB - correct, as per product_copy_confirm.php:
```
        $db->Execute("INSERT INTO " . TABLE_PRODUCTS_TO_CATEGORIES . " (products_id, categories_id)
[...]

PR #3787, opened 2020-07-25
While admin product search ignores master cat, product listing needs it

drbyte:
p2c is still needed in order to show cross-linked categories

Updates #3786
Updates fix to #3729

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -676,6 +676,9 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                               OR p.products_model LIKE '%:search%'
                             ) ";
                 $where = $db->bindVars($where, ':search', $_GET['search'], 'noquotestring');
+            } else {
+                $products_query_raw.= " LEFT JOIN " . TABLE_PRODUCTS_TO_CATEGORIES . " p2c USING (products_id) ";
+                $where .= " AND p2c.categories_id=" . (int)$current_category_id;
             }
 
             $products_query_raw .= $where . $order_by;


scottcwilson:
Not sure if we could further tweak this to solve the following issue: 
- product has master categories id set in products table, corresponding entry exists in the P2C table - all is well.
- P2C entry gets clobbered
- product will no longer show in Admin > Catalog > Categories/Products > [product master category]

scottcwilson:
btw when this happens, catalog search also fails for this product. The only way to get to it is to go to the product_info page with the product id.

torvista:
How about a check in the header of the admin category listing but only for the current category with products (to minimise overhead), for missing mcats?

If I can't find a product by search, I would navigate to its category manually...and then would get the message about the missing mcat/or a message reporting it having been auto-fixed.

drbyte:
> How about a check in the header of the admin category listing but only for the current category with products (to minimise overhead), for missing mcats?

@torvista You've mentioned a couple times that you're doing this "audit" on your own site, but never shared the code you're using to do it. Mind sharing? In whatever form it's in now ... doesn't have to be massaged in any way ... seeing it would add context to your point.

drbyte:
> btw when this happens, catalog search also fails for this product. The only way to get to it is to go to the product_info page with the product id.

I thought in prior discussion that preceded this set of commits, that it was said that catalog search is fine. Are you saying that (if p2c doesn't have master cat then) catalog search is "always broken" or just certain kinds of searches?

scottcwilson:
If no record for the product id exists in the p2c table, it breaks search.

drbyte:
> Not sure if we could further tweak this to solve the following issue:
> 
> * product has master categories id set in products table, corresponding entry exists in the P2C table - all is well.
> * P2C entry gets clobbered
> * product will no longer show in Admin > Catalog > Categories/Products > [product master category]

To clarify: what you're reporting is not "new", right? It's always operated like this, right? eg: the symptom you describe is not a regression/bug created by new code changes, right?

scottcwilson:
Yes - this is a long standing issue.  I just thought you had fixed it because you were working in the area.

drbyte:
In the product-listing then, instead of it "missing", should we show the product, along with another column (yes, this makes the output even wider than it is now! ... alert! for tiny screens!) showing that there's an error, alerting that it needs fixing?

scottcwilson:
Steve, maybe you & I should open issues separately about our concerns - this is getting quite hard for me to track.

torvista:
My code for auditing mcats....under review...

Issue #3789, opened 2020-07-26
New boolean define PRODUCT_THROWS_200_WHEN_DISABLED does not use quoted values


scottcwilson:
References to this value in the following files: 

./includes/functions/functions_products.php
./includes/modules/pages/product_free_shipping_info/main_template_vars.php
./includes/modules/pages/document_general_info/main_template_vars.php
./includes/modules/pages/document_product_info/main_template_vars.php
./includes/modules/pages/product_music_info/main_template_vars.php
./includes/modules/pages/product_info/main_template_vars.php

compare to `true` rather than `"true"` as is traditionally done for other defined constants.

This could be confusing.

drbyte:
Use of boolean was intentional for this. It's not a configuration switch (which have heretofore required strings). It's an optional and technically undocumented override one could add.

Using boolean words as strings is very often misinterpreted. Boolean is very clear.

Aside: in hindsight I think the use of a constant here was ill-conceived. I wish I'd leaned towards a notifier hook instead, so that the logic could be product-specific.

scottcwilson:
Very good - closing then, but please document with an example of how to configure this.  (Or just make a note here and I'll add it to the documentation.)

drbyte:
Update: Given a notifier event does already fire, I'm inclined to leave this constant undocumented, in favor of using the observer instead.

scottcwilson:
Isn't an observer kind of storeowner unfriendly?  What about a flag instead in the Admin > Configuration > Stock area? 

If you're set on using the observer, can you add a quick note to (say) /user/products/products_misc.md telling people how to enable this behavior?  It's completely undocumented right now.  I just stumbled on it trying to answer Steve's question.

mc12345678:
The way I have read the code that uses/supports `PRODUCT_THROWS_200_WHEN_DISABLED` is that its use must be very purposefully implemented... If just the constant is defined, then a "disabled" product becomes visible and a 200 is thrown. To have the disabled product "hidden" then requires addition action in advance of the code at: https://github.com/zencart/zencart/blob/cff05fe30df4938e9053f5d23690bb4c975a3a80/includes/modules/pages/product_info/main_template_vars.php#L21

[...]

PR #3790, opened 2020-07-26
PRODUCT_THROWS_200_WHEN_DISABLED check condition is backwards

scottcwilson:
Logic should be: 

```
IF throw_200 is true THEN
   IF disabled THEN 
       SET not found 
   FI
FI 
```

### DIFF
--- a/includes/modules/pages/product_info/main_template_vars.php
+++ b/includes/modules/pages/product_info/main_template_vars.php
@@ -20,7 +20,7 @@
 
   $product_not_found = $product_info->EOF;
 
-  if (!defined('PRODUCT_THROWS_200_WHEN_DISABLED') || PRODUCT_THROWS_200_WHEN_DISABLED !== true) {
+  if (!defined('PRODUCT_THROWS_200_WHEN_DISABLED') || PRODUCT_THROWS_200_WHEN_DISABLED == true) {
       if (!$product_not_found && $product_info->fields['products_status'] != 1) {
           $product_not_found = true;
       }


scottcwilson:
This can be easily tested simply by setting a product to disabled and adding 

`define('PRODUCT_THROWS_200_WHEN_DISABLED',false);`

to the configure.php, and then attempting to load the product_info page for the disabled product.  

You will observe that it does not work until this fix is applied.

drbyte:
ALERT: changing this is will break whatever people have already implemented from the forum discussions that inspired these changes in the first place. Need to locate those discussions and discern why they were satisfied with the current implementation if indeed it is wrong.

drbyte:
Hmmm ... While the code was heavily discussed on github, I was fairly certain there was a forum thread specific to this feature, where we tested its implementation.
Maybe I was only dreaming.

scottcwilson:
If you can document how to make the current logic work I'd be interested.  It sure seems backwards to me.   I was one of the people who suggested such a flag but I don't recall any specific implementation discussions.

drbyte:
Given there's no forum implementation, and that there's a notifier `NOTIFY_PRODUCT_INFO_PRODUCT_STATUS_CHECK` for it which provides all the same power, and more, maybe it's best to just remove that boolean constant check.

scottcwilson:
As I said in #3789 this seems storeowner unfriendly - can we just do a config setting?

drbyte:
Closed by #4037

PR #3792, opened 2020-07-27
Correct tpl_columnar_display.php comparison

lat9:
- is_array returns a boolean, not suitable for a > comparison.


drbyte:
Good point.

I'm inclined to also refactor it to use a `foreach` instead of `for`, which makes following the bouncing ball easier when looking at the code.

PR #3802, opened 2020-07-30
Set the samesite cookie parameter

### LINKED ISSUE #3721: setting samesite cookie attribute
https://github.com/zencart/zencart/blob/e60c81b6dbac7b1a61fb891397fe20de7b02c623/includes/init_includes/init_sessions.php#L44

we do not seem to be setting the samesite cookie attribute.  perhaps we should?  not sure if we need a new config value (:confounded:) or not....

in addition, we would have to account for the php version, as 7.3 allows for an associative array for use:

```
    $samesite = CONFIG_SAMESITE;  // Lax or Strict or None

    if(PHP_VERSION_ID < 70300) {
        session_set_cookie_params(0, $path .'; samesite='.$samesite, (zen_not_null($cookieDomain) ? $domainPrefix . $cookieDomain : ''), $secureFlag, TRUE);
    } else {
        session_set_cookie_params([
            'lifetime' => 0,
            'path' => $path,
            'domain' => (zen_not_null($cookieDomain) ? $domainPrefix . $cookieDomain : ''),
            'secure' => $secureFlag,
            'httponly' => TRUE,
            'samesite' => $samesite
        ]);
    }
```

or something like this.

drbyte:
The way Zen Cart operates this will almost always need to be set to `lax`, so no config switch is being added for it at this time.

Fixes #3721

### DIFF
--- a/admin/includes/init_includes/init_sessions.php
+++ b/admin/includes/init_includes/init_sessions.php
@@ -4,7 +4,7 @@
  * @copyright Copyright 2003-2016 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: Author: DrByte  Sun Jan 10 02:53:32 2016 -0500 Modified in v1.5.5 $
+ * @version $Id: Author: DrByte  Sun Jan 10 02:53:32 2016 -0500 Modified in v1.5.8 $
  */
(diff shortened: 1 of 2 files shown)


drbyte:
Related:
https://www.youtube.com/watch?v=Fet6-IiX69E&list=PLNYkxOF6rcIDJHOcBzho38p6WTn3vESvQ
https://web.dev/samesite-cookies-explained/
https://www.chromium.org/updates/same-site/test-debug

PR #3804, opened 2020-07-31
improve admin product search

proseLA:
the store side has a better product search than the admin side.  similar to the work done on the admin customer search, this makes much more sense to me, and makes it easier for admins to find products.

it would be nice to more mimic the store side but the store side function ```zen_parse_search_string``` is not available.

```$keywords``` is assigned well above this line, and i do not see it in use in this script.

finally, it might behoove us, considering the plethora of functions inherent in ZC, to consider refactoring this logic into a function; the function takes 2 arrays as inputs: column names and keywords; and returns a find string for use in all of these areas.

i see this var ```$keywords``` used in the following scripts, and somehow having a function that creates the where clause seems like a good idea.

```
orders.php
specials.php
category_product_listing.php
reviews.php
products_price_manager.php
downloads_manager.php
attributes_controller.php
customers.php
featured.php
```

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -670,12 +670,16 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
             $where = " WHERE pd.language_id = " . (int)$_SESSION['languages_id'];
 
             if ($search_result && $action != 'edit_category') {
-                $where .= "  AND (pd.products_name LIKE '%:search%'
-                              OR pd.products_description LIKE '%:search%'
-                              OR p.products_id = ':search'
-                              OR p.products_model LIKE '%:search%'
+                $parts = explode(" ", trim($keywords));
+                foreach ($parts as $k => $v) {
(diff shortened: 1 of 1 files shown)


scottcwilson:
> the store side has a better product search than the admin side. 
Can you share some queries that work on the store side but not the admin side so we're clear on the store-side advantage?

proseLA:
https://zc157.com/index.php?main_page=advanced_search_result&search_in_description=1&keyword=cut+nke

https://zc157.com/admin/index.php?cmd=category_product_listing&search=cut+nke

or try on the 157 test data, search for ```blade link``` on the store side and the admin side.

scottcwilson:
Note that the function `zen_parse_search_string` is now available in the admin thanks to the reorganization and de-duplication of the `functions` folders.   So this could be used in any planned admin-side refactoring.

proseLA:
this is available in v1.5.8 going forward?

scottcwilson:
Yes. See admin/includes/init_includes/init_general_funcs.php

PR #3806, opened 2020-07-31
Remove LIMIT from also purchased query

### LINKED ISSUE #3805: bug in also purchased products.
https://github.com/zencart/zencart/blob/89621e6e8cc96dfdaa202655026737174195f168/includes/modules/also_purchased_products.php#L15

see:
https://www.zen-cart.com/showthread.php?227029-Customers-who-bought-this-product-also-purchased-doesn-t-show-any-products

i have confirmed.  but i do not have the solution at this hour.

scottcwilson:
Fixes #3805 - error in removing hardcoded limit.


scottcwilson:
Suggested for backport to 1.5.7.

PR #3819, opened 2020-08-02
Corrects missing/disabled categories result in noindex,nofollow

lat9:
... as identified in [this](https://www.zen-cart.com/showthread.php?227008-Deleted-category-results-in-noindex-nofollow) Zen Cart bug report.

Suggest back-port to zc157; the base module had no change in the transition to zc158.

### DIFF
--- a/includes/modules/pages/index/header_php.php
+++ b/includes/modules/pages/index/header_php.php
@@ -61,7 +61,7 @@ if (isset($cPath) && zen_not_null($cPath)) {
     //
     $category_redirect_handled = false;
     $zco_notifier->notify(
-        'NOTIFY_HEADER_INDEX_HTTP_STATUS_CHECK', 
+        'NOTIFY_INDEX_CATEGORY_STATUS_CHECK', 
         array('cPath' => $cPath, 'current_category_id' => $current_category_id),
         $category_redirect_handled,
         $current_category_not_found,
@@ -84,8 +84,8 @@ if (isset($cPath) && zen_not_null($cPath)) {
     //    - Set the flag to cause noindex/nofollow to be included for the page
     //    - Issue a 410 (Gone).
     //
-    //    Note: A pseudo-configuration setting is used to allow stores to operate as in Zen Cart versions
-    //    prior to v157a, where a disabled category is still displayed.
+    //    Note: Stores that wish to to operate as in Zen Cart versions prior to v157a, where a disabled 
+    //    category is still displayed can comment the above section 'out'.
     //
     // 3. Otherwise, the category is present and not disabled. Determine the category 'depth' to be displayed.
     //    a. If the current category contains at least one product, display a products' listing.
@@ -103,10 +103,12 @@ if (isset($cPath) && zen_not_null($cPath)) {
             $breadcrumb->reset();
             $robotsNoIndex = true;
             header('HTTP/1.1 404 Not Found');
-        } elseif ($current_category_is_disabled && (!defined('DISPLAY_DISABLED_CATEGORIES') || DISPLAY_DISABLED_CATEGORIES != 'true')) {
+//-bof-Comment the following four (4) lines out to display disabled categories
+        } elseif ($current_category_is_disabled) {
             $category_depth = 'products';
             $robotsNoIndex = true;
             header('HTTP/1.1 410 Gone');
+//-eof-Comment the above four (4) lines out to display disabled categories
         } elseif ($current_category_has_products) {
             $category_depth = 'products';
         } else {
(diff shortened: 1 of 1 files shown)


includes/modules/pages/index/header_php.php
```diff
@@ -13,40 +13,118 @@
...
+    //      of display for stores that have an 'invalid' mix of products and categories within a
+    //      category.
+    //
+    if (!$category_redirect_handled) {
+        if ($current_category_not_found) {
+            unset($_GET['cPath']);
+            $breadcrumb->reset();
+            $robotsNoIndex = true;
+            header('HTTP/1.1 404 Not Found');
+        } elseif ($current_category_is_disabled && (!defined('DISPLAY_DISABLED_CATEGORIES') || DISPLAY_DISABLED_CATEGORIES != 'true')) {
```
drbyte, line 106: Do you have any strong objection to removing this "pseudo-configuration setting"?

I'd prefer that people just use an observer if they want this sort of thing, particularly since it also provides infinitely more flexibility.

(I realize it was inspired by a similar setting for products .. which I'm very close to removing for similar reasons, particularly since it's not actually even working right now.)

lat9: Depends on what you consider `strong`.  It seems like having some quick-and-easy way for stores to revert to previous behavior is a good thing.

drbyte: I'm not sure much fallback-to-old is necessary here.
Here's my reasoning:
- in either case it's going to cascade into `$category_depth = 'products'` as before
- the `$robotsNoIndex` isn't a destructive thing, so no harm in it being triggered without writing code to force old approach
- the `410` response is indeed the correct response for the situation, and will fine-tune SERP listings by the cleanup it offers, resulting in a better customer shopping experience. The notifier allows store-specific specialized handling if something more sophisticated is required.

includes/modules/pages/index/header_php.php
```diff
@@ -13,31 +13,104 @@
...
-            $category_depth = 'nested'; // navigate through the categories
+        $current_category_has_subcats = !$category_parent->EOF;
+    }
+    
+    // -----
+    // Give an observer the chance to override the default handling for the category.
+    //
+    $category_redirect_handled = false;
+    $zco_notifier->notify(
+        'NOTIFY_HEADER_INDEX_HTTP_STATUS_CHECK',
```
drbyte, line 64: Do you mind having a discussion about naming of this?
`NOTIFY_HEADER_INDEX_HTTP_STATUS_CHECK`

Given the location in code where it fires, the name makes perfect sense. That would be my initial first thought too, especially given the product-related pattern this code was inspired from.

But given its very specialized purpose and that all the data passed to it is strictly category-related, are `INDEX` and `HTTP_STATUS_CHECK` both good namings here?

lat9: What do you have in mind?  Something more like `NOTIFY_INDEX_CATEGORY_MISSING_DISABLED_CHECK`?

includes/modules/pages/index/header_php.php
lat9: I don't quite understand.  Without that final bit, you'd never get to a `nested` sub-category listing.

drbyte: > you'd never get to a nested sub-category listing

Argh. I guess. 

I really don't like adding more constants. At least this one's not another admin configuration switch. 

I also don't like giving constants a string value of `'true'` or `'false'` when those words should only be booleans. (Notwithstanding the horrible precent in existing code, inherited from osC.) Better would be `'yes'` or `'no'` because there's no ambiguity or accidental misuse.
Better still is not to implement them at all. 

> It seems like having some quick-and-easy way for stores to revert to previous behavior is a good thing.

I'm fully compassionate to storeowners' resistance to change, and their varied "requirements"

... but do you think there's a very strong likelihood that droves of stores will scream for previous behavior?
If so, why are we even changing things?

Trying here to avoid adding technical debt of code that can never be removed for fear of breaking something one random store might be using.

I'd kinda rather tell the few people who need to "revert to previous" how to comment-out 3 lines of code for their specific use-case, instead of offering a switch that we're obligated to maintain for forever.

includes/modules/pages/index/header_php.php
drbyte: > What do you have in mind? Something more like `NOTIFY_INDEX_CATEGORY_MISSING_DISABLED_CHECK`?

Ya, something along that line feels context-relevant. 
Or maybe `NOTIFY_INDEX_CATEGORY_STATUS_CHECK` because it's a bit shorter.

lat9: Updated in that update just pushed.

includes/modules/pages/index/header_php.php
lat9: Done, review requested.

drbyte:
Thanks.
Feel free to post a PR for this to the 157 branch to have it included in the 157a release.

PR #3825, opened 2020-08-03
Code referenced non-existent field

scottcwilson:
I simplified the message instead of printing the array.


drbyte:
Hmm ... ya, the previous refactor should have used `count($check_products_top_categories)` instead of `$check_products_top_categories->fields['products_in_top_level_error']`

Will people find your proposed less-descriptive message equally helpful?

scottcwilson:
If it's a count, then let me restore it - I thought it was intended to list the products.

PR #3831, opened 2020-08-04
V158 upgrade to laravel 7

zcwilt:
This upgrades Laravel from 5.8 to 7.2 

The move to use a new directory. app/ to laravel/app was prompted by the fact that Laravel only expects some of it's directories
to be in the app directory, with others one level above,
The original use of just the app directory meant we had to make some internal hacks to paths and namespaces to get Laravel working, and while I didn't see any errors because of that. The possibility remains that  some configurations, especially on Windows may well have,

I've also taken the opportunity to now completely disable Laravel's internal error reporting/Logging as while I do prefer the 500 error pages and the graphical stack traces I know this was causing issues for some people.

In the future we can add back the Laravel logging/reporting, but in a way that benefits all.


drbyte:
NOTE: if you had an `/app` from 158 before this, you can just delete that dir.

PR #3834, opened 2020-08-04
Remove PRODUCT_THROWS_200_WHEN_DISABLED

drbyte:
(It was introduced in 1.5.7 but didn't fully work. In retrospect was probably a bad idea because it forces supporting something into future that should be left behind. Thus removing in 1.5.7a.)

Also removes duplicate logic since the http response is already set in `header_php.php`

Closes #3790 
Ref #3789

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -62,10 +62,6 @@ function zen_product_set_header_response($product_id, $product_info = null)
         $response_code = 410;
     }
 
-    if (defined('PRODUCT_THROWS_200_WHEN_DISABLED') && PRODUCT_THROWS_200_WHEN_DISABLED === true) {
-        $response_code = 200;
-    }
-
     if ($product_status === -1) {
         $response_code = 410;
     }
@@ -85,7 +81,7 @@ function zen_product_set_header_response($product_id, $product_info = null)
     if ($should_throw_404) {
         // if specified product_id doesn't exist, ensure that metatags and breadcrumbs don't share bad data or inappropriate information
         unset($_GET['products_id']);
-        unset($breadcrumb->_trail[sizeof($breadcrumb->_trail)-1]['title']);
+        unset($breadcrumb->_trail[count($breadcrumb->_trail)-1]['title']);
         $robotsNoIndex = true;
         header('HTTP/1.1 404 Not Found');
         return;
--- a/includes/modules/pages/document_general_info/main_template_vars.php
+++ b/includes/modules/pages/document_general_info/main_template_vars.php
@@ -20,12 +20,6 @@
 
   $product_not_found = $product_info->EOF;
 
-  if (!defined('PRODUCT_THROWS_200_WHEN_DISABLED') || PRODUCT_THROWS_200_WHEN_DISABLED !== true) {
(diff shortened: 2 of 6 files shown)


scottcwilson:
Could we document (with some sample code) the approved way to return 200 on a disabled product?  This is an issue because Google complains bitterly about getting a 404 and it freaks clients out.

mc12345678:
I thought that part of the desire was to offer multiple header responses as well as visualization of the product page.

It seems that if the variable is removed from main_template_vars then the condition previously expressed as a concern where this constant was set to false and the product "disabled" was that the product was displayed and a 410 was thrown. In the same vein if the constant was set to true and the product disabled then the product would be displayed but a 200 would be thrown. If the constant was set to anything but true or false or not present then the disabled product would be hidden and throw a 410. To throw a 200 but have the product hidden would require additional effort. 

I'm not so sure it "didn't" work as much as the desired outcome to be known and appropriate action/code implemented.

BTW, I'm posting this now and may either revise or revisit, but it appears from just this change alone that a disabled product that exists in the database will still be shown instead of being identified as not existing. Begs the question (again there is another PR proposed related to this constant that I haven't looked at yet) of how one can have a product in the database but it not be presented to the public in a default install of ZC.

scottcwilson:
I'd love to see a plugin that handles this if someone has time to build one.

proseLA:
i like this config.  i think you are trying to do too much with it.  what does this switch have to do with determining the products status?

example:
store sell product A.  A becomes very popular....  gets indexed by various search engines.  response code 200.

store sells out of product A.  product A is now disabled.  but it still has a very high ranking on search engines.   due to high indexing, new customers come to store via search engine link to disabled product A.  do you now want to give that page a 404?  and lose your ranking?  i say NO!  

i have modified all of my clients sites to throw 200s (using this config) for all disabled products.  once on the site, disabled products can NOT be added to the cart, nor do they come up with in search results from the ZC site.

i think this config is properly named; whether or not it is implemented correctly is a different story.  it should only be used for the response code.  there are plenty of other config switches as to whether a store wants to display a disabled product (the only time that this config should be used, and only after the store determines they DO want to display disabled products), and/or add it to the cart.

some discussion here:

https://www.zen-cart.com/showthread.php?227129-Google-says-I-have-Redirect-errors-but-I-cannot-find-them

scottcwilson:
> i have modified all of my clients sites to throw 200s (using this config) for all disabled products. once on the site, disabled products can NOT be added to the cart, nor do they come up with in search results from the ZC site.

This is pretty much my desired end state too - we might quibble about the search results, but it captures the spirit of the change.

scottcwilson:
Expanding products.products_status to 2 bits (or more) so it could capture an additional value, like "hidden" might also be worth doing in the context of this change.

lat9:
> 
> 
> Expanding products.products_status to 2 bits (or more) so it could capture an additional value, like "hidden" might also be worth doing in the context of this change.

Er, a `tinyint` is already 8 bits (one byte) and can range from (signed) -128 to +127 or (unsigned) 0 to 255.

scottcwilson:
ah right you are - for some reason I thought it was a bit rather than a byte.

drbyte:
Closed by #4037

PR #3839, opened 2020-08-05
undefined variable $tax

scottcwilson:
Note: looks as though ot_coupon has the same issue (l. 106) but wanted to get this one in the pipeline first.


drbyte:
If two modules have the same issue, was there a refactoring that caused both to become broken in the same way?

scottcwilson:
No this issue has been there for a long time.  It's in 1.5.1.

drbyte:
Oy. The next question then is what "breaks" if this gets "fixed" - LOL

PR #3843, opened 2020-08-06
Fix log created by missing space before ORDER BY

scottcwilson:


scottcwilson:
Log occurs when admin > catalog > option name manager is viewed (likely other pages too).

PR #3844, opened 2020-08-06
Fix syntax error in language file

### LINKED ISSUE #3842: lang.developers_tool_kit, PHP warning
Seeing the following with the file-set downloaded this morning:
image

scottcwilson:
Fixes #3842


scottcwilson:
To test, go to Admin > Tools > Developer's Tool Kit.   Before the fix, you will see a log as noted by Cindy.  After the fix, no log.

drbyte:
While the fix itself is correct, I'm leaving this for @zcwilt because it may indicate a need to update the converter script.

drbyte:
Or maybe this doesn't need to mention the specific language at all ...... cuz the only option is the current language, which is self-evident because the UI is displaying in that same language.

scottcwilson:
I agree but I was striving to preserve the designer's intent.  We'll see what Ian says.

zcwilt:
I do have some updates to the script that generates the lang. files, but only in the sense that it will flag when it finds content
it can't understand. Haven't tested yet.

However on a general note, would prefer over time to remove stuff like this and rely more on sprintf for interjecting 
this type of content into language strings.

Apart from that have no problems with this pr for now.

PR #3848, opened 2020-08-07
Fixes #3805 - also purchased products broken

### LINKED ISSUE #3805: bug in also purchased products.

scottcwilson:
1.5.7 backport of #3806

### DIFF
--- a/includes/classes/db/mysql/define_queries.php
+++ b/includes/classes/db/mysql/define_queries.php
@@ -29,6 +29,5 @@ DEFINE('SQL_ALSO_PURCHASED', "SELECT p.products_id, p.products_image, max(o.date
                      AND opb.orders_id = o.orders_id
                      AND p.products_status = 1
                      GROUP BY p.products_id, p.products_image
-                     ORDER BY date_purchased desc, p.products_id
-                     LIMIT %u");
+                     ORDER BY date_purchased desc, p.products_id"); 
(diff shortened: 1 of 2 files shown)


includes/classes/db/mysql/define_queries.php
```diff
@@ -29,6 +29,5 @@
                      AND opb.orders_id = o.orders_id
                      AND p.products_status = 1
                      GROUP BY p.products_id, p.products_image
-                     ORDER BY date_purchased desc, p.products_id
-                     LIMIT %u");
```
lat9, line 33: Why did the `LIMIT %u` get removed?  It seems that now **_all_** also-purchases products are going to be pulled with this query.

scottcwilson: All are pulled and then MAX_DISPLAY_ALSO_PURCHASED are selected from the set.  See `includes/modules/also_purchased_products.php` line 15.

drbyte: A better way to describe the change is that the LIMIT has been relocated from the DEFINE SQL to a parameter on the `$db->ExecuteRandomMulti` function call.

Technically the result is the same.

PR #3850, opened 2020-08-08
Consolidate customer-related functions into a Customer class

drbyte:


admin/includes/functions/general.php
```diff
@@ -2575,21 +2566,6 @@ function zen_geo_zones_pull_down_coupon($parameters, $selected = '') {
...
-    $customer_address_book_count_query = "SELECT c.*, ab.* from " .
-                                          TABLE_CUSTOMERS . " c
-                                          left join " . TABLE_ADDRESS_BOOK . " ab on c.customers_id = ab.customers_id
-                                          WHERE c.customers_id = " . (int)$customer_id;
-
-    $customer_address_book_count = $db->Execute($customer_address_book_count_query);
-    return $customer_address_book_count;
-  }
-
 /**
```
scottcwilson, line 2569: Perhaps preserve the interfaces for these old functions but use the new class under the hood?

includes/modules/pages/account/header_php.php
```diff
@@ -31,44 +31,9 @@
...
-  'date_purchased'=>$orders->fields['date_purchased'],
-  'order_name'=>$order_name,
-  'order_country'=>$order_country,
-  'orders_status_name'=>$orders->fields['orders_status_name'],
-  'order_total'=>$orders->fields['order_total']
-  );
-
-  $orders->MoveNext();
-}
+$customer = new Customer;
```
scottcwilson, line 35: Customer() ?

includes/modules/pages/login/header_php.php
```diff
@@ -57,26 +57,20 @@
...
-    // Check if email exists
-    $check_customer_query = "SELECT customers_id, customers_firstname, customers_lastname, customers_password,
-                                    customers_email_address, customers_default_address_id,
-                                    customers_authorization, customers_referral
-                           FROM " . TABLE_CUSTOMERS . "
-                           WHERE customers_email_address = :emailAddress";
 
-    $check_customer_query  =$db->bindVars($check_customer_query, ':emailAddress', $email_address, 'string');
-    $check_customer = $db->Execute($check_customer_query);
+    $customer = new Customer;
```
scottcwilson, line 61: Customer() ?

includes/modules/pages/account/header_php.php
drbyte: Ya, I'm leaving it off when not passing any args.

admin/includes/functions/general.php
drbyte: Yes. It's in the functions_customers file.

includes/functions/functions_customers.php
```diff
@@ -171,118 +181,144 @@ function zen_customer_greeting() {
...
+    return false;
 }
 
 /**
  * This function identifies whether (true) or not (false) the current customer session is
  * associated with a guest-checkout process.
+ * @alias Customer::isInGuestCheckout()
  */
-function zen_in_guest_checkout()
+function zen_in_guest_checkout(): bool
```
lat9, line 293: Shouldn't this function be the procedural way to invoke the `Customer` class method?

includes/functions/functions_customers.php
```diff
@@ -171,118 +181,144 @@ function zen_customer_greeting() {
...
+    $zco_notifier->notify('NOTIFY_ZEN_IN_GUEST_CHECKOUT', null, $in_guest_checkout);
     return (bool)$in_guest_checkout;
 }
 
 /**
  * This function identifies whether (true) or not (false) a customer is currently logged into the site.
+ * @alias Customer::someoneIsLoggedIn()
  */
-function zen_is_logged_in()
+function zen_is_logged_in(): bool
```
lat9, line 305: Shouldn't this function, also, be the procedural way to invoke the like-named `Customer` method?

drbyte:
> How is this going to work when multiple plugins require an override to the `Customer` class? I'm envisioning, for instance, a store that uses both **_One-Page Checkout_** and **_Dual Pricing_**.

Ya, I'd like to understand better what both of those are needing to "do" to the class. 
Handling of including extra fields in the $data property is probably best suited for an Observer.
Passing extra fields for display in the Admin customers listing should work the same as it did before.
Performing additional actions at various points can probably be handled by some new Notifier hooks (granted, all the prior ones are already included, but I guess some plugins were still directly altering some customer-related code so maybe more hooks are still needed but haven't been identified yet).

Extending this class isn't quite ideal yet, as it's only being instantiated "when needed", not registered globally to be available "all the time" and therefore replaceable/overridden later. I'm interested in seeing that happen, but was considering leaving that kind of logic to a different PR.

drbyte:
Another phase of refactoring this class into more of an Object is planned. 

If you can identify specific things to keep in mind related to plugins impacted by the change, I'm happy to incorporate those, or at least engage a discussion in preparation for incorporating.

Merging this phase for now. Continue the dialog as required!

PR #3854, opened 2020-08-08
Warn if downloadable product misconfigured

### LINKED ISSUE #2936: Should Downloads Manager page also display values for Product is Virtual/Always Free Shipping?
When set incorrectly (e.g. Product is Virtual = Yes or Always Free Shipping = Yes), shipping calculations will be wrong for a mixed cart of downloadable and physical products.  If the Downloads Manager page showed this, it would be easier to track down this misconfiguration.

scottcwilson:
Fixes #2936.

### DIFF
--- a/admin/downloads_manager.php
+++ b/admin/downloads_manager.php
@@ -62,6 +62,9 @@ if (zen_not_null($action)) {
           ?>
           <?php echo '</form>'; ?>
       </div>
+      <div class="row text-left">
+        <?php echo zen_image(DIR_WS_IMAGES . 'icon_status_yellow.gif') . TEXT_WARNING_PRODUCT_MISCONFIGURED_SHORT; ?>
+      </div>
       <div class="row text-center">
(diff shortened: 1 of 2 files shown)


admin/downloads_manager.php
```diff
@@ -184,6 +191,10 @@
                   '<a href="' . zen_href_link(FILENAME_DOWNLOADS_MANAGER, zen_get_all_get_params(array('padID', 'action')) . 'padID=' . $padInfo->products_attributes_id . '&page=' . $_GET['page'] . '&action=edit') . '" class="btn btn-primary" role="button">' . IMAGE_EDIT . '</a>' .
                   '&nbsp;<a href="' . zen_href_link(FILENAME_ATTRIBUTES_CONTROLLER, 'products_filter=' . $padInfo->products_id . '&current_categories_id=' . $padInfo->master_categories_id) . '" class="btn btn-primary" role="button">' . IMAGE_EDIT_ATTRIBUTES . '</a>'
                 );
+                $contents[] = array('align' => 'center', 'text' => '<a href="'  . zen_href_link(FILENAME_PRODUCT, 'cPath=' . $cPath . '&product_type=' . $padInfo->products_type . '&pID=' . $padInfo->products_id . '&action=new_product&page=' . $_GET['page']) . '" class="btn btn-primary" role="button">' . IMAGE_EDIT_PRODUCT . '</a>');
```
drbyte, line 194: Should we leave out the `$cPath` parameter here since it's not known in this context?

PR #3855, opened 2020-08-08
Deprecate zen_db_output()

drbyte:
It's only being used for screen output, not db-related.
Therefore should use zen_output_string_protected() instead.

### DIFF
--- a/includes/functions/database.php
+++ b/includes/functions/database.php
@@ -28,13 +28,15 @@ function zen_db_input($string)
     return $db->prepare_input($string);
 }
 
-  function zen_db_output($string) {
-    if (IS_ADMIN_FLAG) {
-       return htmlspecialchars($string, ENT_COMPAT, CHARSET, TRUE);
-    } else {
-       return htmlspecialchars($string);
-    }
-  }
+/**
+ * @deprecated use zen_output_string_protected() instead
+ * @param string $string
+ * @return string
+ */
+function zen_db_output(string $string)
+{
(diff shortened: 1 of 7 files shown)


includes/functions/database.php
```diff
@@ -28,13 +28,15 @@ function zen_db_input($string)
...
-    }
-  }
+/**
+ * @deprecated use zen_output_string_protected() instead
+ * @param string $string
+ * @return string
+ */
+function zen_db_output(string $string)
+{
+    return zen_output_string_protected($string);
```
lat9, line 38: If you're _really_ planning on deprecation, shouldn't this function be issuing an `E_USER_DEPRECATED` log?

drbyte: I suppose.

My IDE flags anything that's deprecated so I don't rely on logs for that, hence it's not top of mind.

Plus, I tend to resist littering the code with more logs about deprecated stuff. Only devs will pay any attention: End-users won't like having their logs filled up with stuff they can't do anything about.

I'll do it for this one.

lat9: Thanks, @drbyte.  IMO, it's better to signal impending doom (the deprecation log) so that when that rug is yanked (i.e. removing the function/method) at least the store-owner had a chance to either correct the issue or have the issue fixed.

drbyte: As I've thought more about it after writing my previous reply, I've shifted and agree strongly with your suggestion.

PR #3860, opened 2020-08-09
Functions refactor

drbyte:
Relocate functions to context-appropriate files

Most functions are now shared by both admin and catalog. (Exceptions are: `html_output.php` and `general`/`functions_general`)
Documented function signatures
Cleaned up majority of argument types
Deprecated some functions that were different between admin/catalog and refactored (usually admin) usages to use new shared equivalent. 
Functions that have been deprecated trigger an `E_USER_DEPRECATED` log.
Formatted files as PSR2 and updated SQL syntaxes

Also added various `@TODO` notations for identifying some planned additional deprecating or refactoring.


proseLA:
i do not know if this commit was the one that added the `functions_products` to the admin side or the one before it; as it was loaded twice on the catalog side (once with a `require_once`) and not at all on admin on v157; but thank you.

:thumbsup: :thumbsup:

drbyte:
Yes, I feel like it's all more logically arranged in this new structure in v158.
Particularly since the admin and catalog now share the same code for almost all functions.

Issue #3820, opened 2020-08-02
1.5.8 Pre-Release Documentation Tasks


scottcwilson:
* Pull all configuration description changes from 1.5.8 into upgrade script 
See the following PRs: 

#3800 
#3735 
#3888 - #3911 

* Other doc tasks as per linked PRs.

scottcwilson:
Please assign to me.

scottcwilson:
All done!

PR #3866, opened 2020-08-11
Add shop-by-brand page and add ability to Feature brands

drbyte:
`main_page=brands`

Brands marked as "featured" show at the top of the list

Screen Shot 2020-08-11 at 12 31 26 AM

Admin Manufacturers 2020-08-11 at 12 35 24 AM

Admin Manufacturers 2020-08-11 at 12 35 53 AM


zc_install/sql/install/mysql_zencart.sql
```diff
@@ -885,6 +885,7 @@ CREATE TABLE manufacturers (
   manufacturers_image varchar(255) default NULL,
   date_added datetime default NULL,
   last_modified datetime default NULL,
+  featured tinyint,
```
lat9, line 888: Shouldn't this have a default value?

zc_install/sql/updates/mysql_upgrade_zencart_158.sql
```diff
@@ -36,6 +36,7 @@ TRUNCATE TABLE db_cache;
 
 
 ALTER TABLE layout_boxes ADD plugin_details varchar(100) NOT NULL default '';
+ALTER TABLE manufacturers ADD COLUMN featured tinyint;
```
lat9, line 39: Likewise.

zc_install/sql/install/mysql_zencart.sql
drbyte: Yes, while core code won't have any issue with it setting a default will allow custom code that doesn't account for the column to not throw errors.
Thanks.

zc_install/sql/updates/mysql_upgrade_zencart_158.sql
```diff
@@ -36,6 +36,7 @@ TRUNCATE TABLE db_cache;
 
 
 ALTER TABLE layout_boxes ADD plugin_details varchar(100) NOT NULL default '';
+ALTER TABLE manufacturers ADD COLUMN featured tinyint default 0;
```
lat9, line 39: Just a question:  Will this cause issues if I backport this to a zc157 installation when the site's upgraded to v158?

drbyte: While that line will fail, zc_install "should" just note it in the `upgrade_exceptions` table and carry on.  Of course, we should test that. I don't recall off the top of my head whether `ADD` vs `ADD COLUMN` is the specific syntax it checks for (or both).

scottcwilson:
Doc update task #3820

PR #3877, opened 2020-08-12
Brands: Correcting PHP notices and empty sections

lat9:
The initial implementation didn't account for stores that have
- No brands defined
- No brands featured
- No brands other than those featured

### DIFF
--- a/includes/languages/english/brands.php
+++ b/includes/languages/english/brands.php
@@ -11,3 +11,5 @@ define('BREADCRUMB_BRANDS', 'Shop By Brand');
 
 define('FEATURED_BRANDS', 'Featured Brands');
 define('OTHER_BRANDS', 'Related Brands');
+
+define('NO_BRANDS_AVAILABLE', 'No brands are currently active for this store.');
--- a/includes/templates/template_default/templates/tpl_brands_default.php
+++ b/includes/templates/template_default/templates/tpl_brands_default.php
@@ -8,26 +8,49 @@
  */
 ?>
 <div id="indexBrandsList" class="centerColumn">
-
-    <h1 id="indexBrandsList-pageHeading" class="pageHeading"><?php echo FEATURED_BRANDS; ?></h1>
-
+    <h1 id="indexBrandsList-pageHeading" class="pageHeading"><?php echo HEADING_TITLE; ?></h1>
+<?php
+    // -----
+    // Display a message if no brands (aka manufacturers) are defined for the current store.
+    //
+    if (empty($brands['featured']) && empty($brands['other'])) {
+?>
+    <p><?php echo NO_BRANDS_AVAILABLE; ?></p>
+<?php
+    }
+    
+    // -----
+    // Display the list of featured brands, so long as at least one exists.
+    //
+    if (!empty($brands['featured'])) {
+?>
     <div class="featuredBrands">
+        <h2><?php echo FEATURED_BRANDS; ?></h2>
 <?php
     $list_box_contents = $brands['featured'];
     $title = '';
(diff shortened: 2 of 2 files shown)


lat9:
Within the page's `header_php.php`, there are currently hard-coded values for 'no_picture.gif' as well as the width/height (126/126) in which to display the manufacturer's image.

Shouldn't these, at a minimum, be definitions within the page's language file?

scottcwilson:
+1.  More questions: 
Should the brands page have a define page? 
Should it be controlled from Admin > Configuration > Define Pages to address @lat9 concern re: stores that don't use manufacturers? 
Should it be added to the site map page ?

drbyte:
@lat9 thanks for this!

scottcwilson:
Thanks @drbyte and @lat9 for this fun new feature.  Nice addition!

drbyte:
> Within the page's `header_php.php`, there are currently hard-coded values for 'no_picture.gif' as well as the width/height (126/126) in which to display the manufacturer's image.
> 
> Shouldn't these, at a minimum, be definitions within the page's language file?

Good points. I hard-coded "no_picture.gif" because I didn't want to go look up the name of the constant that defines it. 

The image size 126 would ideally become just CSS-controlled instead of hard-coded in HTML via PHP.
We could just use the standard product-small-image setting size. 
For this initially I wanted slightly larger images and didn't want to create "more constants, and more configuration settings". In the end I realized it doesn't matter much when the images are smaller than the provided dimensions, but becomes more important to force them down to the smaller dimensions if they're larger than specified.

Very inclined to just use the product-image-thumbnail constant, and people can change that if they want.
And would prefer to find a way to let CSS constrain it without having to hard-code it in PHP. That way, since it's an aesthetic templating thing really, it can be template-specific instead of store-specific.

drbyte:
> Should the brands page have a define page?

That would indeed make it "consistent with capabilities on other pages".
But I really wanna scream "NO!!!! No more define pages! (and associated additional config switches, files)."
That's my first reaction anyway.

> Should it be added to the site map page ?

I'd considered that. Then I worried that people would want to expand it out to list all manufacturers as links in that page, and that can make the page even more unwieldy than it already is. Then I plain forgot about it.

> Should there be an "information sidebox switch" added?

Again, I wanna scream "no more config switches" ...


... 
That said, I'm open to more discussion with arguments for and against these things.

scottcwilson:
It makes no sense to make a prohibition on config switches that are needed.  I know a number of stores who don't use manufacturers would would actively not want this feature.  There needs to be a place to disable it, and Define Pages seems like the best match.

drbyte:
> I know a number of stores who don't use manufacturers would would actively not want this feature. 

Naturally.

> It makes no sense to make a prohibition on config switches that are needed. There needs to be a place to disable it, and Define Pages seems like the best match.

We need to fix the perception that Define Pages Settings are for "turning this thing on or off". 
"Define Pages" are for creating special message content to appear above the normal page content. They have nothing to do with "show a link or not". 
That configuration group name is entirely unintuitive.

scottcwilson:
New group for disabling pages! :)

drbyte:
Renaming the current group is probably a good start for that.

The problem with then making that same group be what end-users "expect" it to be is that it requires extensive updates to all the catalog-side templates, which then also makes all prior templates obsolete (er, well, they won't support the feature)

scottcwilson:
> We need to fix the perception that Define Pages Settings are for "turning this thing on or off".

The Link OFF settings disable the creation of links to their respective page.  This means unless you know the URL structure of Zen Cart, you can't get to the associated page.  That's pretty close to off.  

Should we go the last mile and update the header_php.php files for each associated page to throw a 404 if the Link is set to OFF?   That's almost certainly the storeowner's intent. 

At any rate, yes you're right that this area is confusing - I have worked to build out the documentation to make it less so.

https://docs.zen-cart.com/user/template/define_pages/

PR #3879, opened 2020-08-12
Cast float values before INSERT statement, just as ints are casted

scottcwilson:


drbyte:
Perhaps good for specificity, but @zcwilt isn't this redundant?

drbyte:
Also, should be applied to other product types as well

scottcwilson:
Yes - missed the music one.  That's the only other one.

drbyte:
> What about casting to the other data types associated such as discussed at [13d204f#r41423083](https://github.com/zencart/zencart/commit/13d204fb44fdc06d42bbdba2323b8db2ced52d68#r41423083)

@zcwilt  ?

zcwilt (approved):

Issue #3873, opened 2020-08-12
redirect is incorrect when product type not default


proseLA:
https://github.com/zencart/zencart/blob/dd3875ab3457523e0be7853ff1c75a63492ac27d/admin/product.php#L36-L42

perhaps, this is an edge case scenario.  but i am trying to work within the confines of ZC.

let's say you attempt to create a new product, different from the default, that gets caught in this error.  on the redirect, you would lose the product type, and hence be in a different `type_handler` for the `new_product` action. (i have not tested it with the meta tags preview error; as i find it hard to imagine a separate meta tags collection based on the `type_handler`; but i could be wrong.  and the error would exist there as well.)  

in my testing, a viable solution would be to change the action, and use the messageStack to add the message rather than add session.  my testing shows that changing lines 40-41 to: 

```
          $messageStack->add(ERROR_NO_DATA_TO_SAVE, 'error');
          $action = "new_product";
          break;
```

would now, show the message on this load; as well as not do the redirect and lose the type handler.  i have no problem doing a PR for this change (i can include the `new_product_meta_tags` as well); but i was curious about thoughts prior to proceeding.

mc12345678:
Something else I noticed in the existing redirect is that even if none of the "necessary" information is entered, that anything that has been entered is lost on the redirect, so it would seem that a solution that doesn't lose posted information such as this may be less aggravating, even if the individual repetitively pressed the "preview" button as it floats in the one location (at least on my screen it does I haven't verified CSS has been updated).

It would seem that by allowing the rest of execution to occur outside of this `select`/`case` condition that the display would be loaded/reloaded as if originally entering the `new_product` editor but retaining any other existing entered information.

BTW, to see this in action, create a new Document General product scroll down the screen and note the "missing" fields, alter the product's status away from whatever was shown on load, then without entering any other information select preview.  The error message is displayed at the top, but now when scrolling down, all of the standard product fields are displayed (including price which was otherwise hidden by default (a completely separate issue which has removed some previously convenient workarounds)).

I'm on board with the modification as suggested (though maybe using single quotes instead of double. :P)

proseLA:
i think this solution is correct.  if others agree, i feel this line needs to change as well:

https://github.com/zencart/zencart/blob/3ef5f58ebeaf6505985e66d198b10719b9468432/admin/product.php#L15

[...]

PR #3882, opened 2020-08-14
no redirect on error; change action and continue #3873

proseLA:


drbyte:
Interesting - what's the use-case for changing `$action`?

proseLA:
in the first commit, we established that by changing the action, we can preserve all of the post data, and still provide the error message without the need to reload the page via a redirect.  if we redirect the page as is, we would need to rebuild all of the post data which is currently lost the way ZC stands right now.

to get to a preview page we would normally have something like `$product->isValid()` returning true prior to doing a preview.

the second commit, allows us to do that validation through the notifier; if the validation fails, we can change the action again from the `new_product_preview` to `new_product`.

classic use case would be something like UPC or EAN numbers, which ideally need to be validated prior to saving into the database.  although there could be any number of additional fields that are required and would need to be validated.

drbyte:
@proseLA eventually I'd like to remove the "preview" page altogether. 
Ideally validation would happen still on submit, but would come back to this same page if there were a problem, instead of using an intermediary preview page. I'm mentioning it just to encourage you to architect whatever you're working on to not expect the presence of a preview page.

proseLA:
> @proseLA eventually I'd like to remove the "preview" page altogether.
> Ideally validation would happen still on submit, but would come back to this same page if there were a problem, instead of using an intermediary preview page. I'm mentioning it just to encourage you to architect whatever you're working on to not expect the presence of a preview page.

- appreciate the heads up
- i'm comfortable with whatever decision you make on this/direction you take.
- i'm confident that i will be able to work with whatever gets done.  

thanks again for the heads up.

Issue #3891, opened 2020-08-17
Bug: admin function zen_draw_products_pull_down_categories_attributes, passing of parameters


torvista:
This is only used in 
option name manager global tools
option value sorter

When I overhauled/keelhauled option name manager I modified this function to allow filtering, and thought I had preserved the legacy functionality used by option value sorter and any a.n.other add-ons.
Apparently not.
So in option value sort, the category drop down is broken as the switch case default does not work as I imagined it did, in the function. Obvious now...

As I am not sure how best to fix/improve this between the three files, in terms of passing parameters or not, and maybe others can say whether this function is ever used elsewhere to worry about legacy support....some discussion please.

torvista:
I see there is similar stuff in zen_draw_products_pull_down_attributes, so I'll use that as a basis.

PR #3895, opened 2020-08-18
fix for non-display of category dropdown for option values sorter

torvista:
fixes https://github.com/zencart/zencart/issues/3891

### DIFF
--- a/includes/functions/functions_categories.php
+++ b/includes/functions/functions_categories.php
@@ -505,7 +505,7 @@ function zen_draw_products_pull_down_categories($field_name, $parameters = '', $
  * @param string $filter_by_option_name
  * @return string
  */
-function zen_draw_products_pull_down_categories_attributes($field_name, $parameters = '', $exclude = [], $show_full_path = false, $filter_by_option_name = '')
+function zen_draw_products_pull_down_categories_attributes($field_name, $parameters = '', $exclude = [], $show_full_path = false, $filter_by_option_name = null)
 {
     global $db, $currencies;
 
@@ -544,7 +544,7 @@ function zen_draw_products_pull_down_categories_attributes($field_name, $paramet
             if (!in_array($category['categories_id'], $exclude, false)) {
 
                 $select_string .= '<option value="' . $category['categories_id'] . '">';
-                if ($show_full_path) {
+                if ($show_full_path) {//todo full sorting on category paths
                     $select_string .= zen_output_generated_category_path($category['categories_id']);
                 } else {
                     $select_string .= $category['categories_name'];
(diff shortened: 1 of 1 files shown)


drbyte:
How did you arrive at setting the value to `null` instead of `''` ?

It seems the code is testing for specific cases on lines 528-537.

Was `null` selected because of something specific in the code? or just a guess?

I'm trying to understand your thinking, to be sure we're not introducing incomplete changes.

torvista:
Well, it was the simplest fix from broken to non-broken, since I imagine this will need to be back ported.
null  was used in the function zen_draw_products_pull_down_attributes.

drbyte:
I suspect that the "bug" is in the fact that the value is being cast to `int` when the filter is set into the session:

`$_SESSION['selectedOptionId'] = (int)$_POST['options_id'];`

So, testing for `=== ''` probably is failing because of that.

And given that `-1` and `0` are valid as `int` values and `''` is not, it's probably best to rework it to use an integer.

drbyte:
I'm still puzzled about the proposed fix. And the original refactor.

Empty string is used in application code calling the function. Neither `-1` or `null` are valid insomuch as the dropdown choices which use the value are always only `''` or numeric strings matching `products_options_id` from the db table.

mc12345678:
@DrByte, on the other hand I'm not sure I understand what is not understood.

> I'm still puzzled about the proposed fix. And the original refactor.
> 
> Empty string is used in application code calling the function. Neither -1 or null are valid insomuch as the dropdown choices which use the value are always only '' or numeric strings matching products_options_id from the db table.


The function `zen_draw_products_pull_down_categories_attributes` is used in two locations as currently distributed. Those two are both in the admin, option_values.php and options_name_manager.php. They both use the function differently by having a different set (quantity) of parameters. By having only 2 parameters in option_values.php the last parameter (one being discussed) needs to have a default value that allows the function to do something (otherwise what's the purpose of calling it) which means that the default value can not equate to the first case identifier which does nothing because nothing was "selected", and obviously shouldn't fall into the second case which is where an item has been selected and so, it seems that the default case should be activated. 

For a use where all parameters are in use, then whatever value appears in the first case statement should be "reachable" or align with the provider of that/those values. Because all of the parameters are used in the second application, the actual default parameter value no longer is important as yes some "known" value will be fed.

If there is concern about the data type of this last parameter (`int` versus `string` or some other), then yes may want to try to force/cast the data after evaluation of the special assignment.

torvista:
While I'm in no position to contribute at the moment, I would point out that since I made many changes to name manager, code in there should not be necessarily treated as a point of reference/base for other implementation....

drbyte:
@mc12345678 So what's the point of setting the default to `null`, other than to no-longer-match the first `''` case in the `switch` loop? It could really have been set to anything that doesn't match the first check.

Anyway, we're going in circles here.

I'll accept the `null` default.
Other problems can be addressed separately.

@torvista can you remove the stray "todo" comment and force-push the change to this PR?

Issue #3896, opened 2020-08-18
bug: missing space before AND in zen_draw_products_pull_down_attributes


torvista:
`$sql .= "AND pa.options_id = " . (int)$filter_by_option_name;`

1064 You have an error in your SQL syntax; .....

drbyte:
Closed in #3897

PR #3897, opened 2020-08-18
Add missing space before AND, in zen_draw_products_pull_down_attributes

torvista:
fixes:https://github.com/zencart/zencart/issues/3896

`$sql .= "AND pa.options_id = " . (int)$filter_by_option_name;`

1064 You have an error in your SQL syntax; .....


drbyte:
Since https://github.com/zencart/zencart/issues/3896 serves no purpose on its own, and requires no discussion there, and this PR's meaning is moot without that description, can you copy that description into the first textbox above?

drbyte:
Thanks for updating. It's helpful, and preferable, to have things self-contained.

PR #3898, opened 2020-08-20
fix error on wrong cast type

Zen4All:
It is exepected that ```$pInfo->products_id``` is an integer, or else the functions will fail.


drbyte:
Given that there are probably calls to that function all over the place we might need to just change the function signature for zen_get_products_description to not force the scalar int for that field.
I keep forgetting that queryFactory essentially returns everything as strings, regardless of how they're stored in the db ... thus this strict feature of PHP7 isn't fully leverageable without some complexity.

Zen4All:
You're probably right, I just came across this while loading the page. I did not do any further investigation.

PR #3899, opened 2020-08-20
Add an asterisk to fields

Zen4All:
I noticed that in zc 1.5.8 on the admin side there is no longer an asterisk shown on required fields.
These changes add back the asterisk on fields that are set as required through the input or select function.

The asterisk is placed in an input addon span. This ensures the alignment on the right is always the same as fields that are not required

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -247,30 +247,45 @@ function zen_image_submit($image, $alt = '', $parameters = '')
     return $form;
   }
 
-////
-// Output a form input field
-  function zen_draw_input_field($name, $value = '~*~*#', $parameters = '', $required = false, $type = 'text', $reinsert_value = true) {
-    $type = zen_output_string($type);
-    if ($type === 'price') $type = 'number" step="0.01';
+  /**
+ * 
+ * Output a form input field
+ * @param string $name field name
+ * @param string $value field value
+ * @param string $parameters extra parameters like classes or id
+ * @param boolean $required field is required
+ * @param string $type filed type
+ * @param boolean $reinsert_value
+ * @return string
+ */
+function zen_draw_input_field($name, $value = '~*~*#', $parameters = '', $required = false, $type = 'text', $reinsert_value = true)
+{
+  $type = zen_output_string($type);
+  if ($type === 'price')
+    $type = 'number" step="0.01';
+  $field = ($required ? '<div class="input-group alert-danger">' . PHP_EOL : '');
+  $field .= '<input type="' . $type . '" name="' . zen_output_string($name) . '"';
 
-    $field = '<input type="' . $type . '" name="' . zen_output_string($name) . '"';
-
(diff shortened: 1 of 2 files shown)


admin/includes/functions/html_output.php
```diff
@@ -247,30 +247,45 @@ function zen_draw_form($name, $action, $parameters = '', $method = 'post', $para
...
+ * @param string $parameters extra parameters like classes or id
+ * @param boolean $required field is required
+ * @param string $type filed type
+ * @param boolean $reinsert_value
+ * @return string
+ */
+function zen_draw_input_field($name, $value = '~*~*#', $parameters = '', $required = false, $type = 'text', $reinsert_value = true)
+{
+  $type = zen_output_string($type);
+  if ($type === 'price')
```
lat9, line 264: Could you include braces around this clause?  Makes it more obvious that the single statement is to be executed for the condition.

admin/includes/functions/html_output.php
```diff
@@ -247,30 +247,45 @@ function zen_draw_form($name, $action, $parameters = '', $method = 'post', $para
...
+  if ($value == '~*~*#' && (isset($GLOBALS[$name]) && is_string($GLOBALS[$name])) && ($reinsert_value == true)) {
+    $field .= ' value="' . zen_output_string(stripslashes($GLOBALS[$name])) . '"';
+  } elseif ($value != '~*~*#' && zen_not_null($value)) {
+    $field .= ' value="' . zen_output_string($value) . '"';
+  }
 
-    if ($required && strpos($parameters, 'required') === false) {
-        $field .= ' required';
-    }
+  if (zen_not_null($parameters))
```
lat9, line 275: Ditto.

admin/includes/functions/html_output.php
Zen4All: I left the original code alone, but I like to have braces as well. Will put them in.

drbyte:
Do you mind including a screenshot of the before/after?

drbyte:
I wonder if the "required" 'asterisk' is even needed.
IMO it would be better to just output the `required` HTML attribute so that the browser enforces the requirement, and maybe use the `placeholder` to show either the asterisk or the word "required".

Zen4All:
Before:
before
After:
after

drbyte:
On the admin Users page, where you can add new admins, change admin user passwords, and set admin profile assignments, does this still display correctly?
I.e.: does the single-row form still work if fields are marked as required in this way?

Zen4All:
> On the admin Users page, where you can add new admins, change admin user passwords, and set admin profile assignments, does this still display correctly? I.e.: does the single-row form still work if fields are marked as required in this way?

They stay on one line.
admin users

drbyte:
Great. Thanks!

scottcwilson:
This really looks fantastic.  Thanks for doing it.

PR #3906, opened 2020-08-21
PayPal changed their page so the instructions need a tweak

scottcwilson:
Suggested for backport as well.


drbyte:
Looks like line 717 probably needs the same updates.
And an update statement added to the upgrade SQL.

PR #3908, opened 2020-08-22
Missed define for Admin > Account page

scottcwilson:


admin/includes/languages/english/lang.admin_account.php
```diff
@@ -8,6 +8,10 @@
 
 $define = [
     'HEADING_TITLE' => 'Admin Account',
+    'TEXT_ADMIN_NAME' => 'Admin Username',
```
drbyte, line 11: Was there a reason for changing this from `Username` to `Admin Username`?

I could see changing it if it reduced ambiguity, but I'm not seeing where there's any potential confusion, so I'm curious what your reasoning was.
Screen Shot 2020-08-23 at 1 31 46 PM

scottcwilson: It was Admin Username in another file.  It wasn't really a change - the define was missing before.

drbyte: The file formerly called `include('user.php')` language file. In 1.5.7 that contains just `Username`, hence it appeared to be a change.

PR #3909, opened 2020-08-22
change hard-coded required attribute to use function parameter

Zen4All:
The required attribute is hard-coded in the parameters. IMO it is better to use the required switch in the input-field function. When using the function, the asterisk will be put behind the field.


drbyte:
Yes, this was done because the prior formatting was putting a asterisk on a new line that didn't align with the input field.
Now that the function has been updated to accommodate inline formatting, this is probably fine.

Issue #3894, opened 2020-08-18
Questions about updating coupon_admin.php


Zen4All:
As promised in issue #3888, I started to update the coupon pages.
I have some questions about some things I see in the old code.
1. In the code below all results are set in $_POST variables. Is there a good reason for this, or can I convert them from ```$_POST['coupon_name'][$language_id]``` to ```$coupon_name[$language_id]```
https://github.com/zencart/zencart/blob/e43c21354e43d924a7648a3cbc48612aec472127/admin/coupon_admin.php#L1051-L1091
2. The next code makes no sense to me.
The way I understand, we set to ```$_GET['status']``` or 'Y'. After that we always set ```$status``` to ```$status[0]```, which is empty, as it is not defined earlier.
https://github.com/zencart/zencart/blob/e43c21354e43d924a7648a3cbc48612aec472127/admin/coupon_admin.php#L1243-L1249

drbyte:
1. Is that $_POST array used later for display purposes? eg: is this query just stuffing the POST array with db content in order to match an output-loop used later? 
If there's no need for the POST because you're converting it all to a local array (you mentioned $coupon_name), then it would make sense not to stuff it into POST here.
Granted, it's a multidimensional array for language-purposes, so it's a bit complex.

2. We don't use that construct very often: when working with strings `$status = $status[0];` is a shorthand way of saying "take just the first character of that string". 
I too find that confusing, and would be comfortable with seeing it changed to substr().
Or rewritten to check "is in_array(GET[status], ['Y', 'N']) then use that; else set it to Y". Probably easier to understand quickly that way. Or if the only options are Y/N, then check "if === 'N' then set it to 'N'; else Y".  Even clearer.

drbyte:
Closed in #3911

Issue #3888, opened 2020-08-16
Coupon Pages still using old styling


scottcwilson:
Most of the admin has been refreshed and looks so much better, but the coupon page is still looking like it did 15 years ago.  @Zen4All help us!  You're our only hope.

Zen4All:
:smile: I'll take look at all admin files. I was standing in my corner waiting for changes to slow down a bit, as the development is going in a fast pace lately. But if people can stay away from the coupon pages for a few days, I will make a PR

scottcwilson:
LOL I know what you mean! Much appreciated.

drbyte:
👍

scottcwilson:
Once complete, I will update screenshots in documentation under #3820

drbyte:
Closed in #3911

PR #3911, opened 2020-08-23
Styling updates for GV and coupon pages (admin)

Zen4All:
As discussed here #3888, and #3894, I made updates to the pages.

Please test and review, as it is very possible some things need tweaking


drbyte:
> Please test and review, as it is very possible some things need tweaking

What things do you anticipate need tweaking?

Zen4All:
> What things do you anticipate need tweaking?

Since I do not use GV coupons myself very often, I thought It my very well be possible I missed something.

PR #3912, opened 2020-08-23
Create define page for down for maintenance

scottcwilson:


scottcwilson:
- Editing the language file is not really storeowner friendly.
- You might want to provide additional details about what you are up to.
- You might want to tell them there will be a sale after maintenance is over.

wiztechinc:
I completely and totally approve of this. Clients sometimes need to put up specific messages temporarily. Editing a language file is a no go for most. I’m always for giving site owners as much control as possible.

scottcwilson:
- You might want to tell them they can call or email you and order that way.
- You might want to reassure them that you haven't been hacked.

drbyte:
@wiztechinc wrote:
> Clients sometimes need to put up specific messages temporarily. 

They can already do that directly in the admin config screen. Just most don't realize they already can.

> Editing a language file is a no go for most.

They don't need to edit any language file presently.

drbyte:
> You might want to tell them they can call or email you and order that way.

That seems like a reasonable thing to include in the default template for this page. Doesn't require a define-page to do that.

drbyte:
### Pros:
- add a second place for maintenance messaging, in addition to the existing one
- allows rich-text markup for the message

### Cons:
- zc_install automatically enables maintenance mode during the db-upgrade step, and waits for the storeowner to turn it off. A define-page would mean an out-of-date message will automatically be displayed.
- controlling maintenance messaging would now require using multiple places (ie: also the define-page-editor), which they will forget to edit once maintenance is done and thus will end up showing an "old" maintenance message the "next" time they use it.
- confusion about why the maintenance screen shows content from an old template, if they clone the file from elsewhere
- new support requests about how to remove that maintenance message they forgot they'd put up months ago
- increased core code to maintain and document

simon1066:
If the 'Website Maintenance Period' field is free-form text then perhaps its description should not include the (hh:mm) requisite.

drbyte:
For consideration:

- maybe we change the default text of `Maintenance period: ` to something else, perhaps even blank. 
- then they could replace the `2h00` with a whole paragraph of their own content
- maybe intentionally make that be a textarea field instead of a single-line field
- as simon1066 says, update the field description accordingly

Still, whichever approach is taken, would need to consider what to do about zc_install turning on the maintenance mode. Maybe it needs to also wipe out the previously-configured message with its own?

scottcwilson:
> Maybe it needs to also wipe out the previously-configured message with its own?

Tough but fair. :)

drbyte:
Closing this.
I don't want the core application having yet another define-page to maintain. I'd rather see all define-pages removed and replaced with something better. So, rather than adding more things that would get taken away later, I'm closing this.

I'd be open to a project discussion about improved ways that Maintenance Mode could be implemented, starting with a list of "things merchants are saying they need and why (about maintenance mode), ranked in order of importance".

Issue #3915, opened 2020-08-24
ZC 1.5.8: Bug in updating customer status


Zen4All:
When tying to update the customer status in the admin, by using the red and green buttons. The status is no longer changed as the sql query always sets the status to 4.
https://github.com/zencart/zencart/blob/a7330066d2cfb8d4373e8213eac25fe07dd569f5/includes/classes/Customer.php#L380-L387

This probably happend when copying the query from the function above this code. The ```4``` should be replaced with ```$status```.
I will make a PR shortly.

Zen4All:
So the above commit solde the bug partially.
There is also a problem with ```$customer['customers_authorization']```. It is an integer with value 0-4. If the value is not 0 it whows as the value in the hidden field named 'current'. But if the value is 0, the field value is empty. This is strange, because when I echo ```$customer['customers_authorization']```, it prints as 0
https://github.com/zencart/zencart/blob/4bb1c4e5e964f6244f57a7941eb0bc1e014e40f9/admin/customers.php#L1178-L1188

PR #3916, opened 2020-08-25
Fix bug: if the value is 0, the field value is empty. 

Zen4All:
replace hardcoded ```4``` with ```$status```
see #3915

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -1183,7 +1183,7 @@ if (zen_not_null($action)) {
                     <?php } else { ?>
                       <input type="image" src="<?php echo DIR_WS_IMAGES ?>icon_red_on.gif" title="<?php echo IMAGE_ICON_STATUS_OFF; ?>" />
                     <?php } ?>
-                    <?php echo zen_draw_hidden_field('current', $customer['customers_authorization']); ?>
+                    <?php echo zen_draw_hidden_field('current', strval($customer['customers_authorization'])); ?>
                     <?php echo '</form>'; ?>
                 </td>
                 <td class="dataTableContent text-right"><?php
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -380,7 +380,9 @@ class Customer extends base
     public function setCustomerAuthorizationStatus(int $status)
     {
         global $db;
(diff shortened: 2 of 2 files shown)


admin/customers.php
```diff
@@ -1183,7 +1183,7 @@ function check_form() {
                     <?php } else { ?>
                       <input type="image" src="<?php echo DIR_WS_IMAGES ?>icon_red_on.gif" title="<?php echo IMAGE_ICON_STATUS_OFF; ?>" />
                     <?php } ?>
-                    <?php echo zen_draw_hidden_field('current', $customer['customers_authorization']); ?>
+                    <?php echo zen_draw_hidden_field('current', strval($customer['customers_authorization'])); ?>
```
drbyte, line 1186: why `strval()`?

drbyte: Oh. I see, you explained that in a separate commit.

drbyte: Actually, why not `(int)` it?

drbyte:
@Zen4All Can you also edit the Title of this PR to reflect its intention? ie: words explaining the bug it's solving? 
The git log and the PR log are more readable that way. (And words are better than just putting an issue number in description for the same reason)

admin/customers.php
Zen4All: I tried int, but for some mysterious reason that just doesn't work for 0.

drbyte: That does seem mysterious.

I try to avoid `strval` because it's less performant than a more explicit cast.

includes/classes/Customer.php
```diff
@@ -380,7 +380,9 @@ public function banCustomer()
     public function setCustomerAuthorizationStatus(int $status)
     {
         global $db;
-        $db->Execute("UPDATE " . TABLE_CUSTOMERS . " SET customers_authorization = 4 WHERE customers_id=" . $this->customer_id, 1);
```
drbyte, line 383: I guess line 368 could be changed to call this function instead? (and remove the `global $db;` statement)

Zen4All: we can

PR #3918, opened 2020-08-27
Ask a Question flags not used

scottcwilson:
The `main_template_vars` pages need to reference these flags.


scottcwilson:
Consider for backport; this is a bug in 1.5.7 as well where this feature was added.

scottcwilson:
Urp.  You're right - I looked at the code and thought this was a reference to a field.

PR #3923, opened 2020-09-01
Add Gmail in list of options for email transport

scottcwilson:
image


scottcwilson:
Add to list for #3820

drbyte:
Any objection to removing Qmail from this list? 
I don't recall ever seeing anyone using it.

scottcwilson:
I'd rather not since that means a data conversion and possibly breaking someone's cart.

drbyte:
My thinking is that its presence just creates end-user confusion. They see it in the list and say "oh, maybe I should be using that instead?" and waste time needlessly exploring a feature that's irrelevant to basically everyone.

scottcwilson:
We could just remove it as an option, but leave it for legacy installs, or we could remove it completely. 

If you want to remove it completely, we should open a new issue and announce it on the forum to see if there's pushback.  If there are no objections, we'd need: 

- update of the configuration text for zc_install (similar to this PR) - I'll handle the parallel change in the 158 upgrade script.
- an update of Qmail values in the upgrade script:
UPDATE configuration SET configuration_value = 'php' WHERE configuration_key = 'EMAIL_TRANSPORT' and configuration_value = 'Qmail'; 
- remove Qmail processing from includes/functions/functions_email.php 

If we want to retain it for legacy installs, we could stop at the first item (and just comment it carefully so as not to trip up future maintainers).  Wasn't sure the direction you wanted to go here.

PR #3925, opened 2020-09-02
EZPages display control for mobile

scottcwilson:
You may not want ezpages in your header bar in desktop view, but you still need them in
the hamburger menu.

### DIFF
--- a/includes/modules/responsive_classic/ezpages_bar_header.php
+++ b/includes/modules/responsive_classic/ezpages_bar_header.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * ezpages_bar_header - used to display links to EZ-Pages content horizontally as a header element
+ *
+ * @copyright Copyright 2003-2020 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: DrByte 2020 May 19 Modified in v1.5.7 $
+ */
+if (!defined('IS_ADMIN_FLAG')) {
+  die('Illegal Access');
+}
+$zco_notifier->notify('NOTIFY_START_EZPAGES_HEADERBAR');
+
+$var_linksList = array();
+
+// test if bar should display:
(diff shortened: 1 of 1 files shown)


scottcwilson:
This newly created file `includes/modules/responsive_classic/ezpages_bar_header.php` is identical to the core file `includes/modules/responsive_classic/ezpages_bar_header.php` except for line 18, which adds a check for mobile.

dbltoe:
Works a charm but, I would put it in the
emplate_default or it could be lost shortly after installation.

scottcwilson:
I'm not sure you can count on `$detect` being defined, which would be required if this were in template_default.

scottcwilson:
@dbltoe Maybe you could build a help file specifically for responsive_classic that would point out issues like this?

drbyte:
> @dbltoe Maybe you could build a help file specifically for responsive_classic that would point out issues like this?

No, let's not create docs specific to a template.

drbyte:
> I'm not sure you can count on `$detect` being defined, which would be required if this were in template_default.

Perhaps `isset($detect) && $detect->...` would make it generic enough?

scottcwilson:
Let's start with what's there, which is a valuable fix.  I'm not sure why @dbltoe thinks things will get overwritten and lost - that's the whole point of override folders is to prevent that.

dbltoe:
Upon further review... Out of the Doctor's waiting lounge.
I can see where there is no need for a responsive file in the non-responsive template_default UNLESS we wanted to make sure this operation happened no matter what template the owner chooses.
And, since we have no idea if the chosen template will even use that type menu, I suppose a your_template would be out of the question as well.

wiztechinc:
I figured out some unusual ways of doing this kind of thing with my template. I’m extremely interested in this possibility. Seems to me this can be handy for template developers?

scottcwilson:
Go for it!  Thank you, Delia.   
John, this doesn't mean you're off the hook. :)

scottcwilson:
Possibly useful pages to cross link: 

https://docs.zen-cart.com/user/template/mobile_mode/
https://docs.zen-cart.com/user/template/responsive/

wiztechinc:
> Go for it! Thank you, Delia.
> John, this doesn't mean you're off the hook. :)

Well, as much as I like this idea, I have never used the responsive template so any documenation may be far in the future if I do it.

Issue #3928, opened 2020-09-03
Updates to multiple admin pages


Zen4All:
I am making updates to multiple admin pages. Each page will have it's own PR, but since the changes are fairly common acros all the pages, I am using this issue as as placeholder for the changes.

- Removing braces from include and require lines
- Update status icons from images to font-awesome icons
- Replace info- and current arrow images for font-awesome icon buttons
- Remove quotes on integers in queries
- Replace ```$_GET['page']``` with ```$currentPage```
- Make sql queries uppercase
- unify the seach box feature on pages where it is used. (idea for future update: put this in a function or class to prevent repetitive coding)
- Edit page replace labels with plain text where multiple inputs are used for 1 label
- Add id's to input fields
- Sorting drop-down remembers current selection
- Type casting multiple variables
- Replace Spiffycalender with datePicker

references to this issue will be made in the PR's

Zen4All:
That;s what my plan is.
There may of-course still be add-ons that use Spiffy, so maybe leave spiffy in 1.5.8 as deprecated, and remove completely by 1.5.9?

drbyte:
> * Edit page replace labels with plain text where multiple inputs are used for 1 label

Can you explain this further?

Zen4All:
Object like multi language inputs, and radio fields are not allowed to have a label with a ```for=""``` attribute. So these label neet to be converted to something that looks like a label, but is a ```<div>``` or a ```<p>```.
This was already done on pages like collect_info, I think by @zcadditions.

mc12345678:
Realize this is closed.

Noting that in some of the updates, the following code "template" has been removed:
````php
// Split Page
// reset page when page is unknown
                  if ((!isset($_GET['page']) or $_GET['page'] == '' or $_GET['page'] == '1') && !empty($_GET['IDParameter'])) {
                    $check_page = $db->Execute($zones_query_raw);
                    $check_count = 1;
                    if ($check_page->RecordCount() > MAX_DISPLAY_SEARCH_RESULTS) {
                      foreach ($check_page as $item) {
                        if ($item['association_id'] == $_GET['IDParameter']) {
                          break;
                        }
                        $check_count++;
                      }
                      $_GET['page'] = round((($check_count / MAX_DISPLAY_SEARCH_RESULTS) + (fmod_round($check_count, MAX_DISPLAY_SEARCH_RESULTS) != 0 ? .5 : 0)), 0);
                    } else {
                      $_GET['page'] = 1;
                    }
[...]

PR #3930, opened 2020-09-03
Styling updates to Admin featured page

Zen4All:
see #3928


admin/featured.php
```diff
@@ -6,138 +6,158 @@
...
+          $dt = DateTime::createFromFormat($local_fmt, $expires_date_raw);
+          $expires_date_raw = 'null';
+          if (!empty($dt)) {
+            $expires_date_raw = $dt->format('Y-m-d');
+          }
+        }
+        $expires_date = (date('Y-m-d') < $expires_date_raw) ? $expires_date_raw : '0001-01-01';
+
+        $db->Execute("INSERT INTO " . TABLE_FEATURED . " (products_id, featured_date_added, expires_date, status, featured_date_available)
+                      VALUES (" . (int)$products_id . ", now(), 1, '" . zen_db_input($featured_date_available) . "')");
```
drbyte, line 53: Somehow you dropped the $expires_date from the provided fields, but it is still specified in the insert query.

drbyte (changes requested):

includes/functions/featured.php
```diff
@@ -9,81 +10,92 @@
...
+ * @param int $featured_id
+ * @param int $status
+ */
+function zen_set_featured_status(int $featured_id, int $status)
+{
+  global $db;
+  $sql = "UPDATE " . TABLE_FEATURED . "
+          SET status = " . (int)$status . ",
+              date_status_change = now()
+          WHERE featured_id = " . (int)$featured_id;
```
lat9, line 26: Should this update include a 'LIMIT 1' for performance?

drbyte: Probably, yes.  And probably similar for all such pages.
But this PR is primarily about design changes, particularly making it consistent with design on other admin pages.

So, a separate PR for optimizing those update queries would be appropriate.

lat9: OK.  When I've got time, I'll start that/those reviews.

drbyte: No rush. Probably simpler after these PRs are merged anyway, to avoid merge conflicts.

admin/featured.php
drbyte: I added it when merging other updates.

PR #3934, opened 2020-09-03
Styling updates to Admin specials.php

Zen4All:
see #3928


admin/specials.php
```diff
@@ -6,194 +6,216 @@
...
+        if (DATE_FORMAT_DATE_PICKER != 'yy-mm-dd' && !empty($specials_date_available_raw)) {
+          $local_fmt = zen_datepicker_format_fordate();
+          $dt = DateTime::createFromFormat($local_fmt, $specials_date_available_raw);
+          $specials_date_available_raw = 'null';
+          if (!empty($dt)) {
+            $specials_date_available_raw = $dt->format('Y-m-d');
+          }
+        }
+        $specials_date_available = (date('Y-m-d') < $specials_date_available_raw) ? $specials_date_available_raw : '0001-01-01';
+        $expires_date_raw = zen_db_prepare_input($_POST['expires_date']);
```
drbyte, line 65: I noticed that in the price_manager refactor you used `specials_date_available` and `specials_expires_date`, but here you used `specials_date_available` and `expires_date`.

Same with Featured. 

I'm guessing the discrepancy is from them being done in different coding sessions. 
I'm curious your thoughts on whether to standardize?

```
                     $('#specials_date_available').datepicker();
                     $('#specials_expires_date').datepicker();
                     $('#featured_date_available').datepicker();
                     $('#featured_expires_date').datepicker()
```

Issue #3922, opened 2020-09-01
$_GET method of admin's zen_draw_form does not support action page other than self


mc12345678:
https://github.com/zencart/zencart/blob/b3ed7bc02a0d4fbad68e7a2f086b79d05d5ee33a/admin/includes/functions/html_output.php#L245-L247

(Feel free to revise the title to something that might sound "better").

Basically, if I am on page X in the admin and setup a form using `zen_draw_form` where the form uses `get` parameters instead of `post` parameters, if the action page is not page X, then only page X gets loaded/reloaded when the associated submit button is pressed.  This is instead of loading the alternate page (say page Y) passing along the data that was to appear on the link address...

This was "discovered" through a problem with [Additional Product Images](https://www.zen-cart.com/downloads.php?do=file&id=1063) with a solution posted here: https://www.zen-cart.com/showthread.php?149805-Additional-Product-Images-addon&p=1372382#post1372382

Obviously another way "around" this is to have the current page "process" the self request and then take additional action which really would likely be that the code of the alternate page would be brought into the code of the one file or everything to be modified to use posting instead of getting data...

Note the "use" and results of this style of coding in the following ZC core admin files and note that a 'correction' to how `admin/includes/functions/html_output.php` operates would mean that the following may need adjustment:
https://github.com/zencart/zencart/blob/b3ed7bc02a0d4fbad68e7a2f086b79d05d5ee33a/admin/attributes_controller.php#L775
https://github.com/zencart/zencart/blob/b3ed7bc02a0d4fbad68e7a2f086b79d05d5ee33a/admin/products_price_manager.php#L291

These locations may "search" just fine; however, they do not move to that associated page.  Of course I have found similar "issues" with other code (non-ZC core code) that is currently present, but I haven't used/tried the features that would have revealed this problem.

I'm not sure that I follow why the referenced html_output code section was added, but...

FWIW, I have also looked into the potential use of `' get'` (adding a space either before or after the word) to prevent the use of the above code as the form's method will be properly populated, but this results in the admin's index page being loaded upon attempting to submit the form.

scottcwilson:
I'm pretty sure I have been stung by this too.  Thanks for writing it up, @mc12345678. 

Possible documentation impact to:
https://docs.zen-cart.com/dev/code/forms/
https://docs.zen-cart.com/dev/plugins/upgrading_to_1.5/#rewriting-addon-admin-pages-to-use-form-posts-instead-of-gets

drbyte:
[...]

PR #3940, opened 2020-09-06
refactor product search

proseLA:
pursuant to a conversation on the above referenced PR, i thought about refactoring the building of where clauses for sql statements on keyword searches.  my thoughts are to have a function, where you pass an array of field names, and a string of keywords.  the function would then build the where part of the sql statement.

in the conversation on PR #3804, there are a bunch of scripts where this might be beneficial.

this PR shows proof of concept and i'm certainly open to input.  

it is built off code from:
https://github.com/zencart/zencart/blob/4717032046d64e92ff5a71552b019636ef3fc8f5/includes/modules/pages/advanced_search_result/header_php.php#L317-L359
and that could certainly be part of the refactoring as well.

### DIFF
--- a/includes/functions/functions_search.php
+++ b/includes/functions/functions_search.php
@@ -157,7 +157,7 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
     return false;
 }
 
-    function zen_build_where($fields, $string)
+    function zen_build_keyword_where_clause($fields, $string)
     {
         global $db;
         if (zen_parse_search_string(stripslashes($string), $search_keywords)) {
@@ -172,25 +172,26 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
                         break;
                     default:
                         $sql_add = " (";
-                        $first = true;
-                        foreach ($fields as $k => $v) {
-                            if (!$first) {
+                        $first_field = true;
+                        foreach ($fields as $field_name) {
+                            if (!$first_field) {
                                 $sql_add .= ' OR ';
                             }
-                            $first = false;
-                            if (strpos($v, '_id')) {
-                                $sql_add .= " :field = :keyword_num";
-                                $sql_add = $db->bindVars($sql_add, ':keyword_num', $search_keywords[$i], 'integer');
+                            $first_field = false;
+                            if (strpos($field_name, '_id')) {
+                                $sql_add .= " :field_name = :numeric_keyword";
+
                             } else {
-                                $sql_add .= " :field LIKE '%:keyword%'";
+                                $sql_add .= " :field_name LIKE '%:keyword%'";
                             }
-                            $sql_add = $db->bindVars($sql_add, ':field', $v, 'noquotestring');
+                            $sql_add = $db->bindVars($sql_add, ':field_name', $field_name, 'noquotestring');
                         }
                         $sql_add .= ") ";
 
                         $where_str .= $sql_add;
 
                         $where_str = $db->bindVars($where_str, ':keyword', $search_keywords[$i], 'noquotestring');
+                        $where_str = $db->bindVars($where_str, ':numeric_keyword', $search_keywords[$i], 'integer');
                         break;
                 }
             }
--- a/includes/functions/functions_search.php
+++ b/includes/functions/functions_search.php
@@ -156,3 +156,45 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
 
     return false;
 }
+
+    function zen_build_where($fields, $string)
+    {
+        global $db;
+        if (zen_parse_search_string(stripslashes($string), $search_keywords)) {
+            $where_str = " AND (";
+            for ($i = 0, $n = sizeof($search_keywords); $i < $n; $i++) {
+                switch ($search_keywords[$i]) {
+                    case '(':
+                    case ')':
+                    case 'and':
+                    case 'or':
+                        $where_str .= " " . $search_keywords[$i] . " ";
+                        break;
+                    default:
+                        $sql_add = " (";
+                        $first = true;
+                        foreach ($fields as $k => $v) {
+                            if (!$first) {
+                                $sql_add .= ' OR ';
+                            }
+                            $first = false;
+                            if (strpos($v, '_id')) {
+                                $sql_add .= " :field = :keyword_num";
+                                $sql_add = $db->bindVars($sql_add, ':keyword_num', $search_keywords[$i], 'integer');
+                            } else {
+                                $sql_add .= " :field LIKE '%:keyword%'";
+                            }
+                            $sql_add = $db->bindVars($sql_add, ':field', $v, 'noquotestring');
+                        }
+                        $sql_add .= ") ";
+
+                        $where_str .= $sql_add;
+
+                        $where_str = $db->bindVars($where_str, ':keyword', $search_keywords[$i], 'noquotestring');
+                        break;
+                }
+            }
+            $where_str .= " )";
+        }
+        return $where_str;
+    }
--- a/includes/functions/functions_search.php
+++ b/includes/functions/functions_search.php
@@ -168,7 +168,7 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
                     case ')':
                     case 'and':
                     case 'or':
-                        $where_str .= " " . $search_keywords[$i] . " ";
+                        $where_str .= " " . strtoupper($search_keywords[$i]) . " ";
                         break;
                     default:
                         $sql_add = " (";
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -676,7 +676,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                     'pd.products_description',
                     'p.products_id',
                 ];
-                $where .= zen_build_where($fields, trim($keywords));
+                $where .= zen_build_keyword_where_clause($fields, trim($keywords));
             } else {
                 $products_query_raw.= " LEFT JOIN " . TABLE_PRODUCTS_TO_CATEGORIES . " p2c USING (products_id) ";
                 $where .= " AND p2c.categories_id=" . (int)$current_category_id;
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -670,16 +670,13 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
             $where = " WHERE pd.language_id = " . (int)$_SESSION['languages_id'];
 
             if ($search_result && $action != 'edit_category') {
-                $parts = explode(" ", trim($keywords));
-                foreach ($parts as $k => $v) {
-                    $sql_add = " AND (pd.products_name LIKE '%:part%'
-                              OR pd.products_description LIKE '%:part%'
-                              OR p.products_id = ':part'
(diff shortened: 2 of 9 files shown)


drbyte:
Can you give some commentary on the meaning of the code that's parsing for `'_id'`? 
Actually, commentary on all the things you've added beyond the header_php.php code you modeled it from would be helpful.

I'm guessing at the meaning of `:keyword_num` is something like `the current iteration thru keywords`? Is there something more descriptive that can be used for faster understanding when reading the code?

While `zen_build_where()` is brief, it's very generic when considered across the landscape of all available global functions in Zen Cart. Perhaps making it more specific will lead to less potential misuse later -- perhaps `zen_build_where_for_search()` or `zen_build_search_where_clause()` ?

proseLA:
i was actually thinking that the naming could be better.  i like `zen_build_keyword_where_clause`.  although the two you have proposed work for me as well.

i do a search in the fieldname for `_id`, so that is someone passes `customers_id` or `products_id` in the field names, i move from sql `like '%keyword%'` to `= keyword`. at that point, i'm changing the placeholder from keyword to keyword_num and binding it immediately to an integer value.   in looking at it the code, i think moving line 183 to below 193 would make it clearer without ill effect.  so the binding of the vars happens together.

if the field has `_id` in it, i want an `=` and the keyword should be bound integer.  else the comparison is `like` and bound as a string.  i make use of the var `$first` to not add or in-between field names, as the first field should not have one.  

perhaps changing `$v` to `$field_name` or `$field` might make it a bit clearer.  

there is an assumption being made that this is part of an existing where clause.  as the return of the functions starts with `AND`.

proseLA:
i have been thinking about this function for a bit, and i see the potential for use a number of places.

drbyte:
Thanks. That's clearer, easier to follow.

includes/functions/functions_search.php
```diff
@@ -156,3 +156,46 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
...
+                        $where_str .= " " . $search_keywords[$i] . " ";
+                        break;
+                    default:
+                        $sql_add = " (";
+                        $first_field = true;
+                        foreach ($fields as $field_name) {
+                            if (!$first_field) {
+                                $sql_add .= ' OR ';
+                            }
+                            $first_field = false;
```
drbyte, line 180: Should this `$first_field=false` be inside the `if` above it?

proseLA: no.  that would not work.  we would never hit it.  `$first_field` would be set to true, and then we would never set it to false.  it has to be outside of the if statement and after it.

drbyte: Oh duh. Right 😊

drbyte:
I like it.

Is there anything further you'd like to do with it? 

Do you foresee any challenges when working with multiple tables' fields being passed through? (It seems that as long as table-joins are already taken care of, and ideally tablename aliases are used/passed appropriately, it should be fine. Just curious whether you'd been anticipating any issues we should discuss.)

proseLA:
i have thought about perhaps doing some error checking, such as using the sniffer to check for fields, but i think that could get very carried away.  i also thought the start of the return which is `AND `; and whether there could be a spot where there this return is the whole of the where clause.  but that does not seem to be the case in the places i have looked.

i did look at some of the other places where this function would be used and i came across this bit of code:

https://github.com/zencart/zencart/blob/4717032046d64e92ff5a71552b019636ef3fc8f5/admin/customers.php#L1062-L1079

i am not that familiar with `RLIKE` and whether we need to perhaps incorporate that into the function.  i can only assume someone is searching there using regular expressions?  it seems that type of search pre-dated PR #3079 which started this journey.

drbyte:
Ya, the RLIKE syntax is for regex stuff, but I don't see why that's even beneficial here. More specifically I'm pretty certain nobody uses regex patterns in their searches, so I'm pretty much unconcerned with retaining that. 

Pinging @zcwilt to see if he has any historical insights regarding why RLIKE was implemented, since it's been there since at least v1.1.0.

mc12345678:
For capture of limitation and perhaps pointing out the obvious, note that the following current database table->fields will likely not be usable by this function (regardless of how [un]likely they are to be used):
authorizenet->transaction_id
customers_basket->products_id
customers_basket_attributes->products_id, products_options_id
paypal->payer_id, receiver_id, txn_id, parent_txn_id
paypal_payment_status_history->txn_id, parent_txn_id
paypal_session->session_id
paypal_testing->payer_id, receiver_id, txn_id, parent_txn_id
whos_online->session_id

proseLA:
> For capture of limitation and perhaps pointing out the obvious, note that the following current database table->fields will likely not be usable by this function (regardless of how [un]likely they are to be used):
> authorizenet->transaction_id
> customers_basket->products_id
> customers_basket_attributes->products_id, products_options_id
> paypal->payer_id, receiver_id, txn_id, parent_txn_id
> paypal_payment_status_history->txn_id, parent_txn_id
> paypal_session->session_id
> paypal_testing->payer_id, receiver_id, txn_id, parent_txn_id
> whos_online->session_id

the genesis of this function was to have consistent keyword searching when looking for products.  currently, when looking for a product using the search on the storefront, one would receive different results, then if an admin was looking for products under the featured products admin.  the intention was for consistent searching across various pages.

its potential usage has now expanded.

you are correct in your assertion.

if we wanted to address these edge cases, we could modify the passed array to the function to also have the table name for each field being used.  we could then sniff out every field for its type and adjust the sql statement accordingly.  

i am not sure the processing cost is worth it for those edge cases.  but i am open...

drbyte:
Ignore those edge-cases. If a real legit business issue arises that requires addressing them, that can be handled "then".
Also no need to get into table schema-inspection at this stage.

Both are over-engineering for a hypothetical situation, not an actual situation.
[YAGNI](https://www.martinfowler.com/bliki/Yagni.html)

proseLA:
i left out the admin/customers for now.

i have done some testing, and corrected errors.  i will continue to look; but i think this provides a more robust searching in numerous areas.  i am always open for additional thoughts and suggestions.

i will now be using YAGNI in conversation with all clients at all times.  thank you for that.

drbyte:
If the reason for leaving out admin/customers.php is just the RLIKE stuff, I'd say go ahead with dropping the RLIKE.
I think these improvements probably give sufficient power that the other won't be missed.

drbyte:
Looking good.

In #3804 you mentioned attributes_controller and price_manager, but I don't see those here. Curious about your thinking on these.

proseLA:
i did re-look at both of those scripts.  the keywords var there are part of the form that looks to call the `FILENAME_CATEGORY_PRODUCT_LISTING`  i was mistaken that those 2 scripts could benefit from this refactor.  the only time the `$keywords` var is used on those 2 scripts is below.

https://github.com/zencart/zencart/blob/689b11c7b738bd67b3ae62baad90efb2d980ccdc/admin/attributes_controller.php#L705-L715

https://github.com/zencart/zencart/blob/689b11c7b738bd67b3ae62baad90efb2d980ccdc/admin/products_price_manager.php#L276-L286

mc12345678:
Personally not sure what the continued reasoning for not including the two other admin files, but let me reference the above two search sections to an issue that may or may not modify reasoning to further the application of the search functionality. See issue #3922 where the "destination" identified by those two searches is effectively ignored and only currently offers complete html structure. So if it was because they attempt to go to FILENAME_CATEGORY_PRODUCT_LISTING that incorporation was skipped, the real functionality at the moment is that because it is a 'get' method form that the `cmd` parameter on the browser is what will process the request.

proseLA:
> Personally not sure what the continued reasoning for not including the two other admin files, but let me reference the above two search sections to an issue that may or may not modify reasoning to further the application of the search functionality. See issue #3922 where the "destination" identified by those two searches is effectively ignored and only currently offers complete html structure. So if it was because they attempt to go to FILENAME_CATEGORY_PRODUCT_LISTING that incorporation was skipped, the real functionality at the moment is that because it is a 'get' method form that the `cmd` parameter on the browser is what will process the request.

i am not really sure why you are commenting here, as the problem you are referencing has nothing to do with this PR.

this PR refactors the construction of an SQL statement that utilizes keyword searches and makes it better.  it takes an array of field names and a string, and from that constructs a where statement and makes it consistent across the multiple places that ZC does this kind of search.  if the keyword search did NOT work prior to this PR, ie return results, this PR will not fix it.

when i look at the above 2 files, it seems that the attributes controller does return to the FILENAME_CATEGORY_PRODUCT_LISTING (and does the search better as defined by this PR), but the price manager does not.  

if the exhibited behavior is not what is intended, that is completely separate from this PR.  so i am not sure why you are commenting here, or what your expectation is here.

mc12345678:
*Sigh*, so much for getting any discussion on the topic, nothing helps better at trying to see improvement than to basically tell someone to go away. Unfortunately, its obvious that didn't understand the writeup about the referenced issue.

Let me try this.. From your earlier writing: 
> i did re-look at both of those scripts. the keywords var there are part of the form that looks to call the `FILENAME_CATEGORY_PRODUCT_LISTING` 

This appears to indicate that because there is a form call to `FILENAME_CATEGORY_PRODUCT_LISTING`, it was determined that no additional query modification should be made to use the change that is/was proposed in this PR... So... To try to not limit the applicability of this PR, I was trying to figure out what the halibut the reasoning was for not addressing attributes_controller and price_manager as otherwise asked about.  Again, to me, it appears from what was written that `echo zen_draw_form('searchForm', FILENAME_CATEGORY_PRODUCT_LISTING, '', 'get');` and/or `echo zen_draw_form('search', FILENAME_CATEGORY_PRODUCT_LISTING, '', 'get', 'class="form-horizontal"');` were the reasons...

Unfortunately, the way the admin's `zen_draw_form` code has been rewritten in ZC 1.5.7, it doesn't matter what constant is put in there when a get type form is used, because the content in the second parameter doesn't really do anything associated with the form's action... Meaning could use:
`echo zen_draw_form('search', 'Call_my_uncle_bob', '', 'get');` in the attributes_controller code of a default ZC install and it would act/respond the same way as it did previously... Presumably the result is/was as desired, but the "path" to get there is not desired which is partially why that particular issue is marked as a bug.  

Whatever, it's been merged, if any further improvement is seen applicable likely will be in another commit.

proseLA:
@mc12345678 over the years, i have developed a tremendous respect for your knowledge and your pedantic nature, despite not understanding you at times. 

i have very limited knowledge and experience with the attributes controller; perhaps a bit more on the price manager.

you already have an issue open for the issue you are discussing.  if you want to engage my help/opinion, feel free to do so there.

this PR dealt with the formation, improvement and consistent application of **mysql** statements.  nothing more.

i have learned not to hijack a thread, despite how strongly i may feel about an issue.  

had we been able to stay on topic, perhaps there could have been more conversation with regards to the use of `ADVANCED_SEARCH_DEFAULT_OPERATOR` which i've been thinking is more applicable for the customer facing side and perhaps not as applicable on the admin side.  i think `AND` is much more applicable as i would think admins always know what they are looking for.  and need better tools to find them.

but as you say, it's already merged.  and it seems that conversation will have to happen elsewhere.

drbyte:
I merged what was here because it was largely canonical, and the extra discussion around zen_draw_form seems tangential to the issue at hand. 

and Yes: conversation around the additional considerations makes sense in a new issue/PR.

PR #3943, opened 2020-09-07
Match strings on pages with those on menu (singular)

scottcwilson:


drbyte:
Initial thought:
While I'm a fan of using Singular in thousands of places where Plural is misused, in this case I wonder if these should be flipped the other way. "Option Values Manager" seems slightly better than "Option Value Manager" here, because we are indeed "managing" "many" (plural).

scottcwilson:
The singular is used in a whole lot of places.  It would be better to make a lower impact change and just fix this one case of using the plural.

torvista:
Bearing in mind the quantity and breadth of commits since the release of 157, does the importance of a low-impact change have any relevance these days? I think not, it seems prime time to make sweeping changes...this is only text after all.

mc12345678:
> While I'm a fan of using Singular in thousands of places where Plural is misused (eg: "products_id" 🤮 )

I actually consider the 's' in `products_id` to be possessive not plural (the id of a product)... Unfortunately apostrophes are definitely not supported in such a field.

There are also changes that have been seen here that were made to relate back to expectations versus functionality, I agree that if going to make a change to offer a "correction", the final state should be correct *and* consistent not just consistent.

drbyte:
Decided to leave it as-is for now.

PR #3944, opened 2020-09-08
Allow SQL patch FROM processing to not require capitalization.

mc12345678:
For the FROM "rule" had considered: `$line = 'FROM ' . $table_prefix . ltrim(substr($line, 5));`
but that only works for one replacement not multiple like the
str_replace/str_ireplace would do. I don't know exactly why multiple
would need to be replaced, but kept with the same concept.

### DIFF
--- a/admin/sqlpatch.php
+++ b/admin/sqlpatch.php
@@ -242,8 +242,8 @@ function executeSql($lines, $database, $table_prefix = '') {
             }
           }
           break;
-        case (substr($line_upper, 0, 7) == 'SELECT ' && substr_count($line, 'FROM ') > 0):
-          $line = str_replace('FROM ', 'FROM ' . $table_prefix, $line);
+        case (substr($line_upper, 0, 7) == 'SELECT ' && substr_count($line_upper, 'FROM ') > 0):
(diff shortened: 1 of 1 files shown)


drbyte:
Why is this suddenly a problem that didn't exist before?

mc12345678:
Good question, though would think that either the majority of such sql queries either would use all capital terms (like so many of the changes that have been made to queries captured in ZC code recently), that this particular portion of the query wasn't used frequently (starting a line with `from` along with a database that used a prefix).

Really, it was the initial reporting of the new message about a line not ending with a semicolon, followed by attempts to generate real SQL that used rules in `sqlpatch.php` causing a deeper look into the code.

Initially, I was ok with the error message popping up but it wasn't until brittainmark pushed further for a solution that didn't emit an error message about a missing semicolon because a query spanned multiple lines with reference to multiple tables and the database had a `DB_PREFIX`.


Alternatively figure could put some sort of message on the `sqlpatch.php` admin screen to indicate the use of all capital letters (which was my first consideration, but then realized that it was only a partial fix to the overall issue.

PR #3945, opened 2020-09-08
Correct strict PHP error

mc12345678:
Corrects a strict PHP error.

### DIFF
--- a/admin/sqlpatch.php
+++ b/admin/sqlpatch.php
@@ -833,7 +833,7 @@ if (zen_not_null($action)) {
           <div class="col-sm-12 text-danger"><strong><?php echo HEADING_WARNING2; ?></strong></div>
         </div>
         <?php
-        if ($action == 'execute' && $_POST['query_string'] != '') {
+        if ($action == 'execute' && !empty($_POST['query_string'])) {
           ?>
           <div class="row">
             <div class="col-sm-12 text-danger"><strong><?php echo TEXT_QUERY_RESULTS; ?></strong></div>


admin/sqlpatch.php
```diff
@@ -833,7 +833,7 @@ function init() {
           <div class="col-sm-12 text-danger"><strong><?php echo HEADING_WARNING2; ?></strong></div>
         </div>
         <?php
-        if ($action == 'execute' && $_POST['query_string'] != '') {
+        if ($action == 'execute' && isset($_POST['query_string']) && $_POST['query_string'] != '') {
```
lat9, line 836: Why not just use `!empty($_POST['query_string'])`?

mc12345678: Didn't take the opportunity to investigate a `falsey` `'query_string'` that was appropriate for processing.

lat9: I don't understand your response.

mc12345678: Why not? Because I hadn't evaluated if a problem would occur by using `!empty`. Still haven't, but responded at least.

drbyte:
Okay, I accept `Corrects a strict PHP error` as a description of sorts.
But the title confuses me: `Observed necessary after logout while monitoring a query's result`.
What do "Observed" and "necessary" and "after logout while" and "monitoring a query's result" have to do with each other? and this code PR?

admin/sqlpatch.php
drbyte: In this case, if a falsey value was all that was posted, it would never be a valid query, so `empty()` would be fine.

mc12345678:
The occurrence happened while "logging in again" from a timeout following a sql query execution...  That condition is basically necessary to come across this in a default installation. Otherwise, I'd be interested to know how it would be possible to have an action parameter of execute on the url and *not* have posted `query_string`.

Further in my git editor, the content in the first row becomes the title which was where I was trying to explain all that I thought necessary for this commit then realized that if I didn't have either strict PHP or php 7.4 enabled then it may not have been noticed. Apologies for having been lazy, please advise what would help future readers.

drbyte:
> Further in my git editor, the content in the first row becomes the title

Yes, it is typical behavior for git editor apps to use the first line of comments as the title. Therefore it is always best to start with a brief title that explains a summary of what is changing or what problem is being fixed. That way when reviewing the git log in the future the commit titles provide some context.
Then start the 2nd line (or leave 2nd blank and start on 3rd line) with explanatory details and steps-how-to-recreate etc.

While I can edit the PR title, it doesn't "fix" the git log which uses the actual commit's content.

PR #3953, opened 2020-09-13
do not return empty array on credit selection

proseLA:
if all of the modules that use the function `credit_selection` did not return an empty array, then there would be no need for checking them later in use, ie here:

https://github.com/zencart/zencart/blob/25233dbc9d848a2b31cbfaf29038c2bc91f8e903/includes/templates/template_default/templates/tpl_checkout_payment_default.php#L66

we could get rid of these `isset` checks if we only returned an array when properly populated.

### DIFF
--- a/includes/modules/order_total/ot_gv.php
+++ b/includes/modules/order_total/ot_gv.php
@@ -189,7 +189,6 @@ class ot_gv {
    */
   function credit_selection() {
     global $db, $currencies;
-    $selection = array();
     $gv_query = $db->Execute("SELECT coupon_id FROM " . TABLE_COUPONS . " WHERE coupon_type = 'G' AND coupon_active='Y'");
     // checks to see if any GVs are in the system and active or if the current customer has any GV balance
     if ($gv_query->RecordCount() > 0 || $this->use_credit_amount()) {
@@ -201,9 +200,8 @@ class ot_gv {
                          'field' => zen_draw_input_field('gv_redeem_code', '', 'id="disc-'.$this->code.'" onkeyup="submitFunction(0,0)"'),
                          'tag' => 'disc-'.$this->code,
                          )));
-
+      return $selection;
     }
-    return $selection;
   }
   /**
    * Verify that the customer has entered a valid redemption amount, and return the amount that can be applied to this order


mc12345678:
Although neither solution seems to resolve the need to validate that the `credit_selection` method within the associated `order_totals` module returns a `fields` key, it seems like if the intent is to reduce the occurrence of entering this section of code, then the fix would be in `includes/classes/order_total.php` by changing
https://github.com/zencart/zencart/blob/25233dbc9d848a2b31cbfaf29038c2bc91f8e903/includes/classes/order_total.php#L116-L128
to something like:
````diff
  function credit_selection() {
    $selection_array = array();
    if (is_array($this->modules)) {
      foreach($this->modules as $value) {
        if (isset($GLOBALS[$class]->credit_class) && $GLOBALS[$class]->credit_class == true ) {
          $selection = $GLOBALS[$class]->credit_selection();
-          if (is_array($selection)) $selection_array[] = $selection;
+          if (!empty($selection) && is_array($selection)) $selection_array[] = $selection;
        }
      }
    }
    return $selection_array;
  }
````

Because even an empty array is still an array which causes the "`$selection`" to be added to the `$selection_array` in the current code.

Looking over the version 1.5.8 code further, not all order_total modules return an empty array. Some return `false` such as `ot_group_pricing` which itself appears would be a problem if the class didn't filter that result out in the either version of the above changed line.

As far as "simplifying" the identified code, when either PHP 7.4 is required as a minimum PHP version or Laravel 5.7+ is used (with what appears to be version 8.0.4 incorporated into ZC 1.5.8), then could also use the null coalesce operator that was added by PHP 7.4 (and backfilled by Laravel 5.7+): `count($selection[$i]['fields'] ?? [])` in this area:
````diff
-for ($j=0, $n2=(isset($selection[$i]['fields']) ? sizeof($selection[$i]['fields']) : 0); $j<$n2; $j++) {
+for ($j=0, $n2=count($selection[$i]['fields'] ?? []); $j<$n2; $j++) {
````

Only potential downfall for that change (and not saying that it is important) is the loss of identifying the number of modules evaluated to have a `credit_selection` or to "directly" determine if a module had a `credit_selection` method without using some other inspection code (e.g. `if (method_exists('credit_selection', $selection[$i]['id'])) {`.  But that code depends on `$selection[$i]` containing the `id` key which also isn't known until tested...  

As for why this "check" is even present, well either can cause the page to stop loading and lose any access to information beyond the problem or can in a way silently omit that section of code because the `fields` key isn't present.  Which way is proposed if neither of those two?

proseLA:
yes.  i see your point.  i had noticed the return of `false` on that other total module.  i needed to dig a little deeper into it.

currently the return could either be `false`, an array or an empty array.  

i am personally a fan of having consistent return values, and cleaning up the code so that there are not always checks as to what we are looking at when we start start processing those returns.  however, i understand the need for them with regards to legacy code.

perhaps we can just add a return of false if we are outside of the if statement and move on?  unless the team prefers a more comprehensive refactor?

the null coalesce looks to be part of php 7; are u sure 7.4 is required for it?  i was unaware of it.  thanks for pointing it out.

scottcwilson:
Bear in mind that there are likely plugins which rely on the current way of doing things.

mc12345678:
> the null coalesce looks to be part of php 7; are u sure 7.4 is required for it? i was unaware of it. thanks for pointing it out.

The resource that I thought I was using was from the php.net website and I thought it indicated that it was a "new" feature as part of 7.4, although I was surprised to think that it took that many sub-versions to introduce it. Upon further review specifically of the php manual at example 5 of: https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce

Identifies that it was as early as version 7... so no, obviously not "sure" of it being just finally in 7.4. It does however, require some level of "incorporation" either through incorporation of things like Laravel (which appears to require 7.3 as I thought I read in one of its documents about version 8.) or at least php 7 to work.

As to existing plugins, yes, that's why I identified some of the potential pitfalls in modifying the order_totals class. While it would eliminate the attempt to get to the referenced line (because a credit_select method that didn't return an array with at least one element would not be recognized), it would modify the general behaviour as well as not exactly identify the presence of the fields key/array. The current implementation even expects certain things to be present including the fact that the fields key is also an array or countable.

Issue #3937, opened 2020-09-05
adding inactive items to cart....


proseLA:
i do not have a v158 up and running as of yet.  i will get to that soon...

i have tested and confirmed this behavior on v157 relatively thoroughly.  using the test data, set `DISPLAY_CART` to false.  then take a product (i used a mouse, pid = 3), and make it inactive.  now use the following link (as recommended to me by Dr Byte in a post on the board):

https://zc157.com/index.php?main_page=product_info&action=buy_now&products_id=3

the page is not found, but the product is added to the cart, as well as a success message is displayed.

i am not a fan of allowing inactive items to be added to a cart, and then removing them only when you get to the shopping cart page.  it seems wrong.   i am curious about others thoughts on this issue.

screenshot below:

![image](https://user-images.githubusercontent.com/1095136/92294211-e1671100-eede-11ea-81a1-3bd25aff3bb4.png)

lankeeyankee:
that is definitely unwanted behavior!

scottcwilson:
Here be monsters.   You can also add more than the max qty, less than the min, quantities that don't match the specified units, etc.  This stuff is checked on the shopping cart page.  In some cases it's fixed, in some cases checkout is blocked.  I agree it's not ideal but when you start digging in, it's easy to see why they took this short cut.

proseLA:
well, with all of the activity going on in the repo, perhaps it is something that we can look at and address.  if the logic is already embedded in the cart, then perhaps we need this script to access that same logic in an eloquent way.  without code duplication.  perhaps i will look at closer; time willing.

proseLA:
> Here be monsters. You can also add more than the max qty, less than the min, quantities that don't match the specified units, etc. This stuff is checked on the shopping cart page. In some cases it's fixed, in some cases checkout is blocked. I agree it's not ideal but when you start digging in, it's easy to see why they took this short cut.

ok, i do have a v158 machine spun up and the behavior is the same/similar.

@scottcwilson i can not confirm anything that you are saying.  in fact, i can confirm the function does respect the max quantity.  so i'm not real sure about the rest of your statement.  

with respect to my initial comment about this issue, a check on the products status could easily prevent this situation.

```
$check_valid = zen_products_lookup($_GET['products_id'], 'products_status');
```

[...]

PR #3954, opened 2020-09-14
disallow add to shopping cart for actionBuyNow and fix redirect

proseLA:
pursuant to conversation on #3937; i have added a message for attempting to add an invalid product to the cart, and also addressed uncalled for redirects.  i think we should respect the created links and remove the action.  

in the referenced issue, creating a link to add 1 to the cart and go directly to the product page for that item fails; as the `products_id` gets added to the parameter exclude list.  apparently there is an attempt to make this not happen for reviews, however, i believe the `$parameters` array will already contain the `products_id`, so in my testing the if statement does nothing.

the addition of a notifier in the `zen_get_products_allow_add_to_cart` function is great, and was the exact idea that i had.  i think the function should check for an inactive product as it is already hitting the products table.  it seems that the function might be able to do more, prior to handing over to the observer.

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -2036,7 +2036,9 @@ class shoppingCart extends base
         if (isset($_GET['products_id'])) {
             if (zen_requires_attribute_selection($_GET['products_id'])) {
                 zen_redirect(zen_href_link(zen_get_info_page($_GET['products_id']), 'products_id=' . $_GET['products_id']));
-            } else {
+            }
+            $allow_into_cart = zen_get_products_allow_add_to_cart((int)$_GET['products_id']);
+            if ($allow_into_cart == 'Y') {
                 $add_max = zen_get_products_quantity_order_max($_GET['products_id']);
                 $cart_qty = $this->in_cart_mixed($_GET['products_id']);
                 $new_qty = zen_get_buy_now_qty($_GET['products_id']);
@@ -2064,15 +2066,17 @@ class shoppingCart extends base
             }
         }
         // display message if all is good and not on shopping_cart page
-        if ((DISPLAY_CART == 'false' && $_GET['main_page'] != FILENAME_SHOPPING_CART) && $messageStack->size('shopping_cart') == 0) {
+        if ((DISPLAY_CART == 'false' && $_GET['main_page'] != FILENAME_SHOPPING_CART) && $messageStack->size('shopping_cart') == 0 && ($allow_into_cart == 'Y')) {
             $messageStack->add_session('header', ($this->display_debug_messages ? 'FUNCTION ' . __FUNCTION__ . ': ' : '') . SUCCESS_ADDED_TO_CART_PRODUCTS, 'success');
         } else {
-            if (DISPLAY_CART == 'false') {
-                zen_redirect(zen_href_link(FILENAME_SHOPPING_CART));
+            if (DISPLAY_CART == 'false'  && ($allow_into_cart !== 'Y')) {
+                //zen_redirect(zen_href_link(FILENAME_SHOPPING_CART));
+                $messageStack->add_session('header', ($this->display_debug_messages ? 'FUNCTION ' . __FUNCTION__ . ': ' : '') . FAILURE_ADDED_TO_CART_PRODUCTS, 'error');
             }
         }
+        $exclude[] = 'action';
         if (is_array($parameters) && !in_array('products_id', $parameters) && !(strpos($goto, 'reviews') > 5)) $parameters[] = 'products_id';
-        zen_redirect(zen_href_link($goto, zen_get_all_get_params($parameters)));
+        zen_redirect(zen_href_link($goto, zen_get_all_get_params($exclude)));
     }
 
     /**
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -2071,11 +2071,10 @@ class shoppingCart extends base
         } else {
             if (DISPLAY_CART == 'false'  && ($allow_into_cart !== 'Y')) {
                 //zen_redirect(zen_href_link(FILENAME_SHOPPING_CART));
-                $messageStack->add_session('header', ($this->display_debug_messages ? 'FUNCTION ' . __FUNCTION__ . ': ' : '') . FAILURE_ADDED_TO_CART_PRODUCTS, 'error');
+                $messageStack->add_session('header', ($this->display_debug_messages ? 'FUNCTION ' . __FUNCTION__ . ': ' : '') . FAILED_TO_ADD_UNAVAILABLE_PRODUCTS, 'error');
             }
         }
         $exclude[] = 'action';
-        if (is_array($parameters) && !in_array('products_id', $parameters) && !(strpos($goto, 'reviews') > 5)) $parameters[] = 'products_id';
         zen_redirect(zen_href_link($goto, zen_get_all_get_params($exclude)));
     }
 
--- a/includes/languages/english.php
+++ b/includes/languages/english.php
@@ -336,6 +336,7 @@ define('ARIA_PAGINATION_','');
   define('SUCCESS_ADDED_TO_CART_PRODUCT', 'Successfully added Product to the cart ...');
 // only for where multiple add to cart is used:
   define('SUCCESS_ADDED_TO_CART_PRODUCTS', 'Successfully added selected Product(s) to the cart ...');
+  define('FAILURE_ADDED_TO_CART_PRODUCTS', 'The selected Product(s) are not currently available for purchase...');
 
   define('TEXT_PRODUCT_WEIGHT_UNIT','lbs');
 
--- a/includes/languages/english.php
+++ b/includes/languages/english.php
@@ -336,7 +336,7 @@ define('ARIA_PAGINATION_','');
   define('SUCCESS_ADDED_TO_CART_PRODUCT', 'Successfully added Product to the cart ...');
 // only for where multiple add to cart is used:
   define('SUCCESS_ADDED_TO_CART_PRODUCTS', 'Successfully added selected Product(s) to the cart ...');
-  define('FAILURE_ADDED_TO_CART_PRODUCTS', 'The selected Product(s) are not currently available for purchase...');
+  define('FAILED_TO_ADD_UNAVAILABLE_PRODUCTS', 'The selected Product(s) are not currently available for purchase...');
 
   define('TEXT_PRODUCT_WEIGHT_UNIT','lbs');
 


includes/languages/english.php
```diff
@@ -336,6 +336,7 @@ function zen_date_raw($date, $reverse = false) {
   define('SUCCESS_ADDED_TO_CART_PRODUCT', 'Successfully added Product to the cart ...');
 // only for where multiple add to cart is used:
   define('SUCCESS_ADDED_TO_CART_PRODUCTS', 'Successfully added selected Product(s) to the cart ...');
+  define('FAILURE_ADDED_TO_CART_PRODUCTS', 'The selected Product(s) are not currently available for purchase...');
```
drbyte, line 339: I'm puzzled by the name of this new constant: `FAILURE_ADDED_TO_CART_PRODUCTS`

Perhaps `FAILED_TO_ADD_UNAVAILABLE_PRODUCTS` might be more intuitive?

drbyte (changes requested):

mc12345678:
Isn't that line:
````php
if (is_array($parameters) && !in_array('products_id', $parameters) && !(strpos($goto, 'reviews') > 5)) $parameters[] = 'products_id';
````
Needed (with associated variable change(s)) if this method is called from outside of an `action=buy_now` page call to support the same logic that is executed in `includes/main_cart_actions.php`?

**Comment by drbyte** [comment]
> Isn't that line:
> 
> ```
> if (is_array($parameters) && !in_array('products_id', $parameters) && !(strpos($goto, 'reviews') > 5)) $parameters[] = 'products_id';
> ```
> 
> Needed (with associated variable change(s)) if this method is called from outside of an `action=buy_now` page call to support the same logic that is executed in `includes/main_cart_actions.php`?

If the next line after it is a zen_redirect() to another page, which aborts all current processing and loses all local variables state, is it needed?

File: includes/classes/shopping_cart.php [hunk 2]
```diff
@@ -2064,15 +2066,17 @@ public function actionBuyNow($goto, $parameters = [])
...
-                zen_redirect(zen_href_link(FILENAME_SHOPPING_CART));
+            if (DISPLAY_CART == 'false'  && ($allow_into_cart !== 'Y')) {
+                //zen_redirect(zen_href_link(FILENAME_SHOPPING_CART));
+                $messageStack->add_session('header', ($this->display_debug_messages ? 'FUNCTION ' . __FUNCTION__ . ': ' : '') . FAILURE_ADDED_TO_CART_PRODUCTS, 'error');
             }
         }
+        $exclude[] = 'action';
         if (is_array($parameters) && !in_array('products_id', $parameters) && !(strpos($goto, 'reviews') > 5)) $parameters[] = 'products_id';
-        zen_redirect(zen_href_link($goto, zen_get_all_get_params($parameters)));
+        zen_redirect(zen_href_link($goto, zen_get_all_get_params($exclude)));
```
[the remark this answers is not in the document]
proseLA: i agree with the conclusion; perhaps not with the logic.  prior to my change, how could the 2nd part of the if ever be true?  the function is `actionBuyNow` and `!in_array('products_id', $parameters)` would never be true.

includes/languages/english.php
proseLA: :+1:

drbyte (approved):

mc12345678:
> If the next line after it is a zen_redirect() to another page, which aborts all current processing and loses all local variables state, is it needed?

In order for this method to perform the expected actions, `$_GET['products_id']` must be set to the value of the product to be added to the cart. Regardless the success (or failure) of the product to be added to the cart, with that value still set (which in this revision is guaranteed to be true by the handling of `$exclude`) then the redirect will take the customer to the product information page (or whatever page is identified by `$goto` and/or what remains/is returned from `zen_get_all_get_params` but with `products_id` still set).  

As a result a customer may be on an accumulation page (where multiple product are shown), select buy now, and instead of returning back to that same page or whatever page is identified by `$goto` for potentially adding additional items from that "list" will be taken into the product page itself.  If this is the expected/desired direction to go, then https://github.com/zencart/zencart/blob/b241f01602d9024dbdd29da1fe940dc8468d98df/includes/init_includes/init_cart_handler.php#L32
should be changed to remove `products_id` along with the same removal in https://github.com/zencart/zencart/blob/b241f01602d9024dbdd29da1fe940dc8468d98df/includes/init_includes/init_cart_handler.php#L37

> the function is `actionBuyNow` and `!in_array('products_id', $parameters)` would never be true.

It would be true if the method were called from a load that does not include `includes/init_includes/init_cart_handler.php` with `$parameters` not including `products_id`. Loading of `includes/init_includes/init_cart_handler.php` is not a prerequisite for calling the method `$_SESSION['cart']->actionBuyNow`. Even here though, if `$parameters` *does* include `'products_id'`, it doesn't matter as `$exclude` doesn't include it nor does it include any other values of `$parameters` that were fed to the method.

drbyte:
> This makes the variable `$parameters` unnecessary for this method as it is not otherwise used anywhere in this method which makes the assignment of `$parameters` unnecessary

By "in this method" are you actually referring to `actionBuyNow`? Or the entire `shopping_cart` class? The code you quoted from `init_cart_handler.php` is not scoped only to the `actionBuyNow` method.

drbyte:
I too am a little uncomfortable with the removal of honoring `$parameters` and wholesale replacing it with `['action']`.

In looking at it again I'm still trying to understand the intended purpose.

mc12345678:
> By "in this method" are you actually referring to `actionBuyNow`? Or the entire `shopping_cart` class? The code you quoted from `init_cart_handler.php` is not scoped only to the `actionBuyNow` method.

Specifically the `actionBuyNow` method, but it calls to question the applicability of `products_id` being associated with `$parameters` as the store operation would be inconsistent between `actionBuyNow` and other operations.

> In looking at it again I'm still trying to understand the intended purpose.

In looking at the logic between a "page click" and direct use of the `actionBuyNow` method, it appears that the goal is generally to return the customer to either `FILENAME_SHOPPING_CART` or to the value of `$_GET['main_page']` which in a page click would be expected to be the current "page" (`products_all`, `specials`, `index` with `cPath` possibly set, etc..) In this way, a customer, having determined that they just want to buy this product right now and the store has been set to *not* go directly to the `FILENAME_SHOPPING_CART` would "keep" the customer at the current page where perhaps they could add some other product to their cart by selecting `buy now` again on some then current product.

drbyte:
> > In looking at it again I'm still trying to understand the intended purpose.
> 
> In looking at the logic between a "page click" and direct use of the `actionBuyNow` method, it appears that the goal is generally to return the customer to either `FILENAME_SHOPPING_CART` or to the value of `$_GET['main_page']` which in a page click would be expected to be the current "page" (`products_all`, `specials`, `index` with `cPath` possibly set, etc..) In this way, a customer, having determined that they just want to buy this product right now and the store has been set to _not_ go directly to the `FILENAME_SHOPPING_CART` would "keep" the customer at the current page where perhaps they could add some other product to their cart by selecting `buy now` again on some then current product.

I believe @proseLA had also mentioned in linked issues that the purpose there was to send a buy-now link in a newsletter email for quick add-to-cart for the item on promotion.
In that context there's nothing fancy required for the redirect handling. Ideally probably it should go to the Cart page so the buyer is encouraged to continue on to checkout.

proseLA:
> Setting the parameter to this function to the integer of `$_GET['products_id']` removes all attribute related information associated with that product and prevents the full use of
> 
> https://github.com/zencart/zencart/blob/b241f01602d9024dbdd29da1fe940dc8468d98df/includes/functions/functions_products.php#L780
> 
> 
> To fully evaluate the product being added from the provided information and instead to have to pull the value again from `$_GET['products_id']` instead of the passed or provided `$product_id` within that function.

so, if a product has attributes that can be added to the cart using the default values for those attributes, the allow to cart will fail from that function?  due to passing the full `product_id` to that function?  as opposed to the stripped out value of the `products_id` w/o attributes?

perhaps you can provide a url parameter list from the test data, where these unintended consequences happen, and i can explore further?

proseLA:
> > > In looking at it again I'm still trying to understand the intended purpose.
> > 
> > 
> > In looking at the logic between a "page click" and direct use of the `actionBuyNow` method, it appears that the goal is generally to return the customer to either `FILENAME_SHOPPING_CART` or to the value of `$_GET['main_page']` which in a page click would be expected to be the current "page" (`products_all`, `specials`, `index` with `cPath` possibly set, etc..) In this way, a customer, having determined that they just want to buy this product right now and the store has been set to _not_ go directly to the `FILENAME_SHOPPING_CART` would "keep" the customer at the current page where perhaps they could add some other product to their cart by selecting `buy now` again on some then current product.
> 
> I believe @proseLA had also mentioned in linked issues that the purpose there was to send a buy-now link in a newsletter email for quick add-to-cart for the item on promotion.
> In that context there's nothing fancy required for the redirect handling. Ideally probably it should go to the Cart page so the buyer is encouraged to continue on to checkout.

this is true.

mc12345678:
> Interesting point. When I refactored functions files I specified this param as int because that's how this function uses it. I hadn't considered/expected a hash id to be desired in an observer.
>
> Do you have a use-case for this in-the-wild? Or speaking just hypothetically?

I have not incorporated it yet, but would expect to use it in [Stock By Attributes](https://github.com/mc12345678/Stock_By_Attributes_Combined/) possibly in place of the [extra_cart_actions file](https://github.com/mc12345678/Stock_By_Attributes_Combined/blob/master/includes/extra_cart_actions/stock_by_attributes.php) currently in use. The notifier within the `zen_get_products_allow_add_to_cart` function is an opportune time/means to evaluate the applicability of the product proposed to be added to the cart (which would probably need to be incorporated into the other cart's `action...` methods or otherwise in the process flow).

Thing is that that plugin development has more or less come along with what has been incorporated into Zen Cart without trying to push the software too much.

mc12345678:
> I believe @proseLA had also mentioned in linked issues that the purpose there was to send a buy-now link in a newsletter email for quick add-to-cart for the item on promotion.
In that context there's nothing fancy required for the redirect handling. Ideally probably it should go to the Cart page so the buyer is encouraged to continue on to checkout.

Correct, an email that was sent at some point in the past (nature of an email of course :P) has been opened by a customer where the link contains a products_id of a product that could be in any state of existence/availability. The same "issue" was identified by providing a bogus `products_id` in the url.  That said though, a standard cart that offers such advertising (why wouldn't one) is subject to the same issue(s) as one navigating a site where the buy now option is offered.  So, the current software doesn't "recognize" whether the url originated from an email/outside source or from the active site.  Therefore if the site is set *not* to go to the shopping cart, which is one mindset in encouraging customers to add product to the cart, there is nothing that would provide a flag to the software to instead go to the shopping cart because of a link in the email... Meaning, more software and/or changes needed (even if education) to encourage sites that use the buy now type link to have their settings at least for buynow related operations (`includes/init_includes/init_cart_handler.php`) set to go to the shopping cart when a buy now operation is executed regardless of other normal operation...

Meaning, its possible to force buy now actions to go to the shopping cart and other add to cart operations *not* go to the shopping cart by a software change at https://github.com/zencart/zencart/blob/b241f01602d9024dbdd29da1fe940dc8468d98df/includes/init_includes/init_cart_handler.php#L27-L33

To adjust/set `$goto` to `FILENAME_SHOPPING_CART` within the else statement of that or to have the email that goes out (targeting only email related operations) to have parameters such as `main_page=shopping_cart&action=buy_now&products_id=x` set in the email.  It doesn't resolve the problem identified that basically the customer is incorrectly told of a successful add-to-cart of a product that didn't make it to the cart and a page that reflects nothing to be in the cart...

drbyte:
> It doesn't resolve the problem identified that basically the customer is incorrectly told of a successful add-to-cart of a product that didn't make it to the cart and a page that reflects nothing to be in the cart...

Hmmmm

mc12345678:
> so, if a product has attributes that can be added to the cart using the default values for those attributes, the allow to cart will fail from that function? due to passing the full `product_id` to that function? as opposed to the stripped out value of the `products_id` w/o attributes?
>
> perhaps you can provide a url parameter list from the test data, where these unintended consequences happen, and i can explore further?

The function will not fail in performing an evaluation of the overall ability of the product having an integer value of `products_id` to be able to be added to the cart; however, a product is specific down to all aspects that qualify the item as a product which includes attributes.  By passing the fully formed (hashed) value of `$_GET['products_id']` that information can be used to even look at existing items within the cart or to compare against other data about the possibility/acceptability of the chosen product to be added to the cart.  Setting it to an integer at this point is unnecessarily restrictive where the need for the use of the integer value is handled at the level where it is applicable: database lookup in `TABLE_PRODUCTS` and `TABLE_PRODUCT_TYPES` at least within `zen_get_products_allow_add_to_cart` and basically `TABLE_PRODUCTS_ATTRIBUTES` within `zen_requires_attribute_selection`. 

the function `zen_requires_attribute_selection` contains a notifier within it that can be used to have the function return whatever is desired/expected based on the provided `products_id` information, for example a product that has one option name that normally has multiple option values within it, but only one of the option values has available stock could/would be a product that "fits" the one attribute condition and could be "auto-added" to the cart. 

Further, if the function `zen_get_products_allow_add_to_cart` is called here with the integer value and not elsewhere/everywhere, then there is nothing within that function that identifies the existence nor anything that needs `$_GET['products_id']` to be set such that either it would become complicated to identify whether that value should be pulled/checked or if the data provided is accurate in being an integer instead of a product that really should have attributes but none were provided...

Basically casting the value to an integer at this point does not prevent standard operation of a store, but requires one to modify code to reverse something that appears (in absence of further discussion/explanation) unnecessary and does not support further use/expansion off of the Zen Cart code instead of modifying it.

proseLA:
> > It doesn't resolve the problem identified that basically the customer is incorrectly told of a successful add-to-cart of a product that didn't make it to the cart and a page that reflects nothing to be in the cart...
> 
> Hmmmm

this is half true.  the customer is told of a successful add-to-cart.  and the item is added to the cart; but it is removed when one displays the cart.  so, you are correct, it does not resolve that problem.

i noticed this in my initial testing.

> the addition of a notifier in the `zen_get_products_allow_add_to_cart` function is great, and was the exact idea that i had. i think the function should check for an inactive product as it is already hitting the products table. it seems that the function might be able to do more, prior to handing over to the observer.

my initial comments on this PR referenced this exact issue.  for my own purposes, the addition of the observer allowed me to override at that time.  but i was aware of this situation at the time.

when doing this PR, i attempted to stay within the parameters of the code that came before me; hopefully not upsetting too much.  it does seem that i was not successful in that goal.

mc12345678:
FYI, there is a wrinkle that still exists (not specifically caused by this PR chain)... It appears that while on the shopping cart page, if a product is shown with a buy now button where the product is available (by whatever checks are to be performed), the buy now button clicked, and the site is set to not display the shopping_cart, then the product would be added to the cart, but no notification of success would occur (same action as previously) though the shopping_cart would be displayed because the `main_page` value would have been `FILENAME_SHOPPING_CART` from back in `includes/init_includes/init_cart_handler.php` causing `$goto` to support going back to the shopping_cart. Guess that's something to address in a separate issue.

This is a result of this section of code and the above logic conditions:
https://github.com/zencart/zencart/blob/f4e01a44307315eca1ef4027ae4f9cf1aed6c052/includes/classes/shopping_cart.php#L2074-L2081

drbyte:
> It appears that while on the shopping cart page, if a product is shown with a buy now button where the product is available (by whatever checks are to be performed), the buy now button clicked

If they're on the shopping_cart page, where are they seeing a Buy Now button to click?

> FYI, there is a wrinkle that still exists (not specifically caused by this PR chain)
> Guess that's something to address in a separate issue.

Yes let's keep issues separate from each other. Feel free to open a new issue describing it in detail, including a use-case where this would occur.

PR #3955, opened 2020-09-15
Template and function updates to pricemanager

Zen4All:
See #3928

Also added a function to datepicker.php to prevent repetition of code. This could also be used in other files saving dates to database (like product_info, specials and featured).


admin/products_price_manager.php
```diff
@@ -106,54 +103,37 @@
         $master_categories_id = $_POST['master_categories_id'];
       }
 
-      $products_date_available = ((!isset($_POST['product_start']) || zen_db_prepare_input($_POST['product_start']) == '') ? '0001-01-01' : zen_date_raw($_POST['product_start']));
+      $products_date_available = zen_prepare_date($_POST['products_date_available']);
 
-      $specials_date_available = ((!isset($_POST['special_start']) || zen_db_prepare_input($_POST['special_start']) == '') ? '0001-01-01' : zen_date_raw($_POST['special_start']));
-      $specials_expires_date = ((!isset($_POST['special_end']) || zen_db_prepare_input($_POST['special_end']) == '') ? '0001-01-01' : zen_date_raw($_POST['special_end']));
+      $specials_date_available = zen_prepare_date($_POST['specials_date_available']);
+      $specials_expires_date = zen_prepare_date($_POST['specials_expires_date']);
```
mc12345678, line 109: Perhaps `specials_date_expires`? I can see how code could be "easily" written to consider these three sets of dates using an array or perhaps two, cycling through the prefix and checking/using the suffix of `available` or `expires`.

````php
$prefix = [
    'products_date',
    'specials_date',
    'featured_date',
];
$suffix = [
    'available',
    'expires',
];
foreach ($prefix as $pre) {
    foreach ($suffix as $suf) {
        if (evaluation_of_posted_data($_POST[$pre & '_' & $suf])) {
            // do something and or break or break 2;
        }
    }
}
````

FYI, my first concern was whether the sanitizer needed any updating; however, these variables are not directly identified in the sanitizer.

Comment (Zen4All) [reply admin/products_price_manager.php:109]: Maybe this could be done in a next separate PR. I am currently working through all admin files, and where needed, try to on update what was mentioned in #3982. Although that is not always easy.

File: admin/products_price_manager.php [hunk 2]
```diff
@@ -106,54 +103,37 @@
...
-                  product_is_call = :isCall:,
-                  products_quantity_mixed = :qtyMixed:,
-                  products_priced_by_attribute = :pricedByAttr:,
-                  products_discount_type = :discType:,
-                  products_discount_type_from = :discTypeFrom:,
-                  products_price_sorter = :discPriceSorter:,
-                  master_categories_id = :masterCatId:,
-                  products_mixed_discount_quantity = :discQty:
+                  products_status = " . (int)$_POST['products_status'] . ",
+                  products_quantity_order_min = " . (int)$_POST['products_quantity_order_min'] . ",
```
mc12345678, line 123: Huh, my previous attempt at commenting here didn't "stick" it went something like:

wooohhhh, wait a minute, why is products_quantity_order_min, products_quantity_order_units, and at least products_quantity_order_max being forced to an integer??? What about units of less than 1 (0.01 or 0.5)? What about quantities of any of these being partial values? Even say 1.5?

Why are any of these type casts that are covered by $db->bindVars being removed/modified? 

Even the assignment of say products_date_available from using the string casting where if any portion of the string contains `NULL` to the new validation that ignores if say there are spaces around the 'null' and also don't account for capitalization? Further such use of $db->bindVars provides database sanitization for text which is omitted in this new rewrite...

admin/products_price_manager.php
mc12345678: Okay, but where in that issue is the revision of these fields identified as to be modified in such a way? Currently they are/were like "object" & "_" & "position" where "object" = (product, featured, or special) and "position" = ("start" or "end").

Zen4All: It does not matter if it is set or not, the new function takes care of that. And this is also how it is already done in collect_info for the products. I merely copied it, and put into a function.

admin/products_price_manager.php
Zen4All: You are right, I missed the fact they should be floats. so I will change that. I don't like using the bindvars where it is in my vision unnecessary where a simple cast to integer or float can be done.

admin/products_price_manager.php
mc12345678: > It does not matter if it is set or not, the new function takes care of that. 

In order to get to the new function, php has to recognize the variable that is being passed. If `$_POST[$variable_name]` does not exist, then in stricter operations this raises an issue that it is not set and nothing (under current operations) within the function will correct/address attempting to provide a non-existent value.

Would be another opportunity to use the null coalesce operator to address this:
````diff
-$specials_expires_date = zen_prepare_date($_POST['specials_expires_date']);
+$specials_expires_date = zen_prepare_date($_POST['specials_expires_date']??null);
````

By the way, why are these dates being deleted for all dates upon saving of the information. (I note that the same thing is now being done in featured #3930 and specials #3934.) If this is the route to go (removing/modifying information even on an update), two things, shouldn't they be logged in the admin log, and shouldn't the same occur when a product is made active as a result of such date passing? I personally don't think that this should be happening, but I also am interested in understanding why that bit of history is being forced to be lost.  For example, I know that in a site I manage we have a product that is enabled, made a feature, etc... each year on a date and then disabled or removed from that additional grouping on a date.  If/when going to update any aspect of that information, I would have to on "first" edit/update set/reset that information so as not to lose it.... That would then also run an issue with say featured/specials that have a start/end date where if it were edited in the middle of an active period would have to have a start date that is greater than the end date so that the "record" could be kept, but that defeats logic of the sequence and wouldn't align with activating the feature/special...

Comment (Zen4All) [reply admin/products_price_manager.php:109]: > In order to get to the new function, php has to recognize the variable that is being passed. If $_POST[$variable_name] does not exist, then in stricter operations this raises an issue that it is not set and nothing (under current operations) within the function will correct/address attempting to provide a non-existent value.

I have tested this code in all kinds of scenarios, and so far nor errors or warnings, so if you can produce them plese let me know.

>By the way, why are these dates being deleted for all dates upon saving of the information. (I note that the same thing is now being done in featured #3930 and specials #3934.) If this is the route to go (removing/modifying information even on an update), two things, shouldn't they be logged in the admin log, and shouldn't the same occur when a product is made active as a result of such date passing? I personally don't think that this should be happening, but I also am interested in understanding why that bit of history is being forced to be lost. For example, I know that in a site I manage we have a product that is enabled, made a feature, etc... each year on a date and then disabled or removed from that additional grouping on a date. If/when going to update any aspect of that information, I would have to on "first" edit/update set/reset that information so as not to lose it.... That would then also run an issue with say featured/specials that have a start/end date where if it were edited in the middle of an active period would have to have a start date that is greater than the end date so that the "record" could be kept, but that defeats logic of the sequence and wouldn't align with activating the feature/special...

I did not add anything to delete dates. They are updated in a query not deleted.

Comment (mc12345678) [reply admin/products_price_manager.php:109]: >I have tested this code in all kinds of scenarios, and so far nor errors or warnings, so if you can produce them plese let me know.

Agree no errors nor warnings, but an issue (notice) is raised.
````php
<?php

$array = [];

echo $array['fun'];

function this_test($data = '') {
  echo ":" . $data . ":";
}

this_test($array['test2']);
````

Run on php 7.0.1 and 7.4.7 this results in:
````
Notice:  Undefined index: test in [...][...] on line 5

Notice:  Undefined index: test2 in [...][...] on line 11
::
````

> I did not add anything to delete dates. They are updated in a query not deleted.

In the update section of `products_price_manager.php`, each date is sent through `zen_prepare_date` which is the following code:
https://github.com/zencart/zencart/blob/836b509398676a9081dded5cabfcdf87ccc1d13e/admin/includes/functions/datepicker.php#L41-L59

The second to last line of functional code evaluates the current date as formatted to `Y-m-d` against the date that has been "entered" or provided.. If the current date is less than the provided date, then the provided date is kept (`$date = $date`). If the current date is greater than or equal to the provided date, then the date is set to `'null'` (`$date = 'null'`)... The date (`$date`) is then returned back and that value is what is stored in the update query...

So, yes, the date gets removed/deleted if the provided (`$_POST`) date is less than or equal to the current date and yes that value is updated into the field.  Historically this has been the case for the future availability of a product, but not for specials and featured.

Comment (Zen4All) [reply admin/products_price_manager.php:109]: Ok, I now see what you mean. That is indeed unwanted behavior. Back to the drawing-board 😄

Comment (mc12345678) [reply admin/products_price_manager.php:109]: Pssst... Thinking that perhaps a parameter in `zen_prepare_date` defaulting to leave the date alone and requiring some value for the current process path.

File: admin/products_price_manager.php [hunk 3]
```diff
@@ -106,54 +103,37 @@
...
-      $sql = $db->bindVars($sql, ':orderMax:', $_POST['products_quantity_order_max'], 'string');
-      $sql = $db->bindVars($sql, ':isFree:', $_POST['product_is_free'], 'integer');
-      $sql = $db->bindVars($sql, ':isCall:', $_POST['product_is_call'], 'integer');
-      $sql = $db->bindVars($sql, ':qtyMixed:', $_POST['products_quantity_mixed'], 'integer');
-      $sql = $db->bindVars($sql, ':pricedByAttr:', $_POST['products_priced_by_attribute'], 'integer');
-      $sql = $db->bindVars($sql, ':discType:', (isset($_POST['products_discount_type']) ? $_POST['products_discount_type'] : 0), 'integer');
-      $sql = $db->bindVars($sql, ':discTypeFrom:', (isset($_POST['products_discount_type_from']) ? $_POST['products_discount_type_from'] : 0), 'integer');
-      $sql = $db->bindVars($sql, ':discPriceSorter:', $products_price_sorter, 'string');
-      $sql = $db->bindVars($sql, ':masterCatId:', $master_categories_id, 'integer');
-      $sql = $db->bindVars($sql, ':discQty:', (isset($_POST['products_mixed_discount_quantity']) ? $_POST['products_mixed_discount_quantity'] : 0), 'integer');
```
drbyte, line 155: I'm not sure I understand why you removed the `bindVars` implementation here.

What problem did it create that made you decide to remove it completely?

Zen4All: I removed it because all are simple integers or floats, which can be casted directly. No need to use ```$db->bindVars``` which does the same with a detour. Also the strings are floats

drbyte: Thank you for your efforts. I wish you hadn't done that. I will probably change it back because it is cleaner to read a mass group of bindVars calls than it is to read inline casts interpolated with query strings. I'll have a chat with @zcwilt first.

admin/products_price_manager.php
```diff
@@ -407,157 +366,137 @@
...
-                    grossValue = grossValue * ((taxRate / 100) + 1);
-                }
+              const taxRate = getTaxRate();
+              let grossValue = $('input[name="products_price"]').val();
+              if (taxRate > 0) {
+                grossValue = grossValue * ((taxRate / 100) + 1);
+              }
 
-                document.forms["new_prices"].products_price_gross.value = doRound(grossValue, 4);
+              $('input[name="products_price_gross"]').val(doRound(grossValue, 4));
```
drbyte, line 459: I'm curious: You converted some of these to jQuery syntax, but left others alone. 
Was there a reason?

drbyte (changes requested):

PR #3961, opened 2020-09-18
Query is misformed, giving blank page

scottcwilson:


Zen4All (commented):
Actually that line should move up right below the ```LEFT JOIN " . TABLE_PRODCTS_DESCRIPTION . "```

scottcwilson:
Not sure what you mean.  It's part of a where clause.  The code I submitted has been tested and works.

drbyte:
The point @Zen4All was making is that the language field is part of query join between the products and products_description tables, and he was suggesting that therefore it should be part of the JOIN criteria (eg the USING statement).

However... while that's not wrong, it changes the JOIN from being based on a single field that shares the same name between the current table and the one listed before it in the query which can use the more terse USING syntax, thus the change requires changing the syntax from USING to ON and then specifying the field names from both tables, plus the additional criteria.

There's an expressiveness that comes from reading the code when written that way, but I'm not convinced yet that it's the best way to write it. IMO putting such "where" criteria in a JOIN clause isn't actually telling the db how to "join" the tables; rather it's specifying filtering criteria, which is indeed really more for a "where" clause.
In the end, the db's query-optimizer tends to sort it out anyway. 

In practice I've seen leaving the "where" criteria in a WHERE clause be slightly more reliable. Perhaps it's more a factor of whether/what indexes are present for the appropriate tables. By "more reliable" I mean I've seen in a couple cases that moving  the "AND field='value'" part into the JOIN/ON clause actually bring back no results when in fact there were results that suited the criteria. Again, while I don't recall checking on indexes or other factors, putting the "AND field='value'" back into a WHERE clause instead of the JOIN/ON resolved it.

PR #3962, opened 2020-09-18
Fix explanatory text in TEXT_INFO_MASTER_CATEGORY_CHANGE

scottcwilson:
- This description suggests pricing can be modified in a linked category, which is not true.

- It is confusing to put the instructions for changing the table layout on this page in the middle of instructions for adding linked categories.


torvista:
> This description suggests pricing can be modified in a linked category, which is not true

This text is legacy....and I had always assumed (but never used) that Salemaker could apply a sale discount to a category: so any products (linked or master) in that category would be discounted.
On cursory testing I find that 
a) If a selected salemaker category holds a master product, the same discount is shown on linked products (in other categories).
b) If a selected salemaker category holds a linked product, the discount is not applied to that linked product, nor its master. i.e. it is ignored.

Not what I expected...

scottcwilson:
Interesting but unrelated to this change.  Salemaker behavior on linked products should be written up as a separate issue.

drbyte:
If the other issue needs addressing, please do create an issue for it. Or if a PR is self-evident, a PR.

PR #3965, opened 2020-09-19
Change option-name-values screen pagination URLs and remove page=1.

mc12345678:
Ensures that regardless the parameter assignments that the end of the
url is appropriate (e.g. doesn't end with '&').

This also removes the use of page=1 to eliminate duplication of
address meaning (no page parameter and page=1 identifying the same thing)


mc12345678 (commented):
Incorporated discussed change.

PR #3966, opened 2020-09-19
Ensure value provided is of the database's data type

mc12345678:


lat9 (commented):
Are you saying that the value just pulled from the database might not be of the correct type?

drbyte:
Questions:
1. What brought about the desire to submit this?  Was there an error message? A data problem? 
What were you exploring that caused this to come to the fore?

2. Why the v157 branch? (for all your recent PRs)?

mc12345678:
> Questions:
>
> What brought about the desire to submit this? Was there an error message? A data problem?
What were you exploring that caused this to come to the fore?
>
> Why the v157 branch? (for all your recent PRs)?

1) Well, it came from a series of events/for consideration; however, having just run a separate test, it does not appear that this particular change was needed for the reason perceived... So, I'll try to explain and then likely close it:
In the process of working primarily towards #3964, I caused an action to insert 0 value type data into the `products_options_values` table, which had me thinking about how `0` is often handled and/or lost when not cast to an integer when trying to combine string with variable.  Seeing this variable assignment so close to the code being modified seemed irresponsible not to at least pose the suggested change (of course didn't provide much background on it).  So, today, I created a test where I read from a database a value from one of the tables where the field was an integer and the value was 0, then created a string of text with the retrieved field value (without casting) and the `0` as a string datatype was displayed. So, my choice of casting the value was not so dependent on its recent retrieval (as code could be added between the two points to adjust the value), but more of to ensure that if an unexpected 0 value was to be incorporated, that it would be properly assigned and not cause a mysql error... But, again, this doesn't appear to be an issue in practice, so not really necessary here.

2) Because of a not too long ago direct request to focus on 1.5.7 understanding that perhaps a subsequent release would follow...

drbyte:
Thanks for the context.

I'm satisfied that this is fine for v1.5.8 because it's not a critical bug that needs to be backported into v1.5.7.

I'll commit it into the 158 branch.

PR #3968, opened 2020-09-21
Add grid-layout option for product-listing.

drbyte:
Controlled by 1 admin config setting: number of columns per row. When set to 1 (default), shows in "rows" mode. Otherwise, works as a grid.

This incorporates the bulk of the features of the "Column-Grid Layout Product Listing" mod, but without the customer-selectable display option.

Screen Shot 2020-09-20 at 10 03 55 PM

Screen Shot 2020-09-20 at 10 07 05 PM


scottcwilson:
Beautiful!

drbyte:
While normally a "feature change" like this wouldn't be pushed into a "sub-release" (eg: 1.5.7a), I'm targeting this for 1.5.7a because it's a sensible feature, and it retires the need to update a popular plugin, and it is easily integrated (or ignored) for people with an already-functioning template (eg: if they don't merge this part, that's fine; and if they do merge it but don't set a columnar value in the config, it will not have negative side-effects).

dbltoe:
I've been doing it through the css alone.  This is WAY better and I concur with the 157a

PR #3971, opened 2020-09-21
Missed HTML defines for emails

scottcwilson:
To reproduce the issue, update an order from the Admin Order Details page and observe the HTML email.  The two values         `$EMAIL_SALUTATION` and `$EMAIL_ORDER_UPDATE_MESSAGE` are not filled in. 

image


scottcwilson:
Suggested for backport.

drbyte:
This seems new. When did these symptoms start? Surely people have been screaming about this if it was a problem in previously released versions?

torvista:
It works in my local ZC157 (release version).
Relevant PR here:
https://github.com/zencart/zencart/pull/2850

scottcwilson:
Did you use the repro steps I described in the first post?  Also you must *not* have an order message to see the order message issue.

scottcwilson:
> This seems new. When did these symptoms start?  

This is 1.5.7

torvista:
Made an order, updated status from both order listing and order details. Checked for two languages.
I note that my vanilla html email is not like yours (broken logo as it's local):
Clipboard01

scottcwilson:
Please check for the block 

  <div class="order_message">
    $EMAIL_ORDER_MESSAGE
  </div>

in email/email_template_checkout.html
It looks like it's missing, which is why you don't see the second error.

scottcwilson:
Well, this happened once and now I can't repro it.  Closing.

scottcwilson:
Root cause was that the order was updated using Super Orders, not the built-in order flow management code.

PR #3977, opened 2020-09-26
Simplify private testing for templates

scottcwilson:
Current code does not work.  This works as tested with 

http://localhost:8888/gh_demo_158/index.php?main_page=index&t=classic

then 

http://localhost:8888/gh_demo_158/index.php?main_page=index&t=off

Suggested for backport with low priority.

### DIFF
--- a/includes/init_includes/init_templates.php
+++ b/includes/init_includes/init_templates.php
@@ -33,19 +33,14 @@ $template_dir = $result->fields['template_dir'];
  * Allow admins to switch templates using &t= URL parameter
  */
 if (zen_is_whitelisted_admin_ip()) {
-    $templates = array();
-    foreach($result as $row) {
-        $templates[] = $row['template_dir'];
-    }
-    // check if a template override was requested and that it matches an available choice and it exists on the filesystem
-    if (isset($_GET['t']) && in_array($_GET['t'], $templates, true) && file_exists(DIR_WS_TEMPLATES . $_GET['t'])) {
+    // check if a template override was requested and that the template exists on the filesystem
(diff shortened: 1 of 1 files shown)


scottcwilson:
Instructions on using this feature: 
https://docs.zen-cart.com/user/template/private_testing/

includes/init_includes/init_templates.php
```diff
@@ -33,19 +33,14 @@
  * Allow admins to switch templates using &t= URL parameter
  */
 if (zen_is_whitelisted_admin_ip()) {
-    $templates = array();
-    foreach($result as $row) {
-        $templates[] = $row['template_dir'];
-    }
```
drbyte, line 39: Excluding the check of this array means the only check for templates is against the filesystem.

I intended it to require that the template be registered in the admin in order for catalog-side to be able to switch to it.

scottcwilson: This seems like a really confusing and hard to figure out additional step.  We already have the protection that only admins can do this - why not make it easy for them?

drbyte: I'll concede.

Please also remove `and that it matches an available choice` from the comment.

And change the commit and PR title from "Fix..." to "Simplify..." or "Change...".

scottcwilson: Ugh sorry for the double commit - please squash.

drbyte: I'd hoped you would have updated the commit title as well (part of the amend+forcepush), not just the PR title.

scottcwilson:
Please consider for backport - this is a great feature I was unaware of prior to working on the doc page reference above.

Issue #2397, opened 2019-06-02
MySql 5.7+ and date/datetime dates (again)


lat9:
Just reading through the MySQL 5.7 documentation on date/datetime fields (https://dev.mysql.com/doc/refman/5.7/en/datetime.html) and I think we might still have some issues.  From that documentation:
```
 The DATE type is used for values with a date part but no time part. MySQL retrieves and displays DATE values in 'YYYY-MM-DD' format. 
The supported range is '1000-01-01' to '9999-12-31'.

The DATETIME type is used for values that contain both date and time parts. MySQL retrieves and displays DATETIME values in 'YYYY-MM-DD hh:mm:ss' format. 
The supported range is '1000-01-01 00:00:00' to '9999-12-31 23:59:59'. 
```
There are still (as of zc156b, anyway) a slew of date-related fields that use a default date of `0001-01-01`, which will (from the documentation) cause issues.

Zen4All:
I have been using 5.7 for years now, and it never gave any issues with 0001-01-01

lat9:
That's great, @Zen4All, I'm just pointing to the documentation that says differently.

Zen4All:
I read the link above to, and did some searching on internet, but could not find any good references to 0001-01-01. I am not sure why we used 0001-01-01, but I think I copied it form the zen cart code for my mods.
Maybe someone made a typo in the past?

drbyte:
We inherited using `0000-00-00` from osc.
When MySQL started forcing strict no-zero-dates rules, they still had not yet documented an acceptable alternative to zero-dates, and we did a lot of testing to be sure that `0001-01-01` would work without negative side-effects.
Now it seems MySQL has gone back and defined what they wish to be seen as acceptable values. It's unfortunate that they chose something different, particularly when nobody's going to use either of those choices for "real" data anyway.

Anyway ...

The better approach would be to use `null` instead of 0000-00-00 or 0001-01-01 (or 1000-01-01 IMO), because the whole point of these "old" dates is to indicate that "nothing has been supplied". (And an empty string is the wrong "type".)
Unfortunately converting all the code to handle null dates correctly is also a big undertaking, and affects addons since many mods deal with data concerning dates.

I'd welcome a series of PRs that handle implementing null dates, if the PRs are done feature-by-feature (ie: sale dates in one, coupon dates in another, birthdates in another, specials, then featured products, etc). This will help plugin authors see what's changed related to their code.

drbyte:
[...]

PR #3978, opened 2020-09-26
[158] Fix salemaker disable logic

scottcwilson:
Current code always disables any sale with a start or stop date of 0001-01-01.

### DIFF
--- a/includes/functions/salemaker.php
+++ b/includes/functions/salemaker.php
@@ -32,15 +32,15 @@ function zen_expire_salemaker()
 {
     global $db;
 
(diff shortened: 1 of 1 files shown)


mc12345678:
If in the middle of "correcting" the logic, may I suggest also considering the case(s) of dates being NULL (or as applicable NOT NULL)?

scottcwilson:
These two dates are defined as NOT NULL.

mc12345678:
> These two dates are defined as NOT NULL.

https://github.com/zencart/zencart/issues/2397

PR #3980, opened 2020-09-27
Remove <br> tag from email error message

scottcwilson:
The line break means the message looks like this: 

```
--> PHP Notice: Email Error: SMTP connect() failed. https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting<br /> in /home/username/zencart/includes/functions/functions_email.php on line 377.
```

No need for the line break in the middle.


drbyte:
The `<br>` is to create a new line when the error is rendered as HTML.
We could add an actual line-break `\n` for it to render as such in text-only mode.

PR #3983, opened 2020-09-29
zc157: Ask-a-Question, correct PHP Notices when no 'pid' supplied ...

lat9:
- If no $_GET['pid'], redirect to the site's main page.  Otherwise, various PHP notices are thrown by the template since the `$product_details` isn't set.
- Simplification of antispam conditional
- Pre-fill customer information only if logged in and (added) not in guest checkout.

### DIFF
--- a/includes/modules/pages/ask_a_question/header_php.php
+++ b/includes/modules/pages/ask_a_question/header_php.php
@@ -7,10 +7,16 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: DrByte 2020 Jun 23 Modified in v1.5.7 $
  */
-
-
 $zco_notifier->notify('NOTIFY_HEADER_START_ASK_A_QUESTION');
-if (isset($_GET['pid'])) {
+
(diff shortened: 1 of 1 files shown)


mc12345678:
Curious, why not send to `contact_us` instead of the main page if pid is not provided? Perhaps with consideration of whether the `contact_us` page is activated or not? And/or consideration that the original Ask-a-question plugin used `products_id` as the parameter which was modified to `pid` when incorporated into ZC?

lat9:
If the pid's not provided, then someone's modifying the browser address-bar to try to fuss with the site.  Whatever page is redirected to makes no difference to me, it's just important to not pass through to the template since no information will be gathered.

I've no idea why the zc-incorporation of the plugin version chose to use `pid` instead of `products_id`, but my feeling is that now that Zen Cart's placed a stake-in-the-sand that the name-change should stay.

drbyte:
> I've no idea why the zc-incorporation of the plugin version chose to use `pid` instead of `products_id`,

That was to reduce the URL length

drbyte:
> Curious, why not send to `contact_us` instead of the main page if pid is not provided?

That could be a good option. 
A PR exploring that approach is welcome.

PR #3984, opened 2020-09-29
[158] Ask-a-Question, correct PHP Notices when no 'pid' supplied ...

lat9:
- If no $_GET['pid'], redirect to the site's main page.  Otherwise, various PHP notices are thrown by the template since the `$product_details` isn't set.
- Simplification of antispam conditional
- Pre-fill customer information only if logged in and (added) not in guest checkout.


scottcwilson:
Would it make sense to change `pid` to `products_id` to be more consistent with the rest of the system?

lat9:
Changing from `pid` to `products_id` will add one more thing to check on upgrades if a site's got a template-override version of the page's template.  As such, I don't think it's worth the cosmetic change.

drbyte:
> Would it make sense to change `pid` to `products_id` to be more consistent with the rest of the system? (In the 158 PR only.)

The use of `pid` was intentional.

PR #3986, opened 2020-09-30
Remove greater than sign from query name

### LINKED ISSUE #3798: Greater than symbol in query_builder table causes trouble
This gets sanitized to `&gt;` which doesn't match any row in query_builder, which causes an empty query to be passed to $db->Execute.   

To reproduce in 1.5.7, create a newsletter, then press Send and select the audience "Dormant Customers (>3months) (Subscribers)".  

There are several ways we could fix this but it seems to me that the easiest way is simply to update the query_builder table not to contain a greater than sign.

scottcwilson:
Fixes #3798


drbyte:
This can be backported with a separate PR.

PR #3995, opened 2020-10-03
Update email transport method description to include GMail

scottcwilson:
When I did #3923 I forgot to add this fix to the upgrader.


scottcwilson:
I may have deferred doing this under the auspices of #3820, but I just had to do this update for a client, so we may as well merge this change.

PR #3997, opened 2020-10-03
Add count to dropdown menu for referral reporting

scottcwilson:


drbyte:
Is there any value in also adding ` ORDER BY count DESC`?

scottcwilson:
That way you see the most common ones first in case there's a huge list.

PR #3998, opened 2020-10-03
Fix description of CUSTOMERS_REFERRAL_STATUS

scottcwilson:


zc_install/sql/install/mysql_zencart.sql
```diff
@@ -2535,7 +2535,7 @@ INSERT INTO configuration (configuration_title, configuration_key, configuration
...
 
-INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Customers Referral Status', 'CUSTOMERS_REFERRAL_STATUS', '0', 'Customers Referral Code is created from<br />0= Off<br />1= 1st Discount Coupon Code used<br />2= Customer can add during create account or edit if blank<br /><br />NOTE: Once the Customers Referral Code has been set it can only be changed in the Admin Customer', '5', '80', 'zen_cfg_select_option(array(\'0\', \'1\', \'2\'), ', now());
+INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Customers Referral Status', 'CUSTOMERS_REFERRAL_STATUS', '0', 'Customers Referral Code is created from<br />0= Off<br />1= 1st Discount Coupon Code used<br />2= Customer can add during create account or edit if blank<br /><br />NOTE: Once the Customers Referral Code has been set, it can only be changed in the Admin', '5', '80', 'zen_cfg_select_option(array(\'0\', \'1\', \'2\'), ', now());
```
drbyte, line 2539: - I don't think a comma is needed here. It's still the same thought. 
- perhaps `in the Admin` should be `by the Administrator`? I'm thinking here that the reader will not wonder "where in the Admin", but instead the phrase " by the Administrator" sends their brain into "I will have to do it" rather than "where would I do it". I know, tiny and subtle, but perhaps slightly more clear.

Thoughts?

scottcwilson: Sounds good.  I just didn't want that weird trailing "Customer" word.

PR #4009, opened 2020-10-07
Prevent page parameter when link goes to page=1

### LINKED ISSUE #4008: Page=1 inconsistently displays on browser and in canonical
Stems from this conversation: https://www.zen-cart.com/showthread.php?227157-Canonical-link-now-keeps-the-page-parameter-(zc156-too)

mc12345678:
Prevents display/inclusion of `page=1` in the generated web address(es).
This is to remove the added page parameter into the page address as a
result of this being eventually fed to the canonical link.

This change also addresses the parameter and its appearance in the
canonical link if manually added to the web address.  This does a full
removal from the canonical address when page=1 or if the parameter
is non-numeric.

Fixes #4008

### DIFF
--- a/includes/classes/split_page_results.php
+++ b/includes/classes/split_page_results.php
@@ -107,10 +107,10 @@ class splitPageResults extends base {
 
     $class = '';
 
-    if (zen_not_null($parameters) && (substr($parameters, -1) != '&')) $parameters .= '&';
+    if (zen_not_null($parameters) && (substr($parameters, -1) != '&') && ($this->current_page_number > 1)) $parameters .= '&';
 
     // previous button - not displayed on first page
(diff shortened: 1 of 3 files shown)


lankeeyankee:
I found the init_sanitize casting page to an integer breaks URL rewriting.

drbyte:
> I found the init_sanitize casting page to an integer breaks URL rewriting.

With which URL rewriter?

lankeeyankee:
SSU

mc12345678:
Does the problem go away if the `int` cast variable is then cast "back" to a `string`?

$_GET['page'] = (string)(int)$_GET['page'];

lankeeyankee:
no, that does not fix it

drbyte:
> SSU

As in ... https://www.zen-cart.com/downloads.php?do=file&id=1754 ? (Just checking, as I see it hasn't been updated in a very long time)

PR #4010, opened 2020-10-07
Prevents php notice using PluginManager and a plugin has an upgrade a…

mc12345678:
…vailable.

I haven't been able to fully figure out why this is happening or rather
why the __construct has `tableDefinition['configBox']` which doesn't seem
to be used and that these methods don't reference that particular key, but
when I tried to substitute the `configBox` value(s) in these modifications
then the pluginmanager would not move forward in providing information
or allowing the next action. Therefore, an empty array seemed to be the only
fix to support moving to the next step, although I haven't seen the update
operation successfully performed without first uninstalling and then reinstalling.

### DIFF
--- a/includes/classes/TableViewControllers/BaseController.php
+++ b/includes/classes/TableViewControllers/BaseController.php
@@ -229,12 +229,12 @@ class BaseController implements TableViewController
 
     public function getTableConfigBoxHeader()
     {
-        return $this->tableDefinition['header'];
+        return $this->tableDefinition['header'] ?? [];
     }
 
     public function getTableConfigBoxContent()
     {
-        return $this->tableDefinition['content'];
+        return $this->tableDefinition['content'] ?? [];
     }
 
     public function getSplitPage()


drbyte:
@zcwilt will have to chime in.

I wonder if the intent was to "make sure the `header` and `content` keys were initialized", after receiving the otherwise-unvalidated `$tableDefinition` parameter.

https://github.com/zencart/zencart/blob/34d4b5cce778751dbecd126f8b5c57d120e9e312/includes/classes/TableViewControllers/BaseController.php#L29-L38

If indeed the `configBox` subarray is never actually used, then these other keys could be initialized-if-not-present in another way. (It could be a if-is-not-set call or an array merge using the `+` syntax instead of `array_merge()` since the `+` approach skips merging already-present keys.)

zcwilt:
I think in it's original incarnation the class did use `$tableDefinition['configBox']['header']` and `$tableDefinition['configBox']['content']`
to match how legacy code built those config boxes. However on subsequent refactoring it was simplified. So the
`$this->tableDefinition['configBox'] = ['header' => [], 'content' => []];`  is no longer needed although could be replaced with separate initializations of `$tableDefinition['configBox']['header']` and `$tableDefinition['configBox']['content']`

zcwilt (approved):

PR #4012, opened 2020-10-07
Gray out update button on rows which are not updatable on shopping cart

### LINKED ISSUE #4006: Update icon should not show on shopping cart on items with fixed quantity
Screen Shot 2020-10-06 at 3 12 01 PM

If the quantity is not updatable, the update icon shouldn't show on that row.

scottcwilson:
Fixes #4006 

Regular updatable row: 

updatable

Not updatable: 

not_updatable


includes/modules/pages/shopping_cart/header_php.php
```diff
@@ -132,7 +132,16 @@
...
-  $buttonUpdate = ((SHOW_SHOPPING_CART_UPDATE == 1 or SHOW_SHOPPING_CART_UPDATE == 3) ? zen_image_submit(ICON_IMAGE_UPDATE, ICON_UPDATE_ALT) : '') . zen_draw_hidden_field('products_id[]', $products[$i]['id']);
+  if (SHOW_SHOPPING_CART_UPDATE == 1 or SHOW_SHOPPING_CART_UPDATE == 3) {
+     if (!$showFixedQuantity) { 
+        $buttonUpdate = zen_image_submit(ICON_IMAGE_UPDATE, ICON_UPDATE_ALT); 
+     } else {
+        $buttonUpdate = zen_image_submit(ICON_IMAGE_UPDATE, ICON_UPDATE_ALT, 'style="opacity: 0.25" disabled="disabled"'); 
+     }
+  } else {
+     $buttonUpdate = ''; 
+  }
```
drbyte, line 143: I'd prefer that this final `else` clause be refactored (removed) by moving its operation up before the related `if`. 
ie: initialize it first (with the "else" bit). Then handle conditional logic to override it if appropriate. Therefore no need for `else` clause.

PR #4017, opened 2020-10-08
Make copy product robust for long names

scottcwilson:
a) With newer MySQL versions, you get a white screen from attempting to copy a product with a name that is too long.

```
--> PHP Fatal error: 1406:Data too long for column 'products_name' at row 1 :: INSERT INTO products_description (products_id, language_id, products_name, products_description, products_url) ...
```

b) The new suffix TEXT_DUPLICATE_IDENTIFIER is part of the problem.  Good idea, but the string "[DUPLICATE]" is almost 20% of the max length of the product name.  Shortening to "COPY".


admin/includes/modules/copy_product_confirm.php
```diff
@@ -111,10 +111,15 @@
                                       FROM " . TABLE_PRODUCTS_DESCRIPTION . "
                                       WHERE products_id = " . $products_id);
         foreach ($descriptions as $description) {
+            $name = zen_db_input($description['products_name']); 
+            $newname = $name . " " . TEXT_DUPLICATE_IDENTIFIER; 
+            if (strlen($newname) > zen_field_length(TABLE_PRODUCTS_DESCRIPTION, 'products_name')) {
+               $newname = $name;
```
drbyte, line 117: For feedback: Instead of setting `$newname = $name` (which leaves it less obvious which one is the duplicate), why not truncate at the length of the field?

scottcwilson: Because then you'd wind up with "Name of product CO" or something like that if COPY were half truncated.

drbyte: True. Putting "COPY" at the beginning would avoid that.

PR #4026, opened 2020-10-15
Corrections in "audiences" processing

scottcwilson:
a) get_audiences_list second parameter should be 'true' or 'false', not 'customers_email_address'.  In this case, 'false' makes sense because you're emailing one customer and you don't want to wait for the counts to be created. 

b) in get_audiences_list, second parameter needs exact compare to 'true' or true.  Using == means you always enter the loop, because 'customers_email_address' (or 'false') will == true. 

c) The check for display_counts should be done in one place and reused in determining whether to add the count to the string; current code appends "(0)" when this display_counts is false, which is misleading.


scottcwilson:
If a PR for 1.5.7a is desired, just let me know; due to language changes this can't just be directly merged back.

drbyte:
Yes, please PR this to the v157 branch as well. Thanks

PR #4031, opened 2020-10-17
Allow serving of .woff2 files

scottcwilson:


drbyte:
Go ahead and PR this to 157 as well. (File lineendings are different so I can't cherrypick it back)

PR #4035, opened 2020-10-20
Docs for 1.5.7a

scottcwilson:


docs/changed_files-v1-5-7.html
```diff
@@ -68,13 +68,139 @@ <h1>Zen Cart&reg; Version 1.5.7</h1>
...
+               <li>includes/classes/vendors/PHPMailer/language/phpmailer.lang-de.php</li>
+               <li>includes/classes/vendors/PHPMailer/language/phpmailer.lang-id.php</li>
+               <li>includes/classes/vendors/PHPMailer/VERSION</li>
+               <li>includes/classes/vendors/PHPMailer/composer.json</li>
+               <li>includes/classes/vendors/PHPMailer/get_oauth_token.php</li>
+               <li>includes/classes/vendors/PHPMailer/src/PHPMailer.php</li>
+               <li>includes/classes/vendors/PHPMailer/src/OAuth.php</li>
+               <li>includes/classes/vendors/PHPMailer/src/SMTP.php</li>
+               <li>includes/classes/vendors/PHPMailer/src/Exception.php</li>
+               <li>includes/classes/vendors/PHPMailer/src/POP3.php</li>
```
drbyte, line 96: For the `vendors/PHPMailer` directory, let's not list the files, but rather just say "entire directory". This is the typical pattern followed for code that comes from other sources.

drbyte:
Thanks for this.

While it doesn't bother me much because I never use these lists (I always compare files for myself, never relying on checklists), does it make sense to sort the changed files list?

PR #4037, opened 2020-10-22
Add DISABLED_PRODUCTS_TRIGGER_HTTP200 config setting to "Stock" section

### LINKED ISSUE #3789: New boolean define PRODUCT_THROWS_200_WHEN_DISABLED does not use quoted values

drbyte:
Add `DISABLED_PRODUCTS_TRIGGER_HTTP200` config setting to "Stock" section:
```
When a product is marked Disabled (status=0) but is not deleted from the database,
 should Search Engines still show it as Available?
True = Return HTTP 200 response
False = Return HTTP 410   (THIS IS THE DEFAULT, AND MATCHES PREVIOUS BEHAVIOR)
(Deleting the product will return HTTP 404)
```

Yes, it's a global setting, and will be applied to all products unless other code overrides it at a per-product level.

Replaces undocumented `PRODUCT_THROWS_200_WHEN_DISABLED` added in `v157` (and now removed in `v157a`) which was alleged to be coded backwards anyway.

Closes #3834
Closes #3790
Closes #3789


scottcwilson:
Is the 1.5.8 upgrade statement INSERT IGNORE for this config ?

includes/functions/functions_products.php
```diff
@@ -62,7 +62,7 @@ function zen_product_set_header_response($product_id, $product_info = null)
         $response_code = 410;
     }
 
-    if (defined('PRODUCT_THROWS_200_WHEN_DISABLED') && PRODUCT_THROWS_200_WHEN_DISABLED === true) {
+    if (defined('DISABLED_PRODUCTS_TRIGGER_HTTP200') && (DISABLED_PRODUCTS_TRIGGER_HTTP200 === true || DISABLED_PRODUCTS_TRIGGER_HTTP200 === 'true')) {
```
scottcwilson, line 65: why are we testing for true and 'true' ?

scottcwilson (commented):
Also, should we update `zen_get_buy_now_button()` ?

includes/functions/functions_products.php
drbyte: I suppose the boolean is now redundant if it's going to be a Configuration Switch.

drbyte:
> Also, should we update `zen_get_buy_now_button()` ?

Seems like a different topic IMO. Not related to whether we respond to search engines with a 200 vs a 410 response.

That feels more like "Should disabled products be allowed to be purchased?", which is tightly related to "Should products that go out of stock be marked as disabled?" for which there's already a switch.

scottcwilson:
Yes but the switch is designed for the current semantics of "disabled," which means "not visible in the storefront." 

Well, we can always change this later.

Issue #4013, opened 2020-10-07
Proposed: remove HTTPS_SERVER define


drbyte:
Since modern browsers expect SSL all the time, there's no need to support intermingled http/https like we did in original Zen Cart back when SSL was only used by those who were security-minded.

Thus the double set of `HTTP_SERVER` + `DIR_WS_CATALOG` and `HTTPS_SERVER` + `DIR_WS_HTTPS_CATALOG` is redundant.

I'd like to remove the `HTTPS_SERVER` and `DIR_WS_HTTPS_CATALOG` defines and all the internal infrastructure that refers to them. Both Catalog and Admin.

Caveat: there may be plugins that have hard-coded things to reference these defines. 
But chances are those plugins need other updates anyway so I'd be satisfied with simply documenting the change.

scottcwilson:
Can we just define the things which are removed in core the way we did with other variables we removed from the configure file?

proseLA:
i'm a fan of the removal.  either you keep up or you get left behind.  i am a fan of less configuration vars.  plugins either get updated or left behind.

dbltoe:
I'm one of the first to promote keeping current but, what about the showcase sites?  Is it the browsers that are going to be the bad guys in this or is it going to be ZC not working without SSL?  Just would like to be able to blame the "other guy".

drbyte:
> what about the showcase sites?

I'm not sure I follow whatever you meant to say here.

dbltoe:
Searching for "Welcome, please enjoy our online showcase." comes up with some 1600 returns.  About a hundred are how to change it so, around 1500 are not selling and have not changed the default.  If you figure that just a fourth of the actual non-selling sites did change the greeting, there would be about 2000 sites that could operate fine without an SSL.  Some have an SSL regardless of neither selling nor logging folks in but, most that I looked at did not have a link to log in.  Since the lack of SSL would probably not be triggered in this case.  Is it enough of an occurrence to warrant any consideration?

drbyte:
@dbltoe I think you're saying "there are some sites that do operate entirely without SSL".

Yes I agree. Even in a localhost test environment some will not use SSL. That's simple: just set an http URL into HTTP_SERVER.

But when a site is live and has a functional SSL certificate, then HTTP_SERVER would get defined to an `https://` URL, and the store would always generate https URLs when any URL-generating functions are called, such as `zen_href_link`.

[...]

PR #4039, opened 2020-10-24
Ensure correct sub-directory for plugin relative path

### LINKED ISSUE #4023: 1.5.7: getPluginRelativeDirectory returns wrong relative path
The value returned by the getPluginRelativeDirectory method in includes/classes/FileSystem.php is incorrect if ZenCart is not installed in the "root" folder.  

I fixed it (on my site) by changing the method to:

` public function getPluginRelativeDirectory($pluginKey)`
`    {`
`        if (!isset($this->installedPlugins[$pluginKey])) {`
`            return null;`
`        }`
`        $version = $this->installedPlugins[$pluginKey]['version'];`
`        $relativePath = 'zc_plugins/' . $pluginKey . '/' . $version . '/';`
`        if (DIR_WS_CATALOG != '') $relativePath = DIR_WS_CATALOG . '/' . $relativePath;`
`        return $relativePath;`
`    }`

It fixes the issue I ran into (css not loading for the plug in), But it doesn't seem right ....

All the other methods use DIR_FS_CATALOG. So they all deal with file system paths as you would expect for a class named FileSystem.`  But, this method is used to produce the URL to assets like the plug-in's css file.

mc12345678:
Fixes #4023

### DIFF
--- a/includes/classes/FileSystem.php
+++ b/includes/classes/FileSystem.php
@@ -123,7 +123,7 @@ class FileSystem
             return null;
         }
         $version = $this->installedPlugins[$pluginKey]['version'];
-        $relativePath = '/zc_plugins/' . $pluginKey . '/' . $version . '/';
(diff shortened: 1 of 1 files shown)


proseLA:
fyi:
https://github.com/zencart/zencart/issues/4013

zcwilt (approved):

PR #4048, opened 2020-10-28
Ensure that attempt to handle/process attribute array is an array.

mc12345678:


drbyte:
Please provide a description.

I'm not sure I understand how you arrived at this being necessary.

Sure, theoretically it's possible that someone could alter the code on their site in a way that set this value to a non-array.
But how likely is it that they would, considering it would break things.

Under what present-day live-site circumstances is this necessary?

PR #4050, opened 2020-10-28
Remove unnecessary casting and variable use.

mc12345678:
$pid is already cast to the integer of $_GET['pid'], simplifies the code
to use that cast value throughout the header file.


drbyte:
With queries I almost always re-cast anyway just so I know that it has been intentionally addressed.
Else it requires scanning the code to ensure that it's been handled.
This way it also minimizes the chance that a user has customized things insecurely.

PR #4051, opened 2020-10-28
Support handling an empty query result upon attempted delete.

mc12345678:


drbyte:
Has this been triggering a problem in the wild?

mc12345678:
A myDebug-adm file was generated while I was reviewing the changes that were being made to address the issue with assigning a template to a language. I didn't back track sufficiently to determine if the issue was caused by the old code where the language was directly linked to the template, detached from the language or if it occurred as a result of a timeout and subsequent re-login while at a point to commit the deletion. 

Environment was php 7.4, with all issues raised (not ignoring notices).

PR #4082, opened 2020-11-12
admin orders: separate and clarify status update and status filter forms

torvista:
Instigated by a html warning as two anchors shared the "status" id.
I've separated and rationalised the naming for the status filter and status update forms and posts to clarify their function.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -24,7 +24,7 @@ if (isset($_GET['download_reset_on'])) {
 if (isset($_GET['download_reset_off'])) {
   $_GET['download_reset_off'] = (int)$_GET['download_reset_off'];
 }
-if (!isset($_GET['status'])) $_GET['status'] = '';
+if (!isset($_GET['statusFilterSelect'])) $_GET['statusFilterSelect'] = '';
 if (!isset($_GET['list_order'])) $_GET['list_order'] = '';
 if (!isset($_GET['page'])) $_GET['page'] = '';
 
@@ -233,7 +233,7 @@ if (zen_not_null($action) && $order_exists == true) {
       $oID = zen_db_prepare_input($_GET['oID']);
       $comments = !empty($_POST['comments']) ? zen_db_prepare_input($_POST['comments']) : '';
       $admin_language = zen_db_prepare_input(isset($_POST['admin_language']) ? $_POST['admin_language'] : $_SESSION['languages_code']);
-      $status = (int)$_POST['status'];
+      $status = (int)$_POST['statusUpdateSelect'];
       if ($status < 1) {
          break;
       }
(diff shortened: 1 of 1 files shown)


proseLA:
i'm sorry.  what are we gaining here?  how can one reproduce the warning that you are talking about?

i'm seeing a lot of name changes without any real code cleanup.

i do a lot with the orders script on the admin side, and ideally i would like to see the problem that your solution addresses.  because rather than clarifying the any status update issue, it is only confusing it more to me.

torvista:
Clipboard01

This "warning" will not be visible if there are "errors".

I thought the use of "status" in two different aspects (filter and update) was unclear and could be improved by more explanatory names.

proseLA:
- i can not reproduce it.
- its better to start with a url, where this error exists, ie, something like https://zcdev.com/admin/index.php?cmd=orders  because i am not sure if you have an added action to the url or what.  and exactly what you did to get this error.
- in my v158 install, i see 1 instance of `id="status"` and 1 instance of `id="selectstatus"`
- MO is to only fix code that needs fixing.  naming can always be better; but once named do not change just for clarity.
- unless you can accurately show me how to reproduce this error, i'm not a fan of this PR.

torvista:
I have done nothing to the file. It's vanilla ZC158.
I just enter the orders page and open the html validator in chrome.

The source shows this up, I think it is moaning about these two being the same.

Clipboard01

I understand about not changing what is not broken, but especially in light of this being an oft-modified file, is it not worth clarifying the uses of "status"?

torvista:
I know this is irritating...but anyway to muddy things further...seems the name attribute is deprecated on a form anyway (should be id).
Just to illustrate, these are the uses of "status" on the same page:
Clipboard01

Clipboard02

drbyte:
It appears this wasn't an issue until the "quick update status" was added which used the same name.

drbyte:
I'm not a fan of extra-long $_GET parameters. Rather than 3-word renaming perhaps we can shorten things a little? 
For example instead of going from `status` to `statusFilterSelect` it could be just `filter`?

And maybe the quick-update $_POST element could be `new_status` rather than `statusUpdateSelect`.

`filter` and `new_status` feel more semantic and less programmer-speak.

As for the form name, I'm less opinionated. As you say, it's largely moot unless one is trying to target it in the DOM without having an id associated with it.

proseLA:
so many issues...  

with regards to name being deprecated on a form, have at it:

https://github.com/zencart/zencart/blob/42d150d2a3942266eb30b176ef2937f023d17f15/admin/includes/functions/html_output.php#L233-L234

i agree that names (i meant IDs) on forms on less problematic...

and if you change the name of a posted element, make sure you find where that posed element is used so that the proper action takes place.

torvista:
I'm not sure what you wish to illustrate with the function snippet and comment...perhaps you could re-word it for the hard-of-thinking..

drbyte:
> with regards to name being deprecated on a form, have at it:
> https://github.com/zencart/zencart/blob/42d150d2a3942266eb30b176ef2937f023d17f15/admin/includes/functions/html_output.php#L233-L234

@proseLA, I'm not sure what your intention was by the "have at it" comment.
But as for the rest of your quote, if you're advocating for changing the way that function is used, including its parameters, please open an issue for it so we can discuss it there. There are dozens of pros and cons for making that sort of change. None without impacting a ton of code. Best to discuss in a separate issue.

proseLA:
> I know this is irritating...but anyway to muddy things further...seems the name attribute is deprecated on a form anyway (should be id).

i apologize for my wording.

as far as i can tell, @torvista appears to be correct that the name attribute is deprecated for the form element.

i also agree that changing the `zen_draw_form` is rife with potential for problems considering its use throughout the repo, and certainly worthy of separate issue and discussion.

PR #4083, opened 2020-11-13
Rename prod/cat pulldown functions to reduce ambiguity

drbyte:
`zen_draw_products_pull_down` -> `zen_draw_pulldown_products`

`zen_draw_products_pull_down_attributes` -> `zen_draw_pulldown_products_having_attributes`

`zen_draw_products_pull_down_categories` -> `zen_draw_pulldown_categories_having_products`

`zen_draw_products_pull_down_categories_attributes` -> `zen_draw_pulldown_categories_having_products_with_attributes`


scottcwilson:
Should we add stub functions with the old names that call the new functions to ensure that plugins don't break?

torvista:
Doesn't that just postpone the inevitable?
I would agree for 157b, but for a major version, is it not expected to break some eggs, as long as there is a "Breaking Changes" section in the upgrade docs.

scottcwilson:
You can throw a deprecated warning in the stub functions to encourage people to fix these issues, but I would prefer not to break eggs.

includes/functions/functions_categories.php
```diff
@@ -321,7 +321,7 @@ function zen_product_in_parent_category($product_id, $cat_id, $parent_cat_id)
  * @param bool $show_current_category
  * @return string
  */
-function zen_draw_products_pull_down($field_name, $parameters = '', $exclude = [], $show_id = false, $set_selected = 0, $show_model = false, $show_current_category = false)
+function zen_draw_pulldown_products($field_name, $parameters = '', $exclude = [], $show_id = false, $set_selected = 0, $show_model = false, $show_current_category = false)
```
proseLA, line 324: how would you feel about adding another input parm to the end of this function (obviously in a separate PR)?  i would like to propose `$keyword = ''`.  we could then use the `zen_build_keyword_where_clause` to add to the query.

drbyte: @proseLA while I like the idea I wonder about whether we're in a good spot for implementing that.

If we add $keywords to this function, should that be earlier in the list of function parameters? Most of the latter ones are booleans decorating the output. I'm undecided about whether taking $keywords onto the end is just bandaging. Given the function is being renamed here, now's also the best time to change its signature if needed.

Also, if we add it to this one ..... there's a dozen more pulldown menus ... sigh, do we have to add it to all of those too? 

I wonder whether the whole function should be refactored to extract the query from the decorating? Maybe the same for all of these helper functions? Perhaps the helper/decorator parts should be in html_output and the functions that retrieve the data elsewhere.

Note: I'm generally a little reluctant to add any raw query injection capability to any functions just for security reasons.

proseLA: i will take a look when able and give some thoughts.  i think if we are renaming and using stubs for the old functions, it is a good time to look at potential refactoring opportunities; including function signatures.

proseLA: the number of `pulldown`s is not large, maybe a dozen.

IMO, i think the inclusion of `$keywords` and the `zen_build_keyword_where_clause` is only relevant to 4 of them; the 2 product and potentially the 2 category pulldowns.  

it does seem like we could refactor the sql for those to all be in 1 function, perhaps 2.  on a cursory look, it also seems that the logic for the function is used `zen_draw_pull_down_menu` is recreated needlessly in those 4 functions.

with regards to the bandaging comment, perhaps true, but certainly a path of least resistance. 

refactoring certainly has the opportunity to reduce much code duplication.

PR #4084, opened 2020-11-13
Add letter-pagination option to admin split_page_results

nick-cambridge:
The main part of this change is to the split page result page.
Two other pages are then changed to ilistrate how to use the interface.
The class defaults to the same behaviour as the old class.


drbyte:
Thanks for your submission.

What PHP version was this written for? It seems to be using PHP conventions that have been deprecated in recent years.

nick-cambridge:
Yes sorry this was changed many years ago.   Have probably been using this code 10 years.

I have now corrected the constructor and the old style while loop instead of the foreach.

lat9:
@drbyte might also have been referring to the PSR-2 Coding Standards (https://www.php-fig.org/psr/psr-2/) that are not used in your submission.

drbyte:
> I have now corrected the constructor and the old style while loop instead of the foreach.

Thanks.

> @drbyte might also have been referring to the PSR-2 Coding Standards (https://www.php-fig.org/psr/psr-2/) that are not used in your submission.

While consistent formatting is preferable, the present concern is around whether the proposed changes are the best way of doing what is needed.

drbyte:
Thank you for your contribution. Thank you for fixing the old PHP syntax issues.

What follows may sound harsh. It is not intended as such. It is part of code-review. 

This class is used in a lot of places. Making changes has sweeping impact.

A bigger-picture internal goal related to this class is extracting out the generated HTML in a way that can be overridden externally without losing the rest of the core logic. (No, I'm not asking you to do these refactorings. I'm just pointing out something few may be aware of.)

I fear this proposed code, if merged as-is, risks putting a stake in the ground of even more complex function signatures that makes other refactoring even more fragile. Of particular concern here is the passing of additional parameters to the constructor.
(I'm not saying the additional parameters aren't a good way to write code. In fact it's a good place to start.)

My other concern is the numerous nested if/else/if/else/if construction. I'd prefer to see these simplified, perhaps refactored to internal protected methods. This way, even if the if/else/if/else/if stuff must be retained then at least each of those clauses could merely call a named method and thus the reading of the if/else/if constructs is easier to follow.

For the new code you've proposed here, would you be willing to tackle refactoring it further? If not, I can look into it, but it may be awhile before I do. No rush either way.

nick-cambridge:
In order to find a country, for example, you had to basically have to binary chop the page list and finding a zone was even worse. I can't remember what page I originally developed this enharcment for but I released the country page because it was simple and easy to see the advantage.  We use the countries page quite a lot because we have added fields in the countries table for multiple shipping modules so that process was just unacceptable to me as I hate bad user interfaces.

As neither the original code or the proposed code supports sql sub queries (something I found out the hard way) there is a huge hole in the system anyway for future development and really the argument should be more focused on making the system more useable. 

Believe me once you change just these two pages ( I use this code on many others ) you realise this is how the interface should have worked all along and the current SQL page management is just plain poor.
It is important as a programmer to remember that usability is far more important than perfect code and with so much of the legacy code of such a poor standard as it was inherited from oscommerce. 

Yes you can refactor the code to reduce the nested ifs.  (never had a big problem with them myself but that is why I prefer the format
if( condition)
{
 ....
}
as the balanced brackets just make theses ifs and loops easier to check (personal preference).

It would be far better use of your time or mine to refactor one of the many other huge short comings of the code base like splitting  the orders.php file  into an orders.php that lists the orders and an order.php that manages an individual order as I have always found the combined page horrible to edit I would release my version of this but I have many more features on the order.php page to improve our work flow.  Just simple things such as changing product attribute, splitting multiple lines, managing returns, refunds but these are so linked to so many other enharcments and not sure if there are adequate hooks to make these sort of enharments overridable from the core code.

Alternatively maybe put a class in place for stock control as the current value stored in the products table is a very poor design and the direct access of this value makes implementing an alternative very invasive of core code.  I was somewhat surprised  that after so many years this was still the case.

Its all very well saying that the big picture is to ... and I am not sure if this is referring to version 2.0
But version 2.0 has been 'in the pipeline' for more years than I can remember so I have given up even thinking about it.

The point about separating out the html generating code from the core logic.  As I am sure you will agree both the existing code and the new implementation do this to the same degree.  There is the core logic that breaks down the SQL and creates the list of pages  (i.e the constructor ) and then the code that displays the link list/form and the counts where appropriate i.e the display functions.  It is obviously very easy to derive a class from this one and overload any of these functions.

It has always surprised me how little classes have been used in the system.  When you look at something like the coupons where this is functionality that is crying out for a class hyaricy  and we are stuck with this monolithic class with umpteen modes that are full of bugs and unmaintainable.

If  you look at many of the decisions made to 'separate the html from the logic', again much inherited from oscomerce,
rather than creating a system that is simple to maintain and understand  it ended up smearing the html production across several classes and across several directories making working out where html came from very hard.  I refer particularly to the code for category box output. I come out in a cold sweat every time I have to go near that code and our system allows customers to select from 4 formats.

It is sometimes best to KISS.  Query, Loop ( output ) rather than the query loop( store in intermediate format with some html) .... sometime later loop over intermediate format to output with more html.  There is little gained and simplicity is certainly lost in the latter approach.

My big problem at the moment is needing to move large amounts of code into the latest system so that I can then upgrade from 1.3 to 1.5.  The functionality I have added is:-

Full Multi-site implementation with no change to the front end.
Full Drag and drop support for every feature you would want to drag and drop.
Automatic additional document printing with invoices.
Update orders history/status on print.
Obviously batch printing invoices.
Automatic email processes for abandoned baskets etc.
One  scan dispatch with usb scales integration and automatic label print.
Better refund handling -> Xero
Better stock management though a separate class to allow decent overloading for different requirements (I use stock by attributes contrib).
Improved Coupons, that supports include/exclude and manufacture/categories/product and doesn't rely on the poorly understood Master Category ID which gets out of date so eaiserly. 
Full stock change audit.
Automatic invoice processing of incoming invoices.
Decent basic report page with drill down of multiple reports.
Attribute grid for products rather than multiple drop down.
Draggable Warehousing map and stock location management.
Graphical stock levels and order history for each sku.

Sensible basic table management to reduce the pain of adding new table to the system to avoid the repeated cut paste and edit for each new table added to the system.


e.g This is the whole of my counties.php file which has four extra fields on it:-

    $index_field = 'countries_id';
    $table = TABLE_COUNTRIES;
    $order_field = 'countries_name';
    $heading_txt = "countries";
    $dataCleanup = 'simpleCleanUp';
     $fieldsList = "countries_name    countries_iso_code_2 countries_iso_code_3 country_visible shipping_group state_dropdown address_lookup ";
        $formatFieldFunction = 'groupsFieldDisplay';
    $splitField = 'countries_name';
    $statusFieldsList = "country_visible state_dropdown address_lookup";
       
$filename = FILENAME_COUNTRIES;

  include "BasicTableManagement.php";

So I will try any do the requested extra functions at some point but there are a lot of far more pressing things to be done.

Nick

nick-cambridge:
Thank you for the feedback.

Sorry if the following reads  harsh I couldn't find the words to express this without it sounding more critical than it was meant.

 I have found progress on the development zencart frustratingly backward. I started originally with OSCommerse but moved to Zencart many years ago as it seemed to offer so much more especially with the much promised version 2 but it has been stuck with this obsolete architecture/code with slow or no movement to adequate functionality for so many years.

I find it amazing that even things as essential as stock keeping functionality has not moved to even a basic class implementation.  The lack of any concept of SKU's in the base system and stock number maintained  only at the product level still in 1.5.8 is not reassuring.  I did see one suggestion of a class to make overloading this feasible but there seemed little support.

Don't really agree with your issue about adding parameters to the function because the backward compatibility is maintained through the use of the defaults.  The current released functionality is not really fit for purpose.  I have used this new code for 10+ years and it has always amazed me that the standard code was thought of as an adequate functionality.   That the code has been changed  sometime over the years but still remained functionally poor is disapointing.  

In order to find a country, for example, you had to basically have to binary chop the page list and finding a zone was even worse. I can't remember what page I originally developed this enharcment for but I released the country page because it was simple and easy to see the advantage.  We use the countries page quite a lot because we have added fields in the countries table for multiple shipping modules so that process was just unacceptable to me as I hate bad user interfaces.

As neither the original code or the proposed code supports sql sub queries (something I found out the hard way) there is a huge hole in the system anyway for future development and really the argument should be more focused on making the system more useable. 

Believe me once you change just these two pages ( I use this code on many others ) you realise this is how the interface should have worked all along and the current SQL page management is just plain poor.
It is important as a programmer to remember that usability is far more important than perfect code and with so much of the legacy code of such a poor standard as it was inherited from oscommerce. 

Yes you can refactor the code to reduce the nested ifs.  (never had a big problem with them myself but that is why I prefer the format
if( condition)
{
 ....
}
as the balanced brackets just make theses ifs and loops easier to check (personal preference).

It would be far better use of your time or mine to refactor one of the many other huge short comings of the code base like splitting  the orders.php file  into an orders.php that lists the orders and an order.php that manages an individual order as I have always found the combined page horrible to edit I would release my version of this but I have many more features on the order.php page to improve our work flow.  Just simple things such as changing product attribute, splitting multiple lines, managing returns, refunds but these are so linked to so many other enharcments and not sure if there are adequate hooks to make these sort of enharments overridable from the core code.

Alternatively maybe put a class in place for stock control as the current value stored in the products table is a very poor design and the direct access of this value makes implementing an alternative very invasive of core code.  I was somewhat surprised  that after so many years this was still the case.

Its all very well saying that the big picture is to ... and I am not sure if this is referring to version 2.0
But version 2.0 has been 'in the pipeline' for more years than I can remember so I have given up even thinking about it.

The point about separating out the html generating code from the core logic.  As I am sure you will agree both the existing code and the new implementation do this to the same degree.  There is the core logic that breaks down the SQL and creates the list of pages  (i.e the constructor ) and then the code that displays the link list/form and the counts where appropriate i.e the display functions.  It is obviously very easy to derive a class from this one and overload any of these functions.

It has always surprised me how little classes have been used in the system.  When you look at something like the coupons where this is functionality that is crying out for a class hyaricy  and we are stuck with this monolithic class with umpteen modes that are full of bugs and unmaintainable.

If  you look at many of the decisions made to 'separate the html from the logic', again much inherited from oscomerce,
rather than creating a system that is simple to maintain and understand  it ended up smearing the html production across several classes and across several directories making working out where html came from very hard.  I refer particularly to the code for category box output. I come out in a cold sweat every time I have to go near that code and our system allows customers to select from 4 formats.

It is sometimes best to KISS.  Query, Loop ( output ) rather than the query loop( store in intermediate format with some html) .... sometime later loop over intermediate format to output with more html.  There is little gained and simplicity is certainly lost in the latter approach.

My big problem at the moment is needing to move large amounts of code into the latest system so that I can then upgrade from 1.3 to 1.5.  The functionality I have added is:-

Full Multi-site implementation with no change to the front end.
Full Drag and drop support for every feature you would want to drag and drop.
Automatic additional document printing with invoices.
Update orders history/status on print.
Obviously batch printing invoices.
Automatic email processes for abandoned baskets etc.
One  scan dispatch with usb scales integration and automatic label print.
Better refund handling -> Xero
Better stock management though a separate class to allow decent overloading for different requirements (I use stock by attributes contrib).
Improved Coupons, that supports include/exclude and manufacture/categories/product and doesn't rely on the poorly understood Master Category ID which gets out of date so eaiserly. 
Full stock change audit.
Automatic invoice processing of incoming invoices.
Decent basic report page with drill down of multiple reports.
Attribute grid for products rather than multiple drop down.
Draggable Warehousing map and stock location management.
Graphical stock levels and order history for each sku.

Sensible basic table management to reduce the pain of adding new table to the system to avoid the repeated cut paste and edit for each new table added to the system.


e.g This is the whole of my counties.php file which has four extra fields on it:-

    $index_field = 'countries_id';
    $table = TABLE_COUNTRIES;
    $order_field = 'countries_name';
    $heading_txt = "countries";
    $dataCleanup = 'simpleCleanUp';
     $fieldsList = "countries_name    countries_iso_code_2 countries_iso_code_3 country_visible shipping_group state_dropdown address_lookup ";
        $formatFieldFunction = 'groupsFieldDisplay';
    $splitField = 'countries_name';
    $statusFieldsList = "country_visible state_dropdown address_lookup";
       
$filename = FILENAME_COUNTRIES;

  include "BasicTableManagement.php";

So I will try any do the requested extra functions at some point but there are a lot of far more pressing things to be done.

Nick

drbyte:
@nick-cambridge Thanks for sharing your code.
I've massaged it quite a bit, but it still functions the same as you submitted.

You mentioned a number of other enhancements you've made. Community contribution is welcome!

PR #4085, opened 2020-11-13
Do not round concatenated value used in sorting

scottcwilson:
The round operation can fail and return 0 in some cases, making the value field zero.  Removing it prevents this from occurring.

**WITH round:** 

mysql> select options_values_price, round(concat(price_prefix, options_values_price), 5) as value from zen_products_attributes where products_id = '661' and attributes_display_only != '1' and attributes_price_base_included='1' order by options_id, value;
+----------------------+---------+
| options_values_price | value   |
+----------------------+---------+
|             110.0000 | 0.00000 |
|              90.0000 | 0.00000 |
|              50.0000 | 0.00000 |
|              70.0000 | 0.00000 |
+----------------------+---------+
4 rows in set (0.00 sec)

**WITHOUT round:** 

mysql> select options_values_price, concat(price_prefix, options_values_price) as value from zen_products_attributes where products_id = '661' and attributes_display_only != '1' and attributes_price_base_included='1' order by options_id, value;
+----------------------+-----------+
| options_values_price | value     |
+----------------------+-----------+
|             110.0000 |  110.0000 |
|              50.0000 |  50.0000  |
|              70.0000 |  70.0000  |
|              90.0000 |  90.0000  |
+----------------------+-----------+
4 rows in set (0.00 sec)

### DIFF
--- a/includes/functions/functions_prices.php
+++ b/includes/functions/functions_prices.php
@@ -425,7 +425,7 @@ function zen_get_products_base_price($product_id)
     // do not select display only attributes and attributes_price_base_included is true
     $sql = "SELECT options_id, price_prefix, options_values_price,
                    attributes_display_only, attributes_price_base_included,
-            ROUND(CONCAT(price_prefix, options_values_price), 5) AS value
+            CONCAT(price_prefix, options_values_price) AS value
             FROM " . TABLE_PRODUCTS_ATTRIBUTES . "
             WHERE products_id = " . (int)$product_id . "
             AND attributes_display_only != 1


drbyte:
> "The round operation can fail and return 0 in some cases"

Are the "in some cases" documented somewhere? What are those cases?

What's lost by removing the `round()` call?

The call to `round()` has been in this function for 10 years. How did you arrive at this?

scottcwilson:
Round would be appropriate if we were using the value in calculations or display, but we're not - it's not needed. 

I can't tell you the specific data/configuration combination that causes this but I *can* reproduce it as shown above.

drbyte:
Given that the problem hasn't manifested itself in 10 years, I'm curious how you happened upon this "new" situation. Why hasn't it shown up before?

scottcwilson:
I'm guessing the relevant variable is PHP7.  Do you want to mark this "under investigation" and wait to see if there are more reports?

drbyte:
> if we were using the value in calculations or display, but we're not

I was going to ask: what other "use" would there be? 

Turns out we're not using it at all. That function doesn't use that result in any way.

So ... a better refactor would be to remove the line entirely.

But, again, I'm still curious how you arrived at discovering this in the first place.

drbyte:
> Turns out we're not using it at all. That function doesn't use that result in any way.

I was wrong: it is used in the order-by clause.

drbyte:
> I just upgraded a site from 1.5.5 to 1.5.7
> They use the "Show Starting At" setting.
> The customer said, "The starting at price is not the lowest price."
> I investigated and found what I showed above.

Thanks. Context is ALWAYS helpful. Important even.

drbyte:
> I just upgraded a site from 1.5.5 to 1.5.7
> They use the "Show Starting At" setting.
> The customer said, "The starting at price is not the lowest price."

Given the result change appears within a query, I doubt it's a PHP7 thing.
Did they change MySQL versions? eg to something using a stricter mode than previously?

scottcwilson:
I see this behavior locally on Database Engine: MySQL 5.7.26.
Live site has Database Engine: MySQL 5.5.5-10.0.38-MariaDB

drbyte:
Hmmm ... given their older db version I'm surprised that they would have had any different experience prior to the ZC upgrade.

Maybe they just didn't notice it before, but the upgrade gave them cause to peek.

PR #4088, opened 2020-11-15
Correct for when DIR_WS_CATALOG is a subfolder

scottcwilson:
Recommended for backport to 1.5.7b.  
This will allow people to copy  `admin/includes/ckeditor.php`  to older Zen Cart installs that don't have the `zen_catalog_base_link` function.

### DIFF
--- a/admin/includes/ckeditor.php
+++ b/admin/includes/ckeditor.php
@@ -33,7 +33,7 @@ foreach ($var as $key)
 
                 CKEDITOR.replace(jQuery(this).attr('name'),
                     {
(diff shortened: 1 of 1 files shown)


mc12345678:
Perhaps for self edification, perhaps for inquiring minds, why `DIR_WS_HTTPS_CATALOG` instead of possibly `DIR_WS_CATALOG`, depending on whatever setup the system has?

drbyte:
I seem to recall going back and forth with PRs ... using the constant, then not, then adding it back.

PR #4095, opened 2020-11-18
Attributes controller minor html corrections

torvista:


admin/attributes_controller.php
```diff
@@ -1569,7 +1569,7 @@ function popupWindow(url) {
                     ?>
                     <td class="text-right"><?php echo $attributes_price_final . $new_attributes_price . ' ' . $attributes_price_final_onetime; ?></td>
                     <?php if ($action != '') { ?>
-                      <td width="120">&nbsp;</td>
```
drbyte, line 1572: Arbitrarily removing an attribute "just because it's 'deprecated'" doesn't necessarily mean it's the only thing needed.

Why was a width set on this column in the first place? How has that intention been accommodated in the "new" way that the "deprecation" intended? 
ie: if "width" is deprecated it usually means the intended new way is to use CSS or something to replace it. Where's the corresponding CSS to retain the design intention?

drbyte (changes requested):

admin/attributes_controller.php
torvista: To put this in context, this td is the last column displayed when Editing an attribute:

td120

The 120 width is irrelevant as the column width is determined by the "Action" heading text.

Here the 120 is removed, and there seems little visual difference.

td120

So apart from deprecated, it seems unneeded,

drbyte:
🤞

PR #4098, opened 2020-11-20
Admin, Customers, Edit Customer: Fixes for php notices and hide Company column

torvista:
ZC158 and current,
If the fields company, address second line ("suburb"), and state are disabled in admin configuration customers, Edit customer and save produces the notices
PHP Notice: Undefined index: entry_suburb in ..\zencart-158\admin\customers.php on line 97.
PHP Notice: Undefined index: entry_company in ..\zencart-158\admin\customers.php on line 101.
PHP Notice: Undefined index: entry_state in ..\zencart-158\admin\customers.php on line 102.
PHP Notice: Undefined variable: entry_zone_id in ..\zencart-158\admin\customers.php on line 185.

I've also set the Company column to hide on the listing when disabled.

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -704,12 +704,13 @@ if (zen_not_null($action)) {
                     <a href="<?php echo zen_href_link(basename($PHP_SELF), zen_get_all_get_params(array('list_order', 'page')) . 'list_order=firstname', 'NONSSL'); ?>"><?php echo ($_GET['list_order'] == 'firstname' ? '<span class="SortOrderHeader">Asc</span>' : '<span class="SortOrderHeaderLink">Asc</span>'); ?></a>&nbsp;
                     <a href="<?php echo zen_href_link(basename($PHP_SELF), zen_get_all_get_params(array('list_order', 'page')) . 'list_order=firstname-desc', 'NONSSL'); ?>"><?php echo ($_GET['list_order'] == 'firstname-desc' ? '<span class="SortOrderHeader">Desc</span>' : '<span class="SortOrderHeaderLink">Desc</span>'); ?></a>
                   </th>
+                  <?php if (ACCOUNT_COMPANY === 'true') { ?>
                   <th class="dataTableHeadingContent">
                     <?php echo (($_GET['list_order'] == 'company' or $_GET['list_order'] == 'company-desc') ? '<span class="SortOrderHeader">' . TABLE_HEADING_COMPANY . '</span>' : TABLE_HEADING_COMPANY); ?><br>
                     <a href="<?php echo zen_href_link(basename($PHP_SELF), zen_get_all_get_params(array('list_order', 'page')) . 'list_order=company', 'NONSSL'); ?>"><?php echo ($_GET['list_order'] == 'company' ? '<span class="SortOrderHeader">Asc</span>' : '<span class="SortOrderHeaderLink">Asc</span>'); ?></a>&nbsp;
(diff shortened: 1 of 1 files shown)


drbyte:
Disabling the Company field may not be the best approach.
The Admin should be able to see/edit the fields even if they are not normally displayed to the customer. Otherwise if the switch had been enabled at some point but later disabled, how would you be able to correct the previously-entered data?

That said, obviously some of these input fields are not being displayed, thus causing the missing $_POSTs that you're initializing.

I wonder if a better solution is that the conditions which are hiding the fields need to be removed?

torvista:
> The Admin should be able to see/edit the fields even if they are not normally displayed to the customer. Otherwise if the switch had been enabled at some point but later disabled, how would you be able to correct the previously-entered data?

Company Field: is already hidden in the Edit screen when disabled. I thought it would be consistent to also hide it in the listing, where space is at more of a premium. That's nothing to do with the php notices.
Based on your comment, this should be shown always on both screens.

So you are suggesting show all fields always in the admin?

Regarding unused inputs, there are three different approaches used in the code, which I thought should be reviewed to make them consistent, but not wishing to get shouted at for fixing stuff that was not broken, decided to add the minimum fix and leave the rest well alone.

drbyte:
> Company Field: is already hidden in the Edit screen when disabled. 

Correct. My comment was pointing out that maybe we shouldn't have done that.

> Based on your comment, this should be shown always on both screens.

I acknowledge the inconsistency.

> I thought it would be consistent to also hide it in the listing, where space is at more of a premium. 

Understood. If that change had been a separate PR we could have discussed it independently of initializing those other fields.

> So you are suggesting show all fields always in the admin?

My comment was not intended to be a statement suggesting any sweeping changes across the whole of the admin.

> decided to add the minimum fix and leave the rest well alone.

appreciated
👍

drbyte:
I'm going to merge this.

Just be aware that maybe the changes you made for the ACCOUNT_COMPANY switch might get changed again before release.

torvista:
What about hiding the Group Pricing and Coupon columns if those functionalities are not in use?
I have combined the ID and Date created columns as they are essentially the same.
Also the address column needs some tlc.

drbyte:
> What about hiding the Group Pricing and Coupon columns if those functionalities are not in use?

That depends on how you determine that "those functionalities are in use".

There's no "switch" to turn Group Pricing on/off. So how are you choosing whether that column should appear?

By "Coupon" column, do you mean "GV Balance"? What if a customer does have a GV Balance even if you're no longer offering the feature? Should that balance show? 

> I have combined the ID and Date created columns as they are essentially the same.

That means the created-date and last-login-date wouldn't appear together anymore, right?

Are you treating them as "the same" because that makes sorting easier? (ref: your other discussions about changing this UI)

----

While the attention being paid to this interface is appreciated, I think all the changes you're contemplating should be in separate PRs, not merely separate commits within a single PR.
This way the things that are easily reviewed and agreed-to can be quickly merged, and the things that may need discussion can be reviewed and discussed separately, independent of other proposed changes.

torvista:
They would all be in separate PRs.

Just simple checks.
Group Pricing; check if anything is defined in that table.
GV: if the order total module is enabled.

> Are you treating them as "the same" because that makes sorting easier?

No. I thought about the ID column being the only one not sortable, then realised the ids are in parallel with the date created, so why have two columns for what is essentially the same data.

I would not consider the date created and date of last login related to each other.

All of this just reduces use of space / display of unused information, and allows more space for extra custom fields.

PR #4099, opened 2020-11-20
Admin, customer options: clarify description of second address line "Suburb" in admin

torvista:
Personally I think the second address field should be named as such, but all the code handling refers to suburb...


drbyte:
You said you find `suburb` to be confusing, preferring `second address`/etc.

But your PR `adds` suburb to the storeowner's UI. 
IMO you're adding confusion, not removing it.

I don't think this is helpful for the storeowner.

torvista:
The Customer Edit field is named Suburb and all the underlying code uses suburb.
Currently the admin switch references second address line and makes no mention of the Suburb so I consider this a clarification tying together the field name and the field function.

drbyte:
Given that the only place `Suburb` is mentioned in any UI is the admin edit-customer screen, let's instead just update the language file.

```

/admin/includes/languages/lang.english.php
--
Line #174 :     'ENTRY_SUBURB' => 'Suburb:',

```

```/includes/languages/english.php

Line #187 :   define('ENTRY_SUBURB', 'Address Line 2:'); 

Line #188 :   define('ENTRY_SUBURB_TEXT', ''); 
```

PR #4101, opened 2020-11-20
Admin, Customers: move NOTIFY_ADMIN_CUSTOMERS_CUSTOMER_EDIT outside Company clause

torvista:
These fields go in their own group. I thought it best to title it Custom Fields to differentiate from core code.

I also added an additional field to the returned array, to provide the label "for"->"id" link in the input field.

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -383,50 +383,50 @@ if (zen_not_null($action)) {
           </div>
         </div>
         <?php
-        if (ACCOUNT_COMPANY == 'true') {
-          ?>
-          <div class="row">
-            <?php echo zen_draw_separator('pixel_trans.gif', '1', '10'); ?>
-          </div>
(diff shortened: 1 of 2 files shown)


drbyte:
Closed in dfe29b0b51a2a2e96b997459486b2ab1342a0afd

torvista:
1. It seems inconsistent to have observer-added fields not in their own formarea group, visually and code-wise. They seem orphaned/not integrated.

form group orphan

2. zen_draw_label is used to construct the label but with no for value:
https://github.com/zencart/zencart/blob/215eaf529641f48e50c29fb73ea8f335e3b9d8b5/admin/customers.php#L422-L429

empty fors

torvista:
Any observer-added fields are not in a formarea group, as shown above "orphaned".
To correct this, what is preferred: in their own "STORE_NAME" group or added onto the end of the "Options" group?
Since these are always store-specific I would suggest the former, and located after "Options".

PR #4102, opened 2020-11-20
Admin, Customer delete: remove/replace personal data, but do not remove record

torvista:
The entry in table customers_info is left intact as it has no personal info, and the date created field is used in the admin customer listing for sorting.


drbyte:
Implemented slightly differently, to allow choice during delete:
Forget Customer

PR #4108, opened 2020-11-27
Add coupon type 'G'

scottcwilson:


proseLA:
hi scott,
i'm not sure what this code does.  it checks for a different value of the `coupon_type` and then does the default action.  so nothing changes.  if you are building towards doing something else, perhaps you can elaborate?

in addition, it seems this section of code is ripe for a minor refactor.  we have 2 different switch statements that are switching on the same thing, and we could group the cases together.  i see a potential for 1/3 the lines of code with increased readability.

https://github.com/zencart/zencart/blob/186c4a0fa2b630ff58b62cf8e6c0d22fd448f609/includes/modules/order_total/ot_coupon.php#L360-L418

something like:
```
switch ($coupon_result->fields['coupon_type']) {
		case 'S': // shipping
		case 'O': // amount off and free shipping
		case 'E': // percentage and Free Shipping
			// use delivery address
			$check_zone_country_id = $order->delivery['country']['id'];
			$check_zone_id = $order->delivery['zone_id'];
			break;
		case 'F': // amount
		case 'P': // percentage
		default:
			// use billing address
			$check_zone_country_id = $order->billing['country']['id'];
			$check_zone_id = $order->billing['zone_id'];
			break;
	}
```
you can add your case of `G` as well to the 2nd set; but again, i am not sure why it is needed.  in fact, in the code above, we can eliminate the case of `F` and `P`; but again if it's a readability thing...  that's a different story.

drbyte:
I think the "for documentation" in the title is akin to "it's a readability thing". Depends on the angle from which you see it.

I'm okay with this being in the code as it makes it easy to find "how does it handle G?", etc.

proseLA:
no thoughts on combining the 2 switch statements and the cases within the switches?

drbyte:
> no thoughts on combining the 2 switch statements and the cases within the switches?

Noted. I haven't studied the code yet.
While I agree in principle with reducing complexity of code, I felt that some of your suggestions were overly aggressive so I'd be more interested in a hybrid.

Also I find myself curious if there is/was a reason why these sections weren't already combined.

scottcwilson:
> i'm not sure what this code does. it checks for a different value of the `coupon_type` and then does the default action. so nothing changes. if you are building towards doing something else, perhaps you can elaborate?

Sorry for the lack of clarity - my only intention was to be sure each possible value of coupon_type had an explanatory comment.  That's all.  I explained this in my first PR but I fat fingered it so I had to do it again and missed the explanation.

PR #4109, opened 2020-11-28
Combine switch statements in ot_coupon

proseLA:
- i do not think this is overly aggressive
- i think this is much easier to understand 
- it preserves all of the explanatory comments
- from a functional standpoint, i see no difference.
- it removes 39 lines of code and comments.


drbyte:
Thanks

PR #4110, opened 2020-11-29
Missing close single quote

scottcwilson:


drbyte:
Ref https://github.com/zencart/zencart/pull/2977

PR #4111, opened 2020-11-29
[157] Missing single quote

scottcwilson:
Missing single quote.  Backport of #4110 to 1.5.7.


drbyte:
Appears to have been introduced in https://github.com/zencart/zencart/pull/2977

PR #4117, opened 2020-12-03
Catch exception when send() fails

scottcwilson:
includes/classes/vendors/PHPMailer/src/PHPMailer.php line 1364 notes that send() can throw an exception, but we don't catch it in Zen Cart.  This means a fatal error, which could cause the checkout confirmation page to appear to hang (and cause multiple re-submits).

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -359,7 +359,12 @@ use PHPMailer\PHPMailer\SMTP;
       /**
        * Send the email. If an error occurs, trap it and display it in the messageStack
        */
-      if (!$mail->send()) {
+      $success = false;
+      try { 
+         $success = $mail->send(); 
+      } catch (Exception $e) {
+      }
+      if (!$success) { 
         $msg = sprintf(EMAIL_SEND_FAILED . '&nbsp;'. $mail->ErrorInfo, $to_name, $to_email_address, $email_subject);
         if ($messageStack !== NULL) {
           if (IS_ADMIN_FLAG === true) {


includes/functions/functions_email.php
```diff
@@ -359,7 +359,21 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
...
+             if (IS_ADMIN_FLAG === true) {
+               $messageStack->add_session($msg, 'error');
+             } else {
+               $messageStack->add('header', $msg, 'error');
+             }
+           } else {
+             error_log($msg);
+           }
+           $ErrorInfo .= ($mail->ErrorInfo != '') ? $mail->ErrorInfo . "\n" : '';
+         }
```
drbyte, line 375: Is all this duplicated code needed? If it's throwing an exception when it fails to send, then won't it immediately fall back to the `catch` block, making this duplicated code redundant?

drbyte: Edit: It looks like it still returns a boolean, which we will also need to check.

Can we simplify this so there's no copy/paste of the same logic? Perhaps just put the $mail->send() in the try block, and leave all the failure-checking after the closed (empty) catch block?

scottcwilson: I know there's duplication - just wanted to start by getting the concept out there.  We could try your idea.

includes/functions/functions_email.php
```diff
@@ -359,7 +359,13 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
...
        * Send the email. If an error occurs, trap it and display it in the messageStack
        */
-      if (!$mail->send()) {
+      $rc = 1;
+      try { 
+         $rc = $mail->send(); 
+      } catch (Exception $e) {
+         $rc = 0; 
+      }
+      if (!$rc) {
```
drbyte, line 368: Definitely better.

How do you feel about simplifying it further?
Something like initializing to false before calling:

$success = false;
try ... $success = $mail->send()

and then catch requires no code

and then if (!$success) ...

scottcwilson: Great.  This is small and easy to backport.

scottcwilson:
Suggested for backport if there is a 1.5.7c.

PR #4118, opened 2020-12-05
fix empty quote of enabled shipping module

proseLA:
just because a shipping module is ENABLED does not mean the module will return a quote. 

adding a tax element to an empty quote creates more problems down the line.

### DIFF
--- a/includes/classes/shipping.php
+++ b/includes/classes/shipping.php
@@ -164,7 +164,7 @@ class shipping extends base {
         if (FALSE == $GLOBALS[$include_quotes[$i]]->enabled) continue;
         $save_shipping_weight = $shipping_weight;
         $quotes = $GLOBALS[$include_quotes[$i]]->quote($method);
-        if (!isset($quotes['tax'])) $quotes['tax'] = 0;
+        if (!isset($quotes['tax']) && !empty($quotes)) $quotes['tax'] = 0;
         $shipping_weight = $save_shipping_weight;
         if (is_array($quotes)) $quotes_array[] = $quotes;
       }


drbyte:
Ref #2585

proseLA:
it is curious, as here:
https://docs.zen-cart.com/dev/code/modules/shipping_modules/#quote
the tax element is not required; but we then check for it and add it if not there.  none of the ZC included shipping modules (that i looked at) return a 0 rate.

as to following down the checkout path as to why @scottcwilson added the tax element to the quote, i'm sure he found something.  these rabbit holes can get pretty deep, and my time is limited.

design wise, i am not sure what's best and most easily enforced; clearly this PR makes sense (to me) and corrects the small oversight of 2585.

scottcwilson:
The issue I was fixing was caused by one of the templates checking for tax, which was not set.  Because template files are often (sadly) copied and never again updated, the best way to resolve this for all was to just add the array element.

proseLA:
> The issue I was fixing was caused by one of the templates checking for tax, which was not set. Because template files are often (sadly) copied and never again updated, the best way to resolve this for all was to just add the array element.

sadly, i do not agree.

- the issue was not resolved for all.  it introduced a new bug that this PR addresses.
- i am not a fan of changing core to accommodate outdated templates.
- your change specifically contradicts the documentation; by adding an element that the docs say is not required.
- templates not working and needing to be updated when updating core is part of an update process, IMO.  

at some point, it would be nice to see changes that move ZC more towards headless e-commerce as opposed to addressing old templates.  i guess that's the age old story here...  making sure old plugins, templates, etc work with the new core.

PR #4119, opened 2020-12-05
fix selected quote not getting set

proseLA:
looking at this code, i see the arrow anti-pattern; and it gets a bit confusing.  with more time, i would consider a refactor.

however, just looking at the code, one can see how `$selected_quote` might not get set, which has happened to me with my existing datasets.

setting the shipping to the same values as if there was an error in the selected quote seems like the best option for now, and corrects the warnings that i am seeing.


drbyte:
🤞

PR #4120, opened 2020-12-07
Log customer IP for fraud tracking

drbyte:
Log customer IP for fraud/spam/abuse tracking, and delete if customer requests to be forgotten.

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -141,6 +141,10 @@ class Customer extends base
                 WHERE customers_info_id = " . (int)$customer_id;
         $db->Execute($sql);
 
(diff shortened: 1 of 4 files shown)


includes/classes/Customer.php
```diff
@@ -141,6 +141,10 @@ public function login(int $customer_id, $restore_cart = true): bool
                 WHERE customers_info_id = " . (int)$customer_id;
         $db->Execute($sql);
 
+        $sql = "UPDATE " . TABLE_CUSTOMERS . "
```
mc12345678, line 144: This query never gets run.

scottcwilson: Thanks.  Opened issue #6237

PR #4128, opened 2020-12-11
Prevent empty shipping module tax_class property being used for comparison

### LINKED ISSUE #4124: shipping module not getting set, and then used for comparison
https://github.com/zencart/zencart/blob/06937378d8b0a56ac623bd33a5a4dfe46d66ee55/includes/modules/order_total/ot_shipping.php#L65-L81

on line 65, we are setting `$module`; based on the code, it seems possible that it can get set to an empty string.

on line 81, we are now using that potentially empty string as an index within `$GLOBALS`.

we need to check with an `isset` prior to checking the `tax_class` to avoid a PHP notice.

i have seen this behavior out in the wild, with `$module` being set to an empty string.

proseLA:
Fixes #4124

### DIFF
--- a/includes/modules/order_total/ot_shipping.php
+++ b/includes/modules/order_total/ot_shipping.php
@@ -78,7 +78,7 @@ class ot_shipping extends base
                 $shipping_tax_description
             );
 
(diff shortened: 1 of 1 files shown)


proseLA:
i am just the worst at setting titles.  i'll try and be better.

PR #4129, opened 2020-12-13
Pull all fields in query for allow_add_to_cart

scottcwilson:
A common customization of this function is to check other fields.  Let's make it easier to do this by pulling the whole row in.

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -763,25 +763,28 @@ function zen_get_products_allow_add_to_cart($product_id)
 {
     global $db, $zco_notifier;
 
-    $sql = "SELECT * FROM " . TABLE_PRODUCTS . " WHERE products_id=" . (int)$product_id;
-    $type_lookup = $db->Execute($sql);
+    $sql = "SELECT p.*, pt.allow_add_to_cart FROM " . TABLE_PRODUCTS . " p
+            LEFT JOIN " . TABLE_PRODUCT_TYPES . " pt ON (p.products_type = pt.type_id)
+            WHERE products_id=" . (int)$product_id;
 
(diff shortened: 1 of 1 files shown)


drbyte:
While I've no strong objection to the query change, I notice that except for one place in core code, all other calls to this function are done from a spot where all the fields are already available, so in theory this (essentially duplicate) query could even be avoided.

But of course that requires a handful of changes.

The function's original purpose was to enforce broader product-type rules when allowing add-to-cart, and to filter GVs.

What sort of customizations to this function are you seeing in the wild? 
How is the original purpose being altered/extended? What sort of logic is being added?
Are customizations related to default fields in the db, or to custom fields?

proseLA:
i am a fan of this change BUT only if the `$type_lookup` var is then passed to the notifier.

my current reasoning/use case scenario goes something like this:

- clients sell products that once sold rarely come back into stock.
- during the life of the products in-stock status, it achieves an SEO value
- once out of stock, rather than lose that SEO value, the product stays 'active' on the customer side, and returns a 200 status code
- i use this notifier to determine if the product can be sold, ie added to the cart.  that is based on logic from fields within the products table, ie status and inventory.
- i can then use the fields from the previous query to eliminate a DB lookup from my observer code.
- this PR (how i see it) does nothing to help ZC customization without modifying core ZC code.
- adding the result of the query to the notifier addresses that issue.

drbyte:
I was thinking similarly in terms of the notifier.

drbyte:
I've reworked the function for some brevity, and also expanded the notifier. Since the notifier is new to this same branch I'm not worried about breaking changes to it.

I'm still interested in the ways @scottcwilson is seeing this customized in the wild, to determine whether there is more to be done here.

> What sort of customizations to this function are you seeing in the wild?
How is the original purpose being altered/extended? What sort of logic is being added?
Are customizations related to default fields in the db, or to custom fields?

scottcwilson:
This is perfect.   My customizations would be checking additional fields that specific implementations have added, which can now be easily done.

PR #4136, opened 2020-12-16
Salemaker - replace zen_image with css buttons

proseLA:
while i was here....

i do not know why, but i greatly prefer css buttons to those zen_image gifs.  seems like these 2 were hanging around...

### DIFF
--- a/admin/salemaker.php
+++ b/admin/salemaker.php
@@ -519,7 +519,7 @@ if (zen_not_null($action)) {
                 $contents = array('form' => zen_draw_form('sales', FILENAME_SALEMAKER, 'page=' . $_GET['page'] . '&action=deleteconfirm') . zen_draw_hidden_field('sID', $sInfo->sale_id));
                 $contents[] = array('text' => TEXT_INFO_DELETE_INTRO);
                 $contents[] = array('text' => '<br><b>' . $sInfo->sale_name . '</b>');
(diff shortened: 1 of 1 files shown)


admin/salemaker.php
```diff
@@ -519,7 +519,7 @@ function SetCategories() {
                 $contents = array('form' => zen_draw_form('sales', FILENAME_SALEMAKER, 'page=' . $_GET['page'] . '&action=deleteconfirm') . zen_draw_hidden_field('sID', $sInfo->sale_id));
                 $contents[] = array('text' => TEXT_INFO_DELETE_INTRO);
                 $contents[] = array('text' => '<br><b>' . $sInfo->sale_name . '</b>');
-                $contents[] = array('align' => 'center', 'text' => '<br>' . zen_image_submit('button_delete.gif', IMAGE_DELETE) . '&nbsp;<a href="' . zen_href_link(FILENAME_SALEMAKER, 'page=' . $_GET['page'] . '&sID=' . $sInfo->sale_id) . '">' . zen_image_button('button_cancel.gif', IMAGE_CANCEL) . '</a>');
+                $contents[] = array('align' => 'center', 'text' => '<br>' . '<input type="submit" name="submit" value="Delete" id="delete_btn" class="btn btn-danger">' . '&nbsp;<a href="' . zen_href_link(FILENAME_SALEMAKER, 'page=' . $_GET['page'] . '&sID=' . $sInfo->sale_id) . '"' . ' class="btn btn-warning" role="button">' . IMAGE_CANCEL . '</a>');
```
drbyte, line 522: The constant should be used, lest non-English speakers will complain.

PR #4142, opened 2020-12-22
Polyfill PHP functions for mismatched PHP versions

drbyte:
This provides (many, but not all) core PHP functions introduced between 5.6 and 8.0 for the purpose of backward-compatibility.

Revised implementation of #4141

### DIFF
--- a/admin/includes/application_bootstrap.php
+++ b/admin/includes/application_bootstrap.php
@@ -97,6 +97,7 @@ if (file_exists('includes/defined_paths.php')) {
     die('ERROR: /includes/defined_paths.php file not found. Cannot continue.');
     exit;
 }
+require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'php_polyfills.php';
 /**
  * ignore version-check if INI file setting has been set
  */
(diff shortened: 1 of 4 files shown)


torvista:
So, I assume with this in place it would be preferable to develop with php 8 from now, but to spot any more anomalies, not to use new features?

torvista:
BUGS:
1) admin
include DIR_FS_CATALOG . '/includes/functions/extra_functions/php_polyfills.php';

DIR_FS_CATALOG already has a closing /

2) The target file is not in extra_functions.

drbyte:
> BUGS:
> 
> 1. admin
>    include DIR_FS_CATALOG . '/includes/functions/extra_functions/php_polyfills.php';
> 
> DIR_FS_CATALOG already has a closing /
> 
> 1. The target file is not in extra_functions.

I think you're looking at outdated code.

torvista:
No. It's current.
 https://github.com/zencart/zencart/blob/08642cb289c4e3351929d652170c7dfc4c9c2b38/admin/includes/functions/extra_functions/load_php_polyfills.php#L8

 PHP Warning: include(//includes/functions/extra_functions/php_polyfills.php): failed to open stream: No such file or directory in /admin/includes/functions/extra_functions/load_php_polyfills.php on line 8.


And regarding the double slashes in the path, it appears that although incorrect, include/requires/is_dir still works.

scottcwilson:
Fixed in #4151.

drbyte:
That file was (supposed to be) deleted in a subsequent commit.
The polyfill now loads earlier in the boot cycle.

PR #4146, opened 2020-12-25
Rename page: "advanced_search" to "search"

drbyte:
Benefits:
- makes URL shorter
- makes breadcrumb text shorter
- makes H1 shorter
- all these shortenings make display on mobile devices simpler
- adds somewhat better intuitivity to the page name, and what "module" deals with search and results

Kept FILENAME_ defines for backward compatibility


lat9:
Doesn't this change have the possibility to 'confuse' URL obfuscators (e.g. USU and Ceon URI Mapping)?

Even without the obfuscators, aren't you (when the 'advanced_' prefix is removed) going to be causing a bunch of 404 errors where customers have bookmarked a specific search page?

drbyte:
I guess we could add back those old modules/pages/ directories with a header_php.php file that simply does a redirect to the correct page. No templates or language files required for that.

Issue #4147, opened 2020-12-28
Fatal Error on any product


torvista:
ZC158 as per 26/12/2020
[28-Dec-2020 15:45:24 UTC] Request URI: /zencart-158/index.php?main_page=product_info&cPath=22&products_id=34, IP address: 127.0.0.1
#1  trigger_error() called at [...zencart-158\includes\classes\db\mysql\query_factory.php:656]
#2  queryFactory->show_error() called at [...zencart-158\includes\classes\db\mysql\query_factory.php:623]
#3  queryFactory->set_error() called at [...zencart-158\includes\classes\db\mysql\query_factory.php:265]
#4  queryFactory->Execute() called at [...zencart-158\includes\functions\functions_products.php:927]
#5  zen_products_lookup() called at [...zencart-158\includes\modules\pages\product_info\jscript_dynamic_price_updater.php:40]
#6  require(...zencart-158\includes\modules\pages\product_info\jscript_dynamic_price_updater.php) called at [...zencart-158\includes\templates\responsive_classic\common\html_header.php:181]
#7  require(...zencart-158\includes\templates\responsive_classic\common\html_header.php) called at [...zencart-158\index.php:42]
--> PHP Fatal error: 1064:You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '.products_id = p.products_id AND pd.language_id = 1)
                       ...' at line 3 :: SELECT products_qty_box_status AS lookup_field
                              FROM products p
                              INNER JOIN products_description pd USING (pd.products_id = p.products_id AND pd.language_id = 1)
                              WHERE  p.products_id = 34 ==> (as called by) ...zencart-158\includes\functions\functions_products.php on line 927 <== in ...zencart-158\includes\classes\db\mysql\query_factory.php on line 656.

PR #4148, opened 2020-12-28
fix sql on zen_products_lookup

proseLA:
addresses #4147

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -923,8 +923,8 @@ function zen_products_lookup($product_id, $what_field = 'products_name', $langua
 
     $product_lookup = $db->Execute("SELECT " . zen_db_input($what_field) . " AS lookup_field
                               FROM " . TABLE_PRODUCTS . " p
-                              INNER JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd USING (pd.products_id = p.products_id AND pd.language_id = " . (int)$language .")
-                              WHERE  p.products_id = " . (int)$product_id);
+                              INNER JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd USING (products_id) 
+                              WHERE  p.products_id = " . (int)$product_id . " AND pd.language_id = " . (int)$language);
     if ($product_lookup->EOF) return '';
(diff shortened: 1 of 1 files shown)


drbyte:
The better fix is to change "USING" to "ON".
Syntax is:
"USING (field)" where field-name matches between tables. Handy when only one field is joined.
or 
"ON (t1.fieldname=t2.fieldname AND othercondition)"

proseLA:
why is `ON` better than `USING` in this situation?  is there a performance hit or a different result set?

i'm curious.

drbyte:
Slight performance gain by having the AND condition for language linked to the JOIN on the table where the language field is present.

```
SELECT products_id AS lookup_field
FROM " . TABLE_PRODUCTS . " p
INNER JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd USING (products_id) 
WHERE  p.products_id = " . (int)$product_id . " AND pd.language_id = " . (int)$language;
```
Here the join brings back all product descriptions for the `products_id`, including records for all languages.
Then later the WHERE clause filters out the extras.

```
SELECT products_id AS lookup_field
FROM " . TABLE_PRODUCTS . " p
INNER JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (pd.products_id = p.products_id AND pd.language_id = " . (int)$language .")
WHERE  p.products_id = " . (int)$product_id;
```
Here the join brings back only the product description records matching both `products_id` and `language_id`, which means the filtering is done earlier in the process.

Depending on the dataset this could be significantly faster.

proseLA:
thanks.

drbyte:
I should add: we make two common exceptions to this practice.

1. Sometimes we leave the second condition in a WHERE clause if that WHERE clause is likely to get appended-to with some AND statements in subsequent code. This makes it simpler to let the code concatenate more AND statements without having to guess whether WHERE clause already exists.

2. Sometimes we use `ON(t1.field=t2.field)` instead of `USING(field)` if subsequent code is likely to inject more tables into the join, which could confuse the query optimizer about what table the `field` should join on. Adding the specificity allows surrounding code to be slightly more fluid. While the ON syntax is more verbose, the specificity prevents ambiguity.

PR #4158, opened 2021-01-03
Remove inspected entry from whos_online URI if session no longer exists

mc12345678:
If tracking a particular session, when the session disappears or is removed
it remains to be tracked; however, no longer exists. Further, if the session
being tracked is incorrectly entered, it will appear ass if the session
does not have a cart even if there is one.

This becomes a visible problem if there are two or more sessions active,
one of them is being tracked and the page is set to refresh periodically.
When that session expires and eventually only one session exists, at that
point, it is not possible to click on a session to force tracking it.

This clears the tracking data if it doesn't exist.

### DIFF
--- a/admin/whos_online.php
+++ b/admin/whos_online.php
@@ -39,14 +39,14 @@ $selectedView = isset($_GET['q']) ? $_GET['q'] : '';
 $wo = new WhosOnline();
 if (!empty($_GET['inspect'])) {
     $sql = "SELECT s.sesskey
-          FROM " . TABLE_SESSIONS . " s
-          WHERE s.sesskey = :session_id:";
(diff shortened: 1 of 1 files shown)


mc12345678:
FYI, did come across another situation that is not corrected by the above... If there are for example two sessions active, one from an admin and one from a non-admin IP address. When the admin records are made available for viewing, admin record is selected/active, and then the admin option(s) are hidden, well... The session still exists, but it does not exist in the display... Then because there is only one record showing on screen, it is not selectable, and "again" the shopping_cart data is not made available because nothing has "removed" the extra link data... 

I expect correction to the above to be on a separate PR, but wanted to give some up front background.

PR #4159, opened 2021-01-03
Whos-online: Inspect all sessions if no specific session has been provided.

mc12345678:
Reduces the need to provide/attempt all other sessions and doesn't limit
to the first session that is returned. When a session is provided then
as each session is evaluated then when reviewing the cart information
only that session's cart data will be included.

### DIFF
--- a/admin/includes/classes/WhosOnline.php
+++ b/admin/includes/classes/WhosOnline.php
@@ -133,12 +133,17 @@ class WhosOnline extends base
 
         $ip_array = [];
 
+        $sessionInspectAll = false; // Limit the session inspection to just the one $sessionToInspect
         // make the first entry show cart by default
         if (empty($sessionToInspect) && !$results->EOF) {
             $sessionToInspect = $results->fields['session_id']; // uses the fields array because we're not iterating the array yet
+            $sessionInspectAll = true; // No specific session was requested, so make all available.
         }
 
         foreach ($results as $result) {
+            if ($sessionInspectAll === true) {
+                $sessionToInspect = $result['session_id'];
+            }
             $result['time_online'] = (time() - $result['time_entry']);
             $result['is_a_bot'] = empty($result['session_id']);
             $result['time_since_last_click'] = $this->getHumanFriendlyTimeSince($result['time_last_click']);


admin/includes/classes/WhosOnline.php
```diff
@@ -133,12 +133,17 @@ public function retrieve($selectedView = '', $sessionToInspect = '', $exclude_sp
...
 
+        $sessionInspectAll = false; // Limit the session inspection to just the one $sessionToInspect
         // make the first entry show cart by default
         if (empty($sessionToInspect) && !$results->EOF) {
             $sessionToInspect = $results->fields['session_id']; // uses the fields array because we're not iterating the array yet
+            $sessionInspectAll = true; // No specific session was requested, so make all available.
         }
 
         foreach ($results as $result) {
+            if ($sessionInspectAll === true || !in_array($sessionToInspect, $results)) { // Use the session data from each session.
```
drbyte, line 144: I'm confused about the intention behind `|| !in_array($sessionToInspect, $results)`.

and can you further clarify the comment ` // Use the session data from each session.`?

mc12345678: It was something I had tried to add in before realizing that there was an alternate way to address a non-existent session ***and*** that there was no need to individually track that a session had previously been "found" as they are expected to relatively be unique (or at least that was what I was thinking when I didn't do further development)...

On the comment, could you explain a little further about what clarification is desired? The idea behind that if statement is that if all sessions are to be "inspected" then within this loop either the $sessionToInspect value needs to be updated or
````php
            if (!empty($sessionToInspect) && $sessionToInspect === $result['session_id']) {
                $result['cart'] = $this->inspectSessionCart($result['session_id'], $result['session_data']);
            }
````
the code within the above test needs to be executed (which in a way is what the added code above "supports" because there is no code needed for additional empty session "testing".)

Perhaps focus more on collecting cart data? I've been working to fix the User Tracking plugin which review of this supported, so not entirely sure what the best verbiage is. I'm pretty sure I used the wrong words, but wanted to provide some semblance of explanation and as always welcome constructive feedback.

drbyte: If the `if` statement is simplified to just inspectAll === true, then there's probably no need for the comment.

I found the wording "from each session" confusing because we're already in a foreach, so why do we need to explain "each". 

Can the comment be skipped without losing necessary meaning? (espec. given the other two comments which already describe the new variable's purpose)

mc12345678: Upon additional looking, fully agree... Comment can go away (choice of variable name offers some clarity) and yes that was the modification for the `if` statement.

Forced push coming up...

PR #4162, opened 2021-01-07
Refactor ot_coupon to expose validation methods

drbyte:
This refactoring segments several big long blocks of code into smaller functions.

The end result is that it now allows for some external processes to use the module's coupon validation rules.

For example, it would be fairly easy to capture a `?coupon=FOO` from the URL and ask ot_coupon whether it is valid, and if it's valid then apply it to the Session (eg, cart) automatically.
Handy for emailings or in-site clickable links.

**Backward compatible with v1.5.7** save for a small change in the data passed to the `NOTIFY_OT_COUPON_CALCS_FINISHED` notifier that was added in 1.5.7.

### DIFF
--- a/includes/modules/order_total/ot_coupon.php
+++ b/includes/modules/order_total/ot_coupon.php
@@ -8,6 +8,10 @@
  * @version $Id: Modified in v1.5.8 $
  */
 
+/*
+ * NOTE: Notifier NOTIFY_OT_COUPON_CALCS_FINISHED formerly had its first parameter return an array with a 'coupon' entry which was a queryFactoryResult. It is now just an array of the ->fields values
+ */
+
 /**
  * Order Total class to handle discount coupons
  */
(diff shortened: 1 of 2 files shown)


drbyte:
Example of using this to implement `?coupon=FOO` as a $_GET parameter, in say something like `init_headers.php` (or, better, a new file loaded at an appropriate autoload point):

```php

// The 3 instances of $_GET['coupon'] could be replaced with another value for 'coupon', or even a POST

// The messageStack output to 'header' could be redirected to another place

if (!empty(trim($_GET['coupon']))) {
    if (!isset($_SESSION['coupon_redemption_attempts']) || $_SESSION['coupon_redemption_attempts'] < 7) {
        if (!isset($GLOBALS['ot_coupon'])) {
            $lang_file = zen_get_file_directory(DIR_WS_LANGUAGES . $_SESSION['language'] . '/modules/order_total/', 'ot_coupon.php', 'false');
            require_once $lang_file;
            require_once DIR_FS_CATALOG . 'includes/modules/order_total/ot_coupon.php';
            $ot_coupon = new ot_coupon();
        }
        $coupon_id = $ot_coupon->performValidations(trim($_GET['coupon']));
        $errors = $ot_coupon->getValidationErrors();
        if (!empty($errors)) {
            $messageStack->add('header', $errors[0], 'error');
            unset($_GET['coupon']);
        } else if (!empty($coupon_id)) {
            $_SESSION['cc_id'] = $coupon_id;
            $messageStack->add('header', TEXT_VALID_COUPON, 'success');
        }
    }
}
```

Issue #4153, opened 2020-12-29
zen_get_products_sale_discount_type return value causes PHP warning to be issued


lat9:
**Describe the bug**
When going through checkout with the following demo products in the cart
![image](https://user-images.githubusercontent.com/2685585/103305697-cb271880-49d9-11eb-82dd-a06ca8dbff6f.png)

I'm seeing the following PHP warning being issued
```
[29-Dec-2020 18:03:04 UTC] Request URI: /zc157a_bs4/index.php?main_page=checkout_shipping, IP address: 127.0.0.1
#1  shoppingCart->calculate() called at [C:\xampp\htdocs\zc157a_bs4\includes\classes\shopping_cart.php:1660]
#2  shoppingCart->free_shipping_items() called at [C:\xampp\htdocs\zc157a_bs4\includes\functions\functions_general.php:915]
#3  zen_get_shipping_enabled() called at [C:\xampp\htdocs\zc157a_bs4\includes\modules\shipping\freeshipper.php:24]
#4  freeshipper->__construct() called at [C:\xampp\htdocs\zc157a_bs4\includes\classes\shipping.php:68]
#5  shipping->__construct() called at [C:\xampp\htdocs\zc157a_bs4\includes\modules\pages\checkout_shipping\header_php.php:101]
#6  require(C:\xampp\htdocs\zc157a_bs4\includes\modules\pages\checkout_shipping\header_php.php) called at [C:\xampp\htdocs\zc157a_bs4\index.php:35]
--> PHP Warning: A non-numeric value encountered in C:\xampp\htdocs\zc157a_bs4\includes\classes\shopping_cart.php on line 737.
```

**Version**
Zen Cart version: 157b
PHP version:  7.4.x
MySQL version: n/a
Plugins that may be related: n/a
What older/newer versions were also tested? zc158

**To Reproduce**
Add the above products to the cart, then go to checkout.  You'll see the logs generated.

**Expected behavior**
No logs generated

**Screenshots**
See above

**Debug/Error Logs**
Ditto
 
**Devices (please complete the following information):**
 - Operating System: [e.g. Windows, Mac, Android, iOS]
 - Browser [e.g. chrome, firefox] and version number

**Additional context**
Traced it down to the following lines in the subject function in `includes/functions/functions_prices.php`:
```
/*
    $deduction_type_array = array(array('id' => '0', 'text' => DEDUCTION_TYPE_DROPDOWN_0),
[...]

PR #4166, opened 2021-01-08
Correct non-numeric value.

mc12345678:
When using PHP 7.4+ and attempting to perform mathematical operations against an empty string (`''`), a warning message is generated about a non-numeric value being encountered as a result of the returned value from the price function `zen_get_discount_calc`.  Although the thread where this was recently reported identifies problems with the output from that function, there are mathematical operations performed within the function that could also cause similar messages because the empty quote may be multiplied by a number which is expected to also generate a warning.

This appears to resolve the issue identified at this specific post: https://www.zen-cart.com/showthread.php?227580-A-non-numeric-value-encountered-in-shopping_cart-php&p=1376008#post1376008

### DIFF
--- a/includes/functions/functions_prices.php
+++ b/includes/functions/functions_prices.php
@@ -607,12 +607,12 @@
 
     $discount_type_id = zen_get_products_sale_discount_type($product_id);
 
+    $special_price_discount = 0;
     if ($new_products_price != 0) {
       $special_price_discount = ($new_special_price != 0 ? ($new_special_price/$new_products_price) : 1);
-    } else {
-      $special_price_discount = '';
-    }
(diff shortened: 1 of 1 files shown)


mc12345678:
If not mistaken, isn't this already effectively addressed in #4153? Sorry if so and also if so covered/fixed then this could be closed.

drbyte:
Aha! Yes! That's where I'd seen it. And that's why I thought we'd fixed it.
And it's also why I wasn't seeing it in my tests.

Still, I think your suggested fix is also beneficial.

mc12345678:
Not where I can easily follow the additional code, so asking here, in the additional merges did the admin side also get updated?

Issue #4066, opened 2020-11-10
zen_draw_products_pull_down: output order and format for Admin


torvista:
Looking at products to categories, the category selection and the product selection...

I note the category selection uses admin\includes\modules\products_previous_next.php but subsequently the products dropdown uses function zen_draw_products_pull_down which is shared with the shopfront.

I was thinking how to modify/choose this on the page (tiny text-link "order by....") to allow choice of ordering and relevant format on the listing, for example.
eg
if sorted by id: "#123 - ProductName - ProductModel"
if sorted by name: "ProductName - ProductModel ( #123)"
if sorted by model: "ProductModel  - ProductName ( #123)"
etc. ad infinitum
I don't want to enter the "I don't use a model" argument now, just mooting the idea.

But, I've concluded that it is too convoluted to try and modify the shared function zen_draw_products_pull_down to serve both shopfront and admin as the users have different needs and so have used a new one for my purposes.

So suggestions to shape this into a PR with a new function, or what?

drbyte:
> uses function zen_draw_products_pull_down which is shared with the shopfront.

Where is it used by the shopfront?

torvista:
I had assumed since it was moved from Admin, it was being shared with shopfront or was planned to be.

drbyte:
It appears as though it's been available in both admin and catalog for many releases.
But I see no use if it catalog-side.
In fact all of the pulldown functions are only used admin-side.

The functions were moved catalog-side so they could be as centralized as possible. The thinking wasn't primarily around "are they used in the admin vs catalog" but rather "these are category-related functions so they'll go into functions_categories.php ... and almost all of these functions files are stored catalog-side".

drbyte:
Also note my proposal in #4083

proseLA:
> It appears as though it's been available in both admin and catalog for many releases.
> But I see no use if it catalog-side.

i see no use on the catalog side as well.

torvista:
Is a choice of format not of interest generally? Or am I the only one whose shop management revolves around SKUs/models?

scottcwilson:
New class `includes/classes/productPulldown.php` should suffice.  If additional notifiers are required, please PR.

PR #4170, opened 2021-01-11
Refactor prod cat pulldowns into class file.

proseLA:
in looking at the 2 category and 2 product functions, i decided to refactor them into a small class file.  i'm not hugely tied to any of it, but it makes much more sense to me.

persuant to our conversation in #4083, we talked a bit about separating the db functions from the decorating.  i'm not sure if by decorating you were referring to the html parameters; or the names in the pulldown.  the names in the pulldown are harder to separate without doing unnecessary db hits.

in any event, this can easily be tested with the following code:

```
<?php
	require 'includes/application_top.php';
	require 'includes/classes/prod_cat_pulldown.php';

	echo zen_draw_products_pull_down_categories('test');
	echo __FILE__ . ':' . __LINE__ . '<br>';

	$pull_down = new prod_cat_pulldown('category');
	echo $pull_down->pull_down('test');
	echo __FILE__ . ':' . __LINE__ . '<br>';

	echo zen_draw_products_pull_down_attributes('test2');
	echo __FILE__ . ':' . __LINE__ . '<br>';

	$pull_down = new prod_cat_pulldown();
	echo $pull_down->pull_down('test2');

	die(__FILE__ . ':' . __LINE__);
```
you can play with any of the vars in the `__construct` or pass most of them when creating the instance of the class.

### DIFF
--- a/includes/classes/productPulldown.php
+++ b/includes/classes/productPulldown.php
@@ -1,0 +1,129 @@
+<?php
+
+    class productPulldown extends pulldown
+    {
+        private $keyed_allowed_sort_array = [
+            'products_name' => 'pd',
+            'products_model' => 'p',
+            'products_name' => 'p',
+            'products_id' => 'p',
+            'products_price' => 'p',
+            'products_sort_order' => 'p',
+        ];
+
+        function __construct()
+        {
+            parent::__construct();
+
+            $this->show_model = false;
+            $this->show_price = true;
+            $this->set_selected = 0;
+            $this->categories_join = '';
+
+            $this->sort = ' ORDER BY pd.products_name';
+
+            $this->keyword_search_fields = [
+                'pd.products_name',
+                'p.products_model',
+                'pd.products_description',
+                'p.products_id',
+            ];
+        }
+
+        public function setSort($fieldnameArray)
+        {
+            if (empty($fieldnameArray)) {
+                return $this;
+            }
+
+            $first = true;
+            $this->sort = '';
+            foreach ($fieldnameArray as $fieldname) {
+                if (array_key_exists($fieldname, $this->keyed_allowed_sort_array)) {
+                    $this->sort = ($first ? ' ORDER BY ' : ', ') . $this->keyed_allowed_sort_array[$fieldname] . '.' . $fieldname;
+                    $first = false;
+                }
+            }
+            return $this;
+        }
+
+        public function setCategory(int $category_id)
+        {
+            $this->categories_join = " LEFT JOIN " . TABLE_PRODUCTS_TO_CATEGORIES . " ptc ON (ptc.products_id = p.products_id)";
+            $this->condition .= " AND ptc.categories_id = " . (int)$category_id;
+            return $this;
+        }
+
+        public function showModel(bool $status)
+        {
+            $this->show_model = $status;
+            return $this;
+        }
+
+        public function showPrice(bool $status)
+        {
+            $this->show_price = $status;
+            return $this;
+        }
+
+        public function onlyActive(bool $status)
+        {
+        	$condition = " AND p.products_status = 1";
+        	$this->condition = str_replace($condition, '', $this->condition);
+        	if ($status) {
+		        $this->condition .= " AND p.products_status = 1";
+	        }
+            return $this;
+        }
+
+        protected function setSQL()
+        {
+            $this->sql = "SELECT DISTINCT pd.products_id, p.products_sort_order, p.products_price, p.products_model
+                FROM " . TABLE_PRODUCTS . " p"
+                . $this->categories_join . "
+                INNER JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (p.products_id = pd.products_id) 
+				" . $this->attributes_join . "
+				WHERE pd.language_id = " . (int)$_SESSION['languages_id'];
+        }
+
+
+        protected function processSQL()
+        {
+            global $currencies;
+
+            $this->setSQL();
+            $this->runSQL();
+
+            $parm_2 = '';
+            $parm_3 = '';
+
+            if ($this->show_model) {
+                $parm_2 = '%2$s';
+            }
+
+            if ($this->show_price) {
+                $parm_3 = ' (%3$s)';
+            }
+
+            $this->output_string = '%1$s ' . $parm_2 . $parm_3; // format string with name first
+
+            if (strpos($this->sort, 'model') && (!empty($this->attributes_join))) { // show model first only when attributes exist
+                $this->output_string = (!empty($parm_2) ? $parm_2 . '-' : '') . ' %1$s' . $parm_3; // format string with model first
+            }
+
+
+            foreach ($this->results as $result) {
+                if (in_array($result['products_id'], $this->exclude)) {
+                    continue;
+                }
+                $display_price = $this->show_price ? zen_get_products_base_price($result['products_id']) : '';
+                $name = zen_get_products_name($result['products_id']);
+                $this->values[] = [
+                    'id' => $result['products_id'],
+                    'text' => sprintf($this->output_string, trim(zen_clean_html($name)),
+                            ($this->show_model ? ' [' . $result['products_model'] . '] ' : ''),
+                            $currencies->format($display_price)) . ($this->show_id ? ' - ID# ' . $result['products_id'] : ''),
+                ];
+            }
+        }
+    }
--- a/admin/includes/auto_loaders/config.product_category_pulldowns.php
+++ b/admin/includes/auto_loaders/config.product_category_pulldowns.php
@@ -1,0 +1,13 @@
+<?php
+
+$autoLoadConfig[210][] = array('autoType'=>'class',
+                               'loadFile'=> 'pulldown.php',
+                               'classPath'=>DIR_FS_CATALOG . DIR_WS_CLASSES);
+
+$autoLoadConfig[210][] = array('autoType'=>'class',
+                               'loadFile'=> 'productPulldown.php',
+                               'classPath'=>DIR_FS_CATALOG . DIR_WS_CLASSES);
+
+$autoLoadConfig[210][] = array('autoType'=>'class',
+                               'loadFile'=> 'categoryPulldown.php',
+                               'classPath'=>DIR_FS_CATALOG . DIR_WS_CLASSES);
--- a/includes/classes/categoryPulldown.php
+++ b/includes/classes/categoryPulldown.php
@@ -1,0 +1,81 @@
+<?php
+
+    class categoryPulldown extends pulldown
+    {
+
+        function __construct()
+        {
+            parent::__construct();
+
+            $this->show_parent = false;
+            $this->show_full_path = false;
+
+            $this->sort = " ORDER BY categories_name";
+
+            $this->keyword_search_fields = [
+                'cd.categories_name',
+                'c.parent_id',
+                'cd.categories_description',
+                'c.categories_id',
+            ];
+
+        }
+
+        public function showParent(bool $status)
+        {
+            $this->show_parent = $status;
+            return $this;
+        }
+
+        public function showFullPath(bool $status)
+        {
+            $this->show_full_path = $status;
+            return $this;
+        }
+
+        protected function setSQL()
+        {
+            $this->attributes_join = str_replace('p.products_id', 'ptoc.products_id', $this->attributes_join);
+            $this->sql = "SELECT DISTINCT c.categories_id, cd.categories_name
+            FROM " . TABLE_CATEGORIES . " c
+            LEFT JOIN " . TABLE_CATEGORIES_DESCRIPTION . " cd ON (c.categories_id = cd.categories_id AND cd.language_id = " . (int)$_SESSION['languages_id'] . ")
+            LEFT JOIN " . TABLE_PRODUCTS_TO_CATEGORIES . " ptoc on (c.categories_id = ptoc.categories_id) 
+            " . $this->attributes_join . "
+            WHERE TRUE ";
+        }
+
+        protected function processSQL()
+        {
+            $this->setSQL();
+            $this->runSQL();
+
+            foreach ($this->results as $result) {
+                if (in_array($result['categories_id'], $this->exclude)) {
+                    continue;
+                }
+                $this->values[] = [
+                    'id' => $result['categories_id'],
+                    'text' => $this->categoryText($result),
+                ];
+            }
+        }
+
+        private function categoryText($category)
+        {
+            if (!empty($this->attributes_join)) {
+                $text = $category['categories_name'];
+                if ($this->show_full_path) {
+                    $text = zen_output_generated_category_path($category['categories_id']);
+                }
+                return $text;
+            }
+            $parent = '';
+            if ($this->show_parent) {
+                $parent = zen_get_categories_parent_name($category['categories_id']);
+                if ($parent != '') {
+                    $parent = ' : in ' . $parent;
+                }
+            }
+            return $category['categories_name'] . $parent . ($this->show_id ? ' - ID# ' . $category['categories_id'] : '');
+        }
+    }
--- a/includes/classes/pulldown.php
+++ b/includes/classes/pulldown.php
@@ -1,0 +1,101 @@
+<?php
+
+    abstract class pulldown extends base
+    {
+        var $pulldown, $attributes_join, $show_id, $set_selected, $parameters, $condition, $exclude;
+
+
+        function __construct()
+        {
+            $this->exclude = [];
+
+            $this->show_id = false;
+
+            $this->set_selected = 0;
+            $this->values = [];
+
+            $this->keywords = '';
+
+            $this->attributes_join = '';
+
+            $this->condition = ' ';
+
+            // default styling
+            $this->parameters = '';
+            //$this->parameters = 'required size="15" class="form-control" id="products_id"';
+        }
+
+        public function setDefault(int $id)
+        {
+            $this->set_selected = $id;
+            return $this;
+        }
+
+        public function showID(bool $status)
+        {
+            $this->show_id = $status;
+            return $this;
+        }
+
+        public function setOptionFilter(int $filter_id)
+        {
+            $this->includeAttributes(true);
+            $this->condition .= " AND pa.options_id =" . (int)$filter_id;
+            return $this;
+        }
+
+        public function exclude(array $array)
+        {
+            $this->exclude = $array;
+            return $this;
+        }
+
+        public function includeAttributes(bool $status)
+        {
+        	$this->attributes_join = '';
+        	if ($status) {
+		        $this->attributes_join = " RIGHT JOIN " . TABLE_PRODUCTS_ATTRIBUTES . " pa on (p.products_id = pa.products_id)";
+	        }
+            return $this;
+        }
+
+        public function setSearchTerms(string $keywords)
+        {
+            $this->keywords = $keywords;
+            return $this;
+        }
+
+        abstract protected function processSQL();
+
+        abstract protected function setSQL();
+
+        protected function runSQL()
+        {
+            global $db;
+
+            $this->sql .= $this->condition;
+
+            if (empty($this->keywords)) {
+                $this->keywords = ($_REQUEST['keywords'] ?? '');
+            }
+
+            if (!empty($this->keywords)) {
+                $this->sql .= zen_build_keyword_where_clause($this->keyword_search_fields,
+                    zen_db_input(zen_db_prepare_input($this->keywords)));
+            }
+
+            $this->sql .= $this->sort;
+            $this->results = $db->Execute($this->sql);
+        }
+
+        public function generatePullDownHtml(string $name, $parameters = '', bool $required = false)
+        {
+            $this->processSQL();
+
+            if (empty($parameters)) {
+                $parameters = $this->parameters;
+            }
+
+            return zen_draw_pull_down_menu($name, $this->values, $this->set_selected, $parameters, $required);
+        }
+    }
--- a/includes/functions/functions_categories.php
+++ b/includes/functions/functions_categories.php
@@ -325,60 +325,35 @@ function zen_product_in_parent_category($product_id, $cat_id, $parent_cat_id)
  * @param bool $show_current_category
  * @return string
  */
-function zen_draw_pulldown_products($field_name, $parameters = '', $exclude = [], $show_id = false, $set_selected = 0, $show_model = false, $show_current_category = false)
+function zen_draw_pulldown_products($field_name, $parameters = '', $exclude = [], $show_id = false, $set_selected = 0, $show_model = false, $show_current_category = false, $order_by = '', $filter_by_option_name = null)
 {
-    global $currencies, $db, $current_category_id, $prev_next_order;
+    global $prev_next_order, $current_category_id;
+
+    $only_active = false;
 
     if (!is_array($exclude)) {
         $exclude = [];
     }
 
-    $select_string = '<select name="' . $field_name . '"';
-
-    if ($parameters) {
-        $select_string .= ' ' . $parameters;
+    if (empty($order_by)) {
+        $order_by = $prev_next_order;
     }
 
-    $select_string .= '>';
+    $sort_array = array_filter(explode(',', str_ireplace('order by ', '', trim($order_by))));
 
-    $sql = "SELECT p.products_id, pd.products_name, p.products_sort_order, p.products_price, p.products_model
-            FROM " . TABLE_PRODUCTS . " p
-            INNER JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (p.products_id = pd.products_id)
-            WHERE pd.language_id = " . (int)$_SESSION['languages_id'];
+    $pulldown = new productPulldown();
 
     if ($show_current_category) {
-        // only show $current_categories_id
-        $sql = "SELECT p.products_id, pd.products_name, p.products_sort_order, p.products_price, p.products_model, ptc.categories_id
-                FROM " . TABLE_PRODUCTS . " p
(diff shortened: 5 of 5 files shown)


mc12345678:
FYI, posing some comments/questions. They are in no particular order (either by submission or necessarily placement). I am currently only addressing the code that has been provided and not the overall concept (yet). That said, somewhat wondering if it there may be some "improvement" by having two different classes inheriting from a base class of some sort...

includes/classes/prod_cat_pulldown.php
```diff
@@ -0,0 +1,186 @@
...
+	{
+		var $pulldown, $attributes, $order_by;
+		/**
+		 * @var bool
+		 */
+		private $show_id;
+		/**
+		 * @var array
+		 */
+		private $exclude;
```
mc12345678, line 13: Shouldn't there be handlers for one or both of these private variables? Otherwise the class has to be redefined again to take advantage of a change in the variable instead of simply reperforming the necessary step(s) with the adjusted value.

includes/classes/prod_cat_pulldown.php
```diff
@@ -0,0 +1,186 @@
...
+			array $exclude = [],
+			string $keywords = '',
+			string $order_by = '',
+			bool $attributes = false,
+			$filter_by_option_name = null
+		) {
+
+			$this->exclude = $exclude;
+
+			$this->show_model = true;
```
mc12345678, line 26: Might think that ultimately this would be controlled by something else either a session variable or admin switch... May help reduce support questions (though of course also add some in any case).

includes/classes/prod_cat_pulldown.php
```diff
@@ -0,0 +1,186 @@
...
+			$this->condition = " WHERE pa.options_id =" . (int)$this->option_filter;
+
+			if ($pulldown == 'category') {
+				$this->pulldown = $pulldown;
+				$this->category_vars();
+			} else {
+				$this->product_vars();
+			}
+
+			$this->run_sql();
```
mc12345678, line 54: Wouldn't this be better suited at the point of executing `$this->pull_down()`?
There's a few things that appear could benefit by making that move:
- `$this->keyword_search_fields` could be updated/modified to suit,
- `$this->exclude` could be updated/modified (with appropriate handler).
- Ultimately it seems that this class could be pre-defined (instantiated) in page load and not require to be uniquely declared for execution. Would mean that likely some of these parameters may go to the `pull_down` method, but... Then when wanting to use this class only one line of code is needed rather than two (each of which could have problems in it).
- If one wants to obtain differently filtered results, may not have to reinstantiate the variable and display the results.
-

includes/classes/prod_cat_pulldown.php
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		private function run_sql()
+		{
+			global $db, $currencies;
+
+			$search = '';
+			if (!empty($this->keywords)) {
+				$search = zen_build_keyword_where_clause($this->keyword_search_fields,
+					trim(zen_db_input(zen_db_prepare_input($this->keywords))));
```
mc12345678, line 115: What does this outer trim accomplish that is not already done?
````php
function zen_db_prepare_input($string, bool $trimspace = true)
{
    if (is_string($string)) {
        if ($trimspace == true) {
            return trim(stripslashes($string));
        } else {
            return stripslashes($string);
        }
    } elseif (is_array($string)) {
        foreach ($string as $key => $value) {
            $string[$key] = zen_db_prepare_input($value);
        }
        return $string;
    } else {
        return $string;
    }
}
````
This based on the requirement of $keywords being a string by the parameter definition means that the content entering `zen_db_input` will already be trimmed, then `zen_db_input` doesn't add spaces, so why trim all of that?

File: includes/classes/prod_cat_pulldown.php [hunk 5]
```diff
@@ -0,0 +1,186 @@
...
+					break;
+				default: //legacy: show all products with attributes
+					$this->sql .= $this->sort;
+					break;
+			}
+
+			$results = $db->Execute($this->sql);
+
+			foreach ($results as $result) {
+				if (!in_array($result['common_id'], $this->exclude)) {
```
mc12345678, line 136: Recommend making this an "early return" characteristic. Instead of grouping everything within this `if`, can test and if met then `continue` the `foreach`:
````
if (in_array($result['common_id'], $this->exclude)) {
    continue;
}
````
And then the remaining code can be left shifted and be outside of this "larger" if group.

File: includes/classes/prod_cat_pulldown.php [hunk 6]
```diff
@@ -0,0 +1,186 @@
...
+								'id' => $result['categories_id'],
+								'text' => $this->category_text($result),
+							];
+							break;
+						default:
+							$display_price = zen_get_products_base_price($result['products_id']);
+							$name = zen_get_products_name($result['products_id']);
+							$this->values[] = [
+								'id' => $result['products_id'],
+								'text' => sprintf($this->output_string, trim(zen_clean_html($name)),
```
mc12345678, line 149: If a space existed before or after the name, wouldn't that have been on purpose and therefore desired? As a result of this operation doing a trim, two names may appear in the same format though as unique values, further there may be code down stream that uses/processes any additional content that is in the `$name`; however, it is removed from this data here.  The code in `zen_draw_pull_down_menu` has two things, one a notifier allowing an alternate output result and also provides a level of the `zen_clean_html` function by replacement of common characters that would cause a problem with the html equivalent.

Basically wondering why the `$name` is being treated so early in processing? Note that similar management is not done on `products_model` nor when dealing with categories, the `categories_name`.

includes/classes/prod_cat_pulldown.php
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		private function category_text($category)
+		{
+			if ($this->attributes) {
+				$text = $category['categories_name'];
+				if ($this->show_full_path) {
+					$text = zen_output_generated_category_path($category['categories_id']);
+				}
+			} else {
```
mc12345678, line 166: Before this closing parentheses, could just return `$text` and then remove the else statement below.

includes/classes/prod_cat_pulldown.php
```diff
@@ -0,0 +1,186 @@
...
+					if ($parent != '') {
+						$parent = ' : in ' . $parent;
+					}
+				}
+				$text = $category['categories_name'] . $parent . ($this->show_id ? ' - ID# ' . $category['categories_id'] : '');
+			}
+			return $text;
+		}
+
+		function pull_down(string $name, $parameters = '')
```
mc12345678, line 179: All of the other methods have their visibility (`private`, `protected`, `public`) identified, for consistency shouldn't this as well?

includes/classes/prod_cat_pulldown.php
```diff
@@ -0,0 +1,186 @@
...
+		{
+			if (empty($parameters)) {
+				$parameters = $this->parameters;
+			}
+			return zen_draw_pull_down_menu($name, $this->values, $this->set_selected, $parameters);
```
mc12345678, line 184: What about the `$required` parameter?

includes/modules/pages/index/header_php.php
```diff
@@ -37,41 +38,41 @@
...
+    
     // -----
     // Give an observer the chance to override the default handling for the category.
     //
     $category_redirect_handled = false;
     $zco_notifier->notify(
-        'NOTIFY_INDEX_CATEGORY_STATUS_CHECK',
-        ['cPath' => $cPath, 'current_category_id' => $current_category_id],
+        'NOTIFY_INDEX_CATEGORY_STATUS_CHECK', 
+        array('cPath' => $cPath, 'current_category_id' => $current_category_id),
```
mc12345678, line 67: Why has this array convention been undone here, but continued to be used in the new code?

If modifying this, would say that the last array element should have a comma that follows.

Actually, don't see why this file is included in this commit at all (meaning nothing is really changed here that applies to what is being done, looks like this was accidentally included in the commit).

proseLA:
> FYI, posing some comments/questions. They are in no particular order (either by submission or necessarily placement). I am currently only addressing the code that has been provided and not the overall concept (yet). That said, somewhat wondering if it there may be some "improvement" by having two different classes inheriting from a base class of some sort...

first off, thanks for taking a look.
second off, I BLAME MY IDE.  :frowning_face: and my complete lack of expertise in its use!
third off, it seems i have messed up this branch a bit, as there only should be the 1 class file.  i have been in a bit of a git quagmire, and it seems some effects have come into this branch.  which i just noticed.  apologies.
fourth, this code was submitted as a starting point in an attempt to refactor 4 zen functions.  in that refactor, we wanted to additional add a search component in the construction of pull down menus.
fifth, i go back to my first comment, thanks for taking a look.  as my time permits, i will now look at all your comments, and attempt to address accordingly.  :+1:

includes/modules/pages/index/header_php.php
proseLA: this is a mistake in my git handling.  again, apologies.  this page should not have been modified at all by this branch.

includes/classes/prod_cat_pulldown.php
proseLA: this pr is an attempt to refactor 4 functions: `zen_draw_products_pull_down`, `zen_draw_products_pull_down_attributes`, `zen_draw_products_pull_down_categories`, and `zen_draw_products_pull_down_categories_attributes`.  none of these 4 functions implement that attribute.  so i choose to go with the default of `false`.  i am not opposed to a discussion as to whether these functions need that to be true; but i would have to examine closer.

includes/classes/prod_cat_pulldown.php
proseLA: sure.

includes/classes/prod_cat_pulldown.php
proseLA: this is true; and a question of style.  i'm a fan of minimizing the number of places that a function exits, ie only having 1 `return`.  but again, i'm not opposed to this idea.

includes/classes/prod_cat_pulldown.php
proseLA: :+1:

includes/classes/prod_cat_pulldown.php
proseLA: :+1:

includes/classes/prod_cat_pulldown.php
proseLA: i acknowledge that various vars used to control this class need work.  whether the addition of new config switches or session vars or something of the like would need to be addressed.  i did not want to get really built up into that infrastructure prior to getting feedback from the ZC team as to whether this approach will get approved or not.

includes/classes/prod_cat_pulldown.php
proseLA: see comment above.

includes/classes/prod_cat_pulldown.php
proseLA: great point!  i like it.  feeds into your 2 class suggestion as well.

includes/classes/prod_cat_pulldown.php
drbyte: > prior to getting feedback from the ZC team as to whether this approach will get approved or not

Still queuing up a time to review code in depth.

includes/classes/prod_cat_pulldown.php
drbyte: FWIW I'm a big fan of early-return, so that it's easier to follow into what actual logic is happening after all exclusions are accounted for or no further processing is required.

includes/classes/prod_cat_pulldown.php
proseLA: with regards to spaces before and/or after a product name, YAGNI.  such an edge case; i can not think of any use case scenario where a store owner has product A and product B, and the name of said products is only differentiated by a space before or a space after the name.

i decided to do a trim on the name after the `zen_clean_html` as in testing on a different related project, the `zen_clean_html` function left some white space at the beginning (or end... i do not remember).

in looking at this, it is possible that i would need to clean the category name as well.  i may have missed that.  i have clients that embed html in their category name for display purposes, and i would have to look at results from their dataset.  as far as cleaning the products model.... perhaps, again an edge case, are there store owners that embed html in model number?

mc12345678:
> > FYI, posing some comments/questions. They are in no particular order (either by submission or necessarily placement). I am currently only addressing the code that has been provided and not the overall concept (yet). That said, somewhat wondering if it there may be some "improvement" by having two different classes inheriting from a base class of some sort...
> 
> first off, thanks for taking a look.
> second off, I BLAME MY IDE. ☹️ and my complete lack of expertise in its use!
> third off, it seems i have messed up this branch a bit, as there only should be the 1 class file. i have been in a bit of a git quagmire, and it seems some effects have come into this branch. which i just noticed. apologies.
> fourth, this code was submitted as a starting point in an attempt to refactor 4 zen functions. in that refactor, we wanted to additional add a search component in the construction of pull down menus.
> fifth, i go back to my first comment, thanks for taking a look. as my time permits, i will now look at all your comments, and attempt to address accordingly. 👍

Noted and after that post I realized I neglected to indicate recognition of at least 1) a lot of work done 2) work in progress. After all, can't get comment/growth if not shared. 😄

includes/classes/prod_cat_pulldown.php
mc12345678: Agree.. Didn't really want to comment on this at this stage of development, but also didn't want it to get fully lost...

includes/classes/prod_cat_pulldown.php
mc12345678: So, the flip side to this though is that the `pull_down` method calls `zen_draw_pull_down_menu`. That function has its own notifier controls to support editing the content; however, the "restricted" display provided here is not included elsewhere in the Zen Cart flow path, thus allowing "unique" data to be present/incorporated/handled.  I mean sure if dead set on removing things like that by using `zen_clean_html`, then would highly recommend incorporating ways that those who actually want the additional content can work "with" this. (Ie. appropriately placed notifiers so that the data may be captured/updated.  Considering so much of this data is being stored in the class, there is really little that has to be passed by the notifier because most of it will already be captured in the class information. May still need to pass something, but...)

Yeah, depending on what else was in that content, the cleaning of the html could leave behind a space at either beginning or end.  My point though is that in general store owners/operators take action to make things look a specific way or contain specific information (those beginning will have a learning curve of some sort, regardless the actual output) and seems that when/if adding/changing things that it still be possible to get the same results even if in a better way and perhaps through more hoops.

As to data control of including/excluding an additional "outer" space, I see that its possible and could be done on purpose and while I don't use it, I don't see why it has to be forced to be removed.

includes/classes/prod_cat_pulldown.php
proseLA: i am not sure if the code is clearer; but i have done the early return as suggested.

drbyte:
In the main, I think this is a good sort of direction to go.

Having a "base" class from which sub-classes are derived would probably also allow for abstracting out some of the "administrative" vs "functional" logic complexities. 

Given the plethora of "parameters" that beg to be passed, I figure it may be good to allow method-chaining in order to "set up" the object after creation.

Maybe have a function specific to formatting output rows based on the provided sort-order (see OP in #4066).

```php
$foo = new ProductPulldown;
$foo->setSort('fieldname')->exclude([ids])->showModel(false)->showCurrent($id)->showId(true)->showPrice(true);
echo $foo->generatePulldownHtml($name, $parameters, $required);
```

proseLA:
i have cleaned up a couple of bugs and tested against the 4 functions that we were looking at refactoring.

i look forward to continuing this conversation.   this refactor does the exact same thing that the functions did prior with additional flexibility.

includes/classes/productPulldown.php
```diff
@@ -56,21 +56,25 @@ public function setCategory(int $category_id)
...
-        public function showModel()
+        public function showModel(bool $status)
         {
-            $this->show_model = true;
+            $this->show_model = $status;
             return $this;
         }
 
-        public function showPrice()
+        public function showPrice(bool $status)
```
drbyte, line 65: Perhaps this thought is over-engineering, since the the implementation here is more likely to actually be passing these `$status` params via variables ...
The thought is: is it worth defaulting each of these to `true` to allow less verbosity if variables aren't used?

eg: `$foo->showModel()->showPrice()`
instead of requiring
`$foo->showModel(true)->showPrice(true)`

includes/classes/productPulldown.php
```diff
@@ -0,0 +1,132 @@
...
+    {
+        private $allowed_sort_array = [
+            'products_name',
+            'products_model',
+            'products_name',
+            'products_id',
+            'products_price',
+            'products_sort_order',
+        ];
+        private $db_table_prefix_array = ['pd.', 'p.', 'p.', 'p.', 'p.', 'p.'];
```
drbyte, line 13: I'm not sure I like the fragility of splitting these off into a separate property like this.

Perhaps `$allowed_sort_array` could include both?
eg: `'products_name' => 'pd.',`
and then do lookups using isset() or array_key_exists or similar?

includes/classes/productPulldown.php
```diff
@@ -0,0 +1,132 @@
...
+            }
+
+            if ($this->show_price) {
+                $parm_3 = ' (%3$s)';
+            }
+
+            $this->output_string = '%1$s ' . $parm_2 . $parm_3; // format string with name first
+
+            if (strpos($this->sort,
+                    'model') && (!empty($this->attributes_join))) { // show model first only when there are attributes
```
drbyte, line 113: eeks, the formatting here seems odd with a line-break mid-function

drbyte:
Not sure how the functions_search file got included in this PR.

proseLA:
> Not sure how the functions_search file got included in this PR.

i think i already had a mea culpa on that here:

https://github.com/zencart/zencart/pull/4289#issue-615294293

i will do a force push and remove it, after looking at your other comments.

includes/classes/productPulldown.php
proseLA: the nuances of settings in my IDE!  i hope to get it under control before the end of the decade...

includes/classes/productPulldown.php
proseLA: i'll take a look.  it makes sense to me.

includes/classes/productPulldown.php
proseLA: i did look at this prior.  i think the problem is what happens when someone wants to hide the price?  or the model?  do you want to have more methods there? such as `hideModel()` and `hidePrice()`?  by including the boolean, users can more easily specify what they want to do.

proseLA:
as an fyi, i do have this code running live on one of my supported sites with no ill effect.

includes/classes/productPulldown.php
drbyte: I see where you're going there.
I was actually thinking `showPrice()` and `showPrice(true)` would both mean true,
and `showPrice(false)` for false.

But, let's just leave it for now. At least with no default it'll force the param to be specified, so the intent will be required.

proseLA: actually the default is in the constructor.  so there is a default. `$this->show_price` is the default; the method `showPrice()` changes the var.

scottcwilson:
Fantastic, thanks for doing this @proseLA and thanks to reviewers @drbyte and @mc12345678.

torvista:
I take it that thanks to this impressive effort, it is now possible to customise the format of the items in the drop downs as per https://github.com/zencart/zencart/issues/4066

While I am the first to criticise people for asking for step-by-step instructions instead of figuring it out for themselves, I think I'm well out of my comfort zone on this one, so could someone please document how to achieve other formats?
Thanks.

proseLA:
@torvista at some point, documentation will come.  this is unreleased software, and as such, use at your own risk.

as the no. 3 contributor to this project, i would encourage you to test it, break it, and make recommendations for fixes.  i'll make a few points about it and then i would encourage feedback of the sort as to what works and what does not.

- the initial inspiration for this was being able to do keyword searches on a product list and then provide a dropdown of said list.  scrolling through a listing of 500+ products to find the one you want to put on special seems ludicrous to me.
- this PR is the infrastructure for the above inspiration.  i would review #4530.  this PR adds that element which allows for searching of any pulldown.
- this PR separates the sql from the generation of the html.  one initially specifies what one wants to see, and then the html gets generated afterwards.  
- i have been thinking about adding some notifiers.  i'm just not sure where at this moment, as it's been a while since i looked at this closely.
- the functions refactored (of which there are only 4) should work exactly (hopefully) the same as before.
- you and i have gone back and forth over the years with regards to sorting by model number (you) or product name (me).  that configuration is right here: 
https://github.com/zencart/zencart/blob/55c8df87b27ae6dafff26a82fcbef3232a3ef1d2/includes/functions/functions_categories.php#L338-L340
- the use of `$prev_next_order` as the default if no sort is specified is all on you.  however, if one looks at the parm list on line 328, one can now specify the sort.
- in case there is any question of this sort, i refer to the blame on this section of code:
https://github.com/zencart/zencart/blob/a57158b8c80f27e61720a5fc9eb24141ff0a180c/admin/includes/functions/general.php#L241-L244
- just above that on line 332 is something that would be part of the previously mentioned notifier.  right now the pulldown defaults to all products.  if you want only active products, change that to `true`.
- all of that 'stuff' you mention in your first comment on #4066 is addressed.   that section of code is here:
https://github.com/zencart/zencart/blob/55c8df87b27ae6dafff26a82fcbef3232a3ef1d2/includes/classes/productPulldown.php#L108-L112
- i have no idea what you are referring to when you say other formats.  the formatting of the html is done here:
https://github.com/zencart/zencart/blob/55c8df87b27ae6dafff26a82fcbef3232a3ef1d2/includes/classes/pulldown.php#L24-L25
- it is another place that i am thinking about having a notifier.
- i think you have enough to see how great or what crap this PR is.  IMO, hand holding is on the forum.  github is the place for code collaboration.

best.

torvista:
Many thanks for taking the time to produce a detailed reply.

> i have no idea what you are referring to when you say other formats

I meant the structure: name-model or model-name as you already detailed above.
Plenty to keep me busy here.

torvista:
Why is the output string line 110  model-first only when attributes exist?

proseLA:
steve,
this code was a refactor of the existing pulldowns at the time.  the new code:

https://github.com/zencart/zencart/blob/55c8df87b27ae6dafff26a82fcbef3232a3ef1d2/includes/classes/productPulldown.php#L108-L112

is a refactor from here:
https://github.com/zencart/zencart/blob/fc92b5e83364fb8ee35d329bb626af494c2eeaf3/admin/includes/functions/general.php#L281
and here:
https://github.com/zencart/zencart/blob/fc92b5e83364fb8ee35d329bb626af494c2eeaf3/admin/includes/functions/general.php#L2351-L2359

it was an attempt to preserve the existing formats of the pulldowns.

torvista:
Thanks but I think I had a punctuation failure....I was querying the clause being dependent on the product having attributes (not the format).

proseLA:
> Thanks but I think I had a punctuation failure....I was querying the clause being dependent on the product having attributes (not the format).

have i answered your question?  now i'm not sure....

torvista:
ha, ha...no I'm still none the wiser as to why the model-first structure is dependent on having attributes:
https://github.com/zencart/zencart/blob/dee55fbd0b13e3c4a35a89d39c6e20ab9e145daf/includes/classes/productPulldown.php#L109

proseLA:
steve,
do you know how to use blame on github?  if not, i would suggest a bit of a review on that feature.  i find it very useful.

as far as the active contributors, it does seem that model is important to you.  model was always after name (since the start of this repo on github).  until your PR #3246 which now moved it to the front position but only when there were attributes.  i was not involved in said PR but i could guess that is what you could convince chris to do at the time.  perhaps something else; i do not know.

now, why the position of the model is that important to you to ask me about it or to continue to make changes to the repo, i do not know.  i suppose the advantage of contributing to a repo is the ability to shape it to one's needs, while at the same time, hopefully contributing to the community.

this refactor of the product pulldowns was chris' suggestion to me after i had inquired about adding search terms to pulldowns.  stores that have hundreds if not thousands of products and having a pulldown seemed ridiculous to me.  i thought the ability to effectively search and restrict pulldowns to search results would be of value to the community.  i had already done it for some of my clients and thought it would be great to share it with the community as i think its a good feature.  maybe someone else agrees.  i do not know.

in any event, chris thought doing method chaining for the pulldowns would be the way to go and i took that ball and ran with it.  my goal in that refactor was to not change the results as they existed in the repo at the time.

these classes are now, perhaps, easily extended and overwritten for ones own use.  i already use them in new ways.

as to whether having the model in front of the name is something the community wants or needs; i can not say.   it does seem important to you, because every time i'm digging into the code and sorting by model, or placement of model comes up, it does seem that your name is on the code.  i suppose if i was supporting a parts store where everything is really keyed off the model number, i would see more of my name on it. but for me, when doing a PR or recommending any change, i'm always balancing between my own needs and that of the community.  and frankly placement of the model number?  lets leave it alone.  searching by model? yes.  sorting by model?  yes.  displaying model? yes.  despite not using model with my clients, i see the value of them and this PR addresses all of those things.... unless someone tells me differently.

best.
IMC.

torvista:
I'm squirming at the amount of time you have courteously invested in replying, and I still have not twigged what the answer is.

But it appears you understand that I am questioning the display of model-name or name-model in the drop-down.
I'm not. 
I see the clause I referenced above, where the model-first structure is used when $this->sort has model in it 'model' and I am aware that this is mainly due to my comments, for which I am grateful. (But then again, with only half-a-dozen contributors, it's easy to be the "main instigator" for anything at all.)
Anyway, that useful (to me) clause is in there, great thanks...but why is it also needing the $this->attributes_join to be true? Why not just the strpos($this->sort, 'model')?

proseLA:
because that is all you were able to get into #3246.  

personally, i'm not a fan of all of these changing formats based on sorts.  i like `name-> model number -> price -> id`.

PR #4174, opened 2021-01-15
Extract message_stack icon images to defines

zenexpert:
move image/icon to a language file for easier modifications


zenexpert:
move the annoying icon from message_stack.php class file into a language file for easier edits

drbyte:
@zenexpert thanks for this.

Given that we're trying to move away from having too much complexity (including function calls) in language files, I'd like to suggest we explore an additional variation on what you've submitted: use FontAwesome icons instead of images.

1. comment-out these defines that make calls to zen_image() .... leaving the commented-out version of these defines as examples that can be used by people with old templates, if they need it.

2. above the commented-out defines for zen_image, provide defines for the same constants, but with font-awesome icons HTML markup instead. (The default template currently provides FontAwesome 4.7.0).

drbyte:
Another thing I'm considering is whether to keep this as a separate language file or merge it into another.

zenexpert:
Thanks, Doc, I agree completely. This was all just a suggestion caused by the template I'm currently working on so I needed to get the annoying images out of the way and just have control from the language files. Up to you how to use it. :)

PR #4181, opened 2021-01-20
shopping cart notices on product update

### LINKED ISSUE #4180: problems in shopping_cart class with mixed
https://github.com/zencart/zencart/blob/e11afe2b0c8790f7d24246ad9fd9d036f9563e3b/includes/classes/shopping_cart.php#L2643-L2649

in this changed array, there is a boolean, 3 arrays, and 2 elements of an array.

i do not deal with attributes much, so this code is a bit foreign to me.  however using the test data, it is very easy to generate a notice here.  add the russ tippins band (pid: 166) to your cart; go to your cart, and change the quantity.  going up results in a php warning on the line 2647.  going down results in a warning on 1789.

mc12345678:
Address notice at line 1789, ensuring that $attributes is set to an array of some type, either empty or including attribute data. Formerly this looked for the "presence" of the attribute; however, both in current PHP environments and when strict PHP is enabled a notice was expected to be generated when entering this code cycle.

Corrects potential notice associated with product being evaluated if it is not within the `$_POST`ed content (for whatever unusual reason that may exist).
Corrects notice when evaluating decrease but the quantity increases or remains the same or in reverse if evaluating for increase and the quantity decreases or remains the same.

Fixes #4180

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -2595,7 +2595,7 @@ class shoppingCart extends base
         }
 
         $product_changed = [];
-        $product_total_change = [$pr_id => 0,];
+        $product_total_change = [$pr_id => 0];
         $product_tracked_changed = [];
         $product_last_changed = [];
         $product_increase = [];
@@ -2608,7 +2608,7 @@ class shoppingCart extends base
             if (!is_numeric($_POST['cart_quantity'][$i]) || $_POST['cart_quantity'][$i] < 0) {
                 $_POST['cart_quantity'][$i] = $current_qty; // Default response behavior in cart.
             }
-            // Ensure that array key exists before use in assignment.
+            // Ensure array key exists before use in assignment.
             if (!array_key_exists($prs_id, $product_last_changed)) {
                 $product_last_changed[$prs_id] = null;
             }
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -2581,6 +2581,10 @@ class shoppingCart extends base
         global $db;
 
         $pr_id = zen_get_prid($product_id);
+        
+        if ($pr_id === 0) {
+            return true;
+        }
 
         // check if mixed is on
         $product = $db->Execute("SELECT products_id, products_quantity_mixed FROM " . TABLE_PRODUCTS . " WHERE products_id=" . (int)$pr_id, 1);
@@ -2591,7 +2595,7 @@ class shoppingCart extends base
         }
 
         $product_changed = [];
-        $product_total_change = [];
+        $product_total_change = [$pr_id => 0,];
         $product_tracked_changed = [];
         $product_last_changed = [];
         $product_increase = [];
@@ -2603,6 +2607,13 @@ class shoppingCart extends base
             $current_qty = $this->get_quantity($products_id); // $products[$i]['quantity']
             if (!is_numeric($_POST['cart_quantity'][$i]) || $_POST['cart_quantity'][$i] < 0) {
                 $_POST['cart_quantity'][$i] = $current_qty; // Default response behavior in cart.
+            }
+            // Ensure that array key exists before use in assignment.
+            if (!array_key_exists($prs_id, $product_last_changed)) {
+                $product_last_changed[$prs_id] = null;
+            }
+            if (!array_key_exists($prs_id, $product_total_change)) {
+                $product_total_change[$prs_id] = 0;
             }
             if ($_POST['cart_quantity'][$i] != $current_qty) { // identify that quantity changed
                 $product_changed[$products_id] = $_POST['cart_quantity'][$i] - $current_qty;  // Identify that the specific product changed and by how much the customer increased it.
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -1786,7 +1786,7 @@ class shoppingCart extends base
                         // Verify minuses are good, and affect the items to be changed
                         //  This leaves only increases or 'netzero' to be at play.
                         foreach ($change_state[zen_get_prid($_POST['products_id'][$i])]['decrease'] as $prod_id) {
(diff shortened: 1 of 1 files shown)


includes/classes/shopping_cart.php
```diff
@@ -2592,7 +2596,9 @@ public function in_cart_product_mixed_changed($product_id, $chk = false)
 
         $product_changed = [];
         $product_total_change = [];
+        $product_total_change[$pr_id] = 0;
         $product_tracked_changed = [];
+        $product_tracked_changed[$pr_id] = null;
```
drbyte, line 2601: Was this set to null just as a precaution? Or was it actually triggering a notice of some sort without it?

proseLA: see #4181

```
--> PHP Warning: Undefined array key 166 in /var/www/zcdev/includes/classes/shopping_cart.php on line 2646.
PHP Warning: Undefined array key 166 in /var/www/zcdev/includes/classes/shopping_cart.php on line 2647.
```

mc12345678: No notice as a result of presence/absence had been seen and in part was added as a result of code review for possibilities. The value `null` was chosen for that assignment based on the intended meaning of this array.  In essence if there was no product that changed then tracking of that would indicate nothing had happened; however, I guess I see now that perhaps it should/could be set to false? I was thinking a little about tests that might already exist... While it is undefined (in former code) the use of `is_null` would not work quite right or throw a notice if attempted on a non-existent variable. `empty` would work with it being `false` or `null`, `isset` would only test against `null`.

I also recognized that regardless the value, it seems that this modification could be captured differently as perhaps shown below:
````diff
-$product_tracked_changed = [];
-$product_tracked_changed[$pr_id] = null;
+$product_tracked_changed = [
+    $pr_id => null,
+];
````
Part of my thought was the outside use of this method and providing a level of problem prevention.

File: includes/classes/shopping_cart.php [hunk 2]
```diff
@@ -2604,6 +2610,13 @@ public function in_cart_product_mixed_changed($product_id, $chk = false)
             if (!is_numeric($_POST['cart_quantity'][$i]) || $_POST['cart_quantity'][$i] < 0) {
                 $_POST['cart_quantity'][$i] = $current_qty; // Default response behavior in cart.
             }
+            // Ensure that array key exists before use in assignment.
+            if (!in_array($prs_id, $product_last_changed)) {
+                $product_last_changed[$prs_id] = null;
+            }
+            if (!in_array($prs_id, $product_total_change)) {
```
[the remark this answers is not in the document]
mc12345678: YES!!!! Oye... Coding from the keyboard... *shaking head*

Again though, `null` I guess is likely debatable.

includes/classes/shopping_cart.php
drbyte: @proseLA I was referring to `$product_tracked_changed` which is only used internally to this function, not in the one you were reporting in #4180.

drbyte: @mc12345678 regarding:

```diff
-$product_tracked_changed = [];
-$product_tracked_changed[$pr_id] = null;
+$product_tracked_changed = [
+    $pr_id => null,
+];
```
I prefer one-line inits where possible, as IMO it's more readable when we're only initializing something that's tiny, eg:
`$product_tracked_changed = [$pr_id => null];`

But I'm still wondering whether setting it to null (actually setting it to anything at all) is even needed.
Granted, it's not a huge deal, and largely harmless.
I'm just aware that the cognitive effort of studying through all this section of code to follow the bouncing ball is already complex enough, and it's a delicate balance to decide whether adding more is helpful or harmful. LOL
Nevertheless, thanks for tackling this.

mc12345678: So I didn't want to add "more" to my previous statement until I received a little more input (as provided).

I did consider a use case where having something initiated outside of the `for` loop may be of value. Specifically, if a "test" were being performed against the information to see if/what change might occur by accessing this method against a product that was not already present in the list of information obtained using the `for` loop.

Personally, I would prefer if the anticipated result from this method was that if nothing had changed that effectively nothing would be returned which seems would then make the analysis of the returned result "easier", but it also seems/appears that to do that would require even more code in this area.  I further think that while such a result could be obtained that the need is not present (yet).

Now...  Hmmm... Looking again at what all is going on (and I probably need to make an additional look specifically at `$product_track_changed`), it is appearing ever more that I was overzealous in addressing the potential notice generation and that the additional assignment may not need to occur...

In either case, a forced update needs to occur to either simplify the assignment or remove this additional assignment. My next push to this branch will be my considered answer to that. Thanks for the feedback/review!

mc12345678: Yup, thank you for knocking some sense back into me... As you indicated, the `$product_track_changed` array basically is an internal collection of those product that have been changed where the changed product identifier is the key to the array. 

Thus if it has not changed, then there is no value assigned to it and it does not potentially cause additional action later... The other thing is that by initializing the array with a value in it (an integer at that), it is not expected that the value provided in the key would even be found later in processing (or if it were found it may cause an unusual result) because at least by overall (original?) design, a product without attributes generally can't be mixed with other product (with/without attributes). (second thought, a product with checkboxes exactly fits that condition, none checked as compared to a single checkbox marked. Then there are some other conditions such as a radio button group with no default marked, or perhaps a text box that has no text in it compared to one that does).

That said, other code (added to the cart system) *could* support such a condition/situation and/or as stated above the existing cart arranged also supports the concept of a product mixing with a variant of the product, but... There does not seem to be an issue of allowing the mix of product with and without attributes Basically is a situation that doesn't have to be addressed, corrected, or otherwise handled provided that it doesn't directly cause an issue. Now... before one asks, what about a product without attributes and "mixing" with a product without attributes... That's 1) the same product_id and 2) a multiple quantity of that product where the total quantity is the limit whether the switch is flipped or not.

Ok, the above is a lot of explanation about why no further "mod"/prevention has been incorporated to the associated code.

PR #4189, opened 2021-01-23
[157] Additional category_product_listing.php notifiers

lat9:
Additional notifications added for the admin's Category/Product listing, enabling an observer to

- Handle additional form-actions
- Insert columns into the _products_ listing, both before and after the Quantity column

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -773,11 +773,17 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
             
             $products_query_raw = "SELECT p.products_type, p.products_id, pd.products_name, p.products_quantity,
                                           p.products_price, p.products_status, p.products_model, p.products_sort_order,
-                                          p.master_categories_id" . $extra_select;
-            $products_query_raw .= " FROM " . TABLE_PRODUCTS . " p" . $extra_from;
-            $products_query_raw .= " LEFT JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (pd.products_id = p.products_id)" . $extra_joins;
+                                          p.master_categories_id";
+            $products_query_raw .= $extra_select;
 
-            $where = " WHERE pd.language_id = " . (int)$_SESSION['languages_id'] . $extra_ands;
+            $products_query_raw .= " FROM " . TABLE_PRODUCTS . " p";
+            $products_query_raw .= $extra_from;
+
+            $products_query_raw .= " LEFT JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (pd.products_id = p.products_id)";
+            $products_query_raw .= $extra_joins;
+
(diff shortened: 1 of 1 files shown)


lat9:
Think "Dual Pricing" when reviewing.

admin/category_product_listing.php
```diff
@@ -693,13 +758,25 @@
...
+            $extra_select = $extra_from = $extra_joins = $extra_ands = '';
+            $zco_notifier->notify('NOTIFY_ADMIN_PROD_LISTING_PRODUCTS_QUERY', '', $extra_select, $extra_from, $extra_joins, $extra_ands, $order_by);
+            
             $products_query_raw = "SELECT p.products_type, p.products_id, pd.products_name, p.products_quantity,
                                           p.products_price, p.products_status, p.products_model, p.products_sort_order,
-                                          p.master_categories_id
-                                   FROM " . TABLE_PRODUCTS . " p
+                                          p.master_categories_id $extra_select
+                                   FROM " . TABLE_PRODUCTS . " p $extra_from
                                    LEFT JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (pd.products_id = p.products_id)";
```
drbyte, line 777: Formatting request:

Please do not interpolate `$vars` inside double-quoted strings.
We've avoided doing that in Zen Cart core for readability.
Plus, while PHP allows it, IDEs don't consistently recognize them, which leads to potential human error.

admin/category_product_listing.php
```diff
@@ -693,13 +758,25 @@
...
-
-            $where = " WHERE pd.language_id = " . (int)$_SESSION['languages_id'];
+            $products_query_raw .= $extra_joins;
+            $where = " WHERE pd.language_id = " . (int)$_SESSION['languages_id'] .$extra_ands;
```
drbyte, line 779: Formatting request:

Please always leave one space on either side of a concatenating `.` operator.

admin/category_product_listing.php
lat9: Do you want me to re-submit with these (and the following) changes?

admin/category_product_listing.php
```diff
@@ -768,7 +845,67 @@
...
+              //
+              // $extra_data = array(
+              //     array(
+              //       'align' => $alignment,    // One of 'center', 'right', or 'left' (optional)
+              //       'text' => $value
+              //     ),
+              // );
+              //
+              // Observer notes:  
+              // - Be sure to check that the $p2/$extra_headings value is specifically (bool)false before initializing, since
```
drbyte, line 862: Mention of `$extra_headings` here should be `$extra_data`.

Same with below.

admin/category_product_listing.php
drbyte: Yes please.

lat9:
The **two** data-related notifications were using $heading_info instead of $data_info; just tried to push an update for that.

Issue #3914, opened 2020-08-23
Discuss: New name for Define Pages and Define Page Status


drbyte:
The term `define pages` has zero association with what it actually is, or is for.
Specifically, "define pages" are neither "pages" nor "defines".

Further, we have a Configuration menu group named `Define Page Status` which controls "links" and whether the "define page text content" appears on the associated "page".
Scott's done a great job of describing this configuration group at: https://docs.zen-cart.com/user/template/define_pages/

Also, there is some suggestion that when the storeowner disables the "link" to any of these pages they may also mean to disable the page itself entirely. (eg: disabling the `shipping_info` link from the `information` sidebox, might also mean they have no intention of ever using the `shipping_info` page, and therefore wish it to throw a 404-not-found if someone accidentally stumbles upon the URL)

We need 2 new names:
1. What to call the "content" of "define pages". Is it "snippets"? "content"? "text"? "message"? other?
2. What to call the configuration group that controls these "things" ... keeping in mind the idea of maybe expansing the "group" with settings which also control whether a certain page "shows" if visited, even if the "link" is "disabled".

scottcwilson:
It's like grape nuts.  Neither grape nor nut. :)

wiztechinc:
Well, add in the problem that it really is difficult to figure out which setting to use. I remember my first time and have to gear my brain right each time I touch those settings. 

Owner added, admin added content, admin editable content ? 

If nothing else, changing the config settings name to sidebox display links? Or something similar because some of them only work with sideboxes though the issue arises of the various places and actions one has to take to remove the newslette link and gc faqs. Whoops, now I'm going in circles. Getting down into the weeds, I know, but trying to document the settings is a tad complicated.  (I just looked at this while doing the sidebox-list docs page.) 

I am looking at this from that side. Most owners see the box, see the link and then want to find out how to manage those links (if they even realize they can). How to have some nomenclature that will guide them to these answers?

scottcwilson:
> Scott's done a great job of describing this configuration group at: https://docs.zen-cart.com/user/template/define_pages/

After five rewrites I should hope so!  But it is a tricky concept to explain.   New names might help.

wiztechinc:
I'm not complaining that it was you and not me!

scottcwilson:
How about "Define Content Block" for both the editor under Tools and the settings under Configuration?  Then there's a clear difference between what's in html_includes (content blocks) and the full page infrastructure required to build a web page for storefront display.

[...]

PR #4194, opened 2021-01-28
Add a define page for search results

scottcwilson:


drbyte:
I'm not a fan of adding more "define pages".

I'd be content to see the addition of a language constant which accomplishes the same messaging.

scottcwilson:
The advantage of a define page is that it is editable by the end user and doesn't require a developer to help.

scottcwilson:
See update to #3914 as a way to move forward on define pages.

drbyte:
Update: Not gonna do this in 1.5.8.

I'd accept a language define (perhaps the generic `TEXT_INFORMATION` one?) with some universal default text. 

Perhaps that default should be something like "We're happy to help you find what you're looking for. _Contact Us_ for additional help." 
(The previously-proposed "Not finding what you're after..." text will show even BEFORE they try to do any search if they land on the search page via a link, so it's a slightly negative message before they even start, hence the suggestion for "We're happy to help...")

PR #4195, opened 2021-01-29
V158 remove laravel keep eloquent

zcwilt:


zcwilt:
This commit removes the full Laravel Framework that we previously had in v158, while preserving the use of Laravel Eloquent and Laravel Models.

**Why are we doing this.**
The simple answer is **Security**

Zen Cart operates with all of it's code exposed in the web root, while frameworks like Laravel only expose an entry point (e.g. index.php) and the html assets (js/css/etc)
All other code is stored in directories above the web root and is therefore inaccessible regardless.

For core Zen Cart files we can mitigate the exposure of files ourselves, and in most cases this may be a case of adding 

      die('Illegal Access');
    }

to the header of the file, and generally we have ultimate control of the contents of those files.

For files that are managed via Composer, we don't have that luxury. 
Because we cannot put files above the web root, all the files in the Vendor folder are potentially 
exploitable by hackers.

For the vendor folder that now still exists, to support the use of Laravel Eloquent we have surveyed the contents and removed files from packages and added extra security to ensure files are not directly addressable.

Longer term it would be nice to produce a version of Zen Cart that, like other frameworks, stored the bulk of it's code above web root, and we are exploring that option.

lat9:
Thanks for that, @zcwilt.  I was going to ask why the change was needed!

PR #4196, opened 2021-01-30
[158] account: Format order-totals using currency in which order was placed

lat9:
The 'Total' column displaying the customer's previous orders is currently being displayed as a 'raw' number instead of using a currency format..


drbyte:
Thanks.
Apparently the account_history page is doing the same.

PR #4202, opened 2021-02-08
fix move and copy after searching

proseLA:
there is a refactor of search terms here.  it just seems clearer to me.

i would think that we can add the page parm into the same var, but....

**more importantly** on new lines, 920, 921; the inclusion of the `$search_parameter` is necessary.  else, depending on the size of your products table and various other configs, you can end up with `$pInfo` not getting properly populated due to the split page results populating the limit on the raw sql, with results in a search set that does not have the product identified by the `$_GET['pID']` as part of said search results.  (screenshot below.)

i have encountered this bug numerous times over the years.

personally, i think we should carry over search terms after completing a copy, move, delete, etc.  but i did not want to get too aggressive.  but this is a bug that should be fixed.

image


scottcwilson:
Related: #4288

PR #4217, opened 2021-02-18
IGNORE Seeing cases where shipping_attributes_price is not numeric

scottcwilson:


lat9 (commented):
Various castings to floats are the primary cause of penny-off calculations.

When the value is not numeric, what value does it take on?  What are the cases that expose this unwanted value?

scottcwilson:
If you don't like the cast, we could just set it to 0 if it's not numeric.  Seems to happen at checkout time: 

#1  shoppingCart->calculate() called at [/home2/storename/public_html/store/includes/classes/shopping_cart.php:1660]
#2  shoppingCart->free_shipping_items() called at [/home2/storename/public_html/store/includes/functions/functions_general.php:918]
#3  zen_get_shipping_enabled() called at [/home2/storename/public_html/store/includes/modules/shipping/item.php:25]
#4  item->__construct() called at [/home2/storename/public_html/store/includes/classes/shipping.php:68]
#5  shipping->__construct() called at [/home2/storename/public_html/store/includes/modules/checkout_process.php:52]
#6  require(/home2/storename/public_html/store/includes/modules/checkout_process.php) called at [/home2/storename/public_html/store/includes/modules/pages/checkout_process/header_php.php:14]
#7  require(/home2/storename/public_html/store/includes/modules/pages/checkout_process/header_php.php) called at [/home2/storename/public_html/store/index.php:35]
--> PHP Warning: A non-numeric value encountered in /home2/storename/public_html/store/includes/classes/shopping_cart.php on line 737.

line 737 is the changed line (in version 1.5.7b)

$freeShippingTotal += (float)$shipping_attributes_price;

drbyte:
If you `var_dump()` the value, what is it?  Is it null? empty string? text?

scottcwilson:
These are just log files I'm going off.  I can't reproduce it locally.  But I'll add a call to `error_log` that triggers if it's not numeric and see what happens.

drbyte:
What version of Zen Cart is in use on the site?
If it's v157b, does it have the patches from github that address this same issue from https://github.com/zencart/zencart/issues/4153 ?

scottcwilson:
It is 1.5.7b but no patches. OK will try those.

scottcwilson:
Those look really promising.  I will close this if the issue doesn't reappear.

drbyte:
There was a follow-up commit by mc12345678 after the one for that issue:
https://github.com/zencart/zencart/commits/v157/includes/functions

scottcwilson:
Added the storefront side fixes from #4153: 
https://github.com/zencart/zencart/commit/e7f7d092acaad7718a2ed455bdce772968d71bae
https://github.com/zencart/zencart/commit/1798d83348869ee018931b844d4ee297a84c2258

and this problem no longer occurs (even without float cast). Closing.

PR #4222, opened 2021-02-20
Show image for products in category_product_listing page

scottcwilson:


torvista:
I did this on my site, but made it switchable on the page. I would have thought not everyone may want this on permanently?

PR #4223, opened 2021-02-22
Parameter for zen_redirect of zen_href_link incorrect.

mc12345678:
The admin version of `zen_redirect` has only one parameter which is the
destination address. The generation of that address through `zen_href_link`
accepts an indicator of whether to use `SSL` or `NONSSL`. That value is
expected to be in the third position of calling the `zen_href_link`
function. The modified line originally has `NONSSL` as the second parameter
to the `zen_redirect` function. When modified as provided, this would
correctly become the third parameter.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -155,7 +155,7 @@ if (zen_not_null($action) && $order_exists == true) {
           break;
         default:
           $messageStack->add_session(sprintf(TEXT_EXTENSION_NOT_UNDERSTOOD, $file_extension), 'error');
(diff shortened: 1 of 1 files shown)


scottcwilson:
I could have *sworn* I fixed that.  Thanks for catching it.

drbyte:
> I could have _sworn_ I fixed that. Thanks for catching it.

@scottcwilson you did in 158 https://github.com/zencart/zencart/commit/4cd14b23c0bb6df2969de2f313fd48bed1842011

scottcwilson:
Ah, 158.  Good to know the old guy's not losing it (at least not at a high rate of speed!).

PR #4225, opened 2021-02-23
address book details:remove placeholder from country dropdown

torvista:


includes/templates/template_default/templates/tpl_modules_address_book_details.php
```diff
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
```
drbyte, line 2: Why?

torvista: My mistake, I put that in my files as I go along.

PR #4230, opened 2021-02-28
Pre-check for attacks

scottcwilson:
Bad guys submitting emails like ð\'\'ð\"\" which cause needless logs.

### DIFF
--- a/includes/modules/pages/login/header_php.php
+++ b/includes/modules/pages/login/header_php.php
@@ -30,8 +30,8 @@ include(DIR_WS_MODULES . zen_get_module_directory(FILENAME_CREATE_ACCOUNT));
 //
 $email_address = zen_db_prepare_input(isset($_POST['email_address']) ? trim($_POST['email_address']) : '');
 
-$error = false;
-if (isset($_GET['action']) && $_GET['action'] == 'process') {
+$error = filter_var($email_address, FILTER_VALIDATE_EMAIL); 
+if (!$error && isset($_GET['action']) && $_GET['action'] == 'process') {
     $loginAuthorized = false;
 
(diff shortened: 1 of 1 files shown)


mc12345678:
Doesn't the inserted "bypass" significantly change the operation of this section of code?

Previously, it was possible that the email address could be assigned to a set of single quotes and if the form was being submitted for processing the inner area would have been executed. Such single quotes would fail the: `filter_var($email_address, FILTER_VALIDATE_EMAIL)` test and therefore the inner area would no longer be processed.

Now I don't pretend to fully understand why it was permissible to move forwards with an empty email address, but at the same time the reported issue is against something other than `empty` (Perhaps that is the solution?)
````php
if ((!$error || empty($email_address)) && …
````

drbyte:
Upon first review I don't think this is in the right location in relation to overall flow.

Also wondered why the ZC function for email validation wasn't used here, for sake of consistency.

scottcwilson:
I can change it to call zen_validate_email if you'll take the PR.  I just wanted a quick check to prevent the log.

drbyte:
I think this needs more study. I'm not comfortable with it as it is now in terms of overall flow for where this validation test is being done.

scottcwilson:
To be clear: I have no preference where this check is done; I just don't want to get a log file.

drbyte:
I meant to ask earlier: what exactly is the message in the log?

scottcwilson:
Can't repro in 1.5.8.  Could be the template in question was lacking the type hint that modern templates use.

scottcwilson:
Updating the template changes document to remind people to check for this improvement.

PR #4231, opened 2021-03-01
EMAIL_ORDER_DATE is no longer used in email templates

scottcwilson:


drbyte:
What might this removal break?

scottcwilson:
This setting of this entry in the `html_msg` array was introduced in zen-cart-v1.5.2-RC1-11262013 but it has never been referenced in any version of Zen Cart.   I don't see it in any plugins either.

drbyte:
Does that mean the define for `ORDER_EMAIL_DATE_FORMAT` should be removed as well?

scottcwilson:
I reckon so.  Good catch.

Issue #4089, opened 2020-11-15
[156-158] Mis-configured downloadable product results in entire order getting 'Free Shipping'


lat9:
As far back as zc156 (probably earlier, but untested), if a product with a 'download' attribute (e.g. demo product at `index.php?main_page=product_info&cPath=54_60&products_id=179`) has been set in its product configuration to have _either_ **Product is Virtual: Yes, skip shipping address** _or_ **Always Free Shipping: Yes, Always Free Shipping** and that product is added to the cart with a product that requires shipping, the order is marked has having 'Free Shipping' (via the freeshipper).

This can be reproduced using the demo products,  modifying the configuration for products_id 179 to have one or the other or both of the above options set.  Place a not-free-shipping (e.g. one of the DVD Movies) into the cart, go to the cart page and see that a shipping selection is required.  Then place one of the (modified) Single Download products in the cart, go to the cart page and see that the order is now "Free Shipping".

The source of the error is in the shopping_cart.php class, since it's updating its `free_shipping_item` count once to account for the product's always_free_shipping/virtual setting and once (again) for the fact that the product has a download.  That results in a `free_shipping_item` value of 2 when there's only one of the Single Download product in the cart, causing the class to believe that the cart's products are all free-shipping (since there are 2 distinct products in the cart).

I realize that the settings described above is 'non-standard' for downloads, but I spent at least an hour trying to determine why one product on a client's site, when added to the cart, resulted in a totally free-shipping cart.

scottcwilson:
I could have *sworn* I updated Admin > Catalog > Downloads Manager screen to check for misconfigurations like this - but apparently not.

scottcwilson:
Oh it's in 1.5.8.

drbyte:
Does that mean this issue is resolved in v158?

scottcwilson:
No but at least you get a warning when you go to Admin > Catalog > Downloads Manager that you have messed up!

scottcwilson:
I think I'd like to fortify this with a similar message in the shipping modules.

scottcwilson:
> No but at least you get a warning when you go to Admin > Catalog > Downloads Manager that you have messed up!

Correction: you used to.  This functionality has been broken.

scottcwilson:
> I think I'd like to fortify this with a similar message in the shipping modules.

See PR #4236

scottcwilson:
Cindy, is the situation now where you get warnings in multiple places satisfactory?  Or do you think more should be done?

[...]

PR #4235, opened 2021-03-07
Display warning icon even if product not selected

scottcwilson:
Found during investigation of #4089.


scottcwilson:
yellow_light

PR #4236, opened 2021-03-07
Add function to modules to check for configuration errors

scottcwilson:
Sample Use: freeshipping/download check.

Note that this could be used for all sorts of things for any of the modules (shipping, payment, order totals).


scottcwilson:
freeshipping

admin/includes/functions/general.php
```diff
@@ -758,3 +758,31 @@ function zen_get_orders_status_pulldown_array()
 
     return $orders_status_array;
 }
+
+/***
+ * Do the misconfiguration check which Admin > Catalog > Downloads Manager
```
drbyte, line 763: - Is this code basically a copy/paste from downloads_manager.php? 
- Should downloads_manager.php simply call this function instead of using separate code?
- Should this function really be in "general.php"? I think probably belongs in the central set of functions files catalog-side that are already loaded admin-side, maybe functions_attributes.php ?

scottcwilson: a) Yes - I tried to keep it as close as possible
b) No because downloads manager already does this query and needs the query results for other reasons
c) I am open to placement - whatever you suggest.  I just chose that place because it's not truly needed storefront side.

admin/includes/languages/lang.english.php
```diff
@@ -534,6 +534,7 @@
     'TEXT_BOOLEAN_VALIDATE' => 'The value is required to be a boolean value or equivalent.',
     'TEXT_ASC' => 'asc',
     'TEXT_DESC' => 'desc',
+    'TEXT_DOWNLOADABLE_PRODUCTS_MISCONFIGURED' => 'Some downloadable products are misconfigured.  See Admin &gt; Catalog &gt; Downloads Manager.',
```
mc12345678, line 537: Is this really the "right" message? What about if the result was intentional or had no impact on the site's normal operation? Wouldn't it be more assistive to indicate the result of this configuration? Such as:
> Configuration of some downloadable product causes free shipping for all cart items. See Admin &gt; Catalog &gt; Downloads Manager.

I say this because for example there are carts that do not ship anything and configuring the product causing this message will have no direct impact on the income to the site (until product is incorporated requiring shipping fee collection not associated with the free shipping module).

drbyte:
Thinking about meaningfulness of function names:
- is `zen_check_downloads_shipping()` descriptive enough? I'm not sure I understand its intent by its name.
- is `verify()` too generic? Is its purpose to "verify whether the module is installed?" or "verify that the module has all necessary language defines?" or "turn itself on or off?" or "verify that there's no new version available for it?", etc.

scottcwilson:
I am open to changing either or both of these names, just let me know.

admin/includes/functions/general.php
```diff
@@ -758,3 +758,31 @@ function zen_get_orders_status_pulldown_array()
...
+   $products_downloads_query = $db->Execute($sql); 
+   $valid = true; 
+   foreach ($products_downloads_query as $products_downloads) {
+
+      if ($products_downloads['product_is_always_free_shipping'] == 1 || $products_downloads['products_virtual'] == 1) {
+         $valid = false; 
+         break;
+      }
+   }
+   return $valid;
```
drbyte, line 787: This `$valid` variable is not needed. The `break;` could just be `return false;`, and then the final return would be `true`.

includes/modules/shipping/freeoptions.php
```diff
@@ -205,6 +205,12 @@ function check() {
       return $this->_check;
     }
 
+    function verify() {
```
drbyte, line 208: Change `verify()` to `get_configuration_errors()`.

admin/includes/functions/general.php
```diff
@@ -758,3 +758,31 @@ function zen_get_orders_status_pulldown_array()
...
+                      FROM " . TABLE_PRODUCTS_ATTRIBUTES_DOWNLOAD . " pad
+                      LEFT JOIN " . TABLE_PRODUCTS_ATTRIBUTES . " pa ON pad.products_attributes_id = pa.products_attributes_id
+                      LEFT JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON pa.products_id = pd.products_id
+                        AND pd.language_id = " . (int)$_SESSION['languages_id'] . "
+                      LEFT JOIN " . TABLE_PRODUCTS . " p ON p.products_id = pd.products_id
+                      WHERE pa.products_attributes_id = pad.products_attributes_id";
+   
+   $products_downloads_query = $db->Execute($sql); 
+   $valid = true; 
+   foreach ($products_downloads_query as $products_downloads) {
```
drbyte, line 780: Rename `$products_downloads_query` and `$products_downloads` to use "singular" for "product".
eg: `$product_downloads`

Also, `$product_downloads_query` is actually the result of a query, and not the query itself, so `$results` would suffice, and thus be consistent with other refactored functions.

Actually, then it would be even better to just `foreach ($results as $result)`, again for consistency with core.

admin/includes/functions/general.php
```diff
@@ -758,3 +758,31 @@ function zen_get_orders_status_pulldown_array()
 
     return $orders_status_array;
 }
+
+/***
+ * Do the misconfiguration check which Admin > Catalog > Downloads Manager 
+ * does to verify that downloads don't have invalid shipping settings.
+ */
+function zen_check_downloads_shipping() {
```
drbyte, line 766: Rename `zen_check_downloads_shipping()` to `zen_check_for_misconfigured_downloads()`

admin/modules.php
```diff
@@ -392,6 +396,10 @@
                    $contents[] = $help_button; 
                 }
                 $contents[] = ['text' => '<br>' . $mInfo->description];
+
+                if (!empty($mInfo->verify_msg)) { 
+                  $contents[] = ['text' => $mInfo->verify_msg . '<br>'];  // warnings, etc.
```
drbyte, line 401: Change `verify_msg` to `configuration_errors`

admin/includes/languages/lang.english.php
drbyte: Yes, probably worth making this suggested change.

admin/includes/functions/general.php
drbyte: And move to `functions_attributes.php`

drbyte (changes requested):

scottcwilson:
I think all the changes have been done.

admin/modules.php
```diff
@@ -215,6 +215,10 @@
                           }
                         }
                         $module_info['keys'] = $keys_extra;
+                        if (method_exists($module, 'verify')) { 
+                          $module_info['configuration_errors'] = $module->verify();
```
drbyte, line 219: Both instances of `verify` need to be changed to `get_configuration_errors`

scottcwilson: got 'em.

drbyte (approved):

scottcwilson:
Nope missed a couple - see #4309

PR #4239, opened 2021-03-09
Add $product to notifier in category_product_listing.php

brittainmark:
Add $product to data notifiers so global not required in observers.


drbyte:
Thanks!

PR #4241, opened 2021-03-12
Mitigate notices for a non-existent gID

mc12345678:
Should amplify that this was considered from a link address where the gID parameter was set, but it was set to a value that is not present in the database.

### DIFF
--- a/admin/configuration.php
+++ b/admin/configuration.php
@@ -54,6 +54,10 @@ $cfg_group = $db->Execute("SELECT configuration_group_title
                            FROM " . TABLE_CONFIGURATION_GROUP . "
                            WHERE configuration_group_id = " . (int)$gID);
 
+if ($cfg_group->RecordCount() == 0) {
+  $cfg_group->fields['configuration_group_title'] = '';
+}
+
 if ($gID == 7) {
   $shipping_errors = '';
   if (zen_get_configuration_key_value('SHIPPING_ORIGIN_ZIP') == 'NONE' or zen_get_configuration_key_value('SHIPPING_ORIGIN_ZIP') == '') {
@@ -178,10 +182,10 @@ if ($gID == 7) {
           $contents = array();
 
           // Translation for contents
-          if (defined('CFGTITLE_' . $cInfo->configuration_key)) {
+          if (isset($cInfo) && isset($cInfo->configuration_key) && defined('CFGTITLE_' . $cInfo->configuration_key)) {
             $cInfo->configuration_title = constant('CFGTITLE_' . $cInfo->configuration_key);
           }
-          if (defined('CFGDESC_' . $cInfo->configuration_key)) {
+          if (isset($cInfo) && isset($cInfo->configuration_key) && defined('CFGDESC_' . $cInfo->configuration_key)) {
             $cInfo->configuration_description = constant('CFGDESC_' . $cInfo->configuration_key);
           }
 


scottcwilson:
Nice!  I'd like to add this check to Mod List too. What was the database entry that triggered the notice you are fixing?

mc12345678:
While I might with a stretch come up with something inherent to base code, the notices were generated when two browser tabs were open and performing the following. 

I executed a "reinstall" of a plugin, which does "cleanup" to remove the plugin's configuration menu option and recreates it while in the other tab I was looking at the list of configuration items. When the second tab was updated, of course nothing was associated with that gID because the "cleanup" did not retain the previous gID. As a result of accessing a configuration menu item with a gID that had no information I received several notices (I should post those as well). It appeared that the minimal modifications necessary to retain the "empty" list (without further user notification) and not to cause the notices was accomplished through the posted changes.

As a result of this discovery, I do also plan to modify the plugin to attempt to retain the gID in the "cleanup" process so that any other potentially generated (hard-coded?) links to that particular configuration menu will be retained. I don't know of any reason that the particular gID during a reinstall would need to be made anew. A removal followed by an install is a different situation and therefore the old gID would likely be lost. I am not specifically for reusing the most recently available record reference just because it is "the next" in line.

scottcwilson:
OK that's a bit of an edge case ... I thought it was something that you had actually encountered in a realistic test.

drbyte:
> I am not specifically for reusing the most recently available record reference just because it is "the next" in line.

agreed: let the auto-incrementing do its job.

> I plan to modify the plugin to attempt to retain the gID in the "cleanup" process so that any other potentially generated (hard-coded?) links to that particular configuration menu will be retained.

Since Zen Cart 1.5.6 we've made an effort to ensure that all core-code modules "update in-place" any of their configuration settings that require updates, and detect-and-auto-add any new settings that need to get added.  This avoids the "old" requirement of remove/reinstall of modules during-an-upgrade that was so inconvenient on older versions.
Note: all such configuration checks ONLY fire admin-side. Firing all the dependency lookups catalog-side adds a lot of overhead and would slow down busy storefronts unnecessarily.

mc12345678:
As "promised", the error log, modified to remove unnecessary details:

````php
[Local Time] Request URI: /admin/index.php?cmd=configuration&gID=47, IP address: IP_ADDRESS
#1  require(/admin_path/configuration.php) called at [/admin/index.php:11]
--> PHP Notice: Undefined index: configuration_group_title in /admin_path/configuration.php on line 89.

[Local Time] Request URI: /admin/index.php?cmd=configuration&gID=47, IP address: IP_ADDRESS
#1  require(/admin_path/configuration.php) called at [/admin/index.php:11]
--> PHP Notice: Undefined variable: cInfo in /admin_path/configuration.php on line 143.

[Local Time] Request URI: /admin/index.php?cmd=configuration&gID=47, IP address: IP_ADDRESS
#1  require(/admin_path/configuration.php) called at [/admin/index.php:11]
--> PHP Notice: Trying to get property 'configuration_key' of non-object in /admin_path/configuration.php on line 143.

[Local Time] Request URI: /admin/index.php?cmd=configuration&gID=47, IP address: IP_ADDRESS
#1  require(/admin_path/configuration.php) called at [/admin/index.php:11]
--> PHP Notice: Undefined variable: cInfo in /admin_path/configuration.php on line 146.

[Local Time] Request URI: /admin/index.php?cmd=configuration&gID=47, IP address: IP_ADDRESS
#1  require(/admin_path/configuration.php) called at [/admin/index.php:11]
--> PHP Notice: Trying to get property 'configuration_key' of non-object in /admin_path/configuration.php on line 146.

````

scottcwilson:
what *was* at gID 47?  How do I dup this?

mc12345678:
> Since Zen Cart 1.5.6 we've made an effort to ensure that all core-code modules "update in-place" any of their configuration settings that require updates, and detect-and-auto-add any new settings that need to get added. This avoids the "old" requirement of remove/reinstall of modules during-an-upgrade that was so inconvenient on older versions.
Note: all such configuration checks ONLY fire admin-side. Firing all the dependency lookups catalog-side adds a lot of overhead and would slow down busy storefronts unnecessarily.

For those reading this PR/issue with interest, for the plugin I'm working on, the above is the expectation that tests, checks, updates be performed predominantly in the admin. That at least is the area that will be "corrected". The code had a "delete" section and a "new" section. Way back when, in attempts to support a "refresh" option, the two were called back-to-back without any communication between the two.  That last part will be changed to support conditions that are to be met. (E.g., if the configuration menu itself needs modification (not the items in the list) then the appropriate action will be considered to support that update part.)

mc12345678:
> what was at gID 47? How do I dup this?

Pick any gID that doesn't exist in a default install, e.g. 999. Conditions of the system that was being used was PHP 7.4, with strict PHP reporting (E_ALL).

What was there was a full configuration menu that before running the "reinstall" had items listed, when the delete portion of the reinstall was performed, all reference to gID=47 was removed from the database and the next available gID (as defined by the database) was created/taken for the configuration menu contents.  In my case, there was no "change" in the menu descriptions, just a reset of the value(s) (which in my case actually had no change as at the point of discovery and further testing this was an exercise of the process and results.

PR #4242, opened 2021-03-12
Rename admin "Email" Menu option name (instead of "Email Options")

scottcwilson:
In 1.5.7 we made the title of the email configuration page "Email."  Let's make the menu under Configuration match that. 
(Trivial but suggested for manual backport to 1.5.7d for completeness.)


drbyte:
I'll note that this makes hundreds of forum support posts incorrect because they all say "Email Options".

That said, I agree that just "Email" is more intuitive.

scottcwilson:
You're right but I think it's not too much of a hurdle to go from Configuration > "Email Options" to Configuration > "Email."  (Honestly, what I thought was really egregious was the 1990s spelling of "E-Mail.")

PR #4243, opened 2021-03-12
Define a notifier so new pages can offer their own help

scottcwilson:


admin/includes/functions/functions_help.php
```diff
@@ -163,5 +164,7 @@ function page_has_help()
     if (isset($pagelist[$page])) {
         return $pagelist[$page];
     }
-    return false;
+    $help_page = false; 
+    $zco_notifier->notify('NOTIFIER_PLUGIN_HELP', $page, $help_page);
```
drbyte, line 168: Can we rename this to `NOTIFIER_PLUGIN_HELP_PAGE_URL_LOOKUP`, to add a little specificity.

PR #4261, opened 2021-03-25
Move charset to start of head

torvista:
I needed to do this long ago for some obscure issue with a payment module, but it is also recommended (and highlighted by Edge Developer Tools Console Issues).
https://webhint.io/docs/user-guide/hints/hint-meta-charset-utf-8/?source=devtools


drbyte:
While I don't disagree with this change, I'm confused about why the charset needed before a dns-prefetch?

torvista:
All I could find was that charset should be first and that dns-prefetch should be in the head, but not that there was any priority or influence between them. I think they are unrelated.

drbyte:
Ya, I'm pretty sure they're unrelated as well, but I'm okay with this change.

PR #4264, opened 2021-03-29
Add new key if still missing

scottcwilson:


drbyte:
Please move that code to the `check()` function instead.

scottcwilson:
Separate PR to change `authorizenet_aim.php` and  `authorizenet.php` to put these operations in `check()` as well?

drbyte:
> Separate PR to change `authorizenet_aim.php` and `authorizenet.php` to put these operations in `check()` as well?

No. I'd rather touch none of these modules for now.
But the one you're PRing here for DP is (I assume a fix to an actual problem on a live site).

The "best" fix is to make the `install()` method check each one and self-repair. **But** I don't want us touching any of these modules for such massive reformatting without actual feature changes.

PR #4265, opened 2021-03-29
Extend country strings in orders table

### LINKED ISSUE #4257: Country name United States Minor Outlying Islands too long
The customers_country, delivery_country and billing_country fields are too short to contain the name "United States Minor Outlying Islands."  

I'd be happy to fix this but want to confirm the core team's preference before moving forward:

- extend these fields in the orders table
- abbreviate this string to "US Minor Outlying Islands"

scottcwilson:
Fixes #4257


lat9:
Is there a reason not to use varchar(63)?  It's got the same overhead as varchar(44) but provides additional 'extension' for the future.

drbyte:
While it's never been needed thus far, for the sake of consistency let's use varchar(64) here.
The only overhead would be indexes on those fields, but at this point we don't have need of those.

PR #4267, opened 2021-03-31
Move checks out of the for loop, check all variables before access.

scottcwilson:


drbyte:
Why?

How did this come up as "needed"?

admin/attributes_controller.php
```diff
@@ -194,15 +194,19 @@
       break;
     case 'add_product_attributes':
       $current_image_name = '';
+      if (!isset($_POST['options_id']) || 
+           !isset($_POST['products_id']) || 
+           !empty($_POST['values_id'])) {
+           break;
+      }
```
drbyte, line 201: - `isset` can take multiple vars as comma-separated parameters.
- Shouldn't your `!empty()` be `empty()` ?
```suggestion
      if (!isset($_POST['options_id'], $_POST['products_id']) || empty($_POST['values_id'])) {
           break;
      }
```

scottcwilson:
Good catch and thanks for the suggestion.  Done.

PR #4269, opened 2021-04-01
Move styles to css classes

Zen4All:
This PR is ment to move a lot of the "style" attributes in to CSS classes. For now I limited this to:
- text color
- font weight
- vertical alignment

Changes are only made on the admin files.
The new classes have been added to the stylesheet

### DIFF
--- a/admin/includes/css/stylesheet.css
+++ b/admin/includes/css/stylesheet.css
@@ -866,18 +866,19 @@ input.faint {
 .txt-transparent {
   color: transparent;
 }
-.bold {
-  font-weight: bold;
-}
-.valign-middle {
-  vertical-align: middle;
-}
-.valign-top {
-  vertical-align: top;
-}
-.valign-bottom {
-  vertical-align: bottom;
-}
 .opacity-25 {
   opacity: 25%;
+}
+/* The classes below are copied from Bootstrap 4 and 5 */
+.font-weight-bold {
+  font-weight: 700 !important;
+}
+.align-middle {
+  vertical-align: middle !important;
+}
+.align-top {
+  vertical-align: top !important;
+}
+.align-bottom {
+  vertical-align: bottom !important;
 }
--- a/admin/includes/css/stylesheet.css
+++ b/admin/includes/css/stylesheet.css
@@ -835,3 +835,49 @@ input.faint {
 .control-label {/*for multiple inputs*/
     font-weight: 700;
 }
+
+.txt-black {
+  color:#000;
+}
+.txt-yellow {
+  color: #ff0;
+}
+.txt-orange {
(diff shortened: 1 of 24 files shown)


admin/includes/css/stylesheet.css
```diff
@@ -835,3 +835,49 @@ input.faint {
...
+.txt-pink {
+  color: #ffbaba;
+}
+.txt-transparent {
+  color: transparent;
+}
+.bold {
+  font-weight: bold;
+}
+.valign-middle {
```
lat9, line 872: Could these various `.valign-` use, instead, the [Bootstrap 4](https://getbootstrap.com/docs/4.0/utilities/vertical-align/) names?  That will make it easier to migrate the admin from BS-3.

admin/includes/css/stylesheet.css
```diff
@@ -835,3 +835,49 @@ input.faint {
...
+.txt-white {
+  color:#fff;
+}
+.txt-pink {
+  color: #ffbaba;
+}
+.txt-transparent {
+  color: transparent;
+}
+.bold {
```
lat9, line 869: The Bootstrap 4 class for bold-text is `.font-weight-bold`; could we use that instead?

lat9 (changes requested):
In cases where it's possible (as a pseudo-future-proofing step), I'll suggest that if there's a Bootstrap-4 class associated, that it be used to make a future admin transition to BS4 a tad less onerous.

admin/includes/css/stylesheet.css
Zen4All: Good plan, I had been looking for standard BS3 classes, but did not think of BS4. Will make the adjustments shortly

drbyte:
It's worth also looking at BS5 syntax as well in case there are relevant patterns used there.  

Example: (contrived, because I've NOT checked BS5 for color-related syntax) if `txt-orange` as proposed here actually has a standard in BS5 as `text-orange` then let's use the BS5 style, even if there isn't one for it in BS4.  Again, I'm only using txt-orange for explanation purposes.

And **I don't mean to create additional work**, but if there's no standard for one of these in BS4 or BS5 then a quick check in Tailwind would be sensible, just to try to keep in sync with strong contemporary approaches, which makes future work less onerous and makes it faster for people to get up-to-speed when using ZC.

drbyte:
@Zen4All I like what you're doing here. Thanks!

Zen4All:
I have not had any look at BS5 yet, but will do in the coming days. But it may well be worth it. Maybe we can also skip BS4.

proseLA:
i am a BIG fan of what you are doing here as well.  THANK YOU!

:+1:

PR #4288, opened 2021-04-13
Fixes delete from search

### LINKED ISSUE #4286: Admin product delete fails when being done on page 2 of multiple pages
All 1.5.7 versions.
Go to Admin > Catalog > Categories/Products 
Do a search for "and".  This will return multiple pages of records.
Go to page 2
Hit the trash can on product 104 "Hide Quantity Box" - this is on row 3
You will get a dialog that indicates the admin doesn't know the ID of the product to be deleted.
The delete will fail. 

bad_delete

Also occurs on product 109 (Hide Quantity Box Methods) but not other products on this page.

scottcwilson:
Fixes #4286.

To Test: Install 1.5.8 with default settings and test data set. 
Go to Admin > Catalog > Categories/Products
Search for the string "and"
Go to page 5 and press the delete icon on product "Hide Quantity Box."
Observe that the dialog with the missing product ID shown in issue 4286 appears.

Add this patch; repeat steps above.
Observe dialog is correct and delete works. 

Recommended for backport.


admin/category_product_listing.php
```diff
@@ -917,7 +917,7 @@
...
                         <i class="fa fa-pencil fa-lg" aria-hidden="true"></i>
                       </a>
-                      <a href="<?php echo zen_href_link(FILENAME_CATEGORY_PRODUCT_LISTING, 'cPath=' . $cPath . '&product_type=' . $product['products_type'] . '&pID=' . $product['products_id'] . '&action=delete_product' . (isset($_GET['page']) ? '&page=' . $_GET['page'] : '')); ?>" class="btn btn-sm btn-default btn-delete" role="button" title="<?php echo ICON_DELETE; ?>">
+                      <a href="<?php echo zen_href_link(FILENAME_CATEGORY_PRODUCT_LISTING, 'cPath=' . $cPath . '&product_type=' . $product['products_type'] . '&pID=' . $product['products_id'] . '&action=delete_product' . (isset($_GET['page']) ? '&page=' . $_GET['page'] : '') . (isset($_GET['search']) ? '&search=' . $_GET['search'] : '')); ?>" class="btn btn-sm btn-default btn-delete" role="button" title="<?php echo ICON_DELETE; ?>">
```
drbyte, line 920: @scottcwilson shouldn't this actually be using `$search_parameter` instead of `$_GET['search']`?

scottcwilson: hmm... maybe - but why does edit also use the style delete does?  Should both be changed to look like move and copy?

PR #4289, opened 2021-04-14
fixes php notice and improves search results on fields ending in _id

proseLA:
the previous change on line 163 (eliminating the early return) introduced a PHP notice if no valid keywords are submitted to the function.  the coalesce function on line 205 addresses that issue.

in my previous testing for my category and product pulldowns PR, i also discovered a situation that text being cast as integers would result in `0` which gave results that were not as accurate as i like.  i pushed that change in that PR, but probably should have created its own PR.  given that i wanted to address the (minor) PHP notice i added that fix here.


proseLA:
thanks. i’ll take a look at the parenthesis situation in a bit.

Issue #4287, opened 2021-04-12
ZC158: admin, category-product listing debugs due to function zen_get_discount_calc


torvista:
ZC158, strict reporting
Admin, search for "test" in product/category listing.
The three categories returned generate three pairs of errors:

> [12-Apr-2021 21:40:04 UTC] Request URI: /zencart-158/admin/index.php?cmd=category_product_listing&search=test, IP address: 127.0.0.1
#1  zen_get_products_base_price() called at [...zencart-158\includes\functions\functions_prices.php:685]
#2  zen_get_discount_calc() called at [...zencart-158\admin\category_product_listing.php:654]
#3  require(...zencart-158\admin\category_product_listing.php) called at [...zencart-158\admin\index.php:11]
--> PHP Notice: Undefined index: products_price in ...zencart-158\includes\functions\functions_prices.php on line 421.

> [12-Apr-2021 21:40:04 UTC] Request URI: /zencart-158/admin/index.php?cmd=category_product_listing&search=test, IP address: 127.0.0.1
> #1  zen_get_products_base_price() called at [...zencart-158\includes\functions\functions_prices.php:685]
> #2  zen_get_discount_calc() called at [...zencart-158\admin\category_product_listing.php:654]
> #3  require(...zencart-158\admin\category_product_listing.php) called at [...zencart-158\admin\index.php:11]
> --> PHP Notice: Undefined index: products_priced_by_attribute in ...zencart-158\includes\functions\functions_prices.php on line 423.

torvista:
I see that apart from this, the table rows are unbalanced on a search, along with other issues which I am looking at on this file....

torvista:
Regarding this issue:
ZC157
The category sale/discount information is returned by
zen_get_products_sale_discount('', $category['categories_id'], true);
from
  function zen_get_products_sale_discount($product_id = false, $categories_id = false, $display_type = false) {

ZC158
This function is deprecated

```
function zen_get_products_sale_discount($product_id = false, $categories_id = false, $display_type = false)
{
    trigger_error('Call to deprecated function zen_get_products_sale_discount. Use zen_get_discount_calc() instead', E_USER_DEPRECATED);
```

But the replacement function does not handle categories:
`function zen_get_discount_calc($product_id, $attribute_id = 0, $attributes_amount = 0, $check_qty = 0)`

This is the function being called in the category/product listing file, hence the errors.

mc12345678:
Suggest changing:
[...]

PR #4290, opened 2021-04-15
category-product listing: correct unequal columns for search results

torvista:
Initiated by the unequal row count between category and product results, I've attempted to simplify this mixmatch between the category and product information.

a) Added images and a title to the category listing to bring it closer to the product listing. Show/hide the images I intend to add to another PR. 
b) Added a preview icon to the cats and changed the icon to the popup icon for all: as it is not a **pre-**view but a real view of the catalog, and I didn't think the current icon reflected the function.
Put that icon/link with the name instead of in its own table cell/column, removing colspan.

This makes it much easier to understand the layout.
simplify table row layout (add category image, merge preview icon with name), correct unequal rowcount for search results, change review icon to popup for category and product.

Before:
1 cats

2 prods

3 search

After:
4 cats

5 prods

6 search

Ok, let the shouting begin....

### DIFF
--- a/admin/includes/languages/english/lang.category_product_listing.php
+++ b/admin/includes/languages/english/lang.category_product_listing.php
@@ -10,6 +10,7 @@ $define = [
     'HEADING_TITLE' => 'Categories / Products',
     'HEADING_TITLE_GOTO' => 'Go To:',
     'TABLE_HEADING_ID' => 'ID',
+    'TABLE_HEADING_IMAGE' => 'Image',
     'TABLE_HEADING_CATEGORIES_PRODUCTS' => 'Categories / Products',
     'TABLE_HEADING_PRICE' => 'Price/Special/Sale',
     'TABLE_HEADING_QUANTITY' => 'Quantity',
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -564,7 +564,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
           //     ),
           // );
           //
-          // Observer notes:  
+          // Observer notes:
           // - Be sure to check that the $p2/$extra_headings value is specifically (bool)false before initializing, since
           //   multiple observers might be injecting content!
           // - If heading-columns are added, be sure to add the associated data columns, too, via the
@@ -599,7 +599,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
           //     ),
           // );
           //
-          // Observer notes:  
+          // Observer notes:
           // - Be sure to check that the $p2/$extra_headings value is specifically (bool)false before initializing, since
           //   multiple observers might be injecting content!
           // - If heading-columns are added, be sure to add the associated data columns, too, via the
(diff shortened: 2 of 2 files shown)


proseLA:
tor,
for starters, should not the name be correct unequal **columns** for search results?  the 'problem' only exists if categories and products are displayed on the search results.

and is not the problem right here:

https://github.com/zencart/zencart/blob/35dc5cbc67b99851217a12e580c6130f7b67fd63/admin/category_product_listing.php#L656-L659

and could the column issue simply be resolved with the additional of the following line right after 656:
```
<?php echo $search_result ? '<td class="hidden-sm hidden-xs">&nbsp;</td>' : ''; ?>
```

taking a valid problem, and then trying to stuff additional things into addressing the problem, ie category images, seems trying to do a bit much to me.  and frankly i love being aggressive with changes (just in case you have not noticed!)

in taking a thorough examination of your PR, i also do not understand why this line is in the code (not part of your PR, btw):

https://github.com/zencart/zencart/blob/35dc5cbc67b99851217a12e580c6130f7b67fd63/admin/category_product_listing.php#L654

i do not really deal with product attributes, but that functions parms are 
```
($product_id, $attribute_id = 0, $attributes_amount = 0, $check_qty = 0)
``` 
and we are passing an empty string, the category_id and true.  which is no doubt why i am seeing a bunch of debug-logs while testing.

torvista:
> taking a valid problem, and then trying to stuff additional things into addressing the problem, ie category images, seems trying to do a bit much to me.

Yes. 
But, I've delved into this file multiple times to address table faults, and each time found it a headache/difficult to figure out/not intuitive at all to correct. 
Maybe I got brain freeze or am just getting old, but did not readily see the simple fix. I felt that the three scenarios of products/categories/mixed search results had been complicated further by the addition of the product images, and would be easier to work with if they **both** had that column and the preview column was merged.

Subsequently I found it easier to figure out and correct the problem.

The other thing about the bad function call I've mentioned here
https://github.com/zencart/zencart/issues/4287

admin/includes/languages/english/lang.category_product_listing.php
```diff
@@ -10,6 +10,7 @@
     'HEADING_TITLE' => 'Categories / Products',
     'HEADING_TITLE_GOTO' => 'Go To:',
     'TABLE_HEADING_ID' => 'ID',
+    'TABLE_HEADING_IMAGE' => 'Image',
```
drbyte, line 13: Is it necessary to have the word "Image" as a column-heading?  ie: do people need to know that the pictures below that heading are images by giving the heading a title, instead of leaving it blank?

drbyte: (I'm thinking here about minimizing the addition of language strings that need translating/defining/upgrading.)

drbyte: I guess it's moot. If it's not there then someone's gonna want to add it for OCD reasons. :D

torvista: On a tangent, I could prepare a PR that provides a link on the page to hide/show the images colum. I generally do not show images, for speed.

PR #4292, opened 2021-04-16
Non-numeric value on max_length of datetime field type

mc12345678:
Assigns the proposed maximum PHP length for a datetime field as proposed
in the MySql documentation at:
https://dev.mysql.com/doc/refman/8.0/en/datetime.html

Solution here is as posted at:
https://www.zen-cart.com/showthread.php?227778-A-non-numeric-value-encountered-in-functions_general-php&p=1378411#post1378411

Incorporation resolved warning logging seen on a live site.

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -715,5 +715,8 @@ class queryFactoryMeta {
     $rgx = preg_match('/^[a-z]*/', $type, $matches);
     $this->type = $matches[0];
     $this->max_length = preg_replace('/[a-z\(\)]/', '', $type);
+    if ($type == 'datetime') {
+      $this->max_length = '26';
+    }
   }
 }


drbyte:
https://github.com/zencart/zencart/commit/fadade0e55e79f00bab24826c546123327345d07#diff-4dfb6ff3bda4dd5d8c5413191a41586396e3f3a46425bd7cefd7dcbe5cd7990c

mc12345678:
Any particular reason that one of the core of core files pre-limits the length that can be entered instead of one choosing to limit the length in the template if it exceeds a certain value? Yes, I'm trying to think of "good" examples of how the extra time fractions could "best" be used. All thought of so far have been related to some sort of "go" or "stop" button...

drbyte:
Isn't that over-thinking the situation?

mc12345678:
Fine with this being closed. Point was why should Zen Cart unnecessarily limit/exclude capability instead of just ensuring that the data going through is of the correct type/size? If the template side result is too large, then a store owner can correct the specific field(s), too small and the core code has to be touched.

PR #4298, opened 2021-04-21
Make old template use HTML doctype

scottcwilson:


lat9:
For what purpose?  There's much, much more to an HTML-5 rendering than just the HTML doctype.

scottcwilson:
The Google Lighthouse test complains if it's not there.   This will also bring it into alignment with Responsive Classic.

PR #4303, opened 2021-04-26
Add "Customer Groups" feature

drbyte:
Customers can be assigned to multiple groups.

Modules can use group-related functions to check customer group membership in order to enable/disable certain features as required.


torvista:
So would I be correct in concluding that this base is 
a) how in-built Group Pricing could/should have been implemented (and so will continue to be available in parallel with this for legacy support),
b) how plugin Dual Pricing should be implemented?

drbyte:
Given that `ot_group_pricing` is directly tied to `group_pricing`, it would require some careful work to convert things over.
However, it would be better implemented this way instead of being so tightly limited as it was.
No decision has been made regarding legacy support or running it in parallel.

As to "Dual Pricing", there are a ton of things that one could be referring to when mentioning it. It does dozens of things independent of what group a customer is assigned to; however the part of it that links to specific customers in some sort of grouping strategy could probably be easily changed to use this infrastructure.

PR #4308, opened 2021-05-01
Don't call php_uname if exec disabled

scottcwilson:
Doing so will cause a PHP log on systems where php_uname() "has been disabled for security reasons."


drbyte:
Hmmm ... on second thought, this doesn't actually solve the issue does it?

scottcwilson:
It's a pretty rare condition.  Hopefully hosters are becoming less clueless over time. :)

scottcwilson:
Disabling exec is not uncommon at all.  Disabling php_uname is unique to this induhvidual.

PR #4311, opened 2021-05-02
V157 category product listing c path navigation

### LINKED ISSUE #4310: ZC1.5.7c Catalog navigation within admin is inconsistent, cPath does not represent the full current path
After the significant changes made to ZC 1.5.7 in the catalog navigation area, when navigating the catalog within the admin,
1 the Back button does not consistently return to the category that is the parent of the current category, 
2 the GOTO dropdown sets `cPath` to the category being sought but not the `cPath` path of the destination category and
3 when clicking on a sub-category that has a full link that includes the cPath parent/child relationship the page that is open resolves to a `cPath` of just the current category.

Currently running on PHP 8.0; however, had noticed in PHP 7.2 on up.

MySQL:  5.7.29
No Plugins interacting with `cPath` as through `$_POST/$_GET` modification/sanitization.
Compared against 1.5.6.

Login to admin with demo product.
navigate to Catalog, Categories/Products
To show this as logically as I can think:
Click specifically on the area between the words 'A Top Level Cat' and '0 of 0 active'. (row associated with category 33 that responds to the javascript of the page.
This navigates to the expected `admin/index.php?cmd=category_product_listing&page=0&cPath=33`
Now click in the same fashion for 'SubLevel 2 A' which will navigate to `admin/index.php?cmd=category_product_listing&page=0&cPath=34` which has the unexpected `cPath=34` instead of the full cPath of `cPath=33_34`.
Again do this for 'Sub Sub Cat 2A1'.  Arrive at: `admin/index.php?cmd=category_product_listing&page=0&cPath=40` Again `cPath=40` instead of `cPath=33_34_40`.
Ok, so what's the big deal?
Press the back button at the bottom of the page:
Tada, returned to the root of the catalog.

Now, instead of clicking on the area between the category text and the designation of product within the category, select the text of the category name: 'A Top Level Cat':
Again: `admin/index.php?cmd=category_product_listing&page=0&cPath=33`
Now select the text of the category 'SubLevel 2 A':
This time: `admin/index.php?cmd=category_product_listing&page=0&cPath=33_34`.
Now select the text of the category 'Sub Sub Cat 2A1' and:
This time: `admin/index.php?cmd=category_product_listing&page=0&cPath=33_34_40`
Now press the back button at the bottom of the page...
`admin/index.php?cmd=category_product_listing&page=0&cPath=33_34`

Ok, cool.
Now, go to the top of the screen and use the Go To: dropdown.  Actively select 'Sub Sub Cat 2A1' and note that again the path result is: `admin/index.php?cmd=category_product_listing&page=0&cPath=40`, again selecting the back button at the bottom of the page returns to the root of the store...

So what's wrong?
The javascript process uses the link information of `data-cid` which is populated with the category_id of the destination category instead of building a cPath based off of the current category concatenated with the parent's cPath.
The goto dropdown post's its value as cPath using only the category id for each category and not the cPath information. Correcting this takes two parts, to make the value for each category "record" have the cPath for the 'id' and to feed zen_draw_pull_down_menu the cPath associated with the current category. 

Unfortunately, it appears that to apply corrections to this requires bringing in the catalog side function: `zen_get_generated_category_path_rev`, at least it seems to be the only function that I've found that would assemble that full cPath "correctly". If there is another such function the I could potentially use it in place of what I will be proposing in a moment.

While working that as well and reviewing the associated page links, it got to me that the browser link included a page parameter for anything that wasn't page 2 or greater (e.g., page=0 predominantly). I addressed that as well, but will include as a separate commit where possible.

mc12345678:
Restore category_product_listing navigation 806d6a8
Generated shared functions file to allow admin access to `zen_get_generated_category_path_rev` from the admin.
Fixes #4310

Refactor to display page parameter only on pages 2 and up.


mc12345678:
However closely or distant this may relate, issue #4170 ***may*** provide a portion of the solution attempted here. I have not attempted to apply that where the "dropdown" is used. There is one fix/correction that is unrelated to the "dropdown" but directly related to `cPath` navigation. There are other commits in this PR to address page numbering, code refactoring, SQL statement refactoring, generation/incorporation of a shared function file and just general code spacing (though may not include curly parentheses realignment).

scottcwilson:

Issue #4312, opened 2021-05-02
admin, customer groups: sql error with Mysql Mode and php warnings x 2


torvista:
ZC158. mysql-5.7.32, php8, strict reporting, 

1) Development shop
Added new table using sql from upgrade script, in admin SQL patch. No warnings.

On customer groups page load, two php warnings:

> Request URI: /SHOP/ADMIN/index.php?cmd=customer_groups
> IP address: 127.0.0.1
> #1  require(ADMIN/customer_groups.php) called at [ADMIN/index.php:11]
> --> PHP Warning: Undefined variable $gInfo in ADMIN\customer_groups.php on line 213.
> 
> Request URI: /ADMIN/index.php?cmd=customer_groups
> IP address: 127.0.0.1
> #1  require(ADMIN/customer_groups.php) called at [ADMINindex.php:11]
> --> PHP Warning: Attempt to read property "group_id" on null in ADMIN\customer_groups.php on line 213.

2. Vanilla ZC158.
Clean install.  On customer groups pageload, got SQL error. 

> [02-May-2021 19:38:29 Europe/Madrid] Request URI: /zencart-158/admin/index.php?cmd=customer_groups, IP address: 127.0.0.1
#1  trigger_error() called at [...\zencart-158\includes\classes\db\mysql\query_factory.php:657]
#2  queryFactory->show_error() called at [...\zencart-158\includes\classes\db\mysql\query_factory.php:624]
#3  queryFactory->set_error() called at [...\zencart-158\includes\classes\db\mysql\query_factory.php:265]
#4  queryFactory->Execute() called at [...\zencart-158\admin\customer_groups.php:100]
#5  require(...\zencart-158\admin\customer_groups.php) called at [...\zencart-158\admin\index.php:11]
--> PHP Fatal error: 1055:Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'zencart-158_local.cg.group_id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by :: SELECT cg.*, count(ctg.customer_id) as customer_count                          FROM customer_groups cg                         LEFT JOIN customers_to_groups ctg USING (group_id)                         GROUP BY ctg.group_id                          ORDER BY group_name, group_id LIMIT 0, 1 ==> (as called by) ...\zencart-158\admin\customer_groups.php on line 100 <== in ...\zencart-158\includes\classes\db\mysql\query_factory.php on line 657.

[...]

PR #4313, opened 2021-05-03
Fix query

Zen4All:
See issue #4312


drbyte:
Is the change from `USING` to `ON` really needed?

Zen4All:
No, you're right we can do without. I'll change it back

PR #4320, opened 2021-05-07
admin, layout controller, html corrections

torvista:
Never sure whether to squash these minor things or leave separate.
Doing this prior to other mods.


drbyte:
> Never sure whether to squash these minor things or leave separate.

This format is good because it helps with the review process.

Additional preference would be that when a bug/problem/error is fixed that it be a separate PR from any formatting updates.

drbyte:
> Doing this prior to other mods.

Like what?

torvista:
I'm starting work on a new template,and I know it will be a slow process.
Although in shopfront I can view it privately with parameter t=, I can't edit the layout boxes without making it live in the interim.
I was thinking of doing something to allow selection of template in this page to allow that.
But, after more looking, I think it is non-trivial to implement and not something on which I can justify the time spent.

drbyte:
> Although in shopfront I can view it privately with parameter t=, I can't edit the layout boxes without making it live in the interim.
> I was thinking of doing something to allow selection of template in this page to allow that.
> But, after more looking, I think it is non-trivial to implement and not something on which I can justify the time spent.

Agreed, some parts of it are definitely non-trivial.
This may do what you're seeking though: #4321

PR #4321, opened 2021-05-08
Layout Boxes Controller enhancements

drbyte:
Layout Boxes Controller enhancements
- choose from available templates instead of only editing the live template
- reset by copying settings from another template or from original ZC defaults
- you can now reset `template_default`'s settings as a convenient way to preserve preferences
- "single-column" settings removed, in favor of left/right

Note: template selection choices are based on templates whose directory exists catalog-side, regardless of whether registered in database or enabled in Admin template selection screen.

Screen Shot 2021-05-08 at 5 18 42 PM

Screen Shot 2021-05-08 at 5 18 27 PM

### DIFF
--- a/admin/includes/languages/english/lang.layout_controller.php
+++ b/admin/includes/languages/english/lang.layout_controller.php
@@ -1,24 +1,28 @@
 <?php
 /**
- * @copyright Copyright 2003-2020 Zen Cart Development Team
+ * @copyright Copyright 2003-2021 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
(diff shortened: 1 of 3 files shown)


lat9:
Er, isn't that single-column setting used for various header-based sideboxes (e.g. search_header.php)?

drbyte:
> Er, isn't that single-column setting used for various header-based sideboxes (e.g. search_header.php)?

It is optionally, yes.
There's probably a better way to handle that optional approach. It was done that way only because the 'single' stuff is basically otherwise never used.

PR #4324, opened 2021-05-09
add coupon amount to order->info via ot_coupon class

proseLA:


proseLA:
https://www.zen-cart.com/showthread.php?200407-Edit-Orders-v4-0-Support-Thread&p=1380563#post1380563

perhaps, it is time to rethink the order total module?  it does nothing other than spit out what is in the order->info array.  and needless to say it is very easy for the math to go wrong.  why are we keeping a running total as opposed to recalculating the total provide by all of the active order_total modules?  apologies if this is thread drift, but it seems related.

Issue #4329, opened 2021-05-13
ZC158, Admin, Whos online: PHP Warning: Undefined array key "products_type"  line 298.


torvista:
php8, strict reporting

Admin, whos online: show admin IP addresses.

Shopfront, add product to cart.

Refresh whos online: get debug

> Request URI: /zencart-158/admin/whos_online.php?na=0, IP address: 127.0.0.1
> --> PHP Warning: Undefined array key "products_type" in ....\zencart-158\admin\whos_online.php on line 298.

https://github.com/zencart/zencart/blob/5011ed962f8e10d4931a876c9c713f1904cbd61e/admin/whos_online.php#L298

PR #4331, opened 2021-05-13
add products_type to shopping_cart array

proseLA:
Re: #4329

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -1382,7 +1382,7 @@ class shoppingCart extends base
                     'products_quantity_order_max' => (float)$products->fields['products_quantity_order_max'],
                     'products_quantity_mixed' => (int)$products->fields['products_quantity_mixed'],
                     'products_mixed_discount_quantity' => (int)$products->fields['products_mixed_discount_quantity'],
-                    'products_type' => $products->fields['products_type'],
+                    'products_type' => (int)$products->fields['products_type'],
                 ];
             }
         }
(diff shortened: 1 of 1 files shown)


mc12345678:
Source of the change to whos_online: https://www.zen-cart.com/showthread.php?228000-If-Customer-has-items-in-cart-and-Admin-clicks-on-item-to-view-gets-blank-screen

As I understand/consider, sanitization is a measure taken to ensure that information is of the expected type where/when either it is initially obtained, about to be pushed out and to ensure that if it were something else that it is properly handled. Stated that it is already an integer in the database, but what is in place to ensure that the database field at the time of being accessed is actually of the integer type? What's to say that something hasn't been done outside of "normal" flow to change the data type or make some other modification? Not to say to be paranoid about it, like oh my gosh everything is turned upside down, but to retain a healthy level of verification/assignment.

Further, if a store only used core code, no add ons, no modifications, no plugins, or other software changes then based on the history of the program which was the first eCommerce program to be PCI certified, it would be expected that the data would remain safe and unmodified; however, as it *is* open source, it appears the core developers have been taking the action necessary to preclude the core software from being the source of a data issue.  I mean they went so far as implementing the admin sanitizer code... I mean, protecting the "authorized" admin's data entry?! 

Let's also consider the fact that one of the historical issues with PHP is nearly its absence of strictness to data handling. Sanitizing/casting of the data forces the data to be meet some level of expected type and to be able to be handled in a particular way, arrays for example have characteristics that integers do not (e.g. carry an aspect of counting items within the object which has little meaning for an integer).

drbyte:
Thanks for the idea. 
It's being handled with a bespoke lookup instead.

PR #4336, opened 2021-05-16
Remove unused Define, and some code modernising, in admin header.php

Zen4All:
- In if statement:
  - Remove unused define ```FILENAME_DEFINE_LANGUAGE```
  - Remove ```FILENAME_PRODUCTS_OPTION_NAME``` Seems not really needed in my opinion
- Replace ```and``` with ```&&```
- Replace ```sizeof``` with ```count```
- Short arrays


scottcwilson:
@Zen4All Can you squash this PR so we can merge it?

Zen4All:
I have  now idea how to do that.

PR #4345, opened 2021-05-21
Provide variable to disable product tax column display

scottcwilson:
People using external tax calculation software like Taxcloud will find it confusing to see the an inaccurate tax column for products.  We can make this easily switchable in the orders and invoice admin modules.


lat9 (commented):
Shouldn't that `$show_product_tax` variable have a means via notify/observe for an observer to make its custom modifications?

scottcwilson:
I'm just using the pattern from invoice.php and orders.php for the suppression of tax info.  Should all three support a notifier?  Maybe.

lat9:
I'm not sure which 'all three' are involved, but a notification would surely make it easier for customizations.

PR #4347, opened 2021-05-22
Remove php notice when searching for non-existent description

mc12345678:
If function is used to look up tax rates based on a description, even
TEXT_UNKNOWN_TAX_RATE is not present and therefore leads to the problem
described at this ZC forum thread:
https://www.zen-cart.com/showthread.php?228051-tax_rate-undefined-index

Chose the specific use of setting the tax rate to 0.0 so that one could
"easily" consider other rates to apply if necessary or some other
variation. Another alternative to this could have been to upfront
evaluate `$description` and if it equaled TEXT_UNKNOWN_TAX_RATE to then
continue the loop which would skip the tax assignment.

### DIFF
--- a/includes/functions/functions_taxes.php
+++ b/includes/functions/functions_taxes.php
@@ -2,10 +2,10 @@
 /**
  * functions_taxes
  *
- * @copyright Copyright 2003-2020 Zen Cart Development Team
+ * @copyright Copyright 2003-2021 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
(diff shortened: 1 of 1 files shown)


drbyte:
When looking at that thread yesterday I found myself wondering: how does a non-existent tax rate even get called?

mc12345678:
Wish I knew/understood their situation, although the claim seemed to be that the unknown tax rate description was being fed to this function which would cause this problem, but also I didn't find anything in a default install that would have led to or hinted that there may be this issue with the function...

I also haven't installed the plugin to see how the original problem could be generated on the purpose.

Issue #4349, opened 2021-05-28
admin, countries, disable on other than page A: php warning and incorrect page redirect


torvista:
ZC158, php8, strict

Go to any page (not A), disable a country: page reverts to A, no infobox and debug warning:

Request URI: /zencart-158/admin/index.php?cmd=countries&cID=16
--> PHP Warning: Undefined variable $cInfo in \zencart-158\admin\countries.php on line 193.

torvista:
This error is not fixed by the indicated PR.  4350
It is fixed by this one.https://github.com/zencart/zencart/pull/4351

PR #4351, opened 2021-05-29
ZC158: admin, countries: fix handling of page as letter instead of number

torvista:
fixes https://github.com/zencart/zencart/issues/4349

### DIFF
--- a/admin/countries.php
+++ b/admin/countries.php
@@ -8,7 +8,7 @@
 require 'includes/application_top.php';
 
 $action = (isset($_GET['action']) ? $_GET['action'] : '');
-$currentPage = (isset($_GET['page']) && ctype_alpha($_GET['page'][0]) ? $_GET['page'][0] : '');
+$currentPage = (!empty($_GET['page']) && ctype_alpha($_GET['page'][0]) ? $_GET['page'][0] : '');
 
 if (!empty($action)) {
   switch ($action) {
--- a/admin/countries.php
+++ b/admin/countries.php
@@ -8,7 +8,7 @@
 require 'includes/application_top.php';
 
 $action = (isset($_GET['action']) ? $_GET['action'] : '');
-$currentPage = (isset($_GET['page']) && $_GET['page'] != '' ? (int)$_GET['page'] : 0);
+$currentPage = (isset($_GET['page']) && ctype_alpha($_GET['page'][0]) ? $_GET['page'][0] : '');
 
 if (!empty($action)) {
   switch ($action) {
@@ -44,7 +44,7 @@ if (!empty($action)) {
                         status = " . (int)$status . "
                     WHERE countries_id = " . (int)$countries_id);
       zen_record_admin_activity('Country updated: ' . $countries_iso_code_3, 'info');
-      zen_redirect(zen_href_link(FILENAME_COUNTRIES, ($currentPage != 0 ? 'page=' . $currentPage . '&' : '') . 'cID=' . $countries_id));
+      zen_redirect(zen_href_link(FILENAME_COUNTRIES, ($currentPage !== '' ? 'page=' . $currentPage . '&' : '') . 'cID=' . $countries_id));
       break;
     case 'deleteconfirm':
       $countries_id = zen_db_prepare_input($_POST['cID']);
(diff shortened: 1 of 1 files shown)


drbyte:
Please explain how you decided upon `ctype_alpha($_GET['page'][0])`, 
particularly:
- `ctype_alpha()`
- `$_GET['page'][0]`

torvista:
It was one of several ways I found to get a single character and only allow a letter....seemed the shortest/clearest.

proseLA:
> It was one of several ways I found to get a single character and only allow a letter....seemed the shortest/clearest.

i still do not understand it.  `$_GET['page']` is not array, or at least it should not be.  but you are asking for the first element of a non-array.  

in addition, the page should be a number.  so now you are asking for a letter?

so, i'm not real sure what you are doing.  and it does not seem clear to me at all.

drbyte:
In v1.5.8 the pagination for the Countries page is alphabetic.

torvista:
> $_GET['page'] is not array
You can treat a string as an array.
https://www.php.net/manual/en/language.types.string.php#language.types.string.substr

In any case, no doubt you can achieve the same result with preg_match, but that would have taken me too long to figure out, so its up for grabs. Or maybe there is a built-in sanitation for only-letters?

proseLA:
> You can treat a string as an array.
> https://www.php.net/manual/en/language.types.string.php#language.types.string.substr
> 
interesting, you learn something new every day!

I rarely use countries, so I will give my thoughts, and then move on.

- while I like the new letter filter, I’m not crazy about the reuse of the `$_GET[‘page’]` var for this task.  

- the problem is now the footer and how many countries are displayed on the page?

- I suck at regular expressions as well.  but I would never use it as an excuse when committing to a repo.  I’ve spent an inordinate amount of time on some things in this repo, with no benefit to any of my client sites.

- in looking at `$currentPage`, its only use is in the following code `($currentPage != 0 ? 'page=' . $currentPage . '&' : '')` which is used 13 times.  can we just set a new var for the current page parameter after the assignment of the currentPage (or rename the currentPage var), and then use that var throughout the script?  I think it makes the code far more legible.

drbyte:
> > You can treat a string as an array.
> > https://www.php.net/manual/en/language.types.string.php#language.types.string.substr
> 
> interesting, you learn something new every day!

I probably would have just used `substr($foo, 0, 1)`

> while I like the new letter filter, I’m not crazy about the reuse of the `$_GET[‘page’]` var for this task.

Not sure why it matters. Given that `page` is used consistently for pagination across all admin pages, it makes sense to leave it as-is.

> the problem is now the footer and how many countries are displayed on the page?

I don't see any problem.

> regular expressions

`substr` is faster than regex

> in looking at `$currentPage`, its only use is in the following code `($currentPage != 0 ? 'page=' . $currentPage . '&' : '')` which is used 13 times.  can we just set a new var for the current page parameter after the assignment of the currentPage (or rename the currentPage var), and then use that var throughout the script?  I think it makes the code far more legible.

This has merit.

proseLA:
> > while I like the new letter filter, I’m not crazy about the reuse of the `$_GET[‘page’]` var for this task.
> 
> Not sure why it matters. Given that `page` is used consistently for pagination across all admin pages, it makes sense to leave it as-is.

https://github.com/zencart/zencart/blob/bcfbb0081d01f7b5dc3ca3fc5e8225570ac4c33c/admin/countries.php#L112
https://github.com/zencart/zencart/blob/bcfbb0081d01f7b5dc3ca3fc5e8225570ac4c33c/admin/includes/classes/split_page_results.php#L65-L72

again, i have no dog in this fight.  current page is expected to be integer.  there have been problems in the past with the split_page_results and mods that attempt to use it with complex queries.  i think introducing letters as page numbers and then passing them to this class makes it more confusing, especially if one is attempting to build some sort of logic into the `$_GET['page']` var, which it looks like we are trying to do.  i do not like this approach.

showing all countries that start with a 'D' for example, IMO, is not pagination.  if there were 25 countries that started with a D, and the max search results was 20, then we would have 2 pages.  and then we are talking pagination.

combining filtering and pagination seems to be a potential recipe for problems down the line.

> 
> > the problem is now the footer and how many countries are displayed on the page?
> 
> I don't see any problem.
> 

every page displays this:

Displaying 0 to 0 (of 245 Countries)

that is all i was referring to...

drbyte:
> Displaying 0 to 0 (of 245 Countries)

Good point.

torvista:
So, in summary the letter links are actually filters, not pages.
Since the quantity of countries per letter is limited and not going to increase, why not just forget pagination for countries, just do the filter and name variables accordingly?

scottcwilson:
I agree that pagination is overkill for countries.  I would even drop the filter.  Load the page, control F, find your country.

torvista:
Drop the filter and just display a complete list?

scottcwilson:
That's what I would do.

PR #4353, opened 2021-06-01
remove extra ampersand added during previous refactor

proseLA:
leading ampersand creates notice in admin/includes/classes/split_page_results.php on line 296

PHP Notice: Undefined offset: 1


proseLA:
as an aside, the extra ampersand was there during the initial commit; which i suppose was a previous refactor:

https://github.com/zencart/zencart/blob/698e46bed1603679d484f5b58ccb918d92f2ca87/admin/coupon_admin.php#L1073

PR #4354, opened 2021-06-01
Support product page return if file upload problem

mc12345678:
If there is a problem with the upload of a file as determined by
the upload class' parse method, then regardless the page to which to
return this presents a message and stops the addition of the remainder of
the product conditions.

Without this code, the product could be added to the cart without the
uploaded information (specifically removed because of issue bringing the
code to this section) even if the file upload attribute was marked as
required.

Discussed in Zen Cart forum at: https://www.zen-cart.com/showthread.php?228049-items-added-to-Shopping-Cart-despite-errors

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -1986,6 +1986,9 @@ class shoppingCart extends base
                                         break;
                                     }
                                 } else {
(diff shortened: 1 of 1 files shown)


drbyte:
If the customer is a malicious user, what could they submit that could get echoed back to the browser in a problematic way? 

Simplest/quickest protection: does any of this actually output back to the browser anything entered by the visitor? Or only things already in the database (or language files) via the Admin?

PR #4355, opened 2021-06-02
ZC158, admin, orders: correct BStrap columns to 3

torvista:
I assume this is wrong:
Clipboard01

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -452,7 +452,7 @@ if (!empty($action) && $order_exists == true) {
 
         <div class="row noprint">
           <div class="form-inline">
-            <div class="form-group col-xs-4 col-sm-3 col-md-3 col-lg-3">
+            <div class="form-group col-xs-3 col-sm-3 col-md-3 col-lg-3">
                 <?php
                 echo zen_draw_form('search', FILENAME_ORDERS, '', 'get', '', true);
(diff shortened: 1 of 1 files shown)


drbyte:
> I assume this is wrong:

Why is it wrong? 
Are you assuming it's wrong merely because it breaks pattern with the other `3`'s used for other breakpoints?
Or are you saying it's wrong because it actually makes the display unusable on the smaller screens to which it applies?

drbyte:
Using `3`'s it looks crammed-together like this:
Screen Shot 2021-06-02 at 3 18 05 PM

Using `6`'s it spreads it out for usability like this:
Screen Shot 2021-06-02 at 3 17 44 PM

torvista:
Does it not always have to add up to 12?

drbyte:
> Does it not always have to add up to 12?

No, but if it's more than 12 then it "line-breaks" to the next "line", allowing up to 12 across in any given row.

If one intentionally always wants a new "row" then insert a new div for that new "row", but if you want things to expand/collapse/span/extend/whatever as-needed for various screen-sizes, mixing various numbers within one "row" will still flow based on a max of 12 per row before automatically breaking to next.

torvista:
Thanks. Must read more.

PR #4358, opened 2021-06-03
ZC158, admin, orders: multiple minor non-functional revisions

torvista:
Based on IDE inspections, prior to submitting a functional mod to allow batch updates of multiple orders (multi-language) from the order listing page.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -30,9 +30,9 @@ if (!isset($_GET['list_order'])) $_GET['list_order'] = '';
 if (!isset($_GET['page'])) $_GET['page'] = '';
 
 include DIR_FS_CATALOG . DIR_WS_CLASSES . 'order.php';
-$show_including_tax = (DISPLAY_PRICE_WITH_TAX == 'true');
+$show_including_tax = (DISPLAY_PRICE_WITH_TAX === 'true');
 // prepare order-status look-up list
-$orders_status_array = array();
+$orders_status_array = [];
 $orders_status = $db->Execute("SELECT orders_status_id, orders_status_name
                                FROM " . TABLE_ORDERS_STATUS . "
                                WHERE language_id = " . (int)$_SESSION['languages_id'] . "
@@ -41,7 +41,7 @@ foreach ($orders_status as $status) {
   $orders_status_array[$status['orders_status_id']] = $status['orders_status_name'];
 }
 
-$action = (isset($_GET['action']) ? $_GET['action'] : '');
+$action = ($_GET['action'] ?? '');
 $order_exists = false;
 if (isset($_GET['oID']) && empty($_GET['oID'])) {
   unset($_GET['oID']);
@@ -49,7 +49,7 @@ if (isset($_GET['oID']) && empty($_GET['oID'])) {
 if (isset($_GET['oID'])) {
   $_GET['oID'] = (int)$_GET['oID'];
 }
-if ($action == 'edit' && !isset($_GET['oID'])) {
+if ($action === 'edit' && !isset($_GET['oID'])) {
   $action = '';
 }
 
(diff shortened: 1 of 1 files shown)


admin/orders.php
```diff
@@ -62,7 +62,7 @@
 if ($oID) {
   $orders = $db->Execute("SELECT orders_id
                           FROM " . TABLE_ORDERS . "
-                          WHERE orders_id = " . (int)$oID);
+                          WHERE orders_id = " . $oID);
```
drbyte, line 65: Our practice is to INTENTIONALLY leave the cast in the SQL queries because then when one is visually scanning the code it is obvious that no SQL-injection risk exists. Otherwise one must visually check around to find out where it was last sanitized ... or not.

admin/orders.php
```diff
@@ -267,7 +267,7 @@
         if (file_exists(DIR_FS_CATALOG_MODULES . 'payment/' . $order->info['payment_module_code'] . '.php')) {
           require_once(DIR_FS_CATALOG_MODULES . 'payment/' . $order->info['payment_module_code'] . '.php');
           require_once(DIR_FS_CATALOG_LANGUAGES . $_SESSION['language'] . '/modules/payment/' . $order->info['payment_module_code'] . '.php');
-          $module = new $order->info['payment_module_code'];
+          $module = new $order->info['payment_module_code']();
```
drbyte, line 262: Meh. While I won't reject the PR for this, I'm not convinced it's necessary. It's safe to "new up" a class without passing empty parentheses to it. Or, maybe I'm missing some upcoming deprecation that's threatened by the lack of parens?

admin/orders.php
```diff
@@ -1270,10 +1270,10 @@ function couponpopupWindow(url) {
                     }
 
                     $show_difference = '';
-                    if ((strtoupper($orders->fields['delivery_name']) !== strtoupper($orders->fields['billing_name']) && trim($orders->fields['delivery_name']) !== '')) {
+                    if (trim($orders->fields['delivery_name']) !== '' && (strtoupper($orders->fields['delivery_name']) !== strtoupper($orders->fields['billing_name']))) {
```
drbyte, line 1273: When rewriting with this swap, it now seems like switching out `trim() !== ''` with `!empty()` would be sensible, on both lines.

admin/orders.php
torvista: Comes from here I think:
https://www.php-fig.org/psr/psr-12
4-classes-properties-and-methods

drbyte: Ya, I suppose that's probably what the IDE is holding to.

scottcwilson:
Please resolve conflicts and squash.

scottcwilson:
Doing this to help you get the functional update in place, but please hold off on more stylistic changes until we're done with 1.5.8.

torvista:
`...until we're done with 1.5.8.`
What does that imply?

scottcwilson:
I thought the meaning was quite plain.  I'm not sure I can make it clearer, but here goes: 

> please hold off on more stylistic changes ...

Fix bugs and add features by all means but please don't submit PRs that just change `array()` to `[]` (or similar) that don't fix any bugs or add new features.

> until we're done with 1.5.8.

We are not yet done with 1.5.8.  At some point in the future, it will be released.  At some point after that, we'll be done with 1.5.8.  No, I don't have a timeline.   Once this all happens and a new branch is created for the following major release, we can do more code cleanup work.

torvista:
The stylistic bit was clear, hence I quoted the unclear bit about "being done".

> We are not yet done with 1.5.8....

But since what "done" means is never (publicly) defined, and since 157d is just released, one would assume that 1.5.8 will be kicking around for quite some time yet.
Obviously "it will be released..." at some point, and at another point it will be "done". Hard to tell if you are being serious there?
Again this touches on the issue I have with this project: "done" and the steps to get there are assumed or guessed at.
(And yes I'll join the discussion group, before you tell me to join the discussion group.)

Meanwhile, functionality is added ad hoc/at the whim of the contributor and here is an example.

https://github.com/torvista/Zen_Cart-Batch_Order_Update

It isn't very elegant but it works, so I can PR it as is or people can poke it with sticks over there....

PR #4359, opened 2021-06-03
Fix expire sql cache only if file/db entry exist

jacobshilitz:
This fixes this PHP Warning that happen when the file never existed:
```
PHP Warning: unlink(/var/www/html/cache/zc_7eb4ed1f36d90140477d9474b61e0e26.sql): No such file or directory in /var/www/html/includes/classes/cache.php on line 92
```

In order to be able to use the `existing sql_cache_exists()` method I update it to be null by default, if no cache time is passed to it.

### DIFF
--- a/includes/classes/cache.php
+++ b/includes/classes/cache.php
@@ -19,14 +19,14 @@ if (!defined('IS_ADMIN_FLAG')) {
  */
 class cache extends base {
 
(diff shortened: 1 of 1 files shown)


scottcwilson:
Thanks for your submission, @jacobshilitz !

PR #4361, opened 2021-06-05
Add more notifiers to shopping_cart for after-add/update

zenexpert:
Add notifiers for successful add to cart in case one wants to hook additional actions based on product(s) added to cart

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -1864,6 +1864,7 @@ class shoppingCart extends base
 // display message if all is good and not on shopping_cart page
                     if ((DISPLAY_CART == 'false' && $_GET['main_page'] != FILENAME_SHOPPING_CART) && $messageStack->size('shopping_cart') == 0) {
                         $messageStack->add_session('header', ($this->display_debug_messages ? 'FUNCTION ' . __FUNCTION__ . ': ' : '') . SUCCESS_ADDED_TO_CART_PRODUCTS, 'success');
+                        $this->notify('NOTIFIER_CART_OPTIONAL_SUCCESS_UPDATED_CART', $_POST);
                     } else {
                         if ($_GET['main_page'] != FILENAME_SHOPPING_CART) {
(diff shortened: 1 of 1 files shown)


drbyte:
Interesting idea.

Do you have some scenarios in mind?

I'm also wondering why `NOTIFIER_CART_ADD_CART_END` or `NOTIFIER_CART_ADD_CART_START` doesn't suffice ... other than specifically wanting to fire only after "all" the adds are done when a group has been added.

zenexpert:
I'm actually already using `NOTIFIER_CART_ADD_CART_END` but you read my mind - if you're adding multiple products from ie product_listing page, it might fire too early if condition is met before adding all products. 

As for scenarios - I'll PM you a link so you can see the idea and how it's being used, although I must admit it IS quite a specific setup... But in general, the idea would be to "force" a product (or any other page for that matter) if a certain product is added to cart.

drbyte:
I wonder if `actionAddProduct()` and `actionUpdateProduct()` should have similar hooks?  In the case of `update`, perhaps handling conditions that need to undo the things you added/forced via those others?  I suppose existing other notifiers might cover the update action already. But maybe actionAddProduct might need its own because without it it'd have to use the NOTIFER_CART_ADD_CART_END already discussed

zenexpert:
LOL, Doc, I need about 2 gallons of coffee and about 3 lifetimes to come up with a reply to THAT. The way I'm using it is reeeeeaaaaaly simplistic and I can't even imagine the scenarios you're probably already considering.

I'm guessing if you need to undo, then `update` could get tricky because of the `remove` when you remove multiple items from cart at once (checkboxes), but TBH, I wasn't really thinking that far...

drbyte:
No worries.
I've added the others I was thinking of.
And did a bit of renaming.

proseLA:
https://github.com/zencart/zencart/commit/b4764716724e6a441cb6b82c61084644257427e9#commitcomment-51775540

sorry, i'm not sure the proper place to post this....  are the `$_POST` vars necessary?

Issue #4370, opened 2021-06-09
Improve zen_get_show_product_switch


scottcwilson:
This function is called 14 times in includes/modules/pages/product_info/main_template_vars.php, and each call has 3 or 4 database queries.  We could check the database once and bring back all relevant SHOW_ values for the product, and store them in memory, and then build the relevant variables from the contents of memory (rather than repeatedly querying).  

Would you accept a PR to do this?

drbyte:
Okay. Eventually my goal is a product class that owns all of this; however, it's too early for that yet, so a micro refactoring could be an acceptable bandage.

In looking at it, it's possible we've already got the data in memory, so I wonder if the simplest refactor might be:

- the queries being run against `configuration` and `product_type_layout` tables in `zen_get_show_product_switch()` may be redundant, as the values are already in-memory as constants. So, testing them with `defined('foo')` and returning `constant('foo')` might suffice.
- the queries in `zen_get_show_product_switch_name()` can probably be avoided where already retrieved by adding something like this to the top of the function, leaving the rest of the function's logic as fallback:
```php
    global $product_info;
    if (!empty($product_info) && !$product_info->EOF && (int)$product_info->fields['products_id'] === (int)$lookup) {
        return strtoupper($prefix . $product_info->fields['type_handler'] . $suffix . $field_prefix . $field . $field_suffix);
    }
```

PR #4373, opened 2021-06-10
reduce queries required by zen_get_show_product_switch

scottcwilson:
Implementation of #4370 for product_info


proseLA:
big fan of this micro refactoring as well as the stated goal of a product class!

includes/classes/ShowData.php
```diff
@@ -0,0 +1,55 @@
...
+
+    public function __construct($lookup, $prefix = 'SHOW_', $suffix = '_INFO', $field_prefix = '_', $field_suffix = '') {
+      global $db; 
+
+      $this->prefix = $prefix; 
+      $this->suffix = $suffix; 
+      $this->field_prefix = $field_prefix; 
+      $this->field_suffix = $field_suffix; 
+
+      $sql = "SELECT products_type FROM " . TABLE_PRODUCTS . " WHERE products_id='" . $lookup . "'";
```
lat9, line 16: Shouldn't the `$lookup` value be cast to an integer, quotes removed from the query (since the products_id is an integer) and a `LIMIT 1` be added?

proseLA: @lat9 in my testing on your comment, the script:
```
includes/init_includes/init_sanitize.php
```
is already sanitizing the `$_GET['products_id']` var, and redirecting you if the products id is not valid.  so i would say casting it to an integer is not necessary.

as far as the `LIMIT 1`, i would also say not a good idea, and specifically because `products_id` is the primary key.

scottcwilson: All good ideas.  Currently looking for directional feedback from @drbyte though.

lat9: > 
> 
> @lat9 in my testing on your comment, the script:
> 
> ```
> includes/init_includes/init_sanitize.php
> ```
> 
> is already sanitizing the `$_GET['products_id']` var, and redirecting you if the products id is not valid. so i would say casting it to an integer is not necessary.

Yes, and that's one usage.  Since we're talking about a class-file, who knows what value could be specified by some other process.  Classes should be 'protectionist' about the inputs they're given.

> 
> as far as the `LIMIT 1`, i would also say not a good idea, and specifically because `products_id` is the primary key.

FWIW, I find the `LIMIT 1` to be descriptive in the reading of the code.

includes/classes/ShowData.php
```diff
@@ -0,0 +1,55 @@
+<?php 
+class ShowData extends base
```
zcwilt, line 2: Can we have a better class name here, that is more descriptive of the action we are doing

drbyte:
Using `LIMIT 1` is also part of the refactoring I mentioned already in v158.
Granted, it's implemented on the `$db->Execute($sql, 1)` statement instead of taking extra lines to say LIMIT 1 on new lines in the SQL strings.

drbyte:
PSR-12 (or at least PSR-2 plus relevant parts of PSR-12 that were never in-scope when PSR-2 was created) is certainly the right approach for new code.

includes/classes/ShowData.php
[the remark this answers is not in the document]
scottcwilson: I used lookup as an homage to the old code. :)

scottcwilson:
Added new listing page.  Will do all and featured pages once approved.

includes/templates/responsive_classic/templates/tpl_modules_products_new_listing.php
```diff
@@ -36,19 +51,19 @@
         $display_products_name = '';
       }
 
-      if (PRODUCT_NEW_LIST_MODEL != '0' and zen_get_show_product_switch($products_new->fields['products_id'], 'model')) {
+      if (PRODUCT_NEW_LIST_MODEL != '0' and $flag_show_model) {
```
proseLA, line 54: like the remainder here in the template, do we really need the var?  considering the var is only used once, can we replace the var with the `$config_switches->getSwitch('model')`?  and if we want to use a var, i think the flag_ prefix is not needed.  i would prefer `$show_model`.

includes/templates/responsive_classic/templates/tpl_modules_products_new_listing.php
```diff
@@ -18,7 +18,22 @@
...
+         if ($products_new->fields['products_type'] != $config_switches->getProductsType()) {
+          $config_switches = new ProductConfigurationSwitch($products_new->fields['products_id']);
+         } 
+      }
+      $flag_show_model = $config_switches->getSwitch('model');
+      $flag_show_weight = $config_switches->getSwitch('weight');
+      $flag_show_quantity = $config_switches->getSwitch('quantity');
+      $flag_show_date_added = $config_switches->getSwitch('date_added');
+      $flag_show_manufacturer = $config_switches->getSwitch('manufacturer');
+      $flag_show_show_free_shipping_image = $config_switches->getSwitch('ALWAYS_FREE_SHIPPING_IMAGE_SWITCH');
```
proseLA, line 36: is there an extra `_show` in here?  ie, `$flag_show_free_shipping_image` or ideally as commented below, `$show_free_shipping_image`.

includes/templates/responsive_classic/templates/tpl_modules_products_new_listing.php
```diff
@@ -36,19 +51,19 @@
...
       }
 
-      if (PRODUCT_NEW_LIST_MODEL != '0' and zen_get_show_product_switch($products_new->fields['products_id'], 'model')) {
+      if (PRODUCT_NEW_LIST_MODEL != '0' and $flag_show_model) {
         $display_products_model = '<b>' . TEXT_PRODUCT_MODEL . '</b>' . $products_new->fields['products_model'] . str_repeat('<br>', substr(PRODUCT_NEW_LIST_MODEL, 3, 1));
       } else {
         $display_products_model = '';
       }
 
-      if (PRODUCT_NEW_LIST_WEIGHT != '0' and zen_get_show_product_switch($products_new->fields['products_id'], 'weight')) {
```
proseLA, line 45: as has been previously stated by you, most people do not update their in-use template files.  considering we are looking are replacing this method, show we not add a soon to deprecated call into the `zen_get_show_product_switch` method?

proseLA (commented):
again, i am a fan of what this PR is doing.

includes/templates/responsive_classic/templates/tpl_modules_products_new_listing.php
scottcwilson: There is, thank you.  Fixed.

includes/templates/responsive_classic/templates/tpl_modules_products_new_listing.php
scottcwilson: I add changes like this to the Release Specific Upgrade Considerations page.  

https://docs.zen-cart.com/user/upgrading/release_specific_upgrade_considerations/

lat9:
@scottcwilson, would you resolve the conflicts for the merge?

includes/templates/responsive_classic/templates/tpl_modules_products_new_listing.php
```diff
@@ -36,19 +51,19 @@
...
+      if (PRODUCT_NEW_LIST_WEIGHT != '0' and $flag_show_weight) {
```
lat9, line 60: Should all these 'and' connectors be changed to `&&`?  I thought that we were trying to move away from `and` and `or` in preference to `&&` and `||`.

scottcwilson: Just trying to keep the scope of this change limited.  What you are suggesting is a good idea, I just didn't do it in this change.

PR #4383, opened 2021-06-15
refactor of order class

proseLA:
this started from looking at this here:

https://www.zen-cart.com/showthread.php?228112-Resend-customer-s-order-confirmation&p=1381502#post1381502

prior to putting any more energy into this refactor, i am curious about what the direction is of what i consider to be an important class?  i would prefer to not continue work if there is already development/plans for this class.

that said, there seems to be a lot of legacy code that could get cleaned up.  i see no need to use session information for sending of an order email confirmation, as the order is already in the database.  the same goes for the GLOBALS vars.

it would be nice if when we instantiate this class, we had all the information associated with the class. 

considering the popularity of plugins like EO, i can see the need for after editing an order, a store owner would want to resend an email confirmation with the new order information.

thoughts?

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -841,6 +841,7 @@ class order extends base
 
         for ($i = 0, $n = sizeof($this->products); $i < $n; $i++) {
             $this->products_ordered_attributes = '';
+            $attributes_exist = '0';
             if (isset($this->products[$i]['attributes'])) {
                 $custom_insertable_text = '';
                 $attributes_exist = '1';
@@ -1010,10 +1011,8 @@ class order extends base
             $this->notify('NOTIFY_ORDER_PROCESSING_ATTRIBUTES_BEGIN');
 
             //------ bof: insert customer-chosen options to order--------
-            $attributes_exist = '0';
             $this->products_ordered_attributes = '';
             if (isset($this->products[$i]['attributes'])) {
-                $attributes_exist = '1';
                 for ($j = 0, $n2 = sizeof($this->products[$i]['attributes']); $j < $n2; $j++) {
                     if (DOWNLOAD_ENABLED == 'true') {
                         $attributes_query = "SELECT popt.products_options_name, poval.products_options_values_name,
@@ -1247,7 +1246,9 @@ class order extends base
         $html_msg['ADDRESS_BILLING_DETAIL'] = zen_address_format($this->billing['format_id'], $this->billing, true, '', "<br>");
 
         $payment = '';
-        if (file_exists(DIR_WS_MODULES . 'payment/' . ($this->info['payment_module_code'] ?? 'NO_PAYMENT') . '.php')) {
+        if (isset($_SESSION['payment']) && is_object($GLOBALS[$_SESSION['payment']])) {
+            $payment = $GLOBALS[$_SESSION['payment']];
+        } elseif (file_exists(DIR_WS_MODULES . 'payment/' . ($this->info['payment_module_code'] ?? 'NO_PAYMENT') . '.php')) {
             require_once(DIR_WS_MODULES . 'payment/' . $this->info['payment_module_code'] . '.php');
             require_once(DIR_WS_LANGUAGES . $_SESSION['language'] . '/modules/payment/' . $this->info['payment_module_code'] . '.php');
             $payment = new $this->info['payment_module_code'];
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -60,6 +60,8 @@ class order extends base
 
         $this->orderId = $order_id = (int)$order_id;
 
+        $this->totals = [];
+
         $totals_query = "SELECT title, text, class, value
                          FROM " . TABLE_ORDERS_TOTAL . "
                          WHERE orders_id = " . (int)$this->orderId . "
@@ -1161,7 +1163,7 @@ class order extends base
             EMAIL_THANKS_FOR_SHOPPING . "\n" . EMAIL_DETAILS_FOLLOW . "\n" .
             EMAIL_SEPARATOR . "\n" .
             EMAIL_TEXT_ORDER_NUMBER . ' ' . $zf_insert_id . "\n" .
-            EMAIL_TEXT_DATE_ORDERED . ' ' . strftime(DATE_FORMAT_LONG) . "\n" .
+            EMAIL_TEXT_DATE_ORDERED . ' ' . strftime(DATE_FORMAT_LONG, strtotime($this->info['date_purchased'])) . "\n" .
             EMAIL_TEXT_INVOICE_URL . ' ' . zen_href_link(FILENAME_ACCOUNT_HISTORY_INFO, 'order_id=' . $zf_insert_id, 'SSL', false) . "\n\n";
 
         $html_msg['EMAIL_TEXT_HEADER'] = EMAIL_TEXT_HEADER;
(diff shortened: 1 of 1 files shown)


includes/classes/order.php
```diff
@@ -1201,18 +1228,19 @@ function send_order_email($zf_insert_id = null)
...
             EMAIL_SEPARATOR . "\n" .
-            zen_address_label($_SESSION['customer_id'], $_SESSION['billto'], false, '', "\n") . "\n\n";
+//            zen_address_label($_SESSION['customer_id'], $_SESSION['billto'], false, '', "\n") . "\n\n";
+            zen_address_format($this->billing['format_id'], $this->billing, false, '', "\n") . "\n\n";
         $html_msg['ADDRESS_BILLING_TITLE'] = EMAIL_TEXT_BILLING_ADDRESS;
-        $html_msg['ADDRESS_BILLING_DETAIL'] = zen_address_label($_SESSION['customer_id'], $_SESSION['billto'], true, '', "<br>");
+        $html_msg['ADDRESS_BILLING_DETAIL'] = zen_address_format($this->billing['format_id'], $this->billing, true, '', "<br>");
 
-        if (is_object($GLOBALS[$_SESSION['payment']])) {
+        if (is_object($GLOBALS[$_SESSION['payment'] ?? ''] ?? $this->info['payment_method'])) {
```
drbyte, line 1243: Is this tested? Seems to be a syntax error here with mismatched square-brackets.

drbyte: Oops. I was wrong.

scottcwilson:
Creating an order email that is identical to the original is not really possible at the current time, since the original may contain output from running the payment method. Downloads could also cause problems since these can be date restricted.

proseLA:
i briefly saw a comment, that now looks to be deleted, with regards to whether this code was tested.  i can assure you that prior to doing any PR, i test my code.  there may/probably/possibly will be bugs; but i do test prior to submitting.

i did place an order that included attributes using the v158 order class.  no attribute detail was listed in my original emails.  pretty clear it is currently not part of the html message.

scottcwilson:
> i briefly saw a comment, that now looks to be deleted ... 
Not me.

scottcwilson:
Note however that what I'm saying is the kind of thing that could be easily missed in testing simply by not using a payment module that adds content to the email or not using downloads.

proseLA:
> Note however that what I'm saying is the kind of thing that could be easily missed in testing simply by not using a payment module that adds content to the email or not using downloads.

the payment module code is addressed.  if you think it is not, please provide an example using the code where you think it is not.

the current code is confusing. we jump from `$GLOBALS[$_SESSION['payment']` to `$GLOBALS[$payment_class]` for no good reason.  and to be clear, i think the use of `GLOBALS` in the base code is not a good practice and we should remove it as possible/necessary.

i left the globals use for the auth_code and transaction_id as that data is only used for the admin copy of the email...  and i see no need to resend that.

with regards to downloads, i will take a look; but again this is something that an admin would do, and i see no reason why it should not work.  if a download is date restricted, and an admin wants to resend an order confirmation email; what exactly is the problem?  could you be more specific?

> not me.
yes, i think the email i received said @drbyte.

drbyte:
> i briefly saw a comment, that now looks to be deleted, with regards to whether this code was tested. 

> yes, i think the email i received said @drbyte.

Correct. I did ask a question about testing, because of an anomaly I thought I was seeing. 
Then upon further review I discovered that the issue I was concerned with was addressed properly.
So I posted another comment, saying "Oops, I was wrong", which should have also sent you an email.
Then I deleted both comments because neither were relevant and leaving them would only serve to muddy the discussion.

includes/classes/order.php
```diff
@@ -1085,25 +1125,7 @@ function create_add_products($zf_insert_id = null, $zf_mode = false)
...
-            $this->products_ordered .= $this->products[$i]['qty'] . ' x ' . $this->products[$i]['name'] . ($this->products[$i]['model'] != '' ? ' (' . $this->products[$i]['model'] . ') ' : '') . ' = ' .
-                $currencies->display_price($this->products[$i]['final_price'], $this->products[$i]['tax'], $this->products[$i]['qty']) .
-                ($this->products[$i]['onetime_charges'] != 0 ? "\n" . TEXT_ONETIME_CHARGES_EMAIL . $currencies->display_price($this->products[$i]['onetime_charges'], $this->products[$i]['tax'], 1) : '') .
-                $this->products_ordered_attributes . "\n";
-            $this->products_ordered_html .=
-                '<tr>' . "\n" .
-                '<td class="product-details" align="right" valign="top" width="30">' . $this->products[$i]['qty'] . '&nbsp;x</td>' . "\n" .
-                '<td class="product-details" valign="top">' . nl2br($this->products[$i]['name']) . ($this->products[$i]['model'] != '' ? ' (' . nl2br($this->products[$i]['model']) . ') ' : '') . "\n" .
-                '<nobr>' .
-                '<small><em> ' . nl2br($this->products_ordered_attributes) . '</em></small>' .
```
drbyte, line 1098: The extraction of this section into its own function needs to also include a way for the notifier logic from the previous 30 lines, else notifiers that are firing to inject attribute-specific information (such as phone cards, software license keys, PINs, etc) won't work.

scottcwilson:
I don't think it's going to be possible to create a high fidelity reproduction of the order confirmation email, because some things might be happening when an order is placed that are one-time-only (dispensing a license key from a limited set), some conditions change over time (the price of a precious metal or currency), etc.  You can do a best-effort reproduction (and I did in the Preview Email plugin - https://www.zen-cart.com/downloads.php?do=file&id=2220) but it won't be identical to what was sent out when the order was placed in all cases.

includes/classes/order.php
proseLA: i will take a look...

proseLA:
> I don't think it's going to be possible to create a high fidelity reproduction of the order confirmation email, because some things might be happening when an order is placed that are one-time-only (dispensing a license key from a limited set), some conditions change over time (the price of a precious metal or currency), etc. You can do a best-effort reproduction (and I did in the Preview Email plugin - https://www.zen-cart.com/downloads.php?do=file&id=2220) but it won't be identical to what was sent out when the order was placed in all cases.

with all due respect, this PR is a refactor of the order class (and so named).  if you do not think this class needs refactoring, then we can just agree to disagree.  as 1 example:
```
    $orderNo = 17554;  // use any valid order number
    $order = new order($orderNo);
    $order->query($orderNo);
    print_r($order->totals);
```

with regards to conditions changing over time, i have no idea what you are talking about.  an order is placed; the data goes into the db; any data that is real time data and is part of an email should be so labeled.  but it is not part of the order class, which IMO needs work to be clearer and without bugs.

includes/classes/order.php
```diff
@@ -1201,34 +1234,42 @@ function send_order_email($zf_insert_id = null)
...
             $cc_num_display = (isset($this->info['cc_number']) && $this->info['cc_number'] != '') ? /*substr($this->info['cc_number'], 0, 4) . */
                 str_repeat('X', (strlen($this->info['cc_number']) - 8)) . substr($this->info['cc_number'], -4) . "\n\n" : '';
             $email_order .= EMAIL_TEXT_PAYMENT_METHOD . "\n" .
                 EMAIL_SEPARATOR . "\n";
-            $payment_class = $_SESSION['payment'];
-            $email_order .= $GLOBALS[$payment_class]->title . "\n\n";
+            $email_order .= $payment->title . "\n\n";
             $email_order .= (isset($this->info['cc_type']) && $this->info['cc_type'] != '') ? $this->info['cc_type'] . ' ' . $cc_num_display . "\n\n" : '';
-            $email_order .= (isset($GLOBALS[$payment_class]->email_footer) && $GLOBALS[$payment_class]->email_footer) ? $GLOBALS[$payment_class]->email_footer . "\n\n" : '';
+            $email_order .= (isset($payment->email_footer) && $payment->email_footer) ? $payment->email_footer . "\n\n" : '';
```
drbyte, line 1263: When this `$payment = new $this->info['payment_module_code']` is run, it creates a fresh new instance of the payment module, not the one that is currently active in the session.
Thus, it is likely that if the payment module has set a customized value (during current session processing) for the `email_footer` then that info will be lost here.

drbyte:
Current review status:
- I have concerns about creating a new instance, as mentioned above
- You've made some updates to the email-content generation sub-function, at my request, but I haven't reviewed those yet.

includes/classes/order.php
proseLA: i think my new commit will handle this concern.

scottcwilson:

PR #4385, opened 2021-06-16
Truncate Paypal/Payflow note length to max acceptable values

robryanx:
Paypal NVP NOTE has a max length of 255 characters
Payflow COMMENT2 has a max length of 128 characters


scottcwilson:
Thanks Rob!

PR #4390, opened 2021-06-20
Render line-breaks in Admin-settings-notifications HTML messages

retched:
This is a bit nitpicky, and I'll understand if this is rejected. Just makes it so that an installation (using default language files with `\n` newlines) send the HTML version of the notification message with actual `<br>` for line breaks in it. (Most email programs don't immediately interpret `\n` newlines as `<br>`. At times they'll just leave it as one block.)


retched:
Found the two problems. One changes the "Settings Added" and the other changes the "Settings Removed".

scottcwilson:
Thanks for your contribution, @retched !

drbyte:
Thanks @retched !

PR #4392, opened 2021-06-23
Fix coupon min. required calculation wrong 

scottcwilson:
when Calculated From = "Based on Full Order Total of All Products"

Without this fix you'll get an error message saying a total of <$X> is required to use this coupon, as though the Calculated From setting were "Based on Total of Allowed Products according to Coupon Restriction Rules".


proseLA:
is this something you want backported?  did you find the same problem in v158?

scottcwilson:
This is a 1.5.7 branch fix.  It works properly in 1.5.8.

PR #4397, opened 2021-06-25
Incorrect use of text field from orders_total

### LINKED ISSUE #4097: ZC158, class Customer: PHP Notice: A non well formed numeric value encountered, for decimal comma
My currency settings are with decimal as comma and separator as decimal point. Don't ask.

So, set the currency as described and make an order.

On the admin customer page, debugs are generated as the text version of the order total extracted from table order_totals text column has a comma in it.

> PHP Notice: A non well formed numeric value encountered in 158\includes\classes\Customer.php on line 310:

https://github.com/zencart/zencart/blob/807ce81b5d96194d701046c823945490e9e5ae9a/includes/classes/Customer.php#L298-L314

and the strings get multiplied with the commas in them.

and also PHP Notice: A non well formed numeric value encountered in 158\includes\classes\currencies.php on line 97.
https://github.com/zencart/zencart/blob/807ce81b5d96194d701046c823945490e9e5ae9a/includes/classes/currencies.php#L95-L98

scottcwilson:
Fixes #4097

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -310,7 +310,6 @@ class Customer extends base
                     'language_code' => $result['language_code'],
                 ];
             }
-            $result['order_total'] = preg_replace('/[^\d.,-]/', '', $result['order_total']);
             $lifetime_value += ($result['order_total'] * $result['currency_value']);
         }
         $this->data['last_order'] = $last_order;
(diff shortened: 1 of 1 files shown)


proseLA:
you might want to double check but i think this line is no longer necessary as well:

https://github.com/zencart/zencart/blob/190bda0360fe1b3ad64c00fbcbd2da9404012884/includes/classes/Customer.php#L313

PR #4407, opened 2021-07-06
Update for PHP 8.0, callback related

mc12345678:
Refactored sanitizeEmail to eliminate a grouped if statement, providing an early escape for processing.

Callback function receives data by value not by reference. In PHP 8.0, this throws a php warning: `Warning:  configurationValidation::sanitizeEmail(): Argument #1 ($val) must be passed by reference, value given in [...][...]`. Result is that any other usage of these functions to sanitize, must consider the returned value or `$GLOBALS['configuration_value']` as modified within these functions instead of "reusing" the variable that was sent.

Note that sanitizeBoolean permits the value to be "representative" of a boolean value.  As a result, the value returned is true (that value is considered of the boolean type or false if it is not of the boolean type (returned from two points, the second of which is a fallback check)), but `$GLOBALS['configuration_value']` may be other than strictly `true` or strictly `false`, where `$GLOBALS['configuration_value']` will represent the actual value that was being checked/sanitized to be boolean.

Also, in testing the revision to this, discovered that there was no clear way to store an empty email address. That issue is/was also addressed in this branch.

### DIFF
--- a/admin/includes/classes/configurationValidation.php
+++ b/admin/includes/classes/configurationValidation.php
@@ -31,55 +31,57 @@ class configurationValidation extends base
                          'flags' => '',
                          );
         
-        if (isset($val)) {
-            $send_to_array = explode(",", $val);
-            // If count($send_to_array) > 1 then there are multiple addresses to be parsed.
-            foreach ($send_to_array as $key => $address) {
-                $send_to_name = '';
-                $send_to_email = trim($address);
-                // Collect the portion within <> symbols
-                preg_match('/\<[^>]+\>/', $address, $send_email_array);
-                // If there are parts to the above, then set/collect them.
-                if (!empty($send_email_array)) {
-                    $send_to_email = preg_replace ("/>/", "", $send_email_array[0]);
-                    $send_to_email = trim(preg_replace("/</", "", $send_to_email));
-                    $send_to_name  = trim(preg_replace('/\<[^*]*/', '', $address));
-                }
-                
-                // Collect the individual name/email as part of an array.
-                $results[$key]['send_to_name'] = filter_var($send_to_name, FILTER_SANITIZE_STRING, $options);
-                $results[$key]['send_to_email'] = filter_var($send_to_email, FILTER_VALIDATE_EMAIL, $options);
-                
-                // Restore the inner email address back to its state for capture.
-                if (!empty($send_email_array) && $results[$key]['send_to_email'] !== false) {
-                    $results[$key]['send_to_email'] = '<' . $results[$key]['send_to_email'] . '>';
-                }
-                
-                // If the email address is not assigned, but there is content in the name, validate that the name is a correct email address.
-                if ($results[$key]['send_to_email'] === false && !empty($results[$key]['send_to_name'])) {
-                    $results[$key]['send_to_name'] = filter_var($results[$key]['send_to_name'], FILTER_VALIDATE_EMAIL, $options);
-                }
-                
-                // Remove array parameters that have failed validation.
-                foreach ($results[$key] as $key2 => $value2) {
-                    if (empty($value2)) {
-                        unset($results[$key][$key2]);
-                    }
-                }
-                unset($key2, $value2);
-                
-                // If this round of review identified no record, then move to the next record.
-                if (empty($results[$key])) {
-                    continue;
-                }
-                
-                // Collect the filtered email information into a single record.
-                $response[$key] = implode(" ", $results[$key]);
+        if (!isset($val)) {
+            return false;
+        }
+        
+        $send_to_array = explode(",", $val);
+        // If count($send_to_array) > 1 then there are multiple addresses to be parsed.
+        foreach ($send_to_array as $key => $address) {
+            $send_to_name = '';
+            $send_to_email = trim($address);
+            // Collect the portion within <> symbols
+            preg_match('/\<[^>]+\>/', $address, $send_email_array);
+            // If there are parts to the above, then set/collect them.
+            if (!empty($send_email_array)) {
+                $send_to_email = preg_replace ("/>/", "", $send_email_array[0]);
+                $send_to_email = trim(preg_replace("/</", "", $send_to_email));
+                $send_to_name  = trim(preg_replace('/\<[^*]*/', '', $address));
             }
             
-            // Collect email addresses entered as a string.
-            $final_result = implode(", ", $response);
+            // Collect the individual name/email as part of an array.
+            $results[$key]['send_to_name'] = filter_var($send_to_name, FILTER_SANITIZE_STRING, $options);
+            $results[$key]['send_to_email'] = filter_var($send_to_email, FILTER_VALIDATE_EMAIL, $options);
+            
+            // Restore the inner email address back to its state for capture.
+            if (!empty($send_email_array) && $results[$key]['send_to_email'] !== false) {
+                $results[$key]['send_to_email'] = '<' . $results[$key]['send_to_email'] . '>';
+            }
+            
+            // If the email address is not assigned, but there is content in the name, validate that the name is a correct email address.
+            if ($results[$key]['send_to_email'] === false && !empty($results[$key]['send_to_name'])) {
+                $results[$key]['send_to_name'] = filter_var($results[$key]['send_to_name'], FILTER_VALIDATE_EMAIL, $options);
+            }
+            
+            // Remove array parameters that have failed validation.
+            foreach ($results[$key] as $key2 => $value2) {
+                if (empty($value2)) {
+                    unset($results[$key][$key2]);
+                }
+            }
+            unset($key2, $value2);
+            
+            // If this round of review identified no record, then move to the next record.
+            if (empty($results[$key])) {
+                continue;
+            }
+            
+            // Collect the filtered email information into a single record.
+            $response[$key] = implode(" ", $results[$key]);
         }
+        
+        // Collect email addresses entered as a string.
+        $final_result = implode(", ", $response);
         
         // If there are no email addresses that are valid, then identify that failed validation.
         if (empty($final_result) && zen_not_null($val)) {
--- a/admin/includes/classes/configurationValidation.php
+++ b/admin/includes/classes/configurationValidation.php
@@ -16,6 +16,11 @@ if (!defined('IS_ADMIN_FLAG')) {
 
 class configurationValidation extends base
 {
+    /**
+     *  Usage setting val_function for the configuration key to something similar
+     *    to the below will call on this code to support storage of the email related information.
+     *    val_function = '{"error":"TEXT_EMAIL_ADDRESS_VALIDATE","id":"FILTER_CALLBACK","options":{"options":["configurationValidation","sanitizeEmail"]}}'
+     **/
     static public function sanitizeEmail($val) {
         $results = array();
         $send_email_array = array();
@@ -99,7 +104,7 @@ class configurationValidation extends base
     
     
     /**
-     *  Usage setting val_function  for the configuration key to something similar
+     *  Usage setting val_function for the configuration key to something similar
      *    to the below will call on this code to support storage of the boolean related value.
      *    val_function = '{"error":"TEXT_BOOLEAN_VALIDATE","id":"FILTER_CALLBACK","options":{"options":["configurationValidation","sanitizeBoolean"]}}'
      **/
--- a/admin/includes/classes/configurationValidation.php
+++ b/admin/includes/classes/configurationValidation.php
@@ -82,7 +82,7 @@ class configurationValidation extends base
         }
         
         // If there are no email addresses that are valid, then identify that failed validation.
-        if (empty($final_result)) {
+        if (empty($final_result) && zen_not_null($val)) {
             return false;
         }
         
--- a/admin/includes/classes/configurationValidation.php
+++ b/admin/includes/classes/configurationValidation.php
@@ -16,7 +16,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 
 class configurationValidation extends base
 {
-    static public function sanitizeEmail(&$val) {
+    static public function sanitizeEmail($val) {
         $results = array();
         $send_email_array = array();
         $send_to_array = array();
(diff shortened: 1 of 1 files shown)


mc12345678:
Broke this up into more discreet commits. Can recombine as seen/where fit. Note that the original comment above did not include discussion of supporting an empty email address.  That issue was discovered during additional testing. I am going to add/update the comment, but wanted to be sure to point it out in the course of time.

scottcwilson:
@mc12345678 what's the use case for storing an empty email address?
Also would you please link to the PHP documentation on the change that is producing the warning?

lat9:
Sorry for the late response.  @mc12345678, please provide a full debug log that includes the backtrace of function calls that resulted in this warning.

mc12345678:
Surprised I didn't include it originally... Must have finally given up on trying to resolve it and just wanted to get a solution out...

````php
[DATE TIME TIMEZONE] Request URI: /admin/index.php?cmd=configuration&gID=12&cID=272&action=save, IP address: IPADDRESS
#1  filter_var() called at [/serverPath/admin/includes/functions/configuration_checks.php:53]
#2  zen_validate_configuration_entry() called at [/serverPath/admin/configuration.php:22]
#3  require(/serverPath/admin/configuration.php) called at [/serverPath/admin/index.php:11]
--> PHP Warning: configurationValidation::sanitizeEmail(): Argument #1 ($val) must be passed by reference, value given in /serverPath/admin/includes/functions/configuration_checks.php on line 53.
````

Line numbers might be a little off from what the files previously had/held, I had to go back to find this particular log file associated to when I suggested the change. I tried to find other ways to implement/correct the issue, but didn't find any other than as suggested.

**Comment by lat9** [comment]
Thanks for that @mc12345678.  I'll note that the `cID` referenced in the log currently identifies (???) the enablement of email archiving.  I ran through that change with no error and also changed the Email address (displayed for Contact Us) with no logs being generated.

That was using the current (20220308 10am EST) repository with a fresh install.  Could you re-validate to confirm that the issue has mysteriously corrected itself?

**Comment by mc12345678** [comment]
@lat9, glad to reattempt. For confirmation, what php version was used for your testing?

**Comment by mc12345678** [comment]
At point of occurrence, (had) encountered with:
````php
admin/index.php?cmd=configuration&gID=12&cID=269&action=save
admin/index.php?cmd=configuration&gID=12&cID=272&action=save
admin/index.php?cmd=configuration&gID=12&cID=274&action=save
````
Representing, respectively:
````php
Email Address (sent FROM)
Send Copy of Order Confirmation Emails To
Send Copy of Create Account Emails To
````
All attempts included attempting storing with a value in the field, though also recognize that in most cases disabling the associated field functions/should function as if the field is empty.

When collecting the call data just before the "problematic" operation, I had the following (email address was fully valid but is modified here):
````php
--> PHP Warning: variable:MYEMAIL[AT)MYDOMAIN(dot>EXTENSION:id:1024:options:Array
(
    [options] => Array
        (
            [0] => configurationValidation
            [1] => sanitizeEmail
        )

)
````
The above, if no data populated each variable, would be: `variable::id::options` to ensure recognition of an included space or other character.

Have not reperformed an install/rebuild to test/retest further,

**Comment by lat9** [comment]
If I'm deciphering your previous post correctly, are you saying that the issue arises when you attempt to save an empty string as an email address?

**Comment by lat9** [comment]
I'm thinking that the underlying issue is that there are some email addresses, e.g. **Email Address (sent FROM)**, that are required, but there are some, e.g. **Send Copy of Create Account Emails To**, that are unfortunately _conditionally_ required.  That is only required to be valid if the associated "Send Copies of [email type]" is set to 1.

**Comment by scottcwilson** [comment]
Good catch @lat9 - hopefully the fix for this will solve #4671 as well.

**Comment by proseLA** [comment]
> I'm thinking that the underlying issue is that there are some email addresses, e.g. **Email Address (sent FROM)**, that are required, but there are some, e.g. **Send Copy of Create Account Emails To**, that are unfortunately _conditionally_ required. That is only required to be valid if the associated "Send Copies of [email type]" is set to 1.

while this may be true, i am personally not a fan of using hard coding for a config value based on the value of another config value.

it just seems that the flag to turn it on/off is superfluous.  if there is a valid email address there send it; and lets get rid of the flag.

i think changing the function to a new function as documented by me in #4671 is an easy solution that solves the blank email problem for those emails that are not required.

similar to lat9 i have a hard time deciphering what mc is saying.

**Comment by lat9** [comment]
> > I'm thinking that the underlying issue is that there are some email addresses, e.g. **Email Address (sent FROM)**, that are required, but there are some, e.g. **Send Copy of Create Account Emails To**, that are unfortunately _conditionally_ required. That is only required to be valid if the associated "Send Copies of [email type]" is set to 1.
> 
> while this may be true, i am personally not a fan of using hard coding for a config value based on the value of another config value.
> 
> it just seems that the flag to turn it on/off is superfluous. if there is a valid email address there send it; and lets get rid of the flag.
> 
> i think changing the function to a new function as documented by me in #4671 is an easy solution that solves the blank email problem for those emails that are not required.
> 
> similar to lat9 i have a hard time deciphering what mc is saying.

That was the direction my head was (er) heading as well; if the conditional email is empty, the associated copy-to emails aren't set ... and the associated on/off flag isn't needed.  That in conjunction with the new email validation that allows an empty value.

**Comment by mc12345678** [comment]
> If I'm deciphering your previous post correctly, are you saying that the issue arises when you attempt to save an empty string as an email address?

Not at all.... wow, no idea how the provided data got so confused. 

Data that was present at line 53 of admin/includes/functions/configurations_checks.php that was modified for this thread to not post my own email address. Modified to limit the amount of spam that might be generated by scraping or other techniques of email address collection was: 
````php
--> PHP Warning: variable:MYEMAIL[AT)MYDOMAIN(dot>EXTENSION:id:1024:options:Array
(
    [options] => Array
        (
            [0] => configurationValidation
            [1] => sanitizeEmail
        )

)
````
That data in a php 8.0 environment using the Zen Cart version 1.5.8 software available on the date this thread was initiated caused the reported problem and was corrected by the proposed pull request that removed the use of the variable reference (`&$variable` going to just `$variable`).  Alternate approaches were attempted but unsuccessful. I didn't document those approaches and at the moment don't recall the five or more ways I tried to resolve the issue identified. 

Further, when developing the original check, the issue of not being able to remove the content was seen as a nonissue for the large majority of the field specifically because there is and was a way to directly disable sending an email to the copy to group by turning it off. If there was concern about someone setting the address in the database even with it turned off, it could still be modified to some fictitious "valid" email. The software does not confirm the existence of the email only that it is in the "proper" format. My modified example pasted above does not meet the proper format, but again that pasted above was modified away from a true, accurate, and functional email address. 

The discussion of format for blank information was to explain and/or help someone that wanted to evaluate the output of my debug call function: 
````php
trigger_error('variable:' . print_r($variable, true) . ':id:' . print_r($id, true) . ':options:' . print_r($options, true), E_USER_WARNING);
````
That's it.

**Comment by mc12345678** [comment]
> > > I'm thinking that the underlying issue is that there are some email addresses, e.g. **Email Address (sent FROM)**, that are required, but there are some, e.g. **Send Copy of Create Account Emails To**, that are unfortunately _conditionally_ required. That is only required to be valid if the associated "Send Copies of [email type]" is set to 1.
> > 
> > 
> > while this may be true, i am personally not a fan of using hard coding for a config value based on the value of another config value.
> > it just seems that the flag to turn it on/off is superfluous. if there is a valid email address there send it; and lets get rid of the flag.
> > i think changing the function to a new function as documented by me in #4671 is an easy solution that solves the blank email problem for those emails that are not required.
> > similar to lat9 i have a hard time deciphering what mc is saying.
> 
> That was the direction my head was (er) heading as well; if the conditional email is empty, the associated copy-to emails aren't set ... and the associated on/off flag isn't needed. That in conjunction with the new email validation that allows an empty value.

But in the code of there is a condition meet where an email should not be sent, at that point it's not possible to unset the define (at least as long as the database data remains hard in the operation of the store). As such the flag to not send the email may also be adjusted more logically (even though that setting also is hard coded from a database defined value). This also makes more sense when attempting to prevent an email being sent by indicating "don't send an email" instead of "send an email to no one". Though it does appear that to take advantage of that "change" in setting (do send to don't send) would require some refactoring in the code to permit an override opportunity.

**Comment by lat9** [comment]
Have you retested with the current v158 code, especially since I can't understand what steps are required to recreate the initially-reported issue?  There have been a slew of changes over the past 9 months.

**Comment by mc12345678** [comment]
> At point of occurrence, (had) encountered with:
> 
> ```
> admin/index.php?cmd=configuration&gID=12&cID=269&action=save
> admin/index.php?cmd=configuration&gID=12&cID=272&action=save
> admin/index.php?cmd=configuration&gID=12&cID=274&action=save
> ```
> 
> Representing, respectively:
> 
> ```
> Email Address (sent FROM)
> Send Copy of Order Confirmation Emails To
> Send Copy of Create Account Emails To
> ```
> 
> All attempts included attempting storing with a value in the field, though also recognize that in most cases disabling the associated field functions/should function as if the field is empty.
> 
> When collecting the call data just before the "problematic" operation, I had the following (email address was fully valid but is modified here):
> 
> ```
> --> PHP Warning: variable:MYEMAIL[AT)MYDOMAIN(dot>EXTENSION:id:1024:options:Array
> (
>     [options] => Array
>         (
>             [0] => configurationValidation
>             [1] => sanitizeEmail
>         )
> 
> )
> ```
> 
> The above, if no data populated each variable, would be: `variable::id::options` to ensure recognition of an included space or other character.
> 
> Have not reperformed an install/rebuild to test/retest further,



> Have you retested with the current v158 code, especially since I can't understand what steps are required to recreate the initially-reported issue? There have been a slew of changes over the past 9 months.

> At point of occurrence, (had) encountered with:
> 
> ```
> admin/index.php?cmd=configuration&gID=12&cID=269&action=save
> admin/index.php?cmd=configuration&gID=12&cID=272&action=save
> admin/index.php?cmd=configuration&gID=12&cID=274&action=save
> ```
> 
> Representing, respectively:
> 
> ```
> Email Address (sent FROM)
> Send Copy of Order Confirmation Emails To
> Send Copy of Create Account Emails To
> ```
> 
> All attempts included attempting storing with a value in the field, though also recognize that in most cases disabling the associated field functions/should function as if the field is empty.
> 
> When collecting the call data just before the "problematic" operation, I had the following (email address was fully valid but is modified here):
> 
> ```
> --> PHP Warning: variable:MYEMAIL[AT)MYDOMAIN(dot>EXTENSION:id:1024:options:Array
> (
>     [options] => Array
>         (
>             [0] => configurationValidation
>             [1] => sanitizeEmail
>         )
> 
> )
> ```
> 
> The above, if no data populated each variable, would be: `variable::id::options` to ensure recognition of an included space or other character.
> 
> ### *****Have not reperformed an install/rebuild to test/retest further,***** ###

**Comment by lat9** [comment]
I believe that the discussion in #4671 expands on and covers this report.

**Comment by mc12345678** [comment]
> I believe that the discussion in #4671 expands on and covers this report.

Negative, that issue relates to entering a blank email address. This issue relates to entering a valid email address. Unrelated.

**Comment by proseLA** [comment]
at some point, perhaps brevity and conciseness will be attained.

v158, php 8.1 inserting:
```
fr <larry@mydomain.com>, mx <prose@mydomain.com>
```
for email config value generates said log.

```
[10-Mar-2022 14:34:21 America/Los_Angeles] Request URI: /admin/index.php?cmd=configuration&gID=12&cID=271&action=save, IP address: 192.168.14.74
#1 /var/www/zcdev/admin/includes/functions/configuration_checks.php(53): filter_var()
#2 /var/www/zcdev/admin/configuration.php(22): zen_validate_configuration_entry()
#3 /var/www/zcdev/admin/index.php(11): require('...')
--> PHP Warning: configurationValidation::sanitizeEmail(): Argument #1 ($val) must be passed by reference, value given in /var/www/zcdev/admin/includes/functions/configuration_checks.php on line 53.
```

bug confirmed.

mc12345678:
> at some point, perhaps brevity and conciseness will be attained.
> 
> v158, php 8.1 inserting:
> 
> ```
> fr <larry@mydomain.com>, mx <prose@mydomain.com>
> ```
> 
> for email config value generates said log.
> 
> ```
> [10-Mar-2022 14:34:21 America/Los_Angeles] Request URI: /admin/index.php?cmd=configuration&gID=12&cID=271&action=save, IP address: 192.168.14.74
> #1 /var/www/zcdev/admin/includes/functions/configuration_checks.php(53): filter_var()
> #2 /var/www/zcdev/admin/configuration.php(22): zen_validate_configuration_entry()
> #3 /var/www/zcdev/admin/index.php(11): require('...')
> --> PHP Warning: configurationValidation::sanitizeEmail(): Argument #1 ($val) must be passed by reference, value given in /var/www/zcdev/admin/includes/functions/configuration_checks.php on line 53.
> ```
> 
> bug confirmed.
> 
> and can it not be fixed by changing this:
> 
> https://github.com/zencart/zencart/blob/23066d01cc9b5c6922eebfa206a087ad3c8e5bbb/admin/includes/classes/configurationValidation.php#L19
> 
> to:
> 
> ```
> static public function sanitizeEmail($val) {
> ```
> 
> or is this to simple?

Sigh.  Please see [https://github.com/zencart/zencart/pull/4407/commits/780b52cb268e46fad99a76dd971abcb2b76de9a7](this commit), which was part of this pull request.

Or was this pr too complex?

proseLA:
i did in fact miss that.  and truly, i did not even notice that this was a PR.  i was responding to @lat9 not being able to reproduce this error, which i was able to do without problem.

but if i was doing a code review, i would move all of your array declarations to using the short syntax `[]` which thankfully seems to be more of the norm; i would add php documentation for entry and return parms (somthing i am guilty of not doing enough of), and i find your return statement to be unclear:

https://github.com/zencart/zencart/blob/23066d01cc9b5c6922eebfa206a087ad3c8e5bbb/admin/includes/classes/configurationValidation.php#L93

i am not a fan of assigning a var in a return statement.  i would separate them out into 2.

i stand by my brevity and concise statement.

lat9:
Why was the pass-by-reference removed from both sanitizeEmail and sanitizeBoolean?  From my review of the various `sanitize` methods, it appears that the intent was to pass-by-reference so that a found-to-be-invalid value can be modified to be valid.

proseLA:
at the risk of alienating sorely needed talented developers on this project, i would be hesitant to merge anything from developers that do not take ownership of their code already merged.

i am referring to #4531 

i vaguely remember a comment that said developer would only address bugs in released code as opposed to merged code.  but alas, that comment was on an already merged PR and despite a polite request to open a new issue, said issue was never opened; and alas, my searching skills for closed PRs on github were not competent enough to find said comment.

again, this is only my opinion.

lat9:
OK, those changes appear to correct the validation issues for the _optional_ email addresses, i.e. you can now set them to be an empty string.  Unfortunately, that update also allows an empty email-address to be set for `STORE_OWNER_EMAIL_ADDRESS` and `EMAIL_FROM`, which is not acceptable.

I'm thinking that a separate validation is needed for the email addresses that (a) need to be set and (b) need to be in a simply joe@example.com fomat.

Issue #4399, opened 2021-06-27
ZC158: malformed url creates debug


torvista:
using this url
https://DOMAIN/index.php?main_page=product_info

is handled ok on the shopfront "product not found", but creates a debug.
I know it is malformed but these spiders are sloppy.

> [27-Jun-2021 16:28:55 Europe/Madrid] Request URI: /zencart-158/index.php?main_page=product_info, IP address: 127.0.0.1
> #1  zen_get_products_base_price() called at [...\zencart-158\includes\modules\products_quantity_discounts.php:62]
> #2  require(...\zencart-158\includes\modules\products_quantity_discounts.php) called at [...\zencart-158\includes\modules\pages\product_info\main_template_vars.php:148]
> #3  require(...\zencart-158\includes\modules\pages\product_info\main_template_vars.php) called at [...\zencart-158\includes\templates\responsive_classic\common\tpl_main_page.php:178]
> #4  require(...\zencart-158\includes\templates\responsive_classic\common\tpl_main_page.php) called at [...\zencart-158\index.php:94]
> --> PHP Warning: Undefined array key "products_price" in ...\zencart-158\includes\functions\functions_prices.php on line 421.
> 
> [27-Jun-2021 16:28:55 Europe/Madrid] Request URI: /zencart-158/index.php?main_page=product_info, IP address: 127.0.0.1
> #1  zen_get_products_base_price() called at [...\zencart-158\includes\modules\products_quantity_discounts.php:62]
> #2  require(...\zencart-158\includes\modules\products_quantity_discounts.php) called at [...\zencart-158\includes\modules\pages\product_info\main_template_vars.php:148]
> #3  require(...\zencart-158\includes\modules\pages\product_info\main_template_vars.php) called at [...\zencart-158\includes\templates\responsive_classic\common\tpl_main_page.php:178]
> #4  require(...\zencart-158\includes\templates\responsive_classic\common\tpl_main_page.php) called at [...\zencart-158\index.php:94]
> --> PHP Warning: Undefined array key "products_priced_by_attribute" in ...\zencart-158\includes\functions\functions_prices.php on line 423.
> 
> [27-Jun-2021 16:28:55 Europe/Madrid] Request URI: /zencart-158/index.php?main_page=product_info, IP address: 127.0.0.1
> #1  zen_get_products_base_price() called at [...\zencart-158\includes\modules\products_quantity_discounts.php:86]
[...]

PR #4408, opened 2021-07-09
address php notices in Product info

proseLA:
#4399 

2nd commit for attempt at reformatting based on 'our' code style...

### DIFF
--- a/includes/modules/pages/product_info/main_template_vars.php
+++ b/includes/modules/pages/product_info/main_template_vars.php
@@ -1,158 +1,161 @@
 <?php
-/**
- *  product_info main_template_vars.php
- *
- * @copyright Copyright 2003-2020 Zen Cart Development Team
- * @copyright Portions Copyright 2003 osCommerce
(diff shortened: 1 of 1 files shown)


includes/modules/pages/product_info/main_template_vars.php
```diff
@@ -1,158 +1,161 @@
...
+            while ($file = $dir->read()) {
+                if (!is_dir($extras_dir . '/' . $file)) {
+                    if (preg_match('~^[^\._].*\.php$~i', $file) > 0) {
+                        $directory_array[] = '/' . $file;
+                    }
+                }
+            }
+            $dir->close();
+        }
+        if (sizeof($directory_array)) {
```
mc12345678, line 126: If the way "we" do it, this should use `count` instead of `sizeof`. Still not entirely sure why most/all are going this route other than `sizeof` is an alias to `count` and some previous "scare" that the alias was going to be removed.

PR #4411, opened 2021-07-13
Change 'installer' to 'Installer'

nickcollie:
I think that this change is required as well as the typo that has already been changed by Dr Byte
e.g.
/includes/classes/PluginSupport/Installer.php:        if (!file_exists($pluginDir . '/Installer/' . $patchFile)) {


scottcwilson:
Thank you Nick!

PR #4412, opened 2021-07-13
Remove sample images from default distribution

scottcwilson:


barco57:
Why? Did I miss something?

torvista:
I see this only installs the images when the demo products are installed...that seems correct to me.

scottcwilson:
> Why? Did I miss something?

They are not needed for real stores.

PR #4414, opened 2021-07-15
product, collect_info: use empty on observer-returned optional values

torvista:
When using this observer and constructing the array to be returned, I think for code clarity it would be prudent to include all the fields in the construction for example
`$p2[] = ['label' => ['text' => $label_text, 'addl_class' => $label_addl_class, 'parms' => $label_parms, 'field_name' => $label_field_name], 'input' => $input];`

In this case unused parameters would create an additional space if isset is used instead of empty.


mc12345678:
While can't disagree that the use of `!empty` instead of `isset` offers more flexibility, the above "construction" of all of the fields would not specifically cause a problem with the use of `isset`. E.g. if the associated variable for example `$label_addl_class === null` then `isset($var['label']['addl_class']) === false`. But if in that scenario `$label_addl_class === ''` then yes it would (have) caused the additional space because `isset()` of that empty quote variable would evaluate to true.

torvista:
I don't think having to use null instead of '' is so difficult.

PR #4418, opened 2021-07-19
Configuration table reads should never be cached

scottcwilson:
You wouldn't want a cached copy of DOWN_FOR_MAINTENANCE, for example.  The entries in these tables need to be read without caching.


drbyte:
This is the most cachable query which improves site performance.
They are intentionally cached for `2.5` minutes, to reduce query burden on every hit to the site.

scottcwilson:
Fair enough.  I may document this in the Down for Maintenance section.

PR #4426, opened 2021-07-23
admin, manufacturers, add column title for ID to listing page

torvista:
TABLE_HEADING_ID is ripe for rationalisation to main language file in due course, I think.


drbyte:
Why does the `id` column need a title?

torvista:
I think it looks odd without a title, consistency with other pages, and takes up no more space.

drbyte:
>  and takes up no more space.

Maybe.

But it does take up a bunch more space in language-translation burden.

That's why I intentionally excluded/removed the ID column from the page.

torvista:
> But it does take up a bunch more space in language-translation burden.

ID?

I can put this in the main lang file and take it out of the others now.

PR #4428, opened 2021-07-23
module/product_listing: remove empty classes

torvista:


drbyte:
Is some online validator complaining and causing some SEO scoreboard to rank poorly?

Or is this just a visual preference?

One could argue that leaving the empty classes makes it abundantly clear to the end-user exactly where to add a class if they wanted to customize the appearance.

torvista:
None of the above, I just saw the empty tags.

PR #4429, opened 2021-07-25
Add notifier to init_admin_auth.php, enabling additional non-superuser actions to be denied

zenexpert:
add an observer to catch any action by non-superuser
if non-superuser is NOT allowed to delete a category, a product or an order, this notifier can easily be used to prevent such actions

### DIFF
--- a/admin/includes/init_includes/init_admin_auth.php
+++ b/admin/includes/init_includes/init_admin_auth.php
@@ -69,7 +69,7 @@ if (basename($PHP_SELF) != FILENAME_ALERT_PAGE . '.php')
         zen_record_admin_activity('Attempted access to unauthorized page [' . $page . ']. Redirected to DENIED page instead.', 'notice');
         zen_redirect(zen_href_link(FILENAME_DENIED, '', 'SSL'));
       }
-        $zco_notifier->notify('NOTIFY_ADMIN_ACTIVITY_ACTION_EVENT');
+        $zco_notifier->notify('NOTIFY_ADMIN_NONSUPERUSER_ACTION');
     }
 
   }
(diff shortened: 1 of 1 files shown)


proseLA:
why would one pass the `$_GET` vars?  are they not already available from within any script?  or are they sanitized and hence the need to pass them to the observer?

zenexpert:
> 
> 
> why would one pass the `$_GET` vars? are they not already available from within any script? or are they sanitized and hence the need to pass them to the observer?

Because my first attempt without passing it wasn't working so I figured I'd just pass it. But yeah, my first attempt wasn't working for a completely different reason and I forgot to revisit after fixing that... Whoops. 
Anyway, yeah, you're right, no need to pass the `$_GET` vars. Updated. Thanks.

zcwilt:
Could we give the notifier name something more indicative of the action
NOTIFY_ADMIN_ACTIVITY_ACTION_EVENT is a little obtuse

zenexpert:
Such as? The idea was it would be related to the actual "action" parameter, and possibly usable for other actions you might want to observe, prevent, trigger something etc. It's not necessarily linked to a "delete" action. But yeah, go ahead and change to whatever you feel would be better. Maybe NOTIFY_ADMIN_ACTIVITY_NONSUPERUSER_ACTION_EVENT ;) I dunno

scottcwilson:
How about `NOTIFY_ADMIN_NONSUPERUSER_ACTION`?  This one is otherwise ready to go.

lat9:
> How about `NOTIFY_ADMIN_NONSUPERUSER_ACTION`? This one is otherwise ready to go.

@zenexpert, were you going to make that notifier name change?

zenexpert:
Sorry, I completely forgot about this... Done.

PR #4431, opened 2021-07-28
Add notifier for category image

scottcwilson:
Having a notifier would make the integration of a plugin like Automatic Category Images possible with no core file changes. 

https://www.zen-cart.com/downloads.php?do=file&id=1977

### DIFF
--- a/includes/modules/category_row.php
+++ b/includes/modules/category_row.php
@@ -28,7 +28,10 @@ if ($num_categories > 0) {
   }
 
   while (!$categories->EOF) {
-    if (!$categories->fields['categories_image']) $categories->fields['categories_image'] = 'pixel_trans.gif';
+    $zco_notifier->notify('NOTIFY_CATEGORY_ROW_IMAGE', $categories->fields['categories_id'], $categories->fields['categories_image']); 
+    if (empty($categories->fields['categories_image'])) {
+       $categories->fields['categories_image'] = 'pixel_trans.gif';
+    }
     $cPath_new = zen_get_path($categories->fields['categories_id']);
 
     // strip out 0_ from top level cats


proseLA (changes requested):
why don't you just add 1 line for the notifier, and move your conditionals to your observer?

ie:
```
$zco_notifier->notify('NOTIFY_CATEGORY_ROW_EMPTY_IMAGE', $categories, $categories);
```
this can be done either before or after the assignment of the `pixel_trans.gif`.

i think doing it this way provides more flexibility for others to use the observer as they see fit, as opposed to just your specific use scenario.

scottcwilson:
There's no need to call the observer if the image is properly set.  And there's no sense setting the image to pixel_trans.gif if the observer has changed the value.

proseLA:
- you are wrong.
- adding your conditional in the core code as opposed to in your observer only pollutes the core code.
- just because **you** can not see the need does not make it so.

> And there's no sense setting the image to pixel_trans.gif if the observer has changed the value.

who said that?  not me....

scottcwilson:
I may not be understanding what you are suggesting.  The notifier is called `NOTIFY_CATEGORY_ROW_EMPTY_IMAGE` and I add it after a conditional that checks that the image is in fact empty.  What would you change about that?

proseLA:
this is a new notifier.  so change the notifier name.  i think this code allows for more flexibility:
```
        $zco_notifier->notify('NOTIFY_CATEGORY_ROW_IMAGE', $categories->fields['categories_id'], $categories->fields['categories_image']); 
        if (empty($categories->fields['categories_image'])) {
           $categories->fields['categories_image'] = 'pixel_trans.gif';
        }
```
inside of your observer you can then do the 2nd conditional to test if it is empty, and change accordingly.

i just do not see the need to test for something prior to setting off an observer event.

i think if we are adding more observes (of which i am a fan), they should be as flexible as possible for others to use.

Issue #4434, opened 2021-08-01
When IS_ADMIN_FLAG === true but catalog side files should naturally be loaded?


mc12345678:
Trying to possibly address an issue; however, one way of correcting it is to detect if `IS_ADMIN_FLAG` is true from a file that is stored in the catalog. The file is always accessed when using either the admin or catalog. Part of the intent of the catalog file is to then reach out to other files based on being accessed by the admin "fileset" or the catalog "fileset". 

So, the question is, when might `IS_ADMIN_FLAG` be "officially" set to true (by overall design/need), a catalog side file be accessed, and as a result that strictly only catalog files be loaded? E.g. does one of the plugins that support the admin being the customer cause this condition? Otherwise, should something session related be evaluated?

This potentially relates to issue #4432.

scottcwilson:
When you go to Admin > Modules > Payment | Shipping | Order Total, catalog side files for the associated module type are loaded.  Does this answer your question?  Otherwise I am not sure what you are asking.

PR #4435, opened 2021-08-03
Correct FileSystem->isXXXDir response.

### LINKED ISSUE #4432: Why is FileSystem->isAdminDir response backwards?

````php
````

````php
````

mc12345678:
When admin attempts to use the FileSystem class to determine if a
provided file is asking to use the Admin or Catalog directoy where the full
folder path has been provided, then now returns true if the full path pointed
to that directory side.

Then in identifying the directory to access for a plugin, the previous response
was to only address a path off of the then `$this->context` which had previously
only identified the directory `zc_plugins/admin`; however, even admin plugins
may need to access catalog files either directly in the ZC catalog or in the
plugin's catalog directory. The `InitSystem` class now at least supports admin
access to catalog side files that are stored as a part of the plugin in:
`zc_plugins/PLUGIN_NAME/VERSION/catalog` BUT, operation/access of the plugin
from the catalog side remains disabled. A file (class) included in the plugin's
directory can override the ZC version by being loaded instead of the catalog version.

I.e. if a class is placed in both the `includes/classes/observers` folder
and `zc_plugins/PLUGIN_NAME/VERSION/catalog/includes/classes/observers` folder
the zc_plugins version will be loaded instead of the includes version at least
as demonstrated by a single `zc_plugins/PLUGIN_NAME/VERSION/admin/includes/auto_loaders` file
that either fully referenced DIR_FS_CATALOG . DIR_WS_CLASSES . 'observers' as the classPath
or just `'loadFile' => 'observers/OBSERVER_NAME.php',`.

Additionally when capturing the `debugList` information, the class being
accessed is identified as a `'PLUGIN'` or `'ZC'` class, though this terminology
is modifiable.  Because some classes are included from the current "directory" (admin/catalog)
I did not amplify the messaging to indicate admin/catalog for the messaging
context.  I believe it would be easy to do because could detect `IS_ADMIN_FLAG`
and assign the appropriate designation (which unfortunately is not provided
defined variable... I did; however, maintain sonme level of consistency about
the term being within a single pair of quotes so that if it is to be changed that it would
be easier/global to adjust.

Lastly, in processing/determining whether the loading code is in the admin
or in the catalog, both `includes/init_includes/init_observers.php` and
`includes/modules/extra_functions.php` were updated to use:
````diff
- $context = (new FileSystem)->isAdminDir(__DIR__) ? 'admin' : 'catalog';
+ $context = IS_ADMIN_FLAG ? 'admin' : 'catalog';
````
Because __DIR__ returns the location of where the file is stored in this
case and not side of the store attempting to access the file. This appears to
cause and/or prevent loading admin related files. By following the setting
of `IS_ADMIN_FLAG`, the plugin file(s) associated with the accessing folder (admin/catalog)
are able to be loaded/collected.

For a relative directory, the plugin's "prefix" to the path is provided to indicate
ensure that the appropriate sub-directory within the plugin directory is identified.

Then because InitSystem->findPluginDirectory needs to return the root of the
plugin, the relative directory location (space within the applicable
plugin sub-directory) is potentially modified to

This fixes #4432 and relates to #4434.

### DIFF
--- a/includes/classes/FileSystem.php
+++ b/includes/classes/FileSystem.php
@@ -66,8 +66,8 @@ class FileSystem extends IlluminateFilesystem
     {
         if (!defined('DIR_FS_ADMIN')) return false;
         $test = str_replace(DIR_FS_ADMIN, '', $filePath);
-        if ($test != $filePath) return false;
-        return true;
+        if ($test != $filePath) return true;
+        return false;
     }
 
     public function isCatalogDir($filePath)
@@ -75,8 +75,8 @@ class FileSystem extends IlluminateFilesystem
         if ($this->isAdminDir($filePath)) return false;
         if (!defined('DIR_FS_CATALOG')) return false;
         $test = str_replace(DIR_FS_CATALOG, '', $filePath);
-        if ($test != $filePath) return false;
-        return true;
+        if ($test != $filePath) return true;
+        return false;
 
     }
 
--- a/includes/classes/InitSystem.php
+++ b/includes/classes/InitSystem.php
@@ -78,16 +78,21 @@ class InitSystem
         if (isset($entry['classPath'])) {
             $filePath = $entry['classPath'];
         }
+        $classResult = 'ZC';
         if ($entry['loaderType'] == 'plugin') {
-            $filePath = $this->findPluginDirectory($filePath, $entry['pluginInfo']['unique_key']);
+            $pluginFilePath = $this->findPluginDirectory($filePath, $entry['pluginInfo']['unique_key']);
+            if (file_exists($pluginFilePath . $entry['loadFile'])) {
+                $filePath = $pluginFilePath;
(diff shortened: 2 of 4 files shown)


lat9:
@mc12345678, do you have a simple sample plugin that fails previously and doesn't with your corrections?

lat9:
> @mc12345678, do you have a simple sample plugin that fails previously and doesn't with your corrections?

My question still awaits your answer.

mc12345678:
> > @mc12345678, do you have a simple sample plugin that fails previously and doesn't with your corrections?
> 
> My question still awaits your answer.

That's too much of a loaded question. 

Let's instead go down a path of discovery by my first asking this: 
Is there agreement that there should be a way to identify the requested access path as an admin or catalog location/destination?

scottcwilson:

PR #4440, opened 2021-08-09
add notifier for custom product details in email order confirmation

torvista:
Maybe there is a better way, feel free to point that out!

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -1085,6 +1085,11 @@ class order extends base
 
             $this->notify('NOTIFY_ORDER_PROCESSING_ONE_TIME_CHARGES_BEGIN', $i);
 
+            $products_ordered_email = false;
+            $this->notify('NOTIFY_ORDER_PROCESSING_PRODUCTS_ORDERED_EMAIL', $i, $products_ordered_email);
+            if ($products_ordered_email) {
+                continue;
+            }
             // build output for email notification
             $this->products_ordered .= $this->products[$i]['qty'] . ' x ' . $this->products[$i]['name'] . ($this->products[$i]['model'] != '' ? ' (' . $this->products[$i]['model'] . ') ' : '') . ' = ' .
                 $currencies->display_price($this->products[$i]['final_price'], $this->products[$i]['tax'], $this->products[$i]['qty']) .


scottcwilson:
I don't think you want 
```
            if ($products_ordered_email) {
                continue;
            }
```

I think that if this value is not false you should just append it to `$this->products_ordered` below.

torvista:
$products_ordered_email is only a flag that the observer has been used, since the entire subsequent block of text/html email content is to be **replaced**, not appended, to the vanilla email product detail structure.
The custom code is done in the observer, appending to $class->products_ordered at that point (which contains only the table tags and previous products) and avoiding the vanilla formatting code entirely..

mc12345678:
Umm, isn't the same "action" available by:
collecting the current value of `$this->products_ordered` and `$this->products_ordered_html` in the notifier `NOTIFY_ORDER_PROCESSING_STOCK_DECREMENT_INIT`, determining if something needs to be added/modified compared to what is already in the observer, then observing `NOTIFY_ORDER_PROCESSING_ONE_TIME_CHARGES_BEGIN` to again capture the current value, determine what should be added (perhaps nothing, perhaps only should use the observer's special text) such that if the loop occurs again, any specific information to be captured would be captured/set at `NOTIFY_ORDER_PROCESSING_STOCK_DECREMENT_INIT` (in loop) or afterwards (if on the last item and last item needs modified) at `NOTIFY_ORDER_AFTER_ORDER_CREATE_ADD_PRODUCTS` (loop completion)?

A little difficult to provide exact internal control without fully understanding why/when the subsequent text would need to be modified/replaced. 

Just makes me wonder how many miniscule notifiers should be incorporated throughout the code to accomplish something that is *possible* even if *not practical*. (YES, I recognize that what I propose above is by no means as "simple" as forcing the code to ignore a section in support of allowing complete modification/override of the core/base code.)

torvista:
I'm adding an extra column to the products description table in the email.

While I thought that it may be just possible using other notifiers as you describe, the process is so convoluted that it would not encourage anyone to do it....I would just alter core code to avoid such a pile of work in this case, rather than trying to prove it can be done.

Is there a performance hit from each notifier added?

mc12345678:
> I'm adding an extra column to the products description table in the email.
> 
> While I thought that it may be just possible using other notifiers as you describe, the process is so convoluted that it would not encourage anyone to do it....I would just alter core code to avoid such a pile of work in this case, rather than trying to prove it can be done.
> 
> Is there a performance hit from each notifier added?

I'm sure there is, but perhaps also one for using an existing notifier as well.  Also run the possibility of clashing actions within a given notifier (sequence may be important if controllable). I was thinking more from the "I want a notifier to do this" and "I want a notifier to do that" perspective. (Taking to well above and over) Should there be one every other line? 

That said, I wrote this "little diddy" that specifically works in ZC 1.5.7 and above only at the moment. Do need to fill in some blanks: $criteria_to_cause_replace, $criteria_to_add, NEW_TEXT, NEW_TEXT_HTML, ADDED_TEXT, and ADDED_TEXT_HTML.

Is set to basically auto load by naming the file: includes/classes/observers/auto.order_email_mod_details.php

````php
<?php
/**
 * Code generated by mc12345678 of mc12345678.com 2021
 * This is an observer to support modifying 
 **/


class zcObserverOrderEmailModDetails() extends base
{
    
    function __construct()
    {
        $attachThis = [];
        $attachThis[] = 'NOTIFY_ORDER_PROCESSING_STOCK_DECREMENT_INIT';
        $attachThis[] = 'NOTIFY_ORDER_PROCESSING_ONE_TIME_CHARGES_BEGIN';
        $attachThis[] = 'NOTIFY_ORDER_AFTER_ORDER_CREATE_ADD_PRODUCTS';
        
        $this->attach($this, $attachThis);
    }
    
    /**
     * This observer is where the content is added to the email with the idea that
     *   $order_total_modules->apply_credit();//ICW ADDED FOR CREDIT CLASS SYSTEM
     *   doesn't affect the data being modified.
     * This is also called by earlier processes that want to add/modify the text.
     *
     * May need to do something specific if there are no product though would think
     *   there would be nothing generated.
     **/
    
    //     $this->notify('NOTIFY_ORDER_AFTER_ORDER_CREATE_ADD_PRODUCTS');
    function notify_order_after_order_create_add_products(&$callingClass, $notifier)
    {
        // IF want something special to happen if this is called from inside the loop, then can evaluate the
        //   value of $notifier to see what caused this code to be executed.

        // Early escape if there is nothing to be done with the content.
        if (empty($this->text_action)) {
            return;
        }
        // select case is not used here, because the options are considered to not be mutually exclusive.
        if (in_array('replace', $this->text_action)) {
            $callingClass->products_ordered = $this->text_before . $this->text_replace;
            $callingClass->products_ordered_html = $this->text_before_html . $this->text_replace_html;
        }
        if (in_array('add', $this->text_action)) {
            $callingClass->products_ordered .=‎ $this->text_add;
            $callingClass->products_ordered_html .=‎ $this->text_add_html;
        }
    }

    // $this->notify('NOTIFY_ORDER_PROCESSING_ONE_TIME_CHARGES_BEGIN', $i);
    function notify_order_processing_one_time_charges_begin(&$callingClass, $notifier, $i)
    {
        if (!$criteria_to_modify_output)‎ {
            return;
        }
        $this->i = $i; // Make the looped value of $i available throughout the class for data assembly.

        $this->text_before = $callingClass->products_ordered;
        $this->text_before_html = $callingClass->products_ordered_html;
        if ($criteria_to_cause_replace) {
            $this->text_action[] = 'replace';
            $this->text_replace = NEW_TEXT;
            $this->text_replace_html = NEW_TEXT_HTML;
        ‎}
        if ($criteria_to_add) {
            $this->text_action[] = 'add';
            $this->text_add = ADDED_TEXT;
            $this->text_add_html = ADDED_TEXT_HTML;
        }
    }
 
    //       $this->notify('NOTIFY_ORDER_PROCESSING_STOCK_DECREMENT_INIT', array('i'=>$i), $this->products[$i], $i);
    function notify_order_processing_stock_decrement_init(&$callingClass, $notifier, $i_array, &$productsI, &$i)
    {
        $this->i = $i_array['i']; //used for text generation later
        if (empty($this->i)) {
            $this->text_action = [];
            $this->text_add ='';
            $this->text_add_html ='';
            $this->text_replace = '';
            $this->text_replace_html = '';
        }

        // Cause action associated with the text and internal variables at the beginning of each loop or
        //   as otherwise thought upon restarting the loop.
        $this->notify_order_after_order_create_add_products($callingClass, $notifier);

        // reset tracking variables
        $this->text_action = [];
        $this->text_add ='';
        $this->text_add_html ='';
        $this->text_replace = '';
        $this->text_replace_html = '';
    }
}
````

Now, I wrote it a little "upsidedown" on purpose.. At least it seemed to make sense this way with it going like this:
The code at the end of the method is expected to always run at least at/after the last item in the loop. So, it will always need to be able to address the last item in the loop. Well, might as well just make it the "central processor", So then I backed up to the notifier that was just before all of the content being added by the class. I capture the existing email content referring it as  text_before and text_before_html. Then moving lastly to the notifier just inside the loop. If there was a notifier within this method before that I would have used it to somewhat initialize the associated variables. And then, instead of trying to write the "merging" instruction again, I just call the method that does the handlingl and then reset the variables that are to be cleared for continuing the loop and determining what additional content is needed.

torvista:
I hope you already had that up your sleeve...
It demonstrates what is possible with some (considerable) thought without touching the core code, but I'd say it also demonstrated the need for an easier method.

mc12345678:
> I hope you already had that up your sleeve...
> It demonstrates what is possible with some (considerable) thought without touching the core code, but I'd say it also demonstrated the need for an easier method.

Ehhh, it's similar to other code written for other utilities that have been able to provide useful operation without code modification and what I did write I did so on my mobile device while sitting around. Just had to spruce up a few things. After posting, I considered even adding comment to identify how to preface information not just add afterwards or replace (to me seems pretty clear, but definitely not obvious to all).  Admittedly that code is actually simpler than what I had originally "envisioned/described", but those statements had been without really seeing the involved code and without the better understanding provided by the above "goal".

Note though, if the notifier of this PR is incorporated, there are aspects of the code using that notifier that may be difficult to maintain considering when "replacing" the content via that notifier (where addition to the existing content would still need something similar to the above code or the addition of yet another notifier) the "format" will need to be fully addressed in each "iteration" (version) and/or with unique modification(s)... Yeah, I know nothing really special there and perhaps obvious, just a trade-off of concerns...  

So to my earlier statement of adding notifiers ad nausea, wouldn't it then be prudent to add a notifier at new line 1114 to permit modification just prior to sending to `$order_total_modules->apply_credit()` or to new line 1112 to support adjustment of the content at completion of generation? Then one at line 1052 so that a products_attributes_filename could be stored as empty, but could have some other reason/criteria to take action within that loop while all of the other data is present instead of having to wait for an event outside of that inner loop? 

I'll say this, incorporating this notifier other than whatever processing tradeoff occurs with each notifier will not negatively impact me in any way that I can see and I am not highly familiar with the "secrets of when" a notifier is to be added compared to usage of existing... I tend (try?) to use as much existing code as possible (without core code modification). Perhaps fortunately haven't run into too many issues (yet) of clashing notifiers being used for/by multiple sets of code and having to deal with sequence and value.

scottcwilson:

PR #4442, opened 2021-08-10
Correct notices during installation

mc12345678:
Running PHP 8.0, was installing a brand new store to a new database and
using demo product. Received installation notices that were resolved
when I made these changes.

### DIFF
--- a/zc_install/ajaxAdminSetup.php
+++ b/zc_install/ajaxAdminSetup.php
@@ -12,7 +12,7 @@ define('DIR_FS_ROOT', realpath(__DIR__ . '/../') . '/');
 
 require(DIR_FS_INSTALL . 'includes/application_top.php');
 
-$adminDir = $_POST['adminDir'];
+$adminDir = $_POST['adminDir'] ?? 'admin';
 $wordlist = file(DIR_FS_INSTALL . 'includes/wordlist.csv');
 $max = count($wordlist) - 1;
 $word1 = trim($wordlist[zen_pwd_rand(0,$max)]);
--- a/zc_install/includes/template/templates/admin_setup_default.php
+++ b/zc_install/includes/template/templates/admin_setup_default.php
@@ -8,7 +8,7 @@
 
 <?php require(DIR_FS_INSTALL . DIR_WS_INSTALL_TEMPLATE . 'partials/partial_modal_help.php'); ?>
 
-<?php if (count($errors)) { ?>
+<?php if (!empty($errors)) { ?>
     <div class="alert-box alert">
 <?php
       foreach ($errors as $errormessage) {


zc_install/ajaxAdminSetup.php
```diff
@@ -12,7 +12,7 @@
 
 require(DIR_FS_INSTALL . 'includes/application_top.php');
 
-$adminDir = $_POST['adminDir'];
+$adminDir = $_POST['adminDir'] ?? 'admin';
```
brittainmark, line 15: ?? was introduced in php 7 
Assuming 1.5.8 will have a minimum php version of 7 then fine. If not will need `$adminDir = isset($_POST['adminDir']) ? $_POST['adminDir'] : 'admin';`

brittainmark: Ingore. Have just read the whats new in 1.5.8 says php min version 7.3.

mc12345678:
There are alternate/additional ways to address/remove the cause of the notices/issues addressed by this commit.  I went for the "final" location instead of the originating location as that is associated with one or more files before/within the file that reported the issue.

E.g. the file `zc_install/includes/template/templates/admin_setup_default.php` is expecting `$errors` to be countable (an array); however, it is first called/declared in `zc_install/includes/modules/pages/admin_setup/header_php.php` which is a header file loaded before the template file itself and as a part of "normal" loading.

That variable could be further addressed in any number of ways to include:
Pre-assignment within `zc_install/includes/modules/pages/admin_setup/header_php.php`,
Within a separately called file within the template file,
A check within the template file before use,
as implemented in this commit to use a method of check that does not depend on the variable being countable, or
possibly within the main controller code that calls the header file(s) making the existence of `$errors` possible on all pages (e.g. `zc_install/includes/application_top.php`).

Pre-assignment could be done in any of these locations:
https://github.com/zencart/zencart/blob/ca6442b7947988f529c0c43e05d5233288edc513/zc_install/includes/modules/pages/admin_setup/header_php.php#L11-L13
by adding to the end of this set of code:
````php
$errors = [];
````
or adding the same line of code just before this code block:
https://github.com/zencart/zencart/blob/ca6442b7947988f529c0c43e05d5233288edc513/zc_install/includes/modules/pages/admin_setup/header_php.php#L20-L30

Within the separately called file (which does not appear to have/address the variable in question which also makes it seem like this is an incorrect location in which to place the declaration):
https://github.com/zencart/zencart/blob/ca6442b7947988f529c0c43e05d5233288edc513/zc_install/includes/template/partials/partial_modal_help.php#L9-L19
With suggestion of placing the declaration "up front" of this mostly html code:
````diff
+<?php if (!isset($errors) || !is_countable($errors)) {
+    $errors = [];
+} ?>
<div id="modal-help" class="reveal-modal" data-reveal tabindex="-1">
<div class="modal-header">
<a class="close-reveal-modal">×</a>
<div class="modal-help-title" id="modal-help-title"></div>
</div>
<div id="modal-help-content" class="modal-body">
</div>
<div class="modal-footer">
<button class="radius button dismiss"><?php echo TEXT_CONTINUE; ?></button>
</div>
</div>
````

Then to address a check within the template file before use:
https://github.com/zencart/zencart/blob/ca6442b7947988f529c0c43e05d5233288edc513/zc_install/includes/template/templates/admin_setup_default.php#L9-L11
Could be made as:
````diff
<?php require(DIR_FS_INSTALL . DIR_WS_INSTALL_TEMPLATE . 'partials/partial_modal_help.php'); ?>

+<?php if (!isset($errors) || !is_countable($errors)) {
+    $errors = [];
+} ?>
<?php if (count($errors)) { ?>
````
Note that the content of the above new lines could be moved up or down into an existing php code section instead of being within its own "grouping".

Also note that the code check (`is_countable`) incorporated is based on a minimum PHP version of 7.3 expected for ZC 1.5.8.

scottcwilson:
Thanks for testing in PHP8!

PR #4444, opened 2021-08-12
account order history: make product name link to product as per cart

torvista:


mc12345678:
Any special considerations for the case where the product no longer exists? (E.g. old order pointing to a product that has been discontinued or the like? Or happenstance that was the last instance of the product and site is setup to disable product upon <= 0 quantity?

scottcwilson:
My vote would be KISS.  The cart deals with items that have been deleted, are no longer in stock, etc. already.  Just build a link.

PR #4447, opened 2021-08-15
add notifyers to invoice and packing slip

brittainmark:
Invoice,php changes
1) Restore hardcoded defaults.
2) Additional notifier NOTIFY_ADMIN_INVOICE_DEFAULTS to allow changing of default settings.
3) Additional notifier NOTIFY_ADMIN_INVOICE_HEADING_B4_TAX, NOTIFY_ADMIN_INVOIVE_HEADERS_AFTER_TAX to allow product field headings to be added before and after tax information on the invoice
4) Additional notifier NOTIFY_ADMIN_INVOICE_SORT_DISPLAY and minor code change to allow invoice to be sorted into different sequence. 
5) Additional notifiers NOTIFY_ADMIN_INVOICE_DATA_B4_TAX, NOTIFY_ADMIN_INVOICE_DATA_AFTER_TAX to allow product field to be added before and after tax information on the invoice

packingslip.php changes
1) Additional notifier NOTIFY_ADMIN_PACKINGSLIP_DEFAULTS to allow changing of default settings.
2) Additional notifier NOTIFY_ADMIN_PACKINGSLIP_HEADING to allow product field headings to be added before and after tax information on the packing slip
3) Additional notifier NOTIFY_ADMIN_PACKINGSLIP_SORT_DISPLAY and minor code change to allow packing slip to be sorted into different sequence. 
5) Additional notifiers NOTIFY_ADMIN_PACKINGSLIP_DATA to allow product field to be added before and after tax information on the packing slip


admin/invoice.php
```diff
@@ -124,7 +133,39 @@ function couponpopupWindow(url) { /* just a stub for coupon output that might fi
...
+          if (is_array($extra_headings)) {
+              foreach ($extra_headings as $heading_info) {
+                  $align = (isset($heading_info['align'])) ? (' text-' . $heading_info['align']) : '';
+?>
+            <th class="dataTableHeadingContent<?php echo $align; ?>"><?php echo $heading_info['text']; ?></th>
+<?php
+              }
+          }
+?>
+<?php if ($show_product_tax) { ?>
```
mc12345678, line 166: While I'm not necessarily a fan of requiring all variables to be included in an observer, I don't see in this file where `$show_product_tax` is ever offered an opportunity to not be `true` other than changing (hard coding) the "default" setting above. I say this because, in one respect variables such as `$extra_headings` are included in the latest notifier, both that variable and `$show_product_tax` are accessible as global variables...

So what is the intent of adding the test for `$show_product_tax` if not to support its adjustment within a notifier?

brittainmark: Agreed. I see no reason why it should not be included. I did the mod against 1.5.7 first. $show_product_tax was not in that version. I can add it. My store in the UK does not charge tax (under the vat threshold and do not sell to EU at less that the collection point) so it could simplify my invoices if it were alterable. The only fields that I wanted to alter myself without hardcoding where $show_product_images and $show_attrib_images as I currently hardcode them on my site.

If this were to be added, then would have to move line 26 '$show_product_tax = true;' to above the notifier

if global variable are to be used then the notifier could be removed or changed to ADMIN_INVOICE_START and nothing passed. '$show_product_tax = true;' would still need moving,

mc12345678: I was curious about the possibility that the setting may literally be product based (tax on this product, but not on another). Hadn't looked at how the placement might be modified based on that perspective and its possible that others have more reason to nix the discussion. E.g. if it should be an all or nothing per store, per order, per product, etc..., then placement and consideration should end up being in the right place if observer modifiable at all.

Thanks for explaining at least how it came to be a part of the recommended change.

scottcwilson:
@brittainmark Can you resolve these conflicts so we can merge this in?

brittainmark:
Conflicts resolved. Two show variables product image and attribute image added.

scottcwilson:
Great.  Can you squash these three commits? 

  git checkout <your branch>
  git rebase -i v158
- change all but the first "pick" to "f" (for fix).
  git commit --amend 
  git push -f

brittainmark:
Sorry new to git hub. I have been using the github site to make the changes. If i want to use command line I assume that i need to download and install github on to my local machine. then set up the repository and the branch before i can use the command line s you have given me above. Struggling to get there.  have got git set up. created repository, 
set up **remote.upstream.url=https://github.com/brittainmark/zencart** and checkout v158 rebase does not seem to be working I assume i have done something wrong.

on git hub i can create a new pull request with everything in it.

brittainmark:
Tried again i get a vim session with an empty file as i think it is comparing the same branch. do i need to rebase it with upstream v158? not sure what the command for that would be. Or do i need to checkout upstream v158?

scottcwilson:
probably easier for you to just create a new branch for now.  Maybe once this is done check out Github UI which should make operations like this easier.

brittainmark:
Just to be clear. I think I messed up as i did not read your excellent github docs. Just been reading through them now.
For me to do.
1) create a new branch based on zencart/zencart:develop
2) create a new branch in my fork
3) remake the changes to the file.
4)resubmit the pull request
5)close this pull request?

scottcwilson:
Don't use `zencart/zencart:develop` - that's the 2.0 branch.
Use `zencart/zencart:v158`.  That's the active branch we are working on.

brittainmark:
New pull request #4518 created to replace this one

PR #4454, opened 2021-08-30
admin, category_product_listing, action 'setflag": remove ambiguity

torvista:
I got an error" Undefined array key "flag" for this line. Could not replicate it but I think it needs refactoring so the isset gets parsed first.

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -99,7 +99,7 @@ if (!empty($action)) {
       break;
     case 'setflag':
 
-      if (isset($_POST['flag']) && ($_POST['flag'] == '0') || ($_POST['flag'] == '1')) {
+      if (isset($_POST['flag']) && ($_POST['flag'] === '0' || $_POST['flag'] === '1')) {
         if (isset($_GET['pID'])) {
           zen_set_product_status($_GET['pID'], $_POST['flag']);
         }


proseLA:
pretty easy to replicate:
```
[31-Aug-2021 07:14:34 America/Los_Angeles] Request URI: /admin/index.php?cmd=category_product_listing&action=setflag&pID=1&cPath=1_4&page=1, IP address: 192.168.14.74
#1  require(/var/www/zcdev/admin/category_product_listing.php) called at [/var/www/zcdev/admin/index.php:11]
--> PHP Notice: Undefined index: flag in /var/www/zcdev/admin/category_product_listing.php on line 102.
```
@torvista this is NOT a refactor.  it is a bug.  refactoring does not change logic.

torvista:
> pretty easy to replicate:

How exactly?  I knew it was with enabling, but could not trigger it myself.

proseLA:
> > pretty easy to replicate:
> 
> How exactly? I knew it was with enabling, but could not trigger it myself.

i just used the developer tools to edit the html and remove the flag and then posted the form.

not sure how you got there; but i agree that the logic is wrong.

mc12345678:
Might happen if admin has timed out (also ***assuming*** that keep alive "didn't work") and then product's status was clicked. At that point, the `$_POST` is lost, although the `$_GET` will remain on the link, login is attempted, once complete the `setflag` action is executed but because of left side and (`&&`) happening before evaluating the right side of the or (`||`), the left says no, the right says oops.

PR #4461, opened 2021-09-02
V158 add and incorporate function zen set category image

mc12345678:
Incorporates the idea of reducing number of lines of code and code duplication by use of a single function instead of the same code rewritten multiple times differing only in the image that is assigned. This is similar to the newly added `zen_set_category_status` function.

The second commit in this PR implements the use of the new function `zen_set_category_image` that is in `catalog/includes/functions/functions_categories.php`.

### DIFF
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -162,17 +162,13 @@ if (!empty($action)) {
       }
       // remove the existing image
       if (!empty($_POST['image_delete'])) {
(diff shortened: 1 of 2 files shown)


scottcwilson:
Please smash to a single commit.

scottcwilson:
Love it.  Thanks for tightening up the code!

PR #4462, opened 2021-09-08
Allow image-only category description

zenexpert:
Details discussed here: https://www.zen-cart.com/showthread.php?228328-Category-description-getting-cleared

### DIFF
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -144,7 +144,7 @@ if (!empty($action)) {
         // clean $categories_description of empty tags
         $sql_data_array = [
           'categories_name' => zen_db_prepare_input($categories_name_array[$language_id]),
-          'categories_description' => empty(trim(strip_tags($categories_description_array[$language_id]))) ? '' : zen_db_prepare_input($categories_description_array[$language_id])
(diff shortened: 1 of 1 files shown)


scottcwilson:
I'm not wild about this. The description is supposed to be a block of text.  Could this be a plugin instead?

zenexpert:
It's not a matter of turning it into an additional plugin. Here's a c/p from the forum:

_What I'm saying - we either need a warning in messagestack to let the store owner know that the code was stripped down, or we shouldn't strip it down. Or maybe something else, dunno. Just not a silent strip down._

There's no clear info that it SHOULD be a block of text or anything alike - imagine the frustration not knowing what it *should* be and you add 10-20 images before realizing it's not working and there's no explanation why...

scottcwilson:
This issue has come up exactly once in the 15 years I have been doing this.  It's an edge case and a bad idea (as noted by @dbltoe on the forum).  There's already a category image if you want to display an image.  Descriptions and titles are text.

proseLA:
i disagree.  just because you have only seen it once in 15 years does not mean it's not a problem.

it's another one of those hard coded things that ZC does that has frustrated a valued member of the community.  i think the downside of allowing 4 different tags is minimal.

and more importantly, if you strip them and NOT tell the admin they were stripped, i think that's bad communication.

PR #4463, opened 2021-09-10
correct misuse of ternary operator

proseLA:
use of the ternary operator `?:` does NOT address an isset situation.

use of the null coalescing operator addresses this problem.

from the php page on the operators:

The expression (expr1) ?? (expr2) evaluates to expr2 if expr1 is null, and expr1 otherwise.
In particular, this operator does not emit a notice or warning if the left-hand side value does not exist, just like isset().


scottcwilson:
I think this was a straight-up typo, and the null coalescing operator was what was intended (if you compare vs 1.5.7).  Thanks for catching this.

PR #4469, opened 2021-10-02
Correctly reflect state of slow_query_log global

scottcwilson:
I have seen a number of cases where Admin > Version shows 

MySQL Slow Query Log Status: On

when the true value is Off.

This change fixes that issue.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -318,7 +318,10 @@ function zen_get_system_information($privacy = false)
     $mysql_slow_query_log_status = '';
     $result = $db->Execute("SHOW VARIABLES LIKE 'slow\_query\_log'");
     if (!$result->EOF) {
-        $mysql_slow_query_log_status = $result->fields['Value'];
+       $mysql_slow_query_log_status = '0'; 
+       if (in_array($result->fields['Value'], ['On', 'ON', '1',])) {
+         $mysql_slow_query_log_status = '1'; 
+       } 
     }
     $mysql_slow_query_log_file = '';
     $result = $db->Execute("SHOW VARIABLES LIKE 'slow\_query\_log\_file'");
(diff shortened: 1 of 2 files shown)


scottcwilson:
thanks for catching that.  My eyes kept crossing looking at `$mysql_slow_query_log_status` and `$mysql_slow_query_log_file`. :(

proseLA:
can 7 lines be made into 4?  personally i prefer:
```
        $mysql_slow_query_log_file = '0';
	if (in_array($result->fields['Value'], ['On', 'ON', '1',])) {
		$mysql_slow_query_log_file = '1';
	}
```

admin/includes/functions/general.php
```diff
@@ -318,7 +318,10 @@ function zen_get_system_information($privacy = false)
     $mysql_slow_query_log_status = '';
```
proseLA, line 318: hi scott,
could we not change this line to `= '0'` as opposed to `= ''` and then eliminate line 321 altogether?  in fact, based on the current coding, line 321 is no longer needed, but i think keeping the code a little consistent, ie vars being either `1` or `0` makes it a bit more legible.

mc12345678:
> can 7 lines be made into 4? personally i prefer:
> 
> ```
>         $mysql_slow_query_log_file = '0';
> 	if (in_array($result->fields['Value'], ['On', 'ON', '1',])) {
> 		$mysql_slow_query_log_file = '1';
> 	}
> ```

If attempting to maintain the original strictness, shouldn't this be:
```
        $mysql_slow_query_log_file = '0';
	if (in_array($result->fields['Value'], ['On', 'ON', '1',], true)) {
		$mysql_slow_query_log_file = '1';
	}
```
where `in_array` is used to strictly compare?

proseLA:
i am not a fan of the strictness on the `IN_ARRAY`, as my foo is not that strong on the various mysql, mariadb, etc implementations of server vars.  i think if any of those values comes back and losely matches the array elements we can assume the mysql slow log is turned on.

in addition, good job catching this one @scottcwilson.  the slow log is a useful tool during development.

scottcwilson:
@mc12345678 is correct that his change would make it closer to my original refactor (modulo the variable name error) but I don't think that's truly required.  I think the current PR fixes the bug with a compact change.

PR #4473, opened 2021-10-08
ZC158 options values manager fix for comparison fail from php8

torvista:
When there is no filter set of an option name:
php version <8 : all the option values are listed
php version >8 : no option values are listed
due to string to number comparison change:
https://www.php.net/manual/en/migration80.incompatible.php

In this file the filter is set to an integer or 0 integer, and all comparisons apart from this one in the query are already strict.

### DIFF
--- a/admin/options_values_manager.php
+++ b/admin/options_values_manager.php
@@ -663,14 +663,16 @@ if (!empty($action)) {
           ?>
         </div>
         <?php
+        $filter_condition = '';
+        if ($filter !== 0) $filter_condition = " AND po.products_options_id = " . $filter; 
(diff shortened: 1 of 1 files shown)


scottcwilson:
Good catch.  What about extracting this check so the query is a bit easier to read?

```
$filter_condition = ''; 
if ($filter !== 0) $filter_conditioin =  " AND po.products_options_id = " . $filter; 
```
and then just add $filter_condition to the query 

```
AND pov2po.products_options_values_id != " . PRODUCTS_OPTIONS_VALUES_TEXT_ID . 
                             $filter_condition . " 
```

torvista:
ok, done

PR #4479, opened 2021-11-03
Sold out should not be presented as a button

scottcwilson:
It's confusing to have Sold Out use the same styling as a button with an action.  This change reverts to using the old style image. 

Before:
sold_out_old

After: 
sold_out_new

### DIFF
--- a/includes/functions/functions_general.php
+++ b/includes/functions/functions_general.php
@@ -205,11 +205,11 @@ function zen_get_buy_now_button($product_id, string $buy_now_link, $additional_l
             $return_button = '<a href="' . zen_href_link(FILENAME_ASK_A_QUESTION, 'pid=' . (int)$product_id, 'SSL') . '">' . TEXT_CALL_FOR_PRICE . '</a>';
             break;
         case ($button_check->fields['products_quantity'] <= 0 and SHOW_PRODUCTS_SOLD_OUT_IMAGE == '1'):
-            if ($_GET['main_page'] == zen_get_info_page($product_id)) {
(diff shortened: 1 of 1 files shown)


scottcwilson:
I wonder if there should be template-able functions that return this sort of thing.  This is entirely appropriate for the Bootstrap template but not appropriate for other templates.   I guess we'll see what the reviewers say.

scottcwilson:
@lat9 @proseLA Added a notifier for templates that wish to override.

includes/functions/functions_general.php
```diff
@@ -205,11 +205,10 @@ function zen_get_buy_now_button($product_id, string $buy_now_link, $additional_l
             $return_button = '<a href="' . zen_href_link(FILENAME_ASK_A_QUESTION, 'pid=' . (int)$product_id, 'SSL') . '">' . TEXT_CALL_FOR_PRICE . '</a>';
             break;
         case ($button_check->fields['products_quantity'] <= 0 and SHOW_PRODUCTS_SOLD_OUT_IMAGE == '1'):
-            if ($_GET['main_page'] == zen_get_info_page($product_id)) {
-                $return_button = zen_image_button(BUTTON_IMAGE_SOLD_OUT, BUTTON_SOLD_OUT_ALT);
-            } else {
-                $return_button = zen_image_button(BUTTON_IMAGE_SOLD_OUT_SMALL, BUTTON_SOLD_OUT_SMALL_ALT);
-            }
+            global $template;
+            $image = BUTTON_IMAGE_SOLD_OUT; $alt = BUTTON_SOLD_OUT_ALT;
```
lat9, line 209: Shouldn't this be broken into two separate lines?

PR #4481, opened 2021-11-08
Add trailing space after min

scottcwilson:
Without a trailing space added, products with both a min and units setting are shown as

Min: 5Units: 5

### DIFF
--- a/includes/functions/functions_prices.php
+++ b/includes/functions/functions_prices.php
@@ -575,7 +575,7 @@ function zen_get_products_quantity_min_units_display($product_id, $include_break
         }
 
         if ($check_units != 1) {
-            $the_min_units .= '<span class="qunit">' . ($the_min_units ? ' ' : '') . PRODUCTS_QUANTITY_UNIT_TEXT_LISTING . '&nbsp;' . $check_units . '</span>';
+            $the_min_units .= '<span class="qunit">' . (zen_not_null($the_min_units) ? ' ' : '') . PRODUCTS_QUANTITY_UNIT_TEXT_LISTING . '&nbsp;' . $check_units . '</span>';
         }
 
         // don't check for mixed if no attributes


mc12345678:
Wasn't the purpose of: `($the_min_units ? ' ' : '')` in the below line to specifically address this issue?

````php
            $the_min_units .= '<span class="qunit">' . ($the_min_units ? ' ' : '') . PRODUCTS_QUANTITY_UNIT_TEXT_LISTING . '&nbsp;' . $check_units . '</span>';
````

And to have the space added only if it was needed instead of always being present regardless of something following?

With the suggested correction, a `&nbsp;` will ***always*** be after the last text even if it is desired for something to be adjacent or immediately following... (somewhat thinking of borders on a box or some amount of alignment).

Seems like little has changed in the code from previous versions associated with this area, but would seem that perhaps PHP version has changed (not described in this chain of discussion) and thus either "rules" about `truthiness` or the standard space within the span is not respected and/or CSS is pushing `qunit` to the left enough that yet another space is needed.

My (untested) thought though would be to change the above to:
````diff
-            $the_min_units .= '<span class="qunit">' . ($the_min_units ? ' ' : '') . PRODUCTS_QUANTITY_UNIT_TEXT_LISTING . '&nbsp;' . $check_units . '</span>';
+            $the_min_units .= '<span class="qunit">' . (zen_not_null($the_min_units) ? ' ' : '') . PRODUCTS_QUANTITY_UNIT_TEXT_LISTING . '&nbsp;' . $check_units . '</span>';
````
This allows the "space" to preceed `PRODUCTS_QUANTITY_UNIT_TEXT_LISTING` and not to cause multiple spaces if the text of that constant begins with a standard space (` `) instead of a non-breaking space (`&nbsp;`).

scottcwilson:
> And to have the space added only if it was needed instead of always being present regardless of something following?

This is a fair point. Updated.

PR #4486, opened 2021-11-21
Include Plugins in Search?

mc12345678:
With the addition of Plugins to the admin side of the store, searching
for information within the plugins remains advantageous. However, with the
addition of plugin versions, all plugin software is searched when looking
at the admin and not just the latest or active version nor is there a way
to exclude searching the plugins on a per search basis.

This adds a checkbox to add the zc_plugins directory into the applicable
search which currently is all files, all admin and just the plugins
searches plugins directory. An alternative solution that was temporarily
locally implemented was to add options into the dropdown to describe
performing the search without plugins; however, the list becomes that much
longer and more code to be added.

The choice of add plugins was made to demonstrate that if plugins
was already chosen, it would automatically and already be included, but
if plugins were not originally part of the associated directory (admin for
example doesn't have a plugins sub-directory) then it incorporates the
plugins into the search.

### DIFF
--- a/admin/developers_tool_kit.php
+++ b/admin/developers_tool_kit.php
@@ -22,6 +22,7 @@ if (isset($_GET['configuration_key_lookup']) && $_GET['configuration_key_lookup'
 $configuration_key_lookup = (isset($_POST['configuration_key'])) ? $_POST['configuration_key'] : '';
 $default_context_lines = isset($_POST['context_lines']) ? (int)$_POST['context_lines'] : $default_context_lines;
 $case_sensitive = (isset($_POST['case_sensitive']) && $_POST['case_sensitive']);
+$include_plugins = (isset($_POST['include_plugins']) && $_POST['include_plugins']);
 $q_const = $q_func = $q_class = $q_tpl = $q_all = '';
 
 function getDirList($dirName, $filetypes = 1) {
@@ -149,6 +150,7 @@ function zen_display_files($include_root = false, $filetypesincluded = 1) {
   $file_cnt = 0;
   $cnt_found = 0;
   $case_sensitive = (isset($_POST['case_sensitive']) && $_POST['case_sensitive']);
+  $include_plugins = (isset($_POST['include_plugins']) && $_POST['include_plugins']);
   for ($i = 0, $n = sizeof($directory_array); $i < $n; $i++) {
     // build file content of matching lines
     $file_cnt++;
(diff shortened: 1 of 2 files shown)


admin/developers_tool_kit.php
```diff
@@ -1060,6 +1066,14 @@ function number_pad_with_spaces($number, $n = 0) {
             <?php echo zen_draw_label(TEXT_ALL_FILES_LOOKUPS, 'zv_files', 'class="control-label col-sm-3"'); ?>
           <div class="col-sm-9 col-md-6"><?php echo zen_draw_pull_down_menu('zv_files', $za_lookup, (isset($action) && $action == 'locate_all_files' ? (int)$_POST['zv_files'] : '1'), 'class="form-control"'); ?></div>
         </div>
+        <div class="form-group">
+            <?php echo zen_draw_label(TEXT_INCLUDE_PLUGINS, 'include_plugins', 'class="control-label col-sm-3"'); ?>
+          <div class="col-sm-9 col-md-6">
+            <div class="checkbox">
+              <label for="locate-cs"><?php echo zen_draw_checkbox_field('include_plugins', true, $include_plugins, '', 'id="include_zc_plugins" aria-label="include_zc_plugins"'); ?></label>
```
mc12345678, line 1073: label "for" looks incorrect (`locate-cs` is identified instead of `include_zc_plugins`). This displayed "correctly" in browser, but still is the wrong identifier.

Will correct by forced push.

scottcwilson:
I am inclined to merge. @lat9 any objections ?

admin/developers_tool_kit.php
```diff
@@ -22,6 +22,7 @@
 $configuration_key_lookup = (isset($_POST['configuration_key'])) ? $_POST['configuration_key'] : '';
 $default_context_lines = isset($_POST['context_lines']) ? (int)$_POST['context_lines'] : $default_context_lines;
 $case_sensitive = (isset($_POST['case_sensitive']) && $_POST['case_sensitive']);
+$include_plugins = (isset($_POST['include_plugins']) && $_POST['include_plugins']);
```
lat9, line 25: Why not simplify to `$include_plugins = !empty($_POST['include_plugins']);`?  Same on line 153.

lat9: I suppose that consistency, even if not simplified, is good.

Issue #4488, opened 2021-11-22
Consolidate functions in jscript_addr_pulldowns.php


scottcwilson:
At least some of these files are exact dups.  We should consolidate this code to a single place and reuse it so fixes don't have to be done multiple times. 

includes/modules/pages/checkout_payment_address/jscript_addr_pulldowns.php
includes/modules/pages/create_account/jscript_addr_pulldowns.php
includes/modules/pages/popup_shipping_estimator/jscript_addr_pulldowns.php
includes/modules/pages/checkout_shipping_address/jscript_addr_pulldowns.php
includes/modules/pages/login/jscript_addr_pulldowns.php
includes/modules/pages/address_book_process/jscript_addr_pulldowns.php

torvista:
Previously I had stubs in those places to include a single file from /templates/shared/form_checks
Not done it in my current site, but it should be done.

Seems it was done in a parallel universe.

https://github.com/zencart/zencart/issues/2604
https://github.com/zencart/zencart/pull/999
https://github.com/zencart/zencart/issues/574

scottcwilson:
Ah done in 2.0.  We need to pull this change back to 1.5.x series.

lat9:
Sounds like a good time to convert that javascript to be jQuery-based as well.

lat9:
Does anyone know of a plugin (or other customization) that uses the built-in processing for **_multiple_** address pulldowns on a single page?  I'm guessing **not**, since the `#stBreak`, `#stateLabel` and `#stText` are all (er) unique HTML `id=` references.

I'm just wondering why the current javascript goes for the country/state dropdowns by their name instead of their `id=` attribute.

scottcwilson:
Not AFAIK.  I think even if you had a plugin that allowed shipping different parts of an order to different addresses, it would still use entries from your address book, so it wouldn't be necessary to have two address forms on one page.

lat9:
There's that for review.  Essentially:
1. Moved common processing to `/includes/templates/template_default/jscript/zen_addr_pulldowns.php`.
2. All page-specific `jscript_addr_pulldowns.php` now simply require the above script.
3. Page-specific `on_load_main.js` files updated to remove the `update_zone` function call.  For non-`login` page versions, those are replaced with `javascript:void(0);` calls so that there's still a file to overwrite the no-longer-needed-or-wanted call to `update_zone`.
4. The `shipping_estimator` has been updated to remove its code-based inclusion of the above onload component.

[...]

PR #4492, opened 2021-11-30
v157 Consolidate `jscript_addr_pulldowns.php` processing.

lat9:
Fixes: #4488 

Noting that the various pages' `on_load_main.js` files (other than `login`) are simply `javascript:void(0);` to provide overwrites of the existing onload processing.  

Otherwise, on an upgrade, it would be easy to miss the fact that the file's no longer distributed and the `update_zone` onload-calls will result in a browser console log indicating that the function's missing (it's not available until the document's 'ready').


lat9:
Oops, wrong issue referenced ... should be #4488.

scottcwilson:
Would you please create a 1.5.8 PR for this change?  I don't want to merge in non-critical fixes to 1.5.7 at this point.

scottcwilson:

PR #4495, opened 2021-12-02
Fix incorrect call to get_audiences_list()

### LINKED ISSUE #4494: Audience counts not shown in admin/mail.php even when AUDIENCE_SELECT_DISPLAY_COUNTS = true
admin/mail.php always runs as though the value AUDIENCE_SELECT_DISPLAY_COUNTS were set to false.
This is a result of a bad bug fix in 1.5.7; in 1.5.6 the wrong number of arguments was passed to get_audiences_list() and the fix was to set the second argument to false when it should have been blank. 

This fix can be easily backported to 1.5.7 for anyone who needs it.

scottcwilson:
Fixes #4494 
This bug also exists in Zen Cart 1.5.7. 

Now admin/mail.php will show the counts of each audience if the admin setting Admin > Configuration > Email > Audience-Select Count Display is set to true.


lat9:
Added the Backport label, in case the change becomes warranted for the zc157 branch.

PR #4502, opened 2021-12-20
v158 Begin numbering so that all items on first page area always selectable.

### LINKED ISSUE #4489: Admin page number issue as reported on the forum
For resolution in the future: 
https://www.zen-cart.com/showthread.php?228475-Featured-Products-choosing-deleting-of-10th-item-on-first-page-error-bug/page2

Pages affected, per @mc12345678 : 

banner_manager
category_product_listing
customer_groups (v1.5.8+)
customers
ez_pages
geo_zones
group_pricing
manufacturers 
orders
reviews

mc12345678:
Through observation this counter needs to begin at a value of 0 so that
when a quantity spans more than one page, that the last item on the first
page can be selected allowing the page to reload with it selected and
editable. There may be other arrangements, but this fixes the issue with
the code continuing to exist here.

In some cases, added back the recently removed code to support this operation.

Fixed #4489

### DIFF
--- a/admin/banner_manager.php
+++ b/admin/banner_manager.php
@@ -480,7 +480,7 @@ if (!empty($action)) {
 // reset page when page is unknown
                 if ((empty($_GET['page']) || $_GET['page'] == '1') && !empty($_GET['bID'])) {
                   $check_page = $db->Execute($banners_query_raw);
-                  $check_count = 1;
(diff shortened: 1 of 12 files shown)


scottcwilson:
This PR has caused problems - see #4686 and #4531.

PR #4504, opened 2021-12-22
ZC158, admin, manufacturers: PHP Fatal error: Uncaught TypeError: substr() expects parameter 1 to be string, null given

torvista:
Add 
declare(strict_types=1);
at start of file.

Add a new manufacturer.
Do not set an image.
Save and open again:
PHP Fatal error:  Uncaught TypeError: substr() expects parameter 1 to be string, null given in ADMIN\manufacturers.php:257

https://github.com/zencart/zencart/blob/35b347e7b04903ed50b4c1272f957bec3a53307f/admin/manufacturers.php#L256

Database default is null. This PR just sets that value in the object to an empty string as part of the check as the value is used again a few lines later.

### DIFF
--- a/admin/manufacturers.php
+++ b/admin/manufacturers.php
@@ -253,7 +253,7 @@ if (!empty($action)) {
               $contents[] = ['text' => '<label class="checkbox-inline">' . zen_draw_checkbox_field('featured', '1', $mInfo->featured) . TEXT_MANUFACTURER_FEATURED_LABEL . '</label>'];
               $contents[] = ['text' => zen_draw_label(TEXT_MANUFACTURERS_IMAGE, 'manufacturers_image', 'class="control-label"') . zen_draw_file_field('manufacturers_image', '', ' class="form-control" id="manufacturers_image"') . '<br>' . $mInfo->manufacturers_image];
               $dir_info = zen_build_subdirectories_array(DIR_FS_CATALOG_IMAGES);
-              $default_directory = substr($mInfo->manufacturers_image, 0, strpos($mInfo->manufacturers_image ?? '', '/') + 1);
+              $default_directory = empty($mInfo->manufacturers_image) ? $mInfo->manufacturers_image = '' : substr($mInfo->manufacturers_image, 0, strpos($mInfo->manufacturers_image ?? '', '/') + 1);
 
               $contents[] = ['text' => zen_draw_label(TEXT_PRODUCTS_IMAGE_DIR, 'img_dir', 'class="control-label"') . zen_draw_pull_down_menu('img_dir', $dir_info, $default_directory, 'class="form-control" id="img_dir"')];
 


scottcwilson:
This change is quite difficult to read.  Are you sure?  Why are you setting `$mInfo->manufacturers_image` when you've just verified that it's empty?

torvista:
`Why are you setting $mInfo->manufacturers_image when you've just verified that it's empty?`

Because it is accessed again a few lines down. I was going along the "make minimum changes" idea.

scottcwilson:
How about
```
$default_directory = '';
if (!empty($mInfo->manufacturers_image) {
   $default_directory = substr($mInfo->manufacturers_image, 0, strpos($mInfo->manufacturers_image ?? '', '/') + 1);
}
```

mc12345678:
> How about
> 
> ```
> $default_directory = '';
> if (!empty($mInfo->manufacturers_image)) {
>    $default_directory = substr($mInfo->manufacturers_image, 0, strpos($mInfo->manufacturers_image ?? '', '/') + 1);
> }
> ```

Aren't there "other" issues with this?
```
$default_directory = '';
if (!empty($mInfo->manufacturers_image)) {
   $default_directory = substr($mInfo->manufacturers_image, 0, strpos($mInfo->manufacturers_image ?? '', '/') + 1);
}
```
`strpos($mInfo->manufacturers_image ?? '', '/')` has the potential of returning a `boolean` of `false` which seems likely in some future PHP version that won't add "well" to a value of 1 although currently a boolean of false added to a value is the value. But, there used to be a similar capability associated with numbers and text until PHP 8.x changed that handling...
So I was thinking:
```
$default_directory = '';
if (!empty($mInfo->manufacturers_image)) {
    if (strpos($mInfo->manufacturers_image, '/') === false) {
        $default_directory = $mInfo->manufacturers_image;
    } else {
        $default_directory = substr($mInfo->manufacturers_image, 0, strpos($mInfo->manufacturers_image, '/') + 1);
    }
}
```
or if "considered" necessary, assign another variable to the result of the `strpos` and use that variable so as not to execute the search twice and/or have "duplicate" code. Another reason for this "earlier" test is if the strpos returns false (there is no slash in the name), then simply use the image information and if it is present then use everything up to and including the slash even if the slash is the first character? 

Shouldn't this though be using the `strrpos` function instead? As is, the default directory is only one directory in no matter how many sub-directories deep the image is located... Wouldn't the default directory need to be the information associated with the stored image up to the image's filename? This would make it:
```
$default_directory = '';
if (!empty($mInfo->manufacturers_image)) {
    if (strrpos($mInfo->manufacturers_image, '/') === false) {
        $default_directory = $mInfo->manufacturers_image;
    } else {
        $default_directory = substr($mInfo->manufacturers_image, 0, strrpos($mInfo->manufacturers_image, '/') + 1);
    }
}
```
Hrmm.. Though I see in the image folder selection list that there are no sub-direectories of those sub-directories presented... So, keep with what has been in order to make it easier in the future to change all references at once (meaning don't use `strrpos`)?

scottcwilson:

Issue #4500, opened 2021-12-17
improve performance in admin -> orders


proseLA:
https://github.com/zencart/zencart/blob/35b347e7b04903ed50b4c1272f957bec3a53307f/admin/orders.php#L904-L908

the above query will be returned based on the primary index.  which is the `orders_status_history_id` field.

the primary index and the `date_added` field should mimic their sort and therefore the sort is not required.

i would be interested in examples where those 2 fields do not sort the same; as to whether we should account for those edge cases is a different story.

scottcwilson:
Tested on a large (80K orders) table and didn't see a noticeable difference.

proseLA:
> Tested on a large (80K orders) table and didn't see a noticeable difference.

if i may be so bold, what was your methodology for said test?  and resulting measurements?

i ran the following 3 queries 3 times on a table with 187,953 rows.  the results are below each query.

```
select * from orders_status_history;
```
0.183 sec
0.161
0.161
```
select * from orders_status_history order by date_added;
```
0.260 sec
0.224
0.227
```
select * from orders_status_history order by orders_status_history_id;
```
0.156 sec
0.146
0.162

scottcwilson:
I just loaded the admin > orders page.  The numbers you are posting about would not be noticeable to the human eye.  They're fine improvements, they just don't buy you a faster orders page load.

proseLA:
that's because i am testing on an isolated server speced out reasonably well.  on a shared hosting platform, a 40% reduction in this one query might be helpful.  might not.

i think it is something that people should be aware of as speed, with numerous queries on a single page, is always a concern.  and finding incremental improvements that do NOT affect the operation is, IMO, always worth wild.

i am a bigger fan of changes like this than i am of changing to UPPERCASE in mysql statements to conform to some standard, that while admirable, we never seem to accomodate.  but again, this is merely a personal preference.

torvista:
Everything adds up.
My test showed .0058 to .011 improvement on 35k records.
Seems a no-brainer to implement every performance-related PR/no discussion really necessary if proven.

torvista:
Additionally, should not 
zen_db_input($oID)
be quoted or cast to integer?

proseLA:
> Additionally, should not zen_db_input($oID) be quoted or cast to integer?

i do not think so.

the refactoring of this script, and specifically of `$oID` is an interesting idea.  looking at this section of code:
[...]

PR #4509, opened 2021-12-31
improve query performance #4500

proseLA:
the sort still needs to be there as `$db->Execute` returns the results in a reverse order which is different than running the query in a mysql session.  

this results in faster response with same results.


scottcwilson:
Fantastic.  Thanks for finding and fixing this issue.

Issue #4317, opened 2021-05-06
function zen_create_coupon_code, parameter length hinted as int, but SECURITY_CODE_LENGTH is string


torvista:
ZC158
https://github.com/zencart/zencart/blob/85873df88db6fcafbf70386f41ebef10da79847e/includes/functions/functions_gvcoupons.php#L77-L81

if casting SECURITY_CODE_LENGTH to int, maybe
 $length = (int)$length;
is redundant?

drbyte:
Is PHP reporting a problem?
Or is it just your IDE that's making a suggestion?

torvista:
I didn't get a problem with it, just an IDE red marker.

scottcwilson:
Does casting fix this?

proseLA:
@scottcwilson

IMO, i am not a fan of correcting code because someone's IDE reports a problem.  however, i see a couple of issues.  for testing:  

- i have tested under php 7.4, and 8.0
- create a file in the store root called test.php.  within that file, add the following code:
```
<?php
//    declare(strict_types=1);
    require('includes/application_top.php');

    echo __FILE__ . ':' . __LINE__ . '<br>';
    $test = zen_create_coupon_code();

    echo "-------->" . $test  . "<---------<br>";
    die(__FILE__ . ':' . __LINE__);
```
- now go that url, ie something line https://zcdev.com/test.php
- everything looks fine to me.
- now un-comment out the 1st line with the strict types.  you will now generate an error.
-  this line of code that steve has pointed out will always present a problem in its construction:
https://github.com/zencart/zencart/blob/85873df88db6fcafbf70386f41ebef10da79847e/includes/functions/functions_gvcoupons.php#L77
- if you type hint any method parmater as anything but a string, and use a default any ZC constant, that's a problem.  all zc configuration constants are strings, despite their values potentially looking like integers or boolean.
- how to approach solving this problem is a different story.  one could refactor all of the base ZC calls to cast to int on the call to the function.  i think that's ugly as we now have to refactor all calls to the function.  and i think it defeats the whole purpose of the default value.
- one could remove the type hints.
- an alternative solution could get more to some inherent ways ZC loads the config values.  which seems the most elegant solution for moving forward.  php could soon require all functions to have type hints.  removing them to solve this issue seems kicking the ball down the path.

scottcwilson:
> IMO, i am not a fan of correcting code because someone's IDE reports a problem.  
[...]

PR #4516, opened 2022-01-03
create array of configVals for int config #4317

proseLA:
overriding of defining of configuration values so some can be defined as integer

### DIFF
--- a/laravel/app/Models/Configuration.php
+++ b/laravel/app/Models/Configuration.php
@@ -15,6 +15,7 @@ class Configuration extends Eloquent
     protected $primaryKey = 'configuration_id';
     public $timestamps = false;
     protected $guarded = ['configuration_id'];
+    protected $configAsIntArray = ['SECURITY_CODE_LENGTH',];
 
     // @todo relocate to service class
     public function loadConfigSettings()
@@ -22,7 +23,7 @@ class Configuration extends Eloquent
         $configs = self::all();
         foreach ($configs as $config) {
             $configValue = $config['configuration_value'];
-            if (in_array($config['configuration_group_id'], [2,3])) {
+            if (in_array($config['configuration_group_id'], [2,3]) || in_array($config['configuration_key'],$this->configAsIntArray)) {
                 $configValue = (int)$configValue;
             }
             if (!defined(strtoupper($config['configuration_key']))) {


mc12345678:
For completeness of addressing this issue, shouldn't there also be a `val_function` parameter added for this/these configurations to accommodate integer assignment of initial/update entry? Something like:
````php
'{"error":"TEXT_SECURITY_CODE_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}'
````

**Comment by proseLA** [comment]
> For completeness of addressing this issue, shouldn't there also be a `val_function` parameter added for this/these configurations to accommodate integer assignment of initial/update entry? Something like:
> 
> ```
> '{"error":"TEXT_SECURITY_CODE_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}'
> ```

i agree.

and what might your feelings be with regards to simplifying all these error message constants into 1 constant?  currently there are 59 constants defined that say `Value must be an integer`.  have you seen any shops that have changed that message only for 1 config value and not others? 

i am of the opinion that we do not need 59 constants to say the same thing; especially when i see no one potentially changing them; even in a highly customized shop.

**Comment by scottcwilson** [comment]
> currently there are 59 constants defined that say Value must be an integer.

I could have sworn we squashed these.  Guess not.  Maybe they got lost in the 158 admin rework?

**Comment by proseLA** [comment]
```
charmes: # grepp  'Value must be an integer' | wc -l
59
charmes: # gss
On branch v158
Your branch is up to date with 'origin/v158'.
```

scottcwilson:
LOL so much for having a good memory. 
@mc12345678 Could I impose on you to tighten this up in 1.5.8?

mc12345678:
> > For completeness of addressing this issue, shouldn't there also be a `val_function` parameter added for this/these configurations to accommodate integer assignment of initial/update entry? Something like:
> > ```
> > '{"error":"TEXT_SECURITY_CODE_LENGTH","id":"FILTER_VALIDATE_INT","options":{"options":{"min_range":0}}}'
> > ```
> 
> i agree.
> 
> and what might your feelings be with regards to simplifying all these error message constants into 1 constant? currently there are 59 constants defined that say `Value must be an integer`. have you seen any shops that have changed that message only for 1 config value and not others?
> 
> i am of the opinion that we do not need 59 constants to say the same thing; especially when i see no one potentially changing them; even in a highly customized shop.

In original design, the thought was to make it possible to have unique declarations where in absence of them a generic message would be generated. Others determined that it was necessary to code the individual defines which happen to have the same text for each of those. I am fine with not including the additional defines provided the potential message is possible to be uniquely defined. Of course the current defines lose the added "Min/Max" text that is presented when no define exists.

My thoughts for addressing the additional `integer` defines is to add an additional case within: https://github.com/zencart/zencart/blob/de87916c0c7fc139d88ab8782fc0d5871111f62f/admin/includes/functions/configuration_checks.php#L24-L33

So that these additional integer statements would have a `val_function` beginning with `TEXT_INT_ADMIN` and a "default" statement equal or similar to the above suggestion. I have no plans to add the individual defines basically for the reason stated allowing them to be created if one finds it necessary to provide some alternate text. I have looked back at the history and while I recall reading the proposal(s), I see that I did not respond although I recall at least starting one or more responses.

That said, I can also see that the defines were added so that one could more easily find the "origination" point; however, that would or would have been easier if each of those defines had something unique within them to find the applicable define easily... As is, when the list of text defines were found by string search, then the left side of the define may not provide enough direct information for one to identify the correct one to modify... A bit of a slippery slope in either way, but the idea in design was to permit the flexibility to provide a unique message if one was desired with a generic ideally applicable message in absence of the define. I see that any additional content in the files as some sort of "tool" to help those having interest but minimal coding or code review skills.

scottcwilson:
What I was thinking was change `admin/includes/languages/english/lang.configuration.php` from: 

    'TEXT_MIN_ADMIN_FIRST_NAME_LENGTH' => 'Value must be an integer',
    'TEXT_MIN_ADMIN_LAST_NAME_LENGTH' => 'Value must be an integer',
    'TEXT_MIN_ADMIN_DOB_LENGTH' => 'Value must be an integer',
    'TEXT_MIN_ADMIN_EMAIL_ADDRESS_LENGTH' => 'Value must be an integer',

to

    'TEXT_ADMIN_INTEGER_MSG' => 'Value must be an integer',
    'TEXT_MIN_ADMIN_FIRST_NAME_LENGTH' => TEXT_ADMIN_INTEGER_MSG,
    'TEXT_MIN_ADMIN_LAST_NAME_LENGTH' => TEXT_ADMIN_INTEGER_MSG,
    'TEXT_MIN_ADMIN_DOB_LENGTH' => TEXT_ADMIN_INTEGER_MSG,
    'TEXT_MIN_ADMIN_EMAIL_ADDRESS_LENGTH' => TEXT_ADMIN_INTEGER_MSG,

scottcwilson:
This is a great start, whatever happens with the language defines.  Thanks @proseLA.

PR #4523, opened 2022-01-05
Fix #4487 notices after product removed from cart

### LINKED ISSUE #4487: PHP notices after a product has been automatically removed from cart
ZC version 1.5.7c

On `shopping_cart.php`'s `get_products()` function, if `$check_for_valid_cart == true` and a product that the customer had in cart has meanwhile been disabled, it is removed (with `$this->remove($products_id)`).
However, later on in the same function the product is still being accessed (with `$this->contents[$products_id]`), generating these notices:

```
PHP Notice:  Undefined index: 601:b98d8dcd0327937e0978898c56a40fac in /var/www/html/includes/classes/shopping_cart.php on line 1318
PHP Notice:  Trying to access array offset on value of type null in /var/www/html/includes/classes/shopping_cart.php on line 1318
PHP Notice:  Undefined index: 601:b98d8dcd0327937e0978898c56a40fac in /var/www/html/includes/classes/shopping_cart.php on line 934
PHP Notice:  Trying to access array offset on value of type null in /var/www/html/includes/classes/shopping_cart.php on line 934
```

I would therefore add a check like `if (isset($this->contents[$products_id]))` before line 1305.

marco-pm:


mc12345678:
Wouldn't the same logic/action also need to be applied in the below?:
https://github.com/zencart/zencart/blob/47b8f90e7bd865c6fcb2774c75ea50fa09710ef7/includes/classes/shopping_cart.php#L1288
This area of processing occurs if somehow the product being added to the cart has had attributes "collected" in the page submit where none exist for the product by evaluation of the `product_attributes` table for the product.

The current processing appears that it would cause the same issue described in #4487, but for these attribute "tagged" product. It also seems that the same logic for solution would be applicable.

mc12345678:
Note that the second related item that is discussed above remains unresolved. 

I did at least mentally determine how one might be able to produce a similar notice for that condition: 
As a logged in shopper, place a product in the cart that has at least one attribute. 
Log out. 
From the admin's attribute controller, remove all attributes from the product. 
On the catalog side, have that user login again. 
I would expect similar notices to be generated because of the area of code described above. 

I have not tested nor proven that (yet).

Advise?

scottcwilson:
> Advise?

Test,Reproduce, PR.

mc12345678:
> Note that the second related item that is discussed above remains unresolved.
> 
> I did at least mentally determine how one might be able to produce a similar notice for that condition: As a logged in shopper, place a product in the cart that has at least one attribute. Log out. From the admin's attribute controller, remove all attributes from the product. On the catalog side, have that user login again. I would expect similar notices to be generated because of the area of code described above.
> 
> I have not tested nor proven that (yet).
> 
> Advise?

Upon further review, I've identified the following:
The use of break within the attribute evaluation does not necessarily cause a similar problem. As written, each option value is reviewed for a given product. If any of the option values have been removed, then the (uniquely hashed) product itself is removed and the loop of evaluating the option values selected for that specific product is terminated while the next product is then evaluated.

A potential improvement would be to amplify the messaging to indicate which option value was no longer available (though may be difficult to specify the specifics as apparently that option value is no longer in the database) and allow further notification of either the option that was affected (assuming it remains for further information) or to provide some sort of summation of option information deleted for the product. In either case, the break of line 1288 only stops evaluation of the given product's attributes and not the ability to evaluate other product populated in the cart.

I do not see that the correction related to product removal is needed at [line 1288](https://github.com/zencart/zencart/blob/3748aff13ffd122513abb3adc5ffeb4a12a60446/includes/classes/shopping_cart.php#L1288) making my above suggestion unnecessary. Customer notification is that the product is removed from the cart and a link to the product is provided to allow the customer to navigate to that/those product affected.

mc12345678:
This generally could/should be applied against ZC 1.5.7d+. Have come across this on a ZC 1.5.7 site with PHP 7.4.

PR #4525, opened 2022-01-06
Admin, option values manager: merge cases for Add and Update for common duplicate error message

torvista:
Yes, yes I know.....

I had already expanded the duplicate message to include more information about the duplicate but then realised the code itself was duplicated in both the insert and update cases, so it seemed to make sense to merge them as the only differences are the two sql. I also simplified the loops with foreachs.
Impossible to commit this piecemeal and impossible to see what's changed here in the Github comparison, but much easier with a side-by side file comparison.
Fire away....

Single language store:
dups_single_lang

Dual language store, one duplicated
dups_dual_lang

Dual language store, both duplicated
Clipboard03


scottcwilson:
Great code re-use!  Thanks.

PR #4530, opened 2022-01-08
add a common search form; fix missing closing tag

proseLA:
i decided to use a common function for building the search form.  i'm hopeful of potentially using it in additional places.  it still provides this functionality to the specials and featured pages on the admin side:

image

in addition, both scripts were missing a closing div tag which i added.

### DIFF
--- a/admin/featured.php
+++ b/admin/featured.php
@@ -211,7 +211,12 @@ if (!empty($action)) {
 //            $featured_array[] = $item['products_id'];
 //          }
         }
-        ?>
+
+          if ($action === 'new') {
+              $form = addSearchKeywordForm(FILENAME_FEATURED, $action);
+              echo $form;
+          }
+          ?>
         <div class="row">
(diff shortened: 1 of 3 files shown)


torvista:
Is this result as per design?

Clipboard02

I don't see anything related to "download" in the last two results.

torvista:
In the current PR for a similar search in the Options Values, that also returns unrelated results.

scottcwilson:
@torvista please PR this

proseLA:
> In the current PR for a similar search in the Options Values, that also returns unrelated results.

@torvista the download that you circled is the search filter that you entered and that is by design.  it follows similar styling to a customers page like this:

https://zcdev.com/admin/index.php?cmd=customers&search=prose

if you think there are unrelated results to your search results, i would be curious about that.  the keyword search fields are here:

https://github.com/zencart/zencart/blob/64e09214fd5341622ebf55a2fd22c595b146f7cb/includes/classes/productPulldown.php#L24-L29

at one point i was thinking about notifiers so that one could modify said search fields to their liking.  which is still a possibility.

torvista:
I was not commenting on the location/display of the search term.

I just meant that for the search term "download" I could not see why the last two results would be returned/why they were related to "download"....but the word is in the text, so no problems here.

PR #4534, opened 2022-01-09
Reduce duplication in admin language defines

scottcwilson:


scottcwilson:
Many more opportunities but let's stop here with this PR and get it merged in.

PR #4544, opened 2022-01-11
More de-duping of admin language defs

scottcwilson:
Close to the end for admin side, but this is a decent stopping point.


scottcwilson:
Great question, the answer is not at this time.

torvista:
Perhaps the loader could load lang.filename* instead of just filename. Then an override file could live next to the original and so be in plain view. I'm only thinking admin and the ideal of leaving language pack files untouched from vanilla making comparisons for updates betweeen vanilla and my-site very easy, unlike up to now.

scottcwilson:
I would probably use an overrides folder the way init_includes does if I were to do this.

PR #4548, opened 2022-01-12
checkout_shipping: Use add_session (not add) before redirect

scottcwilson:


lat9:
Shouldn't this also be backported?

scottcwilson:
Good idea- marked for backport.

Issue #4549, opened 2022-01-12
Add search to Option Name Manager, Option Value Manager 


scottcwilson:
I'd like to extend the search capability added by #4530.

proseLA:
@scottcwilson i am not exactly sure what you are asking here.  forgive me, but my foo on option name manager and option value manager is not so strong.  

both of these scripts seem to start with an option name and then create a list of products or categories.  how would you see the searching capability that we implemented be used in this context?

scottcwilson:
Stores that use attributes heavily can have dozens of pages of Names and/or Values. It would be nice to be able to drill in, for example, to see all values associated with a particular name on Option Value manager.

proseLA:
just so we are clear, you are initially talking about the `set_filter` pulldown on the option value manager?  that gets constructed here:

https://github.com/zencart/zencart/blob/64e09214fd5341622ebf55a2fd22c595b146f7cb/admin/options_values_manager.php#L619

this is **VERY** far removed from the PR quoted above.  i created an abstract class called `pulldown` and then extended that class to create 2 new classes, `productPulldown` and `categoryPulldown`.  both of those new classes can make use of the search capability referenced in #4530.  if you wanted the `set_filter` pulldown to make use of searching, one would need to create a new class for the option value pulldown or alternatively build the searching into the existing construction of the pulldown.

considering the existing script to be over 1000 lines, i would be more inclined to do the first rather than the 2nd.  

however, as you can see from the blame, i have never been involved in this script.  and as my experience is rather limited, i am hesitant to do big changes to it with my limited experience.

scottcwilson:
LOL for some reason I had forgotten that this had been addressed by the dropdown already.  Never mind!

PR #4550, opened 2022-01-13
add products options pulldown #4549

proseLA:
#4549 

in my limited experience, i have always felt this page is a little overwhelming.  i'm not sure if this makes it even more so.

in any event, you can see the process of creating a pulldown and how it could work.  the keyword searching does work for the pulldown.


proseLA:
@scottcwilson i'm ready to close this PR unless you feel like you want it.  i have changes i would like to make to the method `addSearchKeywordForm` that make it more readily usable for other users/developers that decide to extend the abstract pulldown class (or use the existing two that already extend it).

scottcwilson:
Sorry for the delay - I will take care of it this week.

scottcwilson:
I don't get it.  You have added a search box to Admin > Catalog > Option Value Manager that reduces the number of items displayed in the Select Option Value to filter dropdown.   Can you provide more details about scenarios where this is helpful? 

I think Select Option Value to filter (added in 157) is enough, no?  When I wrote up #4549 I think I was looking at an old site and getting aggravated.

torvista:
So this search in Values Manager, is actually filtering on Names.
Sure, not necessary if there are few names as the current dropdown filter suffices. 
But evidently there are site with many names, so if it is done and working, why not include it?

PR #4551, opened 2022-01-13
add missing TABLE_HEADING_OPTION_NAME

torvista:
fixes:

> PHP Fatal error: Uncaught Error: Undefined constant "TABLE_HEADING_OPTION_NAME" in D:\GitHub\zencart\admin\options_name_manager.php:530

on Edit Option Name.


scottcwilson:
Thanks - but this is already in #4544.

Issue #4363, opened 2021-06-06
Spiffycal direct date entry issues


scottcwilson:
**Describe the bug**
Go to Admin > Reports > Products Viewed.
In the Start Date, type "3/5/2021".  Hit tab.  Observe that the date is changed to 05/03/2021. 

**Version**
Zen Cart version: 1.5.7
PHP version: 7.3

Same issue observed in older uses of Spiffycal, such as 1.5.6 Admin > Catalog > Specials.

proseLA:
**get rid of spiffyCAL!**

we use datePicker is any number of other places.  of which i am a fan.

https://jqueryui.com/datepicker/

ps, i am able to reproduce the bug as well.

scottcwilson:
It would be great if we could get rid of Spiffycal in Zen Cart 1.5.8 (and replace it with jQuery Datepicker). However, we still have to support 1.5.7 for a while more, so I'd like to get this bug fixed.

scottcwilson:
Closing issue due to lack of interest and activity.

PR #4557, opened 2022-01-18
remove spiffyCal from stats_products_viewed #4363

proseLA:
@swguy this removes spiffycal from this script.  personally i am a fan of having built in functions do all of the validations, so i pulled a lot of that stuff out.  i'm not seeing complaints on datepicker (on other admin scripts) and its error checking so i think we are good there; but some additional testing and another set of eyes is always good.

on another note, in my testing the split page class does not provide accurate results.  which messes up the pager in this script.

in my testing this bug was independent of my branch and should be easily reproduced on the v158 branch.  at least it is for me.

### DIFF
--- a/admin/stats_products_viewed.php
+++ b/admin/stats_products_viewed.php
@@ -26,12 +26,8 @@ if (!function_exists('makeUnixTimestampFromDate')) {
     }
 }
 
-$convertedFormat = str_replace(['mm', 'MM', 'dd', 'yyyy'], ['m', 'm', 'd', 'Y'], DATE_FORMAT_SPIFFYCAL);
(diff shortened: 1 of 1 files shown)


scottcwilson:
Please open a new issue for the split page class problem you are reporting so it can be tracked.  Thank you!

proseLA:
> Please open a new issue for the split page class problem you are reporting so it can be tracked. Thank you!

i will.  but i'm pretty sure i remember bringing up this issue before... or commented on it...  unfortunately i can not find it.  but i will open a new issue.

scottcwilson:
This looks good but can you set decent defaults so the page isn't blank when it loads?  Say CURRENT_YEAR-01-01 to CURRENT_DATE?

scottcwilson:
Great!  Please squash these two commits.

PR #4560, opened 2022-01-18
Allow contact us emails to be cc'ed to others

scottcwilson:
Let's add the cc facility used by order emails, create account, etc. for contact us.


proseLA:
could this not be done in an observer?

in addition, i think if there are cc'ed emails, they should be cc'ed.  here you are creating a completely new email which loses the original to recipient email.

i am not a fan of this approach.  

i would steal from this plugin:
https://www.zen-cart.com/downloads.php?do=file&id=2315

you can look for the `contact_us` and then attach to the existing email.

scottcwilson:
It could be done by an observer but this is core code.  It works the same way the other CC settings do.

proseLA:
those should be changed then as well.  last time i looked (before i placed my original comment) there were a bunch of observers in the core code.

but it is fine.  do not do it in an observer.  

the place to do it is in the `zen_mail` function.  someone wants to contact the store owner, 1 email goes out, multiple recipients.  the same goes for all of those other emails.  the php mail function has the ability to add BCC or CC to the original email.  which seems to be the way to do it.

proseLA:
> that's a lot of risk for a very small return.

is that not the story of a ZC developer?

in answer to your question, yes i think it is worth wild.  if something is worth doing, it is worth doing it properly.  which i TOTALLY think this situation falls into.

the previously referenced plug-in utilizes a notifier from within the zen_mail function.  with out problems.  (at least none are reported on the forum.)

with regards to your question on testing all variants of EMAIL_TRANSPORT, i do not think that is at all necessary.  the mail object gets instantiated here:

https://github.com/zencart/zencart/blob/a57158b8c80f27e61720a5fc9eb24141ff0a180c/includes/functions/functions_email.php#L195

this is well before the transport mechanism is selected (and in turn added to the mail object).

the PHPMailer class handles the addition of BCC or CC; and therefore is transport independent.

i stand by my original opinion including the refactoring of other emails so that only 1 email gets sent with appropriate BCC and CC as required.  

also, please note that order email confirmations sent to the store owner do NOT fall into this category as there is additional information on that email.

scottcwilson:
@proseLA if you would clone your plugin and make it do contact us blind copies instead, we can call this done.

proseLA:
@scottcwilson 

i've looked a bit at the `zen_mail` function.  this `for` loop is over 10 years old and was part of the initial commit:

https://github.com/zencart/zencart/blob/12dd9a46097d3ec900d2b42d2f6be6f81c54c726/includes/functions/functions_email.php#L83-L393

i'm again not a fan of it as we are sending out multiple emails when one would do with multiple to recipients.  

if we ONLY wanted to add the copies to the first contact us, we would need an indicator.  in we wanted to add the copies to all emails (not a fan), we would not need said indicator.

given my preference, prior to line 83, we can add:

```
$sentExtraEmails = false;
```
after line 267, we can add the following code:

```
    if ($module === 'contact_us' && !empty(SEND_EXTRA_CONTACT_EMAILS_TO) && !$sentExtraEmails) {
        $sentExtraEmails = true;
	    foreach (explode(',', SEND_EXTRA_CONTACT_EMAILS_TO) as $key => $value) {
		    if (preg_match("/ *([^<]*) *<([^>]*)> */i", $value, $regs)) {
			    $to_CC_name = str_replace('"', '', trim($regs[1]));
			    $to_CC_email_address = $regs[2];
		    } elseif (preg_match("/ *([^ ]*) */i", $value, $regs)) {
			    $to_CC_email_address = trim($regs[1]);
                $to_CC_name = '';
		    }
		    if (!zen_validate_email($to_CC_email_address)) {
			    continue;
		    }
		    $mail->addCC($to_CC_email_address, $to_CC_name);
	    }
    }
```
if you wanted to log a bad CC address, one could do it before the continue function.

now, we would be sending the contact_us email to the extra recipients in the first contact_us email, and not in any additional ones for the same email.

it would be my opinion that the above block be turned into a function, and use the addAddress function to only send 1 email per request; and just add addresses using the PHPMailer function.

https://github.com/zencart/zencart/blob/12dd9a46097d3ec900d2b42d2f6be6f81c54c726/includes/classes/vendors/PHPMailer/src/PHPMailer.php#L1002-L1016

whether this will rile up the troops, i can not say.  

modifying my plugin is always a possibility to do the same thing; but right now, my plugin is not set up to handle the format that you have established for the new config value.  which again, i think should be part of a function so as to reduce code duplication.

scottcwilson:
Closing without merge - low priority change, don't have time to pursue.

Issue #4558, opened 2022-01-18
admin splitPageResults returns inaccurate query row total


proseLA:
we can not count on this calculation for all queries:
https://github.com/zencart/zencart/blob/2ebaeeb1495c26df624de2624297e139237b5419/admin/includes/classes/split_page_results.php#L114-L119

in my limited testing, any query that already involves a `sum` in the query (or potentially a `count`) will return inaccurate results.  for example, using the test data from a link similar to this: https://zcdev.com/admin/index.php?cmd=stats_products_viewed; prior to calling the `new splitPageResults`, my query is:
```
SELECT p.products_id, pd.products_name, sum(v.views) as total_views, l.name as language, p.products_type, pt.type_handler, pt.allow_add_to_cart FROM products p 
LEFT JOIN products_description pd ON p.products_id = pd.products_id LEFT JOIN languages l ON l.languages_id = pd.language_id 
INNER JOIN count_product_views v ON p.products_id = v.product_id AND v.language_id = l.languages_id LEFT JOIN product_types pt ON p.products_type = pt.type_id 
WHERE date_viewed BETWEEN CAST('2015-12-19' AS DATE) AND CAST('2022-01-18' AS DATE) GROUP BY p.products_id, pd.products_name, language, p.products_type, pt.type_handler, pt.allow_add_to_cart ORDER BY total_views DESC
```
executing this sql statement directly on my test data returns 19 rows.  the `splitPageResults` returns 46.  the cascade effects on the pager as now we have 3 pages with a max result set of 20; when in fact there is only 1 page of results.

scottcwilson:
Would you be willing to take this ticket?

proseLA:
> Would you be willing to take this ticket?

noting that there are 46 instances in 41 different scripts in the admin of `new splitPageResulst` and my limited time will not allow for testing everything, sure, i'll take this ticket.

torvista:
> my limited time will not allow for testing everything

Of course.

If you put up a list of the pages you've not tested, others can take chunks on, hopefully.

proseLA:
@torvista 
i looked at the orders page and products category.  below is a list of pages that use this function.  but in fact, i'm pretty confident my change works.  in fact, i think it's simpler and cleaner.  unfortunately i do not have any benchmarking as to whether it is more or less efficient.

```
admin/music_genre.php
admin/orders_status.php
admin/gv_queue.php
admin/record_artists.php
admin/banner_manager.php
admin/salemaker.php
[...]

PR #4561, opened 2022-01-18
split page results correct count #4558

proseLA:
#4558 

i pulled out code that was not needed; and left in code that was used for `$this->paginateByLetter` which is only used in countries and seems to work fine.

this solution makes use of a mysql CTE in case anyone is curious.


proseLA:
i have now fixed the paginate by letter to at least display results that seem reasonable.

this does **not** change any of the pagination by letter logic.  it only displays accurate information when calling the `display_count` method

scottcwilson:
Is this ready to merge?  If so would you please squash?

proseLA:
squashed...

scottcwilson:
I'd rather not change these files: 
```
admin/downloads_manager.php
admin/ezpages.php
admin/gv_sent.php
admin/paypal.php
```
Those extra fields are sometimes used by plugins and mods. 

I am nervous about this change:
`admin/includes/languages/lang.english.php`

I think seeing "Displaying <start> to <end> of <count>" can be very useful to let you know where you are.

proseLA:
> I'd rather not change these files:
> 
> ```
> admin/downloads_manager.php
> admin/ezpages.php
> admin/gv_sent.php
> admin/paypal.php
> ```
> 
> Those extra fields are sometimes used by plugins and mods.

what extra fields might you be talking about?

`gv_sent` has `c.coupon_id` in there twice.  all i did on that file was remove the extra instance of that field.  the same goes for `paypal` and the field `p.mc_currency`.  even if this PR does not get merged, those 2 scripts should be changed.

the only field missing from `ezpages` is the `languages_id` which the sql selects based on the `$_SESSION['languages_id']`.  i am not sure why a developer would need that field.  but that field could be easily added back.

with regards to the `downloads_manager`, i would be interested to see what plugin or mod uses that script.  and if they are modifying the script, they can very easily change the sql for any fields that they may need.  but in this instance, you are correct: there are fields missing.  i only accounted for the fields used by this script, which is a SQL best practices.

> I am nervous about this change: `admin/includes/languages/lang.english.php`
> 
> I think seeing "Displaying to of " can be very useful to let you know where you are.

with regards to your nerves, lets see what is currently displayed when one clicks on `S` in countries:

image

now lets see that same display with this PR in place:

image

the information is much more accurate.  actually the information is correct.

i was not a fan of this pagination by letter when chris implemented it, although i can see how some people would be.  BUT, this pagination by letter NEVER had the code in there to provide pagination within a specific letter. ie, 2 pages for countries that start with the letter S.

the current pagination class does not work.  and the code is rather difficult to decifer.

this class is much easier to understand and provides accurate results.

torvista:
Regarding Countries, as far as I can see, each letter shows **all** the countries with that letter, so there are no pages (plural) for any single letter, which for this almost permanent/unchanging list where 31 is the max, I think is the best solution.
The second example tells you how may countries have that letter, and the total countries. There's no more information necessary/possible.

scottcwilson:
> what extra fields might you be talking about?

Each of those files has changes to the SELECT statements.  You are right about gv_sent and paypal.  I'll have to look at ezpages and downloads manager.   But why are they in this PR?  They are not related to pagination.

scottcwilson:
To be very clear: please create new PRs for the corrections you are making to the files

```
admin/downloads_manager.php
admin/ezpages.php
admin/gv_sent.php
admin/paypal.php
```
and remove them from this PR.  They are not related to pagination, they are corrections (at least for the last two, I need to review the others still).  They should be in their own PRs.

proseLA:
> To be very clear: please create new PRs for the corrections you are making to the files
> 
> ```
> admin/downloads_manager.php
> admin/ezpages.php
> admin/gv_sent.php
> admin/paypal.php
> ```
> 
> and remove them from this PR. They are not related to pagination, they are corrections (at least for the last two, I need to review the others still). They should be in their own PRs.

i would suggest you review #4558.  the new split page results exposes the problem with the mysql queries as merging the class without the changes to the mysql select statements results in fatal errors.  they are related.

scottcwilson:
I am going to merge these changes I just want them in a separate PR because it's really confusing to view them in the context of this one.  Each one is a small simple change that should be quick to merge.

proseLA:
creating 2 new PRs.  quicker for me than getting into some git debacle.

PR #4563, opened 2022-01-19
Current address should match payment, not shipping

scottcwilson:
If you select a different shipping address from your billing address, the billing address should show up as the selected one when you try to change the payment address. 

See also https://github.com/lat9/ZCA-Bootstrap-Template/pull/106


includes/templates/template_default/templates/tpl_modules_checkout_address_book.php
```diff
@@ -18,6 +18,9 @@
 <?php
   foreach ($addresses as $address) {
     $selected = ($address['address_book_id'] == $_SESSION['sendto']);
+    if ($current_page_base == 'checkout_payment_address') {
```
lat9, line 21: That should use `===` and `FILENAME_CHECKOUT_PAYMENT_ADDRESS`.

scottcwilson: Good catch.

PR #4569, opened 2022-01-22
Storefront constants files: make consistent (idents, spaces, line breaks)

torvista:
While (still) fiddling with my lang.converter it highlighted certain things in the current files that were not consistent in the original files whch I've corrected to:
idents as four spaces, no blank lines in the array, no trailing spaces, blank line after the array closure.

This allows a identical line-by-line comparison with the generated files.


scottcwilson:
Thanks for doing this.  I need to get some earlier work checked in and get you to resolve conflicts, then we'll get this merged in.

PR #4574, opened 2022-01-24
Storefront lang: fix bad conversion on javascript call

torvista:


scottcwilson:
Good catch.  Docs updated.

PR #4580, opened 2022-01-28
Properly position extra customer fields created by NOTIFY_ADMIN_CUSTOMERS_CUSTOMER_EDIT

scottcwilson:
1.5.8 fixes the position of these fields by moving out out of the conditional for ACCOUNT_COMPANY.  However, it just puts them directly below the close of this block, which means they are outside of a row.  

cust_old

This change moves them into the row labelled "Personal" below email address.


scottcwilson:
Layout after fix. 

new_cust

PR #4581, opened 2022-01-28
NOTIFY_ADMIN_ORDERS_DEFAULT_ACTION should pass action just like NOTIFY_ADMIN_ORDER_PREDISPLAY_HOOK does

scottcwilson:


mc12345678:
If it only needs to ***see*** the value, shouldn't the value instead be incorporated in the first parameter as read only? 

Further, shouldn't some reference be included about why it must be incorporated in the notifier and not obtained by way of some global access?

scottcwilson:
> If it only needs to see the value, shouldn't the value instead be incorporated in the first parameter as read only?

No.  This would break any old code which was based on the first parameter being $oID.  
Changes to notifiers should always be additive and compatible with older observers.

torvista:
> Changes to notifiers should always be additive and compatible with older observers.

In a minor update, probably.

But I think 1.57->158 is an ideal moment to make as many breaking changes as necessary to be consistent/better: as long as they are documented in the upgrade notes. Things are already way beyond a simple upgrade.

mc12345678:
> Things are already way beyond a simple upgrade.

Agree, unfortunately.

scottcwilson:
Almost all the other notifiers in this file pass $oID as the first parameter.  Changing that to a fabricated structure just to communicate that the parameter is read-only (and thereby breaking any old code that depends on the old calling convention) is needlessly complicating things.

mc12345678:
> > If it only needs to see the value, shouldn't the value instead be incorporated in the first parameter as read only?
> 
> No. This would break any old code which was based on the first parameter being $oID. Changes to notifiers should always be additive and compatible with older observers.

The first parameter has not been an array, with inclusion of the information about the change in the upgrade notes, a monitoring observer could test for the array status and make assignments as applicable to further use this "required" characteristic and the previously existing value:
````php
function updateNotifyAdminOrdersDefaultAction($callingClass, $notifier, $oID, $order) {
    // BOF Code that can be added to support multiple version usage of observer across Zen Cart versions.
    if (is_array($oID)) {
        $action = $oID['action'];
        $oID = $oID['oID']; // Assign the original value last to retain original code use and not lose other passed information.
    }
    // Funny thing here, is that if the below code *NEEDS* to use `action`, then the not set condition needs to be handled:
    //   Either, escape the observer possibly with some information, or pull in the action value as a global like value...

    // Perform other actions
}
````
If for some reason still using only the `update` method, then becomes even easier to address/revise as more than likely this observer is in combination with other observers and the parameters within the `update` method have been generalized. In that situation, there is likely no reuse of the specific parameter(s) about which to be concerned.

scottcwilson:
So on one hand, we have 

- just add the parameter that was obviously left out.  Maybe it would have been better to have the first parameter as a struct to emphasize the read-only nature of the data, but even this is not obvious because then this observer would have to do parameter handling differently than most of the other observers watching this file. 

and other the other we have: 

- create a bunch of documentation that people won't read, even when they're debugging the problem that your suggestion  creates.
- make your logic much more complicated to deal with the fact that you've recommended a non backwardly compatible change
- break any old code that depends on the original calling conventions.

It's a tough call, but I think I'll do it the first way.

mc12345678:
> So on one hand, we have
> 
> * just add the parameter that was obviously left out.  Maybe it would have been better to have the first parameter as a struct to emphasize the read-only nature of the data, but even this is not obvious because then this observer would have to do parameter handling differently than most of the other observers watching this file.
> 
> and other the other we have:
> 
> * create a bunch of documentation that people won't read, even when they're debugging the problem that your suggestion  creates.
> * make your logic much more complicated to deal with the fact that you've recommended a non backwardly compatible change
> * break any old code that depends on the original calling conventions.
> 
> It's a tough call, but I think I'll do it the first way.

Then the title of this should be changed as the parameter is not provided to just "see" the value of action.

Alternatively, all such parameters be changed to array logic so that additional parameters intended for viewing only can be added with less disruption of effect.

There are any number of other parameters in notifiers that are left out but accessible. Further it is difficult to say that a parameter was "left out", especially since there is already a notifier that specifically provides the action value and there does not appear to be any other changes/modifications in the code that would modify the value of `$action` prior to reaching this point in the code making this PR unnecessary:
https://github.com/zencart/zencart/blob/3494b46607400788cd6997084527dd8bccbee346/admin/orders.php#L78

mc12345678:
Really??? Why should it pass the action in the same way??? What's the point of the first notifier if not to observe and obtain that data or change it if so desired?

Amazing... And with no reason provided why it can not be obtained or adjusted with the existing code.

Shouldn't many of the other notifiers also then pass `$action` in the same way?

PR #4586, opened 2022-01-29
admin, customers, notifier: check for array content

torvista:
It was checking for is_array which was always true.

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -406,7 +406,7 @@ if (!empty($action)) {
           //
           $additional_fields = [];
           $zco_notifier->notify('NOTIFY_ADMIN_CUSTOMERS_CUSTOMER_EDIT', $cInfo, $additional_fields);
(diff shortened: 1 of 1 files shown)


scottcwilson:
Would you please change 

```
if (count($additional_fields) !== 0 ) {
```

to 

```
if (!empty($additional_fields)) {
```

(Covers a greater number of possible problems.) 

Then squash and I'll merge.

proseLA:
> It was checking for is_array which was always true.

i do not agree.  we are passing a var to a notifier.  anyone can do anything they want from within the observer to modify said var.  there is no guarantee at all that the return coming back is an array.

torvista:
Its declared as an array before the notifier. So if there is no modification by an observer, it will continue as an empty array.

https://github.com/zencart/zencart/blob/411ead49f1002a8ca914ee276a9f9905e905a5c7/admin/customers.php#L407-L408

Issue #4531, opened 2022-01-08
#4502 introduces new php notice....


proseLA:
in both of the following examples, the first conditional could be true if `empty($_GET['page']`.  the very following line we are now using that same empty var to assign it to a new var.  that generates a `PHP Notice: Undefined index: page`.

https://github.com/zencart/zencart/blob/7f3d0e06da91bd943af42cbb0fc2eccb6bf7c094/admin/specials.php#L459-L460

https://github.com/zencart/zencart/blob/7f3d0e06da91bd943af42cbb0fc2eccb6bf7c094/admin/featured.php#L371-L372

scottcwilson:
@mc12345678 I think the intention here was `if (!(empty($_GET['page']) ...` wasn't it?

mc12345678:
> @mc12345678 I think the intention here was `if (!(empty($_GET['page']) ...` wasn't it?

No, the intention was to in fact to keep/use `empty()` in the first test. That there were differences between occurrences of the code group (some invoking the notice described above) is/was a direction to address. That additional code adds an interim tracking variable to allow identifying or restoring the original value for some currently unused future situation. I wasn't sure of anyone's use of that variable among the implementations so didn't want to remove without further discussion.

The additional value could be assigned something "appropriate" in these (and perhaps other) uses or it could be removed. While I'm for code simplification I am also for helping the "neighbor" for which this code section that modifies the `$_GET` value(s) may want to evaluate them along the way... Either of these two actions would address the notice that is possibly generated.

By the way, @proseLA what was the use case that generated the notice?

proseLA:
i think i was just clicking around on the pages.  i am away from my development box; once i am back in front of it i can give exact steps to reproduce.  

would not doing
```
$old_page = $_GET['page'] ?? 1;
```
address the issue?

mc12345678:
> would not doing
> 
> ```
> $old_page = $_GET['page'] ?? 1;
> ```
> 
> address the issue?

Unfortunately, no that wouldn't exactly correct the "issue". It wouldn't consistently carry the old/expected value.  further make the need for `$old_page` to not exist because it wouldn't reflect the original value of the page parameter. E.g. a more appropriate tool to capture the old value would be found by:
````php
$old_page = $_GET['page'] ?? null;
````
Again, though that would be if `$old_page` was to be used/kept otherwise that line is not necessary.

proseLA:
@mc12345678

- 23 days ago, you had a PR merged correcting page numbers.
[...]

PR #4588, opened 2022-01-31
Replace isset as able

mc12345678:
Make modifications as suggested in the following post: 

https://github.com/zencart/zencart/pull/4486/files/88be03db7b6cc4ee80d1610711e520b2215e6b58#r782133429

Associated  with pr #4486.

Changing isset followed by a non-falsey test to !empty
Also changing assignments where to use a value of it is set otherwise use some default value to using the format of: `$value = $othervalue ?? $default;`

This is supported by the minimum php version for Zen Cart 1.5.8 and above not expecting to use nor support php version less than 7.

### DIFF
--- a/admin/developers_tool_kit.php
+++ b/admin/developers_tool_kit.php
@@ -13,15 +13,15 @@ if (!empty($_POST) && !isset($_POST['context_lines'])) {
   $_POST['context_lines'] = abs((int)$default_context_lines);
 }
 
-if (isset($_GET['configuration_key_lookup']) && $_GET['configuration_key_lookup'] != '') {
+if (!empty($_GET['configuration_key_lookup']))  {
   $_POST['configuration_key'] = strtoupper($_GET['configuration_key_lookup']);
   $_POST['zv_files'] = 1;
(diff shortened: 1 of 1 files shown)


scottcwilson:
> This is supported by the minimum php version for Zen Cart 1.5.8 and above not expecting to use nor support php version less than 7.

Correct.  Min version 7.3. 
Reference: What's New file for Zen Cart 1.5.8. 
This has been added to the Server Requirements doc as well.

scottcwilson:
I posted a once sentence repro instruction on #4531 so it should be easy to fix.

proseLA:
> I posted a once sentence repro instruction on #4531 so it should be easy to fix.

i agree that it is not hard to fix.  i'm just of the opinion that active contributors should clean up their own code.

i am a huge fan of ownership.  and i would hope that my actions with regards to my own contributions demonstrate that.

scottcwilson:
Agreed - just making sure there was no doubt in @mc12345678 's mind because at first I thought he said he couldn't dup. 
(I see the confusion - my remark above was intended for him, not you.)

PR #4590, opened 2022-01-31
Pagination should be different for search

### LINKED ISSUE #4480: admin/category_product_listing.php uses MAX_DISPLAY_RESULTS_CATEGORIES during search
**Describe the bug**
The appropriate count to use for pagination of search results is MAX_DISPLAY_SEARCH_RESULTS.
Note that this file handles both index listings and searches, so a decision needs to be made at various points whether to use MAX_DISPLAY_RESULTS_CATEGORIES or MAX_DISPLAY_SEARCH_RESULTS.

**Version**
Zen Cart version: 1.5.7
PHP version: 7+

**To Reproduce**
1. In Admin >Configuration > Maximum Values, set a low value for MAX_DISPLAY_RESULTS_CATEGORIES and a high value for  MAX_DISPLAY_SEARCH_RESULTS. 
2. Go to Admin > Catalog > Categories/Products and do a search that you know will return a large number of results.
3. Observe that the lower value MAX_DISPLAY_RESULTS_CATEGORIES is used for pagination.

scottcwilson:
Fixes #4480.
Use `MAX_DISPLAY_RESULTS_CATEGORIES` for pagination only if not searching; otherwise use `MAX_DISPLAY_SEARCH_RESULTS`.

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -17,7 +17,10 @@ $search_result = isset($_GET['search']) && zen_not_null($_GET['search']);
 
 $search_parameter = $search_result ? '&search=' . $_GET['search'] : '';
 $keywords = $search_result ? zen_db_input(zen_db_prepare_input($_GET['search'])) : '';
-
+$max_results = MAX_DISPLAY_RESULTS_CATEGORIES; 
+if (!empty($search_parameter)) {
+   $max_results = MAX_DISPLAY_SEARCH_RESULTS; 
+}
 if (isset($_GET['page'])) {
   $_GET['page'] = (int)$_GET['page'];
 }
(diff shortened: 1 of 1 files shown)


admin/category_product_listing.php
```diff
@@ -17,6 +17,8 @@
 
 $search_parameter = $search_result ? '&search=' . $_GET['search'] : '';
 $keywords = $search_result ? zen_db_input(zen_db_prepare_input($_GET['search'])) : '';
+$max_results = MAX_DISPLAY_RESULTS_CATEGORIES; 
+if (!empty($search_parameter)) $max_results = MAX_DISPLAY_SEARCH_RESULTS;
```
proseLA, line 21: i believe PSR-2 (which i think we follow) requires the use of braces for `if` control statements.  (see 5.1 on https://www.php-fig.org/psr/psr-2/).  (it is my preference as well, else i would not bring it up.)

PR #4601, opened 2022-02-06
remove duplicate and unneeded fields

proseLA:
see #4558 and #4561

### DIFF
--- a/admin/downloads_manager.php
+++ b/admin/downloads_manager.php
@@ -135,7 +135,9 @@ if (!empty($action)) {
               $order_by = " ORDER BY pd.products_name ";
 
 // create split page control
-              $sql = "SELECT pad.*, pa.*, pd.*, p.*
+              $sql = "SELECT pad.products_attributes_filename, pad.products_attributes_maxdays, pad.products_attributes_maxcount, 
+                      pa.products_attributes_id, pa.options_id, pa.options_values_id, pd.products_name, 
+                      p.products_model, p.products_id, p.products_virtual, p.product_is_always_free_shipping
                       FROM " . TABLE_PRODUCTS_ATTRIBUTES_DOWNLOAD . " pad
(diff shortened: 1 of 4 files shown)


scottcwilson:
The changes to `admin/gv_sent.php` and `admin/paypal.php` look good. 

But why are the changes to `admin/downloads_manager.php` and `admin/ezpages.php` required?  There has been a general move away from picking specific fields in the main SELECTs so that people don't have to change them to add a field that they need later on. 

Both 1.5.6 and 1.5.7 have this same query in `admin/ezpages.php`.  It was changed when the ezpages table was split.
The query in `admin/downloads_manager.php` has also been there for a while. 
Do we really need to do those two changes?

proseLA:
best practices for sql suggests to not use `*` in select statements.

more importantly, duplicate column names result in Fatal sql errors as first reported by @torvista and confirmed by me. see:

https://github.com/zencart/zencart/issues/4558#issuecomment-1015903914

scottcwilson:
OK it's the duplicates.  Got it. 
I wonder if it's worth enumerating through the list of fields and just specifying each one so you still get all the data.

proseLA:
> But why are the changes to `admin/downloads_manager.php` and `admin/ezpages.php` required? There has been a general move away from picking specific fields in the main SELECTs so that people don't have to change them to add a field that they need later on.

i would think the move is in the opposite direction.  the use of `select *` in this script goes all the way back to the initial commit:

https://github.com/zencart/zencart/blob/698e46bed1603679d484f5b58ccb918d92f2ca87/admin/downloads_manager.php#L138

> OK it's the duplicates. Got it. I wonder if it's worth enumerating through the list of fields and just specifying each one so you still get all the data.

with no notifiers in `downloads_manager`, any need for additional fields requires a change in base code.  so said modder can also modify the select statement accordingly.

as far as enumerating all fields in the select statement, one could certainly do it.  however, i will not.

PR #4610, opened 2022-02-10
RClassic: rationalise some css properties and remove duplicate properties

torvista:


torvista:
Despite it being far from perfect, Rclassic is still the inbuilt reference at the moment, so minor errors such as this are worth the small effort to correct, I opine.

PR #4619, opened 2022-02-14
email.extras: correct protocol

torvista:


dbltoe:
Sorry to be late on this but, only allowed so many minutes a day at the desk.

If we lose <a href="https://www.zen-cart.com" **rel="noopener noreferrer"** target="_blank">Zen Cart</a>, we go back to an ADA/WCAG violation.

scottcwilson:
This is for admin emails John - it's not storefront content.

PR #4624, opened 2022-02-16
square SCA compliant

brittainmark:
Copy of Rocketsites changes to Square modules to make them SCA compliant.
as per request [The-end-of-Zen-Cart-in-Europe-SCA-kills-payments](https://www.zen-cart.com/showthread.php?228664-The-end-of-Zen-Cart-in-Europe-SCA-kills-payments&p=1386665#post1386665) from @swguy

I have copied the changes to the 1.5.8 stream. I think I have formatted it correctly, please check.
Do you want additional comments in the code as I have removed the //SCA changes 1 of 3 etc. as i thought unnecessary as git shows you the changes

### DIFF
--- a/includes/modules/pages/checkout_payment/jscript_square.php
+++ b/includes/modules/pages/checkout_payment/jscript_square.php
@@ -24,8 +24,19 @@ if (MODULE_PAYMENT_SQUARE_TESTING_MODE === 'Sandbox') {
 
 <script type="text/javascript" title="square">
     var cardNonce;
+    var verificationDetails = { 
+        intent: 'CHARGE', 
+        amount: '<?php echo zen_round ( $order->info['total'], 2 ); ?>', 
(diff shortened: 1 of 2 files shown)


brittainmark:
It appears that some people do not have their location set up properly on the database. 
configuration_key  MODULE_PAYMENT_SQUARE_LOCATION
the configuration_value needs to be in the form "Location Name:[xxxxxxxxxxxx]:Curency code"
Where xxxxxxxxxxxx is the square location value and Curency code is a three letter code for the curency e.g. GBP
it appears that some people have this as blank.

If it is blank then the module fails.

it is easlily fixed in admin>modules>payments
by selecting Square then editing and selecting a location. saving it back sets the field correctly.

I don't know if it would work without a location set. Had a look at the Square docs and it suggests that location id is required.

scottcwilson:
Please squash.

brittainmark:
I'll resubmit.

PR #4626, opened 2022-02-16
Add language define

scottcwilson:
This define is used in includes/functions/functions_osh_update.php which can be called on the storefront side.


scottcwilson:
cc @torvista

torvista:
You don't need to remind me, I'm tracking the merges, thanks,

PR #4627, opened 2022-02-16
Record admin id if Place Order was used when an OSH record is being added 

scottcwilson:
If zen_update_orders_history() is called on the storefront due to an action by an admin logged in using Place Order, the admin id should be recorded in the order status history table.

### DIFF
--- a/admin/includes/functions/admin_access.php
+++ b/admin/includes/functions/admin_access.php
@@ -275,21 +275,6 @@ function zen_read_user($name)
 }
 
 /**
- * Lookup admin user name based on admin id
- * @param int $id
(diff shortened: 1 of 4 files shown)


scottcwilson:
To make this test case pass, I'll have to change not_for_release/testFramework/unittests/testsNotifiers/AdminLoggingTest.php - but I don't want to do that until this PR is approved in concept, because they both have to be merged at once or all the other PR tests will fail.

scottcwilson:
LOL it occurred to me I could just change the test within this same PR so there wouldn't be any weird race conditions.

PR #4629, opened 2022-02-17
square SCA compliant

brittainmark:
Copy of Rocketsites changes to Square modules to make them SCA
compliant.
as per request The-end-of-Zen-Cart-in-Europe-SCA-kills-payments from
@swguy

I have copied the changes to the 1.5.8 stream. I think I have formatted
it correctly, please check.
Do you want additional comments in the code as I have removed the //SCA
changes 1 of 3 etc. as i thought unnecessary as git shows you the
changes

### DIFF
--- a/includes/modules/pages/checkout_payment/jscript_square.php
+++ b/includes/modules/pages/checkout_payment/jscript_square.php
@@ -24,8 +24,19 @@ if (MODULE_PAYMENT_SQUARE_TESTING_MODE === 'Sandbox') {
 
 <script type="text/javascript" title="square">
     var cardNonce;
+    var verificationDetails = { 
+        intent: 'CHARGE', 
+        amount: '<?php echo zen_round ( $order->info['total'], 2 ); ?>', 
+        currencyCode: '<?php echo $_SESSION['currency']; ?>', 
+        billingContact: {
+            givenName: '<?php echo (string)$order->billing['firstname']; ?>',
+            familyName: '<?php echo (string)$order->billing['lastname']; ?>'
+        }
+    };
     var paymentForm = new SqPaymentForm({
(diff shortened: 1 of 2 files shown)


scottcwilson:
Are you satisfied that this is working properly?  There was some back and forth on the support thread and I wasn't sure if it was done.

brittainmark:
I have had it working on mysite for a number of months. The change was written by @rocketsites. I just happened to be the one who still had the code and uploaded it to git hub.
I did discover that if you do not have an entry in the configuration file for location (MODULE_PAYMENT_SQUARE_LOCATION) or that entry is blank or not in the format below, then the module failed.

SELECT * FROM configuration WHERE configuration_key = "MODULE_PAYMENT_SQUARE_LOCATION";

it will return among other fields
configuration_key = MODULE_PAYMENT_SQUARE_LOCATION
configuration_value = Location name:[xxxxxxxxxxx]:£££

where: Location name is the location of the store on your square account
[xxxxxxxxxxx] is the location code from square and
£££ is the three letter currency code e.g. GBP

It can be corrected by:

Log in to your store admin and go to
admin>modules>payment
Select Square then edit and change the location ID.
You should get a list of the location you have set up on Square.
Select the one that represents your online store and save the changes. 

I don't know if a check should be added to ensure that a location is set or not. If it is not, how should it fail? As I said I did not write the fix just copied it over to 1.5.8 for future inclusion.
Currently line 39 of jscript_square.php fails badly with $location is not an array. 

I can look more deelply into it if you want but It will take a bit of time as i will have to fully understand what is going on.

scottcwilson:
Gotcha.  OK, For now let's leave this as an optional modification for those who need it.

Issue #3876, opened 2020-08-12
Product Name centric or Product Model centric?


torvista:
I am under the impression that stores manage their products on names or models, and may in fact not use model at all. 

I am guilty of being model-centric in all my contributions, for both queries and messages (error/confirmation), so the implementation of query results ordering/display and message format is inconsistent across the various admin pages and so guaranteed to not please some people all the time.
Would it be prudent to have an admin switch for priority use name/use model to rationalise query results and message formats?

drbyte:
> Would it be prudent to have an admin switch for priority use name/use model to rationalise query results and message formats?

Probably.

There's already one for catalog-side product-listing results. And it's used in some places in admin query results.

Caveat: there are some admin areas where the sort-order logic is already highly complex based on dropdown order-selection options or column-heading sort selections. Those will have to be handled with care. 

Also, the introduction of a switch can make the code harder to follow if the switch name isn't self-evident in its meaning and implementation.

Another caveat: there are potentially a lot of places affected by such a change.

Also, any such change should focus on ONLY this change, not on changing other things in nearby code.

drbyte:
How do you envision this would be implemented? Let's talk about the approach here before coding it ...

torvista:
Absolutely coincidentally I am looking at catalog sort orders now to fix a bug in something I coded.

> There's already one for catalog-side product-listing results. And it's used in some places in admin query results.

I would have assumed this was PRODUCT_LISTING_DEFAULT_SORT_ORDER, but I don't find it in use admin side (in 156), so you may mean another.

Anyway, in my case at least, how things are displayed (or optionally displayed as customer chooses) catalog-side, is completely different to what I want in the admin. I would not have thought they should be commoned at all.

My hazy idea was to have functions for the messages and the product selection listings, to produce two formats (but allow those to be added to/more formats added).
Then where the product selection listing is actually used, have on-page checkboxes for order by model/name and asc/desc.
The selection of which can set a hidden db constant, subsequently used to set the listing and message formats on all the other admin pages.
That way, no-one has to look for an admin switch, and they can change the option when first feel the need and it stays as so.

drbyte:
Oh, you mean rewrite the whole Admin to use DataTables?

mc12345678:
Likely a LONG way off, but if there were some sort of "onscreen" selections, what's the chance of those being stored on a per admin basis? I know... New curve ball.  Perhaps should just become a new issue?

torvista:
[...]

PR #4630, opened 2022-02-17
low stock email: remove hard-coded text and rejig format

torvista:
Previously

> ID# 2        Matrox G400 32MB              Qty Left: 0

"Qty Left:" was hard-coded text.

Changed to:

> MG400-32MB		 "Matrox G400 32MB" (#2)		 In Stock:  0

Model displays only if used.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -897,8 +897,8 @@ class order extends base
 
                     // for low stock email
                     if ($stock_left <= STOCK_REORDER_LEVEL) {
-                        // WebMakers.com Added: add to low stock email
-                        $this->email_low_stock .= 'ID# ' . zen_get_prid($this->products[$i]['id']) . "\t\t" . $this->products[$i]['model'] . "\t\t" . $this->products[$i]['name'] . "\t\t" . ' Qty Left: ' . $stock_left . "\n";
+                        // add product to low stock email content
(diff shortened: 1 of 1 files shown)


proseLA:
this is another instance of using the hard coded text to reformat an email to one individuals liking.  

there is a lack of consideration to other store owners, who i support, who do not use model for any of their products.  

i have no idea why quotes are added on the product name, there is now an extra space prior to the quoted name if no model exists, and the ID# is now buried (with unneeded parenthesis) as opposed to the beginning.

the code is in fact less legible due to this unneeded formatting, that truly only works for an individuals dataset similar to the author.

at some point, consistency and consideration for others would be nice.  

i reformatted all of the product pulldown output for your model first preference; i would appreciate you doing the same for me.

#4659

torvista:
I don't like upsetting anyone, and that is never my intent, but I understand that the idea of a PR languishing about for a while is precisely to allow criticism and improvement. This one has been here for more than a month.
I may well put something in I am not sure about, so others can add their input.

As I've said before, most people here experience multiple stores in their work, so need to moderate things that come from a narrower outlook like mine.
I have no agenda to push to say what the format should be, it is clear that there are at least two main schools of thought.
I think there should be an agreed default format for both and started a discussion that came to naught:
https://github.com/zencart/zencart/issues/3876

The referenced PR regarding sorting my model for model-centric data is of course gratefully acknowledged, but was, I consider, logical and correct for anyone using models/SKUs, not just me!

Nothing here is cast in stone yet and can be improved, but this issue of formatting product data needs to be discussed and some conclusions agreed on for future modifications and reviewing legacy code.

Perhaps it can be continued in the issue I referenced.

proseLA:
@scottcwilson i have limited bandwidth.  and frankly i was very caught up in addressing the square deprecated library problem.

PR #4637, opened 2022-02-18
remove deprecated html in lang.paypal

torvista:


scottcwilson:
Please describe what you have done in text.  It's really hard to read this change.

torvista:
Replaced the font = green tags with just bold
Removed the nobr tags, some were not nested correctly.

Another PR will be for this call with four parameters instead of three
zen_catalog_href_link(FILENAME_CHECKOUT_PROCESS, '', 'SSL', false)

I assume it doesn't need any?

scottcwilson:
Please post a screenshot of how it looks before and after the change.

torvista:
Before

paypal 1 before

After 
paypal 1 after


Before
paypal 2 before

After
paypal 2 after

PR #4639, opened 2022-02-18
remove type="text/javascript"

torvista:
not required anymore.


scottcwilson:
Agreed.  Thanks for doing this.

Detailed discussion here, including confirmation from Doug Crockford:
https://stackoverflow.com/questions/5265202/do-you-need-text-javascript-specified-in-your-script-tags

PR #4642, opened 2022-02-19
remove type="text/css">

torvista:
and whitespace (auto)


scottcwilson:
Reference: not required in HTML5.
https://stackoverflow.com/questions/7715953/do-we-need-type-text-css-for-link-in-html5

PR #4648, opened 2022-02-21
Admin, sql tool: texts, remove popup

torvista:
**Before**

SQL path before

I found the "details" button did not popup/it showed a new page, and only repeated most of the text on the main page.
So I merged the texts into one warning for the main page, and removed the popup.

**After**

SQL path after

<strike>
And, deep breath....I changed the name from "Install SQL Patches" to "SQL Query Executor" as per the page title as that seems more accurate and does not imply that installing patches is a common thing to do.
</strike>

### DIFF
--- a/admin/includes/languages/english/lang.sqlpatch.php
+++ b/admin/includes/languages/english/lang.sqlpatch.php
@@ -8,14 +8,14 @@
 
 $define = [
     'HEADING_TITLE' => 'SQL Query Executor',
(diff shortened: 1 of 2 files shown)


scottcwilson:
I don't want the rename.  You can close this PR and open one for just the popup or update this PR.

torvista:
Ok, although I still think it is badly named.

PR #4658, opened 2022-02-23
improve search form for use in pulldowns

proseLA:
in further exploring the keyword search, i have found the following changes to make it more universal in its usage.  especially for those developers who might want to use any of the pulldown classes.

i think the parameter input list could be better, ie i think `$includeForm` should be first; but i did not want to change the 3 existing files that currently use this function.

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -503,34 +503,53 @@ function zen_draw_date_selector($fieldname_prefix, $default_date='') {
     return $date_selector;
 }
 
-function addSearchKeywordForm($filename, $action) {
-    $keywords_products = (isset($_POST['keywords']) && zen_not_null($_POST['keywords'])) ? zen_db_input(zen_db_prepare_input($_POST['keywords'])) : '';
-    $fullAction = '';
-    if (!empty($action)) {
-        $fullAction = 'action=' . $action;
-    }
-    $form = '
+    /**
+     * intended to add a universal search for use with any pulldown that extends the
+     * abstract pulldown class.  returns a html string.
+     * @param string $filename
+     * @param string $action
+     * @param bool $includeForm
+     * @param array $extrafieldsArray
+     * @return string
+     */
+
+    function addSearchKeywordForm(string $filename, string $action = '', bool $includeForm = true, array $extrafieldsArray = [])
+    {
+        $keywords_products = (isset($_POST['keywords']) && zen_not_null($_POST['keywords'])) ? zen_db_input(zen_db_prepare_input($_POST['keywords'])) : '';
+        $form = '';
+        $endForm = '';
+        $fullAction = '';
+        if (!empty($action)) {
+            $fullAction = 'action=' . $action;
+        }
+        if ($includeForm) {
+            $form = zen_draw_form('keywords', $filename, $fullAction, 'post', 'class="form-horizontal"');
+            $endForm = '</form>';
+        }
+        $html = '
         <div class="row">
-            <div class="col-sm-offset-2 col-sm-4">' . zen_draw_form('keywords', $filename, $fullAction, 'post', 'class="form-horizontal"') .
-        '<div class="form-group">' .
-        zen_draw_label(HEADING_TITLE_SEARCH_DETAIL, 'keywords', 'class="control-label col-sm-3"') .
-        '<div class="col-sm-9">' .
-        zen_draw_input_field('keywords', '', 'class="form-control" id="keywords"') .
-        '</div>
+            <div class="col-sm-offset-2 col-sm-4">' . $form . '
+                <div class="form-group">' .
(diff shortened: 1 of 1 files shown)


scottcwilson:
Please provide some example use cases as documentation and show some screenshots if appropriate.

proseLA:
scott, in looking at the file diff, perhaps it is a bit confusing.  but understand that this function is called in 3 different ZC places and those calls remain unchanged.  and the resulting output is also unchanged for those 3 function calls in the base ZC code.

these are the differences:

- documentation on input parameters and output parameters
- 2 new input parameters
- defaults for all parameters
- type hinting on all parameters

the new parameters are `$includeForm` and `$extrafieldsArray`.  

in its current form, there is an assumption that we need a call to `zen_draw_form` which is true for the existing calls to base ZC functions.  in working with a new client, they make use of a paid add-on for managing stock levels for products with attributes.  after looking at that code, there is a product pulldown that the plugin author manually creates as part of a form.  when the pulldown value changes, the form gets submitted.  

now the evolution of the function that i am proposing changing was all about being able to search for products prior to creating a pulldown for a user to select something from a much smaller dataset.  pulldowns that have the store's complete product list seem so overwhelming to me; especially when the product count gets into the thousands.

in looking at this paid plugin, i decided to revamp it a bit to use a new pulldown class (that extends the product pulldown class) to create the pulldown and then use this `addSearchKeywordForm` to limit the dataset.  given that the form needs to be submitted upon change of the value, the pulldown needs to be INSIDE of the form.  hence the need to call `zen_draw_form` outside of the function `addSearchKeywordForm` and the need for the changes.

given that it is a paid plugin, i do not want to give up screenshots or other elements; but i feel comfortable sharing how some of my changes were incorporated into it:

```
                        echo zen_draw_form ('choose_pid', FILENAME_PRODUCTS_OPTIONS_STOCK, '', 'post');
                        echo addSearchKeywordForm(FILENAME_PRODUCTS_OPTIONS_STOCK, $action, false);

                        require_once DIR_FS_CATALOG . DIR_WS_CLASSES . 'posmPulldown.php';
                        $pulldown = new posmPulldown();
                        $pulldown->setSort(['products_name',])->showModel(false)->onlyActive(!$include_disabled)->showID(true)->setDefault($pID);
                        echo $pulldown->generatePulldownHtml('pID', 'class="form-control" onChange="this.form.submit()"');
```

given that ANY pulldown that extends the abstract `pulldown` class will incorporate keyword searches, this new pulldown can now reuse the `addSearchKeywordForm` method with no ill effect to it's current uses in the base ZC code.  

i hope that makes sense.

the `$extrafieldsArray` creates new hidden fields that, at the time i thought might be necessary.  they are not used here, but when i did this code, a good deal of time prior to issuing the PR, i thought it made sense.

as always, i stand behind my code and will certainly address and bugs/issues that may come up in a timely fashion.

i am of the opinion that the more objects that can be re-used by developers (or 'modders' as you like to say), the better.  and i think that this one fits that bill.

best.

admin/includes/functions/html_output.php
```diff
@@ -503,34 +503,49 @@ function zen_draw_date_selector($fieldname_prefix, $default_date='') {
...
+     * intended to add a universal search for use with any pulldown that extends the
+     * abstract pulldown class.  returns a html string.
+     * @param string $filename
+     * @param string $action
+     * @param bool $includeForm
+     * @param array $extrafieldsArray
+     * @return string
+     */
+
+    function addSearchKeywordForm(string $filename = '', string $action = '', bool $includeForm = true, array $extrafieldsArray = [])
```
lat9, line 516: Why are there now defaults for the `$filename` and `$action` parameters?  I can 'see' the use for a default/null action, but not the `$filename`, since it's used in both the (now optional) form but also in the 'keyword products' link.

admin/includes/functions/html_output.php
```diff
@@ -503,34 +503,49 @@ function zen_draw_date_selector($fieldname_prefix, $default_date='') {
...
+    {
+        $keywords_products = (isset($_POST['keywords']) && zen_not_null($_POST['keywords'])) ? zen_db_input(zen_db_prepare_input($_POST['keywords'])) : '';
+        $form = '';
+        $endForm = '';
+        if ($includeForm) {
+            $form = zen_draw_form('keywords', $filename, 'action=' . $action, 'post', 'class="form-horizontal"');
```
lat9, line 522: Some forms need no 'action'; why was the `$fullAction` (i.e. include action only if not 'empty') portion removed?

lat9 (changes requested):

admin/includes/functions/html_output.php
proseLA: i agree that `$filename` should be made mandatory.

admin/includes/functions/html_output.php
proseLA: no idea.  it makes sense to have it back in.

scottcwilson:
@proseLA when you're finished, please press the "Resolve Conversation" buttons above so everyone knows it's ready.

PR #4659, opened 2022-02-23
output model first when sorted by model

proseLA:
what @torvista wants, he gets....

i think its better.

### DIFF
--- a/includes/classes/productPulldown.php
+++ b/includes/classes/productPulldown.php
@@ -104,9 +104,9 @@
                 $parm_3 = ' (%3$s)';
             }
 
-            $this->output_string = '%1$s ' . $parm_2 . $parm_3; // format string with name first
+            $this->output_string = '%1$s ' . $parm_2 . $parm_3;  // format string with name first
 
-            if (strpos($this->sort, 'model') && (!empty($this->attributes_join))) { // show model first only when attributes exist
+            if (strpos($this->sort, 'model')) {                  // show model first when sorted by model
                 $this->output_string = (!empty($parm_2) ? $parm_2 . '-' : '') . ' %1$s' . $parm_3; // format string with model first
             }
 


torvista:
In that case, a 30-hour day please?

scottcwilson:
Would you please do before and after screenshots so it's obvious what's happening?

proseLA:
man, you guys are demanding!

before:
image

after:
image

torvista:
Is the first one is not sorted by name?

I had got rid of the brackets when model was leading and ID and space, just to compress it a bit more:
image

Not that I am advocating it, doesn't matter what you do, someone will want to to change it....that was my next step, how to override the default format.

proseLA:
> Is the first one is not sorted by name?
> 
did you really just ask that?

no.

proseLA:
@scottcwilson i do not know how detailed you want to get.  similar to other carts, we could create an open class file for people to override defaults rather easily (ha)...  a little more work, but i can see its value.

scottcwilson:
> @scottcwilson i do not know how detailed you want to get. similar to other carts, we could create an open class file for people to override defaults rather easily (ha)... a little more work, but i can see its value.

Was this regarding another ticket?   I'm not sure I understand what you're saying here ...

proseLA:
> > @scottcwilson i do not know how detailed you want to get. similar to other carts, we could create an open class file for people to override defaults rather easily (ha)... a little more work, but i can see its value.
> 
> Was this regarding another ticket? I'm not sure I understand what you're saying here ...

i'm just thinking out loud.  no new ticket.  right now we have:

`class productPulldown extends pulldown`

we could have 2 separate class files, ie:

`class productPulldownBase extends pulldown`  (this is a rename of the existing `productPulldown` class.)
and
`class productPulldown extends productPulldownBase`

the 2nd one would be completely empty, and then anyone could override any element there for their own needs.  it would make upgrading easier going forward.  its a bit more infrastructure but for those people who want to modify the pulldown, it makes upgrading easier due to not changing any core file.

PR #4661, opened 2022-02-24
Admin, developer email diversion message: add transport info

torvista:
image

1) The constant was wrapped with zen_output_string_protected: I couldn't see why.
2) The additional info is useful after importing a real db for local use and forgetting to change to suit the local server transport method.


scottcwilson:
Good stuff.   Would you please write this up in the dev documentation?

PR #4673, opened 2022-03-03
Guard against first comment being blank

scottcwilson:
Set first false even if initial order comment is blank.


scottcwilson:
Example of problem fixed: 

issue

PR #4674, opened 2022-03-04
ability to change session message on password forget

proseLA:
store owners should be able to make the determination as to whether the session message tells the customer if their email address was in the database or not.


scottcwilson:
Wouldn't this give hints to bad guys about email addresses that could be targeted?

proseLA:
scott, there are plenty of major websites out there that confirm valid email addresses. this PR does NOT do that. it only gives store owners the option to do it without the need to change core ZC code in the event of an upgrade. 

security is always a balancing act. and losing customers who can not retrieve their password because they can not even confirm their address and then relying on an email to get there seems a bridge too far for my clients. and i agree with them. 

forcing developers like myself, who are going to confirm email addresses for various clients, to have to change core ZC code each time there is an upgrade, seems heavy handed to me. 

store owners should have the choice. and even with this small change, they would to know how to use and implement the observer class. it's not as if it's a configuration option. 

i hope that you would approve this PR.

PR #4680, opened 2022-03-07
Remove instructional text from html_includes

scottcwilson:
The default versions of these files are replicated in triplicate and full of content that the storeowner has to delete.  Let's improve this.


lat9:
I remember an issue with previous Zen Carts (not checked for zc158) where if a define-page didn't exist in the template-override directory, the Define Pages editor updated the default version without first copying the file to the template-override.

Could you verify that that handling has been corrected?

scottcwilson:
This behavior has not been changed.  Do you want that included with this PR?

lat9:
Changes to the rendering of the define-pages can be template-specific (e.g. `template_default` vs. `bootstrap`), so I believe that the define-pages editor should also be updated to make a template-override copy of the to-be-edited page if it doesn't already exist.

mc12345678:
Be sure to make this new behavior overridable on a page-by-page basis.

mc12345678:
And similar to #2477... What's changed?

scottcwilson:
Closing without merge - low priority change, don't have time to pursue.

PR #4685, opened 2022-03-09
Redirect to home on bad URL

scottcwilson:
Fixes bad URL:
index.php?inc_subcat=0&keyword=1&main_page=advanced_search_result&page=7&search_in_description=1&sort[]=20a

Reported in 
https://www.zen-cart.com/showthread.php?228698-Error-logs-generated-from-Microsoft-IP

### DIFF
--- a/includes/init_includes/init_sanitize.php
+++ b/includes/init_includes/init_sanitize.php
@@ -15,6 +15,16 @@ use Zencart\Request\Request;
   if (!defined('IS_ADMIN_FLAG')) {
     die('Illegal Access');
   }
+
+  foreach ($_GET as $getvar) { 
+     if (is_array($getvar)) { 
+        $site_array_override = false;
+        $zco_notifier->notify('NOTIFY_INIT_SANITIZE_GET_VAR_CHECK', ['getvarname' => $getvar], $site_array_override);
+        if ($site_array_override === false) {
+           zen_redirect(zen_href_link(FILENAME_DEFAULT));
+        }
+     }
+  }
   $csrfBlackListLocal = array();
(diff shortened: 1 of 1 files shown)


lat9:
I'm wondering whether this should be moved to init_sanitize.php, performing that redirection if **_any_** of the $_GET variables are arrays ... there's nothing (at least in the core) that uses a $_GET array, is there?

proseLA:
i agree with @lat9.

hard coding specific `$_GET` vars seems pretty....  i can't think of a specific word right now...

scottcwilson:
I'd be happy to try this ... we'll have to do some testing to be sure nothing gets broken by it.

torvista:
I've had this in place for a couple of weeks replacing an old change I made for the same reason, and have not detected anything not working (advanced shipper, ceon uri mapping in use).

scottcwilson:
As soon as we find exceptions I'll add them to the list. :)

torvista:
After trying various search permutations on **my** site, I did find this in the Advanced Shipper **admin**

$_POST = array_map('stripslashes_deep', $_POST);
$_GET = array_map('stripslashes_deep', $_GET);   
$_COOKIE = array_map('stripslashes_deep', $_COOKIE);   
$_REQUEST = array_map('stripslashes_deep', $_REQUEST);

which is it's own sanitation.
 but no other uses of GET related to an array.

scottcwilson:
Yes, this is also my recollection.

lat9:
One more (isn't there always?) teeny bit, how about adding a notification just prior to the redirect to enable a site-specific override, e.g.

```php
$site_array_override = false;
$zco_notifier->notify('NOTIFY_INIT_SANITIZE_GET_VAR_CHECK', ['getvarname' => $getvar], $site_array_override);
if ($site_array_override === false) {
   then do the redirect
}
```

PR #4688, opened 2022-03-11
Show Delivery Zone Information for orders

scottcwilson:
With the definition of tax nexus shifting in the US, it may be useful for some shops to see at a glance which state an order is going to.

Feature is switchable for those who may not want it.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -12,9 +12,11 @@ if (isset($module)) {
   unset($module);
 }
 
-$quick_view_popover_enabled = false;
-$includeAttributesInProductDetailRows = true;
-$show_product_tax = true;
+// To override these values, create a file in admin/includes/extra_configures
+if (!isset($quick_view_popover_enabled)) $quick_view_popover_enabled = false;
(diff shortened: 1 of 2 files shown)


scottcwilson:
show_zone

barco57:
This will be helpful for those of us that do a fair bit of troubleshooting and need to check orders that are from the same zone to see if the behavior repeated across multiple orders from the same zone. Quickly spotting orders from the same zone has always been a pain.

admin/orders.php
```diff
@@ -15,6 +15,7 @@
 $quick_view_popover_enabled = false;
 $includeAttributesInProductDetailRows = true;
 $show_product_tax = true;
+$show_zone_info = true;
```
lat9, line 18: This should be at least a pseudo-configuration definition so that stores that don't want the display don't need to 'carve up' this core-file.  Other options would be to add a notification to enable store-specific overrides.

scottcwilson:
OK adding a comment and facilitating using an override file in extra_configures.

admin/orders.php
scottcwilson: OK adding a comment and facilitating using an override file in extra_configures.

PR #4692, opened 2022-03-16
Prevent logs from being created in 1.5.8 testing

scottcwilson:


scottcwilson:
This PR is incorrect and needs to be fixed.

PR #4700, opened 2022-03-24
refactor to correct null order in notifier

proseLA:
`$oID` is cleansed and well established at the top of this script.  continuously casting it to integer makes little sense from a coding standpoint.

the order class is needed here:

https://github.com/zencart/zencart/blob/0b1ff5a1c7c8ef420e7acecc2af5dddd9a41889a/admin/orders.php#L409

and in the current iteration, it is null.  this refactor fixes that oversight.

hopefully in the future, we can eliminate many of these sql queries within this switch action and simply use the order object.


scottcwilson:
Yep, love consolidating the object creation too.  TY.

PR #4713, opened 2022-04-03
Correcting admin/orders.php performance issue when orders have a larg…

lat9:
…e number of products.

Essentially, the orders' listing display takes **_forever_** (and sometimes times out) on stores with (a) a large number of orders, think > 100,000 and (b) orders with a large number of products (like 154).

This is due to the addition made by [this](https://github.com/zencart/zencart/commit/61d2abb213be78e0f1a4e74d439b9a67a45b0856) commit and isn't helped by the addition of the `$includeAttributesInProductDetailRows` flag, since the order-products' gathering query is still performed even if that flag is set to `(bool)false`.

This change moves the product-gathering query so that it's only run when the above flag is set.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -1304,28 +1304,28 @@ if (!empty($action) && $order_exists === true) {
 
                     $show_payment_type = $orders->fields['payment_module_code'] . '<br>' . $orders->fields['shipping_module_code'];
 
-                    $sql = "SELECT op.orders_products_id, op.products_quantity AS qty, op.products_name AS name, op.products_model AS model
-                            FROM " . TABLE_ORDERS_PRODUCTS . " op
-                            WHERE op.orders_id = " . (int)$orders->fields['orders_id'];
-                    $orderProducts = $db->Execute($sql, false, true, 1800);
                     $product_details = '';
                     if ($includeAttributesInProductDetailRows) {
-                    foreach($orderProducts as $product) {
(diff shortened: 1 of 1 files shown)


proseLA:
i too have experienced some orders slowdown on the orders page.  i am not convinced that this single query is the cause.  but i am not opposed.

i remember when this quick popover was implemented and going over it with chris.  i had some other ideas which did not get implemented as he thought the implemented solution resulted in faster results.

the code currently had a bug, and while we are looking at this section, we should address it.  lets say one has this:
```
$quick_view_popover_enabled = true;
$includeAttributesInProductDetailRows = false;
```
we now have a popover, but there is no detail:
image

i think the var `$quick_view_popover_enabled` should be used as the conditional for whether the `$ordersProducts` query gets executed and then `$includeAttributesInProductDetailRows` should be used for the `$productAttributes` getting executed.  and that way the `$products_details` string gets populated if `$quick_view_popover_enabled` is set to true.

lat9:
@proseLA, good point.  I hadn't been keeping up with the slew of zc157/158 changes to the orders.php.  I'll rework either later today or tomorrow.

scottcwilson:
@proseLA please advise when you approve this change.

proseLA:
it looks good to me.

scottcwilson:
@lat9 would you please squash to make one commit?

PR #4716, opened 2022-04-04
reformat popup search help

torvista:
Prompted by no margin.
before:
pre

after:
post


scottcwilson:
This is a nice change.  Is the text still accurate?

torvista:
Functionally yes, physically no: "Microsoft AND (keyboard OR mouse)" does not fit as the search field is limited to 30 chars!!

PR #4727, opened 2022-04-13
v157: Correcting zc_install's zero_dates_cleanup.sql using the zc158 …

lat9:
…version

... otherwise, sites upgrading on strict MySQL installations might have failures during the install process.


scottcwilson:

PR #4729, opened 2022-04-15
add count for pulldown results

proseLA:
i have found the count of the sql results to be useful in further development.


proseLA:
:+1:  thank you.

PR #4733, opened 2022-04-16
Correcting countries' dropdown when a top-of-list country is disabled.

lat9:
... as identified in [this](https://www.zen-cart.com/showthread.php?228769-Empty-entries-in-countries-dropdown-menu&p=1387797#post1387797) Zen Cart forum posting.


proseLA:
in general i am not a fan of issuing logs this way....  but, perhaps in this situation, it's ok...

i do find the message misleading.  country may either be disabled or does not exist.

lat9:
Message updated; thanks, @proseLA, for keeping me honest.

PR #4739, opened 2022-04-21
remove unused array entry

scottcwilson:


proseLA:
seems like edit orders has the same thing (although maybe this version is old):
```
charmes: # grepp stock_reduce
includes/classes/order.php:823:            $this->products[$i]['stock_reduce'] = $this->products[$i]['qty'];
admin/includes/functions/extra_functions/edit_orders_functions.php:952:                $product['stock_reduce'] = $product['qty'];
```

scottcwilson:
That file is not part of the current edit orders so I think we're ok.

proseLA:
i think we were fine, even if it was.  i just found it curious more than anything...  legacy code....oy!

scottcwilson:
Tomorrow's new feature is next year's legacy code. :)

Issue #4743, opened 2022-04-24
ZC158: Admin Activity log creation error when strict_types=1


torvista:
set at start of file:
```
<?php declare(strict_types=1);
```
click on export/Go. Whitescreen,

> PHP Fatal error:  Uncaught TypeError: json_decode(): Argument #1 ($json) must be of type string, bool given in ADMIN\admin_activity.php:184

https://github.com/zencart/zencart/blob/57a150e7fd94bf705b44f01a7266fb40374fd947/admin/admin_activity.php#L184

Reason: $result['gzpost'] is empty.
The only time gzpost is empty in the db is for the entry when the log is reset/cleared by a user: which is always the first line of the log.
So, I'm not sure if the gzpost entry in the db should be corrected for this first line rather than trapping the gzinflate result.

scottcwilson:
Can this be closed?

PR #4744, opened 2022-04-25
admin activity: verify gzinflate result

torvista:
preventing strict error as reported here:
https://github.com/zencart/zencart/issues/4743


torvista:
FYI.
locally I found gzinflate returned false on an empty value (no debug) and so could be tested subsequently.
but remotely, gzinflate on an empty value still caused a debug so I changed this to check for empty first.

PR #4758, opened 2022-04-30
Fixes #4156

### LINKED ISSUE #4156: Brands:  Information sidebox or categories sidebox?
The updates for the `brands` page now includes that page in the information sidebox:

https://github.com/zencart/zencart/blob/514fa65a39f1911a4d37407cb8ad5da0b3d1ad34/includes/modules/sideboxes/information.php#L15

Wouldn't that be better 'placed' in the categories sidebox?

lat9:
Creates an initialization file that sets some "soft" configuration settings, overridable via variables present in a site-specific extra_datafiles module.

Starting with a flag that controls whether/not to display the "Brands" page's link in the information sidebox.

### DIFF
--- a/includes/auto_loaders/config.core.php
+++ b/includes/auto_loaders/config.core.php
@@ -105,6 +105,15 @@ if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
   $autoLoadConfig[50][] = array('autoType'=>'init_script',
                                 'loadFile'=> 'init_sefu.php');
 /**
+ * Breakpoint 55.
+ *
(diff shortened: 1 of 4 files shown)


scottcwilson:
LOVE THIS!  Working on documentation now.

scottcwilson:
Should we do something similar for admin for the flags in orders.php, etc?

lat9:
> Should we do something similar for admin for the flags in orders.php, etc?

That's what I was considering.  That would pull all those 'soft' configuration settings into a single module, making it easier (and more obvious) what site-specific controls are available.

PR #4770, opened 2022-05-03
v158/admin: Disable error-reporting during admin processing ...

### LINKED ISSUE #3159: PHP Notice admin: /local and normal configure.php are both loaded by application_bootstrap
I know I already reported this, but do not find it.
When I have a /local/configure defined, I get all the php notices in the browser, for all the repeated cosntants, eg.:

> Notice: Constant HTTP_SERVER already defined in......0\Dev\public_html\zencart-157\admin\includes\configure.php on line 17 

I just put a clause around the second load and have had no problems, prepare a PR?

lat9:
... until the actual error-reporting handling is loaded to prevent "duplicate constant" warnings when the site uses a /includes/local/configure.php file.

Fixes #3159.

### DIFF
--- a/admin/includes/application_bootstrap.php
+++ b/admin/includes/application_bootstrap.php
@@ -44,6 +44,22 @@ $_SERVER['SCRIPT_NAME'] = str_replace($serverScript, '', $_SERVER['SCRIPT_NAME']
 if (!defined('DIR_FS_ADMIN')) define('DIR_FS_ADMIN', preg_replace('#/includes/$#', '/', realpath(__DIR__ . '/../') . '/'));
 
 /**
+ * set the level of error reporting
+ *
+ * Note STRICT_ERROR_REPORTING should never be set to true on a production site.
+ * It is mainly there to show php warnings during testing/bug fixing phases.
+ * note for strict error reporting we also turn on show_errors as this may be disabled
+ * in php.ini. Otherwise we respect the php.ini setting
+ *
+ */
+if ((defined('DEBUG_AUTOLOAD') && DEBUG_AUTOLOAD === true) || (defined('STRICT_ERROR_REPORTING') && STRICT_ERROR_REPORTING === true)) {
+    @ini_set('display_errors', TRUE);
+    error_reporting(defined('STRICT_ERROR_REPORTING_LEVEL') ? STRICT_ERROR_REPORTING_LEVEL : E_ALL);
+} else {
+    error_reporting(0);
+}
+
+/**
  * Ensure minimum PHP version.
  * This is intended to run before any dependencies are required
  * See https://www.zen-cart.com/requirements or run zc_install to see actual requirements!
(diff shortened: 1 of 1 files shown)


torvista:
I get no errors reported when using this file.
But I still find it odd that the preferred solution is to prevent the error reporting, instead of just loading the local, and if not exists, load the default.

torvista:
I get no errors reported when using this file.
But I still find it odd that the preferred solution is to prevent the error reporting, instead of just loading the local, and if not exists, load the default.

torvista (approved):
works for me.

proseLA:
> I get no errors reported when using this file. But I still find it odd that the preferred solution is to prevent the error reporting, instead of just loading the local, and if not exists, load the default.

i do not love the way that works either.  it can be especially confusing when doing a zc_install.

that said, chris had a reason for doing it this way.  the local was only for other config values and not the complete configuration.  and he seemed pretty adamant about keeping it that way.

when chris made a decision, personally i just accepted it; whether i agreed with it or not.  it's done.

i would encourage you to do the same (although you certainly are free to continue to beat a dead horse, as the saying goes).

i'm not sure why you want to continue to think that your way is the only way of doing things.  for me, none of my config files are ever in any of my repositories.  every server is different; and while i put all of my clients sites into github repos and pull directly from the repo, all config values are in my gitignore.

personally, i would thank lat9 for going above and beyond to incorporate code into this repo just for you.  it seems odd that you could not figure out a way around this issue yourself.  but then again i think a lot of things are odd...

torvista:
> i'm not sure why you want to continue to think that your way is the only way of doing things

I don't.
I have one shop, not multiple sites to maintain. I learn from fixing things that break and do that in the non-existent spare time I have.
If I don't understand, that's it, I don't understand, and if I don't ask I don't learn from the "experts".
I did not invent the concept of a local configure for **me**. 
It is not mine and this fix is not for me personally same as your issues are not for you personally. 

I found out this about years ago and the obvious way to use it (for someone working on their ONE shop/not a pro developer), is to install their site on their local Xampp or equivalent (since a new user wanting a local site will not know or care about github or docker) and copy that **complete**, generated config file to /local. I don’t find any mention of the concept in the docs so maybe I don’t know how it is supposed to be used. I just assumed same name=same content.

Then the production configure can stay put and /local can be hidden in a Compare to the production site, never to be copied by mistake.

Your way of doing things is no doubt better, fine-tuned by your work and need to have a workflow for multiple sites. 
It should be documented for others to also add their tricks and save us all some time.

But I'm still a **user** and maybe I have a better feel for what a user will attempt, to better work with their site as easily as 
 possible.
The most inexperienced user is exactly the one who most needs a local site and having two configures seems as simple as possible.
 I don't expect to get criticised for pointing out a problem with it (nor anything else) nor for not understanding the logic behind the solution. 
I do my best, with my limited knowledge and time.

PR #4775, opened 2022-05-05
Address format changes

brittainmark:


scottcwilson:
Thanks for doing this.  For the files in /docs/, it's better if people just send me changes or post them on the forum because I frequently have checkout out changes to those files.  I'll add this change on the next checkin.

PR #4784, opened 2022-05-08
Add documentation for soft flags

scottcwilson:


lat9:
Please include in the comments for each variable what the valid values are and what the default processing is if a setting is left commented-out.

scottcwilson:
This is done.

PR #4796, opened 2022-05-12
v158: Correcting zen_address_pulldowns.php

### LINKED ISSUE #4792: ZC158: Shipping Estimator layout errors with consolidated javascript address pulldown.
Shipping/Packaging->Shipping Estimator Display Settings for Shopping Cart = 2  Display as listing on Shopping Cart page
Customer Details->State - Always display as pulldown? true

Put an item in the cart.

1) Choose a country with zones: United States.

2022 05 10 Shipping Estimator 1

SHOWN State as dropdown: "zoneLabel" (label), "stateZone" (dropdown)  . Correct.
HIDDEN State as text input: "stateLabel" (label, which has no text!), "state" (text input) . Correct.

This **appears** to be correct visually, but also hidden is the "stBreak" linebreak which is part of the **dropdown** group which may/may not be shown dependant on the admin switch ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN.

2) Change country to a one with no zones defined, like Afghanistan.
2022 05 10 Shipping Estimator 2
The original label "zoneLabel" for the dropdown is still shown. Incorrect, apparently.
The dropdown field is hidden. Correct.
The dropdown linebreak "stBreak" is shown, causing the text field to be misaligned with the (dropdown) label. Incorrect.
The label for the input field is shown. But it has no text. Incorrect.
The input field is shown. Correct.

So, some food for thought there.
 
ps. Does 
`<a name="view"></a>`
serve any purpose?

lat9:
Fixes #4792

### DIFF
--- a/includes/templates/template_default/jscript/zen_addr_pulldowns.php
+++ b/includes/templates/template_default/jscript/zen_addr_pulldowns.php
@@ -85,6 +85,14 @@ jQuery(document).ready(function() {
     }
 <?php
     // -----
+    // Issue a notification to enable site-specific jQuery/javascript to be inserted during the
+    // script's initialization.
+    //
+    $extra_javascript = '';
+    $GLOBALS['zco_notifier']->notify('NOTIFY_ZEN_ADDR_PULLDOWNS_INIT', $c2z, $extra_javascript);
+    echo $extra_javascript;
+
+    // -----
     // This function provides the processing needed when a country has been changed.  It makes
     // use of the country_zones (countries-to-zones) array, built above.  Normally invoked
     // by the template's 'onchange=update_form(this.form)' parameter applied to the countries'
@@ -114,6 +122,16 @@ jQuery(document).ready(function() {
             jQuery('#stateZone').hide();
             jQuery('#stateZone').next('span.alert').hide();
         }
+<?php
+    // -----
+    // Issue a notification to enable site-specific jQuery/javascript to be inserted to provide
+    // additional processing when the base script is run, indicating a change in selected
+    // country.
+    //
+    $extra_javascript = '';
+    $GLOBALS['zco_notifier']->notify('NOTIFY_ZEN_ADDR_PULLDOWNS_UPDATE', [], $extra_javascript);
+    echo $extra_javascript;
+?>
     }
 });
 </script>
--- a/includes/templates/template_default/jscript/zen_addr_pulldowns.php
+++ b/includes/templates/template_default/jscript/zen_addr_pulldowns.php
@@ -11,47 +11,46 @@
  * @version $Id: Zen4All 2019 Jun 03 Modified in v1.5.7 $
  */
 // -----
-// If the site does NOT require a 'State' entry in an address or has NOT enabled states to
-// be displayed as dropdowns, nothing to be done here!
+// If the site does NOT require a 'State' entry in an address, nothing to be done here!
 //
-if (ACCOUNT_STATE !== 'true' || ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN !== 'true') {
+if (ACCOUNT_STATE !== 'true') {
     return;
 }
 
 // -----
-// If the current site has at least one country enabled that uses zones, a JSON-encoded array of
-// countries-to-zones will be created for use by the jQuery section.
+// If the current site is using the state dropdowns, create the array that identifies the various
+// zones for the currently-active countries.  When state dropdowns aren't being used, the
+// empty $c2z array will 'instruct' the jQuery section to presume that no countries have
+// associated zones.
 //
-$countries = $db->Execute(
-    "SELECT DISTINCT zone_country_id
-       FROM " . TABLE_ZONES . "
-            INNER JOIN " . TABLE_COUNTRIES . "
-                ON countries_id = zone_country_id
-               AND status = 1
-   ORDER BY zone_country_id"
-);
-
 $c2z = [];
-foreach ($countries as $next_country) {
-    $current_country_id = $next_country['zone_country_id'];
-    $c2z[$current_country_id] = [];
-
-    $states = $db->Execute(
-        "SELECT zone_name, zone_id
+if (ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN === 'true') {
+    // -----
+    // If the current site has at least one country enabled that uses zones, a JSON-encoded array of
+    // countries-to-zones will be created for use by the jQuery section.
+    //
+    $countries = $db->Execute(
+        "SELECT DISTINCT zone_country_id
            FROM " . TABLE_ZONES . "
-          WHERE zone_country_id = $current_country_id
-       ORDER BY zone_name"
+                INNER JOIN " . TABLE_COUNTRIES . "
+                    ON countries_id = zone_country_id
+                   AND status = 1
+       ORDER BY zone_country_id"
     );
-    foreach ($states as $next_state) {
-        $c2z[$current_country_id][$next_state['zone_id']] = $next_state['zone_name'];
-    }
-}
+    foreach ($countries as $next_country) {
+        $current_country_id = $next_country['zone_country_id'];
+        $c2z[$current_country_id] = [];
 
-// -----
-// If none of the currently-enabled countries use zones, nothing to do here.
-//
-if (count($c2z) === 0) {
-    return;
+        $states = $db->Execute(
+            "SELECT zone_name, zone_id
+               FROM " . TABLE_ZONES . "
+              WHERE zone_country_id = $current_country_id
+           ORDER BY zone_name"
+        );
+        foreach ($states as $next_state) {
+            $c2z[$current_country_id][$next_state['zone_id']] = $next_state['zone_name'];
+        }
+    }
 }
 ?>
 <script>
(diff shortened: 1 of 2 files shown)


torvista:
State - Always display as pulldown? - false
Shipping Estimator as 1-Display as Button on Shopping Cart,  
the empty stateLabel takes up space:
stateLabel

torvista (commented):
jQuery.parseJSON is deprecated from 3.0 onwards

lat9:
That updated commit now uses JSON.parse instead of the deprecated jQuery.parseJSON, corrects the unwanted inclusion of the empty #stateLabel and adds the previously-missing pull-in of the jQuery on the shopping_cart page for the shipping-estimator.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
```diff
@@ -11,47 +11,46 @@
...
+    foreach ($countries as $next_country) {
+        $current_country_id = $next_country['zone_country_id'];
+        $c2z[$current_country_id] = [];
 
-// -----
-// If none of the currently-enabled countries use zones, nothing to do here.
-//
-if (count($c2z) === 0) {
-    return;
+        $states = $db->Execute(
```
proseLA, line 44: this section of code looks very much like the function, `zen_get_country_zones`.  could we not use that function here as opposed to replicating code?  in addition, i am of the belief that we should always include the `zone_code` whenever we are selecting the `zone_id` from this table.  if for no other reason than to appease developers like myself that want to use said field in future development and do not have to jump through hoops to get it.

proseLA (commented):
i am a fan of consolidating code.  it seems we can do more here.

torvista:
looks good on the shipping estimator, thanks.

html validator moans about the unescaped forward slashes on the two  </option>

Clipboard02

torvista:
On a tangent....there are two places where I would want to run additional JavaScript.
a) after the code sets up the fields on first load
b) after the code updates the fields with the change of country,

My code puts country-specific restrictions and real-time checks on the postcode field, with specific error messages per country.
Maybe this is something others may consider in the future.
With this in mind, how would that best be implemented to allow additional code to be included/avoid changing this core file?
Add Notifiers? or a check for the existence of a specifically-named file in the template to include?

While of course another js file can be included in parallel, that gets in a bit of a fight accessing the same fields as the core file, so would be better run in series included into this one.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
lat9: Even though it goes against my "performance-grain", I'm amenable to making the change to use `zen_get_country_zones`.

I'll grant that using the `zone_code` would be preferable, but that wouldn't be downward-compatible with the base Zen Cart templates or any add-on templates in existence.

lat9:
> On a tangent....there are two places where I would want to run additional JavaScript. a) after the code sets up the fields on first load b) after the code updates the fields with the change of country,
> 
> My code puts country-specific restrictions and real-time checks on the postcode field, with specific error messages per country. Maybe this is something others may consider in the future. With this in mind, how would that best be implemented to allow additional code to be included/avoid changing this core file? Add Notifiers? or a check for the existence of a specifically-named file in the template to include?
> 
> While of course another js file can be included in parallel, that gets in a bit of a fight accessing the same fields as the core file, so would be better run in series included into this one.

The `zen_addr_pulldowns.php` is pulled in via various pages' `jscript_addr_pulldowns.php` on a template-specific basis
```php
require $template->get_template_dir('zen_addr_pulldowns.php', DIR_WS_TEMPLATE, $current_page_base, 'jscript') . '/zen_addr_pulldowns.php';
```

You're free to make a template-override in your `/includes/templates/YOUR_TEMPLATE/jscript/zen_addr_pulldowns.php` to create a site-specific version.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
proseLA: i'm not sure i understand your "performance-grain" comment.

in addition, i think i was perhaps mis-understood.  i understand your concerns about use of `zone-code`.  my comment was that the function, `zen_get_country_zones` should also the return the `zone_code` in the return array; despite it not being used here.  it would not affect anything done here.  i can do a PR for said change, if you are amenable.

lat9: I was just referring to the double iteration of the zones' information.  Having `zen_get_country_zones` return the `zone_code` is a totally different 'animal' and should be addressed in a separate issue.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
```diff
@@ -11,58 +11,78 @@
...
+        }
     }
 }
-
-if (count($c2z) !== 0) {
+?>
+<script>
+jQuery(document).ready(function() {
+<?php
     echo '    var country_zones = \'' . addslashes(json_encode($c2z)) . '\';' . PHP_EOL;
```
torvista, line 54: Since the creation of "country_zones" is in php and "hidden" from the IDE, it moans about unresolved variable.
Is there some reason why this is better in php, than javascript with embedded php?

includes/templates/template_default/jscript/zen_addr_pulldowns.php
```diff
@@ -11,58 +11,78 @@
...
+        jQuery('#state').hide();
         jQuery('#stateZone').show();
+        jQuery('#stateZone').next('span.alert').show();
+    } else {
+        jQuery('#state').show();
+        jQuery('#stateZone').hide();
+        jQuery('#stateZone').next('span.alert').hide();
     }
 
     var pleaseSelect = '<?php echo PLEASE_SELECT; ?>';
```
torvista, line 88: This could be a constant.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
```diff
@@ -11,58 +11,78 @@
...
+// 3. Initialize the display for the dropdown vs. hand-entry of the state fields.  If the initially-selected
+// country doesn't have zones, the dropdown will contain only 1 element ('Type a choice below ...').  In that
+// case, the dropdown and associated elements will be hidden and the hand-input 'state' field will be shown.
+//
+// 4. There can be unwanted whitespace, e.g. an &nbsp; prior to the (optional) <span class="alert"> following
+// the 'stateZone' dropdown.  In that case, when the <span> is hidden for unzoned countries, the state input
+// field is slightly offset from the other input fields.
 //
 ?>
-    if (jQuery('#stateZone > option').length == 1) {
```
torvista, line 60: The two .length == could be strict comparisons.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
```diff
@@ -77,23 +97,25 @@
     update_zone = function(theForm)
     {
         var countryHasZones = false;
-        var countryZones = '<option selected="selected" value="0">' + pleaseSelect + '</option>';
+        var countryZones = '<option selected="selected" value="0">' + pleaseSelect + '<' + '/option>';
         var selected_country = jQuery('#country option:selected').val();
-        jQuery.each(jQuery.parseJSON(country_zones), function(country_id, country_zones) {
+        jQuery.each(JSON.parse(country_zones), function(country_id, country_zones) {
             if (selected_country == country_id) {
```
torvista, line 103: Could be strict comparison.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
lat9: You mean a script-based one?  How important is that?

includes/templates/template_default/jscript/zen_addr_pulldowns.php
lat9: Nope, the original implementation had either an empty string or the `var` statement.  Now that the `var` is always set &mdash; just in case a site has no countries with zones &mdash; that change is possible.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
torvista: From what I read, not important in any real sense. Just semantically correct.  
Does it even need to be created at all/could be inline, since it's only used once?

includes/templates/template_default/jscript/zen_addr_pulldowns.php
```diff
@@ -11,58 +11,77 @@
...
+<script>
+jQuery(document).ready(function() {
+    var country_zones = '<?php echo addslashes(json_encode($c2z)); ?>';
+<?php
 // -----
-// Initialize the display for the dropdown vs. hand-entry of the state fields.  If the initially-selected
-// country doesn't have zones, the dropdown will contain only 1 element ('Type a choice below ...').
+// Notes:
+//
+// 1. The '#stBreak' <br> is also never needed/wanted since it will 'throw' the state-field input underneath
```
torvista, line 58: If '#stBreak' always should be hidden, why not remove it from the template file?

lat9: It's `legacy` with existing add-on templates already using it.  Keeping it documented will give the base the opportunity to remove it as 'best practices' in the future.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
lat9: Variable now moved inline, using `addslashes` ... just in case to ensure valid javascript created.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
lat9: Noting, too, that I've changed this to be a `const` instead of `var`.

includes/templates/template_default/jscript/zen_addr_pulldowns.php
lat9: See separate issue created, as identified below.

lat9:
@torvista, do those notifications 'do the trick' for your request?

torvista:
I appreciate the consideration, but while trying to accomplish these days what I already had working/ do a revamp using this, I'm having difficulties with trying to work with this original intact.
I have arrived at the conclusion that as you previously highlighted, this file is indeed best replaced completely in the template rather than trying to add into it and work with/partially override what is already happening. Since a custom version is not core and would not get overwritten on an clumsy upgrade, I don't think it meets the criteria for having notifiers. 
Its a small file and it is already serving it's primary function of consolidation, which is the point of the PR.
Something does not seem right in having notifiers in a file that is already an override.
So, not wanting to be accused of foisting my follies into core, I think you should leave the notifiers out.

lat9:
@torvista, I appreciate the feedback since I kind of cringed when adding notifiers to add javascript; I'll pull that out and then (hopefully) this PR will find its 'forever home' in the Zen Cart base.

scottcwilson:
Thank you team for the hard work on this PR.  I'll get it merged in the next couple of days.

PR #4806, opened 2022-05-18
Remove unused variable

scottcwilson:


lat9:
A high-level description of the problem(s)-to-be-solved would be appreciated.

PR #4807, opened 2022-05-19
shipping estimator template, change deprecated name to id

torvista:
as referenced by the form action
`<?php echo zen_draw_form('estimator', zen_href_link($show_in . '#view', '', $request_type), 'post'); ?>`

### DIFF
--- a/includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
+++ b/includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
@@ -9,7 +9,7 @@
  */
 ?>
 <div id="shippingEstimatorContent">
-<?php echo zen_draw_form('estimator', zen_href_link($show_in . '#view', '', $request_type), 'post'); ?>
+<?php echo zen_draw_form('estimator', zen_href_link($show_in . '#seView', '', $request_type), 'post'); ?>
 <?php if (is_array($selected_shipping)) {
     zen_draw_hidden_field('scid', $selected_shipping['id']);
 } ?>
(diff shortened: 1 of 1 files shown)


lat9 (commented):
Since we're changing it from a `name=`, how about using a relatively unique name so that there's the potential of reducing conflicts ... like if a template already has an `id="view"`.

Something not too wordy, like `id="zc-se-view"`?

proseLA:
> Since we're changing it from a `name=`, how about using a relatively unique name so that there's the potential of reducing conflicts ... like if a template already has an `id="view"`.
> 
> Something not too wordy, like `id="zc-se-view"`?

great minds think alike.  i was thinking then name could be `shippingEstimatorView`.  but anything other than the generic term `view`.  

do not forget that when the name gets changed, it needs to be changed here as well:

https://github.com/zencart/zencart/blob/39f848ccaf0ad86dd0385e6dd14899b1c7f53529/includes/templates/template_default/templates/tpl_modules_shipping_estimator.php#L12

torvista:
shippingEstimatorView gets my vote.

lat9:
Works for me, although I _thought_ that we were heading towards `lower-case-with-minus-signs` type names.  As long as it's fairly unique.

torvista:
>  I thought that we were heading towards lower-case-with-minus-signs type id's

I have no idea, is that documented anywhere?

scottcwilson:
> I have no idea, is that documented anywhere?

Nope.  There's a mix.  Try to stay consistent with the type already used in the file (which is hopefully one or the other and not both).

scottcwilson:
Spec reference:
https://www.w3.org/TR/2011/WD-html-markup-20110525/a.html
("name" on anchor is obsolete).

Thanks for doing this @torvista

PR #4808, opened 2022-05-20
consolidate all jscript form checks into single file

proseLA:
one good refactor deserves another....

the javascript in all of these files is the same.


scottcwilson:
Consolidation FTW!  Thanks for doing this.

PR #4809, opened 2022-05-20
tpl_modules_shipping_estimator.php: remove unused $extra

torvista:
I was correcting for the possible empty value, but then I don't find $extra created anywhere.


lat9 (commented):
Hmm, that's been there for a "dog's age" but it does give a site the option to configure any quote errors' display.  It's not doing any harm, so my suggestion is either to comment its potential use or leave it alone.

torvista:
How does it get used?

lat9:
A site could have a template-override version of `/includes/modules/shipping_estimator.php` that sets that `$extra` class to indicate the CSS class for the styling of those rows if a quote-related error is found.

PR #4811, opened 2022-05-22
Disable Continue Button if Terms not agreed

brittainmark:
Disable the continue button if terms and conditions are required and have
not been checked.
Added labels to continue button and the  tick box for terms and condition.
Added JavaScript to disable and re-enable the continue button when terms and conditions box is checked.
Added change of colour to CSS for disabled button (grey).

### DIFF
--- a/admin/sqlpatch.php
+++ b/admin/sqlpatch.php
@@ -84,6 +84,11 @@ function executeSql($lines, $database, $table_prefix = '') {
      $no_semi_line = rtrim($line, ';');
      $param = preg_split('/\s+/', $no_semi_line);
 
+     // Preset entries 0..5 if not set
+     for ($i = 0; $i <= 13; $i++) {
+        if (!isset($param[$i])) $param[$i] = ''; 
+     }
+
     // The following command checks to see if we're asking for a block of commands to be run at once.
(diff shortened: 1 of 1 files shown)


lat9 (commented):
Rather than 'burdening' the responsibility on the template, since site-specific templates are very common, this functionality should, IMO, be moved to `/includes/modules/pages/checkout_payment/jscript_main.php`.

The page's submit button is already _presumed_ to be wrapped in a `#paymentSubmit` block, so it can be located by the javascript/jQuery portion that enables and disables the submit button.

The conditions' checkbox is identified by its _presumed_ `#conditions` id, so the javascript could us an `on('change')` mechanism to bind to the checkbox's change event.

brittainmark:
Thanks for the feed back @lat9 . I'll rewrite. Should I include code for one page check out like @proseLA has done with SquareWebPay?

lat9:
> Thanks for the feed back @lat9 . I'll rewrite. Should I include code for one page check out like @proseLA has done with SquareWebPay?

I've got the associated change request noted on the OPC GitHub repository; thanks @brittainmark

brittainmark:
replaced by #4833

PR #4814, opened 2022-05-24
cast selected_country to integer

torvista:
When the shipping estimator is first loaded, $selected_country is a string from

https://github.com/zencart/zencart/blob/14d292a970bd5c584636aabe7499f52f98723fd7/includes/modules/shipping_estimator.php#L159

but subsequently, after updating, $selected_country is an integer, from here

https://github.com/zencart/zencart/blob/14d292a970bd5c584636aabe7499f52f98723fd7/includes/modules/shipping_estimator.php#L19

This PR casts the first one to integer too.

### DIFF
--- a/includes/modules/shipping_estimator.php
+++ b/includes/modules/shipping_estimator.php
@@ -156,7 +156,7 @@ if ($_SESSION['cart']->count_contents() > 0) {
     $quotes = $shipping_modules->quote();
 
     // set selections for displaying
-    $selected_country = $order->delivery['country']['id'];
+    $selected_country = (int)$order->delivery['country']['id'];
     $selected_address = $sendto;
     // eo shipping cost
     // check free shipping based on order $total


lat9:
Please identify the problem that this PR corrects.

torvista:
custom code that checks for this variable using a **strict** comparison will fail on one of the events.

lat9:
FWIW, I've started a discussion (https://github.com/zencart/zencart/discussions/4824) to determine whether $_GET/$_POST variables should be sanitized going forward.

scottcwilson:
@lat9 Are you OK with merging this PR?

lat9:
> @lat9 Are you OK with merging this PR?

If it's not affecting core processing, which it sounds like it's not, then I'm not in favor.  I realize that there are other places where the $_GET/$_POST variables are sanitized-in-place via casting, but I believe that we should be heading in the direction of keeping those variables in their `string` representation.

scottcwilson:
What if we say "Keep them as strings *unless* their name ends with '_id'?"  Would that work? Because other than product_id, fields with that name pattern are integer, AFAIK.

lat9:
> What if we say "Keep them as strings _unless_ their name ends with '_id'?" Would that work? Because other than product_id, fields with that name pattern are integer, AFAIK.

... and the reason?  It just feels that once you start with one exception, it's a downward spiral.  Noting also that `products_id` is used more often than `product_id`, so that doubles the number of exceptions!

image

scottcwilson:
OK.  Closing until #4824 determines the direction for handling these variables.

PR #4818, opened 2022-05-25
Remove old Square files

### LINKED ISSUE #4725: Remove old version of Square from 1.5.8
The existing Square implementation is being sunsetted and should be removed from the build.

scottcwilson:
Fixes #4725


proseLA:
you have deleted a file that square_webPay uses.  see:

https://github.com/proseLA/square_webPay/blob/e92cd2e73e9e2254a0d306ad0713857b0bcc0476/for_upload/includes/modules/payment/square_webPay.php#L508

scottcwilson:
Would be better to add this file to the square_webPay repo than to keep it in the core?

scottcwilson:
No worries, will add back.

scottcwilson:
Let's leave this for now - Square was not as fast as we thought they would be in truly removing this functionality.

PR #4820, opened 2022-05-26
alert_page: allow developer overrides to bypass checks

torvista:
Regarding these constants which may optionally be defined for developer use.
```
define('ADMIN_BLOCK_WARNING_OVERRIDE', '1'); //bypasses checks in (admin) init_admin_auth to allow admin access when a) the admin folder has not been renamed and b) installer folder /zc_install exists.
define('WARN_INSTALL_EXISTENCE', 'false'); //bypasses check in (admin) init_errors: prevents a messageStack warning in the admin if the installer folder /zc_install exists.
```

This PR allows the alert page to be refreshed and continue to admin when these constants are defined as shown.

### DIFF
--- a/admin/alert_page.php
+++ b/admin/alert_page.php
@@ -6,12 +6,16 @@
  */
 require('includes/application_top.php');
 $adminDirectoryExists = $installDirectoryExists = false;
-if (substr(DIR_WS_ADMIN, -7) == '/admin/' || substr(DIR_WS_HTTPS_ADMIN, -7) == '/admin/') {
-    $adminDirectoryExists = true;
-}
-$check_path = dirname($_SERVER['SCRIPT_FILENAME']) . '/../zc_install';
-if (is_dir($check_path)) {
-    $installDirectoryExists = true;
+if (!(defined('ADMIN_BLOCK_WARNING_OVERRIDE') && ADMIN_BLOCK_WARNING_OVERRIDE === '1')) {
+    if (substr(DIR_WS_ADMIN, -7) === '/admin/' || substr(DIR_WS_HTTPS_ADMIN, -7) === '/admin/') {
+        $adminDirectoryExists = true;
+    }
+    if (!(defined('WARN_INSTALL_EXISTENCE') && WARN_INSTALL_EXISTENCE === 'false')) {
+        $check_path = dirname($_SERVER['SCRIPT_FILENAME']) . '/../zc_install';
+        if (is_dir($check_path)) {
+            $installDirectoryExists = true;
+        }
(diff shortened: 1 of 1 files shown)


proseLA:
i am not a fan.  not only is the code more confusing, but what is the use case scenario?

currently, my dev box never goes to the alert_page; and my admin is defined as `admin` and my install is `zc_install`.  so i have both dev constants properly defined.  

right now, if i go directly to my alert page, ie:

https://zcdev.com/admin/alert_page.php

i can get there.  

but, if this PR is merged, and if i was a developer, translating ZC into some foreign language, ie gaelic, i would now have to change my constants or override the code to get to the same page and examine the formatting of said page.

so again, what is the use case for this PR?

scottcwilson:
Perhaps an alternate approach - add a link to admin home on the alert page so that after you fix the issue (by removing zc_install or using one of the constants), you could just click the link.

torvista:
The not-really-a-problem, but more of smoothing the "experience" is just as Scott said.
When reinstalling ZC for whatever reason and coming up with alert page, it just seems smoother to fix what is wrong (whether renaming or creating the overrides") then refesh the same page  and see the problem gone, rather than manually editing the address to get back to admin login
Very minor use case, but so what. Small steps to improve the user experience.

drbyte:
> Perhaps an alternate approach - add a link to admin home on the alert page so that after you fix the issue (by removing zc_install or using one of the constants), you could just click the link.

> ... it's aggravating to have to edit the URL to get back to the home page. A "go to home" link would be like the down for maintenance page where you can re-check the site status just by clicking on a link.

Unlike the storefront, the admin URL isn't "known" by the code, so adding a link is somewhat more complicated.

drbyte:
Review consideration: where is the existing code that checks these constants? How is this different? same?

scottcwilson:
> Unlike the storefront, the admin URL isn't "known" by the code, so adding a link is somewhat more complicated.

D-ohhh!  Of course you are correct.  So much for that idea. :(

scottcwilson:
The spirit of this PR is good but it might be more trouble than it's worth.  Not strong one way or the other.

torvista:
Firstly Chris, welcome back.

> Review consideration: where is the existing code that checks these constants? How is this different? same?

WARN_INSTALL_EXISTENCE bypasses one check in init_errors.
ADMIN_BLOCK_WARNING_OVERRIDE bypasses two checks in init_admin_auth
This PR mirrors those same effects.

scottcwilson:
Closing - not enough support to move this forward.

PR #4821, opened 2022-05-26
v158: Correcting account-history unformatted totals

### LINKED ISSUE #4813: Account Order History: order total is unformatted
ZC158
Clipboard01
Previously forgotten here:
https://github.com/zencart/zencart/pull/4196#issuecomment-770245502

lat9:
Fixes #4813.

- Moving the formatting into the `Customers.php` class to address the changes required for both #4196 and #4813.
- Created a separate array element returned by the class' `getOrderHistory` method to also return the 'raw' value of each order's total.
- Removing the template-specific formatting introduced in #4813.

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -487,7 +487,7 @@ class Customer extends base
     public function getOrderHistory(int $max_number_to_return = 0): array
     {
         $language = $_SESSION['languages_id'];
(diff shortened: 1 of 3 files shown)


lat9:
Noting that this change also provides compatibility with zc157 (and prior) templates' rendering.

PR #4823, opened 2022-05-26
Fix "Undefined $current_page_base" notice in PHP8.0+

scottcwilson:


lat9:
Could you include a copy of a generated log?

lat9:
> > Could you include a copy of a generated log?
> 
> look at line 211:
> 
> image

I saw that, but my belief is that a copy of the generated log gives other developers a better 'clue' as to whether the change will also correct any issue they're seeing.

scottcwilson:
It's a good practice to include the log - sorry I missed that this time. 
Of course, I can't dup it right now.  Paul can you repro?

proseLA:
with all due respect, i find the need to justify and comment on (granted what i deem to be) a simple fix of an apparent bug is perhaps a bit tiresome.  

> It's a good practice to include the log - sorry I missed that this time. Of course, I can't dup it right now. Paul can you repro?

i am not going to even try...

either one can get to said line or not...  i trust that you got there and this fixed it.  else why would you have gotten here?

PR #4829, opened 2022-05-27
Fix for (some of) PHP Notice for malformed shipping 

scottcwilson:
PHP Notice:  Malformed value for session-based shipping module; customer will need to re-select: false in includes/classes/order.php on line 327

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -396,7 +396,7 @@ class order extends base
         //
         // If that's not the case, issue a PHP Notice and reset the shipping to its unselected state.
         //
-        if (isset($_SESSION['shipping'])) {
+        if (!empty($_SESSION['shipping'])) {
             if (!empty($_SESSION['shipping']['id']) && strpos((string)$_SESSION['shipping']['id'], '_')) {
                 $shipping_module_code = $_SESSION['shipping']['id'];
             } else {


lat9 (commented):
Isn't `!empty()` good enough?  I'm guessing that `$_SESSION['shipping']` is `(bool)false` as corrected for the 'freeoptions' shipping.

scottcwilson:
`!empty()` would work too yes.

drbyte:
Question: "why" does this "fix" "work"?

scottcwilson:
In the log shown in the first post, the value after the colon (false) is the value that's causing the problem (at least in most of the cases I have seen). Checking !empty() will skip over false values.

lat9:
Is, by chance, the `freeoptions` method enabled?  That had some issues (previously PR'd) that would result in that `false` quote being returned.

scottcwilson:
I believe the store did have freeoptions enabled.  But also any shipping module that was cloned from freeoptions would cause the same problem. 

I have simplified the PR as you suggested.  I think we should merge this as an anti-bugging change and move on.

scottcwilson:
This will give 157 users access to an easy fix for this annoying problem.

njcyx6649250:
My site (v1.57d) doesn't offer free shipping option, but sometimes (once per a few months) I will receive the same warning. Not sure how this bug is triggered. I just updated the order.php file and see how it works.

Thanks for your help.

njcyx6649250:
Update: Unfortunately, I still received another debug log. This log is triggered by the table.php. I'm using the table.php based on @version $Id: DrByte 2020 Sep 29 Modified in v1.5.7a.

[12-Jul-2022 11:12:32 America/New_York] PHP Notice: Malformed value for session-based shipping module; customer will need to re-select: {"id":"","title":"Table Rate (Best Way)","cost":76.9899999999999948840923025272786617279052734375} in /public_html/includes/classes/order.php on line 324

PR #4831, opened 2022-05-28
v158: Resurrecting PayFlow-UK processing

### LINKED ISSUE #4830: Resurrect PayFlow-UK support in paypalwpp
The PayFlow-UK processing in `/includes/modules/payment/paypalwpp.php` has been broken since zc155.  The NVP variables used by the PayFlow processing are using the old, unprefixed, variable names instead of those prefixed by either `PAYMENTINFO_0` or `PAYMENTREQUEST_0_`.

Noting also that
1. For 3DS versions of PayFlow, it's very important that a site is running PHP 7.3 or later with a `samesite cookie` setting of **none** (the default is _lax_).
2. The 3dauth/iframe processing in `paypaldp` also does not support the `collectsCardDataOnsite` AJAX interface, so that variable will need to be set to `false` if a site is currently running in PayFlow-UK mode.

lat9:
Fixes #4830.

### DIFF
--- a/includes/modules/payment/paypaldp.php
+++ b/includes/modules/payment/paypaldp.php
@@ -120,7 +120,7 @@ class paypaldp extends base {
   var $fmfResponse = '';
   var $fmfErrors = array();
   /**
-   * this module collects card-info onsite
+   * this module collects card-info onsite, unless operating in PayFlow-UK mode!
    */
   var $collectsCardDataOnsite = TRUE;
   /**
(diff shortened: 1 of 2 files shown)


scottcwilson:
Thanks for doing this.  Has this been tested live on a UK client's site?

lat9:
It has, on a site running zc157c.  I've merged the changes into the zc158 version of the file.

PR #4832, opened 2022-05-28
Change color for icon when Quantity Discounts enabled

scottcwilson:
Right now when viewing Admin > Catalog > Categories/Products, you cannot tell at a glance if a product has quantity discounts. If a product may be added to the cart, the icon is green.
product_icons_old

This change makes the icon dark green if quantity discounts are present. 
product_icons_new


admin/category_product_listing.php
```diff
@@ -939,7 +939,13 @@
                         <a href="<?php echo zen_href_link(FILENAME_ATTRIBUTES_CONTROLLER, 'products_filter=' . $product['products_id'] . '&current_category_id=' . $current_category_id); ?>" class="btn btn-sm btn-default btn-attributes-off" role="button" title="<?php echo BOX_CATALOG_CATEGORIES_ATTRIBUTES_CONTROLLER; ?>"><strong>A</strong></a>
                       <?php } ?>
                       <?php if ($zc_products->get_allow_add_to_cart($product['products_id']) == "Y") { ?>
-                        <a href="<?php echo zen_href_link(FILENAME_PRODUCTS_PRICE_MANAGER, 'products_filter=' . $product['products_id'] . '&current_category_id=' . $current_category_id); ?>" class="btn btn-sm btn-default btn-pricemanager-on" role="button" title="<?php echo BOX_CATALOG_PRODUCTS_PRICE_MANAGER; ?>">
+<?php 
+                     $ppm_color = 'btn-pricemanager-on'; 
+                     if (zen_has_product_discounts($product['products_id']) == 'true') {
```
lat9, line 944: Let's keep the trend of using exactly-equal-to where the value(s) are known.

PR #4833, opened 2022-05-29
Disable continue if T&C not present

brittainmark:
Javascript to disable the continue button on the checkout payment page.
CSS changes to make button grey when disabled.


brittainmark:
Replaces #4811

scottcwilson:
Thanks for doing this @brittainmark.

PR #4835, opened 2022-05-29
Fix multiplication issue in Customers class

scottcwilson:
In PHP 8.1:
[29-May-2022 14:09:09 UTC] Request URI: /gh_demo_158/admin/index.php?cmd=customers, IP address: ::1
--> PHP Fatal error: Uncaught TypeError: Unsupported operand types: string * string in /Users/scott/sites/gh_demo_158/includes/classes/Customer.php:313

In PHP 8.0: 
--> PHP Warning: A non-numeric value encountered in /Users/scott/sites/gh_demo_158/includes/classes/Customer.php on line 313.

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -310,7 +310,7 @@ class Customer extends base
                     'language_code' => $result['language_code'],
                 ];
             }
-            $lifetime_value += ($result['order_total'] * $result['currency_value']);
+            $lifetime_value += ((float)$result['order_total'] * (float)$result['currency_value']);
         }
         $this->data['last_order'] = $last_order;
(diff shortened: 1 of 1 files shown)


lat9:
What was the 'non-numeric value' that resulted in the error/warning log?

lat9:
> What was the 'non-numeric value' that resulted in the error/warning log?

Looking at the code and the `orders` table schema, I'm guessing that one (or both) of the `order_total` and `currency_value` were defaulted to `NULL`.  Perhaps a better solution (float-casting is 'evil') would be to set both values to 0 during the loop preceding the calculation if found to be `NULL`.

PR #4837, opened 2022-05-30
v158: Adding `zen_define_default`.

### LINKED ISSUE #4825: New function to set a define's default if not already defined
this section of code is something new in v158 in that we are checking if the config key is already defined:

https://github.com/zencart/zencart/blob/5ab07980bdd846ce174ad976456686a339ecf1f5/laravel/app/Models/Configuration.php#L29-L31

looking in v157, we see we do not have that check:

https://github.com/zencart/zencart/blob/742b837868a2b128df96b6635c413c8187972679/includes/init_includes/init_db_config_read.php#L23-L27

in seeing all of the php 8.0 notices for already defined constants, i think it would behoove us to provide an early loading function to do this.  ie, something like:

```
        function defineIfNot(string $name, string $definition)
        {
            if (!defined($name)) {
                define($name, $definition);
            }
        }
```
i see potential for plenty of uses as store owners move to php 8.0 and are inundated with log files.

thoughts?

noted, that i have the declared variable arguments as both strings and in some instances integer is required.  also noting that i WOULD love to refactor all of those boolean strings into actual booleans... and noting that will never happen.  a coder can dream...

lat9:
Fixes #4825

### DIFF
--- a/admin/includes/application_bootstrap.php
+++ b/admin/includes/application_bootstrap.php
@@ -112,6 +112,7 @@ if (file_exists('includes/defined_paths.php')) {
     exit;
 }
 require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'php_polyfills.php';
+require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'zen_define_default.php';
 /**
  * ignore version-check if INI file setting has been set
  */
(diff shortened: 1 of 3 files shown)


scottcwilson:
@lat9 @proseLA Would you please write up a trivial use case for me to show in the documentation?

lat9:
> @lat9 @proseLA Would you please write up a trivial use case for me to show in the documentation?

It's used in /admin/application_bootstrap.php on line#145 to set the default definition of `DIR_WS_TEMPLATES` to `DIR_WS_INCLUDES . 'templates/'`:
```
zen_define_default('DIR_WS_TEMPLATES', DIR_WS_INCLUDES . 'templates/');
```

lat9:
It's just a readability 'helper', a one-liner function instead of
```
if (!defined('DIR_WS_TEMPLATES')) {
    define('DIR_WS_TEMPLATES', DIR_WS_INCLUDES . 'templates/');
}
```

scottcwilson:
OK it looks fine but why a new function file? Wouldn't it fit into an existing file?

PR #4839, opened 2022-06-01
Check second token of name

scottcwilson:
In case you're sending a GV to Cher.  Or Bono.


scottcwilson:
Fixed log: 

```
--> PHP Notice: Undefined offset: 1 in /home/thatsoft3/public_html/site/admin/gv_mail.php on line ... 
```

PR #4850, opened 2022-06-05
Syncronise and stream line packing slip and invoice notifiers

### LINKED ISSUE #4846: Packingslip has lost Notifiers
**Describe the bug**
The Notifiers: 
NOTIFY_ADMIN_PACKINGSLIP_HEADING
NOTIFY_ADMIN_PACKINGSLIP_SORT_DISPLAY
NOTIFY_ADMIN_PACKINGSLIP_DATA
are missing from the admin>packingslip.php
**Version**
Zen Cart version: 1.5.8
 I will recreate and send PR.

**Additionally,** while I am there I noticed that if $attr_img_width was added to the  site_specific_overrides for invoice.php and Packingslip.php then the listeners  NOTIFY_ADMIN_PACKINGSLIP_DEFAULTS and NOTIFY_ADMIN_INVOICE_DEFAULTS could be removed. 
If you want to keep it consistent, you could change the name $attr_img_width to $show_attr_img_width, but that does not sound correct, 
Would prefer to change the documentation to say that this value can be overridden.

brittainmark:
Added notifiers for additional columns to packing slip.
Removed change default values notifiers to use site_specific_admin_overrides
Added $attr_img_width to site_specific_admin_overrides
Removed training ?> from invoice and packing slip
Fixes #4846

### DIFF
--- a/admin/invoice.php
+++ b/admin/invoice.php
@@ -6,16 +6,18 @@
  * @version $Id: DrByte 2020 May 20 Modified in v1.5.7 $
  */
 require('includes/application_top.php');
-// To override the $show_* values, see 
+// To override the $show_* or $attr_img_width values, see 
 // https://docs.zen-cart.com/user/admin/site_specific_overrides/
-if (!isset($show_product_images)) $show_product_images = true;
-if (!isset($show_attrib_images)) $show_attrib_images = true;
+if (!isset($show_product_images)) {
+    $show_product_images = true;
+}
+if (!isset($show_attrib_images)) {
+    $show_attrib_images = true;
+}
 $img_width = defined('IMAGE_ON_INVOICE_IMAGE_WIDTH') ? (int)IMAGE_ON_INVOICE_IMAGE_WIDTH : '100';
-$attr_img_width = '25';
-/*
- * Notifier to allow defaults to be changed
- */
-$zco_notifier->notify('NOTIFY_ADMIN_INVOICE_DEFAULTS', '', $show_product_images, $show_attrib_images, $attr_img_width);
+if (!isset($attr_img_width)) {
+    $attr_img_width = '25';
+}
 
 require(DIR_WS_CLASSES . 'currencies.php');
 $currencies = new currencies();
@@ -25,7 +27,9 @@ $oID = zen_db_prepare_input($_GET['oID']);
 include DIR_FS_CATALOG . DIR_WS_CLASSES . 'order.php';
 $order = new order($oID);
 $show_including_tax = (DISPLAY_PRICE_WITH_TAX == 'true');
-if (!isset($show_product_tax)) $show_product_tax = true;
+if (!isset($show_product_tax)) {
+    $show_product_tax = true;
+}
 
 // prepare order-status pulldown list
 $orders_statuses = array();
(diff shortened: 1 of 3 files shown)


admin/packingslip.php
```diff
@@ -138,11 +138,54 @@
...
         <tbody>
             <?php
-            for ($i = 0, $n = count($order->products); $i < $n; $i++) {
+            /*
+             * Notifier to allow packing slip to be sorted to required order
+             *
+
+             *              * Set $sort_order to the order->products array counter in the sequence you require the invoice to be displayed
+             */
+            $sort_order=false;
```
proseLA, line 181: https://www.php-fig.org/psr/psr-12/#:~:text=6.2.%20Binary%20operators,6.3.%20Ternary%20operators

proseLA (commented):
are we going to make any attempt to follow an accepted coding style such as the ones linked here?  i know there was something at one point in the contributions information.

admin/invoice.php
```diff
@@ -6,16 +6,12 @@
...
 // https://docs.zen-cart.com/user/admin/site_specific_overrides/
 if (!isset($show_product_images)) $show_product_images = true;
 if (!isset($show_attrib_images)) $show_attrib_images = true;
 $img_width = defined('IMAGE_ON_INVOICE_IMAGE_WIDTH') ? (int)IMAGE_ON_INVOICE_IMAGE_WIDTH : '100';
-$attr_img_width = '25';
-/*
- * Notifier to allow defaults to be changed
- */
-$zco_notifier->notify('NOTIFY_ADMIN_INVOICE_DEFAULTS', '', $show_product_images, $show_attrib_images, $attr_img_width);
+if (!isset($attr_img_width)) $attr_img_width = '25';
```
proseLA, line 14: https://www.php-fig.org/psr/psr-12/#51-if-elseif-else

admin/packingslip.php
brittainmark: Sorry. Will correct. Thanks for pointing out.

admin/invoice.php
brittainmark: Does that mean I should change 
lines 11-12 as well to 
if (condition) {
    action;
}
as well I copied the format from these lines. Happy to make change, just trying to be consistent.

Same in packingslip,php

brittainmark:
Additionally, should I continue to use 2 spaces for new code? The original and some new is double-spaced as opposed to 4 spaces, or should I use 4 spaces, as I have, for the new code?

admin/invoice.php
brittainmark: Have changed them all. If its wrong, I'll change the old ones back.

brittainmark:
Updated for standards. Thanks for feedback.

proseLA:
unfortunately the standards are not as clear as one would like.  especially when dealing with legacy code.  here is a link to something that was done before:

https://docs.zen-cart.com/dev/contributing/coding_standards/

i am a fan of the PSR-12 standards which extend PSR-2.  but i am merely a contributor...

i am also a huge fan of the `<?=` syntax for `<?php echo` which is part of one of the PSR standards; but i have not been able to convince anyone here of its usage and ease of readability.

i use it frequently in ALL my plugins with no ill will...

best.

scottcwilson:
@brittainmark can you smash to a single commit?

brittainmark:
Squashed as requested.

PR #4853, opened 2022-06-06
v158: Updating what's new for zc158, identifying store-owner vs. deve…

### LINKED ISSUE #4762: 1.5.8 Whats New list
Please note any items which should be added to the What's New file for 1.5.8.

lat9:
…loper care-abouts

Fixes #4762

### DIFF
--- a/docs/whatsnew_1.5.8.html
+++ b/docs/whatsnew_1.5.8.html
@@ -157,59 +157,237 @@
             <h1>
                 What's New ... Changes from prior versions
             </h1>
-            <h2>
-                The most notable improvements and bugfixes in v1.5.8 since v1.5.7 include:
-            </h2>
-            <ul>
-               <li>Admin: Improved dashboard performance.</li>
-               <li>Admin: Implement "forget" option for customer deletion to support GDPR.</li>
-               <li>Admin: multiple per page javascript now supported in Admin.</li>
-               <li>Admin: multiple per page CSS now supported in Admin.</li>
-               <li>Admin: more notifiers.</li>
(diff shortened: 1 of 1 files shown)


scottcwilson:
I am not in favor of this change.

a) This makes the file harder for me to maintain.
b) Can you point to any other products that make this distinction between fixes a store owner would care about vs fixes a developer would care about? This seems like something that would produce needless argument about which category a fix was in.

torvista:
1) At the end of this page, there is a blasé link about a full changelist being in GitHub.....3,727 changes to date.
At the beginning of this page there is the usual preamble about merging new file changes into your existing shop...would anyone seriously consider that for this release? I think the preamble should make clear that adding customisations to the vanilla files is the recommended option.

2) With the above in mind, a list of breaking changes becomes pretty academic since users will have to review each addition to the vanilla code anyway. 

3) Regarding this PR, I think two sections are required, as I've seen on another project.
a) a new features table with two columns. One column to explain the feature in plain english and the second column with any notes/comments as how it may affect pre-existing code/mods/plugins.
b) a list of changes that are purely under-the-hood/not feature-related.

I think making a distinction between a user and a developer is counter-productive when what is required in general is to encourage users to **be** developers/blur that mental barrier.

lat9:
Never mind then.  I was attempting to make the what's new more amenable to new/existing store owners, since I'm always asked "what's the motivation".

scottcwilson:
The motivation for upgrade - please point them to (and help me improve)

https://docs.zen-cart.com/user/upgrading/about_upgrading/

PR #4855, opened 2022-06-07
Add comment_position field to products_options

scottcwilson:
This field is referenced in code, but, as Cindy noted in a forum post 

not actually in the database.  This PR adds it to the database and handles it on the admin and storefront sides.

I added the comment "This is the color comment" to the Color attribute.  
Here it is in position 0

color_pos_0

and here it is in position 1

color_pos_1


scottcwilson:
Relevant doc:
https://docs.zen-cart.com/user/admin_pages/catalog/option_name_manager/#option-name-settings

mc12345678:
For completeness, the following should also be removed in this modification:
https://github.com/zencart/zencart/blob/fbfa84831fe9ffd593bda9f9e70e526a8aa5cd94/includes/modules/attributes.php#L119-L122

Otherwise, the value could have been and still could be modified by either the second of the below two or combination of both notifiers: 
````php
    $zco_notifier->notify('NOTIFY_ATTRIBUTES_MODULE_START_OPTION', $products_options_names->fields);

        $zco_notifier->notify('NOTIFY_ATTRIBUTES_MODULE_START_OPTIONS_LOOP', $i++, $products_options->fields, $products_options_names->fields, $data_properties, $field_disabled, $attributeDetailsArrayForJson);
````

PR #4860, opened 2022-06-09
admin activity: add admin user filter

torvista:
Clipboard01

Why?
Recently I had reason to check specific user activity and this was useful.

### DIFF
--- a/admin/admin_activity.php
+++ b/admin/admin_activity.php
@@ -42,6 +42,14 @@ $filter_options[3] = array('id' => '3', 'text' => TEXT_EXPORTFILTER3, 'filter' =
 $filter_options[4] = array('id' => '4', 'text' => TEXT_EXPORTFILTER4, 'filter' => 'notice+warning');
 $post_filter = (isset($_POST['filter']) && (int)$_POST['filter'] >= 0 && (int)$_POST['filter'] < 5) ? (int)$_POST['filter'] : ($post_format == 1 ? 0 : 4);
 $selected_filter = $filter_options[$post_filter]['filter'];
+//filter by admin user
+$filter_by_user = empty($_POST['filter_user']) ? 0 : (int)$_POST['filter_user'];
+$filter_by_user = zen_get_admin_name($filter_by_user) !== null ? $filter_by_user : 0; //  check for a valid admin id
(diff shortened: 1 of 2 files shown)


scottcwilson:
Interesting idea, but I would think that exporting and using your spreadsheet's filtering command would be faster.  
Let's see what others think.

torvista:
> I would think that exporting and using your spreadsheet's filtering command would be faster.
Faster than selecting a dropdown? Not at all. Remember this also applies to viewing onscreen.

scottcwilson:
@proseLA @lat9 Are you OK with merging this PR?  Looking for a +1.

proseLA:
without getting too deep into the weeds, in general, i have no problem with what @torvista is doing here.

i do have some changes for issues that i have found.  specifically, somewhere at the top, i would add the following line:

```
$confirmation_needed = false;
```
i would change the following lines:
https://github.com/zencart/zencart/blob/f3cb6df1c59f6a31e2cd3cfc4b32f7576b1116ad/admin/admin_activity.php#L136-L139

to:
```
                if ($records == 0) {
                    $messageStack->add(TEXT_NO_RECORDS_FOUND, 'error');
                    $action = '';
                } else { //process records
```
and i would then change these lines:
https://github.com/zencart/zencart/blob/f3cb6df1c59f6a31e2cd3cfc4b32f7576b1116ad/admin/admin_activity.php#L297-L301
to:
```
                       unset($f);
                    } // endif $save_to_file
                    zen_redirect(zen_href_link(FILENAME_ADMIN_ACTIVITY));
                } //end if $records for processing not 0
                break;
```

if you select a user that has no records, you are redirected and lose all of your post data.  i do not like that.  there is no need to redirect when no records are found.

while i am here, i would change all 33 instances of `<?php echo ` to `<?= `.  i find that format far more legible when echoing php vars within the html.

torvista:
I think both these suggestions warrant separate PRs as they seem additional, so I can do that subsequently.

torvista:
The two suggestions are additional but not part of this PR, as far as I can see. So this PR could be merged as is, I think?

scottcwilson:
OK based on @proseLA 's +1 I will review and merge this PR shortly.

PR #4865, opened 2022-06-11
v158: Add previously-used variable for compatibility with current tem…

lat9:
…plates

As identified in [this](https://github.com/lat9/ZCA-Bootstrap-Template/issues/123) issue raised against the Bootstrap template, update `includes/modules/pages/account_history/header_php.php` to include a variable that's been provided since zc155 (and possibly earlier) for compatibility with pre-existing templates.


scottcwilson:
Thanks for doing this - one less entry in Release Specific Upgrade Considerations!

PR #4866, opened 2022-06-12
admin, customer delete: add customer forget/delete text info

torvista:
Clipboard01


scottcwilson:
Great addition - makes the operation more self documenting.  Thanks.

PR #4872, opened 2022-06-13
admin, customers: bugfix, relocate IP data prior to observer

torvista:
IP registration header is **before** table header observer but IP registration data was **after** table column data.
This PR corrects this.

Also corrected th alignment.
Clipboard01
.


scottcwilson:
Good catch!

PR #4874, opened 2022-06-14
Add missing method outputMessageList

brittainmark:
Fixes 4416


scottcwilson:
Thank you @brittainmark and @proseLA !

PR #4876, opened 2022-06-14
Handles MySQL 8.0.17+

### LINKED ISSUE #4875: MySQL length attribute elimination in mysql 8.0.17
We can no longer rely on the result of `SHOW COLUMNS` to return a length for integer values in recent versions of MySQL.

See https://www.zen-cart.com/showthread.php?228870-Fatal-error-in-admin-includes-functions-general_functions-php for a proposed solution.

scottcwilson:
Newer MySQL versions have deprecated display width. 
Fixes #4875.

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -955,7 +955,7 @@ class queryFactoryResult implements Countable, Iterator
     }
 }
 
-class queryFactoryMeta
+class queryFactoryMeta extends base
 {
     function __construct($field)
     {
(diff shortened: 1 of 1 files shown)


includes/classes/db/mysql/query_factory.php
```diff
@@ -964,10 +964,31 @@ function __construct($field)
...
+                  $this->max_length = 11;
+                  break;
+              case 'TINYINT':
+                  $this->max_length = 4;
+                  break;
+              case 'SMALLINT':
+                  $this->max_length = 4;
+                  break;
+              default:
+                  $this->max_length = 8; // should not get here unless db changed
```
lat9, line 988: Where did this come from and why 8?  If you're going to provide a default value, a notification is in order to allow site-specific updates to deal with their changes.

scottcwilson: If you don't have a value, you'll get a fatal error.   I can take out the default branch if you want but this is really anti-bugging - if 8 is too short, they'll file a bug report and we'll know we have to add a new case.

scottcwilson: I'd be happy to change it to 64 or whatever magic number you think is ideal; I just don't want someone's site to suddenly crash.

lat9: Please add a notification after setting that default value, you'll need to add `extends base` to the `queryFactoryMeta` class statement:
```
$this->notify('NOTIFY_QUERY_FACTORY_META_DEFAULT', ['field' => $field, 'type' => $type], $this->max_length);
```

That will enable site-specific additions to (a) make their required modifications and/or (b) log a PHP notice to identify the missing type.

PR #4878, opened 2022-06-15
Remove hardcoded width setting for .home-image from HTML

dbltoe:
This removes the width setting from code so that it can be controlled in CSS


dbltoe:
Can understand your hesitancy on changes to responsive_classic but this PR, [#4877](https://github.com/zencart/zencart/pull/4877),  and [#4880](https://github.com/zencart/zencart/issues/4880) result in four 100s in Lighthouse.  91 in Performance without them.

PR #4888, opened 2022-06-18
v158: Log 'recipients_failed' emails to differently-named log file ...

### LINKED ISSUE #4803: Notify (via email?) someone when "The email server could not deliver mail ..."
More and more, I'm seeing logs of the ilk
```
[16-May-2022 12:18:41 America/Chicago] Request URI: /create-account, IP address: ---
#1 trigger_error() called at [/home/mysite/public_html/includes/functions/functions_email.php:377]
#2 zen_mail() called at [/home/mysite/public_html/includes/modules/create_account.php:440]
#3 include(/home/mysite/public_html/includes/modules/create_account.php) called at [/home/mysite/public_html/includes/modules/pages/create_account/header_php.php:17]
#4 require(/home/mysite/public_html/includes/modules/pages/create_account/header_php.php) called at [/home/mysite/public_html/index.php:36]
--> PHP Notice: Email Error: SMTP Error: The following recipients failed: joe@example.com: The mail server could not deliver mail to joe@example.com. The
account or domain may not exist, they may be blacklisted, or missing the
proper dns entries.
<br /> in /home/mysite/public_html/includes/functions/functions_email.php on line 377.
```
when a site's configured to use SMTP email and running on a 'smart' email server.  Being lowly PHP notices, they're only logged when full logging is enabled and the site's admin(s) need to scroll through all the logs to identify the issue, i.e. the customer's not going to receive the emails that they think they will, whether create-account or order-confirmations or for any plugins (think BISN) that also send out emails.

As much as I hate to add more configuration settings, I'm considering an new `Configuration :: E-mail Options` setting.  The proposed setting would be set to either an empty string or a collection of email addresses to which the notifications would be sent.

Any other suggestions?

lat9:
... to make the issues more easily identifiable.

Fixes #4803

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -415,12 +415,24 @@ use PHPMailer\PHPMailer\SMTP;
       if (EMAIL_ARCHIVE == 'true'  && $module != 'password_forgotten_admin' && $module != 'cc_middle_digs' && $module != 'no_archive') {
         zen_mail_archive_write($to_name, $to_email_address, $from_email_name, $from_email_address, $email_subject, $email_html, $text, $module, $ErrorInfo );
       } // endif archiving
-    } // end foreach loop thru possible multiple email addresses
-    $zco_notifier->notify('NOTIFY_EMAIL_AFTER_SEND_ALL_SPECIFIED_ADDRESSES');
 
-    if (!empty($ErrorInfo)) {
-      trigger_error('Email Error: ' . $ErrorInfo);
-    }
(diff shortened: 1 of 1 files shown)


scottcwilson:
Note to self: help updates after 1.5.8 release - 
https://docs.zen-cart.com/user/running/mistyped_email/
https://docs.zen-cart.com/user/email/emails_not_received/#symptom-emails-sent-from-the-cart-are-not-arriving

PR #4889, opened 2022-06-18
admin activity: use php echo short tags and remove superfluous semicolons

torvista:
using both makes the code much clearer/cleaner

### DIFF
--- a/admin/admin_activity.php
+++ b/admin/admin_activity.php
@@ -267,7 +267,7 @@ if ($action != '') {
             // HTML
             ?>
             <!doctype html>
-            <html <?php echo HTML_PARAMS; ?>>
+            <html <?= HTML_PARAMS ?>>
               <head>
                   <?php require DIR_WS_INCLUDES . 'admin_html_head.php'; ?>
               </head>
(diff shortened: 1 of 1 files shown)


proseLA:
i'm not sure on the semi-colons.  it seems that the PSR does not make a comment on it.  i'm used to the semi-colons; and one could convince me otherwise i suppose.

torvista:
I'd never heard of dropping the semicolons before, but after reading these
https://stackoverflow.com/questions/52757180/are-semicolons-in-php-optional
https://github.com/prettier/plugin-php/issues/609

and bearing in mind the idea is to make clearer code, I think it has merit when combined with the sort tag and embedded in html.
An admin decision is required, as it would need to be documented in ZC code style.

scottcwilson:
I'm not convinced this is better.  Please don't go on a tear making this change in a bunch of places.

lat9:
> I'm not convinced this is better. Please don't go on a tear making this change in a bunch of places.

Neither am I.  I'm all for clearer/consise code, but recognize that there are a lot of do-it-yourself store owners for which this change will *not* be clearer!  I'm guessing that the use of exactly-equal-to (`===`) and the null-coalescing (`??`) operators might throw them for a loop, much less the short-echo (`<?=`) tags.

torvista:
So, the consensus for now is, do not use short tags or drop semicolons?
Whatever is fine by me, as long as the decision is clear for all.

scottcwilson:
Yes, please do not use short tags.

PR #4893, opened 2022-06-19
check string before running operations

### LINKED ISSUE #4892: admin page registration: php 8.1 PHP Deprecated: strtr()
ZC158, php 8.1.7

> [19-Jun-2022 19:44:27 UTC] Request URI: /zencart/admin/index.php?cmd=admin_page_registration, IP address: 127.0.0.1
> #1 \zencart\includes\functions\functions_strings.php(24): strtr()
> #2 \zencart\admin\includes\functions\html_output.php(418): zen_output_string()
> #3 \zencart\admin\admin_page_registration.php(141): zen_draw_pull_down_menu()
> #4 \zencart\admin\index.php(11): require('zenca...')
> --> PHP Deprecated: strtr(): Passing null to parameter #1 ($string) of type string is deprecated in \zencart\includes\functions\functions_strings.php on line 24.

Back to the unresolved discussion to fix or bin this page:
https://github.com/zencart/zencart/pull/3533

scottcwilson:
Partial fix for issue raised in #4892.

### DIFF
--- a/includes/functions/functions_strings.php
+++ b/includes/functions/functions_strings.php
@@ -15,6 +15,9 @@
  */
 function zen_output_string($string, $translate = false, $protected = false): string
 {
+    if (empty($string)) {
+       return $string; 
+    }
     if ($protected === true) {
         $double_encode = (IS_ADMIN_FLAG ? FALSE : TRUE);
         return htmlspecialchars($string, ENT_COMPAT, CHARSET, $double_encode);


lat9:
Let's address the issue at its source; in /admin/admin_page_registration.php, the initialization of the `$menu_options` array sets the default 'id' to `NULL` instead of a _more proper_ empty string `''`.

https://github.com/zencart/zencart/blob/538601126506f1c3c80ed1a5b956f659b1043d94/admin/admin_page_registration.php#L90

scottcwilson:
See #4898

lat9:
#4898 looks good; please close this one.

scottcwilson:
You don't think we should check inputs for one of our functions?

lat9:
If you're going to 'check', then change the parameter type to `string` instead of blindly returning an invalid value.

scottcwilson:
Let's close for now.  As you noted, the real error was in the caller, so that change might be the only one needed.

PR #4903, opened 2022-06-20
v158: Correcting `contact_us` page's optional "Send to" dropdown

### LINKED ISSUE #4900: Send Email To: Selection Is Not Really Required on Contact_Us
Somewhat of a two-fold problem that I can separate if needed.

1.  Regardless of any selections being set for the Send Email to:, the focus when landing on the Contact Us page is on the Full Name.  Most of the time, users just enter their name, email, message, and hit send.  At topkayaker.com, they really need a separation between kayak questions, order questions, and website concerns.  They are constantly having to forward emails internally that went to kayak questions instead of order questions.  Swapping their position in the Contact Us form would simply swap who would be forwarding emails.  Since the Send Email To: only shows up when items are entered in the Admin, it will never result in an error for not having something in a required field.  Right now, our workaround is to have the first item set to go to an autoresponder that let's them know they have not selected a valid department and gives a little more information.  If they properly select kayak questions, order questions, or website; the proper email is sent.  Perhaps a method to focus on the Send Email To: if it exists is the first step to making this better.
2. The Send Email To: field in the Contact Us form is set as required.  The problem is that, if used by the store, the field will never be blank and there is no way for the owner to insert a "Please Select" like in the sideboxes.  Therefore, the user can fill in the three other required fields and just hit enter.  The email is sent even if the department they want to reach was not selected.

I combined these as I think the change of focus and actually requiring a Send Email To: option should be combined as a final repair for the problem.

Minor is that the responsive_classic propagates the field without emphasis and with a smaller font than everything else on the page.  This contributes to the overlooking of the selection but is easily fixed with CSS.

In this case, I'm seeing the same response in all browsers. 😃

lat9:
Fixes #4900

### DIFF
--- a/includes/modules/pages/contact_us/header_php.php
+++ b/includes/modules/pages/contact_us/header_php.php
@@ -131,10 +131,11 @@ if (zen_is_logged_in()) {
     $telephone = $check_customer->fields['customers_telephone'];
 }
 
-$send_to_array = array();
-if (CONTACT_US_LIST !=''){
-    foreach (explode(",", CONTACT_US_LIST) as $k => $v) {
-        $send_to_array[] = array('id' => $k, 'text' => preg_replace('/\<[^*]*/', '', $v));
+$send_to_array = [];
+if (CONTACT_US_LIST !== ''){
+    $send_to_array[] = ['id' => '', 'text' => PLEASE_SELECT];
+    foreach (explode(',', CONTACT_US_LIST) as $k => $v) {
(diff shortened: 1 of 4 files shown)


dbltoe:
Works great! Can we add the selected item to the email as the Subject?

Pretty Please. 😃

lat9:
> Works great! Can we add the selected item to the email as the Subject?
> 
> Pretty Please. 😃

That's a totally different request, albeit within these modules' processing.  The `contact_us` page has **never** included that selection in the email subject, so another 'issue' identifying this change request should be opened.

PR #4904, opened 2022-06-22
admin activity initialise $confirmation_needed

torvista:
as recommended here
https://github.com/zencart/zencart/pull/4860#issuecomment-1156653809

### DIFF
--- a/admin/admin_activity.php
+++ b/admin/admin_activity.php
@@ -18,7 +18,7 @@ require ('includes/application_top.php');
 if (!defined('DIR_FS_ADMIN_ACTIVITY_EXPORT')) {
   define('DIR_FS_ADMIN_ACTIVITY_EXPORT', DIR_FS_ADMIN . 'backups/');
 }
-
+$confirmation_needed = false;
 $action = (isset($_GET['action']) ? $_GET['action'] : '');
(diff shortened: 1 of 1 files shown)


proseLA:
my comment has a bunch of other code there.  that is all so that the selection does not get reset when there are no records.

in addition, these lines are not needed:
https://github.com/zencart/zencart/blob/cc9cad3729d71faf305a31d5c03a1e9473ebd9ec/admin/admin_activity.php#L119-L120

someone allowed merging of your commit with really poor formatting:
https://github.com/zencart/zencart/blob/cc9cad3729d71faf305a31d5c03a1e9473ebd9ec/admin/admin_activity.php#L125-L127

and while you are here, i would move the `array` to short format, ie, `[]` and reformat the whole page to whatever is the most current version of what we are using, ie 4 spaces, and whatever is in PSR-12.

torvista:
I'm going round in circles to make this more elegant!
https://github.com/zencart/zencart/blob/6cec70b0f9621d706a4fa9c1548c7d951ef7c02f/admin/admin_activity.php#L121-L126

proseLA:
that's funny.  i looked at that section too and bailed on it...

PR #4906, opened 2022-06-22
v158: Correct `options_name_manager` changing option type to 'TEXT' o…

### LINKED ISSUE #4854: Changing an option name from dropdown to text doesn't work
If you edit a dropdown option and change the type to text, you won't be able to add this attribute in the Attribute Manager.

lat9:
…r 'FILE'

Fixes #4854


scottcwilson:
Fantastic catch! Thank you.

PR #4913, opened 2022-06-25
Add name= to Ambiguous Links in Define_Main_Page

dbltoe:
It's a minor infraction to use "use this link" for the three template-viewing options in the includes/languages/english/html_includes/responsive_classic/define_main_page.php.

These three link fixes get us closer to removing the 14 link alerts on the default home page.


dbltoe:
In looking over the 14 ambiguous links reported on the home page of an initial load with demo products, 11 of them are false positives.

3 are banner links going to Zen Cart and 8 are from four products in the "New Products For June in the center column that also appear in the Bestsellers sidebox.

lat9:
Regarding these 'ambiguous links', does this mean that the "more info ..." links on a product listing page when the product(s) have attributes are also considered 'ambiguous'?  I'll note that each of the links specified for that define-page are unique in their associated $_GET variables.

dbltoe:
☹️ Unfortunately, yes.

I was checking those when I found the link error.  Guess I saw a squirrel

dbltoe:
What should be done?
If the hrefs are different, the accessible name and accessible description of each link should be uniquely identified. Note that, if the hrefs are identical, the link names/descriptions can be the same or different.

Perhaps adding the title of the product to the created link?

`<a class="list-moore"href="https://yoursite.com/index.php?main_pagge=producct_info&cPagth=22&products_id=34">... more _insert title here_ info</a>`

scottcwilson:
Please open a new issue for the problem noted above.  Adding the product name is certainly a reasonable solution.

Issue #4912, opened 2022-06-25
shipping estimator postcode field named zip_code


torvista:
On all other instances of a postcode field (create account, address book, checkout shipping/payment address) it is name and id of postcode.
On the shipping estimator it is name zip_code with no id.
Any objection to me changing this to match the rest?
While it is no big deal adding this exception in the postcode validation JS I have watching this field everywhere, it seems a good moment to consolidate it.

scottcwilson:
Thank you.  Please proceed.

PR #4914, opened 2022-06-25
Shipping Estimator: change form field zip_code to postcode

torvista:


proseLA:
why is this necessary?  what is the problem you are trying to correct?

torvista:
https://github.com/zencart/zencart/issues/4912

proseLA:
thanks for the issue reference.  could you please put that in your initial commit message next time?  personally i focus more on PRs than on issues.  and it would be nice to reference other code to see why you are doing what you are doing.

this PR corrects a javascript exception with the shipping estimator.  but now the code looks awful to me.  why are we using a var `$zip_code` that comes from an input field called `postcode`?  is that done in the other referenced scripts?  what about this var `$_SESSION['cart_zip_code']`?  it seems it is only used here in this script.

MO, this change needs more work.  ideally code should be self-documenting.  taking an input field with one name and creating a var using another name seems like pretty bad coding to me.  i see the need for the change, but i would clean up all of this code.

torvista:
While I agree that the mismatch of zip_code and postcode is inelegant, I find it increasingly difficult to know where to draw the line with code changes that do not fix a specific problem.
If I change one line in a file for a reason, I would be fine spending another 10 minutes doing a cleanup on the entire file without changing any functionality but that draws the reaction of making if difficult for upgraders, so I don't bother.
Similarly for this, changing names of things for no specific reason draws complaints.

torvista:
Could you clarify what we disagree about today?
:grin:

proseLA:
@torvista, only if you can agree that i am not disagreeable... :smirk:

you (or we) have successfully removed an input field as well as a var that has the term `zip_code` everywhere except in 2 files.  see:
```
charmes: # grepp zip_code
includes/modules/shipping_estimator.php:42:    $zip_code = (isset($_SESSION['cart_zip_code'])) ? $_SESSION['cart_zip_code'] : '';
includes/modules/shipping_estimator.php:43:    $zip_code = (isset($_POST['zip_code'])) ? strip_tags(addslashes($_POST['zip_code'])) : $zip_code;
includes/modules/shipping_estimator.php:90:                'postcode' => $zip_code,
includes/modules/shipping_estimator.php:105:            $_SESSION['cart_zip_code'] = $zip_code;
includes/modules/shipping_estimator.php:112:                'postcode' => $zip_code,
includes/templates/template_default/templates/tpl_modules_shipping_estimator.php:75:<?php echo  zen_draw_input_field('zip_code', $zip_code, 'size="7"'); ?>
charmes: #
```

now you want to rename 3 of the 10 instances to `postcode` so as to not cause a javascript exception.  which, while it will address the javascript exception, provides code that looks like the coder did not take the time to properly name his vars.

and i find your excuses of taking the additional 10 minutes (which is what time i am now taking to enumerate this problem) to clean up the code as well as the reaction of making it difficult for upgraders not at all convincing.  in fact, i am the one that frequently complains about needless code changes to a file; but i agree that we need to address the javascript exception.  so while we are here, it only makes sense to have vars that properly reflect their input field names.

frankly, i thought i was clear in my initial comment.  but hopefully this makes it clearer in a non-disagreeable way...

torvista:
For me the 10 minute cleanup is doing short array syntax, simple strict comparisons and other IDE recommendations.
I stopped short of touching the cart_zip_code as I assumed it was heading into a rabbit hole, but it seems not.
So, zip_code is no more.

PR #4916, opened 2022-06-25
Update stylesheet.css

dbltoe:
Minor Cleanups

Removes a double border call and fixes some 0 dimensions that have em or px attached.
Removes a margin call that should not be used with display:inline


dbltoe:
This is not what I input

Closing til I can fix

PR #4919, opened 2022-06-25
Update stylesheet_colors.css

dbltoe:
Most ADA/WCAG testers cannot check if a hover is done correctly.

This fix addresses those hovers and their components that were missing, did not comply with minimum color contrast, are needed to match other color combinations of similar items on a page like the back and search buttons on the search page.

As far as I can tell, this fixes all color contrast on all pages but, I welcome any gems you may find.

I did add new lines 32 and 33 simply for a little pizzazz when someone is viewing a new install.


dbltoe:
Found a couple more hidden gems that surfaced during color change

PR #4926, opened 2022-06-26
Update mysql_zencart.sql

dbltoe:
Changing the title of this wide banner may increase exposure to certified services and increase affiliate returns.


scottcwilson:
Not a bad idea but why not do `zc_install/sql/updates/mysql_upgrade_zencart_158.sql` as well?

dbltoe:
Good Point

scottcwilson:
Correct.  Add an update.  
When we update the create script, it's good to also put an update into the current database upgrade script.

scottcwilson:
See #4927 for upgrade script change.

scottcwilson:
This is fine but next time please put both in the same PR, then it's obvious that you have done both.

dbltoe:
Well, this grasshopper has yet to figure out the process of combining or I would have.

Will try to find time to figure it out.

PR #4927, opened 2022-06-26
Update mysql_upgrade_zencart_158.sql

dbltoe:
This will give the proper banners_title for current banners_id 6 in older banner tables.

I used banners_html in case the older DB had been changed.


scottcwilson:
See #4926 for install script change.

PR #4930, opened 2022-06-27
Check output before referencing

scottcwilson:
Fixes this log, seen during `index.php?main_page=checkout_process`: 
```
--> PHP Warning: sizeof(): Parameter must be an array or an object that implements Countable in /var/www/html/myaccount/mystore.com/shop/includes/classes/order_total.php on line 66.
```

### DIFF
--- a/includes/classes/order_total.php
+++ b/includes/classes/order_total.php
@@ -63,6 +63,9 @@ class order_total extends base {
         $class = substr($value, 0, strrpos($value, '.'));
         if (!isset($GLOBALS[$class])) continue;
         $GLOBALS[$class]->process();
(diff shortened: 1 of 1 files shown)


includes/classes/order_total.php
```diff
@@ -63,6 +63,7 @@ function process() {
         $class = substr($value, 0, strrpos($value, '.'));
         if (!isset($GLOBALS[$class])) continue;
         $GLOBALS[$class]->process();
+        if (empty($GLOBALS[$class]->output)) continue;
```
lat9, line 66: Please, no single-line if statements.  Add braces, per current PSR.

scottcwilson: Roger that.

PR #4937, opened 2022-06-28
Remove SquareWebPay_*.log files

scottcwilson:


dbltoe:
Since Square is going to be dropped, shouldn't it be dropped here as well?

PR #4938, opened 2022-06-29
remove preg_replace; use new parm for build search function

proseLA:
the 3rd parm for the zen_build_keyword_search was added after this code was merged.  the conditional is also superfluous, as i think its faster (and cleaner) to just concatenate the empty string....  as that is what we are going to do anyway.


scottcwilson:
Not sure about the third change  (remove preg_replace) - what if `$search` was modified by the notifier NOTIFY_ADMIN_ORDERS_SEARCH_PARMS?

proseLA:
> Not sure about the third change (remove preg_replace) - what if `$search` was modified by the notifier NOTIFY_ADMIN_ORDERS_SEARCH_PARMS?

what if?  people using notifiers need to ensure that they do not break code.  the formatting of notifiers is very specific and their use can be very fragile.  this notifier was added in v157, and OPC only adds an extra field when using it.

in addition, keeping this bit of legacy code breaks the changes that i am proposing.  if one were to search for 2 terms and keep this code the SQL fails.  by ensuring that `$search` starts with `WHERE`, this bit of legacy code is no longer needed.  if `$search` comes back from the notifier and it longer starts with `WHERE`, that's on the 'modder' (as you like to say).

this conversation did get me looking at another bit of this search code that i think can be improved.  specifically the orders_products search.  little bit of background, when one goes to google, and enters a UPS tracking ID, there is no need to preface it with anything.  google already knows it.  given that, i think we should change from:

https://github.com/zencart/zencart/blob/8db03830a5a4f98e8bca2abe37ec7459592c6f23/admin/orders.php#L1200-L1213

to:
```
if (!empty($_GET['search_orders_products'])) {
                    $search_distinct = ' DISTINCT ';
                    $new_table = " LEFT JOIN " . TABLE_ORDERS_PRODUCTS . " op ON (op.orders_id = o.orders_id) ";
                    $keywords = zen_db_input(zen_db_prepare_input($_GET['search_orders_products']));
                      $keyword_search_fields = [
                          'op.products_name',
                          'op.products_model',
                          'op.products_id',
                      ];
                      $search = zen_build_keyword_where_clause($keyword_search_fields, trim($keywords), true);
                    if (substr(strtoupper($_GET['search_orders_products']), 0, 3) === 'ID:') {
                      $keywords = trim(substr($_GET['search_orders_products'], 3));
                      $search .= " OR op.products_id ='" . (int)$keywords . "'";
                    }
                  } elseif (!empty($_GET['search'])) {
```
this PR definitely breaks this part of the search by not returning correct results.  this change eliminates the need to preface the product ID when searching by product_id, fixes the return results, and also preserves that legacy type of search.

PR #4941, opened 2022-06-29
module shipping estimator tidy syntax

torvista:
syntax tidies, no functional changes:
short array syntax
safe strict comparisons
correct method name
null coalesce
list to []

Can be squashed after review.

### DIFF
--- a/includes/modules/shipping_estimator.php
+++ b/includes/modules/shipping_estimator.php
@@ -188,10 +188,10 @@ if ($_SESSION['cart']->count_contents() > 0) {
     // begin shipping cost
     if (!$free_shipping && $_SESSION['cart']->get_content_type() !== 'virtual') {
         if (!empty($_POST['scid'])){
-            list($module, $method) = explode('_', $_POST['scid']);
+            [$module, $method] = explode('_', $_POST['scid']);
             $_SESSION['cart_sid'] = $_POST['scid'];
         } elseif (!empty($_SESSION['cart_sid'])) {
-            list($module, $method) = explode('_', $_SESSION['cart_sid']);
+            [$module, $method] = explode('_', $_SESSION['cart_sid']);
         } else {
             $module = '';
             $method = '';
--- a/includes/modules/shipping_estimator.php
+++ b/includes/modules/shipping_estimator.php
@@ -252,7 +252,7 @@ if ($_SESSION['cart']->count_contents() > 0) {
         $show_in = FILENAME_SHOPPING_CART;
     }
     if (zen_is_logged_in() && !zen_in_guest_checkout()) {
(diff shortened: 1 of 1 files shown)


includes/modules/shipping_estimator.php
```diff
@@ -252,24 +252,24 @@ function shipincart_submit(){
         $show_in = FILENAME_SHOPPING_CART;
     }
     if (zen_is_logged_in() && !zen_in_guest_checkout()) {
-        $addresses = $db->execute("SELECT address_book_id, entry_city AS city, entry_postcode AS postcode, entry_state AS state, entry_zone_id AS zone_id, entry_country_id AS country_id FROM " . TABLE_ADDRESS_BOOK . " WHERE customers_id = '" . (int)$_SESSION['customer_id'] . "'");
+        $addresses = $db->Execute("SELECT address_book_id, entry_city AS city, entry_postcode AS postcode, entry_state AS state, entry_zone_id AS zone_id, entry_country_id AS country_id FROM " . TABLE_ADDRESS_BOOK . " WHERE customers_id = '" . (int)$_SESSION['customer_id'] . "'");
```
proseLA, line 255: #4942 
i am of the opinion, that this code should be:
```
$customer = new Customer($_SESSION['customer_id']);
```
`$addresses` would then be `$addresses = $customer->getAddressBookEntries();`

and then you would have spotted the bug listed above.

if we have a class, **lets use it!**

torvista: Sure, once it is clarified.

scottcwilson: @proseLA PR merged.

torvista: I was waiting for this PR to be cleared/squashed as this class use is a real change.

torvista: This/my branch does not incorporate the class change and although I have attempted to bring in the changes in the z158 branch subsequently, I am not convinced/confident of the commit results.

scottcwilson: The PR #4949 was merged, and the function getAddressBookEntries returns an object. 
Check includes/classes/Customer.php in an updated version of the Zen Cart 158 branch.

torvista: Sure, but I don't have clear how to get that change into this branch.
But that is academic, I've just started to try this and it is not so straightforward as expected, it is not drop-in, so should definately be in a separate PR.

proseLA: is anything so straight forward?

scottcwilson:
I'm going to merge this one so we can start fresh on getting it to use the class.  It might make sense to put this function that Paul has identified in the Customer class in case others want to use it.

PR #4949, opened 2022-06-30
customer class fix typeError

### LINKED ISSUE #4942: customer->getAddressBookEntries returns object when return value is array.
i am a fan of using classes.  this method is speced wrong.  right now it does not seem used anywhere.  but if we want to use it, do you want to return an array or the mysql object?

https://github.com/zencart/zencart/blob/00e07066774211149329516968822e288561cddd/includes/classes/Customer.php#L420-L435

PHP Fatal error: Uncaught TypeError: Return value of Customer::getAddressBookEntries() must be of the type array, object returned in /var/www/zcdev/includes/classes/Customer.php:434

proseLA:
fixes #4942

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -417,7 +417,7 @@ class Customer extends base
         return false;
     }
 
-    public function getAddressBookEntries(int $customer_id = null): array
+    public function getAddressBookEntries(int $customer_id = null): object
     {
         global $db;
 


scottcwilson:
Wouldn't this break line 442 below? 
```
        return count($this->getAddressBookEntries());
```

I think it would be better to fix this function to actually return an array.

proseLA:
function `getNumberOfAddressBookEntries` is already broken.  it calls the broken method `getAddressBookEntries` which this fixes.

i have already looked at returning an array.  if the broken method returns an array, `getNumberOfAddressBookEntries` would now return 32 (or 33) which is the number of data values in each array element as opposed to the number of indexes in the array.  so that method would need to be refactored as well.

in addition, the function `getFormattedAddressBookList` already returns an array; which would be similar if the not the same as what you are now suggesting.

i think the intention of the author was to return an object (a simple error in the return) as this PR corrects.

scottcwilson:
I see your point - the functions ending in List return an array, so the Entries function return a result set.  Thanks.

PR #4965, opened 2022-07-03
admin: category-product listing, add more sorting options

torvista:
Of course I expect a bun-fight, so don't hold back.
I've just added to the existing, but perhaps a rethink would be better.
Feel free to submit your own preferred lists and maybe we'll get to a consensus.

Not doing the sql for the default until that point.
Clipboard01

A sort of the lang array would be a separate PR.

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -414,22 +414,32 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
               	$zc_skip_products = false;
             }
 
-            if ($zc_skip_products == true) {
-              // toggle switch for display sort order
-              $categories_products_sort_order_array = array(array('id' => '0', 'text' => TEXT_SORT_CATEGORIES_SORT_ORDER_CATEGORIES_NAME),
-                array('id' => '1', 'text' => TEXT_SORT_CATEGORIES_NAME)
-              );
+            if ($zc_skip_products === true) {
+                // toggle switch for category display sort order
+                $categories_products_sort_order_array = array(
+                    array('id' => '0', 'text' => TEXT_SORT_CATEGORIES_SORT_ORDER_CATEGORIES_NAME),
+                    array('id' => '1', 'text' => TEXT_SORT_CATEGORIES_NAME),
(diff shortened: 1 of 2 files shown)


proseLA:
will crumpets be served?

i think sorting by products status is silly.  why not an additional pull-down with 3 choices.  don't you think that would be of more value?

image

torvista:
> will crumpets be served?

Of course not, [you can't go back to savory after sweet](https://www.youtube.com/watch?v=L8eh72REd_s).
 
Another drop-down? I can't see me doing that work, adding more sorting options is too easy.

zcwilt:
The last thing I expected was a John Shuttleworth video on Zen Cart.
Maybe the https://www.youtube.com/watch?v=Cj8n4MfhjUc

torvista:
>i think sorting by products status is silly. 

Not sure why. Without going into my use case, I use it. I also use weight, but didn't put that in.

If no-one cares one way or another about this, is this going to be merged or not (when complete)?

lat9:
> > i think sorting by products status is silly.
> 
> Not sure why. Without going into my use case, I use it. I also use weight, but didn't put that in.
> 
> If no-one cares one way or another about this, is this going to be merged or not (when complete)?

Still reviewing, but I think it's a good idea.  I agree with you regarding the status-sort.  There are many times where a finger-check disables a product and that would make it/them easier to locate.

admin/category_product_listing.php
```diff
@@ -711,30 +730,45 @@
...
+                case (4):
+                    $order_by = " ORDER BY p.products_quantity DESC, pd.products_name";
+                    break;
+                case (5):
+                    $order_by = " ORDER BY p.products_price_sorter, pd.products_name";
+                    break;
+                case (6):
+                    $order_by = " ORDER BY p.products_price_sorter DESC, pd.products_name";
+                    break;
+                case (7)://steve added
```
lat9, line 755: "Lose" the `//steve added` comments, please.

torvista: oops

lat9: Thanks, @torvista.

Issue #4946, opened 2022-06-29
zc158: admin, Edit Product->Preview->Back: PHP Fatal error:  Uncaught TypeError: strtr()


torvista:
Vanilla shop with images unpacked.
Admin->Edit product, image should be shown.
No need to edit anything->Preview then use Back button to trigger error.

> PHP Fatal error:  Uncaught TypeError: strtr(): Argument #1 ($string) must be of type string, array given in \zencart\includes\functions\functions_strings.php:24
> Stack trace:
> #0 \zencart\includes\functions\functions_strings.php(24): strtr(Array, Array)
> #1 \zencart\admin\includes\functions\html_output.php(133): zen_output_string(Array)
> #2 \zencart\admin\includes\functions\general.php(23): zen_image('https://www.git...', Array, '', '', '')
> #3 \zencart\admin\includes\modules\product\collect_info.php(357): zen_info_image('matrox/mg200mms...', Array)
> #4 \zencart\admin\product.php(102): require('D:\\GitHub\\zenca...')
> #5 \zencart\admin\index.php(11): require('D:\\GitHub\\zenca...')
> #6 {main}
>   thrown in \zencart\includes\functions\functions_strings.php on line 24

On the collect_info page first load, $pInfo['products_name'] is a string and of the session language.
When going Back to collect_info page, $pInfo['products_name'] is an array of the names in all languages.
$pInfo['products_name'] gets used as the alt text for the image and so causes the error on the Back.

torvista:
I put a check in the name
<?php echo zen_info_image($pInfo->products_image, (is_array($pInfo->products_name) ? $pInfo->products_name[$_SESSION['languages_id']] : $pInfo->products_name)); ?>Unless there is a need to review this process further,

lat9:
Looks reasonable to me.  Any other opinions?

lat9:
Closing, since that PR's been merged.

PR #4970, opened 2022-07-03
admin, collect_info: handle $pInfo->products_name as array or string

torvista:
Fix for problem 

> PHP Fatal error: Uncaught TypeError: strtr(): Argument https://github.com/zencart/zencart/pull/1 ($string) must be of type string, array given

as reported here: 
https://github.com/zencart/zencart/issues/4946


lat9:
Fixes #4946

PR #4971, opened 2022-07-04
v158: CatalogArraysLanguageLoader, extraneous directory-separator used

lat9:
Results in an attempt to load `includes/languages//lang.english.php`, performance wasteful.


scottcwilson:
Per `includes/defined_paths.php`, the define `DIR_WS_LANGUAGES` includes a trailing slash.

PR #4975, opened 2022-07-04
Load partial match language files

### LINKED ISSUE #4973: 158 language loading issue
As reported by @lat9: 

In 1.5.7, includes/modules/require_languages.php line 20 builds a list of language files to load using get_template_part.
1.5.8 does not do that and should do something similar.  

This does work for old style (page_name_plus.php) but not for newer array based files (lang.page_name_plus.php)

Example: includes/languages/english/lang.create_account_register.php

scottcwilson:
example: load create_account_register on create_account page

Fixes #4973


lat9:
Nope, still not loaded.  I'll note that both `/includes/languages/english/create_account_register.php` and `/includes/languages/english/lang.create_account_register.php` exist.  Only the 'legacy' version is being loaded.

scottcwilson:
Ugh sorry declared victory too early.  Coming shortly.

scottcwilson:
should be ok now.

lat9:
That has the extra-added-benefit that if a `lang.something.php` file is present, the (legacy) `something.php` isn't.  That'll make it easier for plugin authors to maintain a single repository-branch that supports both the zc158 and zc157 versions!

PR #4976, opened 2022-07-04
remove dead code

scottcwilson:


includes/modules/require_languages.php
```diff
@@ -13,23 +13,3 @@
 $languageLoader->loadLanguageForView();
 return;
```
lat9, line 14: Might as well get rid of the unneeded `return;` as well.

PR #4977, opened 2022-07-04
v158: Remove duplicate language file loads, performance improvement.

lat9:
... as seen in this load of the `create_account` page:
image

### DIFF
--- a/includes/classes/ResourceLoaders/ArraysLanguageLoader.php
+++ b/includes/classes/ResourceLoaders/ArraysLanguageLoader.php
@@ -113,7 +113,7 @@ class ArraysLanguageLoader extends BaseLanguageLoader
     protected function loadDefinesWithFallback($mainFile, $fallbackFile)
     {
         $defineListFallback = [];
-        if ($mainFile != $fallbackFile) {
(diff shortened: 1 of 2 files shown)


scottcwilson:
The file matching for the array style language files had a bug - it didn't require an additional character after the page name for extra files in loadLanguageForView().  See if includes/classes/ResourceLoaders/CatalogFilesLanguageLoader.php has the same issue.

lat9:
> The file matching for the array style language files had a bug - it didn't require an additional character after the page name for extra files in loadLanguageForView(). See if includes/classes/ResourceLoaders/CatalogFilesLanguageLoader.php has the same issue.

I don't understand the comment, please elaborate.

scottcwilson:
Remove lang.create_account.php and copy in the old create_account.php.  (Not a template override, the base file.). Be sure it's not tried twice on includes/classes/ResourceLoaders/CatalogFilesLanguageLoader.php line 49 

Then do same check for template override on line 43. 

I suspect the listFilesFromDirectory calls should use `'(.+)\.php$~i'` not `'(.*)\.php$~i'`.

scottcwilson:
Kick back to me if too confusing.

PR #4994, opened 2022-07-07
v158: Correct `zen_db_perform` handling of `null` values.

### LINKED ISSUE #4962: zc158: debug for category image = null
In the categories table, categories_image default is null.
admin->edit a category that has never had an image:

> PHP Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in admin\categories.php on line 418.

Since the demo has images for all categories, this error does not show after deleting a category image as the field is then set to an empty string, but if you set the field in the db manually to null and refresh the edit category page, the error shows.

Perhaps the easiest solution is to change the db default to an empty string?

### LINKED ISSUE #4963: zc158: debug for duplicating a product with products_date_available default/null
php 8.1.7, strict reporting

Admin->Product listing->Copy-Duplicate a product.

If it has never has had date_available set, products_date_available is null and causes this debug

> --> PHP Deprecated: mysqli_real_escape_string(): Passing null to parameter #2 ($string) of type string is deprecated in \includes\classes\db\mysql\query_factory.php on line 175.

but it comes from this call
https://github.com/zencart/zencart/blob/0544f29edac64242e286690f0c9d042cf7baebb8/admin/includes/modules/copy_product_confirm.php#L69

Perhaps this default should be changed from null to now/the inserted date?

Additionally if the source product has a date set for this field, that is copied too. Does that make sense?

### LINKED ISSUE #4992: shopfront, class order->create: debugs on null values being passed to zen_db_perform
ZC158, php8.1.7, strict

In the array passed here
https://github.com/zencart/zencart/blob/7a60f75c2358052f9109abb6821e6c0300093502/includes/classes/order.php#L830
these parameters are null (in my shop)
customers_company
delivery_company
billing_company
which consequently causes a debug when passed to prepare_input and mysql_real_escape_string

These things could be fixed globally in the db or in the destination function, or do we fix each case with a check in that specific code, one-by-one...:unamused:?

lat9:
Fixes #4992

### DIFF
--- a/includes/functions/database.php
+++ b/includes/functions/database.php
@@ -92,7 +92,8 @@ function zen_db_perform(string $tableName, array $tableData, $performType = 'INS
         }
         $query = substr($query, 0, -2) . ') VALUES (';
         foreach ($tableData as $value) {
-            switch ((string)$value) {
+            $value = (string)$value;
+            switch ($value) {
                 case 'now()':
                     $query .= 'now(), ';
                     break;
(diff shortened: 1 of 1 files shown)


torvista (commented):
only fixes https://github.com/zencart/zencart/issues/4992  :disappointed:

PR #5001, opened 2022-07-09
Fix sidebox override

### LINKED ISSUE #5000: zc158 modules sideboxes: template override not working
ZC158, php 8.1.8
eg. copy/make
includes\modules\sideboxes\responsive_classic\information.php
Do some change like prefix a link with some text: change does not show, default file is still used
.

scottcwilson:
Fixes #5000

### DIFF
--- a/includes/modules/column_left.php
+++ b/includes/modules/column_left.php
@@ -25,7 +25,7 @@ $sideboxes = LayoutBox::where('layout_box_location', 0)
 $column_width = (int)BOX_WIDTH_LEFT;
 foreach ($sideboxes as $sidebox) {
     $boxFile = (new SideboxFinder(new FileSystem))->sideboxPath($sidebox, $template_dir, true);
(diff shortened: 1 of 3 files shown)


includes/modules/column_left.php
```diff
@@ -25,7 +25,7 @@
 $column_width = (int)BOX_WIDTH_LEFT;
 foreach ($sideboxes as $sidebox) {
     $boxFile = (new SideboxFinder(new FileSystem))->sideboxPath($sidebox, $template_dir, true);
-    if (file_exists($boxFile)) {
+    if ($boxFile !== false && file_exists($boxFile)) {
```
lat9, line 28: The check for `file_exists` is redundant, since the SideboxFinder has already verified that it exists if the return value isn't `(bool)false`.

Ditto for line 28 of the `column_right.php`, below.

PR #5002, opened 2022-07-09
Refactoring listFilesFromDirectoryAlphaSort

brittainmark:
Refactor listFilesFromDirectoryAlphaSorted so it is not a duplication of listFilesFromDirectory
with the sort parameter.
Remove the sort parameter from listFilesFromDirectory


lat9:
Thanks for doing this, @brittainmark!  It's a much better solution than my original.

brittainmark:
It takes people to make the first step for others to be able to contribute. 
Just happy to help.

PR #5020, opened 2022-07-12
version stamping v158

zcwilt:


torvista:
Is it of real value/necessary to change the header comments of **unmodified** files to reflect the release instead of just updating the last modified date?

I've been criticized in the past for making minor non-bug-fixing changes to a file and so increasing update-comparison-hell for upgraders.
Since I am probably the only current upgrader, these 911 file non-changes indeed caused me a big waste of comparison time equating them with my site.
I don't need telling I should not be using an alpha in production, but thanks to that, plenty of real-world bugs have come to light/been fixed.
So, bearing in mind alpha-2, beta-1 and release versions, that is a lot of non-changing files to equate, just for a comment change of questionable value: most files will pass through all phases unmodified apart from that.
Hence my question.

zcwilt:
So I have some sympathy with you here. 
I'm not a fan of the version stamping.
Updating the copyright is pointless and isn't legally required 
I could add @copyright 1762 and it wouldn't affect my copyright in 2022

On the other update stamping, again I'm not a fan, but have been told others rely on this. 
Maybe this is an artifact of poor IDE's in the past. 
So I'm definitely open to a conversation here.

PR #5021, opened 2022-07-12
v158: Correcting zcDate class debug-ouput

### LINKED ISSUE #5017: new Date class provides inaccurate information in debug log.
i am not sure why we do not check if international date is installed on php versions less than 8.1.  it makes no sense to me.

if i turn debug on to play with this class, my debug logs show intl is not installed.  and now i'm on a wild goose chase to try and install it.

if we are less than 8.1 and you want to use sftrtime; fine.  but lets not provide inaccurate data into debug logs.

https://github.com/zencart/zencart/blob/1a6e4803238c742cd799fdff97e14006a3875f38/includes/classes/zcDate.php#L32-L38

lat9:
Fixes #5017

### DIFF
--- a/includes/classes/zcDate.php
+++ b/includes/classes/zcDate.php
@@ -7,9 +7,9 @@
 class zcDate extends base
 {
     protected
-        $intlDateInstalled = false,
+        $useIntlDate = false,
         $useStrftime = false,
-        $locale,                //- Only used when $this->intlDateInstalled is true
+        $locale,                //- Only used when $this->useIntlDate is true
         $strftime2date,         //- Only used when $this->useStrftime is false
         $strftime2intl,         //- Only used when $this->useStrftime is false
         $debug = false;
@@ -33,7 +33,7 @@ class zcDate extends base
             $this->useStrftime = true;
         } else {
             if (function_exists('datefmt_create')) {
-                $this->intlDateInstalled = true;
+                $this->useIntlDate = true;
             }
             $this->initializeConversionArrays();
         }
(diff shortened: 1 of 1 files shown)


proseLA:
i am not sure how this fixes anything.

i think if we move the `intl` conditional outside of the `php8.1` conditional, we are done.

i am not sure what adding additional information to the message gives us.

lat9:
> i am not sure how this fixes anything.
> 
> i think if we move the `intl` conditional outside of the `php8.1` conditional, we are done.
> 
> i am not sure what adding additional information to the message gives us.

Color me confused.  I thought that the original issue was the superfluous information in the class' debug-logging.  Are you now suggesting that the order of the processing be changed as well?

proseLA:
> Color me confused. I thought that the original issue was the superfluous information in the class' debug-logging. Are you now suggesting that the order of the processing be changed as well?

the original and still existing issue is inaccurate information in the debug log.  saying `intl` is not installed when it is in fact installed.

it is fixed by moving the `intl` check as i described above.

that does **NOT** change the order of processing.  and i am not suggesting to change that at all.

the `_construct` method is not part of the processing.  the order of processing is done by the `output` method as seen here:

https://github.com/zencart/zencart/blob/8105434d45118c39409ae66f0e3a7c636a5965c9/includes/classes/zcDate.php#L152-L168

changing the `intl` check does not change the order of processing.

lat9:
OK, so if the class-variable name was changed from `intlDateInstalled` to `usingIntlDate`, would that correct the originally-posted issue?

proseLA:
> OK, so if the class-variable name was changed from `intlDateInstalled` to `usingIntlDate`, would that correct the originally-posted issue?

it would.

lat9:
> > OK, so if the class-variable name was changed from `intlDateInstalled` to `usingIntlDate`, would that correct the originally-posted issue?
> 
> it would.

Super!  I'll get that updated in the morning.

lat9:
Updated, per request.

proseLA:
big fan:

https://hackernoon.com/the-art-of-naming-variables-52f44de00aad

PR #5027, opened 2022-07-15
v158: Refactoring category_row.php module for PSR-12 ...

lat9:
... as identified in [this](https://github.com/lat9/ZCA-Bootstrap-Template/issues/131) ZCA Bootstrap-4 issue.

### DIFF
--- a/includes/modules/category_row.php
+++ b/includes/modules/category_row.php
@@ -11,41 +11,43 @@
  * @version $Id: Scott C Wilson 2021 Jul 28 Modified in v1.5.8-alpha $
  */
 if (!defined('IS_ADMIN_FLAG')) {
-  die('Illegal Access');
+    die('Illegal Access');
 }
+
 $title = '';
 $num_categories = $categories->RecordCount();
 
 $row = 0;
 $col = 0;
-$list_box_contents = array();
+$list_box_contents = [];
 if ($num_categories > 0) {
-  if ($num_categories < MAX_DISPLAY_CATEGORIES_PER_ROW || MAX_DISPLAY_CATEGORIES_PER_ROW == 0) {
-    $col_width = floor(100/$num_categories);
-  } else {
-    $col_width = floor(100/MAX_DISPLAY_CATEGORIES_PER_ROW);
-  }
-
-  while (!$categories->EOF) {
-    $zco_notifier->notify('NOTIFY_CATEGORY_ROW_IMAGE', $categories->fields['categories_id'], $categories->fields['categories_image']); 
-    if (empty($categories->fields['categories_image'])) {
-       $categories->fields['categories_image'] = 'pixel_trans.gif';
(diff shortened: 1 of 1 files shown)


includes/modules/category_row.php
```diff
@@ -11,41 +11,43 @@
...
-  } else {
-    $col_width = floor(100/MAX_DISPLAY_CATEGORIES_PER_ROW);
-  }
-
-  while (!$categories->EOF) {
-    $zco_notifier->notify('NOTIFY_CATEGORY_ROW_IMAGE', $categories->fields['categories_id'], $categories->fields['categories_image']); 
-    if (empty($categories->fields['categories_image'])) {
-       $categories->fields['categories_image'] = 'pixel_trans.gif';
+    if ($num_categories < MAX_DISPLAY_CATEGORIES_PER_ROW || MAX_DISPLAY_CATEGORIES_PER_ROW === '0') {
+        $col_width = floor(100/$num_categories);
```
scottcwilson, line 25: The max values are actually integers aren't they?  See laravel/app/Models/Configuration.php line 26

scottcwilson (changes requested):

includes/modules/category_row.php
```diff
@@ -11,41 +11,43 @@
...
-    $col_width = floor(100/$num_categories);
-  } else {
-    $col_width = floor(100/MAX_DISPLAY_CATEGORIES_PER_ROW);
-  }
-
-  while (!$categories->EOF) {
-    $zco_notifier->notify('NOTIFY_CATEGORY_ROW_IMAGE', $categories->fields['categories_id'], $categories->fields['categories_image']); 
-    if (empty($categories->fields['categories_image'])) {
-       $categories->fields['categories_image'] = 'pixel_trans.gif';
+    if ($num_categories < MAX_DISPLAY_CATEGORIES_PER_ROW || MAX_DISPLAY_CATEGORIES_PER_ROW === '0') {
```
scottcwilson, line 24: The max values are actually integers aren't they? See laravel/app/Models/Configuration.php line 26

scottcwilson (changes requested):

includes/modules/category_row.php
lat9: Sigh, where is this documented?  These are definite anti-patterns, since all other database configuration settings come back as strings.

Thanks for the catch; I will get this updated shortly.

proseLA: noting that it goes back 3 years ago to v156 where config groups 2 and 3 were cast as integers:

https://github.com/zencart/zencart/commit/c2ee7e42fec76bcb74e735ae0c715e3804a3037c

also noting that i have never been a fan of integers, float values and **especially booleans** being strings.

most of my mods have something like this:
```
                protected function convertToBoolean($var)
		{
			if (in_array($var, array('True', 'TRUE', 'true'))) {
				return true;
			} else {
				return false;
			}
		}
```

PR #5032, opened 2022-07-15
refactor for move to plugin controls

### LINKED ISSUE #5029: zc_plugin for admin not creating correct path to include css files if site in sub directory.
**Describe the bug**
When writing an admin plug in if you need to add your own css files for the plugin they should be placed in zc_plugins\YOURPLUGIN\VERSION\admin\include\css. If your site is not top level admin_html_head.php point to the wrong file.

Using display_logs as an example.
I create zc_plugins\display_logs\3.0.0\admin\includes\css\display_logs.css
The admin_html_head.php code creates a relative path of  "/zc_plugins/DisplayLogs/v3.0.0/admin/includes/css/display_logs.css"

This points to "YOURSITE/zc_plugins/DisplayLogs/v3.0.0/admin/includes/css/display_logs.css" which is fine if your site is not a subdirectory.

Options.
1. Change the code to point to "../zc_plugins/DisplayLogs/v3.0.0/admin/includes/css/display_logs.css" This works as you are currenlt in YOURADMIN folder
1. Change the code to include the subdirectory if it exists "/MySiteSubDirectory/zc_plugins/DisplayLogs/v3.0.0/admin/includes/css/display_logs.css"  i.e. DIR_WS_CATALOG minus trailing slash.

**Version**
Zen Cart version: 1.5.8
PHP version: 8.0.20

**To Reproduce**
1. Create a zen cart shop in a subdirectory of the main site.
1. add zc_plugins/DisplayLogs/V3.0.0/admin/includes/css/display_logs.css
1. put something in display_logs.css e.g. table {background:red}
1. install display logs if not already done so (admin>modules>plugin manager, select display logs, Install and install)
1. Create a logfile in your logs directory with a name like myDEBUG-adm-20220708-174655-111752.log
1. Return to admin home page
3. click on the link in the display message
4. log will display with no error, but the background will not be red (assuming you made that change)
5. if you enter the browser console you will see that the display_logs.css failed to load as the file could not be found.

**Expected behavior**
The created css file is loaded from the correct location

**Debug/Error Logs**
None created

brittainmark:
fixes #5029

### DIFF
--- a/admin/includes/admin_html_head.php
+++ b/admin/includes/admin_html_head.php
@@ -36,24 +36,24 @@ if (!defined('IS_ADMIN_FLAG')) {
 <?php
 }
 foreach ($installedPlugins as $plugin) {
-    $relativeDir = $plugin->getRelativePath();
+    $relativeDir = substr($plugin->getRelativePath(), 1);
     $absoluteDir = $plugin->getAbsolutePath();
     $directory_array = $template->get_template_part($absoluteDir . 'admin/includes/css/', '/^global_stylesheet/', '.css');
     foreach ($directory_array as $key => $value) {
         ?>
-        <link rel="stylesheet" href="<?php echo $relativeDir . 'admin/includes/css/' . $value; ?>">
+        <link rel="stylesheet" href="<?php echo ($GLOBALS['request_type'] == 'SSL' ? DIR_WS_HTTPS_CATALOG : DIR_WS_CATALOG) . $relativeDir . 'admin/includes/css/' . $value; ?>">
         <?php
     }
 
-    if (file_exists($absoluteDir . substr($relativeDir, 1) . 'admin/includes/css/' . basename($PHP_SELF, '.php') . '.css')) {
+    if (file_exists($absoluteDir . $relativeDir . 'admin/includes/css/' . basename($PHP_SELF, '.php') . '.css')) {
 ?>
-        <link rel="stylesheet" href="<?php echo $relativeDir . 'admin/includes/css/' . basename($PHP_SELF, '.php') . '.css'; ?>">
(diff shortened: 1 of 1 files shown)


proseLA:
yeah, i'm not real sure about this.  i really do not have any clients that use sub-directories...  however...

i would think that this change belongs in the file that i previously mentioned, ie:

https://github.com/zencart/zencart/blob/v158/laravel/app/Models/PluginControl.php

the `relativePath` is relative to what?  

it does seem that both `DIR_FS_CATALOG` and `DIR_WS_HTTPS_CATALOG`  (or `DIR_WS_CATALOG`) contain the sub-directory when installed in a sub-directory (at least as far as i can tell).  so that could be a problem when getting the `absolutePath` in that same script.

but are you not going to run into the same problem here:

https://github.com/zencart/zencart/blob/v158/admin/includes/javascript_loader.php

specifically lines, 55, 65, 75 and 82?

i am **NOT** a fan of duplicating code the way that you are doing.  i would think the method would return the proper `relativePath` for use; and that same script should also return the proper `absolutePath`.

in addition, we are in 2022.  most stores use SSL for all pages.  even ignoring that, can someone tell me when these 2 vars might be different:

`DIR_WS_HTTPS_CATALOG` and `DIR_WS_CATALOG`?

has anyone seem that in the wild?

any number of issues to discuss....

brittainmark:
Yes was thinking same re javascript. If I change PluginControl.php I think I would have to change javascript_loader.php and admin_html_head.php as well. Will have a look. Need to understand the  javascript_loader.php I was getting confused when I looked at it yesterday. It was not making sence to me.

brittainmark:
Not correct solution

PR #5040, opened 2022-07-21
v158: product_notifications sidebox, use zen_is_logged_in ...

lat9:
... and zen_in_guest_checkout; includes refactoring.

### DIFF
--- a/includes/modules/sideboxes/product_notifications.php
+++ b/includes/modules/sideboxes/product_notifications.php
@@ -9,49 +9,51 @@
  */
 
 // test if box should show
-  $show_product_notifications = false;
+$show_product_notifications = false;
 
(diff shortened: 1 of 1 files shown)


proseLA:
call me a PITA.

in v158, we now have a customer class (not present in v157).

if we want to refactor, lets move all of the db calls with regards to the customer to the customer class.

thoughts?

lat9:
> call me a PITA.
> 
> in v158, we now have a customer class (not present in v157).
> 
> if we want to refactor, lets move all of the db calls with regards to the customer to the customer class.
> 
> thoughts?

Agreed, but that's a separate PR.  This one simply corrects the sidebox so that it's not picking up guest-customer notifications.

PR #5043, opened 2022-07-23
v158: Enabling pagination of `account_history` page

### LINKED ISSUE #5042: v158: account_history page no longer has pagination
In previous versions of Zen Cart, the `account_history` page limits the count of orders displayed to 'MAX_DISPLAY_ORDER_HISTORY' and paginates.

In zc158, the page now displays all orders with no pagination available since the history information is returned by a call to `Customer::getOrderHistory`.  That class method doesn't export the SQL used for the query, so it's not possible to pass that on to the split_page_results class.

I'll note also that the Bootstrap template counts on that base SQL to be returned so that it can use its own split_page_results class to format the output using Bootstrap classes.

lat9:
Fixes #5042

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -493,7 +493,7 @@ class Customer extends base
         return $addressArray;
     }
 
-    public function getOrderHistory(int $max_number_to_return = 0): array
+    public function getOrderHistory(int $max_number_to_return = 0, &$returned_history_split = null): array
     {
         $language = $_SESSION['languages_id'];
         global $db, $currencies;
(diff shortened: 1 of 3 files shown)


scottcwilson:
This is a mandatory change for all templates, right?  Could we avoid this by creating a new method in the Customer class that would return data in the 157 format?  Does doing that make sense?

lat9:
That's actually not a _change_ to the templates, it's correcting the zc158 version to perform the pagination that has occurred in all (?) previous `template_default` versions.

scottcwilson:
Sorry, let me rephrase: The query in `includes/modules/pages/account_history/header_php.php` has changed.  Will an old template still work with the new data returned by this query?

lat9:
They will _now_.  There was a change required to the Bootstrap template's additional `header_php_`, but the template itself did not require change

PR #5061, opened 2022-07-27
v158: Adding ENT_COMPAT - default changed in PHP 8.1

### LINKED ISSUE #4449: [To Do] PHP 8.1 compatibility
The following will need to be addressed for PHP 8.1 compatibility.

**Deprecated in PHP 8.1:**
- [x] `strftime()`. Ref: https://wiki.php.net/rfc/deprecations_php_8_1#strftime_and_gmstrftime
  _**Impact**: Used extensively. Alternatives could be messy._
  Probably best to retire-and-replace `ADMIN_NAV_DATE_TIME_FORMAT`, `DATE_FORMAT_LONG`, `DATE_FORMAT_SHORT` constants with new names since format syntax will likely be different.
  Also, affects all (including non-core) credit card payment modules which display a dropdown of month/year selections.
 
- [x] `FILTER_SANITIZE_STRING`. Ref: https://wiki.php.net/rfc/deprecations_php_8_1#filter_sanitize_string
  _**Impact**: 2 use-cases, which can probably be easily rewritten._

**Changed in 8.1:**
- [x] `htmlspecialchars(), htmlentities(), htmlspecialchars_decode(), html_entitity_decode() and get_html_translation_table()` now use `ENT_QUOTES | ENT_SUBSTITUTE` rather than `ENT_COMPAT` by default. This means that `'` is escaped to `&#039;` while previously it was left alone. Additionally, malformed UTF-8 will be replaced by a Unicode substitution character instead of resulting in an empty string. Ref https://news-web.php.net/php.internals/112782
  _**Impact**: There are only a half-dozen places where no default is specified. Adding specificity is probably straightforward._

- [x] Changed default `mysqli` Error Mode from `silent` to `exception`. Ref: https://wiki.php.net/rfc/mysqli_default_errmode
  _**Impact**: investigation required. Simply overriding the default in `query_factory` may be enough._

**Unrelated to official core code (but may affect some plugins, like image-handler):**
- [x] `file_info` resources converted to objects (started in 8.0, more in 8.1). Ref https://github.com/php/php-src/commit/b85f0d1cd0cc49ad339b8129c3d45181b56eff24

lat9:
As noted in https://www.php.net/manual/en/changelog.strings.php 

Fixes #4449

### DIFF
--- a/admin/featured.php
+++ b/admin/featured.php
@@ -535,7 +535,10 @@ if (!empty($action)) {
                             }
                             $contents[] = ['text' => TEXT_INFO_LAST_MODIFIED . ' ' . zen_date_short($fInfo->featured_last_modified)];
                             $contents[] = ['text' => TEXT_INFO_DATE_ADDED . ' ' . zen_date_short($fInfo->featured_date_added)];
-                            $contents[] = ['align' => 'text-center', 'text' => zen_info_image($fInfo->products_image, htmlspecialchars($fInfo->products_name), SMALL_IMAGE_WIDTH, SMALL_IMAGE_HEIGHT)];
+                            $contents[] = [
+                                'align' => 'text-center',
+                                'text' => zen_info_image($fInfo->products_image, htmlspecialchars($fInfo->products_name, ENT_COMPAT, CHARSET, true), SMALL_IMAGE_WIDTH, SMALL_IMAGE_HEIGHT)
+                            ];
                             $contents[] = [
                                 'align' => 'text-center',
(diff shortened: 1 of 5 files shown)


scottcwilson:
Is this htmlspecialchars change also needed in these places?

admin/includes/application_bootstrap.php line 35
admin/includes/functions/general.php line 671

lat9:
> Is this htmlspecialchars change also needed in these places?
> 

- admin/includes/application_bootstrap.php line 35 (previously missing, just pushed)
- admin/includes/functions/general.php line 671 ('twas part of the initial commit)

PR #5065, opened 2022-07-28
reformat to PSR standard ot_tax

proseLA:
no changes.  just a reformat.

### DIFF
--- a/includes/modules/order_total/ot_tax.php
+++ b/includes/modules/order_total/ot_tax.php
@@ -1,94 +1,101 @@
 <?php
-/**
- * ot_tax order-total module
- *
- * @copyright Copyright 2003-2022 Zen Cart Development Team
- * @copyright Portions Copyright 2003 osCommerce
- * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: DrByte 2020 Jul 10 Modified in v1.5.8-alpha $
- */
-  class ot_tax {
-    var $title, $output;
 
-    function __construct() {
-      $this->code = 'ot_tax';
-      $this->title = MODULE_ORDER_TOTAL_TAX_TITLE;
-      $this->description = MODULE_ORDER_TOTAL_TAX_DESCRIPTION;
-      $this->sort_order = defined('MODULE_ORDER_TOTAL_TAX_SORT_ORDER') ? MODULE_ORDER_TOTAL_TAX_SORT_ORDER : null;
-      if (null === $this->sort_order) return false;
+    /**
+     * ot_tax order-total module
+     *
+     * @copyright Copyright 2003-2022 Zen Cart Development Team
+     * @copyright Portions Copyright 2003 osCommerce
+     * @license   http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+     * @version   $Id: DrByte 2020 Jul 10 Modified in v1.5.8-alpha $
+     */
+    class ot_tax
+    {
+        var $title, $output;
 
-      $this->output = array();
-    }
-
-    function process() {
-      global $order, $currencies;
-
-      $taxDescription = '';
-      $taxValue = 0;
-      if (STORE_TAX_DISPLAY_STATUS == 1)
-      {
-        $taxAddress = zen_get_tax_locations();
-        $result = zen_get_all_tax_descriptions($taxAddress['country_id'], $taxAddress['zone_id']);
-        if (count($result) > 0)
+        function __construct()
         {
-          foreach ($result as $description)
-          {
-            if (!isset($order->info['tax_groups'][$description]))
-            {
-              $order->info['tax_groups'][$description] = 0;
+            $this->code = 'ot_tax';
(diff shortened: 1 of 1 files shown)


includes/modules/order_total/ot_tax.php
```diff
@@ -1,94 +1,101 @@
...
+     * ot_tax order-total module
+     *
+     * @copyright Copyright 2003-2022 Zen Cart Development Team
+     * @copyright Portions Copyright 2003 osCommerce
+     * @license   http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+     * @version   $Id: DrByte 2020 Jul 10 Modified in v1.5.8-alpha $
+     */
+    class ot_tax
+    {
+        var $title, $output;
```
lat9, line 13: Use `public` instead of `var`

includes/modules/order_total/ot_tax.php
```diff
@@ -1,94 +1,101 @@
...
-      global $order, $currencies;
-
-      $taxDescription = '';
-      $taxValue = 0;
-      if (STORE_TAX_DISPLAY_STATUS == 1)
-      {
-        $taxAddress = zen_get_tax_locations();
-        $result = zen_get_all_tax_descriptions($taxAddress['country_id'], $taxAddress['zone_id']);
-        if (count($result) > 0)
+        function __construct()
```
lat9, line 15: Add `public` to this and all other methods

includes/modules/order_total/ot_tax.php
```diff
@@ -1,94 +1,101 @@
...
-          if ($value > 0 || ($value == 0 && STORE_TAX_DISPLAY_STATUS == 1))
-          {
-            $taxDescription .= ((is_numeric($key) && $key == 0) ? TEXT_UNKNOWN_TAX_RATE :  $key) . ' + ';
-            $taxValue += $value;
-          }
+            global $order, $currencies;
+
+            $taxDescription = '';
+            $taxValue = 0;
+            if (STORE_TAX_DISPLAY_STATUS == 1) {
```
lat9, line 34: Use `=== '1'` here and for all other similar comparisons

includes/modules/order_total/ot_tax.php
```diff
@@ -1,94 +1,101 @@
...
+                $taxAddress = zen_get_tax_locations();
+                $result = zen_get_all_tax_descriptions($taxAddress['country_id'], $taxAddress['zone_id']);
+                if (count($result) > 0) {
```
lat9, line 37: Use `!== 0`

includes/modules/order_total/ot_tax.php
```diff
@@ -1,94 +1,101 @@
...
+                            $order->info['tax_groups'][$description] = 0;
+                        }
+                    }
+                }
+            }
+            if (count($order->info['tax_groups']) > 1 && isset($order->info['tax_groups'][0])) {
+                unset($order->info['tax_groups'][0]);
+            }
+            foreach ($order->info['tax_groups'] as $key => $value) {
+                if (SHOW_SPLIT_TAX_CHECKOUT == 'true') {
```
lat9, line 49: Use `===`

includes/modules/order_total/ot_tax.php
```diff
@@ -1,94 +1,101 @@
...
+                        ];
+                    }
+                } else {
+                    if ($value > 0 || ($value == 0 && STORE_TAX_DISPLAY_STATUS == 1)) {
+                        $taxDescription .= ((is_numeric($key) && $key == 0) ? TEXT_UNKNOWN_TAX_RATE : $key) . ' + ';
+                        $taxValue += $value;
+                    }
+                }
+            }
+            if (SHOW_SPLIT_TAX_CHECKOUT != 'true' && ($taxValue > 0 or STORE_TAX_DISPLAY_STATUS == 1)) {
```
lat9, line 64: Use `!== 'true'`; replace `or` with `||`

lat9 (changes requested):

includes/modules/order_total/ot_tax.php
proseLA: why?  is this standard/style documented?

scottcwilson: We are trying to encourage strict style to prevent logs, etc.  

https://docs.zen-cart.com/dev/contributing/coding_standards/#strict-comparisons

proseLA: the code review suggests `!== 0` is preferable to `> 0` (unless i am reading that incorrectly).

i'm not sure where that is documented in a strict comparison.  count always returns an integer.  and `>` looks to be a strict comparison.

https://www.php.net/manual/en/language.operators.comparison.php#:~:text=%24a%20%3E%20%24b,greater%20than%20%24b.

lat9: It's just that `count` is always going to return a number 0 or greater than.

lat9: I guest it goes back to my assembler programming days, checking for equality has a performance advantage over greater/less than.

I don't care anymore which one you choose.

PR #5067, opened 2022-07-28
add stict typing and function visibility

proseLA:
addresses comments in #5065.

i'm just not sure about changing conditionals being the same as what is contained here in the docs:

https://docs.zen-cart.com/dev/contributing/coding_standards/#when-to-reformat-an-entire-file

which is what i was trying to do.

### DIFF
--- a/includes/modules/order_total/ot_tax.php
+++ b/includes/modules/order_total/ot_tax.php
@@ -1,5 +1,6 @@
 <?php
 
+
     /**
      * ot_tax order-total module
      *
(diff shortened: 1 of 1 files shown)


includes/modules/order_total/ot_tax.php
```diff
@@ -46,22 +49,22 @@ function process()
                 unset($order->info['tax_groups'][0]);
             }
             foreach ($order->info['tax_groups'] as $key => $value) {
-                if (SHOW_SPLIT_TAX_CHECKOUT == 'true') {
-                    if ($value > 0 or ($value == 0 && STORE_TAX_DISPLAY_STATUS == 1)) {
+                if (SHOW_SPLIT_TAX_CHECKOUT === 'true') {
+                    if ($value > 0 || ($value === 0 && STORE_TAX_DISPLAY_STATUS === '1')) {
```
lat9, line 53: Since the $value is a float, this is a case to use kinda-sorta-equal-to instead of a strict comparison.  Here and on line 61.

proseLA: good point!.

Issue #5066, opened 2022-07-28
includes\classes\productPulldown.php has no id info/comments


torvista:

scottcwilson:
Same issue for these files: 
```
ProductConfigurationSwitch.php
categoryPulldown.php
productOptionsPulldown.php
pulldown.php
```

PR #5068, opened 2022-07-29
add documentation; reformat to default from cs.symfony.com

proseLA:
#5066

### DIFF
--- a/includes/classes/categoryPulldown.php
+++ b/includes/classes/categoryPulldown.php
@@ -1,9 +1,15 @@
 <?php
 
+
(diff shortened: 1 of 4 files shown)


scottcwilson:
This doesn't close #5066 - the issue that was being reported there was that the comment block at the top of the file that is present in other files is not present in productPulldown.php.

proseLA:
so it's a problem with the version stamper?

great.

scottcwilson:
It just needs a default block of comments to start with.  I will fix.

PR #5073, opened 2022-08-01
v158: Enable 'product_music' to reuse base 'product' processing

### LINKED ISSUE #3551: Improvement: remove need for update_product.php in music type
Since we now have a notifier for adding extra date to the products database table, in 
"admin/includes/modules/update_product.php". We can delete the file from the music folder, and add the changes to an observer file.
This removes the need for updating multiple almost identical files, and provide an example file for users that need to create their own observer files.

lat9:
Fixes #3551


torvista:
Why not do the same thing for document_product and product_free_shipping: they are identical to product.

PR #5074, opened 2022-08-01
v158: Use correct define for products-all/new listings

### LINKED ISSUE #5072: shopfront TEXT_DATE_ADDED: multiple places where date substitution is not implemented
One example
Clipboard01

searching for uses of TEXT_DATE_ADDED shows others.

lat9:
Fixes #5072


dbltoe:
So what do I do since this affects my changes in #5070.

PR #5076, opened 2022-08-01
V158 - currency cron: remove extra semicolon...

torvista:
replace dirname(__FILE__) with __DIR__,
strict comparisons


scottcwilson:
Would you please Squash these three commits?

PR #5081, opened 2022-08-03
ZC158: fix typos - hyphens instead of equals for titles

torvista:


proseLA:
while this fix is clearly needed, i would suggest @dbltoe fix his stuff.  and i would also suggest @swguy do some testing prior to merging.

why would we have a title that equates to the `product_id`?  is that really a title?

in addition, this line now has an extra `$`:

https://github.com/zencart/zencart/blob/4b90a4035b533320747d4e1be40af402b8c0199e/includes/modules/responsive_classic/product_listing.php#L224

that is causing a fatal error:

```
[03-Aug-2022 08:13:47 America/Los_Angeles] Request URI: /index.php?main_page=search_result&search_in_description=1&keyword=bug, IP address: 192.168.14.74
--> PHP Fatal error: Uncaught TypeError: Argument 2 passed to zen_get_buy_now_button() must be of the type string, null given, called in /var/www/zcdev/includes/modules/responsive_classic/product_listing.php on line 327 and defined in /var/www/zcdev/includes/functions/functions_general.php:139
Stack trace:
```

torvista:
Should be ok now.

proseLA:
but what was `title="24"` mean?

torvista:
I just fixed the syntax, but yes, the title should have something relevant in it, like "link to the product description for" with the product name.

proseLA:
which is why @dbltoe should be fixing it. IMO

dbltoe:
Yes, it was my fault that I got a couple of files mixed up with a previous attempt at getting the accessibility corrected.

And yes, I should be fixing my own errors.  I was not able to test and hindsight (always 20/20) says I should have tested first.  I was unaware of the problem and would have fixed it had I known.

@torvista Your suggested addition is an excellent idea and actually might help.  The more info is thrown at the SEs, the better.

Unless someone wishes to jump in, I will look at the fix when allowed hours at the computer later today.

Sorry for the hassle but, I think with the suggestion from Steve, we come away with an even better solution. 👍 

Perhaps I should get my Pobody's Nerfect sign out again. 😃

torvista:
Sometimes I can't be bothered to document a trivial bug like this, easier to fix it. I've no doubt the culprit feels the pain, and I also feel that it's better for someone to submit a faulty PR than none at all. Ideally it gets peer-reviewed and the submitter/others learn something.

dbltoe:
As to my thinking on using the product name, I guess I was assuming too much in thinking someone would want to put full information in the product's name.  I know I probably would have used "24 - The Drama Series Starring Keifer Sutherland".

Still, even that will look better with Steve's suggestion in front.

scottcwilson:
This is fine.  Thanks for the fast fix.  Everyone makes mistakes.

proseLA:
everyone does make mistakes.

if you think having a more info hyperlink whose title is solely a number is fine, i think we have a different opinion as to what fine is.

PR #5083, opened 2022-08-03
update to MobileDetect 2.8.39

torvista:


scottcwilson:
Did you test this?

torvista:
It's in use on my site. 
I don't have a copy of the debug the old version raised but it was the usual passing null thing, which now has a clause to trap it.

PR #5084, opened 2022-08-04
v158: zen_image updates for accessibility and ...

### LINKED ISSUE #5063: Improve Accessibility for Sideboxes
When the sideboxes for New Products, Specials, and Featured are generated, if an items fits the qualifications, the link and information are presented in <div class="sideboxContentItem" and generate an A link using the image as the text for the link,

The image data for the Bug' Life DVD comes out to be:
`A Bug's Life Linked<br>"A Bug's Life Linked"</a>`

The system is taking the Product's Title and uses it for the alt tag, the image title, and the displayed link text underneath.

This generates an alert for Redundant Alternativ Text.  This confuses browser readers and might make the SEs lower rank for keyword stuffing.

If the link is changed to the following, all Accessibility standards would be met.

`<br>"A Bug's Life Linked"</a>`

Most all browsers see the empty alt tag as being presentational.  The newer browsers will accept the role tag.  This will result in about 35 alerts on the main page of the out-of-the-box going away.

I'm assuming the changes would be made in (e.g. Featured) at line 17 of includes/template_default/sideboxes/tpl_featured.php.

Since these are different from the items in the center boxes AND I am not knowledgeable enough to figure the code.  I enter it here.

If someone presents the proper code change, I will do the PR and hopefully not mess it up. 😄 

FYI -- It appears that the center page layout for Featured and others uses the same tpl.

lat9:
- Refactoring to use PSR indentation and exactly-equal-to
- Reduce number of times a file's existence is determined (performance)
- Use is_file instead of file_exists; file_exists indicates existence for directories, too
- Logic restructuring to remove redundant/duplicated checks

Fixes #5063

### DIFF
--- a/includes/functions/html_output.php
+++ b/includes/functions/html_output.php
@@ -114,170 +114,189 @@ function zen_catalog_href_link($page = '', $parameters = '', $connection = 'NONS
  * The HTML image wrapper function for non-proportional images
  * used when "proportional images" is turned off or if calling from a template directory
  */
-  function zen_image_OLD($src, $alt = '', $width = '', $height = '', $parameters = '') {
+function zen_image_OLD($src, $title = '', $width = '', $height = '', $parameters = '')
+{
     global $template_dir;
 
-//auto replace with defined missing image
-    if ($src == DIR_WS_IMAGES and PRODUCTS_IMAGE_NO_IMAGE_STATUS == '1') {
-      $src = DIR_WS_IMAGES . PRODUCTS_IMAGE_NO_IMAGE;
+    //auto replace with defined missing image
+    if ($src === DIR_WS_IMAGES && PRODUCTS_IMAGE_NO_IMAGE_STATUS === '1') {
+        $src = DIR_WS_IMAGES . PRODUCTS_IMAGE_NO_IMAGE;
     }
 
-    if ( (empty($src) || $src == DIR_WS_IMAGES) && (IMAGE_REQUIRED == 'false') ) {
-      return false;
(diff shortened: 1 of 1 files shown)


dbltoe:
While not fully understanding the code, I can see where the items that were discussed elsewhere are better used here.  Especially the use of ratio.
👍

brittainmark:
What happens when $width is set to a % not sure code will work correctly.

lat9:
> What happens when $width is set to a % not sure code will work correctly.

The only time that a % width/height works is if Image Handler is installed.  The current/legacy version also casts those variables to ints:

https://github.com/zencart/zencart/blob/614b2d6f860339ed81e6e39026a89ece8b72e426/includes/functions/html_output.php#L213-L216

Although that implementation also keeps a `''` as part of the `empty` path without conversion.

brittainmark:
Yes saw that. The reason for the question is 'HEADER_LOGO_WIDTH' => '100%'  exists in lang.header.php
I have checked, and it does not appear to be used in the main set of store files. 
However, it gives the impression that % is ok for a width. Maybe need to remove from store language files.

lat9:
> Yes saw that. The reason for the question is 'HEADER_LOGO_WIDTH' => '100%' exists in lang.header.php I have checked, and it does not appear to be used in the main set of store files. However, it gives the impression that % is ok for a width. Maybe need to remove from store language files.

Hmm, that changed since zc157.  I'll open a separate PR to restore that to the pixel width/height of the default logo.gif.

PR #5086, opened 2022-08-05
v158: Enable extra content to be added to the shopping-cart page

lat9:
... as added by additional header_php_*.php files in `/includes/modules/pages/shopping_cart`.


lat9:
After further consideration, I don't like this implementation.

PR #5089, opened 2022-08-05
category-product listing basic review

torvista:
mainly strict comparisons  
null-coalesce  
expansion of a nested ternary operator  
removal of unnecessary checks of $action
removal of optional/default parameters from function calls
short array syntax

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /**
  * @copyright Copyright 2003-2022 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
@@ -10,9 +10,15 @@ $languages = zen_get_languages();
 require DIR_WS_CLASSES . 'currencies.php';
 $currencies = new currencies();
 
-$product_type = (isset($_POST['products_id']) ? zen_get_products_type($_POST['products_id']) : (isset($_GET['product_type']) ? $_GET['product_type'] : 1));
+if (isset($_POST['products_id'])) {
+    $product_type = zen_get_products_type($_POST['products_id']);
+} elseif (isset($_GET['product_type'])) {
+    $product_type = $_GET['product_type'];
+} else {
+    $product_type = 1;
+}
 
-$action = (isset($_GET['action']) ? $_GET['action'] : '');
+$action = $_GET['action'] ?? '';
 $search_result = isset($_GET['search']) && zen_not_null($_GET['search']);
 
 $search_parameter = $search_result ? '&search=' . $_GET['search'] : '';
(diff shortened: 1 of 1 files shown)


admin/category_product_listing.php
```diff
@@ -184,7 +190,7 @@
       zen_redirect(zen_href_link(FILENAME_CATEGORY_PRODUCT_LISTING, 'cPath=' . $cPath));
       break;
     case 'move_category_confirm':
-      if (isset($_POST['categories_id']) && ($_POST['categories_id'] != $_POST['move_to_category_id'])) {
+      if (isset($_POST['categories_id']) && ($_POST['categories_id'] !== $_POST['move_to_category_id'])) {
```
lat9, line 193: Strict comparison will fail, since the admin sanitization has cast 'categories_id' to an int, but not 'move_to_category_id'.

admin/category_product_listing.php
```diff
@@ -397,11 +401,11 @@
             $zc_skip_products = false;
             $zc_skip_categories = false;
 
-            if ($check_products == 0) {
+            if ($check_products === 0) {
               $zc_skip_products = false;
               $zc_skip_categories = false;
             }
-            if ($check_categories == true) {
+            if ($check_categories) {
```
lat9, line 408: $check_categories is a (bool) return from `zen_has_category_subcategories`; use `=== true`.

admin/category_product_listing.php
```diff
@@ -679,30 +683,30 @@
                   <td class="hidden-sm hidden-xs"><!-- no model for categories --></td>
                   <td class="hidden-sm hidden-xs"><!-- no price for categories --></td>
                 <?php } ?>
-                <?php if ($search_result || SHOW_COUNTS_ADMIN == 'true') { ?>
+                <?php if ($search_result || SHOW_COUNTS_ADMIN === 'true') { ?>
```
lat9, line 686: `$search_result` doesn't 'participate' in this section, since output is only rendered if `SHOW_COUNTS_ADMIN === 'true'` and can be removed from this statement.

admin/category_product_listing.php
```diff
@@ -679,30 +683,30 @@
...
                   <td class="text-right hidden-sm hidden-xs">
                     <?php
-                    if (SHOW_COUNTS_ADMIN == 'true') {
+                    if (SHOW_COUNTS_ADMIN === 'true') {
```
lat9, line 689: Can be removed; see comment for line 686.

admin/category_product_listing.php
```diff
@@ -679,30 +683,30 @@
...
                       // show counts
                       $total_products = zen_get_products_to_categories($category['categories_id'], true);
-                      $total_products_on = zen_get_products_to_categories($category['categories_id'], false);
+                      $total_products_on = zen_get_products_to_categories($category['categories_id']);
                       echo $total_products_on . TEXT_PRODUCTS_STATUS_ON_OF . $total_products . TEXT_PRODUCTS_STATUS_ACTIVE;
                     }
```
lat9, line 694: Remove also the closing brace.

admin/category_product_listing.php
```diff
@@ -679,30 +683,30 @@
...
                     ?>
                   </td>
                 <?php } ?>
                 <td class="text-right dataTableButtonCell">
-                  <?php if (SHOW_CATEGORY_PRODUCTS_LINKED_STATUS == 'true' && zen_get_products_to_categories($category['categories_id'], true, 'products_active') == 'true') { ?>
+                  <?php if (SHOW_CATEGORY_PRODUCTS_LINKED_STATUS === 'true' && zen_get_products_to_categories($category['categories_id'], true, 'products_active') == 'true') { ?>
```
lat9, line 699: Why not `===` for the check on zen_get_products_to_categories?

admin/category_product_listing.php
```diff
@@ -816,17 +820,17 @@
...
 
 // reset page when page is unknown
-            if ((empty($_GET['page']) || $_GET['page'] == '1') && !empty($_GET['pID'])) {
+            if ((empty($_GET['page']) || $_GET['page'] === 1) && !empty($_GET['pID'])) {
               $check_page = $db->Execute($products_query_raw);
               if ($check_page->RecordCount() > $max_results) {
                 $check_count = 0;
                 foreach ($check_page as $item) {
-                  if ($item['products_id'] == $_GET['pID']) {
+                  if ($item['products_id'] === $_GET['pID']) {
```
lat9, line 828: `$_GET['pID']` gets sanitized to an integer; the exactly-equal-to compare will never 'hit'.

admin/category_product_listing.php
```diff
@@ -816,17 +820,17 @@
...
                     break;
                   }
                   $check_count++;
                 }
-                $_GET['page'] = round((($check_count / $max_results) + (fmod_round($check_count, $max_results) != 0 ? .5 : 0)), 0);
+                $_GET['page'] = round((($check_count / $max_results) + (fmod_round($check_count, $max_results) !== 0 ? .5 : 0)));
```
lat9, line 833: This one's tough, since `fmod_round` is going to return either an int or a float; I'll suggest leaving the comparison as a lazy one.

admin/category_product_listing.php
```diff
@@ -843,7 +847,8 @@
                 $cPath = $product['master_categories_id'];
               }
 
-              if ((!isset($_GET['pID']) && !isset($_GET['cID']) || (isset($_GET['pID']) && ($_GET['pID'] == $product['products_id']))) && !isset($pInfo) && !isset($cInfo) && (substr($action, 0, 3) != 'new')) {
+              if ((!isset($_GET['pID']) && !isset($_GET['cID']) || (isset($_GET['pID']) && ($_GET['pID'] === $product['products_id']))) && !isset($pInfo) && !isset($cInfo) && (strpos($action, 'new')
```
lat9, line 850: Combine `!isset($_GET['pID']) && !isset($_GET['cID'])` into `!isset($_GET['pID'], $_GET['cID'])`

lat9 (changes requested):
The admin sanitizer doesn't help sometimes in determining when various $_GET/$_POST values should be checked for their submitted string values!

admin/category_product_listing.php
```diff
@@ -114,12 +120,12 @@
...
       // future cat specific deletion
       $delete_linked = 'true';
-      if (isset($_POST['delete_linked']) && $_POST['delete_linked'] != '') {
+      if (isset($_POST['delete_linked']) && $_POST['delete_linked'] !== '') {
         $delete_linked = $_POST['delete_linked'];
       }
 
       // delete category and products
-      if (!empty($_POST['categories_id']) && is_numeric($_POST['categories_id']) && $_POST['categories_id'] != TOPMOST_CATEGORY_PARENT_ID) {
+      if (!empty($_POST['categories_id']) && is_numeric($_POST['categories_id']) && $_POST['categories_id'] !== TOPMOST_CATEGORY_PARENT_ID) {
```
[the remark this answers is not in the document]
torvista: Then no need for subsequent line either ?
      `  $categories_id = zen_db_prepare_input($_POST['categories_id']);`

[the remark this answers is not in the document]
torvista: https://github.com/zencart/zencart/issues/5091

admin/category_product_listing.php
torvista: That function needs attention in general...

But, currently returns string '' or string 'true' or an integer count (function description is incorrect).
In this case, using 'products_active', it returns 'true'.

admin/category_product_listing.php
torvista: This I was leaving separately as I couldn't see if the logic was correct...similarly for the similar test prior to
$cInfo = new objectInfo($category);

torvista:
These are redundant? But product_type is in SIMPLE_ALPHANUM_PLUS.

```
if (isset($_GET['page'])) {
  $_GET['page'] = (int)$_GET['page'];
}
if (isset($_GET['product_type'])) {
  $_GET['product_type'] = (int)$_GET['product_type'];
}
if (isset($_GET['cID'])) {
  $_GET['cID'] = (int)$_GET['cID'];
}
```

admin/category_product_listing.php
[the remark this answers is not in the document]
lat9: Correct, since that variable's pre-santized.

lat9 (approved):

PR #5097, opened 2022-08-07
Add Newer Favicon and Associated Files

dbltoe:
This brings the favicon into 2022 standards to match the current logo and add the icons for desktop/mobile/tablet use.


drbyte:
In the storefront template it's probably better to use a generic icon, since this is a couple dozen files the store owner will have to update. 
Maybe copy the one from zc_install instead?

dbltoe:
There's a free builder that creates these and I will be adding instructions in the docs.

I think it behooves us to show what can be done and explain how in the docs.

dbltoe:
If the store owner uses or clones template_default or responsive_classic, they will not have to touch the template files.  Just uploading the images to the root will suffice.

PR #5100, opened 2022-08-09
Fix Country Codes in the Database

dbltoe:
In checking accessibility deeper in the shopping cart today, I kept getting errors on the country code 223.  Turns out ALL the country codes have changed AND the mysql_utf8.sql had errors that resulted in three tables not being propagated.

This PR corrects the un-escaped apostrophes that caused mysql_utf8.sql to not load and corrects all the country codes/names as well.

The main mysql_zencart.sql is changed to replace the old 223 US code with the new 840.

File mysql_demo.sql is changed to reflect the 840 country code and I put in the name of a friend who passed several years ago of cancer.  He was my family's handyman while I was remote for a year.  Everything from clearing commodes to carting kids to practice, he did it all.  I thought it would be better to show more accurate data in the one default customer and, they say you're not really gone till they forget your name.


dbltoe:
I guess I'm still a ways away from getting the github thing.

I thought I did two separate PRs but got them munged together 🤕

torvista:
>I thought I did two separate PRs but got them munged together

Sometimes it's still black magic to me too when I don't understand the results of what I just did, no matter how many tutorials I read!
 Hence I try and do one at a time, which is frustrating when sometimes they take a long time to get rebuffed or merged.

lat9:
The country-codes should not have changed, in fact the initial install for zc158 has country code 248 as the highest.  Many, many internals and plugins for Zen Cart will roll over and die if those codes were changed.

The real question is, for the database you're using, how the United States became 840 in the first place.

dbltoe:
> The real question is, for the database you're using, how the United States became 840 in the first place.
As stated, I kept getting that 223 was invalid and the Countries page would not load.  Right on the countries page is the link to the current codes.
https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements
I searched the database and 223 was only found in the files that I changed.

lat9:
Right, but what is the history of that database?  I'll note that the country-codes in that wiki article, unfortunately, have no bearing on the codes used in a Zen Cart `countries` table.

lat9:
So it sounds like there's an underlying issue where the zc_install's initial installation process in the database might need revisiting to ensure that various tables are dropped during the initial/re-install processing.

I'll note that the requested changes to the country-codes is a non-starter; there are too many places in the base Zen Cart as well as various plugins and site-specific changes that 'know' that the US is country-code 223.

lat9:
At a minimum, and in a separate PR, the help text for the Admin >> Countries page should be updated to indicate that the country-codes in that wiki article are not what's used in Zen Cart.

dbltoe:
Are we not at some point going to have to meet the ISO?  It's been in effect since November of 2015?

I realize it could be a bear when upgrading.  Most listings in the database are either from grabbing the store's default country or the customer selecting their country from the list.  USPS and PayPal seem to follow this.

This is the kind of "put the lug nuts on the cars" assembly line work that I can do if it helps.  😃

proseLA:
i am a huge fan of [country code standards](https://www.iban.com/country-codes) which does have the US as 840; but frankly, if we were to make a change, using the ISO 3 letter abbreviation makes much more sense to me.  which is already in our countries table.

i am also a fan of the use of [standardized state codes](https://www.iso.org/obp/ui/#iso:code:3166:US) for state pulldowns as opposed to numeric indices that we currently use.  which similar to the numeric country codes are just legacy indices with no connection to any standard.

i will note this has absolutely nothing to do with accessibility.

in addition, replacing the use of 223 in the base code seems trivial to me.  other than the test data, it seems 223 is only hard coded in the pAYpAL modules. 

replacing the hard coded 223 in plugins is another story altogether; and i agree it would inflict a lot of pain to working ZC shops where people maintaining it have no doubt hard coded 223 as opposed to looking up the country index from the db (i myself am guilty as charged).

as far as dean; keep him in your memory as opposed to our test data. but fixing the test data to properly load country names with apostrophes absolutely should be looked at if in fact there is a problem.

dbltoe:
Will do.

Note:  I got the log file while looking at the shopping cart for a final walkthru on accessibility.  Only mention of it was to say how I stumbled upon it.

dbltoe:
Should we consider changing the hard-coded PayPal to STORE_COUNTRY as a beginning?

torvista:
> If we want a better solution it is to update code to access the table through the international standard, e.g. the iso codes

Agreed, it should be so. But unfortunately this is a solution without fixing any real problem.
Either it needs a big bang or piecemeal updates whenever a file gets modified for another reason?
Not sure what is involved for a big bang. As always I think these big changes should be done now on a version change.

PR #5085, opened 2022-08-05
Make admin HTML5 compliant

brittainmark:
I have reworked admin to be html5 compliant and fixed a number of errors
in the code. see below for details.

# general changes:
correct labels to point at id
add table role="listbox"
change/add tr onclick role to "option" and add aria_selected
Change name to id in a tag
remove width and replace with css from \<table>, \<tr>, \<th>,\< td>
replace \<tt> tag  with \<span class="tt">
replace align with css "text-..." in table elements
changed doctype
remove valign from table elements and replace with css where necessary
remove \<button> tag from inside <\a> tags and add css as appropriate

#  specific changes
## admin/coupon_admin.php
fix missing closing tags
fix php warning missing array element 'uid'
move hidden form fields out of table element

## admin/coupon_admin_export.php
add parameters to redirect to prevent php undefined variables
removed extra table tr td with no closing elements
remove incorrect colspan

## admin/coupon_restrict.php
Added column headings to prevent html error table column not used.

## admin/customer_groups.php
added missing spaces before class

## admin/developers_tool_kit.php
created function to output page head so always written first
added action to variables to make unique
added htmlspecialchars so output does not contain html tags
added additional css

## admin/featured.php
added missing space before class

## admin/group_pricing.php
added missing space before class

## admin/gv_mail.php
corrected position of \</form>

## admin/gv_sent.php
changed !isset to empty to remove php warning

## admin/includes/functions/general.php
use new zen_make_id function to create valid id. correct issue with
spaces in id.

## admin/includes/functions/html_output.php
Change zen_image width and height to be style
Change zen_black_line to use $width instead of style in $parameter
Change zen_draw_seperator to use new zen_draw_image
Change zen_draw_input_field to add id = to name if no id present in $parameters
Change zen_draw_password_field to replace autocomplete with 'off' if in $parameters and $autocomplete is false
Change zen_draw_checkbox_field to add id = to name if no id present in $parameters
Change zen_draw_textarea_field to use style if $width is not numeric and default to 100%
Change zen_draw_pull_down_menu to add id = to name if no id present in $parameters
Change zen_draw_label to not output 'for' if $for is empty
Change zen_draw_date_selector to add id = to name if no id present in $parameters
Added new function zen_make_id to make a valid id field

## admin/includes/modules/copy_product.php
Change to use new zen_image format

## admin/includes/modules/.../collect_info.php
correct inputtype to inputmode

## admin/includes/modules/.../collect_info_metatags.php
[...]

PR #5108, opened 2022-08-11
Correct inputtype to inputmode

brittainmark:
Second breakdown #5085 to make it easier to understand, if
you don't want me to do this please let me know.

### DIFF
--- a/admin/includes/modules/document_general/collect_info.php
+++ b/admin/includes/modules/document_general/collect_info.php
@@ -312,7 +312,7 @@ if (zen_get_categories_status($current_category_id) == 0 && $pInfo->products_sta
           <span class="input-group-addon">
               <?php echo zen_image(DIR_WS_CATALOG_LANGUAGES . $languages[$i]['directory'] . '/images/' . $languages[$i]['image'], $languages[$i]['name']); ?>
           </span>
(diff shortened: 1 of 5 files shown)


scottcwilson:
What's the reason for this change? Can you link to a document where this is explained?

barco57:
I am not sure you replace one for the other, they get used together. When I have seen it used both are included.
`<input type="text" inputmode="tel" />`

[Everything You Ever Wanted to Know About inputmode](https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/)

brittainmark:
the 'inputtype' is a typo it should be 'inputmode' this is used on mobile devices to display the correct keyboard for input.
if you look further down the code you will see that all the others are inputmode not inputtype.

> I am not sure you replace one for the other, they get used together. When I have seen it used both are included. `<input type="text" inputmode="tel" />`
> 
> [Everything You Ever Wanted to Know About inputmode](https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/)

It is not a replacement. The code generated currently is 
`<input type="text" name="products_url[1]" value="www.matrox.com/mga/products/mill_g400/home.htm" size="51" maxlength="255" class="form-control" inputtype="url">`
it should be
`<input type="text" name="products_url[1]" value="www.matrox.com/mga/products/mill_g400/home.htm" size="51" maxlength="255" class="form-control" inputmode="url">`

does that explain enough?

PR #5115, opened 2022-08-12
Fix 5114

### LINKED ISSUE #5114: all music genre displayed as selected
**Describe the bug**
when you first enter music genre in admin all of the music genres are shown as selected.

**Version**
Zen Cart version: 1.5.8
PHP version:  8.0.20

**To Reproduce**
1. Go to admin>extras>music genres
on entry all music genres are shown ewith the triangle to indicate selected.

wrong

**Expected behavior**
the first line should be shown as selected with the triangle. The rest should show the information icon "i" in a circle.
Correct

found as part of #5085.

brittainmark:
fixes #5114
set the if in $_get to stop all roes being marked as selected.

### DIFF
--- a/admin/music_genre.php
+++ b/admin/music_genre.php
@@ -91,7 +91,7 @@ if (!empty($action)) {
                 $music_genres = $db->Execute($music_genre_query_raw);
 
                 foreach ($music_genres as $music_genre) {
-                  if ((!isset($_GET['mID']) || (isset($_GET['mID']) && ($_GET['mID'] == $music_genre['music_genre_id']))) && (substr($action, 0, 3) != 'new')) {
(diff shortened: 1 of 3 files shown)


proseLA:
whatever problem you are trying to solve, i would think there has to be a more elegant way to do it.

other than sanitizing `$_GET` vars, i do not think code should assign them.  especially to something that comes from the database.

zcwilt:
Agree with @proseLA here.

While we may have done this in the past, assigning request variables like this has the potential to hide other security problems.

brittainmark:
Yes understand. Serves me right for being lazy. Have understood the logic now and have a better solution. Will adjust and re-pull.

brittainmark:
Modified to check if the $aInfo has been set. This is done the for the first row if the $mId is not set.  So only the first row will show as selected.

brittainmark:
Just noticed same issue exists with record companies, do you want a separate PR?

PR #5098, opened 2022-08-07
admin, category-product listing PSR-12 formatting

torvista:
and removed declare(strict_types=1) from first line, which was only for testing. I leave it in all the files I mess with, seems like a good idea.
No other code changes

This was formatted by phpStorm so maybe there are things in there that are not to everyone's taste. Good to find out now.....

PR #5123, opened 2022-08-15
ZC158: admin, category-product listing - fix strict comparison

torvista:
As per https://github.com/zencart/zencart/pull/5098#discussion_r944373615

Although this PR fixes a mistake, perhaps there is a further discussion to be had...

In this file, $categories_id is ini-cast to integer.
$categories array is generated from zen_get_category_tree/comes from a query and so the 'id' used in the comparison is a string....except in this instance the first element is the passed parameter $categories_id and so is an integer.

Here I have cast the passed parameter integer to string so the returned array is all strings, and made the subsequent strict  comparison as string.

However, regarding the function zen_get_category_tree:
https://github.com/zencart/zencart/blob/22d7eb768062fdf23ab4a2ace5bea2a101d54aeb/includes/functions/functions_categories.php#L648-L657
the first parameter is documented as being an integer, but default is TOPMOST_CATEGORY_PARENT_ID, which is a string.
As this is used in a query, it is cast to integer in the function. 

So not very elegant. 
(I don't find any other uses of this function that may pass an integer.)

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -1,10 +1,12 @@
-<?php declare(strict_types=1);
+<?php
+
 /**
  * @copyright Copyright 2003-2022 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: torvista 2022 Jul 10 Modified in v1.5.8-alpha $
  */
+
 require 'includes/application_top.php';
 $languages = zen_get_languages();
 require DIR_WS_CLASSES . 'currencies.php';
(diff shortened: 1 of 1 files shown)


scottcwilson:
I would really rather we keep the convention that _id variables are integer rather than going backwards and stringifying them.

torvista:
So, what do you suggest: cast the the first parameter to string inside the function?

torvista:
Still requires a conclusion.

scottcwilson:
I know.  It's on my list.

scottcwilson:
I promise this is still on my list.  Ugh too much to do.

torvista:
rabbit hole

drbyte:
Question: Why is strict-typing even important in this context? 
You're right that the db query returns a string, so all comparisons against db results in these and related functions need to be loosely-typed, not strict.

If it is left as-is, where does something break elsewhere? (other than in an IDE because docblocks say `int`)

torvista:
>does something break elsewhere
Not that I can find a the moment
I think leave as is.

PR #5126, opened 2022-08-15
css: remove redundant units of measure for zero values

torvista:


scottcwilson:
I'd rather not do this because I'm worried rookies and inexperienced people will be confused by it.  Let's leave it.

dbltoe:
For over 10 years, IDE's have balked at 0px, 0em, etc.  Why not be correct and let the rookies and inexperienced learn correct syntax?  It may only be milliseconds but CSS is going to try and build a 0px border until it sees that it is actually supposed to not exist and drops the action.  Every millisecond counts in Time to Interactive.
I think the CSS bloating that naturally happens with complicated sites can only be helped by making sure all the CSS is correctly structured.

torvista:
>Why not be correct and let the rookies and inexperienced learn correct syntax?

Why not indeed?

scottcwilson:
This PR sat for a month with no comments and had no description text explaining why it was important. 
Can you guys please help me out here?  Instead of just making a change and submitting it with zero explanation, and not making a case for why something that doesn't obviously fix a bug is a good idea, take a little extra time and argue for it. 

We'll look at this again after 1.5.8 is released.

torvista:
I considered the title was enough information for the people who inhabit this space.
>For over 10 years, IDE's have balked at 0px, 0em, etc.

PRs should only be for fixing bugs? 100% disagree. That criteria is long obsolete.  
Many of the submissions prior to the current monumental effort by brittainmark are non-bug-related.

And, in a general sense, if **anyone** bothers to make the effort to PR **anything**, even if it does almost nothing, it should be treated as 
a) thanks
b) "why not"
if only to reward the interest and effort, not subjected to arbitrary criteria.

Why people bother to spend their own time altruistically improving the code is anyone's guess but easily dissuaded.

dbltoe:
Likewise, I did not know of a need to validate others' work.  Apologies if I should have given a thumbs up.
Is it a problem to merge still?  I do not see any changes that would be a problem and It really helps in promoting ZC in other forums when you can point out the currency and correctness of the product.
I realize that this would not be seen until loaded and set up but how many times have you been disappointed after fully loading software to be faced with a partially working backend?

PR #5129, opened 2022-08-15
HTML5 and Firefox resize

### LINKED ISSUE #5128: Admin pop-up not HTML5, wrong resize for Firefox
**Describe the bug**
The admin pop-up window called from banner manager is not HTML5. Additionally, when using Firefox the window is resized incorrectly. Not high enough.

**Version**
Zen Cart version: 1.5.8
PHP version: 8.0.20

**To Reproduce**
1. Go to 'admin>tools>banner manager.' using the Firefox browser
2. Click on 'banner icon at start of line'

A popup window appears which is too short and not html5 format.

**Expected behavior**
pop-up window is correct size and is HTML5

brittainmark:
Change doc type to HTML and remove Netscape coding from java script as
causing resize to be too small when using Firefox.
fixes #5128 5128

### DIFF
--- a/admin/denied.php
+++ b/admin/denied.php
@@ -9,7 +9,7 @@
 require('includes/application_top.php');
 
 ?>
-<!doctype html public "-//W3C//DTD HTML 4.01 Transitional//EN">
+<!doctype html">
 <html <?php echo HTML_PARAMS; ?>>
 <head>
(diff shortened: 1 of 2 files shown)


brittainmark:
Added admin/denied.php as also has wrong doctype.

PR #5134, opened 2022-08-16
admin server info HTML5 complient

### LINKED ISSUE #5133: Server info is not HTML5 compliant
**Describe the bug**
The following errors are present.

- The font element is obsolete. Use CSS instead.
- Bad character ( after <. Probable cause: Unescaped <. Try escaping it as &lt;.

Additionally, there are many occurrences of these warnings.

- The border attribute is obsolete. Consider specifying img { border: 0; } in CSS instead.
- The name attribute is obsolete. Consider putting an id attribute on the nearest container instead.
- A table row was 1 columns wide, which is less than the column count established by the first row (2).

**Version**
Zen Cart version: 1.5.8
PHP version: 7.4, 8.0, 8.1 & 8.2

**To Reproduce**
1. Go to 'Admin>version' or 'admin>tools>server version'
2. view page source
3. copy page source 
4. paste into https://validator.w3.org/nu/#textarea
5. click validate

**Expected behavior**

brittainmark:
fix #5133

I have:

- removed border=0 from \ tag
- changed the doctype to html
- cleaned up the output from phpinfo
   - changed \<font>\</font> to \<span>\</span>
   - replaced name= with id= in \<a> tags
   - removed border=0 from \ tags
   - added a colspan for one occurrence of table row missing \<td>element
- Added htmlspecialchars to the output of system variables

I tested against php 7.4, 8.0, 8.1, 8.2

I had to define a lot of variables in the classes for 8.2. I just set them to public, and it worked fine. (not included in this release)

### DIFF
--- a/admin/server_info.php
+++ b/admin/server_info.php
@@ -13,7 +13,7 @@
 
 // the following is for display later
   $sinfo =  '<div class="sysinfo wrapper">' .
-         '  <div class="center"><a href="https://www.zen-cart.com"><img border="0" src="images/small_zen_logo.gif" alt=" Zen Cart "></a></div>' .
+         '  <div class="center"><a href="https://www.zen-cart.com"><img src="images/small_zen_logo.gif" alt=" Zen Cart "></a></div>' .
          '  <div class="center"><h2> ' . PROJECT_VERSION_NAME . ' ' . PROJECT_VERSION_MAJOR . '.' . PROJECT_VERSION_MINOR . '</h2>' .
                ((PROJECT_VERSION_PATCH1 =='') ? '' : '<h3>Patch: ' . PROJECT_VERSION_PATCH1 . '::' . PROJECT_VERSION_PATCH1_SOURCE . '</h3>') .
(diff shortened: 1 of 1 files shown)


admin/server_info.php
```diff
@@ -113,7 +116,7 @@
         ?>
         <tr class="db-row">
             <td class="db-info db-name"><?php echo $variable['Variable_name']; ?></td>
-            <td class="db-info db-value"><?php echo empty($variable['Value']) ? '$nbsp;' : $variable['Value']; ?></td>
+            <td class="db-info db-value"><?php echo empty($variable['Value']) ? '$nbsp;' : htmlspecialchars($variable['Value']); ?></td>
```
scottcwilson, line 119: why did you remove the call to `htmlspecialchars`?

scottcwilson (changes requested):

brittainmark:
I didn't I added a call to htmlspecialchars

admin/server_info.php
brittainmark: I didn't I added a call to htmlspecialchars

scottcwilson: sorry I misread this.

PR #5142, opened 2022-08-19
Coupon_restrict HTML5 Validation

### LINKED ISSUE #5141: HTML5 Validation Errors coupon_restrict
**Describe the bug**
coupon_restrict has HTML 5 validation errors
validation errors: 
- Table column 3 established by element td has no cells beginning in it.
- Table columns in range 5…6 established by element td have no cells beginning in them.

**Version**
Zen Cart version: 1.5.8 
PHP version: 8.0.20

**To Reproduce**
1. Go to Admin>Discounds>Coupon Admin
2. If there are no coupons. create one
   1. Click insert
   2. Fill in form
   3. Click preview
   4. Click Confirm
5. Click on Restrict in side box
6. View page source
7. Copy page source 
8. Paste into https://validator.w3.org/nu/#textarea
9. Click validate

**Expected behavior**

brittainmark:
fix #5141

Always display the table heading rows to remove the HTML5 validation errors.
- Table column 3 established by element td has no cells beginning in it.
- Table columns in range 5…6 established by element td have no cells beginning in them.

I have done this for both tables to be consistent, even though it was only necessary for the second table.

Before
before

After
After

**Note:** the movement of the two lower buttons means that they are now more consistently placed on the screen when restrictions are present. If you want the screen to be totally consistent, would have to add bootstrap column widths.

After2

### DIFF
--- a/admin/coupon_restrict.php
+++ b/admin/coupon_restrict.php
@@ -307,15 +307,6 @@ $cr_list = $db->Execute($cr_query_raw);
     <div class="row">
         <h4><?php echo HEADING_TITLE_CATEGORY; ?></h4>
         <table class="table table-hover">
-<?php
-if ($cr_list->EOF) {
-?>
-            <tr class="dataTableHeadingRow">
-                <td colspan="4" class="dataTableHeadingContent text-center"><?php echo TEXT_NO_CATEGORY_RESTRICTIONS; ?></td>
-            </tr>
-<?php
-} else {
-?>
             <tr class="dataTableHeadingRow">
                 <td class="dataTableHeadingContent text-center"><?php echo TABLE_HEADING_CATEGORY_ID; ?></td>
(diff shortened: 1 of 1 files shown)


admin/coupon_restrict.php
```diff
@@ -500,7 +496,7 @@
 }
 unset($products);
 ?>
-            <tr class="smallText">
+            <tr class="smallText" colspan="2">
```
scottcwilson, line 499: Are you sure about this? colspan on a tr?

scottcwilson (changes requested):

admin/coupon_restrict.php
brittainmark: No, should not be there. Good spot! Will remove retest and reissue.

brittainmark:
fixed thanks

PR #5150, opened 2022-08-21
Update products_price_manager.php

### LINKED ISSUE #5149: $p.product_is_cal Not Valid Variable
**Describe the bug**
In admin>price_manager a parameter is passed to objectInfo ` 'p.product_is_cal' => 0,`

it should be  'product_is_call' => 0,

**Version**
Zen Cart version: 1.5.8
PHP version: 8.2

no error produced on earlier versions of php.

brittainmark:
fix #5149
Invalid variable


scottcwilson:
Nice catch!

Issue #5103, opened 2022-08-10
php 8.2: dynamic properties deprecated, multiple warnings


torvista:
[Deprecate dynamic properties](https://wiki.php.net/rfc/deprecate_dynamic_properties)

Thought I'd try the php 8.2beta: 94 warnings on entering the admin, and all for the same error but different files.
It would be good if someone could confirm if anything other than just declaring these variables is necessary....

This is useful, there seem to be only a couple more things to review:
https://stitcher.io/blog/new-in-php-82

lat9:
A copy (or a section) of the logged warnings would help.

torvista:
Well, here's a section of the 1200+ lines.
The format is as generated by my modified error_log: https://github.com/zencart/zencart/issues/5102

> [10-Aug-2022 18:31:43 Europe/Madrid]
> ADMIN
> Request URI: /tienda/ADMIN/
> IP address: 127.0.0.1
> #0 ADMIN/includes/classes/AdminRequestSanitizer.php(79): zen_debug_error_handler()
> #1 ADMIN/includes/classes/AdminRequestSanitizer.php(62): AdminRequestSanitizer->__construct()
> #2 ADMIN/includes/extra_datafiles/advshipper_sanitization.php(4): AdminRequestSanitizer::getInstance()
> #3 ADMIN/includes/init_includes/init_file_db_names.php(45): require('D://OneDrive//Bus...')
> #4 ADMIN/includes/application_bootstrap.php(156): require('D://OneDrive//Bus...')
> #5 ADMIN/index.php(9): require_once('D://OneDrive//Bus...')
> --> PHP Deprecated: Creation of dynamic property AdminRequestSanitizer::$arrayName is deprecated in ADMIN/includes/classes/AdminRequestSanitizer.php on line 79.
> 
> [10-Aug-2022 18:31:43 Europe/Madrid]
> ADMIN
> Request URI: /tienda/ADMIN/
> IP address: 127.0.0.1
> #0 includes/classes/db/mysql/query_factory.php(61): zen_debug_error_handler()
> #1 ADMIN/includes/init_includes/init_database.php(18): queryFactory->connect()
> #2 ADMIN/includes/application_bootstrap.php(157): require('D://OneDrive//Bus...')
> #3 ADMIN/index.php(9): require_once('D://OneDrive//Bus...')
[...]

PR #5153, opened 2022-08-22
PHP8.2 Define Parameters object_info

brittainmark:
Define Parameters

Define all the parameters used by the class

Note this fixes 8.2 PHP Deprecated: Creation of dynamic property

**This does not fix the issue with customisation.** They will still be reported as a warning.

Could add #[AllowDynamicProperties] to disable the warning message.


brittainmark:
Part of #5103

PR #5157, opened 2022-08-22
PHP 8.2 Define Parameters table_block

brittainmark:
Define parameters for table_block and its sub classes

## boxTableBlock
$table_row_parameters, $table_data_parameters are only used in the class and subclass so I have set to protected.
## box
$heading, $contents are only used in this class so I have set them to private.
## messageStack
no changes required.

Do you agree with my logic? While I come from a process background, I have read up on OO and will happily take input if I have not got it right.

Part of #5103


lat9 (approved):
Looks good, @brittainmark

PR #5165, opened 2022-08-24
PHP8.2 Define Parameters queryFactory

brittainmark:
Changed to Private
- $link
- $count_queries
- $total_query_time

Changed to public
- $dieOnErrors
- $error_number
- $error_text

Set to private
- $pConnect   

**N.B. $pConnect  could be completely removed as it is not used anywhere**

Part of #5103

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -19,12 +19,14 @@ if (!defined('IS_ADMIN_FLAG')) {
  */
 class queryFactory extends base
 {
-    var $link; // mysqli object
-    var $count_queries = 0;
-    var $total_query_time;
-    var $dieOnErrors = false;
-    var $error_number = 0;
-    var $error_text = '';
+    private $link; // mysqli object
+    private $count_queries = 0;
+    private $total_query_time;
+    public $dieOnErrors = false;
+    public $error_number = 0;
+    public $error_text = '';
+    
+    private $pConnect;
     /**
      * @var bool
      */


torvista:
plugin sitemap xml uses
count_queries
total_query_time

so bugs out now they are private.
I'd put a link to the current sitemap Github but there are more than one, so some rationalisation required there.  
You know who you are!

brittainmark:
@torvista  Do you want them changed back to public?

torvista:
I guess there is no alternative.

brittainmark:
The plugin could be changed to use ->queryCount() and ->queryTime() methods.
I assume that they do not set these values.

dbltoe:
https://github.com/dbltoe/Sitemap-XML-4 exists only to show what I did to make davewest's version work with 1.5.7.
Let me know if you feel it needs any specific tweak for 1.5.8

brittainmark:
@dbltoe  The change needed for 158 is to replace
`$db->count_queries `with `$db->queryCount() `
and 
`$db->total_query_time` with `$db->queryTime()`

Assuming you only use $db to access the database.

brittainmark:
Probably need to add to 1.5.8 release docs. If leaving as is.

dbltoe:
made a new folder for 1.5.8 and added install/upgrade instructions.

torvista:
That change works for me but also gives warnings

Request URI: admin/index.php?main_page=sitemapxml&rebuild=yes
--> PHP Warning: Undefined property: queryFactory::$queryCount in includes/classes/sitemapxml.php on line 78.
--> PHP Warning: Undefined property: queryFactory::$queryTime in includes/classes/sitemapxml.php on line 79.

plus many more the same on other lines...but this is a sitemap issue/off topic here.

But on-topic:
PHP Deprecated: Implicit conversion from float 23.817535655425836 to int loses precision in includes/classes/db/mysql/query_factory.php on line 444.

https://php.watch/versions/8.1/deprecate-implicit-conversion-incompatible-float-string

brittainmark:
if you change line 444 in query factory to
` return (int)$this->total_query_time;`
should get rid of the implicit conversion issue.

I'll update and create a pr.

Are you using the latest version of queryFactory as $queryCount and $queryTime are not use in query factory except as public functions? Only thing I can think of is are the brackets omitted? Or are you trying to set the values?

brittainmark:
I have an outstanding change of query factory. Is it OK to add to that, or do I need to wait for that one to be merged.
#5186 and create a new pr.

torvista:
`Only thing I can think of is are the brackets omitted?`
Doh, correct. Sorry, busy with the real world. With that and the int cast, all good.

brittainmark:
Probably breaking rules but have included the int cast in #5186

brittainmark:
@scottcwilson Will need to add to the documentation what's new something like
The queryFactory had been changed so that the parameter count_queries, and total_query_time are no longer accessible. Any plugins that use these parameters will need to be changed to us the queryCount() and queryTime() methods respectively.
e.g. assuming you are using `$db = new queryFactory` 
change
`$db->count_queries` to `$db->queryCount()`
and
`$db->total_query_time` to `$db->queryTime()`

scottcwilson:
Using count_queries: 
- sitemapXML 
- printable_price_list
- ultimate_seo_urls

Using total_query_time:
- sitemapXML

brittainmark:
I believe sitemapXML has been changed. See comments above. 
@lat9 are you happy to change printable_price_list and ultimate_seo_urls?

dbltoe:
I wasn't able to tell if some of this discussion is still pointed at sitemapXML but, I will fix any problems I'm made aware of.

mc12345678:
> @scottcwilson Will need to add to the documentation what's new something like The queryFactory had been changed so that the parameter count_queries, and total_query_time are no longer accessible. Any plugins that use these parameters will need to be changed to us the queryCount() and queryTime() methods respectively. e.g. assuming you are using `$db = new queryFactory` change `$db->count_queries` to `$db->queryCount()` and `$db->total_query_time` to `$db->queryTime()`

There is/was a larger change/impact than just those two members becoming private. What about the entire reference to the database itself? The member `link` has gone from `public` (`var`) to `private` which prevents the use of many `mysqli_` functions. For what its worth, this is being addressed in a separate issue (that will reference this one).

scottcwilson:
thanks for the reminder - added this doc task to my todo list. 

@mc12345678 We *may* need to make `link` public, and I am open to it.  Thanks for raising this issue.

brittainmark:
Not providing the link was to try to ensure that all the sql access was through $db. While I understand that this may be difficult for existing mods that have not used the queryFactory but used their own sql directly. It would be better practice to have all of the database access controlled through the queryFactory. While making link public solves the problem for them. it means that the access to the database it not controlled by/through central zen cart functions.

Perhaps we could make it public for now with the intention of changing it to private in the next release to give those who have written addons that make use of the link time to adjust their code and/or ask/provide mods to queryfactory to meet their needs.

scottcwilson:
In spirit I agree, but the updating of old plugins is already going to be so incredibly painful that I'm looking to make it easier where we possibly can.   Only experts will be doing this, truthfully - the average person wouldn't even think to do direct database access.

brittainmark:
You are making my argument stronger. If they are experts, they should recognize the value of keeping the SQL centralized.

brittainmark:
Additionally, if most add-ons are having to be rewritten for 1.5.8, then why not do the changes now.

proseLA:
> thanks for the reminder - added this doc task to my todo list.
> 
> @mc12345678 We _may_ need to make `link` public, and I am open to it. Thanks for raising this issue.

i am weighing in a bit late to the conversation.  and while updating plugins may be painful, i'm not seeing any updates to easypopulate for over 1.5 years.

in briefly looking at that code, there is no class;  so perhaps forcing the author to create their own assessor class is an answer, if the current class does not work for them.

at some point, code modernization is necessary rather than these PRs to preserve the past.

scottcwilson:
You might be right but let's give them a break for 1.5.8.  Like I said there were a ton of changes made to get us ready for 8.2, and some of them may or may not have been the right call.  We can revisit this particular one in the next release.

PR #5166, opened 2022-08-24
reformat customers

proseLA:


brittainmark (commented):
Was this an auto reformat?
Should type checking also have been done on if statements.
line 47 for example 
`if (CUSTOMERS_APPROVAL_AUTHORIZATION == 1 || CUSTOMERS_APPROVAL_AUTHORIZATION == 2) {`
to
`if (CUSTOMERS_APPROVAL_AUTHORIZATION === 1 || CUSTOMERS_APPROVAL_AUTHORIZATION === 2) {`

drbyte:
> Should type checking also have been done on if statements.

Not in this case. All of those constants that are established by virtue of admin configuration settings are treated as strings.

torvista:
As far as I am aware there should be no code changes whatsoever in a **reformat**.
I assume the reformat is prior to some actual code reworking.

admin/customers.php
```diff
@@ -856,253 +863,261 @@ function check_form() {
...
-                    <tr id="defaultSelected" class="dataTableRowSelected" onclick="document.location.href = '<?php echo zen_href_link(FILENAME_CUSTOMERS, zen_get_all_get_params(['cID', 'action']) . 'cID=' . $cInfo->customers_id . '&action=edit'); ?>'" role="button">
-                    <?php } else { ?>
-                    <tr class="dataTableRow" onclick="document.location.href = '<?php echo zen_href_link(FILENAME_CUSTOMERS, zen_get_all_get_params(['cID', 'action']) . 'cID=' . $customer['customers_id']); ?>'" role="button">
-                      <?php
-                    }
+                                    if (isset($cInfo) && is_object($cInfo) && ($customer['customers_id'] == $cInfo->customers_id)) {
+                                        ?>
+                                        <tr id="defaultSelected" class="dataTableRowSelected" onclick="document.location.href = '<?php echo zen_href_link(FILENAME_CUSTOMERS, zen_get_all_get_params(['cID', 'action']) . 'cID=' . $cInfo->customers_id . '&action=edit'); ?>'" role="button">
+                                        <?php
+                                    } else { ?>
```
scottcwilson, line 898: I'd prefer this on one line (i.e. as before).

admin/customers.php
```diff
@@ -856,253 +863,261 @@ function check_form() {
...
-                      if (count($additional_columns) != $additional_heading_count) {
-                        trigger_error("Mismatched additional column heading ($additional_heading_count) and column element (" . count($additional_columns) . ") counts detected for the Customers listing.", E_USER_WARNING);
-                      }
-                      foreach ($additional_columns as $column_data) {
-                        $additional_class = (isset($column_data['class'])) ? (' ' . $column_data['class']) : '';
-                        $additional_parms = (isset($column_data['parms'])) ? (' ' . $column_data['parms']) : '';
-                        $element_content = $column_data['content'];
-                        ?>
-                        <td class="dataTableContent<?php echo $additional_class; ?>"<?php echo $additional_parms; ?>><?php echo $element_content; ?></td>
+                                    $zc_address_book_count = count($customer['addresses']); ?>
```
scottcwilson, line 903: I'd prefer the end tag on a new line (as before).

admin/customers.php
```diff
@@ -856,253 +863,261 @@ function check_form() {
...
+                                            trigger_error("Mismatched additional column heading ($additional_heading_count) and column element (" . count($additional_columns) . ") counts detected for the Customers listing.", E_USER_WARNING);
+                                        }
+                                        foreach ($additional_columns as $column_data) {
+                                            $additional_class = (isset($column_data['class'])) ? (' ' . $column_data['class']) : '';
+                                            $additional_parms = (isset($column_data['parms'])) ? (' ' . $column_data['parms']) : '';
+                                            $element_content = $column_data['content']; ?>
+                                            <td class="dataTableContent<?php echo $additional_class; ?>"<?php echo $additional_parms; ?>><?php echo $element_content; ?></td>
+                                            <?php
+                                        }
+                                    } ?>
```
scottcwilson, line 945: same comment as above.

scottcwilson (changes requested):

proseLA:
> Was this an auto reformat? Should type checking also have been done on if statements. line 47 for example `if (CUSTOMERS_APPROVAL_AUTHORIZATION == 1 || CUSTOMERS_APPROVAL_AUTHORIZATION == 2) {` to `if (CUSTOMERS_APPROVAL_AUTHORIZATION === 1 || CUSTOMERS_APPROVAL_AUTHORIZATION === 2) {`
> 
> otherwise look good.

as others have stated, a reformat should not include any changes to the logic.  which your suggestion does.  and the logic is incorrect, as chris has stated, these constants are strings and so the change should be:

```
if (CUSTOMERS_APPROVAL_AUTHORIZATION === '1' || CUSTOMERS_APPROVAL_AUTHORIZATION === '2') {
```

but again that should should be part of a separate PR.

> > Should type checking also have been done on if statements.
> 
> Not in this case. All of those constants that are established by virtue of admin configuration settings are treated as strings.

ALWAYS a pleasure to see a comment by chris.  i have learned much from him over the years; and not just about coding.

there are some config values that are not strings, see:

https://github.com/zencart/zencart/blob/5151f58c0154ae0fafb29569bde3fafaae56a890/laravel/app/Models/Configuration.php#L23-L28

but in general, he is correct.  hence my comment above...

> As far as I am aware there should be no code changes whatsoever in a **reformat**. I assume the reformat is prior to some actual code reworking.

👍 

@scottcwilson i am sorry, i have no plans to manually code to your whim.  

i have attempted to follow the standards that are set here:

https://docs.zen-cart.com/dev/contributing/coding_standards/

and nowhere are your preferences marked on this document.  if i am wrong, please illuminate me.

in addition, PSR-2 is now deprecated; and PSR-12 is recommended.  see:

https://www.php-fig.org/psr/psr-2/

i know my preference (and torvista) is to use  `<?= ` over `<?php echo `.  which was acceptable as far back as psr-1.  but yet, is not preferred here in this repo.  and considering the massive amounts of our code that mix html with php....  well you got me on that one.

perhaps other contributors might want to weigh in on their preference with regards to this issue.

i know @torvista and i have discussed coding standards recently.  i spent a bunch of time with his PR:

https://github.com/zencart/zencart/pull/5098

but eventually i decided the effort to come up with a standard that is acceptable to everyone was too hard.  i played around with a bunch of tools but none formatted to my liking without digging further into its configuration.

while i acknowledge that coding standards are a difficult issue, it is one that has numerous tools out there to address the problem.

i have used this tool in the past:

https://github.com/FriendsOfPHP/PHP-CS-Fixer

which is IDE independent.  and configurable via this file:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/config.rst

now, perhaps we would like to set our own rules so that someone like me does not have to manually go in, and make changes to conform to someone's coding standard which is not listed in the docs.

and mind you, i am not tied to this tool.  there are others.  i just want something automated to reformat to our standard.

there is much potential refactoring to this script that could be done.  but i have no plans to do it, until the code is reformatted to something close to the new standard, without me having to go in and manually make changes.

in v158 (assuming customer 9989 does not exist), the following url now results in a fatal error:

https://zcdev.com/admin/index.php?cmd=customers&page=1&cID=9989&action=edit

brittainmark:
I to like the idea of a standard formatting tool. My comment on type checking was because array() was replaced with [ ] and TRUE was replaced with true. I see now that these were in fact just formatting changes.

Totally in favour of standard tool. Especially if the formatting file can be held centrally (on Zen cart) so we all know what we are working to.

scottcwilson:
> > i just want something automated to reformat to our standard
> 
> DItto.

Use phpStorm. 
https://stackoverflow.com/questions/57475642/phpstorm-and-psr-12-how-can-i-add-it-as-default-coding-style

scottcwilson:
Replaced by #5173.

torvista:
> Use phpStorm.

I did. That did not go down well!

https://github.com/zencart/zencart/pull/5098

brittainmark:
What about those of us who do not have phpStorm?

scottcwilson:
@torvista I think it mostly did.  There was just left over cruft like 

```
 } ?>
<?php
```

that need to be removed manually.

scottcwilson:
@brittainmark You can purchase it.    Or better yet: don't reformat files!  It's a huge pain in the neck for maintainers and doesn't fix any actual bugs.  Lose-lose!

proseLA:
> @brittainmark You can purchase it. Or better yet: don't reformat files! It's a huge pain in the neck for maintainers and doesn't fix any actual bugs. Lose-lose!

i disagree.  properly formatted code is much easier to read.

as far as huge pain in the neck; that is such a judgement call.  and i can think of things much worse...

with recent discussions about recruiting new developers, i think some of the opinions expressed here does the opposite.

php-cs-fixer has new rule sets for migration to php 8.1; 8,.2, etc.   other more popular frameworks have their own rule sets for code formatting.   symphony has one.  i am sure laravel does.  and they make it easy for developers to ensure their code fits that spec.

telling an active contributor on this forum to purchase an IDE, when they can be using some other IDE is not what i consider supportive.  my environment is an ubuntu desktop, with vagrant as a wrapper to virtualBox, using phpstorm on my client machine to edit code being served up by vagrant.  would i force that specific environment on anyone?  no.  i have never used eclipse for example; and i'm sure there are some great developers that do.

every time someone tries to reformat code, we get into this person likes this, or that person likes that...  i find it tiresome.

@brittainmark you can try using:

https://github.com/FriendsOfPHP/PHP-CS-Fixer

i have a ruleset over here:
https://github.com/proseLA/zencart/blob/formatRules/.php-cs-fixer.php

and you can see how that works for you.

and for people using phpstrom, there is a plugin or something to get that IDE to format the code according to a php-cs-fixer rule set.

i actually run it from the command line:
```
$: php-cs-fixer --allow-risky=yes fix admin/customers.php
```
i downloaded the .phar file, put it my path, and removed the extension.

maybe we can make forward progress on coming up with our own config and ruleset.

brittainmark:
Thanks for the support. I use NetBeans, and it has an add-on that basically allows the use of php-cs-fixer.
What I was thinking was if the .phar file was held in the Zen cart GitHub then everybody would reformat to the same set of rules.
I did start looking at it, but decided it was too much to understand at this stage. Trying to work out what rules to use or not use.
I will often reformat code to understand it then go back to the original to make changes as I don't want to upset the apple cart by reformatting.

It probably would be a good idea for someone like me who has the time to look at the none essential changes to go through and reformat all the code. But without a tool to do it, I would most likely get it wrong.

proseLA:
> Thanks for the support. I use NetBeans, and it has an add-on that basically allows the use of php-cs-fixer. What I was thinking was if the .phar file was held in the Zen cart GitHub then everybody would reformat to the same set of rules. I did start looking at it, but decided it was too much to understand at this stage. Trying to work out what rules to use or not use. I will often reformat code to understand it then go back to the original to make changes as I don't want to upset the apple cart by reformatting.
> 
> It probably would be a good idea for someone like me who has the time to look at the none essential changes to go through and reformat all the code. But without a tool to do it, I would most likely get it wrong.

the ruleset on my fork that i linked to, when put in the proper place overrides the base rule set in the .phar.  which i think is a place to start prior to packing up our own phar.

torvista:
I've oft repeated my opinion that for the 157-158 upgrade, no-one in their right mind should attempt a file merge: they should start with 158 and add their mods.
So, why not format the whole.bloody.fileset in one go. Now. 
Seems the perfect moment.

PR #5180, opened 2022-08-27
HTML5 valign

### LINKED ISSUE #5179: HTML5 Valign Obsolete
**Describe the bug**
HTML5 validation error 
- The valign attribute on the td element is obsolete. Use CSS instead.
**Version**
Zen Cart version: 1.5.8 
PHP version: 8.0.20

**To Reproduce**
1. Go to Admin>Tools>Who's online
2. View page source
3. Copy page source 
4. Paste into https://validator.w3.org/nu/#textarea
5. Click validate

**Expected behavior**

**Additional context**
There are other validation errors which will be dealt with in other PR's

brittainmark:
Remove valign and replace with css where necessary.

Fix #5179

### DIFF
--- a/admin/coupon_admin_export.php
+++ b/admin/coupon_admin_export.php
@@ -313,7 +313,7 @@ require (DIR_WS_INCLUDES . 'header.php');
 <table border="0" width="100%" cellspacing="2" cellpadding="2">
   <tr>
     <!-- body_text //-->
-    <td width="100%" valign="top">
+    <td class="align-top" width="100%">
     <table border="0" width="100%" cellspacing="0" cellpadding="0">
       <tr>
         <td width="100%">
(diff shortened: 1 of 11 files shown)


dbltoe:
I really don't want to seem like I'm being picky here.  The only reason I have probably not addressed all these myself is that I am limited on the time I can sit at the computer and I have not addressed any Accessibilty, HTML, or CSS in the admin unless it specifically came up.

In the various files, some have a TD left with no class while others assume that the existing class will always need to be aligned to the top of the TD.  Would it not be best to add class="align-top" to the TD's with no existing class and add align-top to the list of classes where they exist for a TD?

Then, they could all be controlled with the added .align-top element in the stylesheet.  Those with the align-top AND another class or div could still be controlled by their specific div/class.

brittainmark:
If the table already had td vertical-align=top set, I removed the align and let inheritance do the job. 
If you think I would be easier to understand, I can change the code to include the align-top class.

brittainmark:
Just reread your comment. **I think you are correct**. I will make the change you suggested and resubmit.

dbltoe:
I've busted my allotted time but, will try to get back this evening.

THANX for your work.

brittainmark:
redone, keeping the attribute in the class.

PR #5181, opened 2022-08-27
Remove <span class="tt">

brittainmark:
Remove the \<span class="tt"> and add the class to the preceding element.

### DIFF
--- a/admin/coupon_admin.php
+++ b/admin/coupon_admin.php
@@ -724,7 +724,7 @@ switch ($_GET['action']) {
               <?php } ?>
               <tr>
                 <td class="text-right"><b><?php echo TEXT_MESSAGE; ?></b></td>
-                <td><span class="tt"><?php echo nl2br(htmlspecialchars(stripslashes($_POST['message']), ENT_COMPAT, CHARSET, TRUE)); ?></span></td>
+                <td class="tt"><?php echo nl2br(htmlspecialchars(stripslashes($_POST['message']), ENT_COMPAT, CHARSET, TRUE)); ?></td>
               </tr>
               <tr>
                 <td>
--- a/admin/gv_mail.php
+++ b/admin/gv_mail.php
@@ -228,7 +228,7 @@ if (!empty($_GET['mail_sent_to']) && $_GET['mail_sent_to']) {
             <?php } ?>
             <tr>
               <td class="text-right"><b><?php echo TEXT_MESSAGE; ?></b></td>
-              <td><span class="tt"><?php echo nl2br(htmlspecialchars(stripslashes($_POST['message']), ENT_COMPAT, CHARSET, true)); ?></span></td>
+              <td class="tt"><?php echo nl2br(htmlspecialchars(stripslashes($_POST['message']), ENT_COMPAT, CHARSET, true)); ?></td>
             </tr>
           </table>
           <div class="form-group">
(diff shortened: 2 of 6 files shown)


dbltoe:
BTW, it may have been part of the problem before with tt used as an HTML element and set as a class in the stylesheet vs just tt.  Not sure.

While admin/coupon.php and admin/gv_mail.php were using tt for everything in the TD, in cases where not all of the TD was using the tt (like admin/includes/modules/newsletters/newsletter.php and admin/mail.php) I would use `<code>` to make sure we're only wrapping the items intended to be presented in the different font.

Doesn't adding tt as a class to line 71 of the newsletter.php turn everything into tt versus the intended use of wrapping just the div?

brittainmark:
No, as the \<div> is closed on the same line.
Effectlvly the \<div> replaces the \<span> which replaced the \<tt>

dbltoe:
> Doesn't adding tt as a class to line 71 of the newsletter.php turn everything into tt versus the intended use of wrapping just the div?

The original `$confirm_string .= '<div class="col-sm-12"><tt>' . nl2br($this->content) . '</tt></div>' . PHP_EOL;` would become `$confirm_string .= '<div class="col-sm-12"><code>' . nl2br($this->content) . '</code></div>' . PHP_EOL;` and avoid any confusion with the div's class if used elsewhere.  Or am I thinking incorrectly that this is specific "code" inside a div that is used elsewhere or it would be an ID vs class?

brittainmark:
Yes, I see what you are saying. My concern is that the text ouput is not **code** it is the content of the email to be sent displayed as it would in a text only email. 

The class col-sm-12 is basically width 100% and is used in lots of places throughout the code. 
I could set and id to the div to identify it for future modification. If I do it here, I would suggest that it needs doing to all the sections e.g. the HTML text output as well.

dbltoe:
You may not have taken a look at the link for code and seen that it can be formatted much like the tt of old.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code and click on the CSS under the TryIt section.

Take away the background and border-radius: you have plain text.  Just changing .tt to code in the CSS would make it just like before. 😃

brittainmark:
I did have a look, and it is being used to highlight sections of code.  This is not code. This is just a plain text email. That is why I don't like the code tag. The \<tt> was use to display the email as teletext or fixed format text. So I would think that using the .tt class was the correct way to go.

dbltoe:
Back at it for a few.  To maybe better express where I'm coming from, I think that my definition of "code" may be a little broader.  If I'm off-base, I plead at least semi-ignorance of the goal.

What is code? Is it HTML, CSS, PHP, JavaScript?  All of them?  Isn't the combination of letters we put together in any form really code?  In my mind, Bon Jour is _code_ for G'day in Australia and Bon Giorno in Italy.  Is not plain text a "plain" code?

I realize the goal is to strip the HTML formatting but, what about special characters that we _**want**_ to show up in the plain text message?

To me, the code command says to the browser, "Take whatever I give you and format it without change."  If that's a false assumption on my part then code may, in fact, _not_ be the answer.

brittainmark:
Yes I see where you are coming from. In this case, you have 2 text box so complete, one for rich format text and one for plain text.
The text you complete is for the content of an email. There is no stripping of HTML tags. The idea is to display the two messages as they would be seen. So I still feel that using the class tt is the way to go as we are formatting the output and not displaying code.

scottcwilson:
Where are we on this one?

dbltoe:
While my concern was really over the definition of "code" and not whether or not the solution would work, I see no reason not to merge.

PR #5186, opened 2022-08-30
PHP8.2 Define Parameters queryFactory

brittainmark:
## queryFactory
Necessary to cast queryTime to int to avoid warning.

## queryFactoryResult
No changes

## queryFactoryMeta
public $type; used by zen_field_type in includes\functions\functions_general_shared.php
public $max_length; used by zen_field_length in includes\functions\functions_general_shared.php


brittainmark:
Part of #5103

PR #5187, opened 2022-08-30
PHP8.2 Define Parameters template_func

brittainmark:
$info set to private array. 

I cannot find any use of parameter.


brittainmark:
Part of #5103

PR #5188, opened 2022-08-30
PHP8.2 Define Parameters zcDate

brittainmark:
$dateObject set to protected. 
All other parameters are protected. This will allow for other classes to extend zcDate if needed.


brittainmark:
Part of #5103

PR #5191, opened 2022-08-31
PHP8.2 Define Parameters AdminNotifiers

brittainmark:
$projectNotificationServer set to private and not used outside of class

Also added check to see if set and return empty array if not set.
**Alternative**, if desired, remove both checks and fail if PROJECT_NOTIFICATIONSERVER_URL is not defined. 
Currently defined in include/version.php as "https://ping.zen-cart.com/api/notifications" and
in zc_install/includes/version.php as "https://versionserver.zen-cart.com/api/notifications".


brittainmark:
Part of #5103

PR #5192, opened 2022-08-31
 PHP8.2 Define Parameters statsSalesReportGraph

brittainmark:
All set to public as used outside class $filter_link, $filter_sql = '', $status_available = [], $status_available_size = 0.
I did not set $filter_link as only defined in the class.


brittainmark:
Part of #5103

PR #5195, opened 2022-09-03
PHP8.2 Define Parameters catgory_tree

brittainmark:
All set to private only used in class
- $box_categories_array = [];
- $categories_string;
- $tree = [];


brittainmark:
Part of #5103

PR #5199, opened 2022-09-05
PHP8.2 Define Parameters shipping

brittainmark:
## shipping
Set to public as notifiers may modify values
- $modules
- $enabled
- $abort_legacy_calculations

## All shipping modules
Set to protected not required outside class
- $_check;

Set to public used outside class
- $code;
- $description;
- $enabled;
- $icon;
- $quotes;
- $sort_order;
- $tax_basis;
- $tax_class;
- $title;

### freeoptions
Set to public
- $debug = [];

### freeshipper
Did not have $tax_basis have still included definition

### storepickup
Set protected as not required outside module
- $locations = [];
- $methodsList = [];

### zones
Set to protected only required by class
- $num_zones;

### DIFF
--- a/includes/classes/shipping.php
+++ b/includes/classes/shipping.php
@@ -17,7 +17,21 @@ if (!defined('IS_ADMIN_FLAG')) {
  */
 class shipping extends base
 {
+    /**
+     * $enabled allows notifier to turn off shipping method
+     * @var boolean
+     */
+    public $enabled;
+    /**
+     * $modules is an array of installed shipping module names can be altered by notifier
+     * @var array
+     */
     public $modules;
+    /**
+     * $abort_legacy_calculations allows a notifier to enable the calculate_boxes_weight_and_tare method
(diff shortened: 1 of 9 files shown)


brittainmark:
Part of #5103

brittainmark:
Corrected typo payment module to shipping module

scottcwilson:
Please handle tax basis as with the other modules.

brittainmark:
> Please handle tax basis as with the other modules.

I'll do it in a separate request as it will need changes to install and update  sql files which will complicate this pr.

scottcwilson:
Sorry typed too fast.  Please hold off on additional tax basis change since no one has complained about needing this.  Stay focused on the PHP 8.2 changes you are doing.  Sorry for the confusion.

brittainmark:
Will need to change #5200 as it does not allow store basis. Will do when this one merged. I already have it ready.

PR #5209, opened 2022-09-09
PHP8.2 Define Parameters Payment Method COD

brittainmark:
Set to protected not required outside module
- $_check

Set to public used outside module
- $code
- $description
- $enabled
- $order_status
- $title
- $sort_order


brittainmark:
Part of #5103

PR #5213, opened 2022-09-12
Don't use protected output for admin comments

scottcwilson:


zcwilt:
In terms of PSR-2 there is nothing wrong with the commit.

proseLA:
my reading of PSR-2 suggests the code should be:
```
if (empty($order_history['comments'])) {
    echo TEXT_NONE;
} elseif ($count_comments == 1) {
    echo nl2br(zen_output_string_protected($order_history['comments']));
} else {
    echo $order_history['comments'];
}
```

the nested if makes no sense to me.

proseLA:
outside of PSR-2, it seems with the more strict typing of PHP, the new code should also be:

```
} elseif ($count_comments === 1) {
```

zcwilt:
So my point here is that PSR2 suggests formatting of code 
e.g how to format if you are using an `if` construct and what to do if you are using an `if else` construct
PSR2 does not tell you how to write code
You may have legitimate  concerns over the structure or algorithm the code uses and whether it produces code smells (e.g arrow code)
https://blog.codinghorror.com/flattening-arrow-code/

but its not a PSR2 concern 

Apologies for being anal about this

zcwilt:
There is a bigger picture here 

Do I want to see all code written, to follow PSR2/12 yes
Do I want code written to reduce code smells and cyclical complexity again yes
(and we could add commit hooks to show whether a commit increases/decreases code cleanliness)
However at least for where we are at and where we want to get to, sometimes we compromise

zcwilt:
As an example 

PSR says I should write

``` 
if ($something === false {
    return 
    }
    ```
    
    Whereas
    
   ```
   if ($something === false) return 
   
  ```
  
  Seems eminently more readable
  
  If some one committed the short version above amongst a longer commit, I'm not sure if I would deny the commit just based on the fact that it didn't pass psr

proseLA:
> Apologies for being anal about this

apologies are not necessary.  if anything i should be the one apologizing...

the way i see it, this thread has now morphed into 2 separate discussions:

- the first is whether the code as written in this PR is written according to the standards set out in PSR2/12.
- the second is whether the personal preferences of the maintainers comes into play when merging code into the repo.

with regards to the first point, i think the code does not conform; you (and perhaps others) think that it does.  which is fine.  i still think that code smells, and could be cleaner.

with regards to the second point, what is the point of having standards if we are just going to allow people to code to their personal preferences?  in your example given, i disagree.  i think the single line without the curly braces is not eminently more readable.  and i would say that the maintainers of the PSR standards agree with me.

i think the short echo tag when used in scripts that combine php with html is eminently more readable.  the short echo tag is also acceptable according to the standards in PSR2/12.  and despite a number of contributors to this repo agreeing with me; it is not allowed in this repo.  and i respect that decision.

if we do not want to code to standards, that's fine.  but lets not say that we do.  

in the interest of compromise as well as time suck, i will (hopefully) choose to not respond to this thread any more.  as always, zcwilt, i have tremendous respect for you and whatever you choose to do is fine with me.  thanks for taking the time to respond.

PR #5224, opened 2022-09-18
Fix PHP_OS check to look at first 3 chars only

scottcwilson:
Checking for "win" anywhere in the string returns true for "Darwin" (MacOS). 
This is the only check in the core that doesn't look only at the first three characters.

### DIFF
--- a/includes/functions/functions_dates.php
+++ b/includes/functions/functions_dates.php
@@ -52,7 +52,9 @@ function zen_date_long($raw_date)
 
     global $zcDate;
     $retVal = $zcDate->output(DATE_FORMAT_LONG, mktime($hour, $minute, $second, $month, $day, $year));
-    if (stristr(PHP_OS, 'win')) return utf8_encode($retVal);
+    if (stristr(PHP_OS, 'win') === 0) {
+       return utf8_encode($retVal);
+    }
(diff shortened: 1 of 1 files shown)


torvista:
ZC158: In my testing locally on Windows 11, using spanish, and setting a product expected date for a Wednesday: miércoles.
On the admin product preview page, this function is used.

If this specific conversion for windows is **disabled**:
php 7.4, 8.0.23: the output is mis-encoded as expected: "mi�rcoles"
php 8.1.10, 8.2.0RC2 : the output is correct: "miércoles"

if this specific conversion for windows is **enabled**:
php 7.4, 8.0.23: the output is correct as expected: "miércoles"
php 8.1.10, 8.2.0RC2 : the output is mis-encoded: " miÃ©rcoles"

I would say bin it completely and put this example in the docs section as to more reason why Windows is to be avoided.

scottcwilson:
Thanks for doing these tests.  Good to know this won't be needed in PHP8 since the `utf8_encode` will be deprecated in PHP 8.2.

scottcwilson:
I'm going to merge this (it's impossible to test PHP 8.2 on MacOS without this) and then open a new issue for removal.

PR #5235, opened 2022-09-19
Fix order creation in PHP 8.2

### LINKED ISSUE #5223: PHP 8.2 Deprecations when placing an order
--> PHP Deprecated: Creation of dynamic property order::$subtotal is deprecated in /Users/scott/sites/gh_demo_158/includes/classes/order.php on line 893.
--> PHP Deprecated: Creation of dynamic property order::$total_tax is deprecated in /Users/scott/sites/gh_demo_158/includes/classes/order.php on line 894.
--> PHP Deprecated: Creation of dynamic property order::$total_weight is deprecated in /Users/scott/sites/gh_demo_158/includes/classes/order.php on line 1132.
--> PHP Deprecated: Creation of dynamic property order::$total_cost is deprecated in /Users/scott/sites/gh_demo_158/includes/classes/order.php on line 1134

scottcwilson:
Fixes #5223

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -18,11 +18,12 @@ if (!defined('IS_ADMIN_FLAG')) {
 class order extends base
 {
     protected $orderId = null;
-    var $info, $totals, $products, $customer, $delivery, $billing, $content_type, $email_low_stock, $products_ordered_attributes,
+    public $info, $totals, $products, $customer, $delivery, $billing, $content_type, $email_low_stock, $products_ordered_attributes,
         $products_ordered, $products_ordered_email, $products_ordered_html, $attachArray, $email_order_message, $extra_header_text,
         $doStockDecrement, $send_low_stock_emails, $queryReturnFlag, $bestSellersUpdate, $use_external_tax_handler_only;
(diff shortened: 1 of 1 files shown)


brittainmark (commented):
I had a look at this too. A few comments
- Apart from being set to Zero $subtotal is not used. Should it be removed.
- $products_ordered_attributes_html is never used in the class. Should it be removed.
- $products_ordered_email is never used in the class. Should it be removed.
- I left $total_tax, $total_weight, $total_cost as public in case the notifier "NOTIFY_ORDER_PROCESSING_ONE_TIME_CHARGES_BEGIN" needed to change them
- Finally I also set $statuses to public as if the notifier "NOTIFY_ORDER_BEFORE_QUERY" sets $queryReturnFlag to true then there is no way to set the statuses

torvista:
$statuses = [] needs to be public:
PHP Fatal error: Uncaught Error: Cannot access protected property order::$statuses in ....includes\modules\pages\account_history_info\header_php.php:24

brittainmark:
Submitted  #5236 to fix above

scottcwilson:
Merged.

PR #5239, opened 2022-09-20
Add products_id to pInfo object

scottcwilson:
After a product edit, then preview, then back, the $pInfo->products_id field is not set.  This means that notifiers get called without this value, and fail. 

This should be done for other product types too but I wanted to get this one in first.

### DIFF
--- a/admin/includes/modules/product/collect_info.php
+++ b/admin/includes/modules/product/collect_info.php
@@ -56,6 +56,7 @@ if (isset($_GET['pID']) && empty($_POST)) {
   $pInfo->updateObjectInfo($product->fields);
 } elseif (!empty($_POST)) {
   $pInfo->updateObjectInfo($_POST);
+  $pInfo->products_id = $pInfo->id;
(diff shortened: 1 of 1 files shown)


torvista:
Off the cuff, is this an issue for a new product that does not have an id yet?

brittainmark:
sorry i read the wrong commit. 

Should it be set to $_GET["pId"] as $pInfo->id does not exist! or have I missed something else.

scottcwilson:
You're right- this happened to work on my test cart because of another change I made. 
Follow up PR coming shortly.

scottcwilson:
Fixed in #5243.

brittainmark:
#5243 did not get merged. Is there another fix coming?

scottcwilson:
It got fixed - I had to update the branch directly because there was a merge conflict.  Update your repo and you'll see the fix.

scottcwilson:
Ugh sorry my branch was hosed.  See #5258

PR #5251, opened 2022-09-24
PHP8.2 Define Parameters Paginator

brittainmark:
I don't think this is used any more. LaravalPaginator used instead.

All set to protected


brittainmark:
Part of #5103

PR #5255, opened 2022-09-24
Fix company field definitions

scottcwilson:
In 1.3.8, the lengths of the _company fields were increased but there was no conversion done for people who had older databases.  All customers who have been using Zen Cart since 1.3.7 or prior have this issue.


scottcwilson:
Ugh this has an extra file in it - sorry.  Let me fix.

PR #5260, opened 2022-09-26
PHP8.2 Define Parameters breadcrumb

brittainmark:
# Class breadcrumb
set _trail to protected

# functions_products
modify to use class function removeLast() instead of directly accessing the class parameter.


brittainmark:
Part of #5103

PR #5261, opened 2022-09-26
fixes #4702; missing parentheses in sql; square

### LINKED ISSUE #4702: bug in reset tokens in square
https://github.com/zencart/zencart/blob/dc22072b1a1ea66b3e4fce9efe781ca43557f5e2/includes/modules/payment/square.php#L601-L609

lines 605 and 607 need to end `')");` no `'");`

apparently no one is resetting their tokens.

proseLA:


scottcwilson:
Thanks - I was just thinking about this one the other day.

scottcwilson:
I think only the second line change is required though - the first and the third look wrong.  Would you please check again?

proseLA:
not enough coffee this morning.  

i was initially correct on #4702.  two changes are required.  not four.

good catch.

best.

scottcwilson:
Gold, Jerry, Gold!

PR #5268, opened 2022-09-28
Allow apostrophes in search string

scottcwilson:
Test: Add product called "Tim's Coffee"
Search for "Tim's Coffee"

### DIFF
--- a/includes/functions/functions_search.php
+++ b/includes/functions/functions_search.php
@@ -199,7 +199,7 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
 
                         $where_str .= $sql_add;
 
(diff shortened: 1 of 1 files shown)


lat9:
Just be careful of the Apple "Smart Quotes", e.g &lsquo; and &ldquo;.

scottcwilson:
This is meant to handle an apostrophe which can be typed in from the keyboard.

lat9:
> This is meant to handle an apostrophe which can be typed in from the keyboard.

Unfortunately, that "Smart Quotes" setting/feechur on Apple devices, when enabled, converts the single/double quotes to the &lsquo; and &ldquo; characters when the user presses the quote key on the keyboard.

scottcwilson:
I have smart quotes on and tested with it.

PR #5274, opened 2022-09-30
Set true in zen_prepare_date

### LINKED ISSUE #5273: Sql Error in products_price_manager
**Describe the bug**
A SQL error occours when you leave a date blank in products_price_manager

**Version**
Zen Cart version: 1.5.8 and 1.5.7
PHP version: 7.4,8.0

**To Reproduce**
1. Go to 'Catalog>Products Price Manager'
2. use dropdown to select a category with products
3. Select a product
4. Click on edit product
5. edit product detail leaving one of featured start, featured end, specials start, Specials end blank
6. click update Price changes
7. White screen is displayed with a sql error.
8. log file is produced.

**Expected behavior**
no error

**Debug/Error Logs**
```
[30-Sep-2022 16:10:52 Europe/London] Request URI: /Test158/admin/index.php?cmd=products_price_manager&products_filter=205&current_category_id=67&action=update, IP address: 127.0.0.1
#1  trigger_error() called at [/home/lotus/public_html/innerlightcrystals/Test158/includes/classes/db/mysql/query_factory.php:667]
#2  queryFactory->show_error() called at [/home/lotus/public_html/innerlightcrystals/Test158/includes/classes/db/mysql/query_factory.php:634]
#3  queryFactory->set_error() called at [/home/lotus/public_html/innerlightcrystals/Test158/includes/classes/db/mysql/query_factory.php:275]
#4  queryFactory->Execute() called at [/home/lotus/public_html/innerlightcrystals/Test158/admin/products_price_manager.php:159]
#5  require(/home/lotus/public_html/innerlightcrystals/Test158/admin/products_price_manager.php) called at [/home/lotus/public_html/innerlightcrystals/Test158/admin/index.php:11]
--> PHP Fatal error: 1292:Incorrect date value: '' for column 'specials_date_available' at row 1 :: UPDATE zen_specials

                      SET specials_new_products_price = 95,

                          specials_date_available = '',

                          specials_last_modified = NOW(),

                          expires_date = '2022-10-01',

                          status = 1

                      WHERE products_id = 205 ==> (as called by) /home/lotus/public_html/innerlightcrystals/Test158/admin/products_price_manager.php on line 159 <== in /home/lotus/public_html/innerlightcrystals/Test158/includes/classes/db/mysql/query_factory.php on line 667.
```

brittainmark:
fix #5273

### DIFF
--- a/admin/includes/functions/datepicker.php
+++ b/admin/includes/functions/datepicker.php
@@ -42,13 +42,16 @@ function zen_datepicker_format_forsql()
 }
 
 /**
- * Convet Date to the corect farmat
+ * Convert Date to the correct format
  * @param string $raw_date Date value
- * @param boolean $past_date_null Set the date value to null is the dat is in the past
+ * @param string $past_date is the value to set the date if it is in the past or empty
  * @return string
  */
-function zen_prepare_date($raw_date, $past_date_null = false)
+function zen_prepare_date($raw_date, $past_date = '')
 {
+    if (empty($raw_date)) {
+        return $past_date;
+    }
   $date = zen_db_prepare_input($raw_date);
   if (DATE_FORMAT_DATE_PICKER != 'yy-mm-dd' && !empty($date)) {
     $local_fmt = zen_datepicker_format_fordate();
@@ -58,8 +61,8 @@ function zen_prepare_date($raw_date, $past_date_null = false)
       $date = $dt->format('Y-m-d');
     }
   }
-  if ($past_date_null == true) {
-    $date = (date('Y-m-d') < $date) ? $date : 'null';
+  if (!empty($past_date)) {
+    $date = (date('Y-m-d') < $date) ? $date : $past_date;
   }
   return $date;
 }
--- a/admin/products_price_manager.php
+++ b/admin/products_price_manager.php
@@ -102,14 +102,14 @@ if (!empty($action)) {
       } else {
         $master_categories_id = $_POST['master_categories_id'];
       }
+// Set date to null if empty or in the past
+      $products_date_available = zen_prepare_date($_POST['products_date_available'], 'null');
+// Set dates empty equivelent '0001-01-01' if empty or in the past
+      $specials_date_available = zen_prepare_date($_POST['specials_date_available'], '0001-01-01');
+      $specials_expires_date = zen_prepare_date($_POST['specials_expires_date'], '0001-01-01');
 
-      $products_date_available = zen_prepare_date($_POST['products_date_available'] ?? 'null', true);
-
-      $specials_date_available = zen_prepare_date($_POST['specials_date_available'] ?? 'null');
-      $specials_expires_date = zen_prepare_date($_POST['specials_expires_date'] ?? 'null');
-
-      $featured_date_available = zen_prepare_date($_POST['featured_date_available'] ?? 'null');
-      $featured_expires_date = zen_prepare_date($_POST['featured_expires_date'] ?? 'null');
+      $featured_date_available = zen_prepare_date($_POST['featured_date_available'],'00001-01-01');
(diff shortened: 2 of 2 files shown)


proseLA:
https://github.com/zencart/zencart/blob/fa92a2667deef542e812e3d78255ec25d64e16d7/admin/includes/functions/datepicker.php#L50

why not set the default to true?

scottcwilson:
Because this function is called in a bunch of places where that would be wrong (a real date is required).

proseLA:
> Because this function might be called in other places where that would be wrong (a real date is required).

i do not see that at all.

its a new function.

https://github.com/zencart/zencart/blob/fa92a2667deef542e812e3d78255ec25d64e16d7/admin/includes/functions/datepicker.php#L6

image

scottcwilson:
You're right - I withdraw my objection.  @brittainmark it's up to you; I am open either way.

brittainmark:
I must admit I don't really like the way this is being done. I had been looking at completely refactoring the module to not do all the work of preparing the date when it is not present. That way the work is done once instead of multiple checks. I'll have another look and get back one way or another.

brittainmark:
@proseLA , @scottcwilson 
I have redone this. I think that this is cleaner.
The dates are now set to the value passed to zen_prepare_date. This means that the logic later on in products price manager is less messy as all the work is done upfront.

NB. This does not prevent the entry of invalid dates. For that, you still get a white screen with an error message 
```
1292 Incorrect datetime value: '23' for column 'products_date_available' at row 1
...
If you were entering information, press the BACK button in your browser and re-check the information you had entered to be sure you entered valid data.
```
If you want, I could add additional checks to validate the dates entered as well.

scottcwilson:
> If you want, I could add additional checks to validate the dates entered as well.
One thing at a time. You have already made a nice improvement, let's get that merged.

proseLA:
@scottcwilson, is there a reason why you would merge this PR 4 minutes after it was pushed?  if you would prefer i do not do any code review, by all means, just let me know.  i will stop.

i will make a few comments.

- removal of the coalesce operator `??` is a mistake.  why was that done?
- my initial comment removed the 2nd operator from the call in numerous places.  which reduced code complexity IMO.
- now, we are calling this function with not even a boolean but this hard coded 'date' `(0001-01-01)`, which again why not make that value the function default if that's what you want?
- line 111 has a typo.
https://github.com/zencart/zencart/blob/7f2b75a3958d580598f836a50d81f071361c7312/admin/products_price_manager.php#L111

scottcwilson:
> is there a reason why you would merge this PR 4 minutes after it was pushed? 
That was when I was online.  More PRs welcome.

brittainmark:
Thanks for the feedback. I will correct the typo.
To answer your questions.
I removed the coalesce operator as the value is always there even if empty. The function then checks if it is empty. As the coalesce operator only checks if null or does not exist. It does not check blank.`''`
 I replaced the boolean with a string, so the function could be more flexible. The first check, products_date_available, needs to return "null" if not present while all the other dates need to return "0001-01-01" the equivalent of empty date. I set the default to '' in case the function needs to be used in other places. I could change the function to default to "0001-01-01". Would this seem odd if suddenly you needed "" and had to put that in the call?
On the plus side, it would have saved my typo.

proseLA:
you seem correct on the coalesce.

with most default installs now being configured to have mysql in strict mode, i would think that the whole point of this function is to prepare a value for use in a mysql statement when dealing with a datetime field.  so having the default be `0001-01-01` does not seem odd to me. i am not sure when needing an empty string would be called for. 

i do see where `'null'` might be used, as the `null` might be allowed in the mysql individual field spec.

i am also not sure when we would need to update the date to anything in the past other than the default, but i'm not opposed to having that flexibility.

and if that is true, why even include that check?  take the input; properly format to a mysql datetime without regard to present, past, future and move on.  if it is not in the proper format, return `0001-01-01` and call it a day.

PR #5278, opened 2022-10-01
select does not allow rel element

proseLA:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel

image

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -398,7 +398,7 @@ function zen_draw_input_field($name, $value = '~*~*#', $parameters = '', $requir
 function zen_draw_pull_down_menu($name, $values, $default = '', $parameters = '', $required = false)
 {
   $field = ($required ? '<div class="input-group">' . PHP_EOL : '');
-  $field .= '<select rel="dropdown" name="' . zen_output_string($name) . '"';
+  $field .= '<select name="' . zen_output_string($name) . '"';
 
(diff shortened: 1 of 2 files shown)


scottcwilson:
Seems like a good fix.  No visible change in testing?

Zen4All:
Did you notice your label tag is missing a closing bracket? I see `</label` in your picture instead of `</label>`.

proseLA:
> Did you notice your label tag is missing a closing bracket? I see `</label` in your picture instead of `</label>`.

hi erik,
feel free to bring it up here:
https://www.freeformatter.com/html-validator.html
the input file had all of the necessary closing brackets on every `label` tag.  and the validator did not complain about missing closing brackets.

i would encourage reading what the `rel` attribute is used for and what its potential values are.  `dropdown` is NOT a potential value. and it is not relevant to a `select` tag:

The rel attribute is relevant to the `link`, `a`, `area`, and `form` elements, but some values only relevant to a subset of those elements. Like all HTML keyword attribute values, these values are case-insensitive.

Zen4All:
I was just curious it was not giving any false positives. And after reading the documentation there is no doubt at all.

FYI, the `rel` attribute was added in Zen Cart version 1.3.0

scottcwilson:
I too checked for bad label closes and didn't find any.
Thanks for the fix!

PR #5291, opened 2022-10-06
admin, product-category listing search: add notifier point to modify query "where" string

torvista:
...allowing a plugin to modify the clause.


torvista:
I think this sql section needs reworking to better use one notifier.

PR #5293, opened 2022-10-07
Correct PayPal Checkout Flow

brittainmark:
Changes to set the $_SESSION\['cartID'] if the shopping cart has not had items added to it.

Alter logic to always find the Cart content count before restorecart method. So that if the cart is combined but no message is to be displayed, The before count is available for subsequent logic.

fix #5292

PR #5300, opened 2022-10-09
v158: Correcting redirect back to checkout_shipping when cart content…

### LINKED ISSUE #5292: Incorrect Checkout Flow Using PayPal Express Checkout
**Describe the bug**
If I put an item in the cart then click express checkout. I am taken to the paypal page.
I complete the paypal page, and I am returned to the checkout_payment.
If I complete the checkout_paymant page I am then sent to the checkout_shipping page.
The process then follows the normal route through checkout_payment then checkout_confirmation etc.

**Version**
Zen Cart version: 1.5.8

**To Reproduce**
1. Put item in cart without loging in.
2. Click on paypal express checkout.
3. login to paypal and pay for item
4. On return you are in the in the checkout_payment page with all the shipping etc calculated.
5. Click continue. You are sent to the checkout_shipping page.

**Expected behavior**
At step 5 you should be sent to the checkout_confirmation page.

lat9:
…s unchanged

Fixes #5292

Supercedes PR #5293

### DIFF
--- a/includes/modules/payment/paypalwpp.php
+++ b/includes/modules/payment/paypalwpp.php
@@ -2826,31 +2826,40 @@ if (false) { // disabled until clarification is received about coupons in PayPal
     // click of the PPEC button.
     //
     if ($_SESSION['paypal_ec_temp'] === false) {
-    // bof: contents merge notice
-    // save current cart contents count if required
-        if (SHOW_SHOPPING_CART_COMBINED > 0) {
-          $zc_check_basket_before = $_SESSION['cart']->count_contents();
+        // -----
+        // Determine the number of products currently in the cart, restore the customer's
+        // saved cart contents and determine the number of products now present in the cart.
+        //
+        $zc_check_basket_before = $_SESSION['cart']->count_contents();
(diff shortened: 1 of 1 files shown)


brittainmark (approved):
I don't know why 
```php
$zc_check_basket_before !== $zc_check_basket_after
```
```php
$zc_check_basket_before === $zc_check_basket_after
```
Are returning` false` when both are 1.

As I would rather see === used, I will change class  shopping_cart to return ->count_contents() cast to int

lat9:
@brittainmark, please re-review.  I've added some commentary and changed the comparisons based on the fact that the `count_contents` actually sums the in-cart products' quantities ... which are of type `float` instead of `int`.

brittainmark (approved):
Yes works a treat.  Sorry I missed cartcount being float.

brittainmark:
@lat9 will you make the php8.2 definition changes?

lat9:
@brittainmark, I've made the update to define the class variables for PHP 8.2.  That also resulted in some unused variables being removed and some assignments modified so that the commentary identifying the variable 'type' was accurate.

PR #5303, opened 2022-10-09
Cast ->count_content to int

### LINKED ISSUE #5302: Comparing $_SESSION['cart']->count_contents() with self using strict returns incorrectly
**Describe the bug**
```php
 $zc_check_basket_before = $_SESSION['cart']->count_contents();
 $_SESSION['cart']->restore_contents();
 $zc_check_basket_after = $_SESSION['cart']->count_contents();
```
when there is no change in the basket the comparisons return
 `$zc_check_basket_before !== $zc_check_basket_after` -> true
 `$zc_check_basket_before != $zc_check_basket_after` -> false

 `$zc_check_basket_before === $zc_check_basket_after` -> false
 `$zc_check_basket_before == $zc_check_basket_after` -> true

**Version**
Zen Cart version: 1.5.8

brittainmark:
To allow strict comparison.
Fix #5302
Needed for #5300

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -518,7 +518,7 @@ class shoppingCart extends base
             }
         }
         $this->notify('NOTIFIER_CART_COUNT_CONTENTS_END');
-        return $total_items;
+        return (int)$total_items;
     }
 
     /**


lat9:
Good idea, but a product's quantity value (which is summed up for that method's return) is a `float`, since certain products (like fabric or cheese) are sold in non-integer quantities &mdash; e.g. you might buy 1&half; pounds of cheese (here in the US, anyway).

brittainmark:
Thanks had not spotted that.

PR #5307, opened 2022-10-11
paypalwpp: handle missing data from Paypal

torvista:
Some trapping of missing data that sometimes occurs and causes debug logs.


torvista:
offtopic @brittainmark
for this file I added declarations as per debugs and this is working:
```
class paypalwpp extends base {
 //steve
    protected $_check, $buttonSourceDP, $cards, $cc_type_check, $codeTitle, $codeVersion, $doDPonly, $enableDirectPayment, $emailAlerts, $new_acct_notify;
// if protected causes a failure when incontext window opens
    public $form_action_url;
 //eof
```

brittainmark:
Reply to @torvista off topic. Saw that. Was wondering if @lat9 was going to make the change or waiting for me to.

PR #5318, opened 2022-10-16
Update products_price_manager.php

brittainmark:
Restore coalesce operator for Special and featured as not always present. Have set to `''` as there will be less processing. 
Sorry @proseLA  you were right that they were needed.


proseLA:
no worries.

now, i'm just waiting for you to come around to my way of thinking on the default value....

scottcwilson:
LOL we need this in specials.php too line 108-112.

PR #5323, opened 2022-10-17
Perform the database version update not performed on release.

### LINKED ISSUE #5322: Production distribution of database versioning (again) not up-to-date
During installation of the [1.5.8 production released version of Zen Cart](https://github.com/zencart/zencart/tree/v1.5.8), the database records that are created or touched still reflect the 1.5.8-dev version of the software.  This is a repeat issue of what happened when updating 1.5.7c->1.5.7d where the files did not get updated prior to the official release and did not get updated immediately following. 

For fresh installs and for upgrades to the ZC 1.5.8 release version, the database operations result in reporting that the database is to 1.5.8-dev version instead of the production version which has historically been just the portion up to the last number. (e.g. 1.5.8)

This gives the appearance after installation that the installation was not actually performed or is performed to an older version.

**Version**
Zen Cart version: 1.5.8
PHP version: Doesn't matter
MySQL version: Doesn't matter
Plugins that may be related: Doesn't matter
What older/newer versions were also tested? Doesn't matter

**To Reproduce**
Steps to reproduce/demonstrate the behavior: Install or upgrade version 1.5.8
1. In admin, go to Version link
2. Review the version information/history at the top of the screen area.

Expect that the files are updated before posting for distribution and also that when reviewing the information of an installed version that it report a database version the same as the software version. For an upgraded system, expect to see that the upgrade was performed to the then current sub-version. 

The biggest issue, which is something I dealt with on 1.5.7 is that even if all of the files are brought up-to-date, the fact that a production version is issued identifying that it is still a development version causes end users to really question whether work was complete or not. To independently set the files straight raises similar questions because of the differences then between the installed software and what is released.

mc12345678:
The version that has been released for production use is still reporting that it is a development version (1.5.8-dev). This is a repeat of what happened with 1.5.7d in not updating just prior to packaging the official version.

Fixes #5322


scottcwilson:
TY for catching.  We're building a checklist to make it easier to do this but this step was missed.

PR #5325, opened 2022-10-17
Remove implementation guide from distribution

scottcwilson:


docs/INSTALL.TXT
```diff
@@ -5,19 +5,20 @@ Welcome to Zen Cart(R) E-Commerce
 THE FOLLOWING IS A VERY SIMPLIFIED INSTALLATION DOCUMENT.
 You can find online documentation at: https://docs.zen-cart.com
 
-WE SUGGEST YOU USE THE PDF "Implementation Guide" PDF FILE IN THE /docs FOLDER FOR FULLY DETAILED INSTRUCTIONS.
+WE SUGGEST YOU USE ONLINE DOCS AND THE PDF "Implementation Guide" 
+
+https://docs.zen-cart.com/release/implementation-guide-v157.pdf
```
brittainmark, line 10: Should this be implementation-guide-v158.pdf.
I know it does not exist but may be confusing if implementing 1.5.8 so would need creating and minor mods

docs/INSTALL.TXT
```diff
@@ -5,19 +5,20 @@ Welcome to Zen Cart(R) E-Commerce
...
+
+FOR FULLY DETAILED INSTRUCTIONS.
 
 
 Zen Cart has a built-in automatic installation system. However to use this you must first unpack and upload the code to a compatible web server:
 
 1. Extract the Zen Cart ZIP file (and tell the unzip utility to *retain* folder structures)
 2. Upload the Zen Cart files and folders to your server using your FTP program
-3. You must make some folders writable. The details are in the /docs/Implementation-Guide.pdf and the HTML-formatted documentation in the /docs/ folder
+3. You must make some folders writable. The details are in the Implementation-Guide.pdf and the online documentation.
```
brittainmark, line 19: Here you mention Implementation-Guide.pdf here but above called it implementation-guide-v157.pdf
should you include a link? Or point them to Docs>release Documentation>Implementation Guide (PDF)

docs/INSTALL.TXT
scottcwilson: Thanks. I think we'll just leave off the version number moving forward.

docs/INSTALL.TXT
scottcwilson: Fixed thanks.

docs/INSTALL.TXT
brittainmark: Thats fine. The PDF mentions 1.5.7 so will also need changing if you are going to leave it off.

scottcwilson:
I have made a hash out of this. :( 
Starting again.

brittainmark:
Don't feel bad, I do it all the time : )

PR #5328, opened 2022-10-19
Remove implementation guide from distribution

scottcwilson:
We don't want to distribute the docs - keeping things online means we can keep them fixed and up to date.


mc12345678:
If this is the case, it is yet further incomplete.

In reviewing the process and information provided from the downloaded files of #5473, When accessing the site (pre-installation), the page provides a link to installation documents within the server itself, specifically the docs folder.

Yes, the docs folder now rightfully contains `implementation-guide-v158.pdf`, but while I disagree with the local removal of the documentation the docs folder and `includes/templates/template_default/templates/tpl_zc_install_suggested_default.php` still indicate the presence of the docs that this PR says are no longer to be distributed.

scottcwilson:
This PR had to be un-done.  See #5331

Issue #4474, opened 2021-10-14
ZC158: shopfront warning when php <5.6, triggers php notice for undefined constant HTML_PARAMS


torvista:
Just in case there is anyone still taking any notice (ha), this Warning page references HTML_PARAMS, which has not been loaded. 

https://github.com/zencart/zencart/blob/35b347e7b04903ed50b4c1272f957bec3a53307f/includes/templates/template_default/templates/tpl_zc_phpupgrade_default.php#L19

It should probably use hardcoded values.

proseLA:
> Curious though, 1.5.8 is identified as to use/require PHP 7.3+: [PHP Version Requirements of Zen Cart](https://docs.zen-cart.com/user/first_steps/server_requirements/#php-version)
> 
> Is this/should this issue still be addressed where PHP is < 5.6? Won't there be other aspects of the software that won't work because code is written to support PHP 7.3+?

@mc12345678,  your comment makes little sense to me. this check is done in application_top, so nothing will work on the customer facing side.

@torvista has a valid point in his post, actually two....  a cursory look at the software would validate it.

proseLA:
i will respond last time here on this issue and then move on.  just to clarify my position.

- my comments did not advocate for hard coding the `HTML_PARAMS`.
- steve has a point that `HTML_PARAMS` will generate a notice.
- steve has a point (stated sarcastically) that it seems the lack of activity on this repo possibly portends an end to ZC.
- any discussion of the solution to this issue seems superfluous to me until some time as the aforementioned comment on the lack of activity by the repo's owners get resolved.
- your analysis of the issue is quite thorough.  except as stated above, it's time to step back and look at the big picture of what is actually happening right now.

scottcwilson:
This is really not worth fixing. 
The repo is open and fixes are being merged, but let's focus on things that really move the project forward.

scottcwilson:
This becomes a fatal error in PHP 8 that gives a white screen (yikes).  Fixed in #5332

PR #5332, opened 2022-10-19
The define HTML_PARAMS may not be set if Welcome page is loaded

scottcwilson:
As mentioned in https://www.zen-cart.com/showthread.php?229055-Installer-shows-blank-page


mc12345678:
Wouldn't it be more appropriate use it if it's there and ignore if it is not? 

````php
<?php echo defined('HTML_PARAMS') ? HTML_PARAMS : '';
````
And then revisit issue #4474?

scottcwilson:
This is a pathological case.  Just show the error page; it's not worth optimizing.  No customer will ever see this.

PR #5334, opened 2022-10-20
Fix "duplicate column" issue in Tools > EZPages

scottcwilson:
Patch for #5343.  True fix is in #5342.


scottcwilson:
Commented out `DROP` statements in prior upgrade scripts.
People who upgrade AFTER the next update to 1.5.8 will get this fix as part of their upgrade; others will need to run the statements in Tools > Install SQL Patches.

PR #5346, opened 2022-10-24
Correct Centerbox Title

### LINKED ISSUE #5345: Title on Featured Products Always the Top Level Title Not the Current Sub Category
**Describe the bug**
If you have featured products and your site has sub categories. The title on the Featured centre box is Alway the first category.
e.g. you are in Home, not in a category
 featured title is "Featured Products"
when you go to Hardware
featured title is "Featured Products - Hardware"
when you are in Top Category>Second Level
featured title is still "Featured Products - Hardware" and not as I would expect  "Featured Products - Mice"
this does not matter how far down the tree you go the featured title is still "Featured Products - Hardware" 

Same issue for new products

It is correctly reducing the selection of products as you descent the tree,

**Version**
Zen Cart version: 1.5.8

**To Reproduce**
1. On test system
2. Goto Home:: Hardware
3. Featured and new products say - Hardware
4. Continue down to Mice
5. Featured and new products still say - Hardware
6. 
**Expected behavior**
Center box heading reflects level in tree.

**Debug/Error Logs**
none
 

**Additional context**
The issue is caused by using `$new_products_category_id` which is a string, in example above "1_9" using `(int)$new_products_category_id`  always returns 1. Suggest using` $current_category_id` instead. PR to follow.

brittainmark:
fix #5345

I have not changed the logic as it seemed unnecessary, can do if you want.

### DIFF
--- a/includes/modules/featured_products.php
+++ b/includes/modules/featured_products.php
@@ -79,8 +79,8 @@ if ($num_products_count > 0) {
   }
 
   if ($featured_products->RecordCount() > 0) {
-    if (isset($new_products_category_id) && $new_products_category_id !=0) {
-      $category_title = zen_get_category_name((int)$new_products_category_id);
+    if (!empty($current_category_id)) {
+      $category_title = zen_get_category_name((int)$current_category_id);
       $title = '<h2 class="centerBoxHeading">' . TABLE_HEADING_FEATURED_PRODUCTS . ($category_title != '' ? ' - ' . $category_title : '') . '</h2>';
     } else {
       $title = '<h2 class="centerBoxHeading">' . TABLE_HEADING_FEATURED_PRODUCTS . '</h2>';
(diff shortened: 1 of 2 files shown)


includes/modules/featured_products.php
```diff
@@ -80,7 +80,7 @@
 
   if ($featured_products->RecordCount() > 0) {
     if (isset($new_products_category_id) && $new_products_category_id !=0) {
-      $category_title = zen_get_category_name((int)$new_products_category_id);
+      $category_title = zen_get_category_name((int)$current_category_id);
```
scottcwilson, line 83: Should we be checking `current_category_id` in the if block above?

includes/modules/new_products.php
```diff
@@ -94,7 +94,7 @@
     }
 
     if (!empty($new_products_category_id)) {
-        $category_title = zen_get_category_name((int)$new_products_category_id);
+        $category_title = zen_get_category_name((int)$current_category_id);
```
scottcwilson, line 97: Should we be checking `current_category_id` in the if block above?

brittainmark: yes have changed logic

PR #5349, opened 2022-10-25
Deal with empty lists in extras in PHP8

scottcwilson:


lat9:
Could you provide a scenario for one of those 'scripts' that resulted in failure?  Was it a log and/or incorrect output?

scottcwilson:
If you go to any of the pages under "Admin > Extras" running PHP 8.2 or 8.1 you'll see the logs.  No special setup.

PR #5351, opened 2022-10-26
Add language id to log for multilanguage sites

scottcwilson:
Rationale: newer stricter PHP versions are going to produce a lot more language file logs, and fixing them can be tricky in multilanguage carts if you're not sure which language was being used when the log was created.


lat9 (changes requested):
1. Unconditionally add the current `languages_id`, without casting (just in case there's some kind of issue).
2. Add it with a leading ', ' (no new line), keeping all the 'static' information on a single line; the error message itself will start on the line below.

lat9 (approved):

PR #5361, opened 2022-10-30
Change Load Point to 200

brittainmark:
To conform with the latest load point recommendation, move the load point to 200. 
Change array to []

### DIFF
--- a/includes/auto_loaders/config.products_viewed_counter.php
+++ b/includes/auto_loaders/config.products_viewed_counter.php
@@ -12,8 +12,12 @@
 if (!defined('IS_ADMIN_FLAG')) {
  die('Illegal Access');
 }
-$autoLoadConfig[190][] = array('autoType'=>'class',
-                              'loadFile'=>'observers/class.products_viewed_counter.php');
-$autoLoadConfig[190][] = array('autoType'=>'classInstantiate',
-                              'className'=>'products_viewed_counter',
-                              'objectName'=>'products_viewed_counter');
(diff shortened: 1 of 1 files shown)


lat9:
What's the problem being solved here?

brittainmark:
The module was used as an example for the documentation. I was bringing it in line with the desire for observers to start at 200 not 190. Don't think required any more as the docs have been changed to us a fictious example.

lat9:
I'll note, too, that the load point for an observer depends (er) on what it's observing.  For instance, if an observer is watching for additions/removals from the cart, it should load prior to load-point 140 (where `init_cart_handler.php`) does that processing.

scottcwilson:
I will add this guidance.  Thanks for mentioning it.

scottcwilson:
It's already there.
https://docs.zen-cart.com/dev/code/notifiers/#choosing-when-to-load-an-observer

lat9:
> It's already there. https://docs.zen-cart.com/dev/code/notifiers/#choosing-when-to-load-an-observer

Well ... there's a heading with one sentence.  An example (like above) would provide a tad more help.

scottcwilson:
Cindy, can you give me a good example from one of your plugins to use on 
https://docs.zen-cart.com/dev/code/notifiers/#choosing-when-to-load-an-observer

Actually same request for 
https://docs.zen-cart.com/dev/code/notifiers/#manually-loaded-observers

lat9:
Sure, give me a day or so and I'll hunt something up

PR #5362, opened 2022-10-31
Place update category links button under cat table

scottcwilson:
Go to the multiple categories link manager page on product 2 in the demo data set:

/admin/index.php?cmd=products_to_categories&products_filter=2&current_category_id=4

As a floating button, the update category links button can occlude other parts of the page.
float

It's better to just put it under the table it operates on.

nofloat

### DIFF
--- a/admin/includes/css/products_to_categories.css
+++ b/admin/includes/css/products_to_categories.css
@@ -21,7 +21,7 @@ label, input[type="checkbox"] {
   -webkit-box-shadow: 0 0 6px 0 rgba(0, 0, 0, 0.8);
   -moz-box-shadow: 0 0 6px 0 rgba(0, 0, 0, 0.8);
   box-shadow: 0 0 6px 0 rgba(0, 0, 0, 0.8);
-  bottom: 200px;
+  bottom: 10px;
 }
 
 /*product name and model in Update Categories button*/


torvista:
I disagree. It should be located at the bottom of the viewport.
My categories list is 23 pages long. That's why the submit is not at the bottom of the list.
Probably some clever person can make the box draggable with js or bootstrap?

scottcwilson:
I would be open to putting at the bottom of the right hand box, but where it is now is a mess.

torvista:
That "mess" was the result of many days work to improve one of the most convoluted pages in the project into the condition it is in now.

It is safe to assume that users of this page have multiple categories, which will scroll off screen at some point.
Whether the submit is above or below the categories, or in both locations, will be inconvenient for someone. 
So make the floating box draggable, not static, is my opinion.
Or locate it below the infobox until offscreen/screen is scrolled, then floating.
Changes should be improvements.

scottcwilson:
I wonder if we could put it *above* the table instead of below.

torvista:
what about this
https://jqueryui.com/draggable/

scottcwilson:
Sure.

torvista:
After a load of experimentation, while draggable is easy....putting it so it stays put when scrolling and is draggable has exhausted my patience. None are as good as it is now.

I think the best solution is change the products_to_categories.css to bottom: 10px;
```
.floatButton {
  -webkit-box-shadow: 0 0 6px 0 rgba(0, 0, 0, 0.8);
  -moz-box-shadow: 0 0 6px 0 rgba(0, 0, 0, 0.8);
  box-shadow: 0 0 6px 0 rgba(0, 0, 0, 0.8);
  bottom: 10px;
}
```
So it's always in the bottom right corner.

scottcwilson:
ok, but I thought you didn't like the bottom corner as a placement.

torvista:
don't care anymore

PR #5370, opened 2022-11-02
Allow script to autorun at template change time

scottcwilson:


scottcwilson:
Example use case:  `includes/templates/bootstrap/template_init.php`  is 

```
<?php
$db->Execute("UPDATE " . TABLE_CONFIGURATION . " SET configuration_value='Yes' WHERE configuration_key  = 'IMAGE_USE_CSS_BUTTONS'");
```

and an other template sets the value to 'No' in its init script.  This allows you to change templates but have the design-intended appearance.

scottcwilson:
Documented in 
https://docs.zen-cart.com/user/template/template_switching

PR #5373, opened 2022-11-05
space as opposed to null to output function

proseLA:
in the original square module, the reference id was set to a session id:
https://github.com/zencart/zencart/blob/9910e3d1aa80466f9d59afb53ae494849a2bda8c/includes/modules/payment/square.php#L332
this was done as the order had not been created.  and in the new module, it is still not.

square_webPay makes use of this `square_admin_notification` script; but i am NOT setting a reference id of the session.  in fact, i am not setting a reference id at all.  and there has been no problem with tying requests from square back to zc.

as such, `getReferenceId()` returns null and causes the following deprecated error in php8.1:

```
[04-Nov-2022 17:10:31 America/Los_Angeles] Request URI: /zcdev/admin/index.php?cmd=orders&page=1&oID=133&action=edit, IP address: 192.168.14.74
#1 /var/www/zcdev/includes/functions/functions_strings.php(20): htmlspecialchars()
#2 /var/www/zcdev/includes/functions/functions_strings.php(41): zen_output_string()
#3 /var/www/zcdev/includes/modules/payment/square_support/square_admin_notification.php(43): zen_output_string_protected()
#4 /var/www/zcdev/includes/modules/payment/square_webPay.php(535): require('...')
#5 /var/www/zcdev/admin/orders.php(741): square_webPay->admin_notification()
#6 /var/www/zcdev/admin/index.php(11): require('...')
--> PHP Deprecated: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/zcdev/includes/functions/functions_strings.php on line 20.
```
passing a single space addresses this error and also provides for the future possibility of setting a reference id.

### DIFF
--- a/includes/modules/payment/square_support/square_admin_notification.php
+++ b/includes/modules/payment/square_support/square_admin_notification.php
@@ -40,7 +40,7 @@ if (!empty($transaction) && $transaction->getId()) {
     $outputSquare .= '<tr><td class="main">' . "\n";
     $outputSquare .= 'Reference: ' . "\n";
     $outputSquare .= '</td><td class="main">' . "\n";
-    $outputSquare .= zen_output_string_protected($transaction->getReferenceId()) . "\n";
+    $outputSquare .= zen_output_string_protected($transaction->getReferenceId() ?? ' ') . "\n";
     $outputSquare .= '</td></tr>' . "\n";
 
     $outputSquare .= '<tr><td class="main">' . "\n";


brittainmark:
Should a check be added to includes/functions/function_strings.php, function zen_output_string, to check for null before processing htmlspecialchars, as well.  It could be called from other places?
```php
function zen_output_string($string, $translate = false, $protected = false): string
{
    if ($protected === true) {
        if (is_null($string) === true) {
            return $string; 
        } else {
            $double_encode = (IS_ADMIN_FLAG ? FALSE : TRUE);
            return htmlspecialchars($string, ENT_COMPAT, CHARSET, $double_encode);
        }
    }

    if ($translate === false) {
        return strtr($string, ['"' => '&quot;']);
    }

    return strtr($string, $translate);
}
```

proseLA:
more important people can make that call....

there are more ways of killing a cat than choking it with cream...

scottcwilson:
> Should a check be added to includes/functions/function_strings.php, function zen_output_string, to check for null before processing htmlspecialchars, as well. It could be called from other places?

This seems like it will be needed as PHP gets stricter about these situations.  Please PR.

lat9:
Why a space instead of an empty string (`''`)?

proseLA:
> Why a space instead of an empty string (`''`)?

actually i thought about it yesterday.  and an empty string is better.

in addition, given that @scottcwilson seems amenable to correcting it in the string function, i am going to close this PR and correct it there.

@brittainmark, your solution would not work, and in fact produces a fatal error.  the method requires a return string; your solution returns `null`.  which casues:
```
PHP Fatal error: Uncaught TypeError: zen_output_string(): Return value must be of type string, null returned in...
```

brittainmark:
Is that the latest version? using (string)?

proseLA:
> Is that the latest version? using (string)?

i do not understand your question.

https://github.com/zencart/zencart/blob/e9ffb4bd9ba52f7016615e3626e1b5915eb37130/includes/functions/functions_strings.php#L16

is the latest version.  and it requires a return string as indicated by the return type declaration.

see:

https://www.php.net/manual/en/language.types.declarations.php#:~:text=example%20%233%20basic%20return%20type%20declaration

brittainmark:
I rewrote the function #5375 this uses (string) instead of testing for null. I have tested and it appears to work fine.

proseLA:
that PR is open.

i am not keeping up with very PR that is open.

i am also not a fan of just casting things every time they do not meet whatever condition is expected.

this method is used quite extensively; the idea that `null` gets passed to it, and we want to go through all of these other functions to return an empty string seems a waste of processing resources.

over the years, i have become a fan of the early return.

brittainmark:
Yes agree. Had coded similar to you to start with. I don't think I had spotted the return being string in the first case. Feel free to add to the discussion on #5375. I don't mind recoding again. This time I might get it right.

PR #5375, opened 2022-11-06
Add test for null

brittainmark:
Have added tests for null value in a few places to avoid depreciation errors in php8.2. An alternative is to place (string) before the parameter. e.g.
`strlen((string)$string)`

This also removes the depreciation warning. If you think this is a better approach, let me know and I'll redo.

As requested #5373

### DIFF
--- a/includes/functions/functions_strings.php
+++ b/includes/functions/functions_strings.php
@@ -15,6 +15,10 @@
  */
 function zen_output_string($string, $translate = false, $protected = false): string
 {
+    if (is_null($string) === true) {
+        return '';
+    }
+
     if ($protected === true) {
         $double_encode = (IS_ADMIN_FLAG ? FALSE : TRUE);
         return htmlspecialchars($string, ENT_COMPAT, CHARSET, $double_encode);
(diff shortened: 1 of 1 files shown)


brittainmark:
For info, just ran this test
```php
<?php
$mark = null;
$str = (string)$mark;
if (is_null($str)) echo "<br>is_null";  // not true
if ($str === null) echo "<br>===Null";  // not true
if ($str == null) echo "<br>==Null";
if ($str === '') echo "<br>===''";
if ($str ==' ') echo "<br>==''";

```
returns
==Null
===''
==''

lat9:
I'm finding that last one very interesting.  What version of PHP are you running to get that, essentially, `'' == ' '`?

brittainmark:
Sorry, that was a typo. Should have been `if ($str == '') echo "<br>==''";` otherwise it would have been very interesting. It was run on php 8.2RC5

lat9:
Whew, I was heading to make another pot of coffee trying to wrap my head around that!

For the three functions being corrected, would you please simply cast the presumed-to-be-string input to a `(string)` on entry?  I think it makes the code more readable.

brittainmark:
zen_trunc_string needs two (string) as it possible to reach line 120 `if (strlen((string)$str) <= $len) return $str;`
I have not placed any further down in the function, as you'd need to enter -ve $len to reach any further down the function.

lat9 (changes requested):
I'm liking the approach that @proseLA took with the quick-return; please update.

brittainmark:
Have recoded to follow @proseLA method and do quick return of empty string if null submitted.

lat9 (approved):

PR #5377, opened 2022-11-06
output empty string when fed a null #5373

proseLA:


brittainmark:
I think this is already covered by #5375

scottcwilson:
I actually like them both, but we'll let @lat9 decide since she's already involved.

lat9:
As mentioned in #5375, I'm liking the quick-return for readability.

PR #4641, opened 2022-02-19
change <br /> to <br>

torvista:
and correct 
`<br/>` missing space

Emails are still using  `<br />` although they probably can be changed too?

PR #5383, opened 2022-11-10
Chages to /email and /extras with " />"

dbltoe:
replace all ` />` with `/>`, then replaced all `<br/>` with `<br>`

Where's the emoji for crossed fingers?


scottcwilson:
I'm not positive `<meta>` needs a closing tag - I'm going to merge this, but would you please double check that after you update your branch?

This was a ton of work - thank you so much for doing this.  Little details like this make people realize Zen Cart is as up to date as any other option.

dbltoe:
Since 2013-4, meta has required a closing tag.  Only changes to that in html5 are sheme is no longer valid and charset is added.

https://html.com/tags/meta/

scottcwilson:
The page you linked above says, "This element must not contain any content, *and does not need a closing tag*."  (emphasis mine)

dbltoe:
That can be confusing.  Note that W3C considers the `/` as the closing tag.

From https://www.w3docs.com/learn-html/html-meta-tag.html

> The `<meta>` tag is empty, which means that the closing tag isn’t required. But in [XHTML](https://www.w3docs.com/learn-html/xhtml-extensible-hypertext-markup-language.html), the `(<meta>) `tag must be closed `(<meta/>)`.

scottcwilson:
No worries, just making sure I understood.  
I'll get your next change merged in by the end of the week.  Thanks again!

includes/functions/functions_email.php
```diff
@@ -112,7 +112,7 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
       // Build the email based on whether customer has selected HTML or TEXT, and whether we have supplied HTML or TEXT-only components
       // special handling for XML content
       if ($email_text == '') {
-        $email_text = str_replace(array('<br>','<br />'), "<br>\n", $block['EMAIL_MESSAGE_HTML']);
+        $email_text = str_replace(array('<br>','<br>'), "<br>\n", $block['EMAIL_MESSAGE_HTML']);
```
torvista, line 115: array now has duplicated elements, it was correct previously.

torvista:
How attitudes seem to have changed...https://github.com/zencart/zencart/pull/4641

proseLA:
can we please  be more selective here?  we have changed a file in the vendors directory...  i'm not a fan.  if you want to update the whole class, great.  

in addition, we are changing debug statements and lines that are commented out.  are these really necessary?

PR #5384, opened 2022-11-10
Change " />" in root, zc_install, and Admin root.

dbltoe:
`Changes " />" to "/>" and "<br/>" to "<br>" in the above folders.`

This leaves just a few files that need to be checked by someone more familiar with the format.


admin/mail.php
```diff
@@ -222,7 +222,7 @@ function check_form(form_name) {
               <?php
               $message_preview = empty($_POST['message']) ? $_POST['message_html'] : $_POST['message'];
               $message_preview = (false !== stripos($message_preview, '<br') ? $message_preview : nl2br($message_preview));
-              $message_preview = str_replace(array('<br>', '<br />'), "<br />\n", $message_preview);
+              $message_preview = str_replace(array('<br>', '<br>'), "<br>\n", $message_preview);
```
torvista, line 225: I don't think this change is correct

scottcwilson: I agree with @torvista

scottcwilson (changes requested):
See note from @torvista

dbltoe:
I agree but, as usual, changing the file back has resulted only in creating a new commit that sits with another commit and doesn't want to be sent to zencart/zencart without taking everything with it in a new PR.

There is absolutely no match in the GitHub docs for making a change to a file that has been committed.  At least, none that I can find.

scottcwilson:
Please don't modify 
zc_install/includes/systemChecks.yml
and don't change SHOW_SHOPPING_CART_BOX_STATUS and DEFINE_BREADCRUMB_STATUS in the .sql files or you'll break upgrades. 

It's probably better if you only change things in includes/ - everything 
else has a million pitfalls. Changing the .sql scripts and zc_install 
in particular are fraught. 

For your question
> There is absolutely no match in the GitHub docs for making a change to a file that has been committed. 

you should be able to check out the branch in question, modify the file, and do 

```
git add <file>
git commit --amend
git push --force
```

proseLA:
please do not merge.  please.

scottcwilson:
Agreed.

PR #5386, opened 2022-11-11
remove slash from selfclosingtag in ckeditor fixes #5385

### LINKED ISSUE #5385: Installed ckeditor is not html5 compliant
When one defines the main page, creates a category/product, or simply uses ckeditor to add or edit text in Zen Cart, a line break is saved as `<br/>`.

Even if the Source mode is used, a `<br> `entry is changed to `<br/>` by ckeditor when the item is saved.

https://stackoverflow.com/questions/41496028/how-to-use-br-instead-of-br-in-ckeditor

proseLA:
Fixes #5385


dbltoe:
THANX

PR #5389, opened 2022-11-13
Add legend to ezpages

scottcwilson:
Right now you have to click on an ezpage with the red LED to see the error message and understand what's wrong.  This change adds a legend just like the orders page has. 

legend

### DIFF
--- a/admin/ezpages.php
+++ b/admin/ezpages.php
@@ -505,6 +505,13 @@ if (!empty($action)) {
         </div>
         <?php echo '</form>'; ?>
       <?php } else { ?>
+<?php
+        // Additional notification, allowing admin-observers to include additional legend icons
+        $extra_legends = '';
+        $zco_notifier->notify('NOTIFY_ADMIN_EZPAGES_MENU_LEGEND', [], $extra_legends);
+?>
+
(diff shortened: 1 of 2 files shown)


admin/ezpages.php
```diff
@@ -505,6 +505,12 @@
         </div>
         <?php echo '</form>'; ?>
       <?php } else { ?>
+<?php
+        // Additional notification, allowing admin-observers to include additional legend icons
+        $extra_legends = '';
+        $zco_notifier->notify('NOTIFY_ADMIN_EZPAGES_MENU_LEGEND', [], $extra_legends);
+?>
+        <div class="row"><?php echo TEXT_LEGEND . ' ' . zen_image(DIR_WS_IMAGES . 'icon_status_red.gif', TEXT_WARNING_MULTIPLE_SETTINGS, 10, 10) . ' ' . TEXT_WARNING_MULTIPLE_SETTINGS. $extra_legends; ?></div>
```
lat9, line 513: Needs a space after `TEXT_MULTIPLE_SETTINGS`.

lat9 (changes requested):

lat9 (changes requested):
I guess I wasn't clear in my initial review.  It's not that a blank (' ') needed to be added for that line (since that would insert an unwanted space in the image's HTML tag, but that there needed to be a programmatic space between that constant and the concatenating . following,

lat9:
... specifically:
```php
<div class="row"><?php echo TEXT_LEGEND . ' ' . zen_image(DIR_WS_IMAGES . 'icon_status_red.gif', TEXT_WARNING_MULTIPLE_SETTINGS, 10, 10) . ' ' . TEXT_WARNING_MULTIPLE_SETTINGS .  $extra_legends; ?></div>
```

lat9 (approved):

PR #5394, opened 2022-11-20
Initialize $shipping to prevent log

scottcwilson:


lat9:
An example of the generated log?  Just for the record!

scottcwilson:
Sorry, there were so many logs I didn't capture them all.   But it looks the same as #5388 - something like 

```
--> PHP Warning: Undefined variable $shipping in /home/thatsoftwareguy4/public_html/SITE/includes/modules/shipping/table.php on line 189.
```

PR #5400, opened 2022-11-22
remove unnecessary isset and simplify clause

torvista:
make it quick, more to do.


scottcwilson:
Great, easier to read.  TY.

PR #5406, opened 2022-11-23
Restore `<br />` replacement for `functions_email.php`

lat9:
The `<br />` needs to be there; was removed in PR #5383


scottcwilson:
Thanks to you & @torvista for catching and fixing.

Issue #3664, opened 2020-07-06
Where/how to capture/pass item of concern


mc12345678:
Back in ZC 1.5.6, changes were made to the way upgrades were done through ZC 1.5.3, and through 1.5.5. Specifically, the upgrade script for Zen Cart 1.5.3 adds the country 'South Sudan' whether that country is in the database or not. In the ZC 1.5.5 upgrade the countries 'Curaçao' and 'Sint Maarten (Dutch part)' are added the same way. What this means (and this primarily only affects those in this condition/situation) is that if a store that appears to have a database version less than 1.5.3 and already had added any of those three countries to the store (including using the original default country numbers of 246, 247, and 248, respectively) then as they go through an upgrade to 1.5.6 or above, then the country/countries will be added again.

Now, to also recognize another perspective, whether a number was assigned or not, the upgrade process never ensured that a new country would not be duplicated. Certainly by having an "assigned" number it becomes easier to troubleshoot and determine that there is/has been a difference and offers one the idea to investigate how/why there is/was a difference.  Further, there are those that do/will manually take control of the countries they wish to have listed which may incorrectly include deleting the country.

I was looking for a possible place in the docs files, but then was wondering about the docs site, then perhaps a forum thread, etc... So, in a way looking for a suggestion of where the above may be captured to maybe help those that later find that they have the same country listed twice and/or have problems with zone type configurations that include/associate those three countries (however rare one or more people think it may be).

drbyte:
Is someone actually running into a chaos situation where the numbers assigned for these countries is messing things up? 
Since it created the country's associated zones in tandem with the country, any given store will have its own data in sync with itself. 
Sure, the id numbers may not match the id numbers in other stores, but that's another issue. Canonically the store is still in sync with itself. 
I haven't run into cases where these specific factors you're mentioning (specific countries and their zones) have posed a functional problem.

Have you example cases?

drbyte:
The real root cause is really a design flaw inherited from OsC that we haven't fixed: all country references should have been on ISO-3 code, not on an autoincrementing id. 
Changing that is a big overhaul. But off-topic for the issue at hand.

torvista:
> all country references should have been on ISO-3 code, not on an autoincrementing id.

Exactly.

lat9:
> 
> 
> Is someone actually running into a chaos situation where the numbers assigned for these countries is messing things up?
> Since it created the country's associated zones in tandem with the country, any given store will have its own data in sync with itself.
[...]

PR #5416, opened 2022-11-25
Observe the name change for Türkiye

mc12345678:
United Nations as of 2 June 2022 recognizes the name change and it should therefore be adopted here.

Keeping in line with the way that countries are now handled in Zen Cart, a new install will have Türkiye at country 215, while an upgrade will update the country name if it's countries_iso_code_3 is present. This is in line with the explanation/direction provided in issue #3664.

If the country has been deleted, then reinsertion would need to be by some other path.


scottcwilson:
The string in `zc_install/sql/install/mysql_utf8.sql` looks wrong - please check again.

lat9:
Added that status as a reminder that since it's a database update, it should be held until (a) corrected and (b) the next numbered release.

scottcwilson:
Thanks for that.  We will pull this on the next numbered release; it is listed in issue #5417.

drbyte:
re-opening for inclusion in 2.0.0

scottcwilson:
We shouldn't have put the Turkey change in the 1.5.8 updater - it should be in 2.0.0.  Will you fix @drbyte ?

Issue #5355, opened 2022-10-28
Admin, Option Values Sorter does not take language into account


torvista:
php:8.2.0RC5 /10.6.5-MariaDB
Multi-language store.

Admin, Catalog, Option Values Sorter: Choose a product to update the sort order.
The sort order is updated only for one language (the last installed), as the sql does not factor in the language.
https://github.com/zencart/zencart/blob/8972fc0fceb0f1d8ffd2828339063e3292bbc1d5/admin/option_values.php#L30-L32
Possibly one may expect this sql to update all instances, but it doesn't, in this environment at least.
I've done a fix to allow update of current language or all languages, to PR presently.

scottcwilson:
I'd rather it do all languages and not have a switch to choose current vs all.

torvista:
I have the option to update all as the default. But no doubt there are cases when someone wants something different, and it's no extra overhead, I've already done it.

PR #5426, opened 2022-11-26
Admin, Option Values Sort: fix sort per language, formatting, expand messagestack

torvista:
Impetus was a bug that sorting was updated for one language only
https://github.com/zencart/zencart/issues/5355

I worked on the first sort option only, and the formatting. There are other issues with the other sorting options, not fixed here.

Too much work to try and separate these changes, so...

Before:

aasasa2-11-26 22_59_23-Admin Option Values — Mozilla Firefox

Dropdown is a pointless size, unrelated to content.
2022-11-26 23_00_19-Admin Option Values — Mozilla Firefox

Edit, similar, occupying the whole page.
2022-11-27 00_10_41-Admin Option Values — Mozilla Firefox

Masessage after a successful sort
2022-11-27 00_11_06-Admin Option Values — Mozilla Firefox
as
After PR:
Main Page (other sort options are for another PR)
2022-11-27 00_07_10-Admin Option Values — Mozilla Firefox

Dropdown fits content:
2022-11-27 00_07_27-Admin Option Values — Mozilla Firefox

Edit page: 
2022-11-27 00_11_33-Admin Option Values — Mozilla Firefox

language options are only shown when more than one language installed
2022-11-27 00_13_33-Admin Option Values — Mozilla Firefox

Expanded message:
2022-11-27 00_11_51-Admin Option Values — Mozilla Firefox

### DIFF
--- a/admin/option_values.php
+++ b/admin/option_values.php
@@ -26,12 +26,34 @@ if (!isset($_GET['action'])) {
 switch ($_GET['action']) {
   case ('update_sort_order'):
     foreach ($_POST['options_values_new_sort_order'] as $id => $new_sort_order) {
-
-      $db->Execute("UPDATE " . TABLE_PRODUCTS_OPTIONS_VALUES . "
-                    SET products_options_values_sort_order = " . (int)$_POST['options_values_new_sort_order'][$id] . "
-                    WHERE products_options_values_id = " . (int)$id);
+            if (empty($_POST['language']) || $_POST['language'] == 'current_only') {
+                $db->Execute(
+                    "UPDATE " . TABLE_PRODUCTS_OPTIONS_VALUES . "
+                    SET products_options_values_sort_order = " . (int)$new_sort_order . "
+                    WHERE products_options_values_id = " . (int)$id . " AND language_id = " . (int)$_SESSION['languages_id']
+                );
+            } else {
+                $languages_array = zen_get_languages();
+                foreach ($languages_array as $language) {
+                    $db->Execute(
+                        "UPDATE " . TABLE_PRODUCTS_OPTIONS_VALUES . "
+                    SET products_options_values_sort_order = " . (int)$new_sort_order . "
+                    WHERE products_options_values_id = " . (int)$id . " AND language_id = " . (int)$language['id']
+                    );
     }
-    $messageStack->add_session(SUCCESS_OPTION_VALUES_SORT_ORDER . ' ' . zen_options_name($_GET['options_id']), 'success');
+            }
+            $messageStack->add_session(
+                sprintf(
+                    SUCCESS_OPTION_VALUES_SORT_ORDER_NAME,
+                    htmlentities(zen_options_name($_GET['options_id'])),
+                    (int)$_GET['options_id'],
+                    htmlentities($_POST['options_values_name'][$id]),
+                    (int)$id,
+                    (int)$new_sort_order
+                ),
+                'success'
+            );
+        }
     $_GET['action'] = '';
     zen_redirect(zen_href_link(FILENAME_PRODUCTS_OPTIONS_VALUES));
     break;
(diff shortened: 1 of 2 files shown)


admin/option_values.php
```diff
@@ -26,12 +26,34 @@
...
+                    $db->Execute(
+                        "UPDATE " . TABLE_PRODUCTS_OPTIONS_VALUES . "
+                    SET products_options_values_sort_order = " . (int)$new_sort_order . "
+                    WHERE products_options_values_id = " . (int)$id . " AND language_id = " . (int)$language['id']
+                    );
     }
-    $messageStack->add_session(SUCCESS_OPTION_VALUES_SORT_ORDER . ' ' . zen_options_name($_GET['options_id']), 'success');
+            }
+            $messageStack->add_session(
+                sprintf(
```
scottcwilson, line 46: Please don't change this.  The person doing this knows what he's changing and doesn't need all this feedback; the original simpler language define is enough.

scottcwilson (changes requested):

admin/option_values.php
torvista: >The person doing this knows what he's changing and doesn't need all this feedback

I would not assume to know the profile of the average admin user. 
Sure people know what they want, or what they think they want, and do what they think they want to do, and what happens is what they expected, mostly.
I don't think you can have too much feedback. Not everyone is working in the admin all the time, and is 100% confident of what they are doing/have done. I certainly think twice and check locally before doing a global change.
Since it is already done, is it a negative thing to use it?

torvista:
I must have closed this by accident.

admin/option_values.php
scottcwilson: Yes it is because it forces more language file editing on people who have already upgraded.  Please remove.

scottcwilson:
You'll have to create a new PR.   If you do so, please make the change I requested.

torvista:
I'll try harder not to waste people's time (especially mine) with improvements instead of just bugs.

torvista:
I'd like a second review of the improvements in this offering, given that many changes in the alpha seem to make concern about upgrading redundant.

admin/option_values.php
```diff
@@ -26,12 +26,34 @@
 switch ($_GET['action']) {
   case ('update_sort_order'):
     foreach ($_POST['options_values_new_sort_order'] as $id => $new_sort_order) {
-
-      $db->Execute("UPDATE " . TABLE_PRODUCTS_OPTIONS_VALUES . "
-                    SET products_options_values_sort_order = " . (int)$_POST['options_values_new_sort_order'][$id] . "
-                    WHERE products_options_values_id = " . (int)$id);
+            if (empty($_POST['language']) || $_POST['language'] == 'current_only') {
```
drbyte, line 29: Instead of having two nearly-identical queries here, it can probably be simplified with:
```php
if (empty($_POST['language']) || $_POST['language'] == 'current_only') {
    $languages_array = ['id' => $_SESSION['languages_id']];
} else {
    $languages_array = zen_get_languages();
}
foreach ($languages_array as $language) { 
...
```

torvista: I've not looked at this since my original PR so no doubt there will be further improvements.

I meant a review of the results.

admin/option_values.php
drbyte: > I'd like a second review of the improvements in this offering, given that many changes in the alpha seem to make concern about upgrading redundant.

Agreed on the point about the alpha.

drbyte:
> I'd like a second review of the improvements in this offering, given that many changes in the alpha seem to make concern about upgrading redundant.

Yes, submit a new PR.

I talked with Scott about the message-stack change. I think we'll be fine to go ahead with it.

torvista:
ok in one PR or I need to drip-feed it?

drbyte:
I support the changes you presented in this PR which got closed. 
Let's start with that.

Issue #5335, opened 2022-10-20
admin, attributes controller, additional actions debugs


torvista:
ZC158
php 8.2RC4
Choose graphics cards for example

Copy Options to Another Product

> --> PHP Warning: Undefined array key "copy_attributes" in D:\GitHub\zencart\admin\attributes_controller.php on line 780.

Copy Options to Whole Category

> --> PHP Warning: Undefined array key "products_update_id" in D:\GitHub\zencart\admin\attributes_controller.php on line 812.
> --> PHP Warning: Undefined array key "copy_attributes" in D:\GitHub\zencart\admin\attributes_controller.php on line 813.

scottcwilson:
Can you PR?

scottcwilson:
Can this be closed now?  Does #5450 suffice?

scottcwilson:
Checking back - can this be closed?

PR #5428, opened 2022-11-27
attributes controller, copy attributes: remove hidden field

torvista:
I don't see any need for the hidden field, which causes this debug: https://github.com/zencart/zencart/issues/5335
The field is set by the subsequent non-hidden radio selects.


torvista:
There are multiple similar errors which surface after this and the subsequent one....

scottcwilson:
Do you want to include those fixes in this PR?

torvista:
I think they are better separated.

scottcwilson:
The value `$_POST['copy_attributes']` is referenced on line 553 of admin/attributes_controller.php.   So it needs to be set. 
Would making line 780 

```
          <?php echo zen_draw_hidden_field('copy_attributes', ($_GET['copy_attributes'] ?? '')); ?>
```

work?

admin/attributes_controller.php
```diff
@@ -777,7 +777,6 @@ function popupWindow(url) {
           <?php echo zen_draw_hidden_field('products_filter', $_GET['products_filter']); ?>
           <?php echo zen_draw_hidden_field('products_id', $_GET['products_filter']); ?>
           <?php echo zen_draw_hidden_field('products_update_id', $_GET['products_update_id']); ?>
-          <?php echo zen_draw_hidden_field('copy_attributes', $_GET['copy_attributes']); ?>
```
scottcwilson, line 780: Maybe instead 

```
          <?php echo zen_draw_hidden_field('copy_attributes', ($_GET['copy_attributes'] ?? '')); ?>

```

scottcwilson (changes requested):

torvista:
> The value $_POST['copy_attributes'] is referenced on line 553 of admin/attributes_controller.php. So it needs to be set.

It **is** always set by one of the radio buttons in the same form:
https://github.com/zencart/zencart/blob/c34c15642a1b7dfdb6b042605e55acd9ed449ca4/admin/attributes_controller.php#L804-L810 

Removing it works for me...apart from uncovering the next error.
PHP Warning: Undefined array key "current_category_id" in D:\Github\zencart\admin\attributes_controller.php on line 559.
which should be in the same place!

PR #5429, opened 2022-11-27
v158: Enable template opt-in for single-column ...

### LINKED ISSUE #5262: Layout controller has no single setting anymore
search_header.php and any other header related box required a single setting instead of right and left and that option no longer appears in admin layout boxes controller.

If this has already been addressed, my apologies.

lat9:
... layout settings.

Fixes #5262.

### DIFF
--- a/includes/functions/functions_templates.php
+++ b/includes/functions/functions_templates.php
@@ -32,6 +32,7 @@ function zen_get_catalog_template_directories($include_template_default = false)
             continue;
         }
         if (file_exists($path . '/template_info.php')) {
+            unset($uses_single_column_layout_settings);
             require $path . '/template_info.php';
             // expects the following variables to be set inside each respective template_info.php file
             $template_info[$tpl_dir_name] = [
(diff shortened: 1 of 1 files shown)


scottcwilson:
Are we sure we want to support this, rather than just ripping it out and telling people to use the configuration settings that are already there?   What template uses this?

scottcwilson:
OK.  When you have a minute please write a sentence or two that I can add to https://docs.zen-cart.com/user/template/template_info/#full-specification-for-template_info-file

lat9:
Sure:

If you are going to distribute your template, you should build out the `template_info.php` file more completely, like this:

```
<?php  
$template_name = 'YOURTEMPLATE';  
$template_version = 'Version 1.0';  
$template_author = '<enter your name>';  
$template_description = 'describe your template'; 
$template_screenshot = 'screenshot.jpg';
$uses_single_column_layout_settings = (isset($uses_single_column_layout_settings)) ? $uses_single_column_layout_settings : true;
```
The `$uses_single_column_layout_settings` value indicates whether or not the zc158a (and later) admin's ***Layout Controller*** displays the 'SINGLE COLUMN' sort-order and status fields.  Your template's default (either `true` or `false`) indicates whether the template itself includes of sideboxes (like those which reside in the header) that check for the `layout_box_single_sort_order` and/or `layout_box_single_status` in determining whether or not their content is displayed.  It's a good idea (as in the example above) to check to see if that variable already exists, in case a store needs to further control these settings.

Notice there is no closing PHP tag (`?>`) at the end of the file!

scottcwilson:
See also #5432.   Thank you Cindy!

Issue #5280, opened 2022-10-04
use of NOTIFY_ADMIN_PROD_LISTING_DEFAULT_ACTION


torvista:
I'm trying to add copy category code to the category_product_listing page.. I've added a new notifier for the new icon observer, ok.

Now, the new action "copy_category" requires handling, so I need to use a notifier-observer to replace a hard-coded "case copy_category" in the switch.

This is the ending section of the switch-case that handles the actions.
https://github.com/zencart/zencart/blob/c5cc24e860002674d94981c979b63ba4103df30b/admin/category_product_listing.php#L298-L313

I don't see how to use this notifier to handle a new case/break out to let subsequent code handle $action="copy_category" or rather I don't see the point of this notifier if the action is always deleted subsequently.
Am I supposed to set some new global variable in the observer for the additional new code for the info-box to use?

torvista:
And if the $action is not set/used,  by for example using another POST set in the observer, the formatting breaks with the new infobox:
`<div<?php echo (empty($action)) ? '' : ' class="col-xs-12 col-sm-12 col-md-9 col-lg-9 configurationColumnLeft"'; ?>>`
Making a "clean" plugin is more trouble than it's worth.

mc12345678:
Might suggest adding information into the notifier to allow control of clearing the action like below:

````diff
    case 'setflag_categories':
    case 'delete_category':
    case 'move_category':
    case 'delete_product':
    case 'move_product':
    case 'copy_product':
    case 'attribute_features':
    case 'attribute_features_copy_to_product':
    case 'attribute_features_copy_to_category':
      break;
    default:
+      $clearAction = true;
+      $zco_notifier->notify('NOTIFY_ADMIN_PROD_LISTING_DEFAULT_ACTION', null, $clearAction);
+      if ($clearAction == false) {
+        $action = $_GET['action'] = '';
+      }
+      unset($clearAction);
-      $zco_notifier->notify('NOTIFY_ADMIN_PROD_LISTING_DEFAULT_ACTION');
-      $action = $_GET['action'] = '';
      break;
  }
}
````

torvista:
Thanks for commenting, apologies for not replying sooner.
I'll revisit this asap.

Issue #5267, opened 2022-09-28
Admin: Copy Categories


torvista:
I've added this next to the Category-Products listing Move Category action.
It copies the category and subcategories structure only (to another parent category or duplicates it in the current one).
It does not copy the contained products.
Can I PR this? Adds one function,

scottcwilson:
I know some guy asked for this on the forum but I'm not convinced that it's generally useful.

scottcwilson:
Can you make this a plugin?

torvista:
It would be more useful if it also copied products at the same time, and I may do that at some point/that's more work I don't have time for now.
I was thinking this is a base start, and it's not as if there is a lack of space on the category action buttons and does not complicate the code more or mess up any gui. It seems a natural expansion of the current actions. There is copy for products, so why not for categories.
It's not a major change and is something to put on the new features list, so as I've said before why not add stuff if there is no real reason **not** to, rather than the reverse. Another repository/plugin...yawn.

scottcwilson:
> why not add stuff if there is no real reason not to ... 

Every change we make is that much more code that has to be maintained and debugged.  We want to keep the core small so that the functions it provides are needed by almost everyone.

https://docs.zen-cart.com/user/plugins/why_plugins/

torvista:
A plugin means adding a notifier for the extra button and making it public-ready, and suddenly I've spent more time getting it ready than it took to code it, all for a couple of dozen lines of code. I don't have the time to code it in the first place so the altruistic stuff gets left behind these days. That's why it's quicker to just stuff it in the core since it's already ready for that.

torvista:
Since implementing this, we've been using and liking it **a lot**.

![2022-10-25 16_25_04-Admin Category Product Listing](https://user-images.githubusercontent.com/4391026/197800506-5be112d6-d589-4ab6-953f-88f7dafa5863.gif)


As we say up North, "what's not to like?"

scottcwilson:
You can put out a plugin that's not "drop in" - like I said, I don't want to add things that are only desirable to a small number of people.

torvista:
https://github.com/torvista/Zen_Cart-Copy_Categories

PR #5431, opened 2022-11-28
category_product_listing: modify and add notifier for additional actions

torvista:
To allow an observer to handle additional actions.

NOTIFY_ADMIN_PROD_LISTING_DEFAULT_ACTION: as it stands it cleared any added action, as pointed out here https://github.com/zencart/zencart/issues/5280
so not of much use.

NOTIFY_ADMIN_PROD_LISTING_DEFAULT_INFOBOX: new to allow an infobox to be associated with the new action.

Done specifically to enable this mod:
https://github.com/zencart/zencart/issues/5267

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -306,9 +306,12 @@ if (!empty($action)) {
     case 'attribute_features_copy_to_category':
       break;
     default:
-      $zco_notifier->notify('NOTIFY_ADMIN_PROD_LISTING_DEFAULT_ACTION');
-      $action = $_GET['action'] = '';
-      break;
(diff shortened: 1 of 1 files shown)


proseLA:
i have no idea what formatting you are using here.  it matches none of the standards indicated in the docs.

can we PLEASE stop changing `$_GET` vars?  i think it is a bad practice.  the `$_GET` var gets sanitized, the script can use a localized var, ie `$action` and then we go on our way.

i have no idea where else this script may lead us, but i dislike seeing changing of `$_GET` vars other than their initial sanitizing.

torvista:
Better?

PR #5432, opened 2022-11-29
#5262: Update built-in templates to enable single-column ...

lat9:
... settings to be overridden in the admin's "Layout Controller" ... and serve as an example.


torvista:
// but if a site has set the value, the template honors that setting.

I'm not clear on this.

Bootstrap template info:
`$uses_single_column_layout_settings = true;`
and Layout Settings allows the single column selection to allow location in the header.

If this is changed to:
`$uses_single_column_layout_settings = (isset($uses_single_column_layout_settings)) ? $uses_single_column_layout_settings : false;`

it becomes false as expected, as so no single column option is shown.

Assuming $uses_single_column_layout_settings is set elsewhere.
it is unset prior to the file being read:
https://github.com/zencart/zencart/blob/1ff415427ebc65b214bf6f306c50d83b80fe6359/includes/functions/functions_templates.php#L34-L44

So, I'm not understanding this.

PR #5434, opened 2022-11-29
Restock flag

scottcwilson:
Current logic to inspect is bad; leads to logs like 

```
--> PHP Warning: Undefined array key "restock" in /site/store/admin/orders.php on line 319.
```

when not checked.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -316,7 +316,7 @@ if (!empty($action) && $order_exists === true) {
       break;
 
     case 'deleteconfirm':
-      zen_remove_order($oID, $_POST['restock']);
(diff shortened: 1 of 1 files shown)


admin/orders.php
```diff
@@ -316,7 +316,7 @@
       break;
 
     case 'deleteconfirm':
-      zen_remove_order($oID, $_POST['restock']);
+      zen_remove_order($oID, (!empty($_POST['restock']) && $_POST['restock'] == 'on'));
```
lat9, line 319: Please use exactly-equal-to (`===`).

PR #5436, opened 2022-11-30
Add check for page id parameter

scottcwilson:
Fixes log 

```
--> PHP Warning: Undefined array key "id" in /home/client/public_html/site/includes/modules/pages/page/header_php.php on line 21.
```

### DIFF
--- a/includes/modules/pages/page/header_php.php
+++ b/includes/modules/pages/page/header_php.php
@@ -18,8 +18,8 @@
 // This should be first line of the script:
 $zco_notifier->notify('NOTIFY_HEADER_START_EZPAGE');
 
-$ezpage_id = (int)$_GET['id'];
-if ($ezpage_id == 0) {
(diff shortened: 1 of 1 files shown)


mc12345678 (commented):
Ehhh? Why set to an empty quote instead of `0` when not set?

scottcwilson:
It comes out to the same thing, but if you think it's more readable, fine. Done.

includes/modules/pages/page/header_php.php
```diff
@@ -18,7 +18,7 @@
 // This should be first line of the script:
 $zco_notifier->notify('NOTIFY_HEADER_START_EZPAGE');
 
-$ezpage_id = (int)$_GET['id'];
+$ezpage_id = (int)($_GET['id'] ?? '0');
 if ($ezpage_id == 0) {
```
lat9, line 22: ... and then use exactly-equal-to!

PR #5438, opened 2022-12-01
Enable a plugin to add additional search fields ...

lat9:
... to the admin's category/product listing.


proseLA:
who is loving the `zen_build_keyword_search_function`?

i'm unsure whether you want to check if the return value is an array prior to doing an array merge.

lat9:
> who is loving the `zen_build_keyword_search_function`?
> 
> i'm unsure whether you want to check if the return value is an array prior to doing an array merge.

The thing with that function is that it's used _everywhere_, so you need  to check where you are when you receive any function-based notification.

The 'return value' is initialized to an array; if an observer has changed it to another data type, it should be "punished" with an error log.

scottcwilson:
LOL "punished."

scottcwilson:
Thanks @lat9 and @proseLA !

PR #5439, opened 2022-12-03
Add old signatures for backwards compatibility

scottcwilson:
These function signatures were changed in 1.5.8 in #4083; let's add back the old signatures so that plugins won't break.  We can add a deprecated log in the future.

Fixes issue described in https://www.zen-cart.com/showthread.php?229167-Functions-removed-from-1-5-8-discussion

### DIFF
--- a/includes/functions/functions_general_shared.php
+++ b/includes/functions/functions_general_shared.php
@@ -412,3 +412,28 @@ function zen_get_admin_name($id = null)
     $result = $db->Execute($sql);
     return $result->RecordCount() ? $result->fields['admin_name'] : null;
 }
+
+// Compatibility 
+
+function zen_draw_products_pull_down($field_name, $parameters = '', $exclude = [], $show_id = false, $set_selected = 0, $show_model = false, $show_current_category = false, $order_by = '', $filter_by_option_name = null)
+{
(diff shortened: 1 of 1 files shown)


scottcwilson:
@lat9 Please merge asap.

Note to self: need to update working copy of release specific upgrade considerations, update 158 Known Bugs once this is finalized.

torvista:
https://docs.zen-cart.com/user/upgrading/release_specific_upgrade_considerations/

lat9 (changes requested):
For each, let's not defer the creation of the deprecated log.  Otherwise, it'll just take more time for plugins to "catch up".

scottcwilson:
@torvista The changes were not documented.  But even documenting them would not be sufficient; there's no need to break things unless you have literally no choice.

lat9 (approved):

torvista:
>there's no need to break things unless you have literally no choice.

If it is decided that similar functions should be culled and replaced with something more efficient, that's progress.
But yes the deprecated functions should have remained temporarily with warnings.
And as long as that is documented, I don't see how upgraders can complain. If people are upset about change, why do they upgrade?

scottcwilson:
For search: 
Old name: zen_draw_products_pull_down
New Name: zen_draw_pulldown_products

PR #5441, opened 2022-12-05
Don't show search box during special add by ID

scottcwilson:
Repro steps: 
- Using demo dataset, go to Admin > Catalog > Specials 
- click "Add Special by Product ID" and enter id 15. 
- Observe that add screen shows a search field.  It should not.

### DIFF
--- a/admin/featured.php
+++ b/admin/featured.php
@@ -212,7 +212,7 @@ if (!empty($action)) {
 //          }
         }
 
-          if ($action === 'new') {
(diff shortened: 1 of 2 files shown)


scottcwilson:
Thanks - wasn't aware that was an issue.

torvista:
much of the two files are identical/worth keeping identical

torvista:
While removing duplication is a generic goal, who is ever going to bother for these files, for no gain whatsoever.

PR #5442, opened 2022-12-05
Banner manager

torvista:
Minor code tweaks, no functionality changes yet.
To squash after being reviewed.

### DIFF
--- a/admin/banner_manager.php
+++ b/admin/banner_manager.php
@@ -8,7 +8,7 @@
 require 'includes/application_top.php';
 require 'includes/functions/functions_banner_graphs.php';
 
-$action = (isset($_GET['action']) ? $_GET['action'] : '');
+$action = ($_GET['action'] ?? '');
 if (isset($_GET['flagbanners_on_ssl'])) {
   $_GET['flagbanners_on_ssl'] = (int)$_GET['flagbanners_on_ssl'];
 }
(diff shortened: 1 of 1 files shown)


admin/banner_manager.php
```diff
@@ -142,7 +142,7 @@
           'status' => $status,
           'banners_open_new_windows' => $banners_open_new_windows,
           'banners_on_ssl' => $banners_on_ssl,
-          'banners_sort_order' => (int)$banners_sort_order,
```
scottcwilson, line 145: I don't think you should remove this.

admin/banner_manager.php
```diff
@@ -374,7 +374,7 @@ function popupImageWindow(url) {
             <?php echo zen_draw_label(TEXT_BANNERS_GROUP, 'banners_group', 'class="col-sm-3 control-label"'); ?>
             <div class="col-sm-9 col-md-6">
               <?php echo zen_draw_pull_down_menu('banners_group', $groups_array, $bInfo->banners_group, 'class="form-control" id="banners_group"'); ?><br>
-              <p><?php echo TEXT_BANNERS_NEW_GROUP; ?></p><?php echo zen_draw_input_field('new_banners_group', '', 'class="form-control" id="new_banners_group"', ((count($groups_array) > 0) ? false : true)); ?>
```
scottcwilson, line 377: Can the last parameter just be 
```
(count($groups_array) == 0)
```

scottcwilson (changes requested):

admin/banner_manager.php
torvista: Even though it's already an integer from line 107?
`$banners_sort_order = (int)$_POST['banners_sort_order'];`

admin/banner_manager.php
torvista: Yes. Much better.

torvista:
squashed

Issue #5443, opened 2022-12-05
Attribute Controller, Copy Attributes: PHP Warning: Undefined array key "current_category_id"


torvista:
Request URI: /zencart/admin/index.php?cmd=attributes_controller&action=update_attributes_copy_to_product,
IP address: 127.0.0.1, Language id 1
#0 D:\Github\zencart\admin\attributes_controller.php(559): zen_debug_error_handler()
#1 D:\Github\zencart\admin\index.php(11): require('D:\\Github\\zenca...')
--> PHP Warning: Undefined array key "current_category_id" in D:\Github\zencart\admin\attributes_controller.php on line 559.

PR to follow

PR #5444, opened 2022-12-05
Attribute controller: add missing field current_category_id

torvista:
fixes: https://github.com/zencart/zencart/issues/5443


scottcwilson:
This should be a one line change. Please don't make the other changes. If someone needs to grab this change to fix a debug log, let's make it easier for them by producing a small diff.

PR #5445, opened 2022-12-05
Remove salemaker_info page

scottcwilson:
Migrate contents of this page to the help system.


scottcwilson:
Note updated Sales help at 

https://docs.zen-cart.com/user/admin_pages/catalog/salemaker/

mc12345678:
For those that have historically had multi language admin help information, please be sure to add it to the Zen Cart docs site now that it won't be locally available. If a language other than (American?) English isn't yet made available, perhaps the desired language can be. Or, obviously this is a commit that when restored might offer a method of assistance for that language.

torvista:
Online docs facilitate changes. But the gui should contain the bare bones necessary for use.
This text is necessary for use and is not subject to change. I don't think removing it solves any problem, but may create one.
The language aspect is also valid.
I don't think the removal of this text is helpful.

scottcwilson:

PR #5448, opened 2022-12-07
Add alias for old name

### LINKED ISSUE #5447: Deletion of zen_get_orders_status will cause plugin issues
The function `zen_get_orders_status` was removed in
d79a2cdc4a09184874d77639b0df2b3f0df2b7ca

Plugins like shipworks still use it. 

As reported in 
https://www.zen-cart.com/showthread.php?229181-Quick-question-regarding-third-party-and-zencart-1-5-8

scottcwilson:
Fixes #5447

### DIFF
--- a/includes/functions/functions_general_shared.php
+++ b/includes/functions/functions_general_shared.php
@@ -437,3 +437,9 @@ function zen_draw_products_pull_down_categories_attributes($field_name, $paramet
    return zen_draw_pulldown_categories_having_products_with_attributes($field_name, $parameters, $exclude, $show_full_path, $filter_by_option_name);
 }
 
+function zen_get_orders_status() 
+{
+   trigger_error('Call to deprecated function; please use new names', E_USER_DEPRECATED);
+   return zen_get_orders_status_pulldown_array(); 
+}
+


proseLA:
have you tested this?

includes/functions/functions_general_shared.php
```diff
@@ -437,3 +437,9 @@ function zen_draw_products_pull_down_categories_attributes($field_name, $paramet
    return zen_draw_pulldown_categories_having_products_with_attributes($field_name, $parameters, $exclude, $show_full_path, $filter_by_option_name);
 }
 
+function zen_get_orders_status() 
+{
+   trigger_error('Call to deprecated function; please use new names', E_USER_DEPRECATED);
```
lat9, line 442: How about we give a clue as to what the new name is?

scottcwilson: They only have look at the traceback and go to this function, and they only have to do this once.

scottcwilson:
> have you tested this?
> 
> i'm not sure it will work.

It was a straight rename - you can compare the code in 1.5.7 vs 1.5.8.

proseLA:
> > have you tested this?
> > i'm not sure it will work.
> 
> It was a straight rename - you can compare the code in 1.5.7 vs 1.5.8.
> 
> There are other issues with shipworks, you are right, but this is one thing we can do, just as in the prior PR, #5439.

when i ask questions, rest assured i have already looked at the code.

the potential problem i see is that both of these functions were initially only on the admin side; by now moving the deprecated notice to the common functions, it could potentially cause a problem if done on the customer side.  but then again, i'm a fan of making most functions with limited exceptions available on both.  and for sure i'm a fan of not having different code for the same name on admin v storefront; an area where i think we have a lot of progress over the years.

scottcwilson:
The renames are now documented as a 1.5.8 Release Specific Upgrade Consideration.
https://docs.zen-cart.com/user/upgrading/release_specific_upgrade_considerations/#zen-cart-158

Merging this to keep the code consistent with the doc and forum responses.  We can make the error messages more specific to each rename in a separate PR if people decide it's important.

PR #5453, opened 2022-12-08
fix on full path in pulldown

### LINKED ISSUE #5451: function zen_draw_pulldown_categories_having_products_with_attributes: not showing full category path
For example in Catalog->Option Value Sorter, option For a Category has a category pulldown list

2022-12-07 18_29_50-Admin Option Values — Mozilla Firefox

This is using function zen_draw_pulldown_categories_having_products_with_attributes which eventually leads to the new-for-zc158 pulldown class.
There is an optional parameter in the function call to show the full category path. This is certainly something I need, but the parameter has no effect on the displayed category.

Am I missing something here?

proseLA:
fixes #5451 

i would suggest @torvista review prior to merge.

### DIFF
--- a/includes/functions/functions_categories.php
+++ b/includes/functions/functions_categories.php
@@ -389,15 +389,13 @@ function zen_draw_pulldown_products_having_attributes($field_name, $parameters =
  */
 function zen_draw_pulldown_categories_having_products($field_name, $parameters = '', $exclude = [], $show_id = false, $show_parent = false, $show_full_path = false, $filter_by_option_name = null)
 {
-    global $db, $currencies;
-
     if (!is_array($exclude)) {
         $exclude = [];
     }
 
     $pulldown = new categoryPulldown();
 
-    $pulldown->showID($show_id)->showParent($show_parent)->showFullPath($show_full_path)->exclude($exclude);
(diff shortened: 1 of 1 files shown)


proseLA:
> I do think this list needs to be sorted. It's not practical to find a specific category with them in their default order.

what is your suggestion?  last time i checked there is no model number associated with a category...

torvista:
DIfficult to answer.
There are category listings in multiple places which use the category sort order for the roots.

salemaker by cat id

while products to categories has them alpha-sorted
P2C correct

which would be my preference for categories

Nothing will satisfy everyone apart from being able to change the order next to every place it is used.

I think in this particular instance, using the sort order as per almost everywhere else would be logical.

scottcwilson:
@torvista Is it worth merging this and doing the sort order change later?  Or do you want to wait until sorting is resolved?

proseLA:
@scottcwilson please merge.  we can deal with sort order later and separately.

there was never an ability to sort category pulldowns in the past.

torvista:
What was broken is fixed, so merge.
Further work is optional.

scottcwilson:
Roger that.

scottcwilson:
Thanks for fix @proseLA

PR #5457, opened 2022-12-08
Banner Manager: correct id to match label for

torvista:


scottcwilson:
Nice catch.  Was there a symptom or did you discover it reading code?

torvista:
Your friend and mine the html validator, is still kicking ass.

scottcwilson:
Very good. I vaguely recall that in some cases (checkboxes?) if those two don't match, hilarity ensures. :(

torvista:
Maybe, if you don't go out much.

PR #5458, opened 2022-12-08
Banner Manager: html validation remove role from tr

torvista:
image

added cursor so you get a pointer on the line.

This is the last item for basic changes.

### DIFF
--- a/admin/banner_manager.php
+++ b/admin/banner_manager.php
@@ -511,11 +511,11 @@ if (!empty($action)) {
 
                   if (isset($bInfo) && is_object($bInfo) && ($banner['banners_id'] == $bInfo->banners_id)) {
                     ?>
-                    <tr id="defaultSelected" class="dataTableRowSelected" onclick="document.location.href = '<?php echo zen_href_link(FILENAME_BANNER_MANAGER, 'page=' . $_GET['page'] . '&bID=' . $bInfo->banners_id . '&action=new'); ?>'" role="button">
+                    <tr id="defaultSelected" class="dataTableRowSelected" onclick="document.location.href = '<?php echo zen_href_link(FILENAME_BANNER_MANAGER, 'page=' . $_GET['page'] . '&bID=' . $bInfo->banners_id . '&action=new'); ?>'">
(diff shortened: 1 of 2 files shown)


scottcwilson:
@dbltoe Was this done for ADA? If so we may need to keep even if invalid.

proseLA:
can we have a new policy of no new inline styles?

torvista:
> can we have a new policy of no new inline styles?

Agreed, I didn't notice it had its own stylesheet, so added it there.

scottcwilson:
> can we have a new policy of no new inline styles?
I think this is a good goal and leads to better code.  Would you please open a Discussion in Github on this so others could weigh in?  Let's keep the conversation out of this PR.

dbltoe:
I honestly have not spent much time looking at the admin.

That said, this is a good step in fixing multiple problems on that page.  It removes the WCAG complaint.

A problem that is pervasive in the admin is the use of links without any meaningful (discernible their term) text

`<a href="https://myzencarttemplates.com/temps4JetMar/index.php?cmd=banner_manager&amp;page=1&amp;bID=7&amp;action=setflag&amp;flag=1"><i class="fa fa-square fa-2x txt-status-off" aria-hidden="true" title="Status - Disabled"></i></a>`

So many times a font awesome icon is the only thing in a link.  That seems to be the main culprit over 

Not sure when I will get to the admin.  Got bootstrap storefront up to compliance and working on responsive_classic.  Going deeper than just CSS to meet all through AAA standards (moderate violations)

scottcwilson:
@dbltoe For some reason I thought this was your addition.  Thanks for checking.

dbltoe:
Seems like I have been all over the place lately.  😃 

Since the owner is not likely to sue themselves over ADA/WCAG compliance it's low on my current priorities.

Still it should be done and is on my list.

In the code above, the aria=hidden="true" is a direct A violation.

I will try to remember where I saw it, but somewhere in the storefront that adds a label to all empty alt tags in a certain area.

Code like that could be extremely helpful in the admin.

PR #5461, opened 2022-12-09
Banner stats: minor html

torvista:


admin/banner_statistics.php
```diff
@@ -178,7 +178,7 @@
           });
 
           function showTooltip(x, y, contents) {
-              $('<div id="tooltip">' + contents + '</div>').css({
+              $('<div id="tooltip">' + contents + '<\/div>').css({
```
scottcwilson, line 181: Are you sure? Why?

admin/banner_statistics.php
```diff
@@ -314,7 +314,7 @@ function showTooltip(x, y, contents) {
         </script>
       </div>
 
-<?php if (FALSE) { ?>
+<?php if (FALSE) { //https://github.com/zencart/zencart/issues/5016 ?>
```
scottcwilson, line 317: Let's just get rid of this.  We have source control if someone wants to resurrect it.

admin/banner_statistics.php
torvista: Surprise, surprise:

image

scottcwilson: thanks for clarifying.

PR #5464, opened 2022-12-09
Change access modifiers for PP vars

### LINKED ISSUE #5463: paypal var visibility issues...
as documented in [this forum post](https://www.zen-cart.com/showthread.php?229183-paypalwpp-php-(No-transaction-id)-in-the-email&p=1392057#post1392057), there are some vars whose visibility needs to be changed.

i would do it, but i despise paypal.

scottcwilson:
Fixes #5463 
Credit to @proseLA for debugging this.

As discussed in https://www.zen-cart.com/showthread.php?229183-paypalwpp-php

### DIFF
--- a/includes/modules/payment/paypalwpp.php
+++ b/includes/modules/payment/paypalwpp.php
@@ -121,9 +121,9 @@ class paypalwpp extends base {
   protected $reasoncode;
   protected $numitems;
   protected $amt;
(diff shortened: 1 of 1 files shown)


proseLA (approved):
i think all code should be self documenting!  
but it looks fine to me.

PR #5467, opened 2022-12-12
OrderStatusDashboardWidget.php minor bug

piloujp:
On the left down side of admin main page, when you click on any order status category it will always bring the full orders list instead of the chosen category.
This bug is due to a parameter name mistake on line 24 of '...\admin\includes\modules\dashboard_widgets\OrderStatusDashboardWidget.php' and can be easily fixed by replacing line 24:
`          <td><a href="<?php echo zen_href_link(FILENAME_ORDERS, 'selected_box=customers&status=' . $row['orders_status_id']); ?>"><?php echo $row['orders_status_name']; ?></a>:</td>` by
`          <td><a href="<?php echo zen_href_link(FILENAME_ORDERS, 'selected_box=customers&statusFilterSelect=' . $row['orders_status_id']); ?>"><?php echo $row['orders_status_name']; ?></a>:</td>`
This bug is not much but you can't see it if Zen-Cart is not populated with orders in different status categories.


scottcwilson:
This issue has already been fixed in #5440

piloujp:
Sorry for that.

I did search through github manually as I can't seem to find any search
function.

Is there a way to check efficiently? I am new to github and everything
seems to be counter-intuitive.

I hope I did not make to much trouble with a Pull Request. Afterward I
realized I should have post in issues category although doc says to do a
pull Request when you know the solution.

 

Again sorry for troubles.

scottcwilson:
PRs are always welcome.  But I encourage you to study the one you did, because it contained 276 file changes when it should have only contained 1.   You want to branch off v158, not main - that was your problem.

It can be challenging to know exactly what to search for but if you have identified the file that needs changing, you might want to do a git log to see if the current branch has a more up to date copy than you do.  Then work backwards to see what got changed.

PR #5469, opened 2022-12-12
Fixes default for dropdown Attributes not respected

### LINKED ISSUE #5468: v158 default for dropdown Attributes not respected
See #3734.

Steps to repo:
1. Create a product
1. Go to Admin > Catalog > Option Value Manager and add an option an option called "Test"
2. Add a new Value called "One"
3. Add a new Value called "Two"
1. Go to Attributes Controller, select the new product. Add "One" as an option, setting sort order = 1, Default Attribute = No, Used for Display Purposes = No.
1. Go to Attributes Controller, select the new product. Add "Two" as an option, setting sort order = 2, Default Attribute = Yes, Used for Display Purposes = No.
1. Go to storefront select the new product.

Expected: the selected attribute is "Two"
Actual: the selected attribute is "One"

Same issue but now in 1.5.8 version. The commit that fixes that issue is not included in v158.

Currently only first attribute option is selected on all dropdown attributes on navigating to the product page. And not the default option.

Applying d65a89e3688f39ed7878d798639183d03e3134cc corrects the problem.

nickwhaley:
Fixes #5468

Cherry pick of d65a89e3688f39ed7878d798639183d03e3134cc to add to v1.5.8.

### DIFF
--- a/includes/modules/attributes.php
+++ b/includes/modules/attributes.php
@@ -109,6 +109,7 @@ while (!$products_options_names->EOF) {
     $tmp_radio = '';
     $tmp_checkbox = '';
     $tmp_html = '';
+    $selected_attribute = $selected_dropdown_attribute = false; // boolean, used for radio/checkbox/select
 
     $tmp_attributes_image = '';
     $tmp_attributes_image_row = 0;
@@ -548,8 +549,9 @@ while (!$products_options_names->EOF) {
         // default
         // find default attribute if set for default dropdown
         if ($products_options->fields['attributes_default'] == '1') {
-            $selected_attribute = $products_options_value_id;
+            $selected_dropdown_attribute = $products_options_value_id;
         }
+        $selected_attribute = $selected_dropdown_attribute;
 
         $products_options->MoveNext();
         // end of inner while() loop


scottcwilson:
This doesn't seem like it changes anything.

dbltoe:
from the original https://github.com/zencart/zencart/issues/5468#issuecomment-1347416110

nickwhaley:
> This doesn't seem like it changes anything.

I thought the same thing when I read the code as well, but it will save the value of the default option when the loop finishes. Not really how I would write it, but I wasn't going to write it differently when it was already decided to be fixed this way when it was fixed in v157.

From original commit description:
```
Adds an internal variable to retain the value of $selected_attribute
throughout the loop process for use outside the loop. When the loop is
complete $selected_attribute will be equal to this internally tracked
variable so that the dropdown (and any other external tracking) will
see what the end result is/was for identifying the selected attribute.
```

scottcwilson:
See my comments in https://github.com/zencart/zencart/issues/5468#issuecomment-1347806098
Need repro steps to see the bug this is fixing; seems to work ok to me.

PR #5470, opened 2022-12-13
Restore off-page get forms

### LINKED ISSUE #3922: $_GET method of admin's zen_draw_form does not support action page other than self

### LINKED ISSUE #5454: Can no longer create an admin form that targets a different page with GET request
In past Zen Cart releases I've had a form on our admin dashboard that lets us quickly type an order ID and be taken to the Orders page editing that order, i.e.:

```
<?php echo zen_draw_form('orders', FILENAME_ORDERS, '', 'get'); ?>
<input type="hidden" name="action" value="edit">
  <label>Order ID: <input name="oID" type="number"></label>
</form>
```

However, the latest `zen_draw_form()` insists on writing a hidden `input` with `name="cmd"` and `value` taken from the current requests's `$_GET['cmd']`.  The resulting HTML is:

```
<form name="orders" action="https://dev.site/my-admin/index.php?cmd=orders" method="get">
  <input type="hidden" name="cmd" value="home">
  <input type="hidden" name="action" value="edit">
  <label>Order ID: <input name="oID" type="number"></label>
</form>
```

The cmd=orders in the action is overridden by the hidden field's value, and submitting the form takes you to the `home` target instead of `orders`.

It seems that the way `zen_href_link()` and `zen_draw_form()` work in the latest release conflict with each other, because they both try to insert a `cmd` parameter into the request, the one generated by `zen_href_link` in this case is overridden and ignored.

Ability to generate a form with customised `cmd` value different to the current page.

I wonder what the team think with their deeper knowledge of why and how this feature came about?

mc12345678:
Fixes #5454 and fixes #3922.

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -230,24 +230,36 @@ function zen_image_submit($image, $alt = '', $parameters = '')
 
 ////
 // Output a form
-  function zen_draw_form($name, $action, $parameters = '', $method = 'post', $params = '', $usessl = 'false') {
-    $form = '<form name="' . zen_output_string($name) . '" action="';
-    if (!empty($parameters)) {
-      $form .= zen_href_link($action, $parameters, 'NONSSL');
-    } else {
-      $form .= zen_href_link($action, '', 'NONSSL');
-    }
-    $form .= '" method="' . zen_output_string($method) . '"';
-    if (!empty($params)) {
-      $form .= ' ' . $params;
-    }
-    $form .= '>';
-    if (strtolower($method) == 'post') $form .= '<input type="hidden" name="securityToken" value="' . $_SESSION['securityToken'] . '">';
-    if (strtolower($method) == 'get') {
-      $form .= '<input type="hidden" name="cmd" value="' . (isset($_GET['cmd']) ? $_GET['cmd'] : 'home') . '">';
-    }
-    return $form;
+  function zen_draw_form($name, $action, $href_parameters = '', $method = 'post', $form_parameters = '', $usessl = 'deprecated') {
+      $url = zen_href_link($action, (!empty($href_parameters) ? $href_parameters : ''));
+      $form = '<form name="' . zen_output_string($name) . '" action="' . $url . '" method="' . zen_output_string($method) . '"';
+      if (!empty($form_parameters)) {
+          $form .= ' ' . $form_parameters;
+      }
+      $form .= '>';
+      if (strtolower($method) === 'post') {
+          $form .= '<input type="hidden" name="securityToken" value="' . $_SESSION['securityToken'] . '">';
+      }
+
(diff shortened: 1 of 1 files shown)


proseLA:
i like (and use) my[ javascript solution](https://github.com/zencart/zencart/issues/5454#issuecomment-1348750544)

mc12345678:
> i like (and use) my[ javascript solution](https://github.com/zencart/zencart/issues/5454#issuecomment-1348750544)

Which ***requires*** javascript to be enabled and the page to pass html validation to function out-of-the-box. Sure, could work, but why must core code be bypassed in this way to work?

proseLA:
if someone is running zen-cart v157 or higher without javascript, i have no answers for you. 

i also think valid html (to a point) is a worthy goal. 

as we are talking about the admin side, i find this to be an elegant solution for those who want to modify their admin experience. 

i'm not a fan of further obfuscation of these html functions. but that's just me.

scottcwilson:
Superceded by #5568.

Issue #4756, opened 2022-04-29
Addressing `strftime` deprecation in PHP 8.1 and additional date-formatting options


lat9:
Breaking this issue out as (initially) a discussion regarding the ultimate solution to the issue.  Much discussion is present in #4449 with possibilities presented by @proseLA.

lat9:
Starting by copying this pertinent post from the 'parent' issue:

despite #4553 being closed, i have not given up on this problem.  as i stated above, i believe a class file is the way to handle it.  the class file could be something such as:
```
<?php

    class zcDate
    {
        var $language_code;

        function __construct()
        {
            $this->language_code = setlocale(LC_TIME, 0);
        }

        function output(string $string = 'MMMM/dd/yyyy', int $timestamp = 0)
        {
            if ($timestamp === 0) {
                $timestamp= time();
            }
            $this->dateObject = datefmt_create(
                $this->language_code,
                IntlDateFormatter::FULL,
                IntlDateFormatter::FULL,
                date_default_timezone_get(),
                IntlDateFormatter::GREGORIAN,
                $string
            );
            return $this->dateObject->format($timestamp);
        }
    }
```
within `application_top` or somewhere of your choosing, we could create a var for use such as `$zcDate = new zcDate();` so that it is globally available.  we could then replace all `strftime` calls with `$zcDate->output(FORMAT, TIMESTAMP);`.  we could also eliminate the 12 months in all language files.

what would need to be decided is how to set the language_code in the `_construct` above.  and whether that should be a new constant or how that gets set. 

thoughts?

_Originally posted by @proseLA in https://github.com/zencart/zencart/issues/4449#issuecomment-1029566848_

proseLA:
some thoughts:

- new configuration value, despite how much i abhore adding another one.
- perhaps start minimizing the potential choices store owners have for date fields to 2 or 3.  is this even possible?

back when i had more time and was more motivated, i worked extensively on this problem.  the core of my work is the new class as seen here:

https://github.com/proseLA/zencart/blob/zcDate/includes/classes/zcDate.php

that branch has only 1 commit, which is seem here:

https://github.com/proseLA/zencart/commit/3cf8ccaa903959a931e5a48bbf34cd5937c50a53

i attempted to get at every instance of `strftime` and convert it using the php international date format.  one must have the php international date formatter installed and be running php 8.1 to use said format.

in theory, the output should remain the same.

the class has a `debug` method for comparing formats to see if they work.

[...]

PR #5471, opened 2022-12-13
zcDate class rewrite

piloujp:
zcDate class was built around the deprecated strftime and still uses its %something parameters. It was using Date function that does not support internationalization and for PHP > 8.1 only IntlDate object. Plus if PHP intl extension is not compiled/ activated when usins PHP > 8.1 you got an error calling unknown function which crash everything.

Now it is build around IntlDate object with fall back to Date function in case intl extension is not activated in PHP. A conversion is done too for backward compatibility if old strftime parameter are still used when calling zcDate->output.
For other files than zcDate.php, strftime parameters like %y, %m, %d ... were replaced by their equivalent IntlDate ones.

This has been tested on PHP 7.33 8.0.26 8.1.13 and a little bit with 8.2.0 on Apache 2.54. There were no error or warning in logs.
When intl extension is deactivated months and days of week are displayed in English whatever the local but everything else works fine.

### DIFF
--- a/zc_plugins/DisplayLogs/v3.0.2/admin/includes/languages/english/lang.display_logs.php
+++ b/zc_plugins/DisplayLogs/v3.0.2/admin/includes/languages/english/lang.display_logs.php
@@ -4,17 +4,17 @@
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: Steve 2021 Apr 02 New in v1.5.8-alpha $
-*/
+ */
 
 
 $define = [
     'HEADING_TITLE' => 'Display Debug Log Files',
-    'TABLE_HEADING_FILENAME' => 'File Name',
-    'TABLE_HEADING_MODIFIED' => 'Date Modified',
-    'TABLE_HEADING_FILESIZE' => 'File Size (bytes)',
-    'TABLE_HEADING_DELETE' => 'Delete?',
+    'TABLE_HEADING_FILENAME' => 'Filename',
+    'TABLE_HEADING_MODIFIED' => 'Date',
+    'TABLE_HEADING_FILESIZE' => 'Size (b)',
+    'TABLE_HEADING_DELETE' => 'Selected',
     'TABLE_HEADING_ACTION' => 'Action',
-    'BUTTON_INVERT_SELECTED' => 'Inverse Selection(s)',
+    'BUTTON_INVERT_SELECTED' => 'Invert Selection',
     'BUTTON_DELETE_SELECTED' => 'Delete Selected',
     'DELETE_SELECTED_ALT' => 'Delete all selected files',
     'BUTTON_DELETE_ALL' => 'Delete All',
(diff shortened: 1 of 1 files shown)


proseLA:
welcome to the zen-cart repo!

where were you back in may??

https://github.com/zencart/zencart/issues/4756#issuecomment-1117500593

i too feel that the `zcDate` should be using the international date formats as opposed to the old `strftime` formats; but i lost that battle long ago.

please provide exact setup to produce errors with current implementation, and the associated error logs.  v158 has been out in the wild for enough time that i would have thought errors would be coming up on the boards; which they have not.

welcome to the community!

piloujp:
It is nice to hear from you. At least I am not alone to want an
international Zen Cart.

Although I am a long time user I decided to go the developer side relatively
recently. Using Japanese as main language I can tell you there was never a
properly functioning Japanese Zen-Cart. I always had troubles with date
somewhere or another.

 

I think there is only one problem with actual code. When you have PHP under
8.1 it won't use datefmt_create() but Date() which can only display English
month names and week days. No error here but imagine how nice it looks on an
invoice.

If you use dateobject and PHP >= 8.1 then with datefm_create international
dates are ok.  Why 8.1? Dateobject exist for a while, looking at PHP doc it
was implemented in PHP 5.3.

 

About crash, I was not able to reproduce it which makes me feel it was one
of my test code left over somewhere but now removed. Sorry for that.

 

I have red the thread you mentioned and it was interesting. My position is
generally that rather than fixing old code until the end (deprecated) I
rather write a new one and fix it until it performs as well as the old one
and better. About zc_date I think the actual code is good even great but
based on deprecated parameters that you have to convert to new ones. I'd
rather do the opposite and it is why I submitted this new version. After
reading the thread I doubt it will be commit but well I will use it in my
corner anyway.

By the way I although wrote a new zen_date_raw function (in an
extra_configures file) that accept any date format not only English one but
I don't think I am going to submit this for now.

 

Thanks for your welcome.

 

Pierre

lat9:
Since the initial change was a collaboration between @proseLA and myself, I'd like both of us to review prior to merging.

proseLA (commented):
i think if you read the comments, @piloujp understands that this PR will not get merged.  i completely applaud this effort, and from a cursory look, i'm guessing the code is solid.

i would like to address a couple of his comments though:

> I think there is only one problem with actual code. When you have PHP under
8.1 it won't use datefmt_create() but Date() which can only display English
month names and week days. No error here but imagine how nice it looks on an
invoice.

> If you use dateobject and PHP >= 8.1 then with datefm_create international
dates are ok.  Why 8.1? Dateobject exist for a while, looking at PHP doc it
was implemented in PHP 5.3.

there was a determination made if using less than php 8.1, we would use the strftime function.  that function DOES allow for international dates, and it is done via the `setlocale` [as seen in the php documentation.](https://www.php.net/manual/en/function.strftime.php#:~:text=Examples%20%C2%B6,strftime()%20locale%20examples) 

that gets done here:
https://github.com/zencart/zencart/blob/db3986027440db69fb52677b94b589d1490907f5/includes/languages/lang.english.php#L8

and here:
https://github.com/zencart/zencart/blob/db3986027440db69fb52677b94b589d1490907f5/admin/includes/languages/lang.english.php#L9

if the strftime does not work, that means that the locales may not be installed on your system.

> My position is generally that rather than fixing old code until the end (deprecated) I
rather write a new one and fix it until it performs as well as the old one
and better. About zc_date I think the actual code is good even great but
**based on deprecated parameters that you have to convert to new ones. I'd
rather do the opposite** and it is why I submitted this new version. 

i agree with his sentiment here and expressed it at the time.  using the OLD strftime formats because it was deemed to be less painful, i think will cause more pain in the long run.

whether we want to create a new class, ie `$zcIntlDate` that accepts the new standard for date formats and slowly move over to that class and eliminate this one could certainly be a topic for conversation.

piloujp:
Well, looking at your comment I must admit my remarks on problem with version lower than 8.1 are irrelevant as strftime is used in this case without any problem. Although I am aware of the locale declaration and importance, it is not the problem. For any reason it seems I had already wiped out strfttime from my head... My bad excuse is that when trying to answer real time to your comments it is very late at night here and my brain might be bugging. Just embarrassing...

> i agree with his sentiment here and expressed it at the time. using the OLD strftime formats because it was deemed to be less painful, i think will cause more pain in the long run.

100% sure and this is my original motivation to rewrite zc_date class. Now that more and more people goes PHP 8.1 and higher, each time it is used there are conversion from strftime parameters format to IntlDateFormatter::create one, wasting server resources. Not to mention when you create new plugin or code you have to use old deprecated parameters.

Jus for information I dropped the rewritten zcDate.php file on a live server (Win10, Apache 2.54, PHP 8.1.13, ZC1.5.8) few weeks ago without touching any other files and it has been flowless until now.
In case discussion opens again on the subject you know there is some working code (in my environment) and it is backward compatible.

PR #5477, opened 2022-12-15
Expand installation to support PHP 8.2 and properly report

### LINKED ISSUE #5473: V158 can't install with php 8.2
Uploaded to my server the github zip file: zencart-zencart-v1.5.8-161-g7eb7953.zip
Obtained the file by using: https:// github.com/zencart/zencart/zipball/v158/

Unzipped the files. 
Attempted to access the site, was correctly advised that needed to install and or correct any of other possible issues as to why am seeing the screen. 

Accessed zc_install (without using index.php) and was taken to the system check. 

So far so good. To right identifies that am using zen Cart 1.5.8, but then i get or have:
````
Some problems that need fixing before we continue
Incorrect PHP Version.
The PHP version you are using (8.2.0) is not suitable.

This version of Zen Cart® is compatible with PHP versions 7.3 to 8.0, although 7.4.x or higher is recommended.
Check the [www.zen-cart.com](https://www.zen-cart.com/) website for the latest version of Zen Cart®.

For more information about version compatibility and server requirements see [Zen Cart Server Requirements](https://docs.zen-cart.com/user/first_steps/server_requirements/)
````

There is more text below that, but why bother when coming to that first emboldened error message.

Php version enabled on the server was: 8.2 FastCGI, though that shouldn't matter as version was properly picked up and identified. Really any of the other "normal" characteristics for reporting a problem appear non-applicable. Comes down to the fileset, the limits identified for the version and the text that is supposed to be displayed in reference to the installation. 

Initial file inspection identified that the php version for installation is capped at 8.1.19 and that the above text is mostly hard coded but obviously not updated in initial distribution and not yet corrected. 

Note that the online docs to which everything is now directed says that it should be fully functional with php 8.2 and doesn't advise that installation must be with a lower version and certainly not with 7.3.

mc12345678:
In initial release of Zen Cart 1.5.8, there have been multiple aspects that were not updated to directly support production operation. This is another case where the product is reported in: https://docs.zen-cart.com/user/first_steps/server_requirements/ is to support up to and include PHP 8.2, but the installation is prevented while operating in PHP 8.2 during installation.

Based on the text results and further testing of the underlying version comparison function `version_compare`, other changes were made to the text response for improved understanding.

Purposefully assigned the maximum php version to be 8.3 instead of 8.3.0 in the event that the PHP community releases the first such instance of 8.3 without using the sub-version designator.

Updated the recommended PHP version to 8.1 from 7.4.x. based on active support for 8.0 ending on 26 Nov 2022 per: https://www.php.net/supported-versions.php

Fixes #5473 (at least first issue identified while installing using php 7.2.)


scottcwilson:
After making this change were you able to successfully install under PHP 8.2?

mc12345678:
> After making this change were you able to successfully install under PHP 8.2?

Yes. 

And to ensure the changes "made sense" I had first tested multiple version formats against the expected data, I then purposefully restricted the code to one less php version to observe the messaging including the data in the log file.  Once the code restriction was returned to less than php 8.3, I was permitted to continue on with a fresh install with demo product. Any issues identified along the straight path of installation have been reported against this issue.  

Major change in messaging is that now reporting is the expectation that the software should be using the php version that is at our greater than what is reported as the minimum version, though the maximum is a little trickier. 

For maximum php version, because the ultimate highest subversion is not known or identified, it does not seem appropriate to compare against say 8.2.999999 to do the logic of verifying that the current version is at or below that value. So, the test as it has been is to verify the current version of PHP is less than the version at which it is known or expected to encounter an issue (8.3). But, wait. There's more. Then there is messaging that is to help the end user. Well instead of saying that should be using a version less than 8.3, the reporting has basically been intended to say use a version that is or is less than 8.2.x. of course there is no description about what 'x' represents, but I don't see a need to further explain that character. 

The reason I bring that up though is that moving forwards, ideally should only have to change the actual upper version number in two places, the version number that is one higher in one place, the lower version number in two places (one of which is the same define as for the upper value), and then only the recommended version.

scottcwilson:
Thanks for doing this; I have updated the build documentation so this will be a reviewed item moving forward.

PR #5480, opened 2022-12-15
Notify when <head> tag's been rendered ...

lat9:
... some external observers need to insert their scripts just after that tag's rendering.

Note that this is a template-related change.


scottcwilson:
Will add to docs on next push.

Issue #5450, opened 2022-12-07
Attribute Controller, Copy Attributes to Whole Category: debug Undefined array key "products_update_id


torvista:
php 8.2.0RC7
Select a product with attributes.
Additional Actions->Copy Options to a whole category

--> PHP Warning: Undefined array key "products_update_id" in ...\attributes_controller.php on line 829.

This field is not required for this option.

torvista:
there is another debug for <?php echo zen_draw_hidden_field('copy_attributes', $_GET['copy_attributes']); ?> which should not be there either, same as https://github.com/zencart/zencart/pull/5428.

scottcwilson:
Would you please retest and PR now that a bunch of fixes have been made in this area; not sure what remains.  Thanks.

PR #5481, opened 2022-12-15
Attribute Controller->Copy Attributes to a Category, fixes

torvista:
fixes: https://github.com/zencart/zencart/issues/5450

Note that for the redirect, leaving product_filter empty lets the page set it onload. 

For the moment I've no more bugs/fixes pending for this file.

### DIFF
--- a/admin/attributes_controller.php
+++ b/admin/attributes_controller.php
@@ -575,6 +575,10 @@ if (!empty($action)) {
       }
       $_GET['action'] = '';
       $products_filter = $_POST['products_filter'];
+      if ($_POST['categories_update_id'] !== '') {
+          $_POST['current_category_id'] = $_POST['categories_update_id'];
+          $products_filter = '';
+      }
       zen_redirect(zen_href_link(FILENAME_ATTRIBUTES_CONTROLLER, 'products_filter=' . $products_filter . '&current_category_id=' . $_POST['current_category_id']));
       break;
   }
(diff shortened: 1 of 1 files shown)


scottcwilson:
Please squash

torvista:
ok

mc12345678:
What's the reason behind continuing to send the parameter when there is no value assigned to it? Shouldn't the applicable `empty` parameter (have) be(en) omitted?

torvista:
`What's the reason behind continuing to send the parameter when there is no value assigned to it?`

Nothing more than a bit of making-minimum-changes-so-there's-less-complaining. But, yes the product_filter parameter should be checked for empty and omitted if so.

>  Shouldn't the applicable empty parameter (have) be(en) omitted?

Not sure what you mean/same as above?

mc12345678:
> > What's the reason behind continuing to send the parameter when there is no value assigned to it?
> 
> Nothing more than a bit of making-minimum-changes-so-there's-less-complaining. But, yes the product_filter parameter should be checked for empty and omitted if so.

Unfortunately, everyone complains for different reasons, I prefer if going to do it, try to fix it and the issues the fix causes. At the same time there are those changes to make changes situations. I guess it's not terrible for the parameter to be present, though it could lead one to take action having insufficient code. E.g. verify only the existence of the parameter, not whether it empty or not. That's more likely to cause a logic error than an execution error. 

> 
> > Shouldn't the applicable empty parameter (have) be(en) omitted?
> 
> Not sure what you mean/same as above?

Yeah, basically same as the first sentence. In the second sentence I came out and asked the question in a way to express what I thought was the way for it to go. 

Then after reviewing the changes made since I last uploaded 1.5.8 I got a little punchy with expressing verb tense. Read one way all is in the past and read the other way, in the present. Both passive. 

Anyways, trying to find different and/or expressive ways to write. Feel like I'm becoming too repetitive, asking why. Why is this not standardized, why is/isn't effort being applied here, why are the house rules not being followed? For those that know these, why are they not declaring the reason for non adherence? For those not familiar, honest answers are awesome. Sort of gives that unwritten approval for someone to pick up from there and make the next improvement. 

In this case though, I can't say I've double checked impact of the change (removal of data). I was mostly looking at the change from a "how might I use that data if I needed to" and "did I write or use something that is impacted by this change and what needs to change for it to offer continued operation".

I mean I have yet to understand the negative impact of that data remaining as it was, but if the operation remains functional at least there is some improved clarity.

PR #5485, opened 2022-12-16
modules/attributes: minor code tightening

torvista:
no functional changes


proseLA:
and again, why necessary?

i look at #5490, and since the constant is a string, why did you not do a strict comparison when fixing the bug there? 

it seems that if you want to do no functional changes on something, why not start on your PR that just got merged that has functional changes.  

these commits seem superfluous (and create unnecessary work for people upgrading) if the functionality is the same.

torvista:
After wasting more time with half a dozen attempts to write something respectful that I have not written before, why bother.

PR #5487, opened 2022-12-16
admin, stylesheet: correct rule

torvista:


dbltoe:
Not such a big deal in the admin since the world is not going to see it but, the color for a button is normally a smaller text and should be of higher contrast as a result.

The background color used (#337ab7) only has a 4.55:1 ratio which would be okay if a larger button with larger text.

#0a57a4 as a background color would assure that the button pass A level accessibility with a ratio of 7.21:1.

New favorite tool at https://contrast-ratio.com.  It will even check with rgba's and hex 8's.  Should I PR to replace this private app with the institutional WebAIM checker in the docs?  That particular area needs more "help".  Perhaps add the individual's site as an optional source?

torvista:
I'm afraid I did not employ many neurons other than those necessary to correct the missing hyphen prefix.
Whether the thing should be there at all is for others to argue!

dbltoe:
not a problem here.  Mainly directed at Scott for Doc thoughts.  Should have done it elsewhere.

scottcwilson:
John your PR is welcome.  Thanks to both of you.

PR #5489, opened 2022-12-16
admin, css: remove redundant px

torvista:


proseLA:
please tell me what we are gaining here?

the unit designation is completely optional after a 0, and other than creating diffs for people doing upgrades, is this PR necessary/warranted?

torvista:
> necessary/warranted

No. But a lot of PRs, could have been argued about similarly.

In an ideal world, running an IDE inspection on a bunch of files would reveal a few things that should be examined, not minor stuff like this. So, clear this minor chaff or reduce the inspection detail?
Probably should find the setting to not flag it, but this only needs doing once.
Should we be behoven to default settings on an IDE?  No. But, path of least resistance is doing this. For me at least.

scottcwilson:
> Probably should find the setting to not flag it ...

Yep.

dbltoe:
I'm not a fan of searching for such but, if you are touching a file it makes sense to do corrections that help in the long run.  I don't know of any checker that doesn't flag this as a warning.

We similarly avoided the text javascript and now HTML 5 is really harping on it.

Lastly, if fixing the unit stops a real problem from being lost in a bunch of unit warnings, just that would make it worthwhile to me.  At some point, checkers are going to balk saying that 0 is a null and 0px is not.

proseLA:
> > necessary/warranted
> 
> No. But a lot of PRs, could have been argued about similarly.
> 
i am not commenting about those PRs.  i am commenting about this one.  

i only have so many hours in the day.

> In an ideal world, running an IDE inspection on a bunch of files would reveal a few things that should be examined, not minor stuff like this. So, clear this minor chaff or reduce the inspection detail? Probably should find the setting to not flag it, but this only needs doing once. Should we be behoven to default settings on an IDE? No. But, path of least resistance is doing this. For me at least.

who's idea of an ideal world?  is [this your idea](https://docs.zen-cart.com/dev/contributing/coding_standards/) of an ideal world?

your IDE inspections are NOT the standard; so please stop with that argument.

it has been well established that there is NOT the wherewithal to commit to a coding standard on this repo.  i get the sense that scott thinks it is a waste of time (for the most part).

and this change is NOT a standard.  despite what your IDE says.  there is no performance increase, minifiers remove the px; and until you can make an argument, other than my infinitely configurable IDE says so, i would encourage you to focus on things that would actually make OTHERS lives easier.  which IMO, this PR (and numerous other ones like it), do not.

feel free to have the last word.

torvista:
You are right, I need to focus on making my own life easier.

scottcwilson:
> it has been well established that there is NOT the wherewithal to commit to a coding standard on this repo. i get the sense that scott thinks it is a waste of time (for the most part).

Scott (and the other core committers) have limited amounts of time to do the uncompensated work of maintaining Zen Cart, so we encourage people to focus on *fixing bugs*.  Each PR costs me time and energy, and I only have so much of those things.

PR #5490, opened 2022-12-16
admin, orders: fix for adding string to number

torvista:
whitespace too.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -12,7 +12,7 @@ if (isset($module)) {
   unset($module);
 }
 
-// Override instructions in: 
+// Override instructions in:
 // https://docs.zen-cart.com/user/admin/site_specific_overrides/
 if (!isset($quick_view_popover_enabled)) $quick_view_popover_enabled = false;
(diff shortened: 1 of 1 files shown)


scottcwilson:
Can you be more specific about what this fixes?

torvista:
This should be a sum.
zen_date_diff is a float, DOWNLOAD_MAX_DAYS is a string.

mc12345678:
FYI, have a similar issue with: https://github.com/zencart/zencart/blob/b0b560f603f3b04d3efc3409a36805b7bac7c6aa/admin/orders.php#L199

Database result comes back as a string, date difference is numeric. 

Considering what is addressed through the concern of this PR, similar code uses should be evaluated. I found this by looking at how `products_attributes_maxdays` is used. I did the search in Github which means the results are limited and there may be other places to consider. A search in the Developers Tool Kit would be more exhaustive.

mc12345678:
What's the reasoning behind using `intval` instead of casting with `(int)`?

torvista:
After reading a bit more, none.

PR #5496, opened 2022-12-19
Fix strict comparison of string with int always causing re-fetch of p…

### LINKED ISSUE #5495: string === int comparison causes repeated fetch of product fields
**Describe the bug**
Just a minor issue I ran into.

`product_info/main_template_vars.php` performs this check and fetch of product fields:
```
    if (!isset($product_info->EOF, $product_info->fields['products_id'], $product_info->fields['products_status']) || $product_info->fields['products_id'] !== (int)$_GET['products_id']) {
        $product_info = zen_get_product_details($_GET['products_id']);
    }

```
Since `$product_info->fields['products_id']` is a string representation of an integer from the db, the strict comparison with `(int)$_GET['products_id']` fails.  So, the call to `zen_get_product_details()` that was done in product/header_php.php is repeated for every product page request.

Looks like this was introduced in 46ae2eed 

**Version**
Zen Cart version: 1.5.8 / current master
PHP version: 8.1
What older/newer versions were also tested? I looked at the Line History view in Git and it seems current.

**To Reproduce**
1. Set a breakpoint on product/main_tempate_vars.php line 17 where the comparison is made.
2. Visit any product page in the Demo dataset e.g. my local URL https://zc158.orig/index.php?main_page=product_info&cPath=1_4&products_id=1
3. Observe how the if block is entered and zen_get_product_details is called despite it being fully populated at that time.

**Expected behavior**
The `if` is there to prevent repeated call to this function.

**Screenshots**
Screenshot from 2022-12-19 09-22-13

neekfenwick:
…roduct fields.

Fixes #5495

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -42,7 +42,7 @@ function zen_product_set_header_response($product_id, $product_info = null)
         $product_info = zen_get_product_details($product_id);
     }
     // make sure it's for the current product
-    if (!isset($product_info->fields['products_id'], $product_info->fields['products_status']) || $product_info->fields['products_id'] !== $product_id) {
+    if (!isset($product_info->fields['products_id'], $product_info->fields['products_status']) || (int)$product_info->fields['products_id'] !== $product_id) {
         $product_info = zen_get_product_details($product_id);
     }
 
--- a/includes/modules/pages/document_general_info/main_template_vars.php
+++ b/includes/modules/pages/document_general_info/main_template_vars.php
@@ -14,7 +14,7 @@
   // This should be first line of the script:
   $zco_notifier->notify('NOTIFY_MAIN_TEMPLATE_VARS_START_DOCUMENT_GENERAL_INFO');
 
-  if (!isset($product_info->EOF, $product_info->fields['products_id'], $product_info->fields['products_status']) || $product_info->fields['products_id'] !== (int)$_GET['products_id']) {
+  if (!isset($product_info->EOF, $product_info->fields['products_id'], $product_info->fields['products_status']) || (int)$product_info->fields['products_id'] !== (int)$_GET['products_id']) {
     $product_info = zen_get_product_details($_GET['products_id']);
   }
 
--- a/includes/modules/pages/document_product_info/main_template_vars.php
+++ b/includes/modules/pages/document_product_info/main_template_vars.php
@@ -14,7 +14,7 @@
   // This should be first line of the script:
   $zco_notifier->notify('NOTIFY_MAIN_TEMPLATE_VARS_START_DOCUMENT_PRODUCT_INFO');
 
-  if (!isset($product_info->EOF, $product_info->fields['products_id'], $product_info->fields['products_status']) || $product_info->fields['products_id'] !== (int)$_GET['products_id']) {
+  if (!isset($product_info->EOF, $product_info->fields['products_id'], $product_info->fields['products_status']) || (int)$product_info->fields['products_id'] !== (int)$_GET['products_id']) {
     $product_info = zen_get_product_details($_GET['products_id']);
   }
 
--- a/includes/modules/pages/product_free_shipping_info/main_template_vars.php
+++ b/includes/modules/pages/product_free_shipping_info/main_template_vars.php
@@ -14,7 +14,7 @@
   // This should be first line of the script:
   $zco_notifier->notify('NOTIFY_MAIN_TEMPLATE_VARS_START_PRODUCT_FREE_SHIPPING_INFO');
 
-  if (!isset($product_info->EOF, $product_info->fields['products_id'], $product_info->fields['products_status']) || $product_info->fields['products_id'] !== (int)$_GET['products_id']) {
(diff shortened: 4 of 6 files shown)


neekfenwick:
Same error found in `zen_product_set_header_response`, see new added commit 6b79c6d.

Screenshot from 2022-12-19 09-40-24

includes/functions/functions_products.php
```diff
@@ -42,7 +42,7 @@ function zen_product_set_header_response($product_id, $product_info = null)
         $product_info = zen_get_product_details($product_id);
     }
     // make sure it's for the current product
-    if (!isset($product_info->fields['products_id'], $product_info->fields['products_status']) || $product_info->fields['products_id'] !== $product_id) {
+    if (!isset($product_info->fields['products_id'], $product_info->fields['products_status']) || (int)$product_info->fields['products_id'] !== $product_id) {
```
mc12345678, line 45: Both sides should cast to `int`, not just the left.

neekfenwick: The right is already int, as shown in the Debug Console output I generated in the screenshot.  I could dig back into it but is it really necessary? Are you worried about edge cases where it's not an int?

neekfenwick: As a side note, the way an assignment is done as the parameter is passed into `zen_get_product_details` is horrible, but to help with the type casting situation, we could use type hinting as described in https://www.php.net/manual/en/language.types.declarations.php so the `zen_get_product_details` fn signature would become:

```function zen_get_product_details(int $product_id, int $language_id = null)
```

with both params hinted as int.  This would break backward compatibility below PHP 7, but as 7 is already EOL, do we really care? I'd be happy to add to my PR to do this, just not sure what the Zen Cart community position on type hinting is.  I don't see much around, but is it simply through lack of implementation rather than other overriding philosophy?

neekfenwick: I see from https://docs.zen-cart.com/user/first_steps/server_requirements/ that PHP 7 has been required since Zen Cart 1.5.5, so I think we're safe in that regard :)

proseLA: > As a side note, the way an assignment is done as the parameter is passed into `zen_get_product_details` is horrible, but to help with the type casting situation, we could use type hinting as described in https://www.php.net/manual/en/language.types.declarations.php so the `zen_get_product_details` fn signature would become:
> 
> `function zen_get_product_details(int $product_id, int $language_id = null)`
> 
> with both params hinted as int. This would break backward compatibility below PHP 7, but as 7 is already EOL, do we really care? I'd be happy to add to my PR to do this, just not sure what the Zen Cart community position on type hinting is. I don't see much around, but is it simply through lack of implementation rather than other overriding philosophy?

big fan of type declarations in signatures.

neekfenwick: I thought PHP 8 would barf over wrong datatypes being passed, but a simple test shows you can pass a string into a function with param hinted as int no problem.  It does cast the string to int on the way in, so there is benefit in that the param in the function is guaranteed to be of type int, so strict === comparison would work.

So it seems typehinting the params to core Zen functions would not cause any breakage where things like `$_GET['products_id']` is passed in without regard for type, and would actually help where functions assume it to be int.

We'd have to watch out where the products_id being passed has attribute data hashed into it, I find that passing a string like `123:abcde` into a function throws a `TypeError` with message `"Argument #1 ($products_id) must be of type int, string given"`, so it doesn't cast to `int` as in my last paragraph, and doesn't auto-cast in the same way as a specific `(int)$foo` would.    In this case, because the $_GET['products_id'] is cast to int before being assigned to `$products_id_current` and being passed into `zen_get_product_details`, my testing shows no problem with adding type hinting to the function definitions, even after adding a product like Demo data's Matrox G200 with a few options selected and then clicking the product in the cart to 'edit' it, so $_GET['products_id'] is `"1:83a1db1371ebe7b060dd20664941eca8"` but the int `1` is what's passed around.

However, as Ripley said to Newt, once you start you have to do the whole thing, there are 45 functions in `functions_products.php` alone, and all that noise doesn't belong a PR solving this Issue. Feels like it needs a new Issue/PR, eh?

mc12345678: You referenced php 8. To what level have you tried your experiments? Php 8.2 has added some additional quirks to the type casting madness.

neekfenwick: Fair point, I was testing on PHP 8.1, but since you ask, I've downloaded 8.2 and everything seems to behave exactly the same. https://gist.github.com/neekfenwick/b1d40be1f0cd31a62d09ce3439c8ed34 is the test script I was using for basic tests, and the Zen Cart demo site also behaves the same.  I read up on 8.2 at https://freek.dev/2382-introducing-php-82-all-new-features-and-changes and cannot see any breaking changes relating to type casting.  Seems good.

scottcwilson:
Thanks for tracking this down @neekfenwick.

neekfenwick:
> Thanks for tracking this down @neekfenwick.

Thank you too, no worries :)

Issue #5417, opened 2022-11-25
DB Changes needed 


scottcwilson:
These are things that will need to be added to the database create/update scripts for the next numbered  release.
SOME have been added for 1.5.8a but need to be done again for people who installed 1.5.8. 

~~Fix tables for sites using prefixes:~~
~~https://github.com/zencart/zencart/pull/5334~~
People got this from the upgrade if they upgraded post 1.5.8a, or they did it by hand if they upgraded before.
Adding this again will just cause the upgrade to produce errors.

~~Country Turkey rename: https://github.com/zencart/zencart/pull/5416~~

lat9:
Also #5601

piloujp:
Some zones names for Italia, Swiss and Austria have some mistakes or just changed. 
Here is the list:

Austria (14)
Zone id 103, zone name 'Voralberg' should be 'Vorarlberg'

Italy(105)
Zone id 211 'Carbonia-Iglesias' should be deleted as it does not exist anymore.
Zone id 245 'Medio Campidano' should be deleted as it does not exist anymore.
Zone id 253 'Ogliastra' should be deleted as it does not exist anymore.
Zone id 254 'Olbia-Tempio' should be deleted as it does not exist anymore.

A new zone 'Sud Sardegna' with code SU should be added. Some states were grouped into this new one.

Some other have differences in name spelling more or less important. I am not sure all need to be update but I made a list anyway.
Zone id 193, zone name 'Aosta' should be 'Valle D'Aosta' which in SQL query should be 'Valle D\'Aosta'
Zone id 199, zone name 'Barletta Andria Trani' should be 'Barletta-Andria-Trani'
Zone id 226, zone name 'Forlì Cesena' should be 'Forlì-Cesena'
Zone id 233, zone name 'Aquila' should be 'L'Aquila' or 'L\'Aquila'
Zone id 243, zone name 'Massa Carrara' should be 'Massa-Carrara'
Zone id 261, zone name 'Pesaro Urbino' should be 'Pesaro E Urbino'
Zone id 294, zone name 'Verbania' should be 'Verbano-Cusio-Ossola'

Swiss(204)
All 'cantons' (states) names are in German but intenational designation uses English or main local language (German, French or Italian).
Zone id 110, zone name 'Freiburg' should be 'Fribourg'
Zone id 111, zone name 'Genf' should be 'Genève'
Zone id 113, zone name 'Graubnden' with code 'JU' should be 'Graubünden' with code 'GR'
[...]

PR #5497, opened 2022-12-19
Split Invoice And Packing Image Decision

brittainmark:
Allow the packing Slip and the invoice to have different options for displaying images.
This allows images to be displayed on the invoice but not the packing slip ang vice versa.

### DIFF
--- a/admin/includes/extra_datafiles/dist-site_specific_admin_overrides.php
+++ b/admin/includes/extra_datafiles/dist-site_specific_admin_overrides.php
@@ -17,9 +17,13 @@
 // admin/invoice.php and admin/packingslip.php 
 // to determine whether attribute images should be displayed. 
 //
-// true ...... attribute images are shown on invoice and packingslip (the default) 
-// false ..... attribute images are not shown on invoice and packingslip 
+// $show_attrib_images   $show_attrib_images_pack
+// true or unset         true or unset            ...... attribute images are shown on invoice and packingslip (the default) 
+// false                 false or unset           ...... attribute images are NOT shown on invoice or packingslip 
+// false                 true                     ...... attribute images are NOT shown on invoice but are shown packingslip 
+// true or unset         false                    ...... attribute images are shown on invoice but are NOT shown packingslip
 // $show_attrib_images = false;
+// $show_attrib_images_pack = false;
 
 // Set the width of the attribute image used in packingslip and invoice.
 // the default is 25. if $show_attrinb_images = false is set above then setting this value will have no effect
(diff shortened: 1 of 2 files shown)


scottcwilson:
I like this change, but I wonder if this is in the class of changes (like database changes) that should only be done on a major release, and thus deferred under #5417?  Letting @lat9 decide.   The way it was done is backwards compatible (nice!) so this extra precaution may not be warranted.

admin/includes/extra_datafiles/dist-site_specific_admin_overrides.php
```diff
@@ -17,9 +17,13 @@
 // admin/invoice.php and admin/packingslip.php 
 // to determine whether attribute images should be displayed. 
 //
-// true ...... attribute images are shown on invoice and packingslip (the default) 
-// false ..... attribute images are not shown on invoice and packingslip 
+// $show_attrib_images   $show_attrib_images_pack
+// true or unset         true or unset            ...... attribute images are shown on invoice and packingslip (the default) 
+// false                 false of unset           ...... attribute images are NOT shown on invoice or packingslip
```
lat9, line 22: Change "false _of_ unset" to "false **or** unset".

lat9 (changes requested):
Me, again, being picky.  Great idea on the update!

admin/includes/extra_datafiles/dist-site_specific_admin_overrides.php
brittainmark: Good spot not picky!

admin/includes/extra_datafiles/dist-site_specific_admin_overrides.php
```diff
@@ -29,10 +33,14 @@
 // The flag show_product_images is used by the files 
 // admin/invoice.php and admin/packingslip.php 
 // to determine whether product images should be displayed. 
-//
-// true ...... product images are shown on invoice and packingslip (the default) 
-// false ..... product images are not shown on invoice and packingslip 
+// 
+// $show_product_images  $show_product_images_pack
+// true or unset         true or unset            ...... product images are shown on invoice and packingslip (the default) 
+// false                 false of unset           ...... product images are NOT shown on invoice or packingslip
```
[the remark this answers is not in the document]
brittainmark: Good spot not picky!

lat9 (approved):

scottcwilson:
Thanks @brittainmark.  Added to fixes for next patch release.

mc12345678:
Why wouldn't detection of the page being loaded be used to provide this "difference" instead of adding another configuration switch? Oh wait, did I use that term?

E.g.:
if (isset($_GET['cmd']) && $_GET['cmd'] === FILENAME_ORDERS_PACKINGSLIP) {
    // do packingslip type action.
}
if (isset($_GET['cmd']) && $_GET['cmd'] === FILENAME_ORDERS_INVOICE) {
    // do invoice type action.
}
and as necessary combine/mix the two.

Also, note that the instructions referenced in the changed file have not been updated to reflect the addition of this extra configuration switch. See line 11 of the below:
https://github.com/zencart/zencart/blob/c8edea1529a01007e6f0b5a5692f929d083a1062/admin/packingslip.php#L10-L15

scottcwilson:
The doc issue has been added to the 1.5.8a doc tasks list. Thanks for pointing this out.

scottcwilson:
> Why wouldn't detection of the page being loaded be used to provide this "difference" ... ? 

The spirit of this file is to make it very easy even for non-technical people to make these tweaks.  Certainly you can add arbitrary logic but I'd rather the common use cases be super simple.

brittainmark:
@mc12345678 Thanks for spotting this, I forgot. Have sent an update to the documentation

PR #5498, opened 2022-12-20
Remove redundant assignments

scottcwilson:
The three removed lines are redundant; the setting is done further down in each `case` block.

### DIFF
--- a/includes/modules/shipping/zones.php
+++ b/includes/modules/shipping/zones.php
@@ -236,8 +236,6 @@ class zones
         switch (MODULE_SHIPPING_ZONES_METHOD) {
           case (MODULE_SHIPPING_ZONES_METHOD == 'Weight'):
             if (round($shipping_weight, 9) <= $zones_table[$i]) {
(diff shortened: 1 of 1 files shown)


proseLA:
a small refactor and a good one. 

:+1:

mc12345678:
Why not get rid of the `else` and use the previously set `$shipping` value?

E.g.:
````diff
-              if (strstr($zones_table[$i + 1], '%')) {
-                $shipping = ($zones_table[$i + 1] / 100) * $order_total_amount;
+              if (strstr($shipping, '%')) {
+                $shipping = ($shipping / 100) * $order_total_amount;
-              } else {
-                $shipping = $zones_table[$i + 1];
              }
````

A couple things accomplished: 1) Reassignment to the original value doesn't happen, 2) Modifications between the two existing assignments do not need unnecessary reevaluation. 3) Use/reuse of a single variable where the original value (`$zones_table[$i + 1]`) remains accessible.

PR #5503, opened 2022-12-22
Retain disabled product status after upgrade

### LINKED ISSUE #5502: Disabled product unexpectedly enabled
**Describe the bug**
After updating a database, disabled product that have a `products_available_date` set to `'0001-01-01'` (one of two typical values when "resetting" or clearing `products_available_date`) will become active when the `Stock` configuration setting of `Enable disabled product by available date` is set to `Automatic`.
This unexpectedly causes intentionally hidden/disabled product to become active without notification and without owner direction/choice.

**To Reproduce**
(without going through a database upgrade to potentially randomly observe the occurrence):
1. Install Zen Cart 1.5.8 with demo product.
2. Temporarily change the Configuration->Stock setting `Enable disabled product by available date` to `Manual` (Default on a new install is `Automatic`. An upgrade maintains the existing setting from the upgraded database or, if not present, defaults to `Manual`).
3. For good measure go ahead and restore `products_date_available` for products_id 34 to its demo value:
````
UPDATE products SET products_date_available = '0001-01-01' WHERE products_id = 34;
````
4. Verify the accessibility of the product by navigating the site. (Visit product with products_id of 34: index.php?main_page=product_info&products_id=34)
5. In the admin, disable product with products_id 34.
6. Verify the inaccessibility of product with products_id 34: index.php?main_page=product_info&products_id=34 (session data may require navigating away and then another attempt to access. Refreshing/reloading the page may not newly pull applicable data).
7. Restore the Configuration->Stock setting `Enable disabled product by available date` back to `Automatic`.
8. Navigate the site and again attempt to access that same product to discover that it is now enabled.
9. Review of the products table for products_id 34 will identify that the products_date_available value has been changed to `null` instead of remaining at `'0001-01-01'` or `'0001-01-01 00:00:00'`.

**Expected Behavior**
Root expectation: Leave the product disabled when `products_date_available` is `'0001-01-01'` with the following considerations:
A database is to be upgraded without taking random action on the product catalog such as enabling existent product purposefully disabled. Also, during navigation of the site, data associated with the product that have no impact on continued operation should remain untouched. Until Zen Cart 1.5.8, a `products_date_available` value of `'0001-01-01 00:00:00'` has been acceptable and further has been a means of identifying previous existence of a `products_date_available` as previous operation would convert a foregone date to the Zen Cart "empty date". Additionally, it seems that the upgrade process (`zc_install/sql/install/zero_dates_cleanup.php`) shouldn't affect the existing `'0001-01-01'` dates in `products_date_available` at this time (haven't identified a good reason to yet force that level of conversion).

**Cause(s) of the issue**
Comparison of the applicable Zen Cart 1.5.7 and 1.5.8 code identifies that Zen Cart 1.5.8 incorrectly includes the date value `'0001-01-01'` for evaluating that a product is to be enabled (The difference: changed `!=` to `>=` instead of either keeping `!=` or changing to `>`. I haven't seen anything that suggests that `!=` is incorrect or causes an issue or that omission of the time portion is an issue as initial testing supported `!=`). 

1.5.7:
https://github.com/zencart/zencart/blob/742b837868a2b128df96b6635c413c8187972679/includes/functions/functions_products.php#L120-L121

1.5.8:
https://github.com/zencart/zencart/blob/c8edea1529a01007e6f0b5a5692f929d083a1062/includes/functions/functions_products.php#L132-L133

(Recent) core code that could have caused `products_date_available` to be `'0001-01-01'` prior to upgrade to Zen Cart 1.5.8: use of admin/products_price_manager.php to modify a product in Zen Cart versions before 1.5.8.

ZC 1.5.7:
https://github.com/zencart/zencart/blob/742b837868a2b128df96b6635c413c8187972679/admin/products_price_manager.php#L109
ZC 1.5.0:
https://github.com/zencart/zencart/blob/85b3968832e20fbbb19866d254173884f2f1ad15/admin/products_price_manager.php#L135

ZC 1.5.8 (uses new `zen_prepare_date` function with default to `null` instead of `'0001-01-01'`):
https://github.com/zencart/zencart/blob/c8edea1529a01007e6f0b5a5692f929d083a1062/admin/products_price_manager.php#L106

### PR DESCRIPTION
**Pull request by mc12345678**
Prevents activating disabled product that have a historical `products_date_available` that is equal to `'0001-01-01'`. Initial testing of this (nearly two years ago) had indicated that the same result is obtained if the actual date/time is `'0001-01-01 00:00:00'`.

Fixes #5502

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -129,7 +129,7 @@ function zen_enable_disabled_upcoming($datetime = null)
             FROM " . TABLE_PRODUCTS . "
             WHERE products_status = 0
             AND products_date_available <= " . $zc_disabled_upcoming_date . "
-            AND products_date_available >= '0001-01-01'
+            AND products_date_available != '0001-01-01'
             AND products_date_available IS NOT NULL
             ";
 

### CODE REVIEW

File: includes/functions/functions_products.php [hunk 1]
```diff
@@ -129,7 +129,7 @@ function zen_enable_disabled_upcoming($datetime = null)
             FROM " . TABLE_PRODUCTS . "
             WHERE products_status = 0
             AND products_date_available <= " . $zc_disabled_upcoming_date . "
-            AND products_date_available >= '0001-01-01'
+            AND products_date_available != '0001-01-01'
```
scottcwilson, line 132: Should there be a strictly greater than condition here instead?

scottcwilson: (I think this is OK as is, just making sure.)

scottcwilson:
Related feature doc: https://docs.zen-cart.com/user/products/upcoming_products/

PR #5508, opened 2022-12-27
Don't try to redefine META_TAG_* if previously defined, fixes #5449.

### LINKED ISSUE #5449: Custom page with custom META_TAG_TITLE causes 'Constant META_TAG_TITLE already defined' Warning
**Describe the bug**
includes/modules/meta_tags.php allows you to define a constant named after your custom page, which is used to define META_TAG_TITLE, but later again defines META_TAG_TITLE, causing a duplicate definition Warning to be generated.

**Version**
Zen Cart version: 1.5.8
PHP version: 8.1
MySQL version: n/a
Plugins that may be related: n/a
What older/newer versions were also tested? n/a

**To Reproduce**
1. Create a includes/modules/pages/custom_page/header_php.php which contains: `define('META_TAG_TITLE_CUSTOM_PAGE', 'This is a custom meta title tag');`
2. Create a includes/templates/template_default/tempaltes/tpl_custom_page_default.php which contains Hello World text.
3. Navigate to the page: https://my-site.com/index.php?main_page=custom_page
4. See Warning generated:
`PHP Warning: Constant META_TAG_TITLE already defined in /home/neek/workspace/zen-cart/zencart/includes/modules/meta_tags.php on line 344.`

**Expected behavior**
Constants should not be defined twice.

**Screenshots**
Screenshot from 2022-12-07 10-37-51

**Additional context**
I'm not sure how many people create custom pages with custom meta tags like this.

meta_tags.php line 40 sets META_TAG_TITLE:
`  if (defined('META_TAG_TITLE_' . strtoupper($metatag_page_name))) define('META_TAG_TITLE', constant('META_TAG_TITLE_' . strtoupper($metatag_page_name)));`
then the `default` case in the switch, hit because this is a custom page not matched by the other case blocks, sets it again on line 344:
`      define('META_TAG_TITLE', (defined('NAVBAR_TITLE') ? NAVBAR_TITLE . PRIMARY_SECTION : '') . TITLE . TAGLINE);`

Maybe just guard those lines around 344 with `defined()` checks?

neekfenwick:
For example, by META_TAG_TITLE_PAGENAME in a custom page called 'pagename'.

This works for me with my example `tpl_custom_page.php` example described in Issue #5449 .  Warnings are no longer emitted, and my customised meta tags are used.  It is a little clumsy to test each constant line by line, but I thought this would be less intrusive than introducing more formal branches of logic or control variables such as the existing `$meta_tags_over_ride`.  The logic in that module is already pretty twisty.

### DIFF
--- a/includes/modules/meta_tags.php
+++ b/includes/modules/meta_tags.php
@@ -341,9 +341,9 @@ switch ($metatag_page_name) {
     $metatags_keywords = KEYWORDS . METATAGS_DIVIDER . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '');
     $zco_notifier->notify('NOTIFY_MODULE_META_TAGS_UNSPECIFIEDPAGE', $current_page_base, $metatag_page_name, $meta_tags_over_ride, $metatags_title, $metatags_description, $metatags_keywords);
     if (false===$meta_tags_over_ride) {
-      define('META_TAG_TITLE', (defined('NAVBAR_TITLE') ? NAVBAR_TITLE . PRIMARY_SECTION : '') . TITLE . TAGLINE);
-      define('META_TAG_DESCRIPTION', TITLE . PRIMARY_SECTION . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '' ) . SECONDARY_SECTION . KEYWORDS);
-      define('META_TAG_KEYWORDS', KEYWORDS . METATAGS_DIVIDER . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '' ) );
+      !defined('META_TAG_TITLE') && define('META_TAG_TITLE', (defined('NAVBAR_TITLE') ? NAVBAR_TITLE . PRIMARY_SECTION : '') . TITLE . TAGLINE);
+      !defined('META_TAG_DESCRIPTION') && define('META_TAG_DESCRIPTION', TITLE . PRIMARY_SECTION . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '' ) . SECONDARY_SECTION . KEYWORDS);
+      !defined('META_TAG_KEYWORDS') && define('META_TAG_KEYWORDS', KEYWORDS . METATAGS_DIVIDER . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '' ) );
     }
 }
 
--- a/includes/modules/meta_tags.php
+++ b/includes/modules/meta_tags.php
@@ -23,7 +23,7 @@ if (strlen(SITE_TAGLINE) > 1) {
 }
 $review_on = "";
 $keywords_string_metatags = "";
-if (!defined('METATAGS_DIVIDER')) define('METATAGS_DIVIDER', ', ');
+zen_define_default('METATAGS_DIVIDER', ', ');
 
 // Get all top category names for use with web site keywords
 $sql = "SELECT cd.categories_name FROM " . TABLE_CATEGORIES . " c, " . TABLE_CATEGORIES_DESCRIPTION . " cd WHERE c.parent_id = 0 AND c.categories_id = cd.categories_id AND cd.language_id='" . (int)$_SESSION['languages_id'] . "' AND c.categories_status=1";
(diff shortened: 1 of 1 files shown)


mc12345678:
Ummm, the point was basically that if someone introduced code that negated the need to execute that area (such as might seem like you did in your template file), then an observer should change the value to exclude executing the lines of this pull request. Further, previous discussion on this type of thing has been that if going to define one, then should define all...

While the logic is sound working from left to right through the line, don't be surprised to be requested to break things out a little more. Effectively what has been added is an additional if statement around those defines that is executed on every page load to then define the constant if it is not already defined. And to perform that if test on each of three variables. For any type of future editing/upgrading, seems that an `if` statement should be on one line, with the define on the next and end with a closure for the `if` statement.

mc12345678:
Then, even with my previous statement, there's aspects about it that I still do not like. Yes, it's possible to cause either a notice or warning. I haven't a clue the current thought process. Are all such possible operations to be prevented from causing an issue no matter what is done or some amount of expectation to self mitigate,  or to what level/amount between? 

I mean elsewhere when insufficiently written code caused an issue, core code got unnecessarily changed. So, I wouldn't really be surprised by anything.

neekfenwick:
> Ummm, the point was basically that if someone introduced code that negated the need to execute that area (such as might seem like you did in your template file), then an observer should change the value to exclude executing the lines of this pull request.

The problem I'm fixing is that if you define one of the three main meta tag definitions named after a custom page, e.g. META_TAG_TITLE_CUSTOM_PAGE in my case of a page called 'custom_page', then the code in `modules/meta_tags.php` will define META_TAG_TITLE twice, causing an Warning to be raised, as shown in the Issue this fixes.  It trips itself up over the code around line 40, and line 334, which both define e.g. META_TAG_TITLE.  I don't see how an observer would help.

> Further, previous discussion on this type of thing has been that if going to define one, then should define all...

All should definitely end up defined, but you may override them individually, just one, two, etc, so the code needs to check each one independently.  You can't wrap a single 'if' around the setting of all three based on the definition of one of them.

> While the logic is sound working from left to right through the line, don't be surprised to be requested to break things out a little more. Effectively what has been added is an additional if statement around those defines that is executed on every page load to then define the constant if it is not already defined. And to perform that if test on each of three variables. For any type of future editing/upgrading, seems that an `if` statement should be on one line, with the define on the next and end with a closure for the `if` statement.

I agree the code is fairly terse, and looking again, I see there's already a coding pattern used around line 40 and 336 that does things more explicity, i.e. `if (defined('FOO')) define('BAR', 'BAZ');`.  It's missing brackets for the if block but is at least more readable than what I've done.

I've pushed an improved commit that changes this.

scottcwilson:
This is fine.  Please go ahead and add the braces to the lines you have changed just for readability.  
In your reading of the docs is there an update required to explain this behavior, or is it ok as is?

includes/modules/meta_tags.php
```diff
@@ -341,9 +341,9 @@
...
     $zco_notifier->notify('NOTIFY_MODULE_META_TAGS_UNSPECIFIEDPAGE', $current_page_base, $metatag_page_name, $meta_tags_over_ride, $metatags_title, $metatags_description, $metatags_keywords);
     if (false===$meta_tags_over_ride) {
-      !defined('META_TAG_TITLE') && define('META_TAG_TITLE', (defined('NAVBAR_TITLE') ? NAVBAR_TITLE . PRIMARY_SECTION : '') . TITLE . TAGLINE);
-      !defined('META_TAG_DESCRIPTION') && define('META_TAG_DESCRIPTION', TITLE . PRIMARY_SECTION . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '' ) . SECONDARY_SECTION . KEYWORDS);
-      !defined('META_TAG_KEYWORDS') && define('META_TAG_KEYWORDS', KEYWORDS . METATAGS_DIVIDER . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '' ) );
+      if (!defined('META_TAG_TITLE')) define('META_TAG_TITLE', (defined('NAVBAR_TITLE') ? NAVBAR_TITLE . PRIMARY_SECTION : '') . TITLE . TAGLINE);
+      if (!defined('META_TAG_DESCRIPTION')) define('META_TAG_DESCRIPTION', TITLE . PRIMARY_SECTION . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '' ) . SECONDARY_SECTION . KEYWORDS);
+      if (!defined('META_TAG_KEYWORDS')) define('META_TAG_KEYWORDS', KEYWORDS . METATAGS_DIVIDER . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '' ) );
```
lat9, line 346: There's a function called `zen_default_default`, introduced in zc158, that makes this even more readable:
```php
zen_define_default('META_TAG_KEYWORDS', KEYWORDS . METATAGS_DIVIDER . (defined('NAVBAR_TITLE') ? NAVBAR_TITLE : '' ) );
```

neekfenwick: Nice one @lat9, I have updated the PR.

neekfenwick:
> This is fine. Please go ahead and add the braces to the lines you have changed just for readability. In your reading of the docs is there an update required to explain this behavior, or is it ok as is?

I'm not super familiar with the docs and can't remember where in them I maybe bumped into this feature, many years ago.  I may have discovered it when reading the code.  Perhaps https://docs.zen-cart.com/user/ezpages/ezpages_meta_tags/ but it suggests a plugin.  https://docs.zen-cart.com/user/customizing/add_pages/ mentions adding custom pages but no meta tag details.  I could add some doc, maybe to that first page, `ezpages_meta_tags`?

mc12345678:
> > This is fine. Please go ahead and add the braces to the lines you have changed just for readability. In your reading of the docs is there an update required to explain this behavior, or is it ok as is?
> 
> I'm not super familiar with the docs and can't remember where in them I maybe bumped into this feature, many years ago. I may have discovered it when reading the code. Perhaps https://docs.zen-cart.com/user/ezpages/ezpages_meta_tags/ but it suggests a plugin. https://docs.zen-cart.com/user/customizing/add_pages/ mentions adding custom pages but no meta tag details. I could add some doc, maybe to that first page, `ezpages_meta_tags`?
> 
> edit: see my latest commit using `zen_define_default`, does not require additional braces :)

I found reference to the code that raised the issue associated to this correction/issue at the end of here: https://docs.zen-cart.com/user/admin/meta_tags/

It's not necessarily that there is added content that "causes" it per se, just a capture of the instruction in the file where guidance is provided about how to add data to the Metatag file for pages such as this.

That said, based on how the ez pages controlling code is written to also invoke the default section that raised this issue, it seems there may be (have been) some benefit to add text there. It now doesn't appear necessary.

neekfenwick:
> I found reference to the code that raised the issue associated to this correction/issue at the end of here: https://docs.zen-cart.com/user/admin/meta_tags/

Good find @mc12345678 that `Per-Page meta-tags` bit looks exactly right, and I agree, it doesn't need modifying, it's just that the code that implements it was dodgy :)

Thanks for the merge @scottcwilson

PR #5509, opened 2022-12-28
Write one archive record per email

scottcwilson:
Right now, one record is written *per receipient*.  

We might want to make this behavior switchable - not sure.

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -81,6 +81,7 @@ use PHPMailer\PHPMailer\SMTP;
     if (empty($from_email_name) || $from_email_name == $from_email_address) $from_email_name = STORE_NAME;
 
     // loop thru multiple email recipients if more than one listed  --- (esp for the admin's "Extra" emails)...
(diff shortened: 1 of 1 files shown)


lat9:
What impact would this have on the "Email Archive Manager"?

scottcwilson:
Less records - if your "Send Copy of Order Confirmation Emails To" has 3 emails specified, then instead of 3 records for Email Archive Manager to show, there would be ONE.

scottcwilson:
Trouble is, if we use the `site_specific_overrides` files to allow this behavior to be configured, we'd have to put it in both places (since email can be created from storefront or admin).   Not sure.  What if we put a flag at the top of `includes/functions/functions_email.php` ?

scottcwilson:
I'm a flippin' and a floppin' about this.

lat9:
No flags in functions, unless they're overridable.

scottcwilson:
Let's close this for now.  Even the customer I had in mind might not be a good match for this change after all.

PR #5510, opened 2022-12-31
Check email address before using in query

scottcwilson:


proseLA:
what problem are you trying to solve?  

with the potential number of  `preg_*` functions being called, i am not convinced that this will be any faster than hitting the database (feel free to change my mind).  nor am i convinced that this named function should be doing the validation.

scottcwilson:
This is a response to an attack.  I don't want to spend resources doing a query.

proseLA:
so you would prefer burning resources running 8 regex expressions instead?  do you think that will be faster?

why don't you explain the problem that you are trying to solve instead of providing a solution.

i recently had someone wanting a bug bounty for a potential attack on a clients site on the login page.  perhaps a limiter on the number of requests based on time is more appropriate.

scottcwilson:
I'm trying to prevent bad guys from repeatedly posting garbage in the email address to the login URL.  It's not even an email they're sending, it's a bunch of Cyrillic characters.

scottcwilson:
> i recently had someone wanting a bug bounty for a potential attack on a clients site on the login page. 
> perhaps a limiter on the number of requests based on time is more appropriate.

I am very open to this as well.  But it's a different issue.

scottcwilson:

PR #5512, opened 2023-01-02
Include Plugins in Search?

### LINKED ISSUE #5511: Implement plugin search in 1.5.8 that was in 1.5.7
Seeking to carry over the code change(s) made in PR #4486 to Zen Cart 1.5.8.

PR will be generated momentarily.

mc12345678:
Incorporates the upgraded changes made in #4486.

With the addition of Plugins to the admin side of the store, searching for information within the plugins remains advantageous. However, with the addition of plugin versions, all plugin software is searched when looking at the admin and not just the latest or active version nor is there a way to exclude searching the plugins on a per search basis.

This adds a checkbox to add the zc_plugins directory into the applicable search which currently is all files, all admin and just the plugins searches plugins directory. An alternative solution that was temporarily locally implemented was to add options into the dropdown to describe performing the search without plugins; however, the list becomes that much longer and more code to be added.

The choice of add plugins was made to demonstrate that if plugins was already chosen, it would automatically and already be included, but if plugins were not originally part of the associated directory (admin for example doesn't have a plugins sub-directory) then it incorporates the plugins into the search.

Fixes #5511

### DIFF
--- a/admin/developers_tool_kit.php
+++ b/admin/developers_tool_kit.php
@@ -22,6 +22,7 @@ if (!empty($_GET['configuration_key_lookup']))  {
 $configuration_key_lookup = $_POST['configuration_key'] ?? '';
 $default_context_lines = (int)($_POST['context_lines'] ?? $default_context_lines);
 $case_sensitive = !empty($_POST['case_sensitive']);
+$include_plugins = !empty($_POST['include_plugins']);
 $q_const = $q_func = $q_class = $q_tpl = $q_all = '';
 
 function getDirList($dirName, $filetypes = 1) {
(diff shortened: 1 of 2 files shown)


scottcwilson:
Same comment as other related PR: Should this be switchable, default no?

mc12345678:
For my use, no, I don't see that I would need to support a default value that I would enter into the file just so that it would always be available without having to select it. 

I do think that generally speaking in the Zen Cart code that if the code is to comply with the various PSR expectations, that they do so. For that I would expect that it would be possible in this file to set a default, but really it wouldn't matter if the entered default were changed as long as the actual default were stored elsewhere. 

I've previously assured my recommendations and been told to pack. My stance has not changed. I'd like to follow the guidelines and expectations of the project.  I'm tired.

scottcwilson:
I'm not sure what this has to do with PSR, but it sounds like your answer is "no," which is fine. Cindy also agrees that making it switchable is not valuable.

PR #5514, opened 2023-01-02
Include Laravel in Search?

### LINKED ISSUE #5513: Add search of Laravel in developers tool kit.
A great admin feature of Zen Cart is the ability to search the code associated with operation from within the admin. This makes review of the code (and how it has been modified for the site) at an allowed individual's reach rather than any number of other more difficult methods. E.g. download of the code or accessing via a tool that allows searching the file(s). Even Github has its limitations as it only searches the most recent code of the parent repo. Yes, it could be copied as an independent repo for each version, sub-version, or site being worked on; however, all of these methods generally are not in sync with the site in question, increase overhead on storage, etc...

Anyways, Laravel has been incorporated in a large part to the Zen Cart store, while there are other areas that have been explicitly excluded (`DIR_FS_CATALOG . 'includes/classes/vendors'`), Laravel is certainly an aspect of developing code that would be beneficial to include in the developers tool kit. 

Note, the `zc_install` directory is also excluded; however, following approved installation direction and *NOT* including an override to retain the directory, technically the `zc_install` directory wouldn't be accessible in a live store anyways. I have no issue with that condition, again as the `zc_install` directory becomes non-essential after initial installation. Development within the `zc_install` directory could be made possible with some minor code changes; however, those taking the effort to do so would just be adding one more task to make it possible through the developers tool kit.

mc12345678:
With the addition of Laravel to the store, searching for information within the Laravel code is advantageous. However, the laravel directory (integral to many operations and to become even more so) has been excluded from search.

This adds a checkbox to add the laravel directory into the applicable search which currently is all files and also adds a search option to search just the laravel directory (to potentially exclude other code that uses code from that directory).

The choice of add laravel was made to demonstrate that if laravel was already chosen, it would automatically and already be included, but if laravel was not originally part of the associated directory search then it incorporates the laravel directory into the search.

Fixes #5513. May have conflicts with 5512 because they were added independently. Intention was for this (Laravel) to "follow" any plugin related content at the applicable location. I can rebase and force push if necessary to eliminate the conflict if desired/requested.


scottcwilson:
Should this be switchable, default no?

scottcwilson:
You'll need to update this branch.

PR #5516, opened 2023-01-03
Add inverse selection feature for DisplayLogs plugin

### LINKED ISSUE #5515: Add button to invert display_log selections
Would like to be able to alter the view of the files selected so that could select the files to be kept, then ultimately choose to keep those files. A standard method of implementation is to just inverse the selections.

I have generated a solution that does this, though because it is a change/revision, I propose it as version 3.0.2.

mc12345678:
There are times when there are just a few files to keep, but many others to be removed. Selecting a few to then inverse the selection and then choose to delete all selected improves efficiency.

Added a button to inverse the selection(s). Also updated the script within the main file to not be encapsulated around HTML comment blocks.

Updated as version 3.0.2. Tested the upgrade process which successfully completed.

Fixes #5515

### DIFF
--- a/zc_plugins/DisplayLogs/v3.0.2/admin/includes/languages/english/lang.display_logs.php
+++ b/zc_plugins/DisplayLogs/v3.0.2/admin/includes/languages/english/lang.display_logs.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * @copyright Copyright 2003-2022 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: Steve 2021 Apr 02 New in v1.5.8-alpha $
+*/
+
+
+$define = [
+    'HEADING_TITLE' => 'Display Debug Log Files',
+    'TABLE_HEADING_FILENAME' => 'File Name',
+    'TABLE_HEADING_MODIFIED' => 'Date Modified',
+    'TABLE_HEADING_FILESIZE' => 'File Size (bytes)',
+    'TABLE_HEADING_DELETE' => 'Delete?',
+    'TABLE_HEADING_ACTION' => 'Action',
+    'BUTTON_INVERT_SELECTED' => 'Inverse Selection(s)',
+    'BUTTON_DELETE_SELECTED' => 'Delete Selected',
+    'DELETE_SELECTED_ALT' => 'Delete all selected files',
+    'BUTTON_DELETE_ALL' => 'Delete All',
+    'DELETE_ALL_ALT' => 'Delete all files in the current view',
+    'ICON_INFO_VIEW' => 'View the contents of this file',
+    'DISPLAY_DEBUG_LOGS_ONLY' => 'Display debug-logs only?',
+    'TEXT_HEADING_INFO' => 'File Contents',
+    'TEXT_MOST_RECENT' => 'most recent',
+    'TEXT_OLDEST' => 'oldest',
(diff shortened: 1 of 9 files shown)


lat9:
@mc12345678, I'll get around to reviewing this (hopefully over the weekend).  The main problem I've got with the review is that since this is a plugin and you've properly updated the version, there's no automatic GitHub identification of what's changed, so it will take me some extra time.

scottcwilson:
I keep thinking that's an unfortunate unintended consequence of the encapsulated plugins that we might want to review.

mc12345678:
A path I had considered and could still do was to incorporate in two commits. First updated the software, second the file path to place in the versioned directory.

lat9:
Good idea, @mc12345678.  That would at least give any reviewer a quick look at the changes.  Maybe we should add that to the plugin-update procedure for the future.

BTW, thanks for this ... I can't count the number of times that I wished that this feature was present.

scottcwilson:
Starting discussion on version numbering in encapsulated plugins: https://github.com/zencart/zencart/discussions/5534

zc_plugins/DisplayLogs/v3.0.2/admin/includes/languages/english/lang.display_logs.php
```diff
@@ -0,0 +1,38 @@
...
+
+
+$define = [
+    'HEADING_TITLE' => 'Display Debug Log Files',
+    'TABLE_HEADING_FILENAME' => 'File Name',
+    'TABLE_HEADING_MODIFIED' => 'Date Modified',
+    'TABLE_HEADING_FILESIZE' => 'File Size (bytes)',
+    'TABLE_HEADING_DELETE' => 'Delete?',
+    'TABLE_HEADING_ACTION' => 'Action',
+    'BUTTON_INVERT_SELECTED' => 'Inverse Selection(s)',
```
torvista, line 17: Should be Invert (verb), not Inverse (adjective).

lat9: True, noting that a change will be required in the overall instructional text as well.  Will you PR?

mc12345678: I had wanted to use `inverse of selection(s)` but that seemed too wordy.  I had also toyed with `invert selection(s)` but that could mean to put the items currently at the top of the list at the bottom. 

So, next alternate text suggestion as there is issue with the current?

torvista: I did this

2023-01-15 15_11_19-Admin Display Logs — Mozilla Firefox

I've rejigged the bit about user prefixes so it doesn't look so bad, until someone else fixes that properly/as they see fit.

lat9: How about either "Swap Selections" or "Switch Selections"?  It might help to give an example in the unordered list, e.g.

```
For example, if you want to delete all but one file, tick the selection for the file to be kept, then "Swap Selections" and finally "Delete Selected".
```

torvista: To be honest, I was all for throwing away the explanation completely...are Zen Cart admins really that dumb?
The only non-intuitive info is that which points out that the listing is derived from pre-set and user-defined prefixes.
The rest is bleeding obvious...https://www.youtube.com/watch?v=iJSyGlmT3lk

PR #5519, opened 2023-01-07
zen_get_multiple_tax_rates: add parameter defaults

torvista:
code in function checks for these defaults, as in
function zen_get_tax_description
function zen_get_all_tax_descriptions
function zen_get_tax_rate


scottcwilson:
Seems like a good idea - consistent with the other functions mentioned.

PR #5522, opened 2023-01-09
Enhancement -Do not display Product count for products in subcategories

brittainmark:
Allow categories count to ignore products in sub categories. 
- Added flag COUNT_SUB_CATEGORIES to allow category tree to only count the products in the category specified. 
- Modified Category tree to check it the new flag exists and if not continue to count the products in subcategories. 
- Modified functions_categories to add new parameter to zen_count_products_in_categories to allow subcategories to be ignored in the count. Default true current state count properties.

This could be added now as it checks to see if the new flag is set and if not works as now.


scottcwilson:
We really can't add configs until the next major release.

brittainmark:
Ok. Please don't lose it. I know when we get to the next release I may have to redo. But I don't want to forget.

PR #5524, opened 2023-01-09
add admin footer notifier to admin home

proseLA:


proseLA:
i use it for displaying parse times on queries.

because the home page does not use the normal footer, i do not see the number of parse time stats which is useful...

scottcwilson:
Interesting.

PR #5525, opened 2023-01-09
Fix dynamic properties in test classes

marco-pm:


scottcwilson:
Thanks for doing this.

PR #5527, opened 2023-01-10
"Disable" ask-a-question if disabled for product's type

lat9:
If a site has disabled the `ask_a_question` link (via Catalog :: Product Types :: Layout Settings for a product type), the page still renders.

It's likely that the site's disabled the link because they don't want to deal with additional emails coming from that page, but a knowledgeable person could still navigate to that product-specific page via their browser's address bar.

This update simply redirects back to the product's `_info` page if the feature is disabled for its product-type.  There's also a change to the redirect if a product is not found, since the `_info` pages expect to see a `products_id` parameter, not a `pid` one.

### DIFF
--- a/includes/modules/pages/ask_a_question/header_php.php
+++ b/includes/modules/pages/ask_a_question/header_php.php
@@ -16,23 +16,33 @@ $pid = (isset($_GET['pid'])) ? (int)$_GET['pid'] : false;
 //
 if ($pid === false) {
     zen_redirect(zen_href_link(FILENAME_DEFAULT));
-} else {
-    $sql = "SELECT pd.products_name, p.products_image, p.products_model
-            FROM " . TABLE_PRODUCTS . " p, " . TABLE_PRODUCTS_DESCRIPTION . " pd
-            WHERE p.products_id = pd.products_id
-            AND p.products_id = " . (int)$_GET['pid'] . "
-            AND pd.language_id = " . (int)$_SESSION['languages_id'] . "
-            AND p.products_status = 1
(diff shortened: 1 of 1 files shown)


scottcwilson:
I'm open to this, but would it be better to redirect to contact_us instead?  Not sure.

lat9:
In my mind, if the site's disabled ask-a-question they don't want to be flooded with email questions.  That's why I suggested the redirect back to the product's page.

includes/modules/pages/ask_a_question/header_php.php
```diff
@@ -16,24 +16,34 @@
...
+        WHERE p.products_id = pd.products_id
+        AND p.products_id = " . (int)$_GET['pid'] . "
+        AND pd.language_id = " . (int)$_SESSION['languages_id'] . "
+        AND p.products_status = 1
+        LIMIT 1";
+
+$result = $db->Execute($sql);
+
+if ($result->EOF) {
+    zen_redirect(zen_href_link(zen_get_info_page($pid), 'products_id=' . $pid));
```
scottcwilson, line 42: maybe first param to `zen_href_link` should be `$info_page`?

scottcwilson:
BTW a nice bonus of your approach is that it foils bad guys who use AAQ as a spam mechanism.

includes/modules/pages/ask_a_question/header_php.php
lat9: Certainly, since that's already been gathered.  Thanks for the additional eyes on the change.

Issue #5538, opened 2023-01-15
admin, sales report with graphs:  year display different on hosting/local


torvista:
Hosting: php 8.1.4, Centos
![image](https://user-images.githubusercontent.com/4391026/212548884-b611cb0b-9084-427a-a047-5845fd08a8c2.png)
Local: php 8.1.4/8.2, Windows
![image](https://user-images.githubusercontent.com/4391026/212549004-29759c81-7f46-453b-ac21-9cb4565ce2cc.png)

Identical database.

I've dug down as far as here, from where the error originates:
https://github.com/zencart/zencart/blob/5fb07cf03e3d0e3d7817a8c3411d0e2345147d5d/includes/classes/zcDate.php#L175-L182

Echoing the $this->dateObject->getCalendarObject()
## Hosting/Bad:

> $this->dateObject->getCalendarObject() (object):
> IntlGregorianCalendar Object
> (
>     [valid] => 1
>     [type] => gregorian
>     [timeZone] => Array
>         (
>             [valid] => 1
>             [id] => Europe/Madrid
>             [rawOffset] => 3600000
>             [currentOffset] => 3600000
>         )
> 
>     [locale] => en
>     [fields] => Array
>         (
>             [era] => 1
>             [year] => 2023
>             [month] => 0
>             [week of year] => 2
>             [week of month] => 2
>             [day of year] => 15
>             [day of month] => 15
>             [day of week] => 1
>             [day of week in month] => 3
>             [AM/PM] => 1
>             [hour] => 4
>             [hour of day] => 16
>             [minute] => 21
>             [second] => 12
>             [millisecond] => 711
>             [zone offset] => 3600000
>             [DST offset] => 0
>             [year for week of year] => 2023
>             [localized day of week] => 7
>             [extended year] => 2023
>             [julian day] => 2459960
>             [milliseconds in day] => 58872711
>             [is leap month] => 0
>         )
> 
> )
> $output (string):
> Jan 22

## Local/Good:

> $this->dateObject->getCalendarObject() (object):
> IntlGregorianCalendar Object
> (
>     [valid] => 1
>     [type] => gregorian
>     [timeZone] => Array
>         (
>             [valid] => 1
>             [id] => Europe/Madrid
>             [rawOffset] => 3600000
>             [currentOffset] => 3600000
>         )
> 
>     [locale] => en_GB
>     [fields] => Array
>         (
>             [era] => 1
>             [year] => 2023
>             [month] => 0
>             [week of year] => 3
[...]

PR #5542, opened 2023-01-16
zcDate: fix parameters

torvista:
as per this thread:
https://github.com/zencart/zencart/issues/5538#issuecomment-1384209575


proseLA:
how much testing have you done on this change?

torvista:
Apart from it fixing the reported issue, none.
Reading more of the subject and looking at this indicates that a conversion may be necessary per locale..../one size does not fit all:
https://gist.github.com/ve3/a3b7924a85c3286b554f9e636b919883
So, should there be a big array (based on en_US I suppose), with small arrays per locale/in language packs for overriding the differences?

PR #5543, opened 2023-01-17
Not all group 2,3 values should be ints

scottcwilson:
Fixes issue raised on forum: 
https://www.zen-cart.com/showthread.php?229228-Manufacturers-Sidebox-Shows-Disabled-Manufacturers


torvista:
This idea of casting certain values to integers or not, with no obvious consistency and with random exceptions in the two arrays, seems like multiple accidents waiting to happen and a total block on implementing strict types if you don't know when your constant-string may become a constant-integer or vice-versa, as here.

If this is intended to reduce casting in code, surely it should be 100% numbers become integers, no exceptions, for all constants....or based on the db field definition? But certainly not some array buried in a third-party app.

lat9:
This falls in the realm of the "Law of unintended consequences"

torvista:
Since all configuration constants are strings in the database, I think they should stay as strings. Then everyone knows they have to be cast if strict code/calculations require it. No documentation required, no arrays to be maintained.

scottcwilson:
This change was put in place to guard against a specific database attack.  It had nothing to do with casting or consistency.

proseLA:
> This idea of casting certain values to integers or not, with no obvious consistency and with random exceptions in the two arrays, seems like multiple accidents waiting to happen and a total block on implementing strict types if you don't know when your constant-string may become a constant-integer or vice-versa, as here.
> 
we truly DO know when a constant will be an integer or a string.  all one needs to do is look at the code.  which is a 42 line script that looks pretty easy to read.  especially when compared to the many 1000+ lines of code that many other scripts have; as well as the multitude of duplicate code that exists on the repo.

as i have previously stated, refactoring to reduce this duplicity as opposed to ensuring that 1 set of code is the same as another would be a much better effort.

proseLA:
https://github.com/zencart/zencart/blob/95f5721f05ec6caafb06ea7463f6974cea33504b/laravel/app/Models/Configuration.php#L31

is it really this hard to follow a [rudimentary PSR-12 rule?](https://www.php-fig.org/psr/psr-12/#:~:text=the%20keyword%20elseif%20should%20be%20used%20instead%20of%20else%20if%20so%20that%20all%20control%20keywords%20look%20like%20single%20words.)

PR #5546, opened 2023-01-21
Update zcDate.php

piloujp:
I know this issue had a pull request by Torvista but he deleted it for any reason. There is definitely a bug here and easy to fix.
zcDate parameter conversion to intl date format fix: YY -> yy and YYYY -> y to use standard year instead of year by 'week of year'. See details in issue #5538

Problem/bug (returning wrong year) will append around 1st of January (depending on year) of each year when a date is formatted with zc_date.php transmitting no timestamp parameters or any time using a timestamp parameter set at first days of the year.
Using following Zen Cart files will trigger it (might depends on parameter used):
includes/classes/order.php on lines 1307, 1318
includes/classes/traits/NotifierManager.php on line 126
includes/functions/functions_dates.php on lines 54, 102
includes/functions/functions_files.php on line 354
includes/modules/payment/authorizenet.php on lines 212, 291
includes/modules/payment/authorizenet_aim.php on lines 231, 303
includes/modules/payment/paypaldp.php on lines 443, 476, 684
admin/downloads_manager.php on line 200
admin/includes/classes/stats_sales_report_graph.php on line 310, 314, 318
admin/includes/header.php on line 203
admin/includes/modules/dashboard_widgets/BaseStatisticsDashboardWidget.php on line 35
admin/includes/modules/dashboard_widgets/TrafficDashboardWidget.php on line 29

This year 2023 was kind of a special case when using en_US locale as 1st of January was on Sunday (1st day of the week in this local) making this bug invisible.


piloujp:
Not one comment... What should I do, just delete this? Or is there a chance it will be merge one day?
I don't see why one would continue to use week calendar to change year rather than standard date.

torvista:
I've been using it since you posted it. **I** certainly appreciated the very detailed explanations.

scottcwilson:
Hang tight - I thought I sent this to Cindy but there was probably no notification.  @lat9 Please check.

PR #5548, opened 2023-01-21
Eliminate debug generation

### LINKED ISSUE #5547: v158 zen_get_configuration_key_value_layout throws warning for non-existent configuration_key
if the function `zen_get_configuration_key_value_layout` is called where `$lookup` is a value that does not exist as a `configuration_key`, then a warning debug log is generated.

Discovered using Zen Cart 1.5.8, PHP 8.2.
Reproducing this requires either removal of a `configuration_key` value from the database that is used in a Zen Cart call or creating code that attempts to use the function with a non-existent value to look up.

mc12345678:
Passing a lookup value for `configuration_key` that does not exist in the database generates a warning debug log in PHP 8.2.
````php
#0 /catalog/includes/functions/functions_lookups.php(98): zen_debug_error_handler()
#1 /admin/plugin.php(): zen_get_configuration_key_value_layout()
#2 /admin/index.php(11): require('admin...')
--> PHP Warning: Undefined array key "configuration_value" in catalog/includes/functions/functions_lookups.php on line 98.
````

Fixes #5547

### DIFF
--- a/includes/functions/functions_lookups.php
+++ b/includes/functions/functions_lookups.php
@@ -95,6 +95,9 @@ function zen_get_configuration_key_value_layout($lookup, $type = 1)
 {
     global $db;
     $configuration_query = $db->Execute("select configuration_value from " . TABLE_PRODUCT_TYPE_LAYOUT . " where configuration_key='" . zen_db_input($lookup) . "' and product_type_id='" . (int)$type . "'");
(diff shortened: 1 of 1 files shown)


scottcwilson:
As noted in #5547, the current behavior is appropriate.  If you get this log, either your code is wrong or your database has been damaged.  The log alerts you to this.

PR #5555, opened 2023-01-26
shopping cart: equate form name and id

torvista:
old id shoppingCartForm not used anywhere


scottcwilson:
There's really no reason to do this, and it might break someone's modification.  You opened a couple of issues yesterday that are real bugs - why not fix those instead?

PR #5556, opened 2023-01-26
clean up code page.php

Zen4All:
- Removed unused code for horizontal page links. This solves the unused and undefined variables
- Removed brckets around `require` statement
- shorthand array notation

resolves zencart/zencart/#5552

### DIFF
--- a/includes/modules/pages/page/header_php.php
+++ b/includes/modules/pages/page/header_php.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * ez_pages ("page") header_php.php
  *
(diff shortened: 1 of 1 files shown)


lat9 (changes requested):
Welcome back!  While you're in here anyway, how about also adhering to the 4-space indent?

Zen4All:
Is that the new standard on all files that are getting changed?

lat9:
We're trying to adhere with PSR-12 for refactored and new files.

lat9 (approved):

includes/modules/pages/page/header_php.php
```diff
@@ -20,21 +21,20 @@
...
 }
 
 $chapter_id = isset($_GET['chapter']) ? (int)$_GET['chapter'] : 0;
 $chapter_link = isset($_GET['chapter']) ? (int)$_GET['chapter'] : 0;
 
-
 if (!$sniffer->table_exists(TABLE_EZPAGES_CONTENT)) {
-  return; // early exit; db not upgraded
+    return; // early exit; db not upgraded
 }
```
lat9, line 32: Sorry to be a PITA on this, but this section should be removed.  If that table doesn't exist, the site/site-support should be made aware of the issue instead of silently letting it go by.

lat9 (changes requested):
See comment; lines 30-32 should be removed.

includes/modules/pages/page/header_php.php
scottcwilson: I added this a long time ago when the tables were split because people were getting in trouble trying things before doing the DB upgrade.  I don't have strong feelings but it would be easier just to let it be - it's not causing problems.

Zen4All:
done

lat9 (approved):

lat9:
Thanks for this, @Zen4All!  I appreciate your patience with me.

Zen4All:
Na problem :) I rather make do some additional changes, and make it as good as possible. Then not doing it and have code that could be better.

PR #5558, opened 2023-01-26
Implement search-box module into most files.

Zen4All:
This reduces code repetition in multiple files


lat9 (changes requested):
Nice!

For each instance of
`          <?php include DIR_WS_MODULES . 'search_box.php'; ?>`

... please change to
`          <?php require DIR_WS_MODULES . 'search_box.php'; ?>`

... since an error should occur if the module is not present.

lat9 (approved):

PR #5560, opened 2023-01-26
add cat_id to product category search

proseLA:
use case:

working with a client who has 5726 categories short of 4 gazillion, i have found the ability to input a category_id number into the search field very beneficial.

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -540,6 +540,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
               $keyword_search_fields = [
                 'cd.categories_name',
                 'cd.categories_description',
+                'cd.categories_id',
               ];
               $sql .= zen_build_keyword_where_clause($keyword_search_fields, trim($keywords), true);
           } else {


scottcwilson:
Like it, but one question: should we separate out product id/category id search to make it faster and produce less false positives?

scottcwilson:
(This could be done in a separate PR; not saying this has to change, just want your opinion on the topic.)

proseLA:
> Like it, but one question: should we separate out product id/category id search to make it faster and produce less false positives?

i am not sure i understand the question.

- i think the search happens rather quickly, and i do not think there is a bottle neck here.    
- i'm not sure how you would propose to have separate searches.   
- this code does not do the searching.  adding an array element only adds that element to the where search string.

the simplicity of a google search, is to type anything in there, and you get relevant results.  that is what i was trying to mimic.  i would not be a fan of configuring the search via additional DOM elements.

if you want to elaborate more, perhaps i can better see what you are requesting.

scottcwilson:
I'll write it up separately.

PR #5568, opened 2023-01-30
Fix ability to make admin forms target a different page with GET requ…

### LINKED ISSUE #5454: Can no longer create an admin form that targets a different page with GET request

neekfenwick:
…est fixed #5454.

As requested by @zcwilt

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -244,7 +244,7 @@ function zen_image_submit($image, $alt = '', $parameters = '')
     $form .= '>';
     if (strtolower($method) == 'post') $form .= '<input type="hidden" name="securityToken" value="' . $_SESSION['securityToken'] . '">';
     if (strtolower($method) == 'get') {
-      $form .= '<input type="hidden" name="cmd" value="' . (isset($_GET['cmd']) ? $_GET['cmd'] : 'home') . '">';
+      $form .= '<input type="hidden" name="cmd" value="' . $action . '">';
     }
     return $form;
   }


zcwilt (approved):

scottcwilson:
Thanks Ian.

Issue #4922, opened 2022-06-26
admin, search_box: make sticky?


torvista:
Suggestion to re-populate the search box with the search term?
...makes it easier to correct a minor misspell (i.e. when no results found).
I find it useful.

scottcwilson:
I think this would annoy people who don't misspell and just want to do a new search.  Please make this into a plugin if you feel it would be useful.

PR #5573, opened 2023-02-02
allow repopulate of search box

proseLA:
use case:
you have a database of 12,486 products.  you are trying to find one using search terms.  you search for `red widget`.  that search results in 784 products.  you now want to search for `red widget handle`.  this PR would automatically repopulate the search input field with `red widget `; and remove the checkbox.  so now one can add `handle` to the input field.


torvista:
good idea
https://github.com/zencart/zencart/issues/4922

admin/includes/javascript_loader.php
```diff
@@ -29,6 +29,23 @@
...
+$searchBoxScriptArray = [
+    'specials',
+    'coupon_admin',
+    'reviews',
+    'featured',
+    'customers',
+    'category_product_listing',
+    'downloads_manager',
+];
+$searchBoxJs = 'includes/javascript/searchBox.js';
```
lat9, line 42: Perhaps a notification here to enable the search functionality to be included on other pages, too?

proseLA: one of us is confused.  this functionality is dependent on the following module:

admin/includes/modules/search_box.php

the scripts listed above are the only scripts that make use of that module.  if a new (or existing) script makes use of the search_box module, then by all means, said script should be included in the array.  

if a new script not part of the base ZC makes use of the search_box module, there already is a mechanism for including `js` files, and an include in said file to the `searchBox.js` would be appropriate.

or am i confused?

lat9: It just seemed (to me, at least) that providing a means for a plugin to hook into the already-existing code to prevent duplication would be desired.

admin/includes/modules/search_box.php
```diff
@@ -37,16 +37,18 @@
...
-            <div class="col-sm-3 control-label"><?= TEXT_INFO_SEARCH_FILTER_REPOPULATE; ?></div>
-            <div class="col-sm-9">
-                <?= zen_draw_checkbox_field('repopulateSearch', '', (!empty($_GET['repopulateSearch']) && $_GET['repopulateSearch'] === 'on'), '', ' id="repopulateSearch"'); ?>
+        <div class="row">
+            <div class="form-group col-md-4" id="searchTermRepopulate">
+                <div class="col-sm-9 control-label"><?= TEXT_INFO_SEARCH_FILTER_REPOPULATE; ?></div>
+                <div class="col-sm-1">
+                    <?= zen_draw_checkbox_field('repopulateSearch', '', (!empty($_GET['repopulateSearch']) && $_GET['repopulateSearch'] === 'on'), '', ' id="repopulateSearch"'); ?>
+                </div>
             </div>
```
lat9, line 46: In order for the checkbox's label to also be clickable to toggle the checkbox, the HTML needs to be of the following form:
```
        <div class="checkbox">
            <label>
                <?php echo zen_draw_checkbox_field('repopulateSearch', '', (!empty($_GET['repopulateSearch']) && $_GET['repopulateSearch'] === 'on'), '', ' id="repopulateSearch"'); ?>
                <?php echo ' ' . TEXT_INFO_SEARCH_FILTER_REPOPULATE; ?>
            </label>
        </div>
```
Granted, I've provided neither a col-xx-yy class nor the 'id' attribute, but that's the basic Bootstrap structure that's required for the label to act a the checkbox toggle-extension.

Ditto for the second checkbox, below.

admin/includes/modules/search_box.php
```diff
@@ -35,6 +35,24 @@
...
     <?php
+    if (file_exists($searchBoxJs)) {
+        ?>
+        <div class="row">
+            <div class="form-group col-md-4" id="searchTermRepopulate">
+                <div class="col-sm-9 control-label"><?= TEXT_INFO_SEARCH_FILTER_REPOPULATE; ?></div>
+                <div class="col-sm-1">
+                    <?= zen_draw_checkbox_field('repopulateSearch', '', (!empty($_GET['repopulateSearch']) && $_GET['repopulateSearch'] === 'on'), '', ' id="repopulateSearch"'); ?>
+                </div>
+            </div>
```
lat9, line 46: In order for the checkbox-label to also participate in the toggle of the checkbox, the base HTML structure needs to be similar to
```
        <div class="checkbox">
            <label>
                <?php echo zen_draw_checkbox_field('repopulateSearch', '', (!empty($_GET['repopulateSearch']) && $_GET['repopulateSearch'] === 'on'), '', ' id="repopulateSearch"'); ?>
                <?php echo ' ' . TEXT_INFO_SEARCH_FILTER_REPOPULATE; ?>
            </label>
        </div>
```

Note that I've included neither a width-defining class nor the `id` attribute for that outer `<div>` as I was unclear as to your intentions.  A similar change should be applied to the second checkbox as well.

torvista:
After trying this, my opinion is that 
- selecting re-populate:  the search term should always populate the input/no need for a select box to choose whether it should or not.
- selecting ID filter: perhaps the checkbox could appear on the same line/after the Search Box (permanently or once you start typing). Since that search box is large it could give space to that option. Without trying it, not sure if the input box would jump about/be bad UI if was shown on onkeyup.

So reducing this
image

to this
Untitled-1

proseLA:
@torvista thanks for your input.
> * selecting re-populate:  the search term should always populate the input/no need for a select box to choose whether it should or not.

that fight was already lost.  #4922  and in fact, i like this implementation better than the search term already being in there.  but it is a slight preference.

> * selecting ID filter: perhaps the checkbox could appear on the same line/after the Search Box (permanently or once you start typing). Since that search box is large it could give space to that option. Without trying it, not sure if the input box would jump about/be bad UI if was shown on onkeyup.
> 
eh....
i am not the css king and frankly i'm done playing with the css here.  i suppose it's a question of style.  adding the keyup is an interesting ideal.  maybe something for the future.  someone else can deal with styling.

again, thanks for your input!

torvista:
Search for DVD: returns DVDs as expected.
Click on repopulate: inserts DVD as expected.
Click on Restrict Search to IDs and it appears to return all products and categories instead of nothing, as expected.

proseLA:
you will have to change your expectations.  clicking on restrict id when the search term is non numeric will always return all products. this is expected behavior.

scottcwilson:
In the immortal words of O-Town, I want it all, or nothing at all. :) 
Seems ok to me as is.

torvista:
Hmm...I assumed "restrict" would, er, restrict: return less results than when not restricted!

scottcwilson:
I wouldn't want someone to spend even a minute fiddling with an edge case like this.  It's fine as is.

neekfenwick:
Minor nitpick on this, the checkboxes added to search_box.php don't lay out well for some widths, when the breakpoint is over 992px the `form-group` they are in is set to `col-md-4` which is unnecessarily too narrow.  I've made a couple of minor changes that help the layout both in small (mobile) and medium/large (desktop) sizes.

Please see PR #5853 - I'm not sure how to relate that to this issue without making it look like it 'fixes' the issue, when it's just a cosmetic tweak.

PR #5575, opened 2023-02-03
zen_get_uprid, simplify and return string

### LINKED ISSUE #5452: zen_get_uprid returns a mixed (int|string) value, not a string
Having just spent the last hour trying to figure out why my code wasn't working ... a non-attributed product, when added to the cart, results in an (int) value being returned for an 'id' element of the `$_SESSION['cart']->get_products()` returned array.

Since its results can't be counted on as a string, as it's been that way forever, perhaps it's best to just update the function's definition to indicate that it returns `mixed`.

lat9:
Fixes #5452

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -403,7 +403,9 @@ function zen_unlink_product_from_all_linked_categories($product_id, $master_cate
 function zen_get_uprid($prid, $params)
 {
     $uprid = $prid;
-    if (!is_array($params) || empty($params) || strstr($prid, ':')) return $prid;
+    if (!is_array($params)) {
+        return (string)$prid;
+    }
 
     foreach ($params as $option => $value) {
         if (is_array($value)) {


mc12345678:
Really not going to explain what it is that "got simplified" and how these changes might break code or possibly cause a memory overload?

lat9:
> Really not going to explain what it is that "got simplified" and how these changes might break code or possibly cause a memory overload?

You're right, I should have kept the !empty() clause; I'll re-PR.

torvista:
I drop this
declare(strict_types=1);
in most files I touch, as an additional future-proofing code check.

I just did this in functions_products and got an error in the shopping cart because of the two product ids in the cart array, one was integer and the other was a string (product_id:hash). 

The integer id caused the error when it got here:
function zen_get_prid(string $uprid)

I didn't have time to finish determining where it had been cast to integer (although the add to cart button seems to always send integer) since it was further complicated by both products being simple (no attributes), but one had a hash: https://www.zen-cart.com/showthread.php?229279-What-is-the-hash-added-to-the-product_id-in-the-shopping-cart

So, while I may have oddities in my shop that require more investigation, while you are fettling with this, I suggest you use that strict option to highlight potential casting conflicts.

PR #5582, opened 2023-02-06
get_uprid: Check for empty string is needed

### LINKED ISSUE #5452: zen_get_uprid returns a mixed (int|string) value, not a string

lat9:
Fixes #5452 (again)

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -403,7 +403,7 @@ function zen_unlink_product_from_all_linked_categories($product_id, $master_cate
 function zen_get_uprid($prid, $params)
 {
     $uprid = $prid;
-    if (!is_array($params)) {
+    if (!is_array($params) || $params === []) {
         return (string)$prid;
     }
 
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -402,11 +402,25 @@ function zen_unlink_product_from_all_linked_categories($product_id, $master_cate
  */
 function zen_get_uprid($prid, $params)
 {
-    $uprid = $prid;
-    if (!is_array($params) || $params === []) {
-        return (string)$prid;
+    // -----
+    // The string version of the supplied $prid is returned if:
+    //
+    // 1. The supplied $params is not an array or is an empty array, implying
+    //    that no attributes are associated with the product-selection.
+    // 2. The supplied $prid is already in uprid-format (ppp:xxxx), where
+    //    ppp is the product's id and xxx is a hash of the associated attributes.
+    //
+    $prid = (string)$prid;
+    if (!is_array($params) || $params === [] || strpos($prid, ':') !== false) {
+        return $prid;
     }
 
+    // -----
+    // Otherwise, the $params array is expected to contain option/value
+    // pairs which are concatenated to the supplied $prid, hashed and then
+    // appended to the supplied $prid.
+    //
+    $uprid = $prid;
     foreach ($params as $option => $value) {
         if (is_array($value)) {
             foreach ($value as $opt => $val) {


mc12345678:
So, as before, where is the description about why an already hashed value is to get hashed again? 

If one wants to intentionally rehash the value, shouldn't they have their own means to introduce the prid without the already existing `:`? In that way they have their own means to evaluate equivalency?

lat9:
> So, as before, where is the description about why an already hashed value is to get hashed again?
> 
> If one wants to intentionally rehash the value, shouldn't they have their own means to introduce the prid without the already existing `:`? In that way they have their own means to evaluate equivalency?

If you'll note in the existing function that the string-checking clause would never have been reached.

The original check was for (a) not an array, (b) not empty and (c) not a string containing ':'.  Since a string containing ':' is not an array, that clause would never be executed and the code would (as now) go down the path where the input value is simply cast to a string and returned.

mc12345678:
> > So, as before, where is the description about why an already hashed value is to get hashed again?
> > If one wants to intentionally rehash the value, shouldn't they have their own means to introduce the prid without the already existing `:`? In that way they have their own means to evaluate equivalency?
> 
> If you'll note in the existing function that the string-checking clause would never have been reached.
> 
> The original check was for (a) not an array, (b) not empty and (c) not a string containing ':'. Since a string containing ':' is not an array, that clause would never be executed and the code would (as now) go down the path where the input value is simply cast to a string and returned.

And you should note that the first two checks were against `$params`, the third check was against `$prid`. So, if `$params` was provided as an array having at least one element then the third item ***would*** have been encountered and processing stopped. The two changes proposed this far to make this function "simpler" still don't explain why that third check was removed.

lat9:
Updated with commentary.

lat9:
@mc12345678, what say ye?  Did I _finally_ get this right?  Thanks for keeping an eye out!

mc12345678:
> @mc12345678, what say ye? Did I _finally_ get this right? Thanks for keeping an eye out!

Still considering the effects of the test `!== false`. 

Previously the "`$prid`" could begin with `:` such as removing the beginning integer part of the value and then feeding it into the function to get rehashed. An effort that seems controlled and purposeful. 

By eliminating that feature, the resulting hash will not be possible without revising the core code to restore that feature. 

Trying to see if that has been accidentally happening. e.g. that `$prid` may have been empty resulting in the hash beginning with `:` and not beginning with `0` or some other number. 

I mean, alternatively, if going to change the output of the function given equivalent input and no declared problem of the original output, shouldn't a new function be created? 

Another aspect that am investigating is related to the discussion about how strings are no longer treated as arrays. The implication of that concept is the perception that `is_array` of a string would have returned `true` in older php versions. The PHP library doesn't reflect a change in operation of that function for that purpose. But trying to run some tests to see if this function is now missing a conversion to retain operation.

lat9:
The previous implementation used `strstr($prid, ':')` which would return `false` if there was no ':' in the submitted $prid or the substring of the submitted string starting at the ':' otherwise.  If any ':' was found in the $prid (at _any_ position), that portion of the clause would be 'truthy'.

As such, using the more performant `strpos($prid, ':') !== false` provides the same functionality.

lat9:
@mc12345678, again waiting for your response.  This needs to be 'put to bed'.

PR #5584, opened 2023-02-06
Enable site-specific admin .css customizations

### LINKED ISSUE #5583: Admin:  No way for a site to provide overall styling customizations
While the `/admin/includes/admin_html_head.php` provides a means to provide page-by-page styling overrides, there's no way for a site (without carving up 'core' `.css` files) to provide overall styling changes for their admin.

I propose that the above module look for `/admin/includes/css/site-specific-stylesheet.css` and, if present, include that just after the main `stylesheet.css` is loaded.

lat9:
Fixes #5583

### DIFF
--- a/admin/includes/admin_html_head.php
+++ b/admin/includes/admin_html_head.php
@@ -31,6 +31,13 @@ if (!defined('IS_ADMIN_FLAG')) {
     <link rel="stylesheet" href="includes/css/jAlert.css">
     <link rel="stylesheet" href="includes/css/menu.css">
     <link rel="stylesheet" href="includes/css/stylesheet.css">
+<?php
+if (file_exists('includes/css/site-specific-stylesheet.css')) {
+?>
+    <link rel="stylesheet" href="includes/css/site-specific-stylesheet.css">
+<?php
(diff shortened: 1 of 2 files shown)


scottcwilson:
Love it!  But should we not provide a dist-includes/css/site-specific-stylesheet.css as we did in other cases?

torvista:
I would suggest the use-case for this would be users not liking something. That "something" could be in various css, depending on the page/mods etc., so I would have it as the last-loaded css.

lat9:
> Love it! But should we not provide an `admin/includes/css/dist-site-specific-stylesheet.css` as we did in other cases?

... and, perhaps, changing it to have a `.php` extension, so that the styling could have programmatic modifications?

scottcwilson:
Good idea.

scottcwilson:
Good point.  I will create a stub: 

https://docs.zen-cart.com/user/customizing/admin_customizing/

lat9:
Document link added to commentary.

scottcwilson:
Docs updated.  Thanks for doing this.

PR #5589, opened 2023-02-09
admin,Product Types: fix incorrect parameters to zen_info_image

torvista:
also check for empty:  if empty passed to zen_info_image it incorrectly says image missing (i.e. not found) instead of not defined.


admin/product_types.php
```diff
@@ -324,7 +324,7 @@
                 $dir_info = zen_build_subdirectories_array(DIR_FS_CATALOG_IMAGES);
                 $default_directory = substr($ptInfo->default_image, 0, strpos($ptInfo->default_image, '/') + 1);
                 $contents[] = array('text' => '<br>' . zen_draw_label(TEXT_UPLOAD_DIR, 'img_dir' ,'class="control-label"') . zen_draw_pull_down_menu('img_dir', $dir_info, $default_directory, 'class="form-control"'));
-                $contents[] = array('text' => '<br>' . zen_info_image($ptInfo->default_image, $ptInfo->type_name));
+                $contents[] = array('text' => empty($ptInfo->default_image) ? '' : zen_info_image($ptInfo->default_image, $ptInfo->type_name, SMALL_IMAGE_WIDTH, SMALL_IMAGE_HEIGHT));
```
lat9, line 327: Why not use a more performant `$ptInfo->default_image === ''`?  The schema for the `product_types` table indicates that the default_image is not null.

torvista: ok

lat9:
Thanks, @torvista!

PR #5590, opened 2023-02-09
admin, Product Types: misc non-combative changes

torvista:
use null coalesc
correct method name "insert"
add parameter list new box();
replace ternary test
short array syntax
remove silence operator

### DIFF
--- a/admin/product_types.php
+++ b/admin/product_types.php
@@ -7,7 +7,7 @@
  */
 require('includes/application_top.php');
 
-$action = (isset($_GET['action']) ? $_GET['action'] : '');
+$action = ($_GET['action'] ?? '');
 if (!isset($_GET['cID'])) $_GET['cID'] = '';
 if (!isset($_GET['gID'])) $_GET['gID'] = '';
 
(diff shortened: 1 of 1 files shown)


lat9:
Thanks for the care & feeding of this script!

mc12345678:
Curious, how is the stopping of code execution (unable to `unlink` a file) considered a non-functionality change? Where is the requirement for operation that the file(s) must be able to be unlinked in support of using Zen Cart? Is the delete image option disabled if unable to `unlink` and in parallel prevent the same action from being on the URL? 

If the goal was to ensure that operation could continue (e.g. the information actually deleted from the database), wouldn't it be "better" to provide some level of notification rather than an error log and a non-descript message?

torvista:
You are referring to the removal of the silence operator on the unlink command.
I am of the understanding that silence operators are to be avoided as they only hide problems that should be handled with better code.
The rare failure of this unlink would I think trigger an error message that was understandable enough to investigate. I was not of the mind to do more.

mc12345678:
> You are referring to the removal of the silence operator on the unlink command. I am of the understanding that silence operators are to be avoided as they only hide problems that should be handled with better code. The rare failure of this unlink would I think trigger an error message that was understandable enough to investigate. I was not of the mind to do more.

Yes, I was referring to the removal of the silence operator for the unlink operation and how the pr included this as a "non-functionality change".

Before this commit, if a user had locked down their images directory as suggested by the Zen Cart install and or security instructions, then chose to "remove" an image, the image would be removed from the product reference, but the file would not be deleted. Now, after this commit, neither happens and the user must go and do something else because of the error generated. Hardly a non-functionality change, accepted without further ado and merged. 

I found it as I was incorporating changes that have been made and looking into the why it changed. I expected to find the why identified as part of making the change rather than having to ask. 

Nothing makes a product better to attract more users than something that generates errors and prevents the accomplishment of a needed goal when also setup as highly suggested. Too bad for the end user. Tech answer, oh ftp into your site, allow files to be deleted, if can't do that right now and really need the image to not be shown, then run the to be provided sql. Best of luck next time and don't forget to lock down your files again.

lat9:
I'll note that the [`unlink`](https://www.php.net/manual/en/function.unlink.php) function's documentation indicates that a PHP _Warning_ is generated on failure.  As such, the product-removal will continue in its execution, with the by-product that a log is generated to when a failure occurs.

Given that the "Display Logs" tool is part of the distribution, I don't see what the big kerfuffle is.

PR #5594, opened 2023-02-10
Fix errorContainer not available on plugin upgrade

marco-pm:


scottcwilson:
I'm open to this, but should we set `errorContainer` (somewhere else) instead?

marco-pm:
> I'm open to this, but should we set `errorContainer` (somewhere else) instead?

I honestly don't feel like looking into the PluginManager classes too much, at least for now. This is the same of #5493, I had simply forgotten to include this line too.

scottcwilson:
Understood.  We'll go with this for now.

PR #5598, opened 2023-02-11
collect_metatags: remove invalid % from cols

torvista:
invalid html: no visible difference after removing them.


scottcwilson:
Looks like this is also done in a few other places: 

admin/mail.php
admin/define_pages_editor.php
admin/newsletters.php
admin/ezpages.php

I'm not sure changing 100% to 100 is really an improvement.  The parameter should probably be empty and not used in zen_draw_textarea_field; the width is now set by the enclosing class. 

```
```

torvista:
Yes it can be removed with no change visually, but the function zen_draw_textarea_field will require modifying to deal with the empty $width parameter.
https://github.com/zencart/zencart/blob/bc7fd44f4d0a46a0204565e42f0024ed72edf57d/admin/includes/functions/html_output.php#L351-L352

I could include that in this PR. 

Having no width (on these pages at least), shows no difference.

scottcwilson:
> ... but the function zen_draw_textarea_field will require modifying to deal with the empty $width parameter.

agreed.   And please change the name from `$width` to `$cols` to make it a bit less confusing - I think seeing "width" was what led people to use "100%".

May as well add the other pages listed above too.

torvista:
...better split and start again.

PR #5604, opened 2023-02-15
add notifiers to allow modification of product description/other product fields

torvista:
Use case: replacing placeholders/boilerplate texts embedded in the description.

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -17,7 +17,7 @@
  */
 function zen_get_product_details($product_id, $language_id = null)
 {
-    global $db;
+    global $db, $zco_notifier;
 
     if ($language_id === null) $language_id = $_SESSION['languages_id'];
 
@@ -26,7 +26,10 @@ function zen_get_product_details($product_id, $language_id = null)
             LEFT JOIN " . TABLE_PRODUCT_TYPES . " pt ON (p.products_type = pt.type_id)
             LEFT JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (p.products_id = pd.products_id AND pd.language_id = " . (int)$language_id . ")
             WHERE p.products_id = " . (int)$product_id;
-    return $db->Execute($sql, 1, true, 900);
+    $product = $db->Execute($sql, 1, true, 900);
+    //Allow an observer to modify details
+    $zco_notifier->notify('NOTIFY_GET_PRODUCT_DETAILS', $product_id, $product);
+    return $product;
 }
 
 /**
@@ -683,7 +686,7 @@ function zen_get_products_url($product_id, $language_id)
  */
 function zen_get_products_description($product_id, $language_id = 0)
 {
-    global $db;
+    global $db, $zco_notifier;
 
     if (empty($language_id)) {
         $language_id = $_SESSION['languages_id'];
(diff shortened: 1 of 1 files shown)


scottcwilson:
Nice.

includes/functions/functions_products.php
```diff
@@ -27,6 +27,10 @@ function zen_get_product_details($product_id, $language_id = null)
             LEFT JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (p.products_id = pd.products_id AND pd.language_id = " . (int)$language_id . ")
             WHERE p.products_id = " . (int)$product_id;
     return $db->Execute($sql, 1, true, 900);
```
proseLA, line 29: leaving this line (29) will ensure that your new code will never get executed.

includes/functions/functions_products.php
```diff
@@ -693,6 +697,8 @@ function zen_get_products_description($product_id, $language_id = 0)
                              FROM " . TABLE_PRODUCTS_DESCRIPTION . "
                              WHERE products_id = " . (int)$product_id . "
                              AND language_id = " . (int)$language_id, 1);
+//Allow observer to modify the description
+    $zco_notifier->notify('NOTIFY_GET_PRODUCTS_DESCRIPTION', $product_id, $product);
```
proseLA, line 701: i am always a fan of less database hits rather than more.

if you want an observer to change be able to change the description (because that is all that gets returned by this function), why not do it before the `$db->Execute` and then have an early return?  ie starting on line 695:
```
    $description = '';
    $zco_notifier->notify('NOTIFY_GET_PRODUCTS_DESCRIPTION', $product_id, $language_id, $description);
    if (!empty($description)) {
        return $description;
    }
```
(if you do not want the observer to be able to change the language_id, you can create an array for the first argument along with the product_id.)

also note that on line 690, you will also need to add zco_notifier to the global scope as well as on line 20.  (did not your IDE complain about that??)

proseLA (commented):
i am a fan of adding observers whenever needed.  but the idea that we have a 1200 line script for products functions calls out for a product class.

extending the class would allow for all of these changes.

the question as to who wants to do that and then whether it would ever get merged are reasonable ones.

includes/functions/functions_products.php
lat9: I'm _guessing_ that the change suggested would allow an observer to manipulate the description returned by the database query (I'm thinking like WP 'shortcodes').

Agreed that both functions need to add a `global $zco_notifier;`.

includes/functions/functions_products.php
torvista: late night shoddy merging (again), apologies

torvista:
actually works now.

lat9:
> i am a fan of adding observers whenever needed. but the idea that we have a 1200 line script for products functions calls out for a product class.
> 
> extending the class would allow for all of these changes.
> 
> the question as to who wants to do that and then whether it would ever get merged are reasonable ones.

Agreed, but the class' definition and use of class-extensions can get _complicated_.  What if multiple plugins are installed that have their own extensions to the class?

proseLA:
> > i am a fan of adding observers whenever needed. but the idea that we have a 1200 line script for products functions calls out for a product class.
> > extending the class would allow for all of these changes.
> > the question as to who wants to do that and then whether it would ever get merged are reasonable ones.
> 
> Agreed, but the class' definition and use of class-extensions can get _complicated_. What if multiple plugins are installed that have their own extensions to the class?

an intellectual exercise...

```
abstract class zen_product extends base 
{
// all base zen-cart functions go here
}

class product extends zen_product
{
// this class purposefully comes empty from the base
}

//  in code that requires the product

$product = new product();
```
plugins and modifications could all go into the product class.   if there are multiple plugins, it would require merging of said files.

observers and notifiers would also be preserved and could also be used.

in addition, if a plugin only needed an extension for the product class for its own use, it could name it differently as well.

i'm sure there are other solutions, but this is what quickly comes to mind.

PR #5609, opened 2023-02-15
`zen_draw_form`, correct hidden 'cmd' variable contents

### LINKED ISSUE #5607: Pagination broken for admin pages (products, option names, option values, etc.)
Repro steps: 
a) go to a page that requires pagination - with the demo data and default settings, going to 

Admin > Catalog > Option Value Manager 
or 
Admin > Catalog > Categories/Products > Test Examples 

will show the issue.

b) At the bottom of the page there's a dropdown with page numbers.   Click it and go to the next page (2).

c) Observe that you are redirected to the home page.  The desired behavior is to go to the second page of the page you were just viewing.

It looks as though the issue might be the following:
In the form that allows you to navigate through the pages, it contains

```
<input type="hidden" name="cmd" value="category_product_listing.php">
```
which should be

```
<input type="hidden" name="cmd" value="category_product_listing">
```

lat9:
The issue here was that, for 'get' type forms, the hidden field added by `zen_draw_form` neglected to remove the `.php` from the submitted `$action`, which identified the associated 'cmd=' variable's value.

Fixes #5607

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -230,24 +230,27 @@ function zen_image_submit($image, $alt = '', $parameters = '')
 
 ////
 // Output a form
-  function zen_draw_form($name, $action, $parameters = '', $method = 'post', $params = '', $usessl = 'false') {
+function zen_draw_form($name, $action, $parameters = '', $method = 'post', $params = '', $usessl = 'false')
+{
     $form = '<form name="' . zen_output_string($name) . '" action="';
     if (!empty($parameters)) {
-      $form .= zen_href_link($action, $parameters, 'NONSSL');
+        $form .= zen_href_link($action, $parameters, 'NONSSL');
     } else {
-      $form .= zen_href_link($action, '', 'NONSSL');
+        $form .= zen_href_link($action, '', 'NONSSL');
     }
     $form .= '" method="' . zen_output_string($method) . '"';
     if (!empty($params)) {
-      $form .= ' ' . $params;
+        $form .= ' ' . $params;
     }
     $form .= '>';
-    if (strtolower($method) == 'post') $form .= '<input type="hidden" name="securityToken" value="' . $_SESSION['securityToken'] . '">';
-    if (strtolower($method) == 'get') {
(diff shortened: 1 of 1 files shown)


proseLA:
i would look at #5608.  if the problem only is in the paginator, why not fix the var that gets passed to zen_draw_form as opposed to what you doing here?

lat9:
> i would look at #5608. if the problem only is in the paginator, why not fix the var that gets passed to zen_draw_form as opposed to what you doing here?

I guess because the `zen_draw_form` is where the hidden-field is actually being drawn, so I was having the function do its own clean-up.  Noting that any call to the admin's `zen_href_link` also strips any trailing `.php` from the to-be-linked page.

Either way works, I'm not wedded to my solution.

scottcwilson:
I'm tempted to merge both - seems like both are good bug fixes.  Thanks for jumping on this issue @proseLA and @lat9.

proseLA:
> I'm tempted to merge both - seems like both are good bug fixes. Thanks for jumping on this issue @proseLA and @lat9.

i'm not wedded to any solution as well.  but i do think we should not do overkill.

the problem is in the split_page class; i say we should fix it there.  fixing it in the `zen_draw_form` is imposing a performance penalty for every call to that method when it is not required.

in addition, i have done some testing, and it my testing `str_replace` is twice as fast as `basename`.  run:

```
$time_start = microtime(true);
$action = 'products_categories.php';

for ($i = 0; $i <= 1000000; $i++) {
    // method 1
    $phrase = str_replace('.php', '', $action);
}
$time_end = microtime(true);
$time = $time_end - $time_start;
echo "Did Test 1 in ($time seconds)<br /><br />";

$time_start = microtime(true);
for ($i = 0; $i <= 1000000; $i++) {
    // Method2
    $phrase = basename($action, '.php');
}

$time_end = microtime(true);
$time = $time_end - $time_start;
echo "Did Test 2 in ($time seconds)<br /><br />";
```
the longer the `$action`, the better the performance.

to summarize, my opinion is 1 fix in the split_page class and to use `str_replace`.

do with it as you wish.

PR #5610, opened 2023-02-16
Admin countries, correct missing page= on pages other than first

lat9:
... as reported in [this](https://www.zen-cart.com/showthread.php?229309-Admin-Locations-Taxes-Countries-cannot-edit-country-not-on-first-page) Zen Cart forum bug-report.

### DIFF
--- a/admin/countries.php
+++ b/admin/countries.php
@@ -9,6 +9,7 @@ require 'includes/application_top.php';
 
 $action = (isset($_GET['action']) ? $_GET['action'] : '');
 $currentPage = (!empty($_GET['page']) && ctype_alpha($_GET['page'][0]) ? $_GET['page'][0] : '');
+$page_parameter = ($currentPage !== '') ? ('page=' . $currentPage . '&') : '';
 
 if (!empty($action)) {
   switch ($action) {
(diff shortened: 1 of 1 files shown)


proseLA:
i think this fixes the problem...  however, any mis-constructed url still causes an error, ie:

admin/index.php?cmd=countries&cID=170

is that something that we should address?

scottcwilson:
@proseLA can you be more specific?  I'm not tracking. 

If I go to Admin > Locations > Countries and select "Antarctica" the URL is 

http://localhost:8888/gh_demo_158/admin/index.php?cmd=countries&page=A&cID=8

If I diddle the URL to remove the page parameter

http://localhost:8888/gh_demo_158/admin/index.php?cmd=countries&cID=8

I still works with no log under PHP 8.1.

lat9:
@scottcwilson, try that on one of the non `page=A` countries.  I'm still coffeeing up, but I'm guessing that that's the issue that @proseLA refers to.

proseLA:
> @proseLA can you be more specific? I'm not tracking.
> 

http://localhost:8888/gh_demo_158/admin/index.php?cmd=countries&cID=170

proseLA:
@lat9 well done!

Issue #5611, opened 2023-02-16
Delete product: fatal error in sql 


torvista:
Delete a product (including master category):

> [16-Feb-2023 15:53:42 UTC] Request URI: /zencart/admin/index.php?cmd=category_product_listing&action=delete_product_confirm&product_type=1&cPath=1_17&page=1, IP address: 127.0.0.1, Language id 1
> #0 [internal function]: zen_debug_error_handler()
> #1 D:\Github\zencart\includes\classes\db\mysql\query_factory.php(667): trigger_error()
> #2 D:\Github\zencart\includes\classes\db\mysql\query_factory.php(634): queryFactory->show_error()
> #3 D:\Github\zencart\includes\classes\db\mysql\query_factory.php(275): queryFactory->set_error()
> #4 D:\Github\zencart\includes\functions\functions_products.php(1061): queryFactory->Execute()
> #5 D:\Github\zencart\admin\includes\modules\delete_product_confirm.php(45): zen_remove_product()
> #6 D:\Github\zencart\admin\category_product_listing.php(239): require('D:\\Github\\zenca...')
> #7 D:\Github\zencart\admin\index.php(11): require('D:\\Github\\zenca...')
> --> PHP Fatal error: 1064:You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ''181' at line 2 :: DELETE FROM customers_basket_attributes
>                   WHERE products_id LIKE '181 ==> (as called by) D:\Github\zencart\includes\functions\functions_products.php on line 1061 <== in D:\Github\zencart\includes\classes\db\mysql\query_factory.php on line 667.

from here:
https://github.com/zencart/zencart/blob/216c7c90295dca42460f08f5c62454483416e5f8/includes/functions/functions_products.php#L1061-L1062

PR #5612, opened 2023-02-16
Delete product: fix bug in sql

torvista:
fixes https://github.com/zencart/zencart/issues/5611


torvista:
dunno why there are two in there

PR #5614, opened 2023-02-16
admin function textarea: check for cols parameter

torvista:
cols can only be a positive integer.
Aside, this dumps the % that may be passed, but subsequently I'll review the admin instances to remove those parameters.

Should be followed up by the same on the storefront to equate the functions for a future merge.

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -348,8 +348,9 @@ function zen_draw_input_field($name, $value = '~*~*#', $parameters = '', $requir
 
 ////
 // Output a form textarea field
-  function zen_draw_textarea_field($name, $wrap, $width, $height, $text = '~*~*#', $parameters = '', $reinsert_value = true) {
-    $field = '<textarea name="' . zen_output_string($name) . '" wrap="' . zen_output_string($wrap) . '" cols="' . zen_output_string($width) . '" rows="' . zen_output_string($height) . '"';
+  function zen_draw_textarea_field($name, $wrap, $cols, $height, $text = '~*~*#', $parameters = '', $reinsert_value = true) {
+    $cols = (int)$cols;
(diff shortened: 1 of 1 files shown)


proseLA:
why not use a type declaration?

lat9:
> why not use a type declaration?

That could, unfortunately, be a backwards-compatibility-breaking change.

proseLA:
> > why not use a type declaration?
> 
> That could, unfortunately, be a backwards-compatibility-breaking change.

breaking what?  i'm now running with:
```
function zen_draw_textarea_field(string $name, string $wrap, int $width, int $height, string $text = '~*~*#', string $parameters = '', bool $reinsert_value = true) {
```
and i see no problems (so far) with the base zc code.

are we that concerned with plugins that no one is updating?  either we want to push zc forward or continue with these incremental little changes that really do little for the ecosphere.

torvista:
I understood that this is type hinting, not casting? So it pushes the actual casting back to the caller, and required testing every call to this function and maybe adding (int) to some calls.

If the project plan is to end up with every function parameter being hinted (which I suppose is the ideal), then that should be made clear to developers so people can try and implement it.

>are we that concerned with plugins that no one is updating?

Nope.
Neither for the ones that **are** being updated. Tightening up the core should not bow to looser plugins. php will always be a moving target.

torvista:
I'll not that using the hinting on the function as in the example, did break collect_info as the width and height are not integers, but it's a minutes job to correct all the instances of the calls to this function, so long as the parameters are not variables.
I think hinting should be done where its as easy as this.

scottcwilson:
> either we want to push zc forward or continue with these incremental little changes that really do little for the ecosphere.

This change clarifies an extremely confusing parameter, which was miscoded in a bunch of places and is actually not even needed in the current core admin.  It's a small change, but it makes things better. 

Please proceed with the changes to functions which call `zen_draw_textarea_field()` as discussed @torvista.  Thanks for doing this.

torvista:
actually, as the function stands, you can't pass '' (empty) string if it's hinted as int.

PR #5616, opened 2023-02-18
Admin: remove cols/width from function textarea calls

torvista:
I've checked each of these, to ensure they make no different to the display, which is set by css.


scottcwilson:
This was more changes than I expected.  I thought we were just going to fix the ones that said "100%".

torvista:
On a search, there were mixtures of %, sizes, missing soft, when all were in fact size-fixed by the css, so it seemed more natural to make them all the same/remove the cols when they were doing nothing.

scottcwilson:
I'd really rather not change things that don't have to change.  Adding changes means more diffs and more frustration for upgraders.  Please remove the files that didn't use "100%" as a width so we can make this a smaller change.

torvista:
To clarify, you want the ones with 100% to have the parameter empty or 100?

scottcwilson:
Either empty or zero will work as expected (be dropped) by your change in the last PR.

scottcwilson:
TY

PR #5618, opened 2023-02-22
Update functions_dates.php

brittainmark:
Remove need to edit zen_date_raw function if changing to UK date format.

### DIFF
--- a/includes/functions/functions_dates.php
+++ b/includes/functions/functions_dates.php
@@ -12,8 +12,8 @@ if (!function_exists('zen_date_raw')) {
     /**
      * Return date in raw format
      *
-     * $date should be in format mm/dd/yyyy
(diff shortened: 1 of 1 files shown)


brittainmark:
Added as requested and squashed.

includes/functions/functions_dates.php
```diff
@@ -12,8 +12,8 @@
     /**
      * Return date in raw format
      *
-     * $date should be in format mm/dd/yyyy
-     * raw date is in format YYYYMMDD, or DDMMYYYY
+     * $date should be in format mm/dd/yyyy or dd/mm/yy
```
lat9, line 15: `dd/mm/yy` should be `dd/mm/yyyy`

lat9 (changes requested):

lat9 (approved):

PR #5631, opened 2023-02-25
#5630: Correcting manufacturer's URL redirection

### LINKED ISSUE #5630: `manufacturers_url` treated as 'relative' if it doesn't start with http{s}
If a manufacturer's URL doesn't start with either `https://` or `http://`, it's treated like a relative URL.

With a default installation, using the demo products, navigate to the admin's _Catalog :: Manufacturers_ and edit the _Microsoft_ entry, changing the URL from `http://www.microsoft.com` to simply `www.microsoft.com`.

With the "Responsive Classic" as your default template, navigate to one of the Microsoft-branded products, e.g. `index.php?main_page=product_info&cPath=1_9&products_id=26`.  From the "Manufacturer Info" sidebox, click the "Home Page" link for the manufacturer and notice that you're redirected to your site's page-not-found page, since the (invalid) manufacturer URL  is `https://example.com/www.microsoft.com`.

Of course, the redirect _should_ be made to 'http{s}://www.microsoft.com', with either an `http://` or `https://` request-type, depending on whether the site's currently running https-all-the-time.

lat9:
... and the `fixup_url` function to use https:// rather than http:// if the current page's request-type is 'SSL'.

Fixes #5630.

### DIFF
--- a/includes/functions/functions_strings.php
+++ b/includes/functions/functions_strings.php
@@ -381,7 +381,7 @@ function fixup_url($url)
     global $request_type;
 
     if (!preg_match('#^https?://#', $url)) {
-        $url = (($request_type === 'NONSSL') ? 'http' : 'https') . '://' . $url;
+        $url = '//' . $url;
     }
     return $url;
 }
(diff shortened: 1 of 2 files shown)


dbltoe:
Might I suggest //whatever.com format vs having to know for certain the status of the link’s SSL?

scottcwilson:
Not sure the request type is relevant here.  You're going to someone else's site.

PR #5634, opened 2023-03-03
Review pages should not pull a substring of review text

scottcwilson:
... not arbitrary hardcoded number.

### DIFF
--- a/includes/modules/pages/product_reviews/header_php.php
+++ b/includes/modules/pages/product_reviews/header_php.php
@@ -56,7 +56,7 @@
 
   $review_status = " and r.status = 1";
 
-  $reviews_query_raw = "SELECT r.reviews_id, left(rd.reviews_text, 100) as reviews_text, r.reviews_rating, r.date_added, r.customers_name
+  $reviews_query_raw = "SELECT r.reviews_id, rd.reviews_text, r.reviews_rating, r.date_added, r.customers_name
                         FROM " . TABLE_REVIEWS . " r, " . TABLE_REVIEWS_DESCRIPTION . " rd
                         WHERE r.products_id = :productsID
                         AND r.reviews_id = rd.reviews_id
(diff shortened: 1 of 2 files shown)


lat9:
Er, the `MAX_PREVIEW` constant has absolutely nothing to do with reviews.

scottcwilson:
Look at the review template files: 

includes/templates/responsive_classic/templates/tpl_reviews_default.php
includes/templates/responsive_classic/templates/tpl_product_reviews_default.php
includes/templates/template_default/templates/tpl_reviews_default.php
includes/templates/template_default/templates/tpl_product_reviews_default.php

They all reference that constant.  We should respect this when retrieving the data.

scottcwilson:
I am open to removing the `left` call entirely from the query and allowing the template to decide how much to show if you think that's a better solution.

brittainmark (approved):
good spot!

lat9:
> Look at the review template files:
> 
> includes/templates/responsive_classic/templates/tpl_reviews_default.php includes/templates/responsive_classic/templates/tpl_product_reviews_default.php includes/templates/template_default/templates/tpl_reviews_default.php includes/templates/template_default/templates/tpl_product_reviews_default.php
> 
> They all reference that constant. We should respect this when retrieving the data.

All those template-references use `zen_trunc_string`, e.g.
```php
zen_trunc_string(nl2br(zen_output_string_protected(stripslashes($reviews->fields['reviews_text']))), MAX_PREVIEW);
```

... and that function's a bit smarter than just a 'truncate-after-some-number-of-characters' like the SQL query's going to be.  I'd almost be more comfortable losing the text-related limitation in the SQL query instead.

scottcwilson:
I think you're right.  Why do it twice!  Let's let the template decide.   Coming right up.

lat9:
Were you going to change that query to use a JOIN as @brittainmark suggested?

scottcwilson:
I'd like to just make this one small change now and wrap this up.

lat9:
Should we create a to-do list?  There have been a couple other changes that would be 'good' that have also been shelved.

scottcwilson:
A todo list is always a good idea to get a shared understanding.

PR #5638, opened 2023-03-10
Stop duplication of defines if Square Webpay is installed.

brittainmark:
Modified to stop the already defined messages appearing if you have installed Square webPay. It defines a number of variables as constants that are already defined in this language file.

As I assume that square will be removed completely from Zen Cart at some point. So I thought it better to modify this file rather than have code in a new webPay module.


scottcwilson:
This seems reasonable.  Would you please just post an example log that you are seeing that complains about the duplicate definition?

brittainmark:
Fixes #5640

PR #5643, opened 2023-03-12
Fix admin product sort order issue

scottcwilson:
Fixes issue raised in 
https://www.zen-cart.com/showthread.php?229358-In-admin-categories-sort-category-status-enable-error


lat9 (changes requested):
You need `p.products_weight`, too.

lat9 (approved):

PR #5645, opened 2023-03-12
Update init_sessions.php

### LINKED ISSUE #5644: samesite parameter not set in cookie_test cookie.
**Describe the bug**
in test console get warning message
`Cookie “cookie_test” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite`

**Version**
Zen Cart version: 1.5.8
PHP version: 8.1

**To Reproduce**
1. Go to storefront page with browser test consol open. you must have `SESSION_FORCE_COOKIE_USE = 'True';`
2. in console warning message is displayed

**Expected behavior**
no warning message

**Additional context**
fix supplied

brittainmark:
if PHP7.3 or greater use parameters so samesite is set.

fixes #5644

### DIFF
--- a/includes/init_includes/init_sessions.php
+++ b/includes/init_includes/init_sessions.php
@@ -68,7 +68,14 @@ $_SERVER['REMOTE_ADDR'] = $ipAddress;
  */
 $session_started = false;
 if (SESSION_FORCE_COOKIE_USE == 'True') {
-  setcookie('cookie_test', 'please_accept_for_session', time()+60*60*24*30, $path, (!empty($cookieDomain) ? $domainPrefix . $cookieDomain : ''), $secureFlag);
+    if (PHP_VERSION_ID < 70300) {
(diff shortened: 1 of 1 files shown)


scottcwilson:
Do we need the PHP 7.3 test? ZC 1.5.8 won't run on lower versions. (No strong feelings on this, just checking.)

brittainmark:
admin appilcation top checks for 7.2.5
```
if (PHP_VERSION_ID < 70205) {
    die('Sorry, this version of Zen Cart requires PHP 7.2.5 or greater. <a href="https://www.zen-cart.com/requirements" rel="noopener" target="_blank">Please refer to our website</a> for the PHP versions supported.');
}
```
So for now I would leave it. Unless we want to change application top to be 7.3 up!

PR #5650, opened 2023-03-14
Paypaldp.php INVNUM format change to fight duplicated orders

njcyx:
As mentioned in the forum:

https://www.zen-cart.com/showthread.php?229318-Advise-needed-to-change-the-format-of-INVNUM-to-fight-duplicated-orders&p=1393626

This PR is a walk around solution to reduce/fight the duplicated orders/submissions. It will reduce the resolution for INVNUM value from 1s to 60s. Paypal will auto denied the duplicated transactions with the same INVNUM number.

### DIFF
--- a/includes/modules/payment/paypaldp.php
+++ b/includes/modules/payment/paypaldp.php
@@ -930,7 +930,7 @@ class paypaldp extends base {
       $optionsAll['CUSTOM'] = 'DP-' . (int)$_SESSION['customer_id'] . '-' . time();
 
       // send the store name as transaction identifier, to help distinguish payments between multiple stores:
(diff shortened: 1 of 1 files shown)


njcyx:
time() function used twice in paypaldp.php. One is for INVNUM. Another is CUSTOM. 

Paypal only uses INVNUM to identify whether it is a duplicated order or not, so I only changed INVNUM. CUSTOM remains unchanged. 

paypal statement regarding INVNUM can be found here: 

https://www.paypal.com/us/cshelp/article/how-do-i-avoid-duplicate-transactions-ts1097

Note: I tested this mod only in zen cart 1.5.7d. Not sure if the root issue to cause duplicated orders have been fixed in 1.5.8 or 1.5.8a or not.

PR #5655, opened 2023-03-16
remove extra blank lines at end of files

torvista:
as found by tool 
https://github.com/torvista/Zen_Cart-Find_Extra_Spaces


scottcwilson:
For the sqlpatch change, may as well get rid of the last `?>` as well.  Thanks for submitting this.

PR #5656, opened 2023-03-16
products image efficency

brittainmark:
If there is no product image then there will not be any additional images. Therefore why process the additional image code.


scottcwilson:
Do we need similar checks in template_default?

brittainmark:
Not sure if it's worth it. The additional images are loaded in a different place in the template, not straight after the main image (which has the if statement already there). If you want, I will add another check to the default template to bring it in line. 
Glad you asked, as I found 2 I had missed.

scottcwilson:
Please squash and I will merge.

brittainmark:
squashed as requested

PR #5662, opened 2023-03-22
Fixes for ROBOTS_PAGES_TO_SKIP

scottcwilson:
- ask a question should be in the list (since contact us is)
- order_status needs a comma after it.


torvista:
Is there a mechanism to add site-specific pages to this list?

scottcwilson:
Yes - you modify the file and add them.

PR #5663, opened 2023-03-23
check empty vs using count

scottcwilson:
Prevents log when POST data missing entirely. 

```
--> PHP Warning: Undefined array key "products_id" in /home/client/public_html/includes/classes/shopping_cart.php on line 2112.
```


brittainmark (approved):
nice and efficient.

drbyte (approved):

PR #5664, opened 2023-03-24
Handle undefined array keys in authorizenet_aim

scottcwilson:
This is a PHP8.1+ issue that happens because of increased checking. 

Handle errors like

```
--> PHP Warning: Undefined array key 3 in /home/client/public_html/site_home/_livesite/includes/modules/payment/authorizenet_aim.php on line 440.
```

```
--> PHP Warning: Undefined array key 4 in /home/client/public_html/site_home/_livesite/includes/modules/payment/authorizenet_aim.php on line 441.
```

etc

### DIFF
--- a/includes/modules/payment/authorizenet_aim.php
+++ b/includes/modules/payment/authorizenet_aim.php
@@ -441,7 +441,7 @@ class authorizenet_aim extends base {
     $this->auth_code = $response[4];
     $this->transaction_id = $response[6];
     $this->avs_response= $response[5];
-    $this->ccv_response= $response[38];
+    $this->ccv_response= ($response[38] ?? '');
     $response_msg_to_customer = $response_text . ($this->commError == '' ? '' : ' Communications Error - Please notify webmaster.');
 
     if ($this->display_specific_failure_reason_for_cvv_and_date) {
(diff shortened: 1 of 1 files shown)


lat9 (changes requested):
See comment, essentially the database storage will never be executed.

scottcwilson:
How about `!empty()`?

lat9:
> How about `!empty()`?

That won't work either, since, per the a.net/aim documentation, the field can be 0:
image

I'm thinking that it's best to simply replace `$response[7]` with `$response[7] ?? ''` in the original statement.

includes/modules/payment/authorizenet_aim.php
```diff
@@ -745,18 +745,19 @@ function _debugActions($response, $order_time= '', $sessID = '') {
...
-      $db_response_text = $response[3] . ($this->commError !='' ? ' - Comm results: ' . $this->commErrNo . ' ' . $this->commError : '');
+      $db_response_text = ($response[3] ?? "Unknown") . ($this->commError !='' ? ' - Comm results: ' . $this->commErrNo . ' ' . $this->commError : '');
       $db_response_text .= ($response[0] == 2 && $response[2] == 4) ? ' NOTICE: Card should be picked up - possibly stolen ' : '';
       $db_response_text .= ($response[0] == 3 && $response[2] == 11) ? ' DUPLICATE TRANSACTION ATTEMPT ' : '';
 
       // Insert the data into the database
       $sql = "INSERT INTO " . TABLE_AUTHORIZENET . "  (id, customer_id, order_id, response_code, response_text, authorization_type, transaction_id, sent, received, time, session_id) VALUES (NULL, :custID, :orderID, :respCode, :respText, :authType, :transID, :sentData, :recvData, :orderTime, :sessID )";
       $sql = $db->bindVars($sql, ':custID', ($_SESSION['customer_id'] ?? '-1'), 'integer');
-      $sql = $db->bindVars($sql, ':orderID', preg_replace('/[^0-9]/', '', $response[7]), 'integer');
+      $sql = $db->bindVars($sql, ':orderID', $response_order_id, 'integer');
```
lat9, line 757: Shouldn't this still include the `preg_replace` call?

lat9: So, why did than change on line 757 'disappear' in your most recent commit?

scottcwilson: should be same as original line - no change needed.

PR #5669, opened 2023-03-25
correct date format for functions zen_date_long/zen_date_short

torvista:
as per functions spec:
https://github.com/zencart/zencart/blob/f987315e8f91cf051a548a30db8539c89886b954/includes/functions/functions_dates.php#L51


lat9:
Sigh, the formatting in that file's functions hurts my eyes.  Since the file's been changed to correct the issue identified, I'll add a 'to-do' issue to address the reformatting/refactoring separate.y.

PR #5678, opened 2023-03-30
Single quote for strings

scottcwilson:


torvista:
While I have been removing (unnecessary) double quotes from my own legacy code when I come across them, I would expect a PR from me that **only** "fixed" such a thing to be dis-approved and also receive the usual flak for making an arguably pointless change and potentially remove more hair from hapless upgraders.

I'd come to the conclusion that first one must justifiably fix "something" and then further non-fixing fettling would be permitted once a file has been touched inter-version.

So, argue the point of this, or expect more of the same.

scottcwilson:
If the whole string was alpha only double quoted with no line breaks I wouldn't have bothered.  But the trouble is, not only does it mix and match single and double quotes, it also embeds newlines in quotes so someone could accidentally during translation change it to a single and mess things up.   I actually fat fingered this myself, so I made this PR as a revenge fix.

torvista:
ok

Issue #5671, opened 2023-03-27
`functions_email.php`, refactoring needed


lat9:
The various functions in the subject file contain a slew of one-line-if statements, e.g.
```php
if (!isset($block['EMAIL_SPAM_DISCLAIMER']) || $block['EMAIL_SPAM_DISCLAIMER'] == '')   $block['EMAIL_SPAM_DISCLAIMER']  = EMAIL_SPAM_DISCLAIMER;
```
... as well as other currently non-standard formatting.  Since that file's been updated for the current release, let's take the opportunity to prettify it!

scottcwilson:
Closed by #5682

PR #5683, opened 2023-04-02
functions.email, restructuring

lat9:
Building on PR #5671,

- Use `zen_define_default` where applicable.
- Use `===` and `!==` where applicable
- Remove error suppression; it's removed in PHP 8.2.
- Use `strpos` instead of `preg_match` for better performance
- Replace double-quotes with single-quotes where no in-string replacements are made.
- Replace `size` with `count`
- Use lowercase for false, true and null

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -302,7 +302,7 @@
                     if ($email_mail_server_port !== 25 && $email_mail_server_port !== 0) {
                         $mail->Port = $email_mail_server_port;
                     }
-                    if ((int)$mail->Port < 30 && $mail->Host === 'smtp.gmail.com') {
+                    if ($mail->Port < 30 && $mail->Host === 'smtp.gmail.com') {
                         $mail->Port = 587;
                     }
                     //set encryption protocol to allow support for secured email protocols
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -82,7 +82,7 @@
         }
 
         // if no text or html-msg supplied, exit
-        if (trim($email_text) === '' && (empty($block) || (isset($block['EMAIL_MESSAGE_HTML']) && $block['EMAIL_MESSAGE_HTML'] === ''))) {
+        if (trim($email_text) === '' && empty($block['EMAIL_MESSAGE_HTML'])) {
             return false;
         }
 
@@ -118,19 +118,19 @@
 
             //define some additional html message blocks available to templates, then build the html portion.
             if (is_array($block)) {
-                if (!isset($block['EMAIL_TO_NAME']) || $block['EMAIL_TO_NAME'] === '') {
+                if (empty($block['EMAIL_TO_NAME'])) {
                     $block['EMAIL_TO_NAME'] = $to_name;
                 }
-                if (!isset($block['EMAIL_TO_ADDRESS']) || $block['EMAIL_TO_ADDRESS'] === '') {
+                if (empty($block['EMAIL_TO_ADDRESS'])) {
                     $block['EMAIL_TO_ADDRESS'] = $to_email_address;
                 }
-                if (!isset($block['EMAIL_SUBJECT']) || $block['EMAIL_SUBJECT'] === '') {
+                if (empty($block['EMAIL_SUBJECT'])) {
                     $block['EMAIL_SUBJECT'] = $email_subject;
                 }
-                if (!isset($block['EMAIL_FROM_NAME']) || $block['EMAIL_FROM_NAME'] === '') {
+                if (empty($block['EMAIL_FROM_NAME'])) {
                     $block['EMAIL_FROM_NAME'] = $from_email_name;
                 }
-                if (!isset($block['EMAIL_FROM_ADDRESS']) || $block['EMAIL_FROM_ADDRESS'] === '') {
+                if (empty($block['EMAIL_FROM_ADDRESS'])) {
                     $block['EMAIL_FROM_ADDRESS'] = $from_email_address;
                 }
             }
@@ -352,8 +352,8 @@
 
             // set the reply-to address.  If none set yet, then use Store's default email name/address.
             // If sending from checkout or contact-us, use the supplied info
-            $email_reply_to_address = (isset($email_reply_to_address) && $email_reply_to_address !== '') ? $email_reply_to_address : (in_array($module, ['contact_us', 'ask_a_question', 'checkout_extra']) ? $from_email_address : EMAIL_FROM);
-            $email_reply_to_name = (isset($email_reply_to_name) && $email_reply_to_name !== '') ? $email_reply_to_name : (in_array($module, ['contact_us', 'ask_a_question', 'checkout_extra']) ? $from_email_name : STORE_NAME);
+            $email_reply_to_address = (!empty($email_reply_to_address)) ? $email_reply_to_address : (in_array($module, ['contact_us', 'ask_a_question', 'checkout_extra']) ? $from_email_address : EMAIL_FROM);
+            $email_reply_to_name = (!empty($email_reply_to_name)) ? $email_reply_to_name : (in_array($module, ['contact_us', 'ask_a_question', 'checkout_extra']) ? $from_email_name : STORE_NAME);
             $mail->addReplyTo($email_reply_to_address, $email_reply_to_name);
 
             $mail->setFrom($from_email_address, $from_email_name);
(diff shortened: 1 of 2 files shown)


includes/functions/functions_email.php
```diff
@@ -342,19 +346,19 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
...
-            if ($email_transport == 'sendmail-f' || EMAIL_SEND_MUST_BE_STORE == 'Yes') {
+            if ($email_transport === 'sendmail-f' || EMAIL_SEND_MUST_BE_STORE === 'Yes') {
                 $mail->Sender = EMAIL_FROM;
             }
 
             // set the reply-to address.  If none set yet, then use Store's default email name/address.
             // If sending from checkout or contact-us, use the supplied info
-            $email_reply_to_address = (isset($email_reply_to_address) && $email_reply_to_address != '') ? $email_reply_to_address : (in_array($module, ['contact_us', 'ask_a_question', 'checkout_extra']) ? $from_email_address : EMAIL_FROM);
-            $email_reply_to_name = (isset($email_reply_to_name) && $email_reply_to_name != '') ? $email_reply_to_name : (in_array($module, ['contact_us', 'ask_a_question', 'checkout_extra']) ? $from_email_name : STORE_NAME);
+            $email_reply_to_address = (isset($email_reply_to_address) && $email_reply_to_address !== '') ? $email_reply_to_address : (in_array($module, ['contact_us', 'ask_a_question', 'checkout_extra']) ? $from_email_address : EMAIL_FROM);
```
proseLA, line 355: can't this line and the line after be simplified to:
```
$email_reply_to_address = $email_reply_to_address ?? (in_array($module, ['contact_us', 'ask_a_question', 'checkout_extra']) ? $from_email_address : EMAIL_FROM);
$email_reply_to_name = $email_reply_to_name ?? (in_array($module, ['contact_us', 'ask_a_question', 'checkout_extra']) ? $from_email_name : STORE_NAME);
```

lat9: I wish.  Unfortunately the `??` operator looks for a value to be `null` and the two values could be an empty string.

drbyte: What about dropping `isset()` and using `empty()` instead?

lat9: Thanks, sounds like a plan.

lat9: Commit just pushed to consolidate using empty or !empty.

includes/functions/functions_email.php
```diff
@@ -281,46 +287,44 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
...
-                        }
+                    } elseif (defined('NEWSLETTER_SMTPAUTH_EMAIL_PROTOCOL') && NEWSLETTER_SMTPAUTH_EMAIL_PROTOCOL !== 'none') {
+                        $mail->SMTPSecure = NEWSLETTER_SMTPAUTH_EMAIL_PROTOCOL;
                     }
                     break;
                 case 'smtp':
                     $mail->isSMTP();
                     $mail->Host = trim($email_mail_server);
-                    if ((int)$email_mail_server_port != 25 && (int)$email_mail_server_port != 0) {
+                    if ((int)$email_mail_server_port !== 25 && (int)$email_mail_server_port !== 0) {
```
scottcwilson, line 327: These int casts shouldn't be needed - we have converted when we created this variable.

includes/functions/functions_email.php
```diff
@@ -281,46 +287,44 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
                     $mail->Port = 587;
                     $mail->Host = 'smtp.gmail.com';
                     $mail->Username = (!empty(trim($email_mailbox))) ? trim($email_mailbox) : EMAIL_FROM;
-                    if (trim($email_password) != '') {
+                    if (trim($email_password) !== '') {
```
scottcwilson, line 290: Should we start converting !== '' to !empty()?

includes/functions/functions_email.php
```diff
@@ -385,24 +389,24 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
             if (isset($newAttachmentsList) && is_array($newAttachmentsList)) {
                 $attachments_list = $newAttachmentsList;
             }
-            if (defined('EMAIL_ATTACHMENTS_ENABLED') && EMAIL_ATTACHMENTS_ENABLED && is_array($attachments_list) && sizeof($attachments_list) > 0) {
+            if (defined('EMAIL_ATTACHMENTS_ENABLED') && EMAIL_ATTACHMENTS_ENABLED && is_array($attachments_list) && $attachments_list !== []) {
```
scottcwilson, line 392: should we use count on the last condition?

includes/functions/functions_email.php
```diff
@@ -281,46 +287,44 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
...
                     $mail->SMTPAuth = true;
                     $mail->Username = (!empty(trim($email_mailbox))) ? trim($email_mailbox) : EMAIL_FROM;
-                    if (trim($email_password) != '') {
+                    if (trim($email_password) !== '') {
                         $mail->Password = trim($email_password);
                     }
-                    $mail->Host = (trim($email_mail_server) != '') ? trim($email_mail_server) : 'localhost';
-                    if ((int)$email_mail_server_port != 25 && (int)$email_mail_server_port != 0) {
+                    $mail->Host = (trim($email_mail_server) !== '') ? trim($email_mail_server) : 'localhost';
+                    if ((int)$email_mail_server_port !== 25 && (int)$email_mail_server_port !== 0) {
```
scottcwilson, line 302: cast not needed

includes/functions/functions_email.php
```diff
@@ -281,46 +287,44 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
...
                         $mail->Port = (int)$email_mail_server_port;
                     }
-                    if ((int)$mail->Port < 30 && $mail->Host == 'smtp.gmail.com') {
+                    if ((int)$mail->Port < 30 && $mail->Host === 'smtp.gmail.com') {
```
scottcwilson, line 305: cast not needed

scottcwilson (changes requested):
Let's just clean up the casts a bit more.  Other than that looks fantastic - nice improvements and simplifications.

includes/functions/functions_email.php
lat9: Nope, since it's known that the variable is an array, checking the variable specifically for an empty array is more performant.

includes/functions/functions_email.php
lat9: Nope, the $email_password is known to be a string and checking its trim for specifically an empty string is more performant.

includes/functions/functions_email.php
lat9: Removed in latest push.

includes/functions/functions_email.php
lat9: Removed in latest push.

includes/functions/functions_email.php
lat9: Missed that one, now removed.

includes/functions/functions_email.php
```diff
@@ -64,32 +57,32 @@
...
         foreach ([$from_email_address, $to_address, $from_email_name, $to_name, $email_subject] as $key => $value) {
-            if (preg_match("/\r/i", $value) || preg_match("/\n/i", $value)) {
+            if (strpos($value, "\r") !== false || strpos($value, "\n") !== false) {
                 return false;
             }
         }
 
         // if no text or html-msg supplied, exit
-        if (trim($email_text) == '' && (empty($block) || (isset($block['EMAIL_MESSAGE_HTML']) && $block['EMAIL_MESSAGE_HTML'] == ''))) {
+        if (trim($email_text) === '' && empty($block['EMAIL_MESSAGE_HTML'])) {
```
drbyte, line 85: Should this still check for `empty($block)`?

lat9: I thought long and hard on that and I came to the conclusion that if the EMAIL_MESSAGE_HTML element isn't `empty` that that implies that `$block` isn't empty either and, since that array element is present, that `$block` is also an array.

drbyte: hmmm, I think I see your point

It's still a (tiny) functional difference ... BUT then again why would one pass OTHER $block elements but NOT EMAIL_MESSAGE_HTML .. AND leave $email_text empty. 
ie: why send an html-only content message using only uncommon $block array elements?
I suppose a customized email message could use that, but then they should probably also provide text-only content themselves.
So, i guess that's a long way of thinking-out and saying I think i agree with your change! 👍

lat9: Agreed on the teeny functional difference, but I've been trying to get these highly-used functions and classes to be a tad more performant and that additional `empty` check just felt like overkill.

I appreciate you taking the time to review!

includes/functions/functions_email.php
```diff
@@ -119,50 +112,61 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
...
                     $block['EMAIL_FROM_NAME'] = $from_email_name;
                 }
-                if (!isset($block['EMAIL_FROM_ADDRESS']) || $block['EMAIL_FROM_ADDRESS'] == '') {
+                if (empty($block['EMAIL_FROM_ADDRESS'])) {
                     $block['EMAIL_FROM_ADDRESS'] = $from_email_address;
                 }
             }
             $email_html = (!is_array($block) && substr($block, 0, 6) == '<html>') ? $block : zen_build_html_email_from_template($module, $block);
-            if (!is_array($block) && ($block == '' || $block == 'none')) {
+            if (!is_array($block) && ($block === '' || $block === 'none')) {
```
torvista, line 138: I think !is_array($block) is redundant here.

lat9: True; that needs to be a separate PR.

PR #5692, opened 2023-04-10
Restore T&C Error message in checkout confirmation

brittainmark:
Having found, you can bypass ticking the T&C's box in checkout_payment. (Have bootstrap and squareWebpay installed then checkout with a credit card.)

I thought it would be prudent to restore the check in checkout_confirmation.

### DIFF
--- a/includes/modules/pages/checkout_confirmation/header_php.php
+++ b/includes/modules/pages/checkout_confirmation/header_php.php
@@ -52,7 +52,9 @@ $_SESSION['comments'] = !empty($_POST['comments']) ? $_POST['comments'] : '';
 
 
 if (DISPLAY_CONDITIONS_ON_CHECKOUT == 'true') {
-    $_SESSION['conditions'] = $_POST['conditions'] ?? NULL;
+  if (!isset($_POST['conditions']) || ($_POST['conditions'] != '1')) {
+    $messageStack->add_session('checkout_payment', ERROR_CONDITIONS_NOT_ACCEPTED, 'error');
+  }
 }
 //echo $messageStack->size('checkout_payment');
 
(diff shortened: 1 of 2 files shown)


includes/modules/pages/checkout_confirmation/header_php.php
```diff
@@ -52,7 +52,9 @@
 
 
 if (DISPLAY_CONDITIONS_ON_CHECKOUT == 'true') {
-    $_SESSION['conditions'] = $_POST['conditions'] ?? NULL;
```
lat9, line 55: What happened to setting that into the session?

lat9 (changes requested):
See comment

includes/modules/pages/checkout_confirmation/header_php.php
brittainmark: It was not in the original
```
     if (DISPLAY_CONDITIONS_ON_CHECKOUT=='true') {
 -     if (!isset ($_POST['conditions'])||($_POST['conditions']!='1')) {
 -       $messageStack->add_session ('checkout_payment', ERROR_CONDITIONS_NOT_ACCEPTED, 'error');
 -     }
 - 
 +        $_SESSION['conditions'] = $_POST['conditions'] ?? NULL;
```

 Refactor T&C #5019

Happy to put it in if you think it is required.

brittainmark: Thinking a bit more about it, I do not think there is any value in setting $_SESSIONS['conditions'] as it would either be set to 0 or NULL, neither of which matter as the tick box would be unticked unless $_SESSIONS['conditions'] was set to 1.

still happy to change if you disagree!

lat9: OK, just looked like it was a removed line.

PR #5698, opened 2023-04-13
unset session vars when doing an hmac admin login

proseLA:
i have tracked down the following error:
```
PHP Fatal error: unknown address_book_id (225223) for customer_id (5684) in /var/www/myBloodyDomain/includes/classes/OnePageCheckout.php on line 1063.
```
to some session vars not getting unset when an admin places multiple orders in a row.

there might be other session vars that need to get unset; and destroying the session entirely is not really an option.


lat9:
Holy crow!  I've been pulling hair on that one for *years*!

PR #5703, opened 2023-04-14
Fixes missing insert button in Admin > Manufacturers

### LINKED ISSUE #5699: No insert option in admin when manufacturers list is empty
**Describe the bug**
In ZC v1.5.8a, if the manufacturers list in admin is empty (as would happen if ZC was installed without demo data) then the `Insert` button is not present. There is therefore no method for subsequently adding a new manufacturer.
Issue raised in this thread, https://www.zen-cart.com/showthread.php?229417-Manufacturers-(v1-5-8a)

**Version**
Zen Cart version: 1.5.8a
PHP version: 8
What older/newer versions were also tested? Issue not present in v1.5.8

**To Reproduce**
1. Install ZC without demo products or delete all existing manufacturers
2. Go to admin > catalog > manufacturers
3. No 'insert' button present

**Expected behavior**
1. Install ZC without demo products or delete all existing manufacturers
2. Go to admin > catalog > manufacturers
3. 'insert' button should be present

Bear with me and I will see if I can resolve this,

simon1066:
Fixes #5699
(No insert option in admin when manufacturers list is empty)

### DIFF
--- a/admin/manufacturers.php
+++ b/admin/manufacturers.php
@@ -328,10 +328,10 @@ if (!empty($action)) {
       </div>
        <?php
       if (empty($action)) {
-        $current_mid = (isset($mInfo)) ? 'mID=' . $mInfo->manufacturers_id : '';
+        $current_mid = (isset($mInfo)) ? 'mID=' . $mInfo->manufacturers_id . '&' : '';
      ?>
         <div class="col-sm-12 text-right">
-          <a href="<?php echo zen_href_link(FILENAME_MANUFACTURERS, ($currentPage != 0 ? 'page=' . $currentPage . '&' : '') . $current_mid . ($current_mid != '' ? '&' : '') . 'action=new'); ?>" class="btn btn-primary" role="button"><?php echo IMAGE_INSERT; ?></a>
+          <a href="<?php echo zen_href_link(FILENAME_MANUFACTURERS, ($currentPage != 0 ? 'page=' . $currentPage . '&' : '') . $current_mid . 'action=new'); ?>" class="btn btn-primary" role="button"><?php echo IMAGE_INSERT; ?></a>
         </div>
       <?php } ?>
       <div class="row">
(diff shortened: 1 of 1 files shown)


admin/manufacturers.php
```diff
@@ -326,9 +326,12 @@
...
         <!-- body_text_eof //-->
       </div>
-      <?php if (empty($action) && !empty($mInfo)) { ?>
+       <?php
+      if (empty($action)) {
+        $current_mid = (isset($mInfo)) ? 'mID=' . $mInfo->manufacturers_id : '';
+     ?>
         <div class="col-sm-12 text-right">
-          <a href="<?php echo zen_href_link(FILENAME_MANUFACTURERS, ($currentPage != 0 ? 'page=' . $currentPage . '&' : '') . 'mID=' . $mInfo->manufacturers_id . '&action=new'); ?>" class="btn btn-primary" role="button"><?php echo IMAGE_INSERT; ?></a>
+          <a href="<?php echo zen_href_link(FILENAME_MANUFACTURERS, ($currentPage != 0 ? 'page=' . $currentPage . '&' : '') . $current_mid . '&action=new'); ?>" class="btn btn-primary" role="button"><?php echo IMAGE_INSERT; ?></a>
```
drbyte, line 334: You'll need to move the `&` out of `'&action=new'`, into `$current_mid`, else you'll have double `&&` in some cases and a missing one in other cases.

simon1066: Thanks. So I can better understand why this is needed, and just how to implement it, I'll do some research for my own benefit and submit again once I have it figured out.

drbyte: I've updated to show better what I meant.

Here's how it works:
- note how the zen_href_link ends with checking and setting the page param with a trailing `&`
- now I've added to the end of your `$current_mid` assignment an `&` if/when it gets set
- and now the `action=new` param doesn't get a double-`&&` by mistake

PR #5706, opened 2023-04-16
Coupon admin: html

torvista:
form name and input field name 'status' duplicated

labels for input fields become invalid when there are multiple (language fields).
I don't see any difference changing them from label to div.


drbyte:
> labels for input fields become invalid when there are multiple (language fields). 
> I don't see any difference changing them from label to div.

By removing the `<label>` element, which is used to display the field "title", the title/label becomes no-longer clickable-to-put-the-cursor-into-the-input-field. 
Admittedly in a multi-lang situation that has a small diminishing value of returns.

torvista:
Yes for single language.
But invalid html for multiple language.

This thing of having to substitute the label of a single field, for similarly-formatted text for multiple fields has been done elsewhere for the same reason, ie the product edit.

drbyte:
> labels for input fields become invalid when there are multiple (language fields).

They're unique per language, so it's not from duplication.
Can you post what the validator is saying? 

Also, `coupon_free_ship` and `coupon_amount` are not multi-language.

torvista:
The "label" can't have a for attribute that is valid for multiple inputs/with different names.

2023-04-29 10_01_14-Admin Coupon Admin — Mozilla Firefox

>Also, coupon_free_ship and coupon_amount are not multi-language.

Yes, I don't know what I was (not) thinking there...too much x and not enough y. Corrected.

drbyte:
Valid HTML is important, of course. 
So is usability. 
What's WCAG say regarding best practices for labels and inputs?

What if instead of removing labels, the label was set to the same as the first multi lang input field name?

torvista:
This, and four glasses of wine, is making my head spin, 
https://stackoverflow.com/questions/9004307/two-input-fields-inside-one-label

drbyte:
I think the simplest solution is just to set the label's `for` element to match the `id` that the first language field uses. More specifically, instead of `foo[]` it would be `foo[1]` (where `1` is the first language number in the loop of retrieved active languages).

drbyte:
👍

torvista:
and that validates, as I am sure you already knew!

scottcwilson:
Thanks guys.

PR #5709, opened 2023-04-18
Authorize.net/AIM, correcting order-status-updates for void/auth/capt

### LINKED ISSUE #5708: More authorizenet_aim.php issues; this time with default order status IDs
zc158a, although these issues have been around since _dinosaurs walked the earth_:

1. If `authorizenet_aim` has its `MODULE_PAYMENT_AUTHORIZENET_AIM_REFUNDED_ORDER_STATUS_ID` set to 'default', the order's status (and the associated status-history record) winds up set to 0.  That subsequently results in various PHP warnings when displaying the site's orders-listing containing the mangled order or viewing that order individually.
2. If any of the module's` _ORDER_STATUS_ID` values are set to '0' (i.e. use the site's default order-status), the module assigns a fixed value of 1 (for the default Zen Cart installation, that's "Pending") ... which might not be the site's default.
3. For whatever reason, on the Refund, Capture or Void methods, a customer-viewable order-status-history record is created, but it's not emailed to the customer.  That results in additional back-end work for the site since the customer's not notified via email of the update.

lat9:
Fixes #5708.

### DIFF
--- a/includes/modules/payment/authorizenet_aim.php
+++ b/includes/modules/payment/authorizenet_aim.php
@@ -780,182 +780,199 @@ class authorizenet_aim extends base {
       $db->Execute("ALTER TABLE " . TABLE_AUTHORIZENET . " MODIFY transaction_id varchar(64) default NULL");
     }
   }
- /**
-   * Used to submit a refund for a given transaction.
-   */
-  function _doRefund($oID, $amount = 0) {
-    global $messageStack;
-    $new_order_status = (int)MODULE_PAYMENT_AUTHORIZENET_AIM_REFUNDED_ORDER_STATUS_ID;
-    if ($new_order_status == 0) $new_order_status = 1;
-    $proceedToRefund = true;
-    $refundNote = strip_tags(zen_db_input($_POST['refnote']));
-    if (isset($_POST['refconfirm']) && $_POST['refconfirm'] != 'on') {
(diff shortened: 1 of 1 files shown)


drbyte:
```diff
-             $new_order_status = 1;
+             $new_order_status = (int)DEFAULT_ORDERS_STATUS_ID;
```

Just asking: if `DEFAULT_ORDERS_STATUS_ID` is "paid" (some equivalent thereof), and you refund it, changing it from `1=pending` to some "default" paid-variant is gonna be ... well, confusing.

lat9:
> ```diff
> -             $new_order_status = 1;
> +             $new_order_status = (int)DEFAULT_ORDERS_STATUS_ID;
> ```
> 
> Just asking: if `DEFAULT_ORDERS_STATUS_ID` is "paid" (some equivalent thereof), and you refund it, changing it from `1=pending` to some "default" paid-variant is gonna be ... well, confusing.

It can be an issue regardless the `DEFAULT_ORDERS_STATUS_ID` (which defaults to "Pending").  Since this payment method has no (current) concept of a full vs. partial refund, an order could go from Pending (if the default's the initial status) to Shipped and then back to Pending on any type of refund.  If a site does a fair number of refunds, its setting for these void/capture/refund actions' status should be more specific.

PR #5710, opened 2023-04-19
Fix for #5672-Shipping modules with Zone restrictions won't show up in Estimator if State not shown as pulldown.

### LINKED ISSUE #5672: Shipping modules wi/Zone restrictions won't show up in Estimator if State not shown as pulldown
When

a) Admin > Configuration > Customer Details > State field - Display as pulldown when possible? is set to FALSE
and
b) the zone restriction for a shipping module is set,

that module will not show up on the shipping estimator for non-logged in customers even if they have entered a country and state within the configured zone.

I'm not positive this is worth fixing - it would be a bunch of work for a configuration most people don't use. (It's much better to set Display as pulldown to true).

piloujp:
Fix Shipping modules wi/Zone restrictions won't show up in Estimator if State not shown as pulldown.

Fixes #5672.


scottcwilson:
Can you squash this change?

piloujp:
I think I just did...

scottcwilson:
Would you please close this PR and create a new one?  Sorry, I thought the squash would get rid of the un-needed files.

scottcwilson:
Replaced by #5712.

PR #5711, opened 2023-04-20
States pull down menu error when creating account

### LINKED ISSUE #5707: States pull down menu error when creating account
**Describe the bug**
With a Zen Cart fresh install v1.5.8, v1.5.8a and probably lower, when submitting create account form message  'Please select a state from the States pull down menu.' appear although state field (input text) has been properly filled.

**Version**
Zen Cart version: 1.5.8a
PHP version: 8.2
MySQL version: 8.032
Plugins that may be related: none
What older/newer versions were also tested? 1.5.8

**To Reproduce**
With default install (option Configuration->Customers Details->State field - Display as pulldown when possible? set to false).
1. Go to 'login/Create account' page
2. Fill the form setting country to 'United States', or any country that has zones defined, and submit it.

**Expected behavior**
Create a new account and log in.

**Screenshots**
If applicable, add screenshots to help explain.

**Debug/Error Logs**
 'Please select a state from the States pull down menu.' pop up message.
 
**Additional context**
If admin setting 'Configuration->Customers Details->State field - Display as pulldown when possible?' is set to true or if chosen country has no zone defined, problem disappears, form is accepted normally as a dropdown option has bee chosen. But when country has zones defined but input text is used in form them message popup when submitting and you can't create an account.
Same problem occurs when adding a new address at checkout or in 'My Account->address book', files 'create_account.php' and 'checkout_new_address.php'.

**Fix**
Changing line 201 in 'includes/modules/create_account.php' from:
if ($entry_state_has_zones == true) {
to
if ($entry_state_has_zones == true && ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN === 'true') {

And Changing line 82 in 'includes/modules/checkout_new_address.php' from:
if ($entry_state_has_zones == true) {
to
if ($entry_state_has_zones == true && ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN === 'true') {
will fix the problem.

piloujp:
Fixes #5707
When submitting create account form, message 'Please select a state from the States pull down menu.' appears although state field (input text) has been properly filled. Only when 'Configuration>Customers Details>State field - Display as pulldown when possible?' is set to false and has zones defined.

### DIFF
--- a/includes/modules/checkout_new_address.php
+++ b/includes/modules/checkout_new_address.php
@@ -79,7 +79,7 @@ if (isset($_POST['action']) && ($_POST['action'] == 'submit')) {
       $check_query = $db->bindVars($check_query, ':zoneCountryID', $country, 'integer');
       $check = $db->Execute($check_query);
       $entry_state_has_zones = ($check->fields['total'] > 0);
-      if ($entry_state_has_zones == true) {
+      if ($entry_state_has_zones == true && ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN === 'true') {
       $zone_query = "SELECT distinct zone_id, zone_name, zone_code
                        FROM " . TABLE_ZONES . "
                        WHERE zone_country_id = :zoneCountryID
--- a/includes/modules/create_account.php
+++ b/includes/modules/create_account.php
@@ -198,7 +198,7 @@ if (isset($_POST['action']) && ($_POST['action'] == 'process') && !isset($login_
         $check_query = $db->bindVars($check_query, ':zoneCountryID', $country, 'integer');
         $check = $db->Execute($check_query);
         $entry_state_has_zones = ($check->fields['total'] > 0);
-        if ($entry_state_has_zones == true) {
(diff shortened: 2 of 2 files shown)


piloujp:
Hello everybody,
I would like to attract more attention to this PR that seems to have been forgotten... The bug that is fixed here is what I call a 'business killer', It sends new customers away when they try to create a new account by offering them an infinite loop!
Of course there are necessary conditions for this to happen, but these conditions are there by default on a newly installed cart, at least for customers that live in any country which has zones defined by default, United States included.
The only thing that limits damages is probably the fact that many store owners set the pull down menu option to 'on'.
For those who don't they just loose clients without even noticing it.
I don't know how review and merge flow work but it is two slow for this kind of bug that is critical to Zen Cart function and which fix is relatively simple as it is just adding a condition on an IF test.
Actually I think there should be some kind of 'sub-released' between Zen Cart versions for bugs like this and security ones.

scottcwilson:
Thanks for doing this. Nothing gets forgotten, it's just that all the core committers are busy.  Please be patient.

piloujp:
> Thanks for doing this. Nothing gets forgotten, it's just that all the core committers are busy. Please be patient.

I'll try but it has never been one of my qualities... ;))

PR #5720, opened 2023-04-28
Redirect page updates and corrections

### LINKED ISSUE #5719: Correct `redirect` page redirects and add notifications
While the majority of the 'redirects' handled by the page's `header_php.php` issue a notification prior to the subsequent redirect, a banner click doesn't (i.e. an action of `banner`).

Since this is a significant 'measuring' event, such a notification _should_ be issued.

lat9:
Fixes #5719

### DIFF
--- a/includes/modules/pages/redirect/header_php.php
+++ b/includes/modules/pages/redirect/header_php.php
@@ -12,6 +12,19 @@ if (!defined('IS_ADMIN_FLAG')) {
 }
 
 $_GET['action'] = $_GET['action'] ?? '';
+
+// -----
+// Grab the currently-configured **default** language's id value.  Used
+// to determine if an additional query is needed when a language-specific URL
+// isn't found for the session's current language.
+//
+$configured_languages = new language();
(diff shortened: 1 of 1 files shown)


drbyte:
Maybe I'm overthinking. But since it's new functionality, is there any value in passing a returnable `$bypass` variable to these notifiers that can be checked whether to skip the counter-incrementing and redirect?

lat9:
> Maybe I'm overthinking. But since it's new functionality, is there any value in passing a returnable `$bypass` variable to these notifiers that can be checked whether to skip the counter-incrementing and redirect?

Ah, so you're joining me in the `over-thinking` club?  My initial take is that if an observer wanted to do that bypass, it could simply do its own `zen_redirect`.

drbyte:
> My initial take is that if an observer wanted to do that bypass, it could simply do its own zen_redirect.

Ya, that's probably enough. 
It's not going to trigger the "page is redirecting too many times" because it's internal anyway.

PR #5728, opened 2023-05-09
Add a notification for zen_get_products_special_price

lat9:
Add a notification for the subject function, enabling a site to override the default salemaker sale behavior, e.g. enabling the site to include products linked to a sale category.

Also restores the commentary that describes what the 'sale_deduction_type' values actually mean.

### DIFF
--- a/includes/functions/functions_prices.php
+++ b/includes/functions/functions_prices.php
@@ -52,6 +52,21 @@ function zen_get_products_special_price($product_id, $specials_price_only = fals
     $category = $product->fields['master_categories_id'];
     $sale = zen_get_sale_for_category_and_price($category, $product_price);
 
+    // -----
+    // Give an observer the opportunity to add functionality for the salemaker sales, 
+    // perhaps enabling linked products to be included.
+    //
+    // If the 'sale' price is to be overridden, an observer sets the $sale variable
+    // to an associative array with keys sale_specials_condition, sale_deduction_value and
+    // sale_deduction_type to match the format returned from a query of the
+    // salemaker_sales database table.
+    //
+    // If the observer wishes to negate a sale of the current product, the observer
+    // sets the $sale to contain (bool)false.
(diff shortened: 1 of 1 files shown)


lat9:
OK, so the question of the day (mine, anyways):  Why the heck does the admin not share `functions_prices.php` with the storefront?

marco-pm:
> OK, so the question of the day (mine, anyways): Why the heck does the admin not share `functions_prices.php` with the storefront?

Isn't that already the case for ZC v158?

lat9:
> Isn't that already the case for ZC v158?

Unfortunately, no.  It's one of the few function-files that isn't shared.  That (and for whatever reason) `html_output.php`.

marco-pm:
> Unfortunately, no. It's one of the few function-files that isn't shared.

https://github.com/zencart/zencart/blob/078b509bb514ab344a034e89508eb02f1e63854c/admin/includes/functions/functions_prices.php

🤨

lat9:
OK, so I'm having a bad day/week/month ...

scottcwilson:
> Why the heck does the admin not share functions_prices.php with the storefront?

Fair question.  In at least some cases, there is enough difference between admin/storefront function implementations that it was too just much work (at least at the time, given other commitments).  

PRs would be welcome for either or both.   Maybe @marco-pm needs a project! :)

lat9:
> > Why the heck does the admin not share functions_prices.php with the storefront?
> 
> Fair question. In at least some cases, there is enough difference between admin/storefront function implementations that it was too just much work (at least at the time, given other commitments).
> 
> PRs would be welcome for either or both. Maybe @marco-pm needs a project! :)

Turns out, per @marco-pm's reply that the module in the admin is a "remove me" stub.

scottcwilson:
Yes for functions_prices.php, but no for html_output.php.

PR #5729, opened 2023-05-10
Remove 'selected_box' references

Zen4All:
This seems to obsolete code, that has been there since the beginning of Zen Cart.

### DIFF
--- a/admin/includes/modules/dashboard_widgets/OrderStatusDashboardWidget.php
+++ b/admin/includes/modules/dashboard_widgets/OrderStatusDashboardWidget.php
@@ -21,7 +21,7 @@ if (!zen_is_superuser() && !check_page(FILENAME_ORDERS, '')) return;
           $orders_pending = $db->Execute("SELECT count(*) as count FROM " . TABLE_ORDERS . " WHERE orders_status = " . (int)$row['orders_status_id'], false, true, 1800);
           ?>
         <tr>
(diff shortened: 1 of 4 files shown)


admin/includes/modules/dashboard_widgets/OrderStatusDashboardWidget.php
```diff
@@ -21,7 +21,7 @@
           $orders_pending = $db->Execute("SELECT count(*) as count FROM " . TABLE_ORDERS . " WHERE orders_status = " . (int)$row['orders_status_id'], false, true, 1800);
           ?>
         <tr>
-          <td><a href="<?php echo zen_href_link(FILENAME_ORDERS, 'selected_box=customers&statusFilterSelect=' . $row['orders_status_id']); ?>"><?php echo $row['orders_status_name']; ?></a>:</td>
+          <td><a href="<?php echo zen_href_link(FILENAME_ORDERS, 'customers&statusFilterSelect=' . $row['orders_status_id']); ?>"><?php echo $row['orders_status_name']; ?></a>:</td>
```
lat9, line 24: Shouldn't the `customers&` portion be removed as well?

lat9 (changes requested):

admin/includes/modules/dashboard_widgets/OrderStatusDashboardWidget.php
Zen4All: You are right, I'll update the PR

lat9 (approved):

PR #5744, opened 2023-06-03
SaleMaker: fix issue with sale turning on and immediately turning off.

boomer196:
Additions to the changes @scottcwilson made in #5726.  Issue is that it turns them "on", but because the expires select query was not changed as well, it will immediately turn them "off".  

To Recreate using `Zencart 1.5.8a` with `msql 8.0`:
* Set up a sale to start today and end in a few days.
* Sale turns on https://github.com/zencart/zencart/blob/9ccb250afda600b06c3446b50b992406f6ae661f/includes/functions/salemaker.php#L81
* Sale would immediately turn off
https://github.com/zencart/zencart/blob/9ccb250afda600b06c3446b50b992406f6ae661f/includes/functions/salemaker.php#L95

### DIFF
--- a/includes/functions/salemaker.php
+++ b/includes/functions/salemaker.php
@@ -57,7 +57,7 @@ function zen_start_salemaker()
 {
     global $db;
 
-    $sale_date = date('Ymd', time());
+    $sale_date = date('Y-m-d', time());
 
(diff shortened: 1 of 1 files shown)


scottcwilson:
Thanks for catching this.

PR #5747, opened 2023-06-07
Fix admin prod/cat search with missing whitespace or empty string - Fixes #5715, #5666

### LINKED ISSUE #5666: admin, category-product listing, search with no keyword: debug logs for undefined variable error
> [25-Mar-2023 11:42:21 UTC] Request URI: /zencart/admin/index.php?cmd=category_product_listing&search=, IP address: 127.0.0.1, Language id 1
> #1 D:\GitHub\zencart\includes\functions\functions_search.php(211): substr()
> #2 D:\GitHub\zencart\admin\category_product_listing.php(545): zen_build_keyword_where_clause()
> #3 D:\GitHub\zencart\admin\index.php(11): require('D:\\GitHub\\zenca...')
> --> PHP Deprecated: substr(): Passing null to parameter #1 ($string) of type string is deprecated in D:\GitHub\zencart\includes\functions\functions_search.php on line 211.

> [25-Mar-2023 11:42:21 UTC] Request URI: /zencart/admin/index.php?cmd=category_product_listing&search=, IP address: 127.0.0.1, Language id 1
> #0 D:\GitHub\zencart\includes\functions\functions_search.php(211): zen_debug_error_handler()
> #1 D:\GitHub\zencart\admin\category_product_listing.php(545): zen_build_keyword_where_clause()
> #2 D:\GitHub\zencart\admin\index.php(11): require('D:\\GitHub\\zenca...')
> --> PHP Warning: Undefined variable $where_str in D:\GitHub\zencart\includes\functions\functions_search.php on line 211.

### LINKED ISSUE #5715: Admin product search: keywords with parentheses returns many results/memory error
php 8.2.5

search for
"stuff(" : ok/no results
"stuff(more" : ok/no results
"stuff(more)" : not ok/returns many results or out of memory error.

Maybe the eager beaver who wrestles this one into submission can do two for the price of one: https://github.com/zencart/zencart/issues/5666

Debugs:

[25-Apr-2023 15:08:42 UTC] Request URI: /zencart/admin/index.php?cmd=category_product_listing&search=stuff%28more%29, IP address: 127.0.0.1, Language id 1
#1 D:\Github\zencart\includes\functions\functions_search.php(211): substr()
#2 D:\Github\zencart\admin\category_product_listing.php(545): zen_build_keyword_where_clause()
#3 D:\Github\zencart\admin\index.php(11): require('D:\\Github\\zenca...')
--> PHP Deprecated: substr(): Passing null to parameter #1 ($string) of type string is deprecated in D:\Github\zencart\includes\functions\functions_search.php on line 211.

[25-Apr-2023 15:08:42 UTC] Request URI: /zencart/admin/index.php?cmd=category_product_listing&search=stuff%28more%29, IP address: 127.0.0.1, Language id 1
#1 D:\Github\zencart\includes\functions\functions_search.php(211): substr()
#2 D:\Github\zencart\admin\category_product_listing.php(802): zen_build_keyword_where_clause()
#3 D:\Github\zencart\admin\index.php(11): require('D:\\Github\\zenca...')
--> PHP Deprecated: substr(): Passing null to parameter #1 ($string) of type string is deprecated in D:\Github\zencart\includes\functions\functions_search.php on line 211.

and

[25-Apr-2023 15:08:42 UTC] Request URI: /zencart/admin/index.php?cmd=category_product_listing&search=stuff%28more%29, IP address: 127.0.0.1, Language id 1
#0 D:\Github\zencart\includes\functions\functions_search.php(211): zen_debug_error_handler()
#1 D:\Github\zencart\admin\category_product_listing.php(545): zen_build_keyword_where_clause()
#2 D:\Github\zencart\admin\index.php(11): require('D:\\Github\\zenca...')
--> PHP Warning: Undefined variable $where_str in D:\Github\zencart\includes\functions\functions_search.php on line 211.

[25-Apr-2023 15:08:42 UTC] Request URI: /zencart/admin/index.php?cmd=category_product_listing&search=stuff%28more%29, IP address: 127.0.0.1, Language id 1
#0 D:\Github\zencart\includes\functions\functions_search.php(211): zen_debug_error_handler()
#1 D:\Github\zencart\admin\category_product_listing.php(802): zen_build_keyword_where_clause()
#2 D:\Github\zencart\admin\index.php(11): require('D:\\Github\\zenca...')
--> PHP Warning: Undefined variable $where_str in D:\Github\zencart\includes\functions\functions_search.php on line 211.

neekfenwick:
Regarding #5715, the search parsing code tries to split around whitespace, and then recognise grouped logical terms, e.g. `foo (wibble and bar)` but fails when there is missing whitespace e.g. `foo(wibble and bar)`.

I've taken a very simple approach of using regex replace to separate words from parentheses before the parsing begins, so the search string is turned from `foo(wibble and bar)` into `foo (wibble and bar)` and the splitting and query building proceeds as normal.

Also return early (after notifier is called in case of required modification) when search string is empty to fix #5666.

### DIFF
--- a/includes/functions/functions_search.php
+++ b/includes/functions/functions_search.php
@@ -162,9 +162,6 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
         global $db, $zco_notifier;
 
         $zco_notifier->notify('NOTIFY_BUILD_KEYWORD_SEARCH', '', $fields, $string);
-        if (empty($string)) {
-            return '';
-        }
         $where_str = '';
         if (zen_parse_search_string(stripslashes($string), $search_keywords)) {
             $where_str = " AND (";
--- a/includes/functions/functions_search.php
+++ b/includes/functions/functions_search.php
@@ -11,6 +11,9 @@
 function zen_parse_search_string($search_str = '', &$objects = array()) {
     $search_str = trim(strtolower($search_str));
 
+// Separate any parentheses from adjacent words so splitting works as intended
+    $search_str = preg_replace([ '/(\b)(\()/', '/(\))(\b)/' ], [ '$1 $2', '$1 $2' ], $search_str);
+
 // Break up $search_str on whitespace; quoted string will be reconstructed later
     $pieces = preg_split('/[[:space:]]+/', $search_str);
     $objects = array();
@@ -162,6 +165,7 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
         global $db, $zco_notifier;
 
         $zco_notifier->notify('NOTIFY_BUILD_KEYWORD_SEARCH', '', $fields, $string);
+        if (empty($string)) { return ''; }
         if (zen_parse_search_string(stripslashes($string), $search_keywords)) {
             $where_str = " AND (";
             if ($startWithWhere) {
--- a/includes/functions/functions_search.php
+++ b/includes/functions/functions_search.php
@@ -162,7 +162,9 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
         global $db, $zco_notifier;
 
         $zco_notifier->notify('NOTIFY_BUILD_KEYWORD_SEARCH', '', $fields, $string);
-        if (empty($string)) { return ''; }
+        if (empty($string)) {
+            return '';
+        }
         $where_str = '';
         if (zen_parse_search_string(stripslashes($string), $search_keywords)) {
             $where_str = " AND (";
--- a/includes/functions/functions_search.php
+++ b/includes/functions/functions_search.php
@@ -11,9 +11,6 @@
 function zen_parse_search_string($search_str = '', &$objects = array()) {
     $search_str = trim(strtolower($search_str));
 
-// Separate any parentheses from adjacent words so splitting works as intended
-    $search_str = preg_replace([ '/(\b)(\()/', '/(\))(\b)/' ], [ '$1 $2', '$1 $2' ], $search_str);
-
 // Break up $search_str on whitespace; quoted string will be reconstructed later
     $pieces = preg_split('/[[:space:]]+/', $search_str);
     $objects = array();
@@ -21,7 +18,7 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
     $flag = '';
 
     for ($k=0; $k<count($pieces); $k++) {
-        while (substr($pieces[$k], 0, 1) == '(') {
+        while (substr($pieces[$k], 0, 1) == '(' && strpos($pieces[$k], ')', 1) == false) {
             $objects[] = '(';
             if (strlen($pieces[$k]) > 1) {
                 $pieces[$k] = substr($pieces[$k], 1);
@@ -32,7 +29,7 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
 
         $post_objects = array();
 
-        while (substr($pieces[$k], -1) == ')')  {
+        while (substr($pieces[$k], -1) == ')' && strpos($pieces[$k], '(') == false)  {
             $post_objects[] = ')';
             if (strlen($pieces[$k]) > 1) {
                 $pieces[$k] = substr($pieces[$k], 0, -1);
--- a/includes/functions/functions_search.php
+++ b/includes/functions/functions_search.php
@@ -163,6 +163,7 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
 
         $zco_notifier->notify('NOTIFY_BUILD_KEYWORD_SEARCH', '', $fields, $string);
         if (empty($string)) { return ''; }
+        $where_str = '';
         if (zen_parse_search_string(stripslashes($string), $search_keywords)) {
             $where_str = " AND (";
             if ($startWithWhere) {
@@ -212,5 +213,5 @@ function zen_parse_search_string($search_str = '', &$objects = array()) {
         if (substr($where_str, -7) === '( ()  )') {
             return ' ';
         }
-        return $where_str ?? ' ';
+        return $where_str;
     }
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -549,7 +549,6 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
           }
 
           $sql .= $order_by;
-          echo "DEBUG: $sql<br><br>";
 
           $categories = $db->Execute($sql);
 
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -536,7 +536,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                   LEFT JOIN " . TABLE_CATEGORIES_DESCRIPTION . " cd ON c.categories_id = cd.categories_id
                     AND cd.language_id = " . (int)$_SESSION['languages_id'];
 
-            if (isset($_GET['search'])) {
+          if (isset($_GET['search']) && !empty($_GET['search'])) {
               $keyword_search_fields = [
                 'cd.categories_name',
                 'cd.categories_description',
@@ -549,6 +549,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
           }
 
           $sql .= $order_by;
+          echo "DEBUG: $sql<br><br>";
 
           $categories = $db->Execute($sql);
 


drbyte:
This covers the use-case you describe, but does it break the ability to search for situations where parentheses are supposed to be present without white space? Like a product code or something?

torvista:
Thanks for making the effort to fix this.
Just tried it from the admin. 

With an empty search string it returns a vast amount of results. Would it not be better to just make the browser/js prevent any search when the field is empty?

Using "stuff (and more stuff)" I get the same debugs as here: https://github.com/zencart/zencart/issues/5715#issue-1683334688

neekfenwick:
> stuff (and more stuff)"

You've hit a weird case because the 'and' in your string is treated as a boolean operator when it's actually a search term.  It's expanded to include logical operators to give the array `[ 'stuff', 'and', '(', 'and', 'more', 'and', 'stuff', ')' ]` when it should really be `[ 'stuff', 'and', '(', 'and', 'and', 'more', 'and', 'stuff', ')' ]` but the code doesn't have a way to distinguish between an 'and' that's a search term and 'and' that's a logical operator.

- change the internal operators from 'and' and 'or' to something more unlikely such as '#and#', '#or#'.  This would prevent the search term 'and' from colliding with them.  This would mean a `configuration` change for `ADVANCED_SEARCH_DEFAULT_OPERATOR` which is currently `and`.
- we allow a search string like this to be considered malformed, fail parsing, and result in no meaningful search being done.

WWTD?

In either case I'll submit a new PR because as you say that second case, where the search string was considered malformed, does still trigger an `Undefined variable` warning.

includes/functions/functions_search.php
```diff
@@ -162,6 +162,8 @@ function zen_build_keyword_where_clause($fields, $string, $startWithWhere = fals
         global $db, $zco_notifier;
 
         $zco_notifier->notify('NOTIFY_BUILD_KEYWORD_SEARCH', '', $fields, $string);
+        if (empty($string)) { return ''; }
```
proseLA, line 165: please use the PSR-12 coding style.

scottcwilson:
I wonder if the support for logical operators (includes/functions/functions_search.php line 172-176, 128-133) is worth the trouble.  What if we just removed this?  How many people actually need it?

proseLA:
> I wonder if the support for logical operators (includes/functions/functions_search.php line 172-176, 128-133) is worth the trouble. What if we just removed this? How many people actually need it?



> I wonder if the support for logical operators (includes/functions/functions_search.php line 172-176, 128-133) is worth the trouble. What if we just removed this? How many people actually need it?

i am not a fan of removing the functionality; especially since it is already out in the wild.

i think the blank search terms is easily fixed; i'm also of the opinion as to how many people actually search with parenthesis?  is this another YAGNI?

i would not be opposed to just stripping the parenthesis from the keyword search POST; and in addition, utilizing a placeholder to provide instructions for the search box.  from what i remember, i think after numerous iterations, the search box is a common bit of code.

actually it is used for search form pulldowns:

https://github.com/zencart/zencart/blob/df819c7da9b3d30e0d64daabfde7aeb1f8be47e5/admin/includes/functions/html_output.php#L520

but i still like the idea of placeholder text that provides search instructions as opposed to removing logic.

includes/functions/functions_search.php
neekfenwick: No problem, I presume you mean to break it over three lines.

neekfenwick:
> i would not be opposed to just stripping the parenthesis from the keyword search POST

This would mean you couldn't search for a name or model that actually has parentheses in it, like drbyte's first reply here.  My understanding of that requirement is that if the user wants to search for `thingy(a64)` or whatever then they should be able to do so.  My PR supports this.

Placeholder text for the default `tpl_search_header` input is already editable as lang string `HEADER_SEARCH_DEFAULT_TEXT`, the input is fairly wide on the admin categories page so we could fit some reasonable instructions in there.

@scottcwilson I've absolutely no idea if anyone ever uses it but I feel if it's not actively breaking typical use cases we may as well keep it.

neekfenwick:
> With an empty search string it returns a vast amount of results. Would it not be better to just make the browser/js prevent any search when the field is empty?

Good point. I just pushed an improvement to not bother parsing the search string at all if it's empty, so you get a typical top level cat listing instead of a half arsed attempt at search.  This also removes the need to check for empty within `zen_build_keyword_where_clause`.

includes/functions/functions_search.php
proseLA: coding style has been a hotly debated topic in the past.  what i mean is here:

https://www.php-fig.org/psr/psr-12/

yes, it should be over 3 lines; but i would guess that your editor may be able to help you with this.

proseLA:
@neekfenwick there are times when i think this project solely exists for torvista and his model numbers.  these edge case scenarios seem absurd to me; but this is merely my opinion.  will someone please provide me a manufacturers list (from the manufacturer) where there is a parenthesis in the model number?  i would like to see it.

that said, lets take a model number of `123(456` as an example.  the enhanced searching in v158, which i coded (with drBytes supervision some 2+ years ago) is SO far superior to what was previously done.  it is NOT intended to require exact matches, but complex boolean matches.

if we were to take that model number, replace the `(` with a space, would not the required product show up in the search results?  is that not what we are after?

my opinion is to replace all of the `(` and `)` with spaces and the results will still be clear.  assuming someone is not using an 'OR` search in which case the result set may be a bit larger but the searched for product will still be there.

lets apply some common sense here.  and again, i would like to see a manufacturer's list that has a parenthesis in the model number.  just for my own education, and to know to never get into moto racing...

proseLA:
> > With an empty search string it returns a vast amount of results. Would it not be better to just make the browser/js prevent any search when the field is empty?
> 
> Good point. I just pushed an improvement to not bother parsing the search string at all if it's empty, so you get a typical top level cat listing instead of a half arsed attempt at search. This also removes the need to check for empty within `zen_build_keyword_where_clause`. A browser/js approach seems unnecessary and fiddly, they might be on a page of results from a previous search and simply want to 'clear' the search by searching for an empty string, which this allows them to do.

your improvement has an `echo DEBUG` line in it.  i have found modern IDEs with xdebug enabled within them (or something of the same nature) to be incredibly useful and a real game changer as far as increasing one's productivity.  it also removes the need for these type of lines which somehow get all embedded in the code. 

part of ZC's limitations, again in my opinion, is not the full on embracing of JS.  client side validation and the newer frameworks that embrace them are far more powerful and provide a much better user experience, again in my opinion.  that said, best practices used to include the need to assume JS was not enabled which required PHP to still check for the same thing that the JS checked for.  whether that is still a best practice, i can not say; i have not looked.

and please do not take my opinion as a criticism of the work done by all of the contributors to this repo, especially drByte and zcwilt.  i have tremendous respect for both of them and all they have done to make ZC what it is today.  i have learned much from both of them.

and as far as your comment about `half-arsed`, i can without question confirm that all code has bugs, and as the person most responsible for `zen_build_keyword_search` that my arse is full and fully functioning...

neekfenwick:
> your improvement has an `echo DEBUG` line in it

OK, obvious oversight. You don't need to lecture on IDE's, this was simply a way to view the built SQL over many invocations without having to stop in the debugger and examine variables.  If we want to prevent such commits a pre-commit hook or similar is a good way to go, I'll add one on my end.

I thought my point of the user hitting Search with an empty box with the intent to cancel any existing search was valid and avoided any discussion of JS, which can be a real rabbit hole when we start discussing frameworks :)

My comment about something being half arsed was in no way directed at any of us, I was talking about the code attempting to build a WHERE clause with an empty search string.

neekfenwick:
> there are times when i think this project solely exists for torvista and his model numbers.

It was drbyte who mentioned searches with parentheses in, not torvista.  I do think some businesses will have brackets in their strings, especially those dealing in technical parts or embedded information like year number etc (imagine a "widget(1998)" vs "widget(2005)".  I can't provide immediate usage but it's certainly credible.

scottcwilson:
> ... I was talking about the code attempting to build a WHERE clause with an empty search string.

Changing admin/category_product_listing.php l. 539 to check for empty is a nice improvement.  Thanks for doing that. 

Re: debug statements 
> If we want to prevent such commits a pre-commit hook or similar is a good way to go, I'll add one on my end.

This is not a real problem.  One of the core committers will ask you to clean it up prior to merge.

torvista:
Sorry if my example is such a pita to handle, it was just intended to point out that whatever goes in that search field should not break anything or cause a debug and nothing else.
If it is so esoteric that it would never occur, then don't bother trying to get results for it.

I have not seen brackets in model numbers.

Yes I am forced to hinge everything on my site on model numbers, but Google wants them too, and GTIN, so I don't think it's my caprice.

Yes the common search form is great, much appreciated.

scottcwilson:
True story: A client wanted the model number extended to 255 characters so they could write a sentence.  "The one with the red thing on top ..."

neekfenwick:
> True story: A client wanted the model number extended to 255 characters so they could write a sentence. "The one with the red thing on top ..."

@scottcwilson :) Just an aside, I've extended the core schema too but only because our products have ridiculously long descriptors, like "Glass Splashback in Caribbean Jetty With Ocean View".  The default varchar(32) is very limiting, in the end I added a separate 'extra' table similar to `product_music_extra` so the details could be broken out into distinct fields.

@proseLA I think this is ready to merge, as it supports the requirement of embedded parentheses, and fixes the debug output caused by searching for partial parentheses and empty string.

piloujp:
I am a little bit late but I don't agree that this is an edge case, parantheses are very common in manufacturer listing.
A search in my shop database like this:
SELECT * FROM `zen_products_description` WHERE `products_name` LIKE '%(%';
Returns 1109 products out of 2320... And I am generally using manufacturer's description as product name.

neekfenwick:
> I am a little bit late but I don't agree that this is an edge case, parantheses are very common in manufacturer listing. A search in my shop database like this: SELECT * FROM `zen_products_description` WHERE `products_name` LIKE '%(%'; Returns 1109 products out of 2320... And I am generally using manufacturer's description as product name.

@piloujp  My PR supports searching for a term that contains matching parentheses, e.g. `widget(2005)`, what more do you want?

piloujp:
> @piloujp My PR supports searching for a term that contains matching parentheses, e.g. `widget(2005)`, what more do you want?

@neekfenwick Nothing, I was just trying to support you as there was arguments against your PR. ;)

proseLA:
i want to clarify a few things:

- i do not have merge power in this repo.
- i think no search should result in an error and a debug log.
- i currently do not have time to review this PR.
- i am not opposed to parenthesis in product name, description, etc.
- my opinion is that stripping parenthesis and replacing them with spaces from the posted keyword search terms will provide a more robust search utilizing the current code.
- it would be nice to see discussions that exponentially move search forward as opposed to these incremental bug discussions.
- i think a lot of the coding contained in [marco-pms instant search](https://github.com/marco-pm/zencart_instantsearch) would make such an exponential change.

best.

torvista:
Perhaps when this immediate issue is resolved, there could be an attempt to review/combine and merge to core Marcos' Instant Search and/or/with the Bootstrap template instant search?

neekfenwick:
> i want to clarify a few things:

@proseLA you are named as the 'reviewer' on this issue, not something I did, but I presumed from this you'd be merging it.  I just tried to fix this Issue, as @torvista said if you want to broaden the scope to incorporating this marco-pms thing by all means open a new Issue that aims to do this.  This PR was never intended to be 'exponential'

> i think no search should result in an error and a debug log.

I think this is wrong, if by 'no search' you mean hitting the search button with no text entered, why on earth would that create a debug log?  A good user experience is a reasonable outcome given their input, if they search for nothing they should get the normal result of the page when no search has been performed, which is what we get with my PR.

> i am not opposed to parenthesis in product name, description, etc.
> my opinion is that stripping parenthesis and replacing them with spaces from the posted keyword search terms will provide a more robust search utilizing the current code.

These statements seem to contradict one another.  If you allow parenthesis in name, description etc, but then strip parentheses from search terms, then the search will not match the name, description etc.

I have no idea now who is in charge and frankly and quite tired of discussing this relatively minor change.

proseLA:
> > i want to clarify a few things:
> 
> @proseLA you are named as the 'reviewer' on this issue, not something I did, but I presumed from this you'd be merging it. I just tried to fix this Issue, as @torvista said if you want to broaden the scope to incorporating this marco-pms thing by all means open a new Issue that aims to do this. This PR was never intended to be 'exponential'

- i am the one who made a review comment that stated you should follow PSR-12 coding style.  i "think" that is why github lists me as reviewer as i see no other reason why my name would be there.  i would suggest you stop presuming and understand who has merge access to this repo, i do not.

> 
> > i think no search should result in an error and a debug log.
> 
> I think this is wrong, if by 'no search' you mean hitting the search button with no text entered, why on earth would that create a debug log? A good user experience is a reasonable outcome given their input, if they search for nothing they should get the normal result of the page when no search has been performed, which is what we get with my PR.

- when i say no search should result in an error and a debug log; i meant that nothing that a user does within the context of searching should result in the page errorring out.  that clearly is not the case here.  you have clearly misunderstood what i said, as there is a problem that needs to get fixed.  and in looking at your code, you are only fixing it in the category_product_listing page; what happens with the other places where the method is called?  what happens when a developer decides to use this method elsewhere and accidentally passes an empty string?  why not fix it in the method?

 
> > i am not opposed to parenthesis in product name, description, etc.
> > my opinion is that stripping parenthesis and replacing them with spaces from the posted keyword search terms will provide a more robust search utilizing the current code.
> 
> These statements seem to contradict one another. If you allow parenthesis in name, description etc, but then strip parentheses from search terms, then the search will not match the name, description etc.

your statement above, at least to me, demonstrates you have no idea how the `zen_build_keyword_where_clause` method works.  if a user searchs for `(more languages)` and the only field being fed to the method in the field array was product description, the resulting where string was intended to be something like `where products_description like '%(more%' and products_description like '%languages)%'`.  unfortunately that is not working.  if the search string was stripped of the parenthesis and replaced with spaces, the new result of the method would be `where products_description like '%more%' and products_description like '%languages%'`.  

if you think the result set would not match, i would suggest you do a little more work on it.

torvista:
nothing that a user does within the context of the GUI should result in a debug

proseLA:
@neekfenwick similar to you, i am tired of this thread.  someone with merge power can decide what to do.  i plan on no longer commenting on this thread after this post.

i think literal and exact matches mean less when doing searching.  that is why i created this method.  a user is searching for a product; show said user the most relevant products to that search.  showing additional products, at least to me, is not a problem; as long as you include the product they are looking for.

reviewing this [post](https://www.zen-cart.com/showthread.php?229500-Admin-Discount-Coupons-Search&p=1395325#post1395325) shows that at least some people find this type of searching of value as they now want it in another areas in ZC; which was always my intention.

zcwilt:
Going to merge this. Maybe we need a more in depth discussion around Search/Advanced Search

PR #5751, opened 2023-06-16
Remove UPS and Fedex logs too

scottcwilson:


drbyte:
Argh, I'd meant to submit that as a review proposal you could merge if you agreed. But committed it directly to your branch instead. Apologies.

drbyte:
Q: is there another place where we have the same set of log patterns for delete? Maybe it's in `zc_install`?

scottcwilson:
> Q: is there another place where we have the same set of log patterns for delete? Maybe it's in `zc_install`?

zc_plugins/DisplayLogs/v3.0.2/Installer/ScriptedInstaller.php

Do you think these should be kept in sync?   If so please go ahead and do that - I'm fine with the change you made, it's easier to read.

drbyte:
I agree: probably wise to keep it in sync.  

No PHP changes required. Just need to update that configuration entry.

scottcwilson:
... but the config entry wouldn't be there if Display Logs is not installed.

drbyte:
> No PHP changes required. Just need to update that configuration entry.

Clarification: no programming logic changes required in that plugin for it to use the array that I added above.
But yes, it would be convenient for merchants if your new patterns were added to that plugin's installer.

scottcwilson:
I wonder instead if we should remove that config and give people the filename where the list is stored (we could put it in extra_definitions).  Then they could update both Store Manager and Display Logs according to their own needs.

drbyte:
That may be worth considering separately, for sure. Perhaps start an issue for that, and plugin collaborators can chime in.

drbyte:
If making it user-configurable, might wanna be sure to limit which things they can control vs are forced "anyway"

scottcwilson:
> If making it user-configurable, might wanna be sure to limit which things they can control vs are forced "anyway"

Not sure it's worth the effort.  Who's going to remove entries?   And even if they do they can still use FTP or File Manager to remove unwanted logs.

Anyhow, this has been updated; please merge or amend as you see fit.

PR #5758, opened 2023-06-21
Fix on-page links to use cmd= URL syntax - Fixed #5633

### LINKED ISSUE #5633: Admin, Who's Online, Refresh uses old page link
eg.:
On first page load:

>  .../zencart/admin/index.php?cmd=whos_online

On auto refresh for 15s: 

> .../zencart/admin/whos_online.php?t=15

I assume this should be now using the cmd parameter?

neekfenwick:
As described in Issue #5633 the on page links used the old admin/whos_online.php syntax.  In my case this broke javascript widgets in our common header on that page because the generated <html> element wasn't set up correctly.


drbyte:
Thanks!

PR #5760, opened 2023-06-23
refactoring test framework. Feature tests using symfony browserkit

zcwilt:
This is a major refactor of the test framework to allow Feature testing using symfony browserkit.
It also introduces the concepts of migrations and seeders (taken from Laravel)

It still needs some more work to add more tests and various helpers, especially around Admin Login

the idea here is to increase test code coverage to spot more bugs at the PR level and allow for better TDD


scottcwilson:
exciting stuff!

zcwilt:
Note: Haven't integrated with github CI yet. Will be testing that next.

zcwilt:
So getting the functional tests running turned out to be more of 'mare' than I thought 
Need to update this so that we are seeding with demo.sql data and then build out a few more tests.

currently i'm just running against 8.0, worry that running against a matrix of php versions will be super slow.

scottcwilson:
Can we make the checkin test only run the current PHP, but have nightly runs against older versions?

zcwilt:
@scottcwilson yes
github actions can be set to run on a cron schedule but have never tried this and not sure how it affects passing PR's

torvista:
I note the auto-testing done on a PR is from 7.4, but the server requirements allow 7.3
https://docs.zen-cart.com/user/first_steps/server_requirements/

zcwilt:
We haven't tested against 7.3 since mid 2022.

scottcwilson:
Yep, the intention of 7.3 support is just to get people past the initial install.  I really don't want anyone to spend a ton of time on it.

zcwilt:
Moved this back to DNR for now. 
Want to add the use of custom seeders for feature tests.

Issue #5554, opened 2023-01-25
'TEXT_GREETING_GUEST' missing parameter for sprintf


torvista:
https://github.com/zencart/zencart/blob/0c6e7c422cc7fe4b5252c310d62b90f22f12b206/includes/languages/english/lang.index.php#L3

https://github.com/zencart/zencart/blob/0c6e7c422cc7fe4b5252c310d62b90f22f12b206/includes/functions/functions_customers.php#L17

scottcwilson:
Let's simplify by removing the additional parm from `includes/functions/functions_customers.php`.

scottcwilson:
Would you please PR this @torvista ?

PR #5765, opened 2023-06-29
remove unused sprintf parameter from $greeting_string

torvista:
fixes: https://github.com/zencart/zencart/issues/5554


scottcwilson:
Great!

PR #5801, opened 2023-07-01
v158 ajax closure and refactor

### LINKED ISSUE #5800: Ajax closure incorrectly always uses catalog side
**Describe the bug**
Calls to ajax files will not call the admin side of `includes/application_bottom.php` although will call `includes/application_top.php`.

**Version**
Zen Cart version: 1.5.8+
PHP version: Compatible
MySQL version: Compatible
Plugins that may be related: Any that use it and expect to use the associated `application_bottom.php` file.
What older/newer versions were also tested? similar existed in 1.5.7.

mc12345678:
Call the `includes/application_bottom.php` file applicable to the side of the store used to open the ajax load.

Do away with nested if/elseif logic where return is used for true evaluation.

Eliminate single line if statements and add parentheses to support expanded logic grouping.

Eliminate the parentheses around the filename for calls of `require`.

Fixes #5800.

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -38,23 +38,26 @@ if (!function_exists('utf8_encode_recurse')) {
     {
         if (strtolower(CHARSET) == 'utf-8') {
             return $mixed_value;
-        } elseif (!is_array($mixed_value)) {
-            return utf8_encode((string)$mixed_value);
-        } else {
-            $result = array();
-            foreach ($mixed_value as $key => $value) {
-                $result[$key] = utf8_encode($value);
-            }
-            return $result;
         }
+        if (!is_array($mixed_value)) {
+            return utf8_encode((string)$mixed_value);
+        }
+        $result = array();
+        foreach ($mixed_value as $key => $value) {
+            $result[$key] = utf8_encode($value);
+        }
+        return $result;
     }
 }
 
 function ajaxAbort($status = 400, $msg = null)
 {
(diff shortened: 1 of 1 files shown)


scottcwilson:
This seems reasonable but is there any opportunity to consolidate the two copies of `ajax.php` instead?

mc12345678:
> This seems reasonable but is there any opportunity to consolidate the two copies of `ajax.php` instead?

Please take a moment and revisit this comment. It doesn't make sense when there is only one file called `ajax.php` that is in the root of the store.

scottcwilson:
There's another one called `admin/ajax.php`.

mc12345678:
> There's another one called `admin/ajax.php`.
> 
> Oh geez, all it does is include the base copy. For some reason I thought there was redundant code. OK TY.

Np. I mean the admin version *could* be done away with, but at a cost to programmers and potentially a portion of desired security. The admin folder name/path would need to be passed, likely in posted information, which just has its own issues. Other efforts may also be required. 

The current method (although yes there are two files) offers consistent developer code for the functionality offered on both sides. If/when all operations go to just a catalog side, this current usage will not be impacted.

scottcwilson:
No, no, I confused it with one of the old situations we had in prior releases where there was almost identical copy and pasted code from storefront placed in admin.  No need to do anything further.

PR #5832, opened 2023-07-01
moving filter classes

zcwilt:


zcwilt:
Should be noted, we are not using filters anywhere yet

PR #5835, opened 2023-07-02
Return blank if image is not specified

### LINKED ISSUE #5518: admin, category-product listing, product with no image references directory: 403 erro
create a new product with no details apart from name.
it has no image in the listing:

image

but the src references the directory instead of a missing image:

image

which is not allowed:

image

scottcwilson:
Fixes #5518

Instead of returning the images folder as the URL within an image tag, it returns blank.

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -130,6 +130,9 @@ function zen_catalog_base_link($connection = '')
 ////
 // The HTML image wrapper function
   function zen_image($src, $alt = '', $width = '', $height = '', $params = '') {
+    if ($src === DIR_WS_CATALOG_IMAGES) {
+      return '';
+    }
     $image = '<img src="' . $src . '" alt="' . zen_output_string($alt) . '"';
     // soft clean the alt tag
     $alt = zen_clean_html($alt);


scottcwilson:
I flip-flopped on this one between just returning nothing and returning images/no_picture.gif.  But that file could be missing too, so this seems less intrusive.  Also, tried TEXT_IMAGE_NONEXISTENT and that just seemed obnoxious.

torvista:
> returning images/no_picture.gif. But that file could be missing too,

Shouldn't be, it's in the distribution.

lat9:
> > returning images/no_picture.gif. But that file could be missing too,
> 
> Shouldn't be, it's in the distribution.

Yes, but a site _could have_ removed that file.

scottcwilson:
I'm happy to add a notifier if people think different behavior may be desired.  I just agreed with @torvista that this particular case should not throw a 403.

drbyte:
I think I'm fine with it returning blank.

What's an HTML validator gonna say about a `` tag, with blank src?
(frankly I don't care what a validator says: browser behavior is to just ignore it, and the storeowner solves the problem easily by fixing the image)

Issue #5823, opened 2023-07-01
Consolidate replicated references to external dependencies 


scottcwilson:
https://docs.zen-cart.com/dev/release_process/dependency_checks/

shows multiple files including jQuery and other dependencies. These references could be consolidated, reducing the number of places that need to be changed when we go to a newer version.

neekfenwick:
Related to the FontAwesome work on PR #5845 I notice the `admin/includes/stylesheet.css` file is mentioned in the Dependencies list as relying on FontAwesome.  It does indeed require `css/font-awesome.min.css` which no longer exists in my PR because I moved the fontawesome files into `admin/includes/fontawesome` (may be a bad idea but it seemed a good idea to replicate the directory structure of the FontAwesome download bundle so I gave it a root dir to live in).  Searching for this `admin/includes/stylesheet.css` only a handful of files link to it (`salemaker_info.php` etc), and they don't seem to use the new `admin_html_head.php` file.  So, it seems `admin/includes/stylesheet.css` is obsolete, these files can be updated to use `admin_html_head.php`, remove the `admin/includes/stylesheet.css` link, and we can delete that file.

scottcwilson:
I will begin work on this as soon as #5845 has been merged.

drbyte:
> These references could be consolidated, reducing the number of places that need to be changed when we go to a newer version

There's not much "fluff" in there. Things are already quite consolidated. Apart from self-hosting the files, which is a different approach, there's not much to be gained. If you add "more" files to "centralize" things it creates more points of failure, resulting in the perception that "this template system is hard to use".

scottcwilson:
My intention was to create a single PHP file in a new folder (say, /depend_inc) to pull in jquery: 

depend_inc/get_jquery.php would contain: 
<script src="https://code.jquery.com/jquery-3.6.1.min.js" integrity="sha256-o88AwQnZB+VDvE9tvIXrMQaPlFFSUTR+nldQm1LuPXQ=" crossorigin="anonymous"></script>

Then, the two files 
includes/templates/responsive_classic/common/html_header.php
includes/templates/template_default/common/html_header.php

would use 
require 'depend_inc/get_jquery.php'; 
 
Instead of each specifying a version of jquery that would have to be updated in two places. 

And so forth, for all remotely retrieved dependencies.  

Now you can just look at the depend_inc folder and see all the specific versions being used.

neekfenwick:
[...]

PR #5845, opened 2023-07-03
FontAwesome update

neekfenwick:
First draft. Notes:

- took the 'Free' latest release from https://fontawesome.com/download
- Mostly, the fa-solid fonts seem to match what the v4 font icons looked like.  The docs seem to say only 'solid' is free but 'regular' is also in the free download so I've used some fa-regular fonts as well.
- Obsolete glyphs listed at https://fontawesome.com/docs/web/setup/upgrade/whats-changed have been replaced.
- New configuration item FONTAWESOME_V4_SHIM with values 'true'|'false' and included in upgrade SQL.  I don't know what to do about people who already have v158 and may miss running the SQL update.  It will likely hard crash their site when `admin_html_head.php` refers to that symbol.  Advice please?


zcwilt:
I don't see code that adds anything to the upgrade sql related to FONTAWESOME_V4_SHIM
where you do <?php if (FONTAWESOME_V4_SHIM == 'true') 
then add a test to see if it is `defined()` as well to mitigate it not existing

neekfenwick:
> I don't see code that adds anything to the upgrade sql related to FONTAWESOME_V4_SHIM where you do <?php if (FONTAWESOME_V4_SHIM == 'true') then add a test to see if it is `defined()` as well to mitigate it not existing

Whoops my mistake, thanks. I've pushed the fix.

zcwilt:
k. so another issue. 
Looks like you are using configuration_group_id = 1
the problem here is that will create a visible configuration setting that a store admin can manipulate. 
We don't want that do we ??

If you look at the configuration group definitions you will see the configuration group 6 ('Modules') is always hidden, so is best to use this for adding settings we don't want the admin to change. Unless you think we want the admin to change that ??

neekfenwick:
> 



> the problem here is that will create a visible configuration setting that a store admin can manipulate.

That was my intention, I think the use case is when a user installs an old addon that doesn't know about the v6 upgrade and the store admin needs to be able to turn on the shim.

zcwilt:
Cool. That makes sense. Just wanted to check

zcwilt:
I guess that leaves the description and whether that conveys what the option is doing. Not saying what you have is wrong, but others might disagree. 
Will pull down code and test myself.

lat9:
Should this be a _soft_ configuration setting that can be overridden via a site's use of `admin/includes/extra_datafiles/site_specific_admin_overrides.php` (distributed in `admin/includes/extra_datafiles/dist-site_specific_admin_overrides.php`)?

This 'setting' feels like it has the possibility of being removed after time (when the majority of plugins have caught-up with the FA usage) plus adding yet-another database configuration setting for this just doesn't sit well with me.

lat9:
Noting that using a site-specific 'soft' setting, e.g. `$font_awesome_4_required`, would enable plugins as well as site-specific customizations to set that flag to a `truthy` value to enable the 'shim' to be loaded.  That also enables the FA4-to-FA6 change to occur without a _major_ version change since it doesn't require changes to the database.

The base Zen Cart processing would then check for `!empty($font_awesome_4_required)` to see if the shim should be loaded.

dennisns7d:
Should the CDN links in admin_html_head.php include an integrity attribute?

lat9:
> Should the CDN links in admin_html_head.php include an integrity attribute?

Absolutely!

scottcwilson:
At some point I will migrate these references out of their current places and put them in a single place so they are more easily tracked and updated.  See #5823.

neekfenwick:
> Should the CDN links in admin_html_head.php include an integrity attribute?

Good point. PR Updated. Also added integrity for jquery-ui.css.

neekfenwick:
> Should this be a _soft_ configuration setting that can be overridden

This seems better to me too @lat9, the `configuration` entry did seem a bit heavy handed. I've updated the PR to use `site_specificadmin_overrides` and updated the `dist_` file.

Also added the missing `v4-shims.min.css`, doh!

neekfenwick:
> At some point I will migrate these references out of their current places and put them in a single place so they are more easily tracked and updated. See #5823.

I just noticed the `catalog` files also use fontawesome, at least in `responsive_classic`, so I have pushed an update to the PR to convert `catalog` too: https://github.com/zencart/zencart/pull/5845/commits/4bce32c0aa06d7676aa971090ab5d55b2f744370.  I had thought incorrectly that since `admin` seemed the only place fontawesome files had been baked into ZC, then only `admin` needed attention when upgrading FontAwesome.

`catalog` has a CDN link to fontawesome (in `responsive_classic/common/html_header.php`), while `admin` has the option to self host the files.  Is this for any good reason, or just the result of some haphazard development over the years?  It looks like we should consolidate the self-hosting option to encompass `catalog` and `admin`, and have only one set of files shared by both.

admin/includes/admin_html_head.php
```diff
@@ -22,12 +22,12 @@
     <link rel="stylesheet" href="<?php echo $value; ?>">
     <link rel="stylesheet" href="includes/fontawesome/css/solid.min.css">
     <link rel="stylesheet" href="includes/fontawesome/css/regular.min.css">
-    <?php if (defined('FONTAWESOME_V4_SHIM') && FONTAWESOME_V4_SHIM == 'true' && file_exists($value = 'includes/fontawesome/css/v4-shims.min.css')) { ?>
+    <?php if (isset($fontawesome_v4_shim_required) && $fontawesome_v4_shim_required == true && file_exists($value = 'includes/fontawesome/css/v4-shims.min.css')) { ?>
```
lat9, line 25: Use `!empty($fontawesome_v4_shim_required)` instead of `isset` and the follow-on test.

admin/includes/admin_html_head.php
```diff
@@ -22,12 +22,12 @@
...
         <link rel="stylesheet" href="<?php echo $value; ?>">
     <?php } ?>
 <?php } else { ?>
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.0/css/all.min.css" integrity="sha256-HtsXJanqjKTc8vVQjO4YMhiqFoXkfBsjBWcX91T1jr8= sha384-iw3OoTErCYJJB9mCa8LNS2hbsQ7M3C0EpIsO/H5+EGAkPGc6rk+V8i04oW/K5xq0 sha512-iecdLmaskl7CVkqkXNQ/ZH/XLlvWZOJyj7Yy7tcenmpD1ypASozpmT/E0iPtmFIB46ZmdtAc9eNBvH0H/ZpiBw==" crossorigin="anonymous">
-    <?php if (defined('FONTAWESOME_V4_SHIM') && FONTAWESOME_V4_SHIM == 'true') { ?>
```
lat9: Ditto.

lat9 (changes requested):
Per comments.

admin/includes/admin_html_head.php
neekfenwick: I checked https://www.php.net/manual/en/function.defined.php when making the change, which states "If you want to see if a variable exists, use isset() as defined() only applies to constants" .. `empty`does not seem a good fit to me, see https://www.php.net/manual/en/function.empty.php

neekfenwick: PR updated.

admin/includes/admin_html_head.php
neekfenwick: PR updated.

lat9 (approved):

lat9:
> `catalog` has a CDN link to fontawesome (in `responsive_classic/common/html_header.php`), while `admin` has the option to self host the files. Is this for any good reason, or just the result of some haphazard development over the years? It looks like we should consolidate the self-hosting option to encompass `catalog` and `admin`, and have only one set of files shared by both.

I was wondering that, too, since we're then "carrying the load" for those self-hosted files.

Anybody know why we're supporting self-hosting of the font-awesome, bootstrap and jquery payloads?

lat9 (dismissed):
Rather than introduce an `admin/includes/fontawesome` sub-directory, please move the files from `fontawesome/css`  into the `includes/css` sub-directory and those in `fontawesome/webfonts` into the `includes/fonts` one.  No sense in changing the admin directory structure as well.

Also, remove the `zc_install` updated file since nothing's actually changed.

lat9:
The _not loading_ of the FA-4 shim by default is starting to give me the [heebie jeebies](https://www.dictionary.com/browse/heebie-jeebies).  There are a significant number of plugins (both admin and storefront) that make use of those FA-4 classes and without that shim's load, this could turn into a [goat rodeo](https://www.dictionary.com/e/goat-rodeo/) _very quickly_.

includes/classes/message_stack.php
dennisns7d, line 1: To complete the transition to FA6, change `fas` to `fa-solid`.

dennisns7d: > Rather than introduce an `admin/includes/fontawesome` sub-directory, please move the files from `fontawesome/css` into the `includes/css` sub-directory and those in `fontawesome/webfonts` into the includes/fonts one.

There are a number of references to `../webfonts/` in `fontawesome/css` files that may argue against this recommendation.

lat9: Sigh, could we just do away with this self-hosted version?

dennisns7d: Since the Font Awesome values are commented out, it would probably be better to remove the comment lines which imply a combination of Bootstrap and Font Awesome neither of which are part of the catalog core. The ZCA Bootstrap 4 template extends this class to do this.

neekfenwick:
> Anybody know why we're supporting self-hosting of the font-awesome, bootstrap and jquery payloads?

Self hosting has benefits in SEO, because you can reduce the number of resources fetched for a page by consolidating them into single files.  I know the current ZC approach doesn't do this, but in our shop for example we aggressively compress up to 8 CSS files into one minified one so our shop pages only fetch a single CSS file that's customised to the page being loaded.  This was mostly for Google Page Speed to get a high performance ranking.

For FontAwesome we actually built a 'Kit' of glyphs so we serve our own custom CSS and webfonts, different to what the CDN would give you.  Of course I've had to modify core ZC to achieve all this.

Losing the self hosting approach would make persuing these kinds of optimisations as part of core ZC more difficult.

includes/classes/message_stack.php
neekfenwick: I'll remove the comments from this core file for now.

The FontAwesome references in comments were introduced in https://github.com/zencart/zencart/commit/8bf4f7711e8e34085dab351113909e73fe451f94 by @drbyte, can I ask what the thinking was around that time? Looks ilke you were playing with introducing FA to the `catalog` files, or maybe just copied some code from `admin/classes/message_stack.php` then commented it out.  Only `responsive_classic/common/html_header.php` brings in FontAwesome, presumably other template would not and so core ZC cannot rely on it.

neekfenwick: > There are a number of references to `../webfonts/` in `fontawesome/css` files that may argue against this recommendation.

Yes, though this was not specifically on my mind at the time, it seemed wise to mimic the FontAwesome release directory structure.  The actual release is much larger than what I've included (about 10 subdirs, lots of SVG and JS stuff etc) and when updating in future it will be easier to maintain if the files aren't mixed in with ZC.  @lat9 What is the harm in having this `admin/includes/fontawesome` subfolder?

neekfenwick:
> Also, remove the `zc_install` updated file since nothing's actually changed.

Done, damn whitespace 😀

neekfenwick:
> The _not loading_ of the FA-4 shim by default is starting to give me the [heebie jeebies](https://www.dictionary.com/browse/heebie-jeebies). There are a significant number of plugins (both admin and storefront) that make use of those FA-4 classes and without that shim's load, this could turn into a [goat rodeo](https://www.dictionary.com/e/goat-rodeo/) _very quickly_.

@lat9 this sounds valid if true. I don't know what's out there.  We could certainly make 'load the v4-shim' on by default (i.e. omitted from `site_specific_overrides`) and change the flag to `$fontawesome_suppress_v4_shim`.  Anyone else support this?

scottcwilson:
Please remove the change to zc_install/sql/updates/mysql_upgrade_zencart_158.sql.

@neekfenwick This is a heroic change.  Thank you.  Yikes I wish we didn't have to make all these changes just because FA renamed some things, but there you are.

lat9:
I'm of the belief that, for compatibility, **the 'shim' be loaded by default** and a site/developer given the opportunity to _not load_ the shim for testing purposes.  Pulling the plug on FA-4 is, as mentioned before, going to be a goat-rodeo.

lat9:
> > The _not loading_ of the FA-4 shim by default is starting to give me the [heebie jeebies](https://www.dictionary.com/browse/heebie-jeebies). There are a significant number of plugins (both admin and storefront) that make use of those FA-4 classes and without that shim's load, this could turn into a [goat rodeo](https://www.dictionary.com/e/goat-rodeo/) _very quickly_.
> 
> @lat9 this sounds valid if true. I don't know what's out there. We could certainly make 'load the v4-shim' on by default (i.e. omitted from `site_specific_overrides`) and change the flag to `$fontawesome_suppress_v4_shim`. Anyone else support this?

I know I'm being selfish, but I've got a **_bunch_** of admin-related plugins that, as of zc156, use the FA-4 glyphs.  I'd also really, really, like that variable's name to be a tad more generally descriptive instead of developer-centric like  `$font_awesome_load_not_required`.

scottcwilson:
Yep.  Please do it this way.

lat9:
How about the name I suggested?

neekfenwick:
> How about the name I suggested?
> $font_awesome_4_load_not_required

I thought it was a bit crap, to me honest :stuck_out_tongue: I tripped over almost every part of it.
- `font_awesome` seems wrong by current naming standards, they seem to have moved from names like `font-awesome.css` to `fontawesome.css`, so lose the `_`.
- the `4` on its own seems wrong, better to have `v4`.
- `load` is too vague, we're talking about the shim but yours seems to indicate the entirety of fontawesome v4.
- `not_required` isn't very imperative and does not clearly indicate that turning the flag on will actively prevent something from being included.  

`$fontawesome_suppress_v4_shim` doesn't have any of these problems.  What do you think?

lat9:
I hear you, it's just that `shim` is so techno-speak.

lat9:
OK, how about `$disable_fontawesome_v4_compatibility`?

neekfenwick:
> OK, how about `$disable_fontawesome_v4_compatibility`?

Yeah I see your point. On the PSR-12 note should it be camelCased at this point?  So, `$disableFontAwesomeV4Compatibility`?

Note to self, catalog `-dist.php` has mention of admin due to my lazy copy/paste, needs updating.

lat9:
> > OK, how about `$disable_fontawesome_v4_compatibility`?
> 
> Yeah I see your point. On the PSR-12 note should it be camelCased at this point? So, `$disableFontAwesomeV4Compatibility`?
> 
> Note to self, catalog `-dist.php` has mention of admin due to my lazy copy/paste, needs updating.

Sounds good, @neekfenwick.  Thanks for putting up with my picky-bits!

neekfenwick:
No worries @lat9 I've pushed the change, the meaning of the flag is now inverted.  Admin and `responsive_classic` files updated. Can you give it a check please, make sure it's doing what you need it to do?

admin/includes/admin_html_head.php
```diff
@@ -22,12 +22,13 @@
     <link rel="stylesheet" href="<?php echo $value; ?>">
     <link rel="stylesheet" href="includes/fontawesome/css/solid.min.css">
     <link rel="stylesheet" href="includes/fontawesome/css/regular.min.css">
-    <?php if (!empty($fontawesome_v4_shim_required) && file_exists($value = 'includes/fontawesome/css/v4-shims.min.css')) { ?>
+    <?php if ((empty($disableFontAwesomeV4Compatibility)) &&
+        file_exists($value = 'includes/fontawesome/css/v4-shims.min.css')) { ?>
```
lat9, line 26: Missed this before; the file name should be `DIR_WS_INCLUDES . 'fontawesome/css/v4-shims.min.css'`

neekfenwick: I've pushed this change, it extended well beyond this line, I guess all those old lines also need updating.  Didn't touch the plugin includes stuff as presumably that directory structure doesn't use DIR_WS_INCLUDES in the same way.

neekfenwick:
@lat9 I just noticed you seem to have done a normal merge onto v158, which has put all my "work in progress" commits on that branch.  I was intending a squash commit to hide away some of the ugliness and fiddling about.  Is it too late for a bit of rebasing and squashing?

drbyte:
I'll explore squashing it manually.

neekfenwick:
> I'll explore squashing it manually.

Thank you @drbyte  .. just to try to help, I normally do this from the command line, here's my view of the required squash, while on the current v158 branch, run `git rebase -i HEAD~16` to start interactive rebasing over the last 16 commits, and identify the commits from my branch:

Screenshot from 2023-07-14 08-45-52

edit 'pick' to 'squash' for all but the **first** commit, these will then be squashed up into that first commit.

Screenshot from 2023-07-14 08-46-23

Edit the commit message to indicate it's a squash commit of PR 5845 Fixing Discussion 5763, and commit.  Force push.

drbyte:
Yes, I did basically that. 
Unfortunately the force-push created some chaos with open pull-requests where there had been merge-conflicts. Now the conflicts are even more complex. LOL. Oh well, we'll get them sorted out.

PR #5851, opened 2023-07-04
Check that provided cID, pID exist, redirect if not - fixes #5743

### LINKED ISSUE #5743: Admin->Move Product: after move, debug created with browser Back button
If you move a product from a category, then after completing the move, go back to that category using the Back button (twice), you pass the Move infobox on the way, creating the debug:

> [02-Jun-2023 16:01:48 UTC] Request URI: /zencart/admin/index.php?cmd=category_product_listing&cPath=17&product_type=1&pID=181&action=move_product&page=1, IP address: 127.0.0.1, Language id 1
> #0 D:\GitHub\zencart\admin\includes\modules\move_product.php(12): zen_debug_error_handler()
> #1 D:\GitHub\zencart\admin\category_product_listing.php(1094): require('D:\\GitHub\\zenca...')
> #2 D:\GitHub\zencart\admin\index.php(11): require('D:\\GitHub\\zenca...')
> --> PHP Warning: Undefined variable $pInfo in D:\GitHub\zencart\admin\includes\modules\move_product.php on line 12.

+ another 21 warnings.

I thought it would be straightforward to add a redirect before 12,  back to the original category if $pInfo not set, as handily $current_category_id still exists.
But, no, another error is created related to zen_redirect and the keepalive module...

neekfenwick:
Guards against using 'back' button to revisit page with previous values that are no longer valid, e.g. after move_product to another category.

### DIFF
--- a/includes/classes/sniffer.php
+++ b/includes/classes/sniffer.php
@@ -74,5 +74,53 @@ class sniffer extends base {
     }
     return false;
   }
+
+    /**
+     * Return true if the specified row exists in the table.
+     *
+     * @param string $table_name The table to query.
+     * @param string $key_name The key to check.
+     * @param int    $key_value The value that key_name must equal.
+     * @return void
+     */
+    public function rowExists(string $table_name, string $key_name, int $key_value): bool
+    {
+        global $db;
+        $sql = 'SELECT COUNT(*) AS cc FROM :table_name WHERE :key_name = :key_value;';
+        $sql = $db->bindVars($sql, ':key_name', $key_name, 'passthru');
+        $sql = $db->bindVars($sql, ':key_value', $key_value, 'integer');
+        $sql = $db->bindVars($sql, ':table_name', $table_name, 'passthru');
+        $rs = $db->Execute($sql);
+        return $rs->fields['cc'] != 0;
+    }
+
+    /**
+     * Return true if the specified row exists in the table.
+     * Key column names taken from $key_names are matched against equivalent
+     * key values in $key_values.
+     *
+     * @param string $table_name The table to query.
+     * @param array  $key_names The array of keys to check.
+     * @param array  $key_values The array of values that key_names must equal.
+     * @return void
+     */
+    public function rowExistsComposite(string $table_name, array $key_names, array $key_values): bool
+    {
+        global $db;
+        $sql = 'SELECT COUNT(*) AS cc FROM :table_name WHERE ';
+        $sql .= join(' AND ', array_map(
+            function ($key, $value) {
+                global $db;
+                $bit = ':key = :value';
+                $bit = $db->bindVars($bit, ':key', $key, 'passthru');
+                $bit = $db->bindVars($bit, ':value', $value, 'integer');
+                return $bit;
+            },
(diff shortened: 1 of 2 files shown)


lat9 (changes requested):
Since the changes to the sniffer class comprise the creation of a new method, that method should adhere to [PSR-12](https://www.php-fig.org/psr/psr-12/).

In addition, since zc158 supports PHP versions 7.3 to 8.2, the use of the `union types` should not be used, since that feature was introduced in [PHP 8.0](https://www.php.net/manual/en/language.types.declarations.php).

neekfenwick:
> Since the changes to the sniffer class comprise the creation of a new method, that method should adhere to [PSR-12](https://www.php-fig.org/psr/psr-12/).
> 
> In addition, since zc158 supports PHP versions 7.3 to 8.2, the use of the `union types` should not be used, since that feature was introduced in [PHP 8.0](https://www.php.net/manual/en/language.types.declarations.php).

That's an awfully long spec, and I doubt we adhere to every part of it (80 char line length?) .. Can you point out what I need to change please?

For union types, sorry I thought we were supporting 8.0+, my mistake.  What would you suggest here, introduce two functions, one that takes strings and another that takes arrays?  I admit I coded myself into a corner when I needed to match against a composite key in `products_to_categories` :)

includes/classes/sniffer.php
```diff
@@ -74,5 +74,40 @@ function field_type($table_name, $field_name, $field_type, $return_found = false
...
+   * If arrays are passed, each element in $key_name will be matched against
+   * its corresponding index entry in $key_value, similar to how str_replace()
+   * works.
+   *
+   * @param [string] $table_name The table to query.
+   * @param [string|array] $key_name The key, or array of keys, to check.
+   * @param [int|array] $key_value The value, or array of values, that key_name must equal.
+   * @return void
+   */
+  function row_exists(string $table_name, string|array $key_name, int|array $key_value) : bool {
```
lat9, line 90: - Use `public function`
- Opening brace on new line
- No space between ending parenthesis and `:`
- Remove `union types`, not supported until PHP 8.0

lat9 (changes requested):
Besides the specific comment for the class method definition, code should use an indent of 4 spaces (no tabs).  It's also preferable, but not required, to use `camelCased` class-method names, e.g. `rowExists` vs. `row_exists`.

includes/classes/sniffer.php
neekfenwick: Ah gotcha, thanks for the patience. I've pushed a fix for most issues. (also installed a PSR-12 linter to make this easier in future, though vscode integration is a bit clunky)

On the other two, the sniffer.php file is aready using 2 spaces for indent and function names like `table_exists` not `tableExists` .. should I copy that, or use new standards in my addition only, or convert the whole file to the newer standard?

lat9 (approved):

scottcwilson:
Please squash so we can merge.  Thanks!

neekfenwick:
> Please squash so we can merge. Thanks!

Squashed and pushed.

PR #5850, opened 2023-07-04
Issue 5823 dep consolidation

neekfenwick:
Suggested change to remove obsolete `admin/includes/stylesheet.css`, which has been replaced by `admin_html_head.php` and its specific `<link>` elements including `admin/includes/css/stylesheet.css`.

Also it seems the old `admin/includes/menu.js` and `menu.css` are obsolete, with the JS completely removed and CSS moved to `admin/includes/css/menu.css`, so I've removed the old files too.

PR #5852, opened 2023-07-04
Create new ModList version that uses admin_html_head.php

scottcwilson:
Also removes older versions.

Prompted by #5850


zcwilt:
looks good so far

PR #5853, opened 2023-07-05
Fix layout of search_box elements.

neekfenwick:
The container, outside of search_box.php for example customers.php, is col-md-4 which means it will be 4/12 width in medium breakpoint, and full page width in smaller breakpoints.  The columns here were also set to col-md-4, so they were 4/12 each in a space that was already quite narrow. In small breakpoints they jumped to 100% width which is also not optimal.

The approach here is:
- extra small and small, 50% width so the checkboxes lay out side by side.
- other breakpoints - 100% width so the checkboxes stack on top of each other in the narrow screen space they have been allotted by their parent.

Screenshots:
Small screen - before, the checkboxes are on top of each other:
230605-search-box-small-before
Small screen - after, the checkboxes are side by side, saving some vertical screen space:
230605-search-box-small-after

Wide screen - before, the checkboxes are squished into 4/12 the available width:
230605-search-box-medium-before
Wide screen - after, the checkboxes are allowed to lay out horizontally:
230605-search-box-medium-after

### DIFF
--- a/admin/includes/modules/search_box.php
+++ b/admin/includes/modules/search_box.php
@@ -36,18 +36,18 @@ if (isset($_GET['search']) && zen_not_null($_GET['search'])) {
   </div>
     <?php
     if (file_exists($searchBoxJs)) {
-        ?>
+    ?>
         <div class="row">
-            <div class="form-group col-xs-6 col-sm-12" id="searchRestrictIds">
-                <div class="col-sm-9 control-label"><label for="restrictIDs"><?= TEXT_INFO_SEARCH_FILTER_RESTRICT_IDS; ?></label></div>
-                <div class="col-sm-1">
-                    <?= zen_draw_checkbox_field('restrictIDs', '', (!empty($_GET['restrictIDs']) && $_GET['restrictIDs'] === 'on'), '', ' id="restrictIDs"'); ?>
+            <div class="form-horizontal col-xs-6">
+                <div class="form-group" id="searchRestrictIds">
+                    <label for="restrictIDs" class="col-xs-11 control-label"><?= TEXT_INFO_SEARCH_FILTER_RESTRICT_IDS; ?></label>
+                    <?= zen_draw_checkbox_field('restrictIDs', '', (!empty($_GET['restrictIDs']) && $_GET['restrictIDs'] === 'on'), '', ' id="restrictIDs" class="col-xs-1"'); ?>
                 </div>
             </div>
-            <div class="form-group col-xs-6 col-sm-12" id="searchTermRepopulate">
-                <div class="col-sm-9 control-label"><label for="repopulateSearch"><?= TEXT_INFO_SEARCH_FILTER_REPOPULATE; ?></label></div>
-                <div class="col-sm-1">
-                    <?= zen_draw_checkbox_field('repopulateSearch', '', (!empty($_GET['repopulateSearch']) && $_GET['repopulateSearch'] === 'on'), '', ' id="repopulateSearch"'); ?>
+            <div class="form-horizontal col-xs-6">
+                <div class="form-group" id="searchTermRepopulate">
+                    <label for="repopulateSearch" class="col-xs-11 control-label"><?= TEXT_INFO_SEARCH_FILTER_REPOPULATE; ?></label>
+                    <?= zen_draw_checkbox_field('repopulateSearch', '', (!empty($_GET['repopulateSearch']) && $_GET['repopulateSearch'] === 'on'), '', ' id="repopulateSearch" class="col-xs-1"'); ?>
                 </div>
             </div>
         </div>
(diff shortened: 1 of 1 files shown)


proseLA:
@neekfenwick at some point, i would like to see you take a big whack at something.  it seems you have the energy for it.

one of my issues has always been changes made to accommodate a particular user, as opposed to the community as a whole.  i try to think of the community when i make changes; granted i am not always successful, but i do try.

the current version of bootstrap is 5.3.  we are running 3.4.  hmmm....

i acknowledge that design is not my strong suit; and i hate working with css.  there are others out there who are much stronger in it than i.

this PR works for your screen widths; it does not work for mine.  i spent some time reviewing [the 3.4 docs on grid systems.](https://getbootstrap.com/docs/3.4/css/#grid-options)  i could behind the changing the class to `form-group col-xs-12 col-sm-6 col-md-4 col-lg-4`.  that seems to work across the various screen widths. 

how does that work for you?

neekfenwick:
@proseLA if design isn't your strong suit and you don't like working with CSS, maybe someone else should review this :)

I touched on this because the search box looked ugly in my v158 when I used it, and tracking the line history of the culprit led to this Issue, so I PR'd in an effort to help where the original change seemed to have awry and provided screenshots of it working.  I cannot tell from your description above where you applied the class, perhaps you could be more specific or PR?

It does sound like moving to a more recent bootstrap would be a good idea.  That would belong in another Issue or whatever.

In response to your comments about me personally, I'll take a whack at what I think is a good fit for me.  I've approached a couple of minor Issues I saw as low hanging fruit to try to help reduce the list, since sometimes people raise issues while working on something else and don't have the headspace for them at the time.  I've also bitten off a couple of larger pieces of work to see what that's like in this community.

scottcwilson:
> the current version of bootstrap is 5.3. we are running 3.4. hmmm....

This will be reviewed when we do the next release, and doing this review is now part of the release process.

neekfenwick:
> while css is not my strong suit; i would argue that the change i recommended above is the way to go that addresses various view ports.

I appreciate that, thanks.  I'm at the end of my work day but it looks like you're suggesting the immediately container of the checkboxes have col-md-4, which I believe it already has.  At that window width the parent, that contains the entire search box, is also something like col-md-4, so it's 25% window width.  Having the checkboxes then be 25% of that width makes them squish, as in my screenshot.  I'd be very happy to try your suggestion if you can be a little more specific about where you'd apply that class list :)

proseLA:
> I'd be very happy to try your suggestion if you can be a little more specific about where you'd apply that class list :)
```
             <div class="form-group col-xs-12 col-sm-6 col-md-4 col-lg-4" id="searchRestrictIds">
....
            <div class="form-group col-xs-12 col-sm-6 col-md-4 col-lg-4" id="searchTermRepopulate">
```
this still needs work, see [this screencast.](https://www.dropbox.com/s/8qwn5j9hmjejfi3/search_box.webm?dl=0)

as such, i think this PR needs work.

neekfenwick:
> this still needs work, see [this screencast.](https://www.dropbox.com/s/8qwn5j9hmjejfi3/search_box.webm?dl=0)

@proseLA This screencast shows different markup to my commit https://github.com/zencart/zencart/pull/5853/commits/a3b44d47c0bc948c9282cd3e7eee745691603bec .. I don't know what's gone wrong but you're not running the changes I intended.

For example, my commit has `<div class="form-group col-xs-6 col-sm-12" id="searchRestrictIds">` but your screencast shows four different col-* classes.

proseLA:
i am not a fan of your changes.  my screen is wide.  rather than using the width of my screen, you want to take up the height.

neekfenwick:
> i am not a fan of your changes. my screen is wide. rather than using the width of my screen, you want to take up the height.

Ah I misunderstood your comment "this still needs work", I thought you meant my PR, but you seem to be talking about your attempt.

You're right, my PR had the controls stacking vertically when they shouldn't.  After fiddling around for a while and reviewing the 3.4 docs at https://getbootstrap.com/docs/3.4/css/#forms-horizontal I've got good results using `form-horizontal` and restructuring things a little, which has resulted in fixing a problem we had with the label and checkbox wrapping to a new line when `col-xs-11` and `col-xs-1` were applied.  Seems pretty sweet now.

https://www.dropbox.com/s/ixr4p3t434fzbz7/Screencast_06_07_23_09%3A48%3A04.mp4?dl=0

Just for the record:
- with bootstrap, one only has to apply the smallest `col-*` breakpoint class and it will apply that to larger sizes if unspecified, so `col-xs-6` implies `col-sm-6`, `col-md-6`, `col-xl-6` etc. So, I've only got col-xs-6 because I want the two to be 50% of available width for all screen sizes.
- After re-thinking, I don't see any need for the control groups to be anything *other* than 50% .. so I've not used your suggestion of `col-md-4 col-lg-4`, I just don't see why they'd want to be 4/12 width. (I had a brain fart earlier saying 4/12 was 25%, one of those days).

What do you think of the latest changes?

proseLA:
the latest changes are great.  in my testing, they display quite nicely across multiple view ports.  nicely done.

:+1:

PR #5860, opened 2023-07-07
zenCssButton

### LINKED ISSUE #5545: Sold Out button not respecting admin CSS Button setting
I could be barking up the wrong tree with this as I don't have a vanilla site of ZC v1.5.8 with which to test.

During the process of upgrading from v1.5.7d to v1.5.8 it appeared that the Sold Out button on the product listing and product info pages displays as an image (gif) rather than the CSS button it should be (CSS Button switch in admin is on). 

If it is relevant, the button image has an empty alt tag even though `BUTTON_SOLD_OUT_ALT` is defined.

I'll note that in line 211 of functions_general.php changing the `zen_image` function to the old `zen_image_button` allows the sold out button to be displayed as a CSS button.

Line 211
`$return_button = '<span class="text-center">' . zen_image($template->get_template_dir($image, DIR_WS_TEMPLATE, $current_page_base, 'buttons/' . $_SESSION['language'] . '/') . $image, $alt, '', '', '') . '</span>'; `

Perhaps someone with a ZC v1.5.8 test site could check this.

brittainmark:
fixes  #5545

### DIFF
--- a/includes/functions/functions_general.php
+++ b/includes/functions/functions_general.php
@@ -208,8 +208,13 @@ function zen_get_buy_now_button($product_id, string $buy_now_link, $additional_l
             global $template;
             $image = BUTTON_IMAGE_SOLD_OUT; 
             $alt = BUTTON_SOLD_OUT_ALT; 
-            $return_button = '<span class="text-center">' . zen_image($template->get_template_dir($image, DIR_WS_TEMPLATE, $current_page_base, 'buttons/' . $_SESSION['language'] . '/') . $image, $alt, '', '', '') . '</span>'; 
+            if (strtolower(IMAGE_USE_CSS_BUTTONS) === 'yes') {
+                $return_button = zen_image_button($image, $alt);
+            } else {
+                $return_button = '<span class="text-center">' . zen_image($template->get_template_dir($image, DIR_WS_TEMPLATE, $current_page_base, 'buttons/' . $_SESSION['language'] . '/') . $image, $alt, '', '', '') . '</span>'; 
+            }
             $zco_notifier->notify('NOTIFY_ZEN_SOLD_OUT_IMAGE', array_merge($button_check->fields, ['products_id' => (int)$product_id]), $return_button);
+                
             break;
         default:
             $return_button = $buy_now_link;
--- a/includes/templates/classic/css/stylesheet_css_buttons.css
+++ b/includes/templates/classic/css/stylesheet_css_buttons.css
@@ -37,3 +37,7 @@ span.normal_button {
 span.normal_button:hover {
   background-color:#b4bc8b;  /* Hover color for link-related buttons */
 }
+span.button_sold_out, span.button_sold_out:hover{
+  background:#8b0000;
+  color:#ffffff;
+}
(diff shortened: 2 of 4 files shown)


scottcwilson:
Product 108 in the test dataset (a sold out product).  

Before: 
before

After: 

after

scottcwilson:
Looks right to me.

torvista:
Should there not be some visual difference between Ask a Question (link) and Sold Out (information/warning)?

brittainmark:
Can add CSS. What do you suggest, red background white text?

scottcwilson:
Fair enough, just make the button background color red.

brittainmark:
Added Id and css. I have changed the background so it stays the same colour on hover as it did not look right changing colour with no action.
Responsive Classic Template
image

Classic Contemporary Green
image

scottcwilson:
Love it!  Thank you.

scottcwilson (approved):

includes/functions/functions_general.php
```diff
@@ -208,8 +208,13 @@ function zen_get_buy_now_button($product_id, string $buy_now_link, $additional_l
             global $template;
             $image = BUTTON_IMAGE_SOLD_OUT; 
             $alt = BUTTON_SOLD_OUT_ALT; 
-            $return_button = '<span class="text-center">' . zen_image($template->get_template_dir($image, DIR_WS_TEMPLATE, $current_page_base, 'buttons/' . $_SESSION['language'] . '/') . $image, $alt, '', '', '') . '</span>'; 
+            if (strtolower(IMAGE_USE_CSS_BUTTONS) == 'yes') {
+                $return_button =  zenCssButton($image, $alt, 'button', '', 'id=soldOutButton');
```
lat9, line 212: An `id=` is going to generate invalid HTML if multiple products are 'sold out' on a products' listing page.  Suggest using `class=` instead.

lat9 (changes requested):
Change `id=` to `class=` for valid HTML with multiple sold-out products on a listing.

includes/functions/functions_general.php
brittainmark: Good spot thanks.
Cannot use class as html_output line 451 will create a duplicate class.
` $css_button = '<input class="' . $mouse_out_class . '" ' . $css_button_js . ' type="submit" value="' . $text . '"' . $tooltip . $parameters . '>';`

I suggest the i remove the parameter. 
The button is identified in css by `span.button_sold_out`

lat9: Hold off; this should be using
```php
$return_button = zen_image_button($image, $alt);
```
That will generate a `button` with a class of `button_sold_out`.

lat9: ... and the comparison should be strict, i.e. `=== 'yes'`.

brittainmark: I tried zen_image_button that creates a clickable button that takes you to the cart. 
That was why I used zenCssButton. 
ZenCssButton also generate a button with class button_sold_out. Clicking it has no action, the same as the image

lat9: I'm not seeing that generation of a 'clickable' button.  When I make the change that I suggested (less the CSS updates), I'm presented with an unclickable `span` that says "Sold Out"; here's the hover image
image

brittainmark: OK I'll try it again. That is where I started. Ignore last push. I'll check it out and push again if ok.

brittainmark: Why should I doubt the master. Works a treat!

lat9 (approved):

PR #5864, opened 2023-07-09
v158b Add string PRODUCT_URL_REGEX sanitizing

### LINKED ISSUE #5849: function filterProductUrlRegex debug for empty parameter
While delousing the Flexible Footer, I was getting a debug when updating a page definition,

> --> PHP Warning: Invalid argument supplied for foreach() in ADMIN/includes/classes/AdminRequestSanitizer.php on line 515.

https://github.com/zencart/zencart/blob/e27bdb54a1f92f14a548f5473eb5bab9f3c85d3f/admin/includes/classes/AdminRequestSanitizer.php#L511-L515

the parameter being passed was
`    [page_url] => `

so set, but empty.
Changing the isset on line 511 to !empty fixed it.

mc12345678:
Original design of the admin sanitizer associated with PRODUCT_URL_REGEX was written to directly support only the data type that is specifically used by Zen Cart which is a language based array of values. Use of this filter type on a non-array variable causes errors.

Four methods of resolution came to mind:
1. Other software attempting to use this filter should make the data an array, filter it, then return it to not being an array.
2. Add/expand filtering by use of class overriding.
3. Add an additional filter that only addresses stringss (not arrays). This appears would complicate the code and control. Either the new code would reference the existing (logging a "non-existent" sanitization) or would effectively duplicate the code such that a change in one would require a change in the other.
4. Expand the method's code to perform the same operation but with the data formatted solely as a string instead of an array and without recursion.

As part of the feedback from the initial poster, also corrected the capitalization of the call to the method `setCurrentArrayName` to ensure systems that have begun enforcing or will enforce this.

I normally would have placed the smaller/shorter code portion first and then allowed the `foreach` loop to follow; however, based on so many micro-optimisations recently implemented I am not sure what (re-)sequencing is "best". Considering the base code was expected to process an array, I put/kept that first so that inner code would not always get bypassed during normal operation.

I also left the assignment to $urlRegex before the early escape to align with perhaps old coding style recommendations of defining a variable as soon as possible within the code. A micro-optimisation would be to move that variable assignment to after the first check so that it is created only when it would become useful. Though perhaps there are other reasons to declare it each method call.

Fixes #5849

### DIFF
--- a/admin/includes/classes/AdminRequestSanitizer.php
+++ b/admin/includes/classes/AdminRequestSanitizer.php
@@ -508,12 +508,15 @@ class AdminRequestSanitizer extends base
     private function filterProductUrlRegex($parameterName)
     {
         $urlRegex = '~([^0-9a-z' . preg_quote("'.!@#$%&()_-~/;:=?[]", '~') . ']|[><])~i';
-        if (isset($_POST[$parameterName])) {
-            // Add the parameterName to the base arrayname.
-            $this->arrayName = $this->setCurrentArrayName($parameterName);
-            $this->debugMessages[] = 'PROCESSING PRODUCT_URL_REGEX == ' . $this->arrayName;
+        if (!isset($_POST[$parameterName])) {
+            return;
+        }
+        // Add the parameterName to the base arrayname.
+        $this->arrayName = $this->setCurrentArrayName($parameterName);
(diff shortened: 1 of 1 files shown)


drbyte:
Seems like a reasonable solution to the problem.

I might suggest a code comment be added before the non-array processing to note that it's intentionally basically a repeat of the array processing. Just so the human coder who comes along doesn't have to wonder whether it's the same logic or not... just saves on questions, but is minor.

PR #5865, opened 2023-07-09
V158 sanitizer regex url

mc12345678:
v158b Refactor Admin Sanitizer and add String Only sanitization
    
In association with fixing #5849, applied the same type of correction
to other unique filters that only managed array type sanitization.

With incorporation of a change, took an opportunity to refactor the
code using some of the principles that have been applied by others in the
community.
early escape
remove empty lines
use elseif instead of else if

### DIFF
--- a/admin/includes/classes/AdminRequestSanitizer.php
+++ b/admin/includes/classes/AdminRequestSanitizer.php
@@ -508,12 +508,15 @@ class AdminRequestSanitizer extends base
     private function filterProductUrlRegex($parameterName)
     {
         $urlRegex = '~([^0-9a-z' . preg_quote("'.!@#$%&()_-~/;:=?[]", '~') . ']|[><])~i';
-        if (isset($_POST[$parameterName])) {
-            // Add the parameterName to the base arrayname.
-            $this->arrayName = $this->setCurrentArrayName($parameterName);
-            $this->debugMessages[] = 'PROCESSING PRODUCT_URL_REGEX == ' . $this->arrayName;
+        if (!isset($_POST[$parameterName])) {
+            return;
+        }
+        // Add the parameterName to the base arrayname.
+        $this->arrayName = $this->setCurrentArrayName($parameterName);
+        $this->debugMessages[] = 'PROCESSING PRODUCT_URL_REGEX == ' . $this->arrayName;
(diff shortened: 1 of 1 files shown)


torvista:
This works without the errors generated as described in https://github.com/zencart/zencart/issues/5849

I think the error log time style should be 24 hr and not 12 as currently set.

proseLA:
> This works without the errors generated as described in #5849
> 
> I think the error log time style should be 24 hr and not 12 as currently set.

i think the error log time should be the same as whatever is configured in the admin header bar:

image

lat9:
> I think the error log time style should be 24 hr and not 12 as currently set.

I think that that's a separate PR!

mc12345678:
> I think the error log time style should be 24 hr and not 12 as currently set.

Also then missing the timezone information. 

Should this be a discussion topic? Having it log date/time differently than error logging or not by a single format with just a difference in prefix/suffix could be either informative or distracting.

proseLA:
i think the error log should format the same as the date in the header bar in the admin, ie:

image

however someone has the configured, that should be the same for the error logging.

that to me makes the most sense...

drbyte:
Why are we talking about error log formats or date/time formats in this PR?

mc12345678:
> Why are we talking about error log formats or date/time formats in this PR?

I saw it as "oh by the way", one more thing to fix? Though I personally thought the statement was made in the original issue and not in the PR to fix the original issue.

drbyte:
I tried to fix the merge conflicts by merging `v158` back into this branch, but it's rejecting my submission. Not sure if it's because you've got the "allow maintainers to update PR's" turned off, or if it's something else.

mc12345678:
> I tried to fix the merge conflicts by merging `v158` back into this branch, but it's rejecting my submission. Not sure if it's because you've got the "allow maintainers to update PR's" turned off, or if it's something else.

Suspecting because I purposefully chose to deselect the option "Allow edits and access to secrets by maintainers". I have force pushed the two commits. Note that comments had been made against only the first commit and believing that the comment was intentionally only against the first commit, I did not make similar changes to the second commit and I also did not squash the commits. Trying to make it easy to be bring it all together or reject in whole.

PR #5866, opened 2023-07-09
Replace tab with space

mc12345678:
Somehow tab characters were introduced into these files instead of the routine 2 or 4 spaces. This exchanges those tab characters for the associated spaces.

### DIFF
--- a/admin/includes/languages/english/lang.salemaker_info.php
+++ b/admin/includes/languages/english/lang.salemaker_info.php
@@ -27,12 +27,12 @@ $define = [
                       </li>
                       <li>
                         You must choose the action to take if a product is a special <i>and</i> is subject to this sale:
-						<ul>
+                        <ul>
                           <li>
                             <strong>Ignore Specials Price - Apply to Product Price and Replace Special</strong><br>
-							The sale deduction will be applied to the regular price of the product.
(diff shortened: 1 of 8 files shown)


drbyte:
Note: one of the reasons why those standalone template files used tabs for CSS definitions is because it's a pretty common convention to use tabs instead of spaces in CSS files in order to keep files smaller, so they load faster.
That's insignificant in the case of these template files because they're rarely used, and only under very specific circumstances.

mc12345678:
If they were 100% present or used, I wouldn't have thought twice about the tabs. They were and have been Inconsistently added, some times after two or four spaces to then maybe have a space after. 

Even faster for css is to almost not have any space or return and basically for everything to stack up. A lot harder to evaluate or read for changes, but definitely on the top end of shorter.

drbyte:
> If they were 100% present or used, I wouldn't have thought twice about the tabs. They were and have been Inconsistently added, some times after two or four spaces to then maybe have a space after.

Yup. Agreed.
 
> Even faster for css is to almost not have any space or return and basically for everything to stack up. A lot harder to evaluate or read for changes, but definitely on the top end of shorter.

True enough. Hence PurgeCSS etc.

PR #5868, opened 2023-07-09
Allow intended error logging before `$_SESSION` set

mc12345678:
If a log is requested to be generated before `$_SESSION` has been set or specifically if `$_SESSION['languages_id']` is not set, then the absence of that data causes an issue/error such as the below:

[Date Time Timezone] PHP Warning:  Undefined global variable $_SESSION in /catalog-path/includes/extra_configures/enable_error_logging.php on line 84
[Date Time Timezone] PHP Warning:  Trying to access array offset on value of type null in /catalog-path/includes/extra_configures/enable_error_logging.php on line 84

This commit allows that to not exist and adds more information onto the reporting to indicate that it didn't exist at the time of log generation.

### DIFF
--- a/includes/extra_configures/enable_error_logging.php
+++ b/includes/extra_configures/enable_error_logging.php
@@ -81,7 +81,7 @@ function zen_debug_error_handler($errno, $errstr, $errfile, $errline)
     if (!empty($backtrace)) {
         $backtrace = PHP_EOL . rtrim($backtrace);
     }
-    $language_info = ', Language id ' . $_SESSION['languages_id'];
+    $language_info = ', Language id ' . ($_SESSION['languages_id'] ?? 'not set');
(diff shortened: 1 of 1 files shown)


includes/extra_configures/enable_error_logging.php
```diff
@@ -81,7 +81,7 @@ function zen_debug_error_handler($errno, $errstr, $errfile, $errline)
     if (!empty($backtrace)) {
         $backtrace = PHP_EOL . rtrim($backtrace);
     }
-    $language_info = ', Language id ' . $_SESSION['languages_id'];
+    $language_info = ', Language id ' . ($_SESSION['languages_id'] ?? 'SESSION[languages_id] NOT DEFINED');
```
lat9, line 84: How about a simple 'not set', like the IP address?

lat9 (changes requested):
See comment for line 84.

includes/extra_configures/enable_error_logging.php
mc12345678: At the time of entry, it seemed more informative to have something other than "not set" which can occur in a fully executed and completed load as compared to a call occurring before the value is available.

Modified it anyways to align with the other "not set" type values.

lat9 (approved):

lat9:
Thanks, @mc12345678.

PR #5869, opened 2023-07-10
correct check for valid cID

torvista:
Fixes Bug:
Create a new category and save it: browser redirects to the top category listing due to this check for valid category being incorrect.


drbyte:
Manually merged separate from this PR due to merge conflicts.

torvista:
It does not look fixed:
https://github.com/zencart/zencart/blob/d427cbfe884e00e95d35774636474727935e807b/admin/category_product_listing.php#L38-L40
should be
`if (!$sniffer->rowExists(TABLE_CATEGORIES, 'categories_id', $_GET['cID'])) {`

drbyte:
Thanks for checking. Should be there now.

PR #5871, opened 2023-07-10
Categories listing, column heading/data notifications

lat9:
While there are notifications to allow the _gathering_ of additional category-related fields during an insert/edit, there are currently no notifications to allow additional columns to be added to a category-listing in the admin.

This PR defines those notifications.


lat9:
That's not going to work, since when there's a search string there are potentially a combination of products and categories.

lat9:
I'm just going to start over.

PR #5875, opened 2023-07-12
Correcting `checkout_success` logs when a customer has no previous orders

### LINKED ISSUE #5742: debug for $orders_id not set in checkout_success\header_php.php
Now and again I get:
```
#0 includes/modules/pages/checkout_success/header_php.php(54): zen_debug_error_handler()
#1 index.php(35): require('...')
--> PHP Warning: Undefined variable $orders_id in includes/modules/pages/checkout_success/header_php.php on line 54.
```

https://github.com/zencart/zencart/blob/9ccb250afda600b06c3446b50b992406f6ae661f/includes/modules/pages/checkout_success/header_php.php#L43-L54

When there is no session order_id, it tries to use the result from the query, which may be not set if there were no previous orders.

So, some clause needs to be in there. Not sure what would be best....

lat9:
Fixes #5742.


drbyte:
Seems like an ideal solution to me.

PR #5878, opened 2023-07-17
ot_group_discount: Correct tax calculations when "Calculate Tax" when "None"

lat9:
The `include_tax` property is set to the value of the configuration setting, a 'true' or 'false' string, so its value is always interpreted as _truthy_.

Add a strict comparison to determine whether/not the taxes are to be included.  As the module currently sits, tax is _always_ included in the deduction when the "Calculate Tax" setting is "None".

### DIFF
--- a/includes/modules/order_total/ot_group_pricing.php
+++ b/includes/modules/order_total/ot_group_pricing.php
@@ -160,7 +160,7 @@ class ot_group_pricing {
        */
       switch ($this->calculate_tax) {
         case 'None':
-          if ($this->include_tax) {
+          if ($this->include_tax === 'true') {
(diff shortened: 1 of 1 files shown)


scottcwilson:
Is this also a defect in ot_coupon.php?

lat9:
> Is this also a defect in ot_coupon.php? (Not the exact same code , but missing logic in the 'None' case.)

I dunno, I was making a clone of ot_group_pricing.php and came across that boo-boo there.

scottcwilson:
No worries, I looked at  ot_coupon quickly and it seems ok; it doesn't do anything for the None case.

drbyte (approved):

PR #5880, opened 2023-07-17
Update index_dashboard.php

dbltoe:
More and more, we are having to move orders to the top in the admin as it simplifies access without scrolling.

If New Orders and Customers are easily seen above the fold, why not order statuses

### DIFF
--- a/admin/index_dashboard.php
+++ b/admin/index_dashboard.php
@@ -41,15 +41,15 @@ if (empty($currencies)) {
     <div id="colone" class="col-xs-12 col-sm-6 col-md-4 col-lg-4">
 
         <?php
+        include DIR_WS_MODULES . 'dashboard_widgets/OrderStatusDashboardWidget.php';
+        ?>
+
+        <?php
         include DIR_WS_MODULES . 'dashboard_widgets/BaseStatisticsDashboardWidget.php';
         ?>
 
         <?php
         include DIR_WS_MODULES . 'dashboard_widgets/SpecialsDashboardWidget.php';
-        ?>
-
-        <?php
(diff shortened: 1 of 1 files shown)


lat9:
What's really needed here, IMO, is a means for a site to easily (another 'soft setting') identify which of the widgets should appear on the dashboard and in which column/sort-order.

scottcwilson:
I'm not positive I want to take this PR.  I think the guidance should be "update to taste until we have a configurable display."

dbltoe:
I can't think of any site where Statistics or Specials are more important to the storeowner than accessing the information in Orders quickly.

When you're dealing with having to find the 9 processing out of 80,989 shipped, having to scroll down the page every time is just frustrating. (one actual site)

I agree that this should be done similarly to lat9's Login-Page Layout in OPC but, until we can do that, why not put pertinent data above the fold?

lat9:
> I can't think of any site where Statistics or Specials are more important to the storeowner than accessing the information in Orders quickly.
> 
> When you're dealing with having to find the 9 processing out of 80,989 shipped, having to scroll down the page every time is just frustrating. (one actual site)
> 
> I agree that this should be done similarly to lat9's Login-Page Layout in OPC but, until we can do that, why not put pertinent data above the fold?

What's [one man's trash is another man's treasure](https://en.wiktionary.org/wiki/one_man%27s_trash_is_another_man%27s_treasure).  I'm going with @scottcwilson on this one.  Until such time as we have a means to provide some layout customization, sites that want a different view of their world are welcome to alter the base file (with appropriate comments and/or a backup of the original, of course).

PR #5881, opened 2023-07-18
Remove usage of the deprecated `utf8_encode`

### LINKED ISSUE #5387: PHP8.2 utf8_encode and utf8_decode are deprecated
**Describe the bug**
If PHP 8.2 is used and the CHARSET is not UTF-8 a depreciation error `This function has been DEPRECATED as of PHP 8.2.0. Relying on this function is highly discouraged.` will be raised in ajax.php and includes/functions/function_strings.php

**Version**
Zen Cart version: 1.5.8
PHP version: 8.2

**Additional context**
`utf8_encode` could be replaced with `iconv` to perform the necessary conversion. Assuming a conversion is necessary at all.
Additionally, does ajax.php need its own copy of the function?

lat9:
Fixes #5387

Notes:
- Issuing an `E_USER_DEPRECATED` log for any usage of `utf8_encode_recurse`, just in case some plugin is using that function.
- The removal of the `testEnsureDataIsUtf8` method from the test framework's AdminLoggingTest.php will most likely result in a unit-test failure.  I don't know where that test is invoked!

### DIFF
--- a/admin/includes/classes/class.admin.zcObserverLogEventListener.php
+++ b/admin/includes/classes/class.admin.zcObserverLogEventListener.php
@@ -74,7 +74,6 @@ class zcObserverLogEventListener extends base {
      */
     $postdata = $_POST;
     $postdata = self::filterArrayElements($postdata);
-    $postdata = self::ensureDataIsUtf8($postdata);
     $notes = self::parseForMaliciousContent(print_r($postdata, true));
     /**
      * Since the POST data was an array, we json-encode the parsed POST data for storage in the logging system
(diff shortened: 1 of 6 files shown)


lat9:
Yes, I think that change "blew up" the Functional Tests.

drbyte (approved):

drbyte:
>  I don't know where that test is invoked

In the test suite, for filenames ending in `Test.php` and function names starting with `test`, those test functions are invoked automatically by virtue of naming convention.
(And if the test function's name doesn't begin with `test`, then if a docblock annotation precedes the function declaration and contains `@test`, then the function will be invoked by virtue of that annotation. Both conventions are very common practices with test suites based on phpunit.)

PR #5884, opened 2023-07-21
Suppress error logs in the case of additional MySQL error codes

### LINKED ISSUE #5883: More $db failures to 'squish'
As a follow-on to #5722, which identified the 'database connection failed' as being problematic, I'm also seeing logs of the ilk:
```
--> PHP Warning: mysqli_connect(): (HY000/2002): Connection refused in /home/xxxx/public_html/includes/classes/db/mysql/query_factory.php on line 64.
--> PHP Warning: mysqli_query(): MySQL server has gone away in /home/xxxx/public_html/includes/classes/db/mysql/query_factory.php on line 45.
--> PHP Warning: mysqli_connect(): (08004/1040): Too many connections in /home/xxxx/public_html/includes/classes/db/mysql/query_factory.php on line 64.
```

These SQL server issues also result in a slew of 7-10Kb log files when the SQL server has the identified issues.  Could these also be added to the logs' size-shrink and force a `die()`?

drbyte:
Fixes #5883

### DIFF
--- a/includes/classes/db/mysql/query_factory.php
+++ b/includes/classes/db/mysql/query_factory.php
@@ -37,6 +37,16 @@ class queryFactory extends base
     private $user = '';
     private $password = '';
     private $zf_sql = '';
+    private $ignored_error_codes = [
+        2002, // connection refused via socket
+        2003, // cannot connect to host
+        2006, // server has gone away
+        2013, // lost connection during query
+        1040, // too many connections
+        1053, // server shutdown in progress
+        1141, // grant-type not allowed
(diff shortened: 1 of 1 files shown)


scottcwilson:
We still ought to report the error code one time (so you know whether it's a host-side issue like 2006 or an attack like 1203).

lat9:
> We still ought to report the error code one time (so you know whether it's a host-side issue like 2006 or an attack like 1203).

The failed `mysqli_connect` does that "for us".

Issue #5807, opened 2023-07-01
Admin Side pages with HTML issues 


scottcwilson:
- [x] Developer's Toolkit.  (original issue #5540)
- [x] Invoice (original issue #5520)
- [ ] Options Value Manager (original issue #5424)

brittainmark:
developer's toolkit needs ticking above.

PR #5885, opened 2023-07-21
Make DevelopersToolKit HTML5 compliant

brittainmark:
Part of #5807

Changes
1) Added variables `$output` & `$outCount` to display output from zen_diaplay_files to allow output after `<body>` tag separated count from matches.
1) Moved `pixel_black.gif` to  `zen_diaplay_files` function and added to `$output`
1) Moved matches to display below menu bar.
1) Replaced `<table> border, cellspacing, cellpadding` tags with css 
1) Added missing` id` tags to fields and made unique, changing label where necessary 
1) Added `htmlspecialchars `to configuration value so you can see value if it contains html

### DIFF
--- a/admin/developers_tool_kit.php
+++ b/admin/developers_tool_kit.php
@@ -8,6 +8,8 @@
 require('includes/application_top.php');
 
 $default_context_lines = 0;
+$output = '';
+$outCount = '';
 
 if (!empty($_POST) && !isset($_POST['context_lines'])) {
   $_POST['context_lines'] = abs((int)$default_context_lines);
@@ -57,7 +59,7 @@ function getDirList($dirName, $filetypes = 1) {
 }
 
 function zen_display_files($include_root = false, $filetypesincluded = 1) {
-  global $check_directory, $found, $configuration_key_lookup;
+  global $check_directory, $found, $configuration_key_lookup, $outCount, $output;
   global $db;
   $max_context_lines_before = $max_context_lines_after = abs((int)$_POST['context_lines']);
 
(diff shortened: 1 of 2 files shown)


admin/developers_tool_kit.php
```diff
@@ -8,6 +8,8 @@
 require('includes/application_top.php');
 
 $default_context_lines = 0;
+$output = '';
+$outCount = '';
```
lat9, line 12: How about a bit more descriptive variable names, like `$matches_output` and `$matches_header`, respectively.

admin/developers_tool_kit.php
```diff
@@ -206,7 +207,7 @@ function zen_display_files($include_root = false, $filetypesincluded = 1) {
 
         // use to debug for UTF-8 NO BOM on files: test search on a, e, s change if below to true
         if (false && htmlspecialchars($line, ENT_QUOTES, CHARSET) == '') {
-          echo '<br>SOMETHING BROKE in: ' . $file . '<br>on: ' . $line_num . ' - ' . $line . '<br>';
+          $output = '<br>SOMETHING BROKE in: ' . $file . '<br>on: ' . $line_num . ' - ' . $line . '<br>';
```
lat9, line 210: Shouldn't this be `$output .=` instead of `$output =`?

admin/developers_tool_kit.php
```diff
@@ -765,9 +765,13 @@ function number_pad_with_spaces($number, $n = 0) {
   <?php require DIR_WS_INCLUDES . 'admin_html_head.php'; ?>
   </head>
   <body>
+    <?php echo $outCount;?>
```
lat9, line 768: Move this under the header, otherwise, it gets 'lost' as it's above the admin menu and any messages displayed for the page.

admin/includes/css/developers_tool_kit.css
```diff
@@ -14,3 +14,7 @@
 .dtk-contextline {font-family: Verdana, sans-serif;}
 .dtk-linenum {font-family: Verdana, sans-serif;}
 .dataTableGroupChange {border-top: 2px solid #000 !important;}
+table.results {border:2px solid;}
+table.blank {width:100%;}
```
lat9, line 18: There's no table here with a `blank` class, please remove.

If it turns out to be needed, suggest using `w-100` as the class name, since that's a Bootstrap 4/5 included class for a 100% width.

lat9 (changes requested):
Just a couple of refinements, looks good (and valid) otherwise.

admin/developers_tool_kit.php
brittainmark: yes done.

admin/developers_tool_kit.php
brittainmark: done

admin/includes/css/developers_tool_kit.css
brittainmark: done

admin/developers_tool_kit.php
brittainmark: Happy to move. As will only need the one new variable.
Is it still ok if nothing found?

lat9: If nothing's found, there's a messageStack message indicating as such and the 'header' + 'output' will just show the searching for and then found 0.

brittainmark: Excellent. Just checking you were happy.

lat9 (approved):

PR #5886, opened 2023-07-22
admin, product preview, display of description, embedded images: substitute relative image paths for shopfront

torvista:
...otherwise embedded image references in the description do not display/point to the admin images folder.

Use-case for embedded images: screenshots of software etc., in a relevant language to the description.

### DIFF
--- a/admin/includes/modules/product/preview_info.php
+++ b/admin/includes/modules/product/preview_info.php
@@ -90,10 +90,11 @@ $form_action = (isset($_GET['pID'])) ? 'update_product' : 'insert_product';
           $pInfo->products_name = $products_image_name;
         }
         if ((isset($_POST['image_delete']) && $_POST['image_delete'] == '1') || ($products_image_name == '' && PRODUCTS_IMAGE_NO_IMAGE_STATUS == '1')) {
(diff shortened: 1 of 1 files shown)


admin/includes/modules/product/preview_info.php
```diff
@@ -89,10 +89,11 @@
           $products_image_name = $_POST['img_dir'] . $_POST['products_image_manual'];
           $pInfo->products_name = $products_image_name;
         }
+        $display_products_description = str_replace('images/', DIR_WS_CATALOG_IMAGES, $display_products_description);
```
lat9, line 92: This line should be changed to
```php
$display_products_description = str_replace('src="images/', 'src="' . DIR_WS_CATALOG_IMAGES, $pInfo->products_description);
```

- Correcting the source-variable for the `str_replace`
- Limiting the replacement to the `src="` attribute of an `` tag.  Otherwise, a fully-specified `https://mysite.com/images/image_name.jpg` would have _unwanted_ replacements.

Note that this change also needs to be applied to the like-named .php file for the `document_general`, `document_product`, `product_free_shipping` and `product_music` types.

lat9 (changes requested):
See changes requested for line 92.

admin/includes/modules/product/preview_info.php
drbyte: May I also propose moving this new line 92 down to line 97/98 after the `if/else` as a new `echo` statement?

This is because this "description" var is being output in both cases, so isn't really part of the if/else.

lat9 (approved):

PR #5891, opened 2023-07-23
Consolidate product-type modules

drbyte:
This causes all the admin product-type files to look for an override inside the product-type-specific folder, but if not found then to use the parent one.

Since all inherit from the "`product`" base type, that folder is now empty by default.

**Impact on plugins:**
All plugins that overrided these files will still "work". In fact, simply dropping the plugin's file into the directory will be enough, meaning that the "core" file (like `/admin/includes/modules/collect_info.php`) can be left intact (and should NOT be touched by plugins), and the plugin only supplies the altered copy of the file, placed in the appropriate product-type's directory.

**Impact on upgrades:**
People should be encouraged to upgrade by deleting the removed files after first inspecting their own files to see if they had previously customized them in some way, or had replaced them with a version supplied by a plugin/addon.


admin/product.php
```diff
@@ -97,13 +97,29 @@
     <!-- body_text //-->
     <?php
     if ($action == 'new_product_meta_tags') {
-      require(DIR_WS_MODULES . $zc_products->get_handler($product_type) . '/collect_info_metatags.php');
+      if (file_exists(DIR_WS_MODULES . $zc_products->get_handler($product_type) . '/collect_info_metatags.php')) {
```
proseLA, line 100: it seems all of this code repetition is ripe for a refactor.  could we not create a function like:

```
function zenRequireFile($fileName)
    {
        global $zc_products, $product_type;
        if (file_exists(DIR_WS_MODULES . $zc_products->get_handler($product_type) . '/' . $fileName)) {
            require(DIR_WS_MODULES . $zc_products->get_handler($product_type) . '/' . $fileName);
        } else {
            require(DIR_WS_MODULES . $fileName);
        }
    }
```
and then call that function?  ie:

```
zenRequireFile('update_product_meta_tags.php');
```
i'm seeing 7 times the exact same code is being called within this script.  unless i am missing something...

scottcwilson:
Questions:

a) Should we stub out (instead of removing) these files?   Perhaps for one release only?
b) Would you please provide an example for documentation - perhaps something like 'add the field "cost" to the products table to capture the cost of goods sold."

admin/product.php
drbyte: I was okay with that initially since it was contained to just the one file. 
But then I found it also in one other file. Granted, in only 1 other file.
But yes, function added.

drbyte:
> Would you please provide an example for documentation - perhaps something like 'add the field "cost" to the products table to capture the cost of goods sold."

I'm not sure I understand.  Just like our other overrides, if you want to replace the default functionality with something of your own, copy the default file into the override folder and apply the changes there.
There are good examples in each of the product-type folders now in their respective `collect_info.php` files, as well as in `document_general` and `product_music`.

Adding a completely new field not currently in the db is way more involved than merely editing these particular override files. Is there something about adding such fields already in the docs?

drbyte:
> a) Should we stub out (instead of removing) these files? Perhaps for one release only?

I toyed with that idea. What do you think of only stubbing out the ones for `product`, but not for the others?

scottcwilson:
I would say do them all the same way.  Either stub out like 
```
<?php
// No longer required; will be removed in a future release 
```

or the way you suggested.  But use the same approach for all.

drbyte:
Oh, I see what you mean. Hmmmm. Pondering..

includes/functions/functions_files.php
```diff
@@ -144,6 +144,29 @@ function zen_get_module_sidebox_directory($check_file)
...
 }
 
+/**
+ * find module directory for admin product-type modules
+ * @param string $product_type
+ * @param string $filename_to_check
+ * @param bool $dir_only
+ * @return string
+ */
+function zen_get_admin_module_from_directory($product_type, $filename_to_check, $dir_only = false)
```
proseLA, line 154: how would you feel about starting to add types to the signature?  ie:

```
function zen_get_admin_module_from_directory(string $product_type, string $filename_to_check, bool $dir_only = false): string
```

drbyte: Good catch. I was cloning from other functions in the same file. I won't touch those in this PR though, to keep this canonical.

drbyte:
> // No longer required; will be removed in a future release 

Can't do that, because if the file exists then it gets loaded instead of the default, and if this is all that's in the file then that'll break things.

admin/category_product_listing.php
```diff
@@ -175,9 +175,7 @@
 
             if ($do_delete_flag) {
               //--------------PRODUCT_TYPE_SPECIFIC_INSTRUCTIONS_GO__BELOW_HERE--------------------------------------------------------
-              if (file_exists(DIR_WS_MODULES . $zc_products->get_handler($product_type) . '/delete_product_confirm.php')) {
-                require DIR_WS_MODULES . $zc_products->get_handler($product_type) . '/delete_product_confirm.php';
-              }
+              require zen_get_admin_module_from_directory($product_type, 'delete_product_confirm.php');
```
drbyte, line 178: Not sure if this was a bug, but this never fired before, because (probably nobody) ever had an override copy of this file.

Or, does that mean what I've done here actually CAUSES a bug, because it WILL now fallback to the default when it was only intended to run IF there was a file specific to the product-type?

drbyte: @zcwilt Thoughts?

drbyte: I think I answered my own question: it was a bug, but completely obscured because of duplicated code.

I've removed the duplication in favor of having it fall back onto the module-specific override behavior.

Issue #5727, opened 2023-05-07
Add MPN to products table and


torvista:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Important: Explain what "Business Problem" is being solved by this solution. 
What kind of merchant (small business, retail-only, digital-products-only, etc) would most benefit from this solution. Who might "not" benefit much?

**Describe alternatives you've considered**
If you have working code that could be a possible solution, include a link to where that code can be viewed.

**Additional context**
Add any other context or screenshots about the feature request here. Include URLs where possible solutions can be seen, if any.

lat9:
Good idea, but since it changes the database schema, it'll need to be deferred until the next unsuffixed version.

scottcwilson:
I want to include this in the next release.

lat9:
Noting also that it makes no sense to add an MPN to a `products_attributes` record.  An MPN for an attributed product would apply to the product+_**attribute-combinations**_, e.g. shirt+size+color,  Having a value in each attributes' record is a non-starter.

drbyte:
Remember that you may want the `orders_products` table to also store the information as it was at the time of an order.

lat9:
Does the inclusion of `products_mpn` necessitate the addition of a "SHOW_PRODUCT{_type}_INFO_MPN" setting in the `product_type_layout` table and related changes to the various `tpl_product{_type}_info_display.php` templates?

![image](https://github.com/zencart/zencart/assets/2685585/4046da8e-c114-47fc-9fb8-7cc1bd974018)

scottcwilson:
I would vote no.  People who want it can just update their templates, which they'll have to do anyhow.

lat9:
How about the function addition?  Noting that any database update would start with the `products_mpn` value as `NULL` to distinguish 'unset' vs. not supplied.

scottcwilson:
Sure, function is fine.

lat9:
Sigh, that addition will also necessitate updates to the admin's various product types' `collect_info.php` modules as well as the `update_product.php` module.

scottcwilson:
Yep.

proseLA:
> Remember that you may want the `orders_products` table to also store the information as it was at the time of an order.

i disagree.  MPNs are assigned by the manufacturer, and therefore rarely change for a specific product.  and if there is a change, in theory it is a different product.  ensuring that you have the MPN at the time of the order in the orders_products table is a wasted bit of data, at least to me.

[...]

Issue #5601, opened 2023-02-12
Adding the `shipping_tax_rate` to the orders


lat9:
I can't believe that I didn't submit this before, but there are various elements of processing (e.g. coupons) and plugins (e.g. Edit Orders) that can't properly do their job without knowing what (if any) tax percentage was applied to an order's shipping.

I propose that a `shipping_tax_rate` field `decimal(15,4) DEFAULT NULL be added to the `orders` table.  The `shipping` class would be updated to record the selected shipping method's tax-rate so that a decomposition of the order can be performed without various calculations that result in the penny-off rounding errors.

In the Zen Cart base, the coupon order-total could use that value when it's calculating the deduction for a shipping-included order.

lat9:
It's already there; I'd forgotten to submit the change request so that the value would be in the base.

drbyte:
Closed by #5893

lat9:
Re-opening, since only the database portion has been completed.

PR #5893, opened 2023-07-24
Initial v2.0.0 upgrade

scottcwilson:
Satisfies db portions of #5601 and #5727

Tested with upgrade and fresh install.

### DIFF
--- a/includes/version.php
+++ b/includes/version.php
@@ -14,8 +14,8 @@
  */
 
 define('PROJECT_VERSION_NAME', 'Zen Cart');
-define('PROJECT_VERSION_MAJOR', '1');
-define('PROJECT_VERSION_MINOR', '5.8a');
+define('PROJECT_VERSION_MAJOR', '2');
+define('PROJECT_VERSION_MINOR', '0.0');
 define('PROJECT_VERSION_PATCH1', '');
 define('PROJECT_VERSION_PATCH2', '');
 define('PROJECT_VERSION_PATCH1_SOURCE', '');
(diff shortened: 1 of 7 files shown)


zc_install/sql/install/mysql_zencart.sql
```diff
@@ -1156,6 +1157,7 @@ CREATE TABLE orders_products (
   orders_id int(11) NOT NULL default '0',
   products_id int(11) NOT NULL default '0',
   products_model varchar(32) default NULL,
+  mpn varchar(32) DEFAULT NULL,
```
lat9, line 1160: Should be `products_mpn`.

zc_install/sql/install/mysql_zencart.sql
```diff
@@ -1620,6 +1622,7 @@ CREATE TABLE products (
   products_type int(11) NOT NULL default '1',
   products_quantity float NOT NULL default '0',
   products_model varchar(32) default NULL,
+  mpn varchar(32) DEFAULT NULL,
```
lat9, line 1625: Should be `products_mpn`.

lat9 (changes requested):
Two name changes

includes/version.php
```diff
@@ -14,8 +14,8 @@
  */
 
 define('PROJECT_VERSION_NAME', 'Zen Cart');
-define('PROJECT_VERSION_MAJOR', '1');
-define('PROJECT_VERSION_MINOR', '5.8a');
+define('PROJECT_VERSION_MAJOR', '2');
+define('PROJECT_VERSION_MINOR', '0.0');
```
lat9, line 18: Shouldn't this be `0.0-alpha1`?

scottcwilson: I'd rather not go down that road unless the group insists.  We're trying to move away from letters, and this is one more thing that has to be remembered on release date that has been messed up in the past.

zc_install/sql/install/mysql_zencart.sql
scottcwilson: Yep.  TY.

zc_install/sql/install/mysql_zencart.sql
scottcwilson: Yep.  TY.

includes/version.php
lat9: That's part of the standardized versioning.  As we move through alpha/beta versions of v2.0.0, I believe that the alpha/beta designation is quite important.

drbyte:
I agree that this needs to be named "-alpha" in the db.

Since it touches several files, I will do it as a separate PR after merging.

torvista:
As of today, version check is incorrect
image

PR #5896, opened 2023-07-25
Updating per requested changes to #5843

### LINKED ISSUE #5799: 1.5.8b? Additional image code no longer reachable
**Describe the bug**
Because of incorporation of PR #5656, unable to display additional images without having a primary image in a default ZC install. This includes that no additional images notifiers are executed as a result of a change to the default template for all recognized product types.

**Version**
Zen Cart version: 1.5.8b? (code modified after the github tag of 1.5.8a)
PHP version: Anything compatible
MySQL version: Anything compatible
Plugins that may be related: 
What older/newer versions were also tested? 1.5.8a and below

**To Reproduce**
Install the Zen Cart software that is available on the current v158 branch: https://github.com/zencart/zencart/tree/v158 that incorporates PR #5656.

Install Additional Product Images plugin: https://www.zen-cart.com/downloads.php?do=file&id=1063
As necessary upgrade for any php related issues, though cursory glance doesn't appear to have any obvious specific issues.
Add an additional image using the plugin to a product that has a main image.
Navigate the store to look at the results of that product having both a main image and an additional image.
Remove the primary image from that product.
Attempt to navigate the product again and notice that the additional image(s) are no longer presented.
Return to the admin side and note how additional, additional images can be added to the product that no longer has a main image.

**Expected behavior**
Able to independently address the main image and the additional images on any product page without requiring that a path to a main image file must exist and if it must exist, the a way to bypass the html code that is generated to display the image even though the expectation may be that no image would be displayed for the particular product. This has not been a requirement before, there is no identified level of improvement gained by this being made now part of the core code and is done at the template level rather than either deeper in the code or not at all. 

**Screenshots**
None that seem would add much value.

**Debug/Error Logs**
None generated specifically in relation to this issue.
 
 - Operating System: Doesn't matter
 - Browser [e.g. chrome, firefox] and version number: Doesn't matter

**Additional context**
There is another "similar" plugin about which I know exists; however, I do not know how it interfaces with the store.

lat9:
... also updating the previously-forgotten `template_default` modules.

Fixes #5799 

Co-authored-by: @mc12345678

### DIFF
--- a/includes/extra_datafiles/dist-site_specific_overrides.php
+++ b/includes/extra_datafiles/dist-site_specific_overrides.php
@@ -52,3 +52,12 @@
 // true ..... no link will be created.
 // false .... a <link> element will load the v4-shims.min.css file.
 //$disableFontAwesomeV4Compatibility = true;
+
+// -----
+// Indicate whether or not a product's additional images should be displayed even if the product
+// has no 'main' image defined in the database.
+//
+// true .... Additional images are displayed regardless the definition of a product's 'main' image
+// false ... Additional images are displayed *only* if a product's 'main' image is defined (i.e. not `''`); this is the default.
+//
+//$enable_additional_images_without_main_image = false;
--- a/includes/templates/responsive_classic/templates/tpl_document_general_info_display.php
+++ b/includes/templates/responsive_classic/templates/tpl_document_general_info_display.php
@@ -47,7 +47,7 @@ require($template->get_template_dir('/tpl_products_next_previous.php',DIR_WS_TEM
 <div id="pinfo-left">
 <!--bof Main Product Image -->
 <?php
-  if (!empty($products_image)) {
+  if (!empty($products_image) || !empty($enable_additional_images_without_main_image)) {
   ?>
 <?php
 /**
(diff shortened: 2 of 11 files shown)


mc12345678:
Remarkable. 

This ***does not*** fix the referenced issue, instead breaking #5799.

Took a developed functional PR, providing developer breadcrumbs in a meaningful location; narrowed down an intended useful value to become just a hidden boolean; and then, when implemented, stuffs it all up creating debug logs. Why? Ha, because didn't want to change "a lot" of files though changed just as many now in a completely breaking way.

Shouldn't this have gotten some level of independent review before incorporation into the core code? 

I take no credit in this broken, non-functional PR that now creates debug logs when used and personally have *miniscule* appreciation for so-called credit. 

The changes in the `includes/templates/default_template/` folder were unnecessary, in resolution to this issue and to the original problem code. Further, the ignored original solution specifically prevented the error logs of this now useless feature.

This has outright become a complete waste and should be removed. Just forget that I was looking ahead at offering solutions to a created problem. The inconsiderate control of this repo is too much. The half-baked, rushed coding shows no development consideration or planning and is a complete disappointment. 

I resign to report formally available plugin functionality has been unnecessarily removed with no core functional replacement or override provided/accepted. Further, functionality loss resolution should uselessly be sought by the core code maintainers.

PR #4943, opened 2022-06-29
admin, server_info: add database variables info from plugin

torvista:
based on
https://github.com/lat9/database_server_info/issues
as discussed
 https://github.com/zencart/zencart/issues/4924

PR #5902, opened 2023-07-28
admin, server info: correct $nbsp;

torvista:
corrects this 
e.g._
image


drbyte:
Good catch. Problem was created in https://github.com/zencart/zencart/pull/4943

PR #5903, opened 2023-07-28
Rename temporary helper files

scottcwilson:
Using spaces in filenames only "works" in Windows.  It's a nightmare for Unix/Linux/MacOS users.  Please use underscores.


drbyte:
Um ... these are temporary helpers, not important files.

And, they were created on a Mac 🤣 

but, if it somehow helps everyone, alright

PR #5904, opened 2023-07-28
Enable additional actions for Catalog :: Manufacturers

lat9:
... by adding a notification.


drbyte:
I gather the intent would be to interact with global variables, since none are being passed for update?

lat9:
Right, if there's a separate action then there was another form to gather posted data in support of that form's action.

PR #5907, opened 2023-07-30
Refactoring admin/customers.php

lat9:
Refactoring `admin/customers.php`:

- Use `===` where warranted.  Note that the customers_id element returned by the `Customer` class is an integer!
- Use the null coalesce (`??`) operator where warranted.
- Use short-array syntax on those missed in the reformatting.
- Remove list_addresses_done action; simply redirect to the listing from the list_addresses action.
- Cast posted values that _should be_ integers to integers.
- Use "LIMIT 1" on updates for performance on sites with a large number of customers.
- Preserve search keywords on list_addresses, status, edit, delete and pwreset actions.
- Cast fax_number and suburb to string for htmlspecialchars (they can be null when editing and were triggering PHP Deprecated logs).
- Use ctype_digit instead of is_numeric for status action processing.
- Cast customers_authorization to a string instead of using strval.
- Remove duplicated page= parameter from info-icon link.

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -37,16 +37,10 @@ if (!empty($action)) {
             $customer = new Customer($_GET['cID']);
             $addressArray = $customer->getData('addresses');
             break;
-        case 'list_addresses_done':
-            $action = '';
-            zen_redirect(
-                zen_href_link(FILENAME_CUSTOMERS, 'cID=' . (int)$_GET['cID'] . '&page=' . $_GET['page'], 'NONSSL')
-            );
-            break;
         case 'status':
-            if (isset($_POST['current_status']) && is_numeric($_POST['current_status'])) {
-                if ($_POST['current_status'] == CUSTOMERS_APPROVAL_AUTHORIZATION) {
-                    if (CUSTOMERS_APPROVAL_AUTHORIZATION == 1 || CUSTOMERS_APPROVAL_AUTHORIZATION == 2) {
+            if (isset($_POST['current_status']) && ctype_digit($_POST['current_status'])) {
+                if ($_POST['current_status'] === CUSTOMERS_APPROVAL_AUTHORIZATION) {
+                    if (CUSTOMERS_APPROVAL_AUTHORIZATION === '1' || CUSTOMERS_APPROVAL_AUTHORIZATION === '2') {
                         $customers_authorization = 0;
                     } else {
                         $customers_authorization = 4;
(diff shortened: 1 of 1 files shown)


admin/customers.php
```diff
@@ -168,17 +159,17 @@
...
             $entry_state_error = false;
-            if (ACCOUNT_STATE == 'true') {
+            if (ACCOUNT_STATE === 'true') {
                 $entry_state_has_zones = count(zen_get_country_zones($entry_country_id)) > 0;
                 if ($entry_state_has_zones) {
                     $zone_query = $db->Execute(
                         "SELECT zone_id
                            FROM " . TABLE_ZONES . "
-                           WHERE zone_country_id = " . (int)$entry_country_id . "
+                           WHERE zone_country_id = " . $entry_country_id . "
```
drbyte, line 168: I've always preferred to keep the cast, even if redundant, inside SQL queries ... because then there's never any doubt about whether it's been properly sanitized .... and it doesn't require searching the entire file or the files that called this file in order to find out whether it was already the right type, and that 3rd party plugin code can not have tampered with it in any way that could leave any bSLQi vulnerabilities.

admin/customers.php
```diff
@@ -253,20 +244,20 @@
...
-                        )),
+                        'value' => ($customers_dob === '0001-01-01 00:00:00') ?
+                            '0001-01-01 00:00:00' : zen_date_raw($customers_dob),
                         'type' => 'date'
                     ];
                 }
 
                 $zco_notifier->notify('NOTIFY_ADMIN_CUSTOMERS_CUSTOMER_UPDATE', $customers_id, $sql_data_array);
-                $db->perform(TABLE_CUSTOMERS, $sql_data_array, 'update', "customers_id = " . (int)$customers_id);
+                $db->perform(TABLE_CUSTOMERS, $sql_data_array, 'update', "customers_id = " . $customers_id . " LIMIT 1");
```
drbyte, line 254: ditto on keeping the cast.

admin/customers.php
```diff
@@ -253,20 +244,20 @@
...
 
                 $db->Execute(
                     "UPDATE " . TABLE_CUSTOMERS_INFO . "
                       SET customers_info_date_account_last_modified = now()
-                      WHERE customers_info_id = " . (int)$customers_id
+                      WHERE customers_info_id = " . $customers_id . "
```
drbyte, line 259: ditto on keeping the cast.

admin/customers.php
```diff
@@ -357,14 +348,14 @@
                     TABLE_ADDRESS_BOOK,
                     $sql_data_array,
                     'update',
-                    "customers_id = '" . (int)$customers_id . "' AND address_book_id = '" . (int)$default_address_id . "'"
+                    "customers_id = " . $customers_id . " AND address_book_id = " . $default_address_id . " LIMIT 1"
```
drbyte, line 351: ditto on keeping the cast.

admin/customers.php
```diff
@@ -357,14 +348,14 @@
...
                 );
 
                 if (isset($_POST['customer_groups']) && is_array($_POST['customer_groups'])) {
                     zen_sync_customer_group_assignments($customers_id, $_POST['customer_groups']);
                 }
 
-                zen_record_admin_activity('Customer record updated for customer ID ' . (int)$customers_id, 'notice');
+                zen_record_admin_activity('Customer record updated for customer ID ' . $customers_id, 'notice');
```
drbyte, line 358: ditto on keeping the cast.

admin/customers.php
lat9: I agree on string-related values which need to be properly sanitized, but the processing prior to this (and the following) SQL statements has guaranteed that the values are already of an integer type.  If some rogue add-on has made modifications to set one or more of the values to a non-integer value, the SQL query/update will fail.

drbyte: >  If some rogue add-on has made modifications to set one or more of the values to a non-integer value, the SQL query/update will fail.

It may fail, but if the non-integer value happens to be a string containing a rogue SQL injection, while the intended core code payload may fail, unintended side-effects may take place even without knowing.

lat9:
Changes submitted; some additional `===` applied.

PR #5910, opened 2023-07-31
Add ability to count distinct products for category counts

brittainmark:
Currently, the category count counts the number of products in each sub category individually. This means that if you have a product in more than one sub category of the parent category, it is counted once for each category that it is in. If, like me, you have products that you put into many sub categories to allow searching by a feature of the product, the category count can get highly inflated. 
This mod counts the distinct products_id(s) in a category and all its sub categories.


scottcwilson:
Interesting idea.  Thanks for your contribution!

PR #5912, opened 2023-08-01
Countries zones names update

piloujp:
Update to Italian, and Austria zones.
Swiss zones deleted.
See issue #5417


drbyte:
Thanks!

drbyte:
Reworking the syntax of this in a new PR ...

zc_install can't parse table names if they're not in the first 3-4 words of a "new line"
So, it's best if keywords like FROM or LEFT JOIN start on a new line.

Also, MariaDB 10.3 doesn't like the INSERT WHERE NOT EXISTS syntax. See #6292

PR #5914, opened 2023-08-02
Fix misspelled GV Release button text

drbyte:
On **Admin->Discounts->GV Queue** page:

before:
Screen Shot 2023-08-01 at 8 03 21 PM

after:
Screen Shot 2023-08-01 at 8 04 54 PM


scottcwilson:
Thank you!

PR #5917, opened 2023-08-02
Adding listbox role to rest of admin files

### LINKED ISSUE #5905: `role="button"` on `<tr>` tags results in invalid HTML (e.g. admin/customers.php)
What's the motivation for the `role="button"` addition on various `<tr>` tags in `customers.php` (probably repeated elsewhere):
https://github.com/zencart/zencart/blob/9dff1c8397def4e2408fed49a594be622450e14f/admin/customers.php#L1537-L1542

I ask because those result in various HTML validation issues

- The role attribute must not be used on a tr element which has a table ancestor with no role attribute, or with a role attribute whose value is table, grid, or treegrid.
- The element button must not appear as a descendant of an element with the attribute role=button.
- The element a must not appear as a descendant of an element with the attribute role=button.

brittainmark:
Follow on from #5915 same fixes #5905 for rest of admin

additionally customer_groups.php had missing space.

### DIFF
--- a/admin/configuration.php
+++ b/admin/configuration.php
@@ -104,7 +104,7 @@ if ($gID == 7) {
       <div class="row">
         <div class="col-xs-12 col-sm-12 col-md-9 col-lg-9 configurationColumnLeft">
 
-          <table class="table table-hover">
+          <table class="table table-hover" role="listbox">
             <thead>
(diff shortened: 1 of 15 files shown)


drbyte:
Thanks!

PR #5918, opened 2023-08-02
Fix version-check compare for minor number

drbyte:
While it was correctly handling `1 < 2`, it wasn't subsequently correctly handling `5.8 < 0.0`.


drbyte:
Addresses comment by @torvista at https://github.com/zencart/zencart/pull/5893#issuecomment-1661857546

PR #5921, opened 2023-08-03
Refactoring init_add_crumbs.php

lat9:
- Using null-coalesce where appropriate
- Use `count` instead of `sizeof`
- Using multiple variables on `isset` where appropriate
- Use foreach instead of while/$db->MoveNext on multiple-record query.
- Gather a single query where either a single or no record is expected to be returned
- Use `===` instead of `==`
- Simplify the 'get-terms-to-filter' section
- Ensure that `$cPath` is set for the product-name portion
- Use $result->EOF instead of $result->RecordCount()

Next up, adding a notification so that **_Ceon URI Mappings_** can discontinue its use of an unwanted override of this file!

### DIFF
--- a/includes/init_includes/init_add_crumbs.php
+++ b/includes/init_includes/init_add_crumbs.php
@@ -63,7 +63,7 @@ foreach ($get_terms as $next_get_term) {
 /**
  * add the products name to the breadcrumb trail
  */
-if (isset($_GET['products_id'], $cPath)) {
+if (isset($_GET['products_id'])) {
     $productname_query =
         "SELECT products_name
            FROM " . TABLE_PRODUCTS_DESCRIPTION . "
--- a/includes/init_includes/init_add_crumbs.php
+++ b/includes/init_includes/init_add_crumbs.php
@@ -16,23 +16,21 @@ $breadcrumb->add(HEADER_TITLE_CATALOG, zen_href_link(FILENAME_DEFAULT));
 /**
  * add category names or the manufacturer name to the breadcrumb trail
  */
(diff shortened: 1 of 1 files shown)


includes/init_includes/init_add_crumbs.php
```diff
@@ -47,33 +45,33 @@
...
     }
-    $get_terms->MoveNext();
 }
 
 /**
- * add the products model to the breadcrumb trail
+ * add the products name to the breadcrumb trail
  */
-if (isset($_GET['products_id'])) {
+if (isset($_GET['products_id'], $cPath)) {
```
drbyte, line 66: This could be a breaking-change, right?  Granted, it's probably rare that `$cPath` won't be defined.

Perhaps if the `$cPath` is missing, we should just skip the `"&cPath=" .$cPath` part of the breadcrumb URL?

(The canonical URL of a product page doesn't include a cPath, so doesn't introduce any new URLs to worry about.)

lat9: It _should_ be set; perhaps better to leave as-is and a PHP Warning will be issued if the `$cPath` isn't set?

lat9: Your call, @drbyte.  I've got no skin in the game.

drbyte: Let's leave it as-is. I wanna look at some related code before making a call on it. 
No need to have this hold up the other things you're doing.

lat9: As-is in the PR or amend the PR to remove that addition?

Issue #5923, opened 2023-08-04
Admin, Extras: all pages missing Insert button


torvista:
php7.4-8.3
ZC2.0.0.

Delete all demo items on any page so there are no records: no Insert button.

proseLA:
#5349

scottcwilson:
Thanks guys. I thought we had fixed this but apparently not.

proseLA:
> Thanks guys. I thought we had fixed this but apparently not.

there are 4 more files that need fixing for this issue that "we" introduced.

simon1066:
As long as there's no rush, I don't mind making the edits to the other files - using @proseLA's media_types.php as an example.

When this issue arose for manufacturers.php #5699, @drbyte made changes that moved the `&` out of `'action=new'` and into `$current_mid` (`$mType_parameter` in this case), #5703

> You'll need to move the `&` out of `'&action=new'`, into `$current_mid`, else you'll have double `&&` in some cases and a missing one in other cases.

Is that something that needs to be done here?

torvista:
Once fixed, perhaps a discussion is warranted (in the forum) who still uses this stuff / splitting this stuff out to a plugin.

simon1066:
Now that the changes are finalised and merged, #5924, I will look at doing the same for the other affected files under 'Extras'

PR #5924, opened 2023-08-04
Fix for empty media types

proseLA:
see #5923 

addresses problems created by #5349 

someone else can handle the other scripts.

### DIFF
--- a/admin/media_types.php
+++ b/admin/media_types.php
@@ -10,179 +10,181 @@ require('includes/application_top.php');
 $action = (isset($_GET['action']) ? $_GET['action'] : '');
 
 if (!empty($action)) {
-  switch ($action) {
-    case 'insert':
-    case 'save':
-      if (isset($_GET['mID'])) {
-        $type_id = zen_db_prepare_input($_GET['mID']);
-      }
-      if (isset($_POST['type_ext'])) {
-        $type_ext = zen_db_prepare_input($_POST['type_ext']);
-      }
-      if (isset($_POST['type_name'])) {
-        $type_name = zen_db_prepare_input($_POST['type_name']);
-      }
+    switch ($action) {
+        case 'insert':
+        case 'save':
+            if (isset($_GET['mID'])) {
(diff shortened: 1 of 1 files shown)


torvista:
Doing an add, save, edit, save, I get an error on the save.

> [04-Aug-2023 17:25:24 Europe/Madrid]
> ADMIN
> Request URI: ADMIN/index.php?cmd=media_types&page=1&action=save
> IP address: 127.0.0.1, Language id 1
> #1  require(ADMIN/media_types.php) called at [ADMIN/index.php:21]
> --> PHP Notice: Undefined variable: type_id in ADMIN/media_types.php on line 40.
> 
> [04-Aug-2023 17:25:24 Europe/Madrid]
> ADMIN
> Request URI: ADMIN/index.php?cmd=media_types&page=1&action=save
> IP address: 127.0.0.1, Language id 1
> #1  require(ADMIN/media_types.php) called at [ADMIN/index.php:21]
> --> PHP Notice: Undefined variable: type_id in ADMIN/media_types.php on line 43.

proseLA:
not real sure what that has to do with the insert not being there.

admin/media_types.php
```diff
@@ -10,173 +10,181 @@
...
+                                echo '<a href="' . zen_href_link(FILENAME_MEDIA_TYPES, zen_get_all_get_params(['mID']) . 'mID=' . $media_type['type_id']) . '">' . zen_image(DIR_WS_IMAGES . 'icon_info.gif', IMAGE_ICON_INFO) . '</a>';
+                            }
+                            ?>
+                        </td>
+                        </tr>
+                        <?php
+                    }
+                    $mType_parameter = '';
+                    if (!$media_types->EOF) {
+                        $mType_parameter = '&mID=' . $mInfo->type_id;
```
drbyte, line 113: Is it possible that `$mInfo` still could be undefined here?  
Ref line 87

proseLA: i really do not think so; but i will adjust the code to reflect that possibility.

PR #5926, opened 2023-08-05
Change customer groups look and feel to use font Awesome

brittainmark:
Changed row action to just select row (remove &action="edit")
Removed edit delete buttons from rows
Change icon so Arrow is selected row
Add information icon to not selected rows
Before
customerGroupsBefore
After
customerGroupsAfter

### DIFF
--- a/admin/customer_groups.php
+++ b/admin/customer_groups.php
@@ -114,19 +114,31 @@ if (!empty($action)) {
                         $role = 'role="option" aria-selected="true"';
                     } 
                     ?>
-                    <tr <?php echo $class_and_id; ?> onclick="document.location.href='<?php echo zen_href_link(FILENAME_CUSTOMER_GROUPS, $href_page_param . 'gID=' . $group['group_id'] . '&action=edit'); ?>'" <?php echo $role;?>>
+                    <tr <?php echo $class_and_id; ?> onclick="document.location.href='<?php echo zen_href_link(FILENAME_CUSTOMER_GROUPS, $href_page_param . 'gID=' . $group['group_id']); ?>'" <?php echo $role;?>>
                         <td class="dataTableContent text-center"><?php echo $group['group_id']; ?></td>
                         <td class="dataTableContent"><?php echo $group['group_name']; ?></td>
                         <td class="dataTableContent text-center"><?php echo $group['customer_count']; ?></td>
                         <td class="dataTableContent"><?php echo $group['group_comment']; ?></td>
                         <td class="dataTableContent text-right"><div>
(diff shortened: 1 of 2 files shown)


drbyte:
Doesn't this make it less user friendly?

proseLA:
> Doesn't this make it less user friendly?

beyond user friendly, i do not think we really have a standard for look and feel on the admin.  at one point, we were using glyficons as well as font-awesome.  personally i think we should move away from the use of the image icons; especially since we recently had a bunch of work put into upgrading to fa 6.

i think the standard should be the use of bootstrap and the font-awesome icons.  

the use of these image icons, as seen in this PR, are part of what make the look and feel seem so dated...  at least to me.

brittainmark:
The issue for me was that the arrow on all the other screens indicated the selected row. But on this screen it indicated the rows that were not selected.

Happy to change to Font Awsome.
Also happy to restore the buttons edit and delete I just feel that if my simple brain struggels with the format others may to.

Other options would be
1) remove the arrow icon completely and just leave the edit delete buttons.
2) put the arrow only on the record that has been selected as the other screens (but use font awsome).

As for user friendly I suppose you have to make one more click to select the rwo before you edit.

For a minimul change I would suggest option 2. but open to any sugestions.

drbyte:
I agree with the fix to the arrows being on the wrong row. And to the other indicators of "selected" status.
And I agree we should use fontawesome across the whole of the admin (aside: that'll be complicated for the cat/prod button bar, and is a completely separate topic for its own isolated PR, if tackled)

I was mainly objecting to the dropping of the edit/delete buttons in the row. I'd be happy for them to become merely icons if the text-buttons are too-large-looking. My apologies for not being more specific in my first comment.

brittainmark:
Not a problem. Will look at making those changes. 
Are we happy with the following icons (Image is example of icons)
image
```
<i class="fa-solid fa-trash-can fa-2x fa-fw txt-red align-middle"></i>
<i class="fa-solid fa-circle-info fa-2x fa-fw txt-black align-middle"></i>
<i class="fa-solid fa-pencil fa-2x fa-fw txt-blue align-middle"></i>
<i class="fa-solid fa-caret-right fa-2x fa-fw txt-navy align-middle"></i>
```

I would also suggest adding to languages/lang.english.php

'ICON_CURRENT' => 'Selected Row',

brittainmark:
Redone now look like.
image
I have not looked at changing to bootstrap, just font awesome and added ICON_SELECTED to indicate selected row.

drbyte:
I'm satisfied with that look!

drbyte:
> Sorry being thick. What code changes?

I meant: I like the layout proposed in the revised picture you last posted. 
But the code in this PR is for the original proposal. 
You haven't updated this PR with the code changes to make it work like the picture you posted.

brittainmark:
Sorry, I don't know what happened there, it looks like the changes did not get pushed up to the branch. I have retested and this now appears to be correct.

drbyte:
Looks good. Thanks.

PR #5928, opened 2023-08-06
early return for ot_coupon and add notifier

### LINKED ISSUE #5925: ot_coupon does not need to process if no coupon is entered.
is there a reason why we do not just exit if there is no cc_id?  ie:

```
            if (empty($_SESSION['cc_id'])) {
                return;
            }
```
at the beginning of this method.
 https://github.com/zencart/zencart/blob/7afd3a914c8ba09ecdc1ce46b038eeaf027df1dc/includes/modules/order_total/ot_coupon.php#L133-L138

proseLA:
fixes #5925 and moves notifier to correct place.  notifier was added AFTER v158a release.  intention was to allow a observer to prevent coupon box from appearing during checkout.

i am not opposed to changing the notifier name if so desired.

### DIFF
--- a/includes/modules/order_total/ot_coupon.php
+++ b/includes/modules/order_total/ot_coupon.php
@@ -133,6 +133,11 @@ class ot_coupon extends base
     function process()
     {
         global $order, $currencies;
+
+        if (empty($_SESSION['cc_id'])) {
+            return;
+        }
+
         $od_amount = ['tax' => 0, 'total' => 0];
 
         $order_total = $this->get_order_total(isset($_SESSION['cc_id']) ? $_SESSION['cc_id'] : '');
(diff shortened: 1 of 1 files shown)


drbyte:
> not opposed to changing the notifier name if so desired

Ya, it'd probably be wise to change it, as _START implies some other context. Probably _CREDIT_SELECTION would be more logical.

lat9:
> > not opposed to changing the notifier name if so desired
> 
> Ya, it'd probably be wise to change it, as _START implies some other context. Probably _CREDIT_SELECTION would be more logical.

... except that the processing would be similar and there _might_ be current usage of the _START version.  Better, IMO, to leave _START (and mark it deprecated) and add _CREDIT_SELECTION.

drbyte:
> notifier was added AFTER v158a release

> there _might_ be current usage of the _START version

Yes there's a very minor edge-case where that might be possible if someone's using code from an unreleased version.

No harm in leaving the _START one in the constructor. 
No need for deprecating it. If it's not used, it'll be ignored.

lat9:
> > notifier was added AFTER v158a release
> 
> > there _might_ be current usage of the _START version
> 
> Yes there's a very minor edge-case where that might be possible if someone's using code from an unreleased version.
> 

Yeah, like me!

PR #5929, opened 2023-08-07
store paypal payment date in local time fixes #5927

### LINKED ISSUE #5927: paypal time in paypal table...
[as referenced here](https://www.paypal.com/us/cshelp/article/what-format-does-the-api-timestamp-use-ts1085#:~:text=Each%20API%20response%20contains%20a,to%20as%20%22Zulu%20time.%22), paypal always uses ZULU time.

we are then posting that time:

https://github.com/zencart/zencart/blob/4d8945cf2732ec40c1a1a5f009eff4b12fdeb33e/includes/modules/payment/paypal/paypal_functions.php#L285

after converting it here:

https://github.com/zencart/zencart/blob/4d8945cf2732ec40c1a1a5f009eff4b12fdeb33e/includes/modules/payment/paypal/paypal_functions.php#L13-L22

is anyone else bothered by that table storing timestamps from different timezones?  or is it just me?

date_added in the same table makes use of the mysql now function.

https://github.com/zencart/zencart/blob/4d8945cf2732ec40c1a1a5f009eff4b12fdeb33e/includes/modules/payment/paypal/paypal_functions.php#L287

the [now function](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_now), The value is expressed in the session time zone.

proseLA:


drbyte:
Your logic looks good.

I'm just having trouble testing it 🤪...
Maybe I should be looking at this with fresher eyes, but when I pass a Zulu formatted date (eg: `2020-08-05T00:00:00Z` or `2014-10-15T16:42:56Z`) into the original `datetime_to_sql_format()` function (even before your changes), it badly fails to parse it. Obviously I'm missing something. Will have to dig more tomorrow.

proseLA:
> Your logic looks good.
> 
> I'm just having trouble testing it zany_face... Maybe I should be looking at this with fresher eyes, but when I pass a Zulu formatted date (eg: `2020-08-05T00:00:00Z` or `2014-10-15T16:42:56Z`) into the original `datetime_to_sql_format()` function (even before your changes), it badly fails to parse it. Obviously I'm missing something. Will have to dig more tomorrow.

yeah, i had that same problem.

i only have one client using paypal, and they use express.  none of their transactions seem to go through that `datetime_to_sql_format()` method.  there are some calls to it through the handler, but the situations seem to be real edge case stuff from what i can tell.

my guess is that method is no longer called.

there are some direct calls to it from paypal standard.  [but we know how ZC feels about standard.](https://docs.zen-cart.com/user/payment/paypal_standard/) 

in addition, that method is close to 20 years old without having been updated (at least that is what it looks like to me).  it's hard to imagine it still works...

perhaps someone else can weigh in on its use.

but just in case we are missing something, i added the logic.

drbyte:
Turns out, for `IPN`s the `POST`ed `payment_date` is formatted like:

`[payment_date] => 15:37:58 May 18, 2020 PDT`

What are your thoughts on the following then?

```diff
  function datetime_to_sql_format($paypalDateTime) {
      //Copyright (c) 2004 DevosC.com
+     // converts from '15:37:58 May 18, 2020 PDT' to '2020-05-18 15:37:58'
      $months = array('Jan' => '01', 'Feb' => '02', 'Mar' => '03', 'Apr' => '04', 'May' => '05',  'Jun' => '06',  'Jul' => '07', 'Aug' => '08', 'Sep' => '09', 'Oct' => '10', 'Nov' => '11', 'Dec' => '12');
      $hour = substr($paypalDateTime, 0, 2);$minute = substr($paypalDateTime, 3, 2);
      $second = substr($paypalDateTime, 6, 2);
      $month = $months[substr($paypalDateTime, 9, 3)];
      $day = (strlen($day = preg_replace("/,/" , '' , substr($paypalDateTime, 13, 2))) < 2) ? '0'.$day: $day;
      $year = substr($paypalDateTime, -8, 4);
      if (strlen($day)<2) $day = '0'.$day;
+     $tz = substr($paypalDateTime, -3);
-     return ($year . "-" . $month . "-" . $day . " " . $hour . ":" . $minute . ":" . $second);
-     return convertToLocalTime($year . "-" . $month . "-" . $day . " " . $hour . ":" . $minute . ":" . $second);
+     return convertToLocalTime($year . "-" . $month . "-" . $day . " " . $hour . ":" . $minute . ":" . $second, $tz);
   }

-   function convertToLocalTime($dateTime)
+   function convertToLocalTime(string $dateTime, string $fromTz = 'UTC')
   {
-      $localDateTime = new DateTime($dateTime, new DateTimeZone('UTC'));
+      $localDateTime = new DateTime($dateTime, new DateTimeZone($fromTz));
       $loc = (new DateTime)->getTimezone();
       $localDateTime->setTimezone($loc);
       return $localDateTime->format('Y-m-d H:i:s');
   }
```

PayPalDP:
- DoDirectPayment: no timestamp provided. Generated on line 986: `$this->payment_time = date('Y-m-d h:i:s');`
- Error responses: `[TIMESTAMP] => 2023-08-08T19:05:48Z`
- Line 999 handles 'TIMESTAMP': `$this->payment_time = urldecode($response['TIMESTAMP']);`
- GetTransactionDetails: `[SETTLE_DATE] => 2023-08-08 08:19:58`, which is not stored by ZC

PayPal  EC:
- SetExpressCheckout/GetExpressCheckoutDetails/DoExpressCheckout: `[TIMESTAMP] => 2023-08-08T18:43:35Z`
- step EC5 `[TIMESTAMP] => 2023-08-06T01:51:37Z`
- DoExpressCheckout: `[PAYMENTINFO_0_ORDERTIME] => 2023-08-08T18:43:36Z`, which correlates to `$this->payment_time = urldecode($response[$this->infoPrefix . 'ORDERTIME']);` on line 514.

proseLA:
my opinion would be to change the logic on the method to:

```
function datetime_to_sql_format(string $dateString): string
{
    $dateTime = DateTime::createFromFormat('H:i:s M d, Y e', $dateString);
    $dateTime->setTimezone((new DateTime)->getTimezone());
    return $dateTime->format('Y-m-d H:i:s');
}
```

with regards to the `convertToLocalTime`, i have no problem with that change.  perhaps add return signature as string.

drbyte:
Agreed on both counts!

drbyte:
Thanks!

PR #5935, opened 2023-08-15
Fix for missing buttons in empty product-type Extras sub-pages

### LINKED ISSUE #5923: Admin, Extras: all pages missing Insert button

simon1066:
Fixes #5923

Admin > Extras > Media Types was fixed in #5924, this PR fixes the remaining sub-menus

`media_manager.php` may need a closer check as it's possible that I may have lost my way, slightly.

### DIFF
--- a/admin/record_company.php
+++ b/admin/record_company.php
@@ -149,7 +149,7 @@ if (!empty($action)) {
                                              ORDER BY record_company_name";
                 $record_company_split = new splitPageResults($_GET['page'], MAX_DISPLAY_SEARCH_RESULTS, $record_company_query_raw, $record_company_query_numrows);
                 $record_companies = $db->Execute($record_company_query_raw);
-
+                $rCompany_parameter = '';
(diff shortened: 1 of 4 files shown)


admin/record_company.php
```diff
@@ -158,6 +158,7 @@
 
                     $aInfo_array = array_merge($record_company, $record_company_products->fields);
                     $aInfo = new objectInfo($aInfo_array);
+					$rCompany_parameter = '&mID=' . $aInfo->record_company_id;
```
drbyte, line 161: Indentation is strange here. Best to use spaces, not tabs.

simon1066:
Thank you. Annoyed that I missed the spaces-not-tabs as I was aware of that, I'll resubmit. Better next time.

drbyte:
Thanks!

PR #5941, opened 2023-08-18
refactor use of orders status array

proseLA:
no functional change...

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -344,9 +344,9 @@ function zen_get_system_information($privacy = false)
         }
     }
     if (!$exec_disabled) {
-        list($system, $host, $kernel) = array('', $_SERVER['SERVER_NAME'], php_uname());
+        [$system, $host, $kernel] = array('', $_SERVER['SERVER_NAME'], php_uname());
         @exec('uname -a 2>&1', $output, $errnum);
-        if ($errnum == 0 && count($output)) list($system, $host, $kernel) = preg_split('/[\s,]+/', $output[0], 5);
+        if ($errnum == 0 && count($output)) [$system, $host, $kernel] = preg_split('/[\s,]+/', $output[0], 5);
         $output = '';
         if (DISPLAY_SERVER_UPTIME == 'true') {
             @exec('uptime 2>&1', $output, $errnum);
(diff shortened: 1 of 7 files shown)


admin/includes/functions/general.php
```diff
@@ -738,23 +738,31 @@ function zen_set_ezpage_status(int $pages_id, int $status, string $status_field)
...
+            $orders_statuses[] = [
+                'id' => $orders_status_query->fields['orders_status_id'],
+                'text' => $orders_status_query->fields['orders_status_name'] . ' [' . $orders_status_query->fields['orders_status_id'] . ']',
+            ];
+            $orders_status_array[$orders_status_query->fields['orders_status_id']] = $orders_status_query->fields['orders_status_name'] . ' [' . $orders_status_query->fields['orders_status_id'] . ']';
+        } else {
+            $orders_statuses[$orders_status_query->fields['orders_status_id']] = $orders_status_query->fields['orders_status_name'];
+            $orders_status_array[$orders_status_query->fields['orders_status_id']] = $orders_status_query->fields['orders_status_name'];
+        }
+        $orders_status_query->MoveNext();
```
lat9, line 763: Use `foreach ($orders_status_query as $next_status) {` instead of the `while/MoveNext()` loop, replacing each instance of `$orders_status_query->fields` with `$next_status`.

lat9 (changes requested):
No sense in propagating that legacy database-query looping style.

admin/includes/functions/general.php
```diff
@@ -738,23 +738,30 @@ function zen_set_ezpage_status(int $pages_id, int $status, string $status_field)
...
-                                   ORDER BY orders_status_id");
-
-    while (!$orders_status->EOF) {
-        $orders_status_array[] = [
-            'id' => $orders_status->fields['orders_status_id'],
-            'text' => $orders_status->fields['orders_status_name']
-        ];
-        $orders_status->MoveNext();
+    $orders_status_query = $db->Execute('SELECT orders_status_id, orders_status_name FROM ' . TABLE_ORDERS_STATUS . '
+                                 WHERE language_id = "' . (int)$_SESSION['languages_id'] . '" ORDER BY sort_order, orders_status_id');
```
drbyte, line 751: Would it make sense to include the caching parameters here, as used in the dashboard widget?

`$db->Execute(SQLQUERY, false, true, 3600)` // for one hour

proseLA: i'm going to wait an hour before i answer this...

proseLA: i am really on the fence on this one.  i do not feel strongly 1 way or another.  i think maintaining order statuses is rarely done.  but i also do not think that caching this query saves much time in the grand scope of things.  i could see people questioning why their statuses are not showing up right after entering them and wanting to change an order status.

that said, i could see adding another optional input parm to effect caching on the method call from the widget.

but i am open.

drbyte: Understood. The only time the cache would need a reset would be directly after an edit.
And the reset has to be done to the call to the exact same query string
So yes passing a parameter, maybe `$resetCache=false`, would make sense ... and that would only ever be passed as true from a fresh call to this same function immediately after an insert/edit/delete and before the usual zen_redirect that follows such actions.

drbyte: I also agree that this is not an expensive query so the db's own internal caching is probably sufficient.

drbyte: Will leave it for now

PR #5942, opened 2023-08-18
shipping estimator module - formatting etc, no functional changes

torvista:
to squash, once reviewed.

### DIFF
--- a/includes/modules/shipping_estimator.php
+++ b/includes/modules/shipping_estimator.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Shipping Estimator module
  *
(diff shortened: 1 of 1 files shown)


includes/modules/shipping_estimator.php
```diff
@@ -209,7 +209,6 @@ function shipincart_submit() {
                         foreach ($selected_quote[0]['methods'] as $qkey => $qval) {
                             if ($qval['id'] == $method) {
                                 $selected_quote[0]['methods'] = [$qval];
-                                continue;
```
drbyte, line 211: Should that have been `continue 2;` instead? ie: abort the current foreach (because it's satisfied) and jump to the next iteration of the outer foreach. Perhaps the redundant continue statement was a bug that should have specified the '2' part?

drbyte: Other than that question, I'm satisfied with these formatting updates.

torvista: I've added the continue in back in for now so this PR does not touch functionality. I'll look at the continue shortly, separately.

PR #5948, opened 2023-08-21
early return if adminID not set

proseLA:
if we are now making this method available on the storefront, we should not assume that there will always be a `$_SESSION['admin_id']` set.

i have opted for an empty return but i'm not opposed to something else.

i think the check should definitely be done in this method as opposed to the call to the method.

### DIFF
--- a/includes/functions/functions_general_shared.php
+++ b/includes/functions/functions_general_shared.php
@@ -389,8 +389,11 @@ function request()
     return \Zencart\Request\Request::getInstance();
 }
 
-function zen_updated_by_admin($admin_id = null)
+function zen_updated_by_admin($admin_id = null): string
 {
+    if (empty($admin_id) && empty($_SESSION['admin_id'])) {
+        return '';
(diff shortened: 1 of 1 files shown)


drbyte (approved):

includes/functions/functions_general_shared.php
```diff
@@ -389,8 +389,11 @@ function request()
     return \Zencart\Request\Request::getInstance();
 }
 
-function zen_updated_by_admin($admin_id = null)
+function zen_updated_by_admin($admin_id = null): string
 {
+    if (empty($admin_id) && empty($_SESSION['admin_id'])) {
+        return '';
```
scottcwilson, line 395: Should this be `return null; `  as is done on line 416?

drbyte: No, because this function returns a string, and the `zen_update_orders_history()` function that calls it always expects a string.

PR #5951, opened 2023-08-22
shipping estimator template: non-functionality changes

torvista:
tr tags: IDE was unable to balance the tr tags as correct, this is easier to read.


torvista:
Do I need to do something with this, PR the tr change separately? 
As that's the only thing that is not easily checked by eye?

brittainmark (commented):
Looks good to me. I like the <tr> change looks logically correct and easier to understand.

torvista:
I'll squash when requested.

torvista:
**>Do I need to do something with this**...

PR #5953, opened 2023-08-23
Sort all auto-loaded "extra" folders by alphabetical

### LINKED ISSUE #5946: Change load-order of extra_configures etc folders?
@lat9 wrote:
> The problem with the files in the /extra_configures directory is that they're loaded in file system order ... not necessarily alphabetically.

With 2.0, is it time to add an alpha-sort to the code that auto-loads files in various filesystem directories?

drbyte:
Closes #5946

### DIFF
--- a/admin/define_pages_editor.php
+++ b/admin/define_pages_editor.php
@@ -7,27 +7,20 @@
  */
 require('includes/application_top.php');
 
-function zen_display_files() {
-  global $check_directory, $found, $configuration_key_lookup;
-    for ($i = 0, $n = sizeof($check_directory); $i < $n; $i++) {
+function zen_display_files(): array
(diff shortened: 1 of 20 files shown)


drbyte:
Notes:
- the main function handling this is `zen_get_files_in_directory()` in `functions_files.php`
- it uses `glob()` to get the directory listing, already sorted
- a handful of places in the early-bootstrapping process have to use `glob()` directly because `functions_files.php` isn't loaded yet; fortunately the code is also greatly simplified in those cases vs the old method
- leftover uses of `dir()` to iterate via a loop are cases where a specific regex was needed, or files were being deleted for cleanup and therefore a sort would be a waste of time

PR #5956, opened 2023-08-24
For ADMIN_CONFIGURATION_KEY_ON, Label module name as code instead of key

torvista:


lat9:
What's the problem you're solving with this change?

torvista:
The module code is misnamed as a key.

2023-08-24 21_03_05-Admin Modules — Mozilla Firefox


2023-08-24 21_03_39-Admin Modules — Mozilla Firefox

PR #5960, opened 2023-08-26
Auto-expire specials/featured/sales once per admin session/day

lat9:
Semi-replicating the once-per-day/session handing on the storefront for the specials/featured/salemaker auto-enable/expire checks.  Otherwise, the expirations (which normally happen only once/day) are checked on *every* admin page-load.

Discovered this issue on a zc158 site whose admin was taking 1-2 seconds to load each-and-every admin page.  The site had minimal (i.e. < 20) of these automatic updates to perform, but adding this change, the other pages in the admin averaged less than 250ms to load.

### DIFF
--- a/admin/includes/init_includes/init_special_funcs.php
+++ b/admin/includes/init_includes/init_special_funcs.php
@@ -12,24 +12,21 @@ if (!defined('IS_ADMIN_FLAG')) {
 // set a default time limit
 zen_set_time_limit(GLOBAL_SET_TIME_LIMIT);
 
-// auto activate and expire banners
-require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'banner.php';
-zen_activate_banners();
-zen_expire_banners();
-
 // -----
-// Functions are still needed for specials/featured/salemaker tools'
-// processing!
+// Load required "special" function-files, associated with features that
+// auto-expire and auto-activate.
 //
+require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'banner.php';
 require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'specials.php';
 require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'featured.php';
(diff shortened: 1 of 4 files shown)


scottcwilson:
This is a really interesting find.  Do we know the conditions under which the slowdown occurs? (i.e. PHP 8+ only or ZC 158 only?). Is it related to the issue Melanie reported here: https://www.zen-cart.com/showthread.php?229632-Stumped/ ?

scottcwilson:
Once we have more information, this might be a good candidate for adding to https://docs.zen-cart.com/user/speed/admin_slow/

admin/includes/init_includes/init_special_funcs.php
```diff
@@ -6,29 +6,53 @@
...
+zen_activate_banners();
+zen_expire_banners();
+
+// -----
+// Functions are still needed for specials/featured/salemaker tools'
+// processing!
+//
+require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'specials.php';
+require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'featured.php';
+require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'salemaker.php';
```
drbyte, line 26: I'm inclined to relocate these up above the expiring of banners: ie: include/require all the files ... cuz this init file is tasked with loading those functions files.
And then do all the activating/expiring/etc of banners and specials/etc in the last half of the file.

I'm not "bent" on this suggestion. It was your "are still needed" comment that led me to consider "what do you mean by 'still needed'?" ... cuz those files must always get loaded for core code to work too, not just for expiring/etc.

lat9: My "still needed" comment was after-the-fact (on the client's update) realizing that the `require` of the functions needed to be outside of the does-this-need-to-be-run section, since the Specials/SaleMaker tools use the functions therein to 'toggle' an elements status.

PR #5966, opened 2023-08-27
Extract error-handling code into core functions file

### LINKED ISSUE #5944: load problem with error logging....
i like to look at logs.  it helps solve problems.

i also like to look at logs correctly time stamped.  unfortunately the following code:

https://github.com/zencart/zencart/blob/bd417ab84933856dcea06cd412373af821cb91bd/includes/extra_configures/enable_error_logging.php#L146-L149

gets loaded before this code:

https://github.com/zencart/zencart/blob/bd417ab84933856dcea06cd412373af821cb91bd/includes/extra_configures/set_time_zone.php#L22-L28

which means i end up looking at logs that look like this:

```
237 Aug 18 16:46 myDEBUG-20230818-234640-944827.log
```
now i am running a version of alma linux that has the timezone set to PDT:

```
[pRose@alma logs]$ date
Fri Aug 18 05:04:26 PM PDT 2023
```

and yet the timestamp on the error log is wrong.  so perhaps i have set the timezone incorrectly on my alma linux box; or something with something else...

but that said, would it not make more sense to set the timezone PRIOR to enabling the error logs which make use of a date time stamp?

drbyte:
This leaves the extra_configures/enable_error_logging.php file as merely a simple configuration file.

Also fixes #5944

### DIFF
--- a/includes/functions/functions_error_handling.php
+++ b/includes/functions/functions_error_handling.php
@@ -1,0 +1,160 @@
+<?php
+/**
+ * @copyright Copyright 2003-2023 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id:  $
+ */
+if (!defined('IS_ADMIN_FLAG')) {
+    exit('Invalid Access');
+}
+
+function zen_debug_error_handler($errno, $errstr, $errfile, $errline)
+{
+    if (!(error_reporting() & $errno)) {
+        return;
+    }
+
+    static $last_log_suffix;
+    if (!isset($last_log_suffix)) {
+        $last_log_suffix = '.log';
+    }
+    $ignore_dups = false;
+    if (IS_ADMIN_FLAG === true) {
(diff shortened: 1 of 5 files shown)


includes/functions/functions_error_handling.php
```diff
@@ -0,0 +1,160 @@
...
+
+function zen_enable_error_logging(array $pages_to_debug = ['*'], $logging_level = E_ALL & ~E_NOTICE): void
+{
+    global $current_page_base, $debug_logfile_path;
+
+    if (empty($debug_logfile_path)) {
+        $debug_logfile_path = zen_set_error_logging_filename();
+    }
+
+    if (in_array('*', $pages_to_debug, true) || in_array($current_page_base, $pages_to_debug, true)) {
```
lat9, line 146: `$pages_to_debug` should be set in global scope.

proseLA: why does it need to be in the global scope?  the var is passed as an input parm to the method.

what am i missing?

proseLA: as a side note, via the terminal, i was able to download this pr (and test) using the following command:

```
git fetch upstream pull/5966/head:error-handling
```
i am sure there is something similar in the git client that you are using.

PR #5968, opened 2023-08-28
paypaywpp: initial IDE inspections/PSR-12

torvista:
replace isset with ??
short array syntax
use short form of type keywords
add parameter list (empty) to new classes
return types
some else if to elseif

**no logic changes.**

(Adding public visibility to functions changes whole file layout, so better as a separate PR).


brittainmark (commented):
There may be other places that you could change and if (!isset($var)) {
$var=value; }

This is just an observation up to you if you want to make the changes.

torvista:
On the first pass I try to offer only as much stuff as possibly that can be easily visually checked/non-contentious to get it out of the way quickly before adding stuff that requires some attention/more time.
This is a big file with multiple things to fettle with...

brittainmark:
Fine with me.

torvista:
I've been using it for some time.

torvista:
Is this waiting for something?

PR #5982, opened 2023-09-03
Rename $customer var to avoid conflicts and ambiguity

### LINKED ISSUE #5981: Fatal error:  Call to undefined method queryFactoryResult::getData() in functions_gvcoupons.php:61
Strict_reporting, php8.3.0RC1
To make it easier, use Classic template, but it occurs in all.
Enable the shopping cart sidebox, (on the left/so it is displayed).
Delete anything in the cart and Login, to satisfy this and call function:
https://github.com/zencart/zencart/blob/b3cd1f31aa5578acc51c3c7c5f29c6c338e311d5/includes/modules/sideboxes/shopping_cart.php#L16

Go to any product and Write Review: debug and partial whitescreen.

Multiple similar errors:

> [03-Sep-2023 15:18:29 UTC] PHP Fatal error:  Uncaught Error: Call to undefined method queryFactoryResult::getData() in \zencart\includes\functions\functions_gvcoupons.php:61
> Stack trace:
> #0 zencart\includes\modules\sideboxes\shopping_cart.php(16): zen_user_has_gv_account(1)
> #1 zencart\includes\modules\column_left.php(30): include('D:\\GitHub\\zenca...')
> #2 zencart\includes\templates\template_default\common\tpl_main_page.php(109): require('D:\\GitHub\\zenca...')
> #3 zencart\index.php(94): require('D:\\GitHub\\zenca...')
> #4 {main}
>   thrown in includes\functions\functions_gvcoupons.php on line 61

At this point (functions_gvcoupons.php on line 61), $customer is indeed a queryFactoryResult and not a Customer object.

drbyte:
The typechecking added here to the `Customers` class is probably sufficient to resolve those errors.

However, I've also included additional fixes to reduce ambiguity.

Fixes #5981

### DIFF
--- a/includes/modules/pages/checkout_confirmation/header_php.php
+++ b/includes/modules/pages/checkout_confirmation/header_php.php
@@ -20,13 +20,14 @@ if ($_SESSION['cart']->count_contents() <= 0) {
   if (!zen_is_logged_in()) {
     $_SESSION['navigation']->set_snapshot(array('mode' => 'SSL', 'page' => FILENAME_CHECKOUT_PAYMENT));
     zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
-  } else {
-    // validate customer
-    if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
-      $_SESSION['navigation']->set_snapshot();
-      zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
-    }
   }
+
+$customer = new Customer($_SESSION['customer_id']);
+// validate customer
+if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
+    $_SESSION['navigation']->set_snapshot();
+    zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
+}
 
 // Ensure no cart content changes during the checkout procedure by checking the internal cartID
 if (isset($_SESSION['cart']->cartID) && isset($_SESSION['cartID'])) {
--- a/includes/modules/pages/checkout_payment/header_php.php
+++ b/includes/modules/pages/checkout_payment/header_php.php
@@ -24,13 +24,14 @@ if ($_SESSION['cart']->count_contents() <= 0) {
   if (!zen_is_logged_in()) {
     $_SESSION['navigation']->set_snapshot();
     zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
-  } else {
-    // validate customer
-    if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
-      $_SESSION['navigation']->set_snapshot();
-      zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
-    }
   }
+
+$customer = new Customer($_SESSION['customer_id']);
+// validate customer
+if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
+    $_SESSION['navigation']->set_snapshot();
+    zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
+}
 
 // if no shipping method has been selected, redirect the customer to the shipping method selection page
 if (!isset($_SESSION['shipping']) || !$_SESSION['shipping']) {
@@ -119,6 +120,7 @@ if (isset($_GET['payment_error']) && is_object(${$_GET['payment_error']}) && ($e
 $breadcrumb->add(NAVBAR_TITLE_1, zen_href_link(FILENAME_CHECKOUT_SHIPPING, '', 'SSL'));
 $breadcrumb->add(NAVBAR_TITLE_2);
 
+$gv_balance = $customer->getData('gv_balance');
+
 // This should be last line of the script:
 $zco_notifier->notify('NOTIFY_HEADER_END_CHECKOUT_PAYMENT');
-?>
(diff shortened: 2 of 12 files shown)


torvista:
Works for me.

includes/modules/sideboxes/shopping_cart.php
```diff
@@ -8,19 +8,25 @@
  * @version $Id: DrByte 2020 Jul 10 Modified in v1.5.8-alpha $
  */
 
+  $gv_balance = 0.00;
+  if (zen_is_logged_in() && !zen_in_guest_checkout()) {
+    $customer = new Customer;
```
proseLA, line 13: i am a fan of line 13, as well as now removing the call to the function `zen_user_has_gv_account` in this script.

if merged, said function will now only have 1 call in the base code, which is in the script `tpl_checkout_payment_default.php`.

given that we are now moving towards a more class based ideology (at least from what i am seeing), i do not understand how this `$customer` object is not available during the checkout payment process.  i would think that if a customer is logged in, we should have a `$customer` object which is an instance of the `Customer` class.

i think eliminating calls to these `zen_` functions would make zc much more flexible; and needless to say i am a fan.

i suppose what i am getting at, is that the conditional for line 12 should be done much sooner during the load process.  if the customer is logged in, why do we not create the Customer object at an earlier time and make it much more available?

drbyte: Done

includes/modules/pages/checkout_confirmation/header_php.php
```diff
@@ -20,14 +20,15 @@
...
-    // validate customer
-    if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
-      $_SESSION['navigation']->set_snapshot();
-      zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
-    }
   }
 
+$customer = new Customer($_SESSION['customer_id']);
+// validate customer
+if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
```
lat9, line 27: Use `===`

includes/modules/pages/checkout_payment/header_php.php
```diff
@@ -24,14 +24,15 @@
...
-    // validate customer
-    if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
-      $_SESSION['navigation']->set_snapshot();
-      zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
-    }
   }
 
+$customer = new Customer($_SESSION['customer_id']);
+// validate customer
+if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
```
lat9, line 31: Use `===`

includes/modules/pages/checkout_payment_address/header_php.php
```diff
@@ -17,16 +17,17 @@
...
-    if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
-      $_SESSION['navigation']->set_snapshot(array('mode' => 'SSL', 'page' => FILENAME_CHECKOUT_SHIPPING));
-      zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
-    }
-  }
+}
+
+$customer = new Customer($_SESSION['customer_id']);
+// validate customer
+if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
```
lat9, line 27: Use `===`

includes/modules/pages/checkout_shipping/header_php.php
```diff
@@ -20,14 +20,15 @@
...
-    // validate customer
-    if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
-      $_SESSION['navigation']->set_snapshot(array('mode' => 'SSL', 'page' => FILENAME_CHECKOUT_SHIPPING));
-      zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
-    }
   }
 
+$customer = new Customer($_SESSION['customer_id']);
+// validate customer
+if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
```
lat9, line 27: Use `===`

includes/modules/pages/checkout_shipping_address/header_php.php
```diff
@@ -17,16 +17,17 @@
...
-    if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
-      $_SESSION['navigation']->set_snapshot(array('mode' => 'SSL', 'page' => FILENAME_CHECKOUT_SHIPPING));
-      zen_redirect(zen_href_link(FILENAME_LOGIN, '', 'SSL'));
-    }
-  }
+}
+
+$customer = new Customer($_SESSION['customer_id']);
+// validate customer
+if (zen_get_customer_validate_session($_SESSION['customer_id']) == false) {
```
lat9, line 27: Use `===`

lat9 (changes requested):
Use strict comparison when checking customer validation.

drbyte:
Will do `===` separately, because those lines are not canonically part of this PR, save for indentation.

includes/modules/pages/checkout_confirmation/header_php.php
drbyte: Done in https://github.com/zencart/zencart/commit/464fb1a4fe00e55abc77c084b2b9c28076aec985

PR #5984, opened 2023-09-04
add test and fix vulnerablitlity

zcwilt:


lat9:
Erm, I just applied this to my zc200 test site and none of the zc_plugin tools are available in the admin.  Clicking on the "Tools :: Display Logs", for instance, just redirects to the admin home page ... like it's not found.

zcwilt:
The problem turns out to be a windows issue with realpath and the directory path.

I've reverted the original PR and will look to fix again.

PR #5987, opened 2023-09-07
Admin/orders: Limit when single order, correct layout

lat9:
On a site that has 120K+ orders, adding the 'limit 1' when a specific order is being viewed is a significant performance improvement.

For the orders' button row, change the middle section to have a large-viewport column-width of 4 (so the column-counts are 4/4/4).  Otherwise, site-specific button additions will overflow.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -59,7 +59,7 @@ if (isset($_POST['oID'])) {
 if ($oID) {
   $orders = $db->Execute("SELECT orders_id
                           FROM " . TABLE_ORDERS . "
-                          WHERE orders_id = " . $oID);
+                          WHERE orders_id = " . $oID, 1);
   $order_exists = true;
   if ($orders->RecordCount() <= 0) {
     $order_exists = false;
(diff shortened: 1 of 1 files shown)


scottcwilson:
> On a site that has 120K+ orders, adding the 'limit 1' when a specific order is being viewed is a significant performance improvement.

I wouldn't have expected this.  Interesting finding! Glad you keep finding ways to make big stores run better.

scottcwilson:
I don't notice a big difference changing includes/classes/order.php `query()` function to limit 1 on my client with 450K orders - can you try this on yours?

lat9:
Try it when the orders' listing is based on an email address search.

scottcwilson:
I was talking about the storefront account_history_info page, which I was hoping would benefit the same way from your change (in `includes/classes/order.php`).  To me, it doesn't look any faster and I just wanted to see if you saw any improvement.

lat9:
Again, the additional factor was the presence of a keyword search on the admin side.

PR #5989, opened 2023-09-10
Admin overrides, add $show_registration_ip_in_listing

torvista:


admin/includes/extra_datafiles/dist-site_specific_admin_overrides.php
```diff
@@ -36,45 +36,45 @@
 // 
 // $show_product_images  $show_product_images_pack
 // true or unset         true or unset            ...... product images are shown on invoice and packingslip (the default).
-// false                 false or unset           ...... product images are NOT shown on invoice or packingslip.
-// false                 true                     ...... product images are NOT shown on invoice  but are shown packingslip.
-// true or unset         false                    ...... product images are shown on invoice and but are NOT shown packingslip.
+// false                 false or unset           ...... product images are NOT shown on invoice nor packingslip.
```
lat9, line 39: Actually, 'or' is correct in this case.  To use 'nor', the phrase would be 'product images are shown on neither invoice nor packingslip'.

PR #5994, opened 2023-09-14
show/hide images on the admin catalog/product page

### LINKED ISSUE #5993: admin/category_products_listing: Shouldn't the image-column be optional?
As indicated by a couple of up-votes in [this](https://github.com/zencart/zencart/pull/4222) PR that added the category/product images to the admin's Categories/Products listing, shouldn't the image column have a soft-setting to allow a quick way to disable that information from the view?

Not all sites want that additional 'clutter'.

proseLA:
fixes #5993

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -18,6 +18,12 @@ if (isset($_POST['products_id'])) {
     $product_type = 1;
 }
 
+if (!isset($_SESSION['imageView'])) {
+    $_SESSION['imageView'] = true;
+}
+$buttonText = $_SESSION['imageView'] ?  TEXT_HIDE_IMAGES : TEXT_SHOW_IMAGES ;
+$additionalClass = $_SESSION['imageView'] ? '' : ' hidden ';
+
 $action = $_GET['action'] ?? '';
 $search_result = isset($_GET['search']) && zen_not_null($_GET['search']);
 
@@ -341,6 +347,12 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                     </div>
                   </td>
                 </tr>
+                <tr>
+                    <td></td>
+                    <td><input type="button" id="imageView" class="btn btn-info btn-xs" value="<?= $buttonText; ?>">
+                    </td>
+                    <td colspan="3"></td>
+                </tr>
               </tbody>
             </table>
           </div>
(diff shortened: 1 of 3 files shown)


scottcwilson:
Not sure a toggle is really needed; I thought this was just going to be another flag.

lat9:
> Not sure a toggle is really needed; I thought this was just going to be another flag.

Per the @torvista comment in the associated issue, I believe that having a toggle (so that an admin has control of their environment) is a good thing.

lat9:
@proseLA, what do you think the effort would be to make this a session-based setting for the admin.  That way, once they've chosen to see/not see the images the selection would be remembered for their current login.

torvista:
This modification still downloads the images and but then offers the option to hide them.
So it saves screen space, offers a more compact listing, but does not have a speed advantage.

I think the most complete implementation of this would be to prevent images being downloaded and be session-maintained.
That's what I did.

scottcwilson:
> > Not sure a toggle is really needed; I thought this was just going to be another flag.
> 
> Per the @torvista comment in the associated issue, I believe that having a toggle (so that an admin has control of their environment) is a good thing.

Literally nothing else in the project works this way, but Steve has yet another personal preference, and he has unlimited time and energy for complaining, so we'll do it his way?  

This doesn't seem like a great process.  But it's not important enough to me to fight over, so do what you think is best.

torvista:
>he has unlimited time and energy for complaining
I'll bypass the fundamental rule of not replying when angry.

Scott, you have an unlimited ability to repeatedly insult people. You just don't learn.
Where is my complaint in this thread?
I don't complain, I contribute. Or rather I did.

proseLA:
@torvista i am no longer a fan of page reloads.  i think modern website design is more about sending all of the data down to the browser, and then using css and javascript/jquery to style the data as needed for the page.  i think all of these image files should be reasonably small, and in addition, browser caching of image files means you are not continuously reloading these image files.  again, my opinion is that the speed savings by not downloading the image files are minimal (but i have not seen your image files), but feel free to prove me wrong.  i know in my experience, other shopping cart software's admin experience is far slower than zc.  

@scottcwilson i did not code this to steve's preference.  i coded it to what made sense to me.  in addition, i think the rationale that nothing else is done this way is not a good rationale.  either we want zc to move forward with new ideas, incorporating more modern ideas of web design, or we do not.  personally i am more a fan of the former.

admin/category_product_listing.php
```diff
@@ -341,6 +347,12 @@
                     </div>
                   </td>
                 </tr>
+                <tr>
+                    <td></td>
+                    <td><input type="button" id="imageView" class="btn btn-info btn-xs" value="<?= $buttonText; ?>"/>
```
lat9, line 352: For an HTML5 input, there's no `/` at the end-of-tag.

admin/category_product_listing.php
```diff
@@ -1293,7 +1305,24 @@
             jQuery("tr.product-listing-row td").not('.dataTableButtonCell').on('click', (function() {
                 window.location.href = productEditLink.replace('[*]', jQuery(this).parent().attr('data-pid'));
             })).css('cursor', 'pointer');
-        })
+        });
+        $(document).ready(function () {
+            $('#imageView').click(function () {
```
lat9, line 1310: Use `$('#imageView').on('click', function() {`

includes/classes/ajax/zcAjaxAdminSessionChange.php
```diff
@@ -0,0 +1,23 @@
...
+
+        } elseif ($_SESSION[$_POST['name']] === true) {
+            $_SESSION[$_POST['name']] = false;
+            return $_POST['name'] . ' set to false!';
+        }
+
+        $_SESSION[$_POST['name']] = true;
+        return $_POST['name'] . ' set to true!';
+
+
```
lat9, line 20: Remove unneeded empty lines, please.

lat9 (changes requested):
A couple of nit-picks, as usual.

lat9 (approved):

PR #5997, opened 2023-09-24
column_right.php seems to reference the LEFT

highburyeye:
column right should use RIGHT
not LEFT

### DIFF
--- a/includes/modules/column_right.php
+++ b/includes/modules/column_right.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * column_right module
+ * column_right module 
  *
  * @copyright Copyright 2003-2022 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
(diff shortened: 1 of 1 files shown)


drbyte:
Closing since it's for v158 branch, and have merged #5998 into v200 branch.

lat9:
The _real_ question, in my mind, is why this change doesn't seem to matter ... at least for the Bootstrap template.

drbyte:
> The _real_ question, in my mind, is why this change doesn't seem to matter ... at least for the Bootstrap template.

Probably in large part because the Bootstrap template does its own thing instead of $column_width_default.

https://github.com/zencart/zencart/blob/45b05d292319dffbd8f7f2ad2c8fd1d33c3f85d9/includes/templates/template_default/common/tpl_box_default_left.php#L18

https://github.com/lat9/ZCA-Bootstrap-Template/blob/89eb1f9d972001f20e041bd9b5b76cfe4abcec4c/includes/templates/bootstrap/common/tpl_box_default_left.php#L18-L27

PR #6000, opened 2023-09-25
Capture payment module final status in order

lat9:
If a payment module, e.g. `authorizenet_aim`, sets a different status during its `before_process` method, that status is not initially recorded into the order &mdash; which can cause customer confusion.

For instance, if an a.net/aim transaction is being held for review, a specific a.net/aim order-status can be set.  Currently, the order's initial status (viewable by the customer) indicates that the order is awaiting processing, but a subsequent (hidden) status-update by the a.net/aim payment module sets the order's status to its "Under Review" setting.

This update corrects that issue.

### DIFF
--- a/includes/modules/checkout_process.php
+++ b/includes/modules/checkout_process.php
@@ -96,7 +96,15 @@ if (!isset($_SESSION['payment']) && $credit_covers === FALSE) {
 
 // load the before_process function from the payment modules
 $payment_modules->before_process();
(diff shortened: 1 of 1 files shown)


drbyte (approved):
I'm "pretty sure" this is okay.

drbyte:
Must also remember that not all status-updates need to be visible to the customer.

PR #6003, opened 2023-10-06
Add mysql version check

### LINKED ISSUE #6002: zc_install: Need to check for minimum MySQL version prior to continuing
As described in [this](https://www.zen-cart.com/showthread.php?229421-1-5-8a-Missing-customer_groups-field-in-DB) Zen Cart bug-report, there have been a couple of sites that have undergone a _successful_ install/upgrade only to find that the `customer_groups` and `customers_to_groups` tables are missing:
```
PHP Fatal error: 1146:Table 'xxx.customer_groups' doesn't exist :: SELECT count(*) as total FROM (SELECT cg.*, count(ctg.customer_id) as customer_count FROM customer_groups cg LEFT JOIN customers_to_groups ctg USING (group_id) GROUP BY cg.group_id ORDER BY group_name, group_id) countresults ==> (as called by) /xxxxxxxxxxxxx/includes/classes/split_page_results.php on line 220 <== in /includes/classes/db/mysql/query_factory.php on line 667.
```

Shouldn't the zc_install process be checking for a minimum MySQL/MariaDB version before proceeding with the install/upgrade to prevent these issues?  This issue, for instance, was reported for a site running MySQL 5.5.68-MariaDB .

brittainmark:
Add a check into the install to ensure that the database is at least the minimum version of Mysql of MariaDB. The check only fails if it is able to perform a version check and the version is less than that specified in systemChecks.yml.

fixes  #6002

### DIFF
--- a/zc_install/includes/classes/class.systemChecker.php
+++ b/zc_install/includes/classes/class.systemChecker.php
@@ -730,4 +730,41 @@ class systemChecker
             }
         }
     }
+    /**
+     * Check installed Database version
+     * The check is only validated if the information is available.
+     * There are other checks that will fail so don't issue spurious error message
+     */
+    public function checkMysqlVersion($parameters) {
+        if (function_exists('mysqli_connect')) {
+            $dbServerVal = $this->getServerConfig()->getDefine('DB_SERVER');
+            $dbNameVal = $this->getServerConfig()->getDefine('DB_DATABASE');
+            $dbPasswordVal = $this->getServerConfig()->getDefine('DB_SERVER_PASSWORD');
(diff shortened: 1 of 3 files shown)


drbyte:
Looks fine at first glance, but I've not tested it with anything yet.

I'd prefer that the mention of "Mysql" be "MySQL" in the error messages, but that's merely a minor cosmetic tweak.

brittainmark:
Changed as requested

drbyte:
Thanks.

drbyte:
Thanks for taking care of this!

PR #6008, opened 2023-10-09
This would have the JS Popup Window replaced with a Modal DIV element.

MarkoKrsic:
Replaces JS popup window with JS Modal DIV, browser stays on product screen/page.
This would replace `additional_images.php` in `includes/modules/` and `tpl_modules_main_product_image.php` in the  `includes/templates/template_default/templates/`

### DIFF
--- a/includes/modules/payment/paypal/paypal_functions.php
+++ b/includes/modules/payment/paypal/paypal_functions.php
@@ -53,7 +53,7 @@ if (!function_exists('convertToLocalTimeZone')) {
 
   function ipn_get_stored_session($session_stuff) {
     global $db;
-    if (!is_array($session_stuff)) {
+    if (!is_array($session_stuff) || !isset($session_stuff[1])) {
       ipn_debug_email('IPN FATAL ERROR :: Could not find Zen Cart custom variable in POST, cannot validate or re-create session as a transaction initiated from this store. Might be from another source such as eBay or another PayPal store using this PayPal account.');
       return false;
     }


Zen4All:
IMHO the styles and javascripts should go in their seperate files, instead of including them in the php file.

MarkoKrsic:
> IMHO the styles and javascripts should go in their seperate files, instead of including them in the php file.

I agree and knew I forgot something right after I submitted just didn't have time yet to circle back to it.

 I am also not sure where to place them for best conformity to existing project standards?  Should I put the CSS in the base `includes/templates/template_default/css/stylesheet.css`? If not where do I declare and include it?

JS in jscript folder?

dbltoe:
Since it is going to be part of the core, I would use the existing stylesheets.
For add-ons and customizations, we add stylesheet_zcustom.css to most of our projects so we can find the code when upgrade time comes.

lat9:
I take my thumbs-up back.  Since that CSS requires 'programmatic' inserts, I'll suggest creating a `.php` file in the `/templates/template_default/css` directory that does the insertions and creating a like-named file in the `/templates/responsive_classic/css` directory that requires the template_default version (so there's really only one file to be maintained).

Each template's `/common/html_header.php` would specifically load the file in its `/css` subdirectory.

drbyte:
Oops. I was trying to update the branch, and github closed it. 

I'll push it as a separate PR, giving @MarkoKrsic commit credit.

PR #6011, opened 2023-10-12
Shopping cart performance enhancements

lat9:
Reformatting
- Use null coalescing (??) operator where indicated
- Use short array syntax
- No one-line conditionals
- No one-line multiple assignments

Refactoring
- Since the `contents` property is keyed on a 'uprid', change the `$products_id` key to be `$uprid` on foreach traversals of that array.  That makes it more obvious when viewing the code as to what the `contents` array is keyed on.
- Limit MySQL updates to a single row for unique operations.
- Use `===` where warranted.
- Use `zen_define_default` where warranted.
- Use `strpos` instead of `preg_match` when checking for gift-certificates in the cart.
- Remove redundant code.
- Use `zen_get_product_details` where warranted to get maximum cached queries.  Results in a roughly 50% reduction in time taken to add/remove/update a product as well as the cart's overall calculations!

Corrections
- `calculate` didn't process a 'base' product pricing if a product wasn't found, but *did* process the attributes!  Should the product be removed from the cart if this condition is found?

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -107,51 +107,50 @@ class shoppingCart extends base
         $this->notify('NOTIFIER_CART_RESTORE_CONTENTS_START');
         // insert current cart contents in database
         if (is_array($this->contents)) {
-            foreach ($this->contents as $products_id => $data) {
-                // $products_id = urldecode($products_id);
-                $qty = $this->contents[$products_id]['qty'];
+            foreach ($this->contents as $uprid => $data) {
+                $uprid_db = zen_db_input($uprid);
+                $qty = $data['qty'];
                 $sql = "SELECT products_id
                         FROM " . TABLE_CUSTOMERS_BASKET . "
                         WHERE customers_id = " . (int)$_SESSION['customer_id'] . "
-                        AND products_id = '" . zen_db_input($products_id) . "'";
+                        AND products_id = '" . $uprid_db . "'";
 
                 $product = $db->Execute($sql);
 
-                if ($product->RecordCount() <= 0) {
-                    $sql = "INSERT INTO " . TABLE_CUSTOMERS_BASKET . "
-                            (customers_id, products_id, customers_basket_quantity,
-                             customers_basket_date_added)
-                             VALUES (" . (int)$_SESSION['customer_id'] . ", '" . zen_db_input($products_id) . "', '" .
-                        $qty . "', '" . date('Ymd') . "')";
+                if ($product->EOF) {
+                    $sql =
+                        "INSERT INTO " . TABLE_CUSTOMERS_BASKET . "
+                            (customers_id, products_id, customers_basket_quantity, customers_basket_date_added)
+                         VALUES
+                            (" . (int)$_SESSION['customer_id'] . ", '$uprid_db', $qty, '" . date('Ymd') . "')";
 
                     $db->Execute($sql);
 
-                    if (isset($this->contents[$products_id]['attributes'])) {
-
-                        foreach ($this->contents[$products_id]['attributes'] as $option => $value) {
+                    if (isset($data['attributes'])) {
+                        foreach ($data['attributes'] as $option => $value) {
 
                             // include attribute value: needed for text attributes
-                            $attr_value = isset($this->contents[$products_id]['attributes_values'][$option]) ? $this->contents[$products_id]['attributes_values'][$option] : '';
+                            $attr_value = $data['attributes_values'][$option] ?? '';
 
-                            $products_options_sort_order = zen_get_attributes_options_sort_order(zen_get_prid($products_id), $option, $value);
+                            $products_options_sort_order = zen_get_attributes_options_sort_order(zen_get_prid($uprid), $option, $value);
                             if ($attr_value) {
                                 $attr_value = zen_db_input($attr_value);
                             }
-                            $sql = "INSERT INTO " . TABLE_CUSTOMERS_BASKET_ATTRIBUTES . "
-                                    (customers_id, products_id, products_options_id,
-                                     products_options_value_id, products_options_value_text, products_options_sort_order)
-                                     VALUES (" . (int)$_SESSION['customer_id'] . ", '" . zen_db_input($products_id) . "', '" .
-                                    $option . "', '" . $value . "', '" . $attr_value . "', '" . $products_options_sort_order . "')";
+                            $sql =
+                                "INSERT INTO " . TABLE_CUSTOMERS_BASKET_ATTRIBUTES . "
+                                    (customers_id, products_id, products_options_id, products_options_value_id, products_options_value_text, products_options_sort_order)
+                                 VALUES
+                                    (" . (int)$_SESSION['customer_id'] . ", '$uprid_db', '$option', '$value', '$attr_value', '$products_options_sort_order')";
 
                             $db->Execute($sql);
                         }
                     }
                 } else {
                     $sql = "UPDATE " . TABLE_CUSTOMERS_BASKET . "
-                            SET customers_basket_quantity = '" . $qty . "'
+                            SET customers_basket_quantity = $qty
                             WHERE customers_id = " . (int)$_SESSION['customer_id'] . "
-                            AND products_id = '" . zen_db_input($products_id) . "'";
-                    $db->Execute($sql);
+                            AND products_id = '$uprid_db'";
+                    $db->Execute($sql, 1);
 
                 }
             }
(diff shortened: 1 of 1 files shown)


lat9:
FWIW, I'm also considering removing the use of `$display_debug_messages`; is anyone actually using this?  There are many better ways to debug the shopping-cart actions.

drbyte:
> FWIW, I'm also considering removing the use of `$display_debug_messages`; is anyone actually using this? There are many better ways to debug the shopping-cart actions.

I've only used it a few times, but it was certainly helpful at the time.

If the class had more smaller methods in it it would be even easier to debug when a problem occurred, instead of having to trace through to figure out which block is doing what ... which that $display_debug_messages var was trying to help with, IIRC.

lat9:
Agreed on the smaller methods.  If the feature has utility, it stays!

drbyte:
Oy, this is a massive bunch of changes in one commit. Super complex to review.  😨

includes/classes/shopping_cart.php
```diff
@@ -610,11 +640,7 @@ public function get_product_id_list()
         if (!is_array($this->contents)) {
             return '';
         }
-        $product_id_list = [];
-        foreach ($this->contents as $products_id => $data) {
-            $product_id_list[] = $products_id;
-        }
-        return implode(',', $product_id_list);
+        return implode(',', array_keys($this->contents));
```
drbyte, line 643: Nice. 👍

includes/classes/shopping_cart.php
```diff
@@ -527,62 +570,49 @@ public function count_contents()
...
-                    AND products_id = '" . zen_db_input($product_id) . "'";
-            $db->Execute($sql);
-
-            $sql = "DELETE FROM " . TABLE_CUSTOMERS_BASKET_ATTRIBUTES . "
-                    WHERE customers_id = " . (int)$_SESSION['customer_id'] . "
-                    AND products_id = '" . zen_db_input($product_id) . "'";
-            $db->Execute($sql);
-        }
+        $this->notify('NOTIFIER_CART_REMOVE_START', null, $uprid);
+        $this->removeUprid(zen_db_input($uprid));
```
drbyte, line 615: Nice de-duplication. 👍

includes/classes/shopping_cart.php
```diff
@@ -634,107 +660,104 @@ public function calculate()
...
 
-            // products price
-            $sql = "SELECT products_id, products_price, products_tax_class_id, products_weight,
-                    products_priced_by_attribute, product_is_always_free_shipping, products_discount_type, products_discount_type_from,
-                    products_virtual, products_model
-                    FROM " . TABLE_PRODUCTS . "
-                    WHERE products_id = " . (int)$products_id;
+            $product = zen_get_product_details($prid);
+            if ($product->EOF) {
+                continue;   //-TODO: Should the product be removed from the cart if this is detected?
```
drbyte, line 682: Hmmm.. good question...

includes/classes/shopping_cart.php
```diff
@@ -634,107 +660,104 @@ public function calculate()
...
+            }
 
-            if ($product = $db->Execute($sql)) {
-                $prid = $product->fields['products_id'];
+            $product = $product->fields;
+            $this->notify('NOTIFY_CART_CALCULATE_PRODUCT_PRICE', $uprid, $product);
+            $prid = (int)$product['products_id'];   //-TODO: Needed?  Just in case the id was changed by the observer?
```
drbyte, line 687: Probably good, although it's more likely that other fields would be changed besides the primary key.

Q: While `(int)` is functionally the same as calling `zen_get_prid()`, why not use `zen_get_prid()` for consistency with line 678 above?

includes/classes/shopping_cart.php
```diff
@@ -744,173 +767,200 @@ public function calculate()
                     $sale_maker_discount = '';
 
                     // bottom total
-                    if ($attribute_price->fields['product_attribute_is_free'] == '1' && zen_get_products_price_is_free((int)$prid)) {
+                    if ($attribute_price->fields['product_attribute_is_free'] === '1' && zen_get_products_price_is_free($prid)) {
                         // no charge for attribute
+                        continue;
```
drbyte, line 772: Brilliant, assuming the bracket-matching indicates this. (This is more a note to myself to trace that back.)

includes/classes/shopping_cart.php
```diff
@@ -744,173 +767,200 @@ public function calculate()
...
-                        $added_charge = 0;
-                        if ($attribute_price->fields['attributes_price_factor'] > 0) {
-                            //echo 'products_id: ' . $product->fields['products_id'] . ' Prices ' . '$chk_price: ' . $chk_price . ' $chk_special: ' . $chk_special . ' attributes_price_factor:' . $attribute_price->fields['attributes_price_factor'] . ' attributes_price_factor_offset: ' . $attribute_price->fields['attributes_price_factor_offset'] . '<br>';
-                            $added_charge = zen_get_attributes_price_factor($chk_price, $chk_special, $attribute_price->fields['attributes_price_factor'], $attribute_price->fields['attributes_price_factor_offset']);
+                    ////////////////////////////////////////////////
+                    // calculate additional attribute charges
+                    $chk_price = zen_get_products_base_price($uprid);
+                    $chk_special = zen_get_products_special_price($uprid, false);
+                    // products_options_value_text
+                    if (ATTRIBUTES_ENABLED_TEXT_PRICES === 'true' && zen_get_attributes_type($check_attribute) == PRODUCTS_OPTIONS_TYPE_TEXT) {
```
drbyte, line 835: ❓ I think there's a bug here. `$check_attribute` got removed from old line 786, and was inlined as just `$attributes_id` on new line 816.

includes/classes/shopping_cart.php
drbyte: Update: yes, definitely must do this, for safety, since you've removed the sanitization from the query on line 751 below.

includes/classes/shopping_cart.php
```diff
@@ -634,107 +660,104 @@ public function calculate()
...
+            $this->notify('NOTIFY_CART_CALCULATE_PRODUCT_PRICE', $uprid, $product);
+            $prid = (int)$product['products_id'];   //-TODO: Needed?  Just in case the id was changed by the observer?
 
-                $this->notify('NOTIFY_CART_CALCULATE_PRODUCT_PRICE', $products_id, $product->fields);
+            $products_tax = zen_get_tax_rate($product['products_tax_class_id']);
+            $products_price = $product['products_price'];
 
-                $products_tax = zen_get_tax_rate($product->fields['products_tax_class_id']);
-                $products_price = $product->fields['products_price'];
+            $is_free_shipping = $product['product_is_always_free_shipping'] === '1' || $product['products_virtual'] === '1' || strpos($product['products_model'], 'GIFT') === 0;
```
drbyte, line 692: Nice de-duplication! 👍

includes/classes/shopping_cart.php
```diff
@@ -989,62 +1038,75 @@ public function attributes_price($product_id)
...
+                    $sale_maker_discount = '';
+                    $new_attributes_price = zen_get_discount_calc(
+                        $prid,
+                        $attribute_price->fields['products_attributes_id'],
+                        $attribute_price->fields['options_values_price'] +
+                            (zen_get_products_price_is_priced_by_attributes($attribute_price->fields['products_id']) ?
+                                zen_products_lookup($attribute_price->fields['products_id'], 'products_price') : 0.0),
+                        $qty
+                    );
+                    $new_attributes_price -= (zen_get_products_price_is_priced_by_attributes($prid)) ? zen_products_lookup($prid, 'products_price') : 0;
```
drbyte, line 1073: I'd rather put the closing `)` at the end of the line, instead of at the end of the function call. Or skip the wrapping parentheses.

includes/classes/shopping_cart.php
```diff
@@ -1071,75 +1133,68 @@ public function attributes_price($product_id)
...
+     * @param mixed $uprid the product ID of the item to check
      * @param float $qty item quantity
      * @return float the price of the items attributes
      */
-    public function attributes_price_onetime_charges($product_id, $qty)
+    public function attributes_price_onetime_charges($uprid, $qty)
     {
         global $db;
 
         $attributes_price_onetime = 0;
```
drbyte, line 1144: For consistency with the other refactorings to early-return, this line 1144 `$attributes_price_onetime = 0;` should be moved below the early-return on line 1149-1150.

includes/classes/shopping_cart.php
```diff
@@ -1212,181 +1259,197 @@ public function get_products($check_for_valid_cart = false)
...
-                $special_price = zen_get_products_special_price($prid);
-                if ($special_price && $products->fields['products_priced_by_attribute'] == 0) {
-                    $products_price = $special_price;
-                } else {
-                    $special_price = 0;
-                }
+        foreach ($this->contents as $uprid => $data) {
+            $products = zen_get_product_details((int)$uprid);
+            if ($products->EOF) {
+                continue;   //-TODO: Product not found, should it be removed from the cart?
```
drbyte, line 1270: Again, good question.

includes/classes/shopping_cart.php
```diff
@@ -1212,181 +1259,197 @@ public function get_products($check_for_valid_cart = false)
...
-                    // Check Quantity Max if not already an error on Minimum
-                    if ($fix_once == 0) {
-                        if ($products->fields['products_quantity_order_max'] != 0 && $check_quantity > $products->fields['products_quantity_order_max'] && !isset($this->flag_duplicate_quantity_msgs_set[(int)$prid]['max'])) {
+                        $chk_attributes_exist = $db->Execute($sql);
+                        if ($chk_attributes_exist->EOF) {
                             $fix_once++;
                             $_SESSION['valid_to_checkout'] = false;
-                            $_SESSION['cart_errors'] .= ERROR_PRODUCT . $products->fields['products_name'] . ERROR_PRODUCT_QUANTITY_MAX_SHOPPING_CART . ERROR_PRODUCT_QUANTITY_ORDERED . $check_quantity . ' <span class="alertBlack">' . zen_get_products_quantity_min_units_display((int)$prid, false, true) . '</span> ' . '<br>';
-                            $this->flag_duplicate_quantity_msgs_set[(int)$prid]['max'] = true;
+                            $chk_products_link =
```
drbyte, line 1333: Q: This removed the check for `products_status`, and now renders a link regardless of status, instead of only displaying the name if the status is 0 (product disabled).
Was this intended?

includes/classes/shopping_cart.php
```diff
@@ -1451,172 +1514,139 @@ public function get_content_type($gv_only = false)
...
-                                default:
-                                    if ($free_ship_check->fields['products_virtual'] == '1') {
-                                        $this->content_type = 'virtual';
-                                    } else {
-                                        $this->content_type = 'physical';
-                                    }
-                            }
+                        // -----
+                        // If the product doesn't have downloads, its virtual setting dictates
+                        // whether it' a virtual/physical product.
```
drbyte, line 1574: Missing `s`

includes/classes/shopping_cart.php
```diff
@@ -1762,114 +1792,141 @@ public function actionUpdateProduct($goto, $parameters)
...
+                    }
 
-                        $attributes = isset($_POST['id'][$_POST['products_id'][$i]]) ? $_POST['id'][$_POST['products_id'][$i]] : [];
-                        $this->add_cart($_POST['products_id'][$i], $new_qty, $attributes, false);
-                    } else {
-                        // adjust minimum and units
-                        $attributes = isset($_POST['id'][$_POST['products_id'][$i]]) ? $_POST['id'][$_POST['products_id'][$i]] : [];
-                        $this->add_cart($_POST['products_id'][$i], $new_qty, $attributes, false);
+                    // -----
+                    // Message the customer regarind the quantity adjustment.
```
drbyte, line 1906: Typo.

drbyte (changes requested):
Looks great. Thanks. That's a huge overhaul, optimizing a ton of things.

A couple small bugs highlighted.

Brings back memories of when a bunch of this code was written ...  ... shaking my head :D

includes/classes/shopping_cart.php
drbyte: Update: Yes. Matched.

includes/classes/shopping_cart.php
torvista: Yes. Why allow a product that does not exist?

includes/classes/shopping_cart.php
torvista: Yes. Why allow a product that does not exist?

includes/classes/shopping_cart.php
lat9: Yes, there is!  Update forthcoming.

includes/classes/shopping_cart.php
lat9: Noting that this section has changed "a fair amount" for #6017 (sigh).

includes/classes/shopping_cart.php
lat9: This section isn't run if the `products_status` is 0; bypassed by the processing on lines 1314-1320 above.

lat9:
I believe that I've got all the corrections/comments addressed in the just-pushed update.

drbyte (approved):

torvista:
If you set 
`declare(strict_types=1);`
to the start of this file, and add a product to the cart, it spits out a series of errors due to mismatched casting.

lat9:
Thanks, @torvista.  I'll address that in the changes associated with #6017, since the shopping_cart class requires updates there as well.

torvista:
There's only 12 errors, but some may generate a "big-picture" discussion, so it may be prudent to hoist it above the parapet as a separate thing to throw mud at, in the interests of mixing metaphors.

PR #6015, opened 2023-10-17
Formatting

Zen4All:
For years we have been trying to get rid of shortcode, and now we are introducing it again?

### DIFF
--- a/admin/featured.php
+++ b/admin/featured.php
@@ -253,7 +253,7 @@ if (!empty($action)) {
             <div class="form-group">
               <?php echo zen_draw_label(TEXT_FEATURED_PRODUCT, 'products_id', 'class="col-sm-3 control-label"'); ?>
               <div class="col-sm-9 col-md-6">
-                <?= zen_draw_pulldown_products('products_id', 'required size="15" class="form-control" id="products_id"', $featured_array, true, (!empty($_GET['add_products_id']) ? $_GET['add_products_id'] : ''), true); ?>
+                <?php echo zen_draw_pulldown_products('products_id', 'required size="15" class="form-control" id="products_id"', $featured_array, true, (!empty($_GET['add_products_id']) ? $_GET['add_products_id'] : ''), true); ?>
               </div>
             </div>
           <?php } ?>
--- a/admin/specials.php
+++ b/admin/specials.php
@@ -324,7 +324,7 @@ if (!empty($action)) {
             <div class="form-group">
               <?php echo zen_draw_label(TEXT_SPECIALS_PRODUCT, 'products_id', 'class="col-sm-3 control-label"'); ?>
               <div class="col-sm-9 col-md-6">
-                <?= zen_draw_pulldown_products('products_id', 'required size="15" class="form-control" id="products_id"', $specials_array, true, (!empty($_GET['add_products_id']) ? $_GET['add_products_id'] : ''), true); ?>
+                <?php echo zen_draw_pulldown_products('products_id', 'required size="15" class="form-control" id="products_id"', $specials_array, true, (!empty($_GET['add_products_id']) ? $_GET['add_products_id'] : ''), true); ?>
               </div>
             </div>
           <?php } ?>


lat9:
FWIW, it was the PHP short tag `<?` that was being replaced with `<?php` due to XML having chosen the same short-code for its operation.  That short tag is actually deprecated in PHP 7.4.

That's quite different from the PHP short echo tag `<?=` to which you refer, which is still alive and well. I'll leave it to the community for a vote as I can (obviously) go either way.  It's kind of like _finally_ accepting the usage of the null-coalesce (`??`) operator.

proseLA:
`<?=` is shorthand for `<?php echo`.  it is NOT a short tag.  and it can not be disabled.

IMO, it makes code, where there is a mix of html and php, far more readable.

i would suggest doing a little research on the subject.  [this link](https://softwareengineering.stackexchange.com/questions/151661/is-it-bad-practice-to-use-tag-in-php#:~:text=before%20the%20misinformation%20train%20goes%20too%20far%20out%20of%20the%20station%2C%20there%20are%20a%20bunch%20of%20things%20you%20need%20to%20understand%20about%20php%20short%20tags.) and [this one](https://www.php.net/manual/en/language.basic-syntax.phptags.php#:~:text=Note%3A,to%0A%20%20%20%20%20%20maximise%20compatibility.) might shed some light.

drbyte:
The Zen Cart code format has always leaned to avoiding `<?=`, in favor of the slightly more verbose `<?php echo`.

You can of course do what you want with plugins or your own code, but core code avoids `<?=` style.

The only uses of `<?=` in core code were introduced in https://github.com/zencart/zencart/pull/5573 and https://github.com/zencart/zencart/pull/5994 .
For consistency, they should be reformatted.

proseLA:
> For consistency, they should be reformatted.

i do not think this is a good enough reason.  

looking at [the coding standards in the docs](https://docs.zen-cart.com/dev/contributing/coding_standards/), there has always been an evolution.  old style to new style...  and yet, i have found developers with write access to this repo writing and merging code that does not conform to the supposed new standard of psr-2.

discussions on coding standards, for whatever reasons, always seem contentious. i am not sure why...

chris, during your 'hiatus', there was a long discussion about coding standards and unfortunately we could not come to a consensus.  i played around with a few tools, ie php-cs-fixer, and the like, in an attempt to provide a tool that all developers, irrespective of individual operating environments, could use to format their code according to some standard.  unfortunately that effort did not bear fruit.

if consistency is what you desire, i would advocate for a wholesale replacement of all uses of `<?php echo ` to `<?= `.  i think one could very easily start that PR using:
```
grep -rl '<?php echo ' ./ | xargs sed -i -e  's/<?php echo /<?= /g'
```
while i can not find the original documentation, Rasmus Lerdorf - endorses the shorthand `<?=` tag.  if it is good enough for him, why is it not acceptable here?

finally, i would advocate for supporting the developers who continue to support this project.  and developers that make attempts to move it forward.  there really is not that many of us left.

drbyte:
> discussions on coding standards, for whatever reasons, always seem contentious. i am not sure why...

Until now there was a respect for the pain that wholesale reformatting would cause to people trying to do upgrades to live stores. And also to maintain the ability to continue to run on older PHP versions whilst performing said upgrades without dropping all older syntax styles.
That's why there was an intentional slower adoption of using all the bleeding-edge stuff that developers love to sink their teeth into. It was actually a very difficult thing to impose, because the core team also wanted (still does) to do more bleeding-edge stuff.

But, I see that there's been a sharp drop in caring about those pain points, in favor of telling end-users to "just upgrade, no matter how hard we've made it by reformatting all the files and rearranging all the goal-posts on the field".

drbyte:
> Rasmus Lerdorf ... if it is good enough for him, why is it not acceptable here?

Nice one.

> finally, i would advocate for supporting the developers who continue to support this project.

So ... just to be clear, are you saying that the use of `<?php echo` syntax over `<?=` is the equivalent of "not supporting the developers"?

proseLA:
> Until now there was a respect for the pain that wholesale reformatting would cause to people trying to do upgrades to live stores. 

i agree that upgrading is a pain.  and i am not for wholesale reformatting; but i am for the allowed use of the shorthand echo tag; and if that means you want to do a wholesale reformat; so be it.  i find code legibility vastly improved.

> And also to maintain the ability to continue to run on older PHP versions whilst performing said upgrades without dropping all older syntax styles. That's why there was an intentional slower adoption of using all the bleeding-edge stuff that developers love to sink their teeth into. 

the shorthand version of the echo tag was introduced into the core in php 5.4.  that was over 11 years ago.  this is hardly bleeding edge stuff.

> 
> But, I see that there's been a sharp drop in caring about those pain points, in favor of telling end-users to "just upgrade, no matter how hard we've made it by reformatting all the files and rearranging all the goal-posts on the field".

as i have previously stated, i am not in favor of reformatting files that have not been touched.  i still care about the pain.  for me, this discussion is about the potentially disallowed use of a shorthand echo tag on a file that was already touched for another reason.

> > finally, i would advocate for supporting the developers who continue to support this project.
> 
> So ... just to be clear, are you saying that the use of `<?php echo` syntax over `<?=` is the equivalent of "not supporting the developers"?

i think this comment is a bit of red herring.  this PR clearly states 'Fix php shortcode'; and i am opposed to this PR.  and i all for the use of the shorthand version of the php echo command.

in looking at the 2 PRs mentioned above, i can clearly see the usefulness and value of both of them.  IMO, they are both improvements to the function of ZC.   i am not sure what signal one would be sending by changing that code now for a stylistic reason.  the repopulate search box PR would certainly be in the camp of causing more upgrade pain for no good reason.  again, IMO.

torvista:
>it makes code, where there is a **mix** of html and php, far more readable.

I would say allow it where php is  **embedded** in html, as files get modified, as it is an improvement.

gburton:
Changing from <?php echo to <?= IS painful.  We started this task of updating to short echo tags and it took a number of releases to get it 100%;

`<?php echo 'blah'; ?>`
sometimes see this
`<?= 'blah'; ?>`
and sometimes this
`<?= 'blah' ?>`

In the end we've had to advise that our preferred "gold standard" is;

`<?= 'blah' ?>`

IE;  short echo tag with no trailing semi-colon.

Doing a grep replace of <?php echo to <?= wasn't enough.
That trailing ; also needs removing.

In the end it was done mostly manually and did cause some shopowner issues, but we got past it.

I'm pretty sure that our core has no instance of <?php echo
Addons are sometimes still using <?php echo

It's hard to learn new ways of doing things.

**The upshot;**

Code is easier to read 
We're still finding semi-colons that need removing  :-\

torvista:
> In the end we've had to advise that our preferred "gold standard" is;
> short echo tag with no trailing semi-colon.

Why?

scottcwilson:
Gary, do you mean osC 4 or osC 2?

gburton:
> Gary, do you mean osC 4 or osC 2?

Never seen osC 4.  
And osC 2 is a Dodo now.

Phoenix.  

We also went down the array(a,b,c) to [a,b,c] avenue.  That was not fun either.

torvista:
> In the end we've had to advise that our preferred "gold standard" is;
> short echo tag with no trailing semi-colon.

I meant why no trailing semi-colon? To the semi-untrained eye it looks unfinished without it.

gburton:
> I meant why no trailing semi-colon? To the semi-untrained eye it looks unfinished without it.

The closing tag of a block of PHP code automatically implies a semicolon;
you do not need to have a semicolon terminating the last line of a PHP block.

As `<?= 'xyz' ?>` is a block of PHP code it needs no ending ;

It is, of course, a choice that another project might decide is not for them.  
One fewer character to have to type is usually "a good thing";

PR #6017, opened 2023-10-17
Wholesale pricing

lat9:
Per the discussion.  Prerequisites #6007 and #6011


lat9:
FWIW, I have no idea what that 8.0 unit-test failure is trying to "tell" me.

torvista:
php 8.3.0RC2
On updating a customer:
--> PHP Warning: Undefined array key "customers_whole" in ...\admin\customers.php on line 103.

torvista:
Attribute pricing does not appear to work as expected.

Matrox 200: set 10% wholesale discount level 1. Customer has wholesale level 1.

Base price not logged in: 231.89
Base price logged in: 208.70 correct, is 10% off

Premium/16MB not logged in: 363.30
Premium/16MB logged in: 340.11 incorrect, 10% off is 326.97

lat9:
> php 8.3.0RC2 On updating a customer: --> PHP Warning: Undefined array key "customers_whole" in ...\admin\customers.php on line 103.
> 
> I see this in the file ` $customers_whole = (int)$_POST['customers_whole'] ?? 0; //- Not present if Wholesale Pricing isn't enabled for the site or a group-pricing group's selected!`
> 
> I assumed the wholesale pricing would use the customer group 10%, so had selected that.
> 
> I see on further investigation there is now a My Store->Wholesale option. With this to true and no group selected, there is no warning. Still, the reported error should be trapped.

Did you run the zc_install updates which added those fields and configuration setting to the database?

lat9:
> Attribute pricing does not appear to work as expected.
> 
> Matrox 200: set 10% wholesale discount level 1. Customer has wholesale level 1.
> 
> Base price not logged in: 231.89 Base price logged in: 208.70 correct, is 10% off
> 
> Premium/16MB not logged in: 363.30 Premium/16MB logged in: 340.11 incorrect, 10% off is 326.97
> 
> Also lots of these
> 
> > #0 ...\includes\classes\shopping_cart.php(829): zen_debug_error_handler()
> > #1 ...\includes\classes\shopping_cart.php(1458): shoppingCart->calculate()
> > #2 ...\includes\modules\pages\shopping_cart\header_php.php(29): shoppingCart->show_weight()
> > #3 ...\index.php(35): require('D:\OneDrive\Dev...')
> > --> PHP Warning: Undefined variable $check_attribute in ...\includes\classes\shopping_cart.php on line 829.
> 
> As I've mentioned before, I did this myself and hacked my way through discount code hell, coming to the conclusion that the whole pricing/discount/attributes code should be overhauled first before attempting to add this functionality in core, much as it is of massive use to me.

Yep, @drbyte had pointed out that error in the previous (#6011) shopping-cart update.  I'll be pulling that in here as well.

lat9:
> php 8.3.0RC2 On updating a customer: --> PHP Warning: Undefined array key "customers_whole" in ...\admin\customers.php on line 103.
> 
> I see this in the file ` $customers_whole = (int)$_POST['customers_whole'] ?? 0; //- Not present if Wholesale Pricing isn't enabled for the site or a group-pricing group's selected!`
> 
> I assumed the wholesale pricing would use the customer group 10%, so had selected that.
> 
> I see on further investigation there is now a My Store->Wholesale option. With this to true and no group selected, there is no warning. Still, the reported error should be trapped.

Yep, needs parentheses around the null-coalesce, since the `int` casting has priority.

lat9:
Rebasing with master, to get changes for #6007.

admin/customers.php
```diff
@@ -91,7 +91,7 @@
                 $customers_fax = zen_db_prepare_input($_POST['customers_fax']);
             }
             $customers_newsletter = zen_db_prepare_input($_POST['customers_newsletter']);
-            $customers_group_pricing = (int)$_POST['customers_group_pricing'];
+            $customers_group_pricing = (int)$_POST['customers_group_pricing'] ?? 0; //- No present if wholesale pricing is selected!
```
drbyte, line 94: 'No' vs 'Not' typo in comment.

drbyte:
> FWIW, I have no idea what that 8.0 unit-test failure is trying to "tell" me.

The test is looking for this string to be present on the Attributes Controller page: 
Screen Shot 2023-10-18 at 5 41 39 PM

But it looks like it wasn't fully drawing that page (probably the bug Torvista mentioned), so that could be why the string didn't get rendered.

lat9:
Previous commit to rebase with master's incorporation of #6011.

admin/customers.php
lat9: Yep, and casting has a higher priority than null-coalesce (sheesh).

lat9:
> > php 8.3.0RC2 On updating a customer: --> PHP Warning: Undefined array key "customers_whole" in ...\admin\customers.php on line 103.
> > I see this in the file ` $customers_whole = (int)$_POST['customers_whole'] ?? 0; //- Not present if Wholesale Pricing isn't enabled for the site or a group-pricing group's selected!`
> > I assumed the wholesale pricing would use the customer group 10%, so had selected that.
> > I see on further investigation there is now a My Store->Wholesale option. With this to true and no group selected, there is no warning. Still, the reported error should be trapped.
> 
> Did you run the zc_install updates which added those fields and configuration setting to the database?

Wholesale pricing is not compatible with customers' group-pricing discounts.  I still need to update the customers.php to check for that condition and present an error to the admin if both are chosen.

torvista:
Testing with latest changes still yields unexpected results.

| Matrox G200: 10% |   guest price      | 10%     |  wh price | %   |
|------------------|---------|---------|:-------:|-----|
| value, 4MB       | 231.89  | 208.70  | 208.70  | 10% |
| + premium, 4MB   | 309.19  | 278.27  | 286.00  | 8%  |
| + premium + 8MB  | 347.84  | 313.06  | 324.65  | 7%  |
| + premium + 16MB | 363.30  | 326.97  | 340.11  | 6%  |
| totals           | 1252.22 | 1127.00 | 1159.46 | 7%  |

I would suggest the cart should show the original price and discount applied.

And for a separate bun fight a more logical column layout would be 

model | name | original unit price | discount applied | final unit price | quantity | total

with a single update button next to the grand total instead of that huge icon wasting space on every line.

torvista:
> What does the top-left column heading (Matrox 200 10%) mean? Where is that 10% applied?

It means that product has a single wholesale price of 10% defined in the product edit.

> What does the 10% column represent?

It shows what 10%-off **should** calculate to.

> What does the wh price % column represent?

It shows the % discount shown in the cart, which is expected to be the wholesale discount of 10%, but is actually something else for all but the base price.

> The cart has never shown the original price and discount applied (that would be a separate PR) for specials and sales (which is what the Wholesale Pricing is modeled after).

Sure, but we use sites that are trade suppliers which is what you are trying implement here, and the original price is always displayed, as that is what the shop will probably use that as the RRP/to charge their customer, and they want to know what their discount (=profit) is:  as this will vary per product/per product family/per manufacturer.....

lat9:
Let's start, then, with the first row:

Matrox 200 (Base price 299.99).  10% off is 269.99 (that's the price for the Value/4mb option-combination).

I'm still not understanding where the pricing you've identified is coming from.

lat9:
> > FWIW, I have no idea what that 8.0 unit-test failure is trying to "tell" me.
> 
> The test is looking for this string to be present on the Attributes Controller page: Screen Shot 2023-10-18 at 5 41 39 PM
> 
> But it looks like it wasn't fully drawing that page (probably the bug Torvista mentioned), so that could be why the string didn't get rendered.

Here's what I see when logging in to the admin and clicking **_Catalog :: Attributes Controller_** (so I still don't get what the functional test failure's trying to "tell me"):
image

torvista:
I'm getting multiple problems in testing, so have to set up a remote site...

proseLA:
> > Testing with latest changes still yields unexpected results.
> > Matrox G200: 10% 	guest price 	10% 	wh price 	%
> > value, 4MB 	231.89 	208.70 	208.70 	10%
> > 
> > * premium, 4MB 	309.19 	278.27 	286.00 	8%
> > * premium + 8MB 	347.84 	313.06 	324.65 	7%
> > * premium + 16MB 	363.30 	326.97 	340.11 	6%
> >   totals 	1252.22 	1127.00 	1159.46 	7%
> 
> I don't understand what you're trying to convey with that table or how you've configured that product.
> 
> * What does the top-left column heading (Matrox 200 10%) mean?  Where is that 10% applied?
> * What does the 10% column represent?
> * What does the wh price % column represent?

i think what steve is saying is the the 10% is only coming off of the base unit price, ie $231.89.  10% of that is $23.19.  so the first line is (value 4 mb) is correct.

in steve's ideas of wholesale pricing the additional cost of the attributes should also have 10% off.  they do not.  each of the next 3 lines, the whole sale price is reduced by 10% of the base price, or $23.19.  (i have ZERO idea why he included the totals line.)

so his 5 column descriptions seem to be:

- description 
- regular price 
- what he thinks the wholesale price should be (ie 10% off the total) 
- what your code shows as the price
- the percentage difference that your code shows against the base

hope that clarifies what steve is showing.

with regards to your point of the item in the test data is $299.99, and steve is showing it as $231.89, i think that's an excellent question.  i think the whole point of the test data is so that everyone can be looking at the same thing.  how he got to that starting point, i do not know.

torvista:
Due to time constraints I've got no further but can discount (ha!) my previous test:
- currency was not updated
- attribute pricing has to be discounted manually in addition to the base price.

Once done, the pricing was as expected.

lat9:
I'm thinking that the help-text on the product-entry pages should contain a disclaimer regarding the attributes' settings being separate.

lat9:
OK, this branch (with all the recent changes) is getting to be quite the PITA to keep it up-to-date.  We're either going to merge this and run with it or I'm just going to withdraw the PR.

torvista:
Apologies for not investigating this further at the time. Too many sidetracks.
Once merged I'll certainly be using and improving it.

torvista:
Great. Next, Product Kits.

drbyte:
For future troubleshooting: If one has not run the `zc_install` db updates, the following 2 kinds of errors will be logged:

This one is slightly more obvious:
```
--> PHP Warning: Undefined array key "products_price_w" in /includes/functions/functions_prices.php on line 468.
```

This one is more obscure, masked by the fact that the NULL is caused by missing db schema updates:
```
--> PHP Fatal error:  Uncaught TypeError: zen_get_retail_or_wholesale_price(): Argument #2 ($wholesale_pricing_tier) must be of type string, null given, called in /includes/functions/functions_prices.php on line 468 and defined in /includes/functions/functions_prices.php:1628
Stack trace:
#0 /includes/functions/functions_prices.php(468): zen_get_retail_or_wholesale_price('3.9500', NULL)
#1 /includes/functions/functions_prices.php(186): zen_get_products_base_price('169')
#2 /includes/templates/template_default/sideboxes/tpl_featured.php(15): zen_get_products_display_price('169')
```

It's possible we may be able to add defensive code to avoid the fatal errors, perhaps just failing silently if the feature isn't usable due to the missing schema updates.

drbyte:
> My feeling is that if `zc_install` (or at least the associated schema updates) hasn't been run, the user of the current branch has shot themselves in the foot.

Agreed. 
Probably the best way is to do our usual: find things in the schema updates that zc_install should detect in order to give the necessary warnings if they're not present.

... so, gotta look for something in those changes that wasn't part of the original wholesale_pricing plugin.
There are already other factors detected which indicate a need for upgrade to v2.0.0 ... but finding one for this feature will be wise too ... if only to flag it for people who are using the unreleased v2 code, whether on local or even live (yes, I see evidence of people using it live 🙈)

drbyte:
Gonna use:
- configuration table `'Wholesale Pricing', 'WHOLESALE_PRICING_CONFIG'`
- orders table `is_wholesale tinyint(1) DEFAULT NULL`

scottcwilson:
> I'm thinking that the help-text on the product-entry pages should contain a disclaimer regarding the attributes' settings being separate.

Cindy, can you be more specific about what you'd like to see on which page?  I'm happy to do it, I'm just not sure I'm tracking what the issue is. 

Would a statement at the bottom of 

https://docs.zen-cart.com/user/products/wholesale_pricing/#setting-wholesale-price-per-product

directing people to 

https://docs.zen-cart.com/user/products/attribute_pricing/#attribute-quantity-discounting-vs-product-quantity-discounting

for attribute discounting suffice?  Or are you thinking more?

Issue #5911, opened 2023-08-01
'master' is now the official branch to push to


drbyte:
UPDATE: We've switched to using `master` for active development on the "next" release.
Therefore we've renamed `v200` to `master`.
Be sure to update your local git configuration. Github's suggestion via command-line is:
<img width="386" alt="Screen Shot 2023-10-09 at 2 10 35 PM" src="https://github.com/zencart/zencart/assets/404472/82a830b0-c62d-40f3-9409-0c4aaa8721cb">

Original post below:
---
Since v2.0.0 will be the number of the next official Zen Cart release, a new `v200` branch has been created.

All new pull requests should be to this branch.

You can clone that `v200` branch into your own repository, from inside your own Github account.
Once you have cloned it, you can pull that branch down to your local PC to work with it. You can periodically click the button to sync that branch with your own as well.

Here's how to clone the new branch into your own github repo, from inside your own github account page:

![1 Go to Branches page](https://github.com/zencart/zencart/assets/404472/25a5fabb-1312-487f-82af-bc5ba461af5f)

![2  New Branch](https://github.com/zencart/zencart/assets/404472/aaa83d67-aea6-4a9b-bed3-48d7268062f2)

Set the fields as shown, and click the button:
<img width="478" alt="3 Create the branch from upstream" src="https://github.com/zencart/zencart/assets/404472/38517c5e-1dee-46c9-bda5-6d085e53c0a9">

And now make that new branch the "Default" branch for the repo:
![4 switch default branch](https://github.com/zencart/zencart/assets/404472/c9982e02-4e39-41f7-87d7-5df9b3440674)

Now you can sync that `v200` branch from your Github to your Local, and on your local you can branch off of it to submit any pull requests.
First check whether the branch on zencart/zencart has been updated (it will show how many commits it is out of sync). Click the button to Sync your own fork with Zen Cart. 

![5 Sync from origin (out of sync)](https://github.com/zencart/zencart/assets/404472/d01734a7-1561-4f61-8093-a1daba488fdb)

Then pull your v200 branch down to your local.
And if you're working on a Pull Request, branch from this newly updated v200 branch before submitting the PR.

@zencart/contributors

drbyte:
> @drbyte: It is not possible to merge into "main"?

Considering it ...

brittainmark:
[...]

PR #6036, opened 2023-11-02
Refactor search SQL operations into helper class, catalog side.

neekfenwick:
Took the bulk of search_result/header.php which built the SQL and put it in a helper class Zencart\Search\Search.
Inputs to the search refactored into SearchOptions so they can be called programatically without needing a GET request to trigger it. Parsing of $_GET retained in the search/header.php and search_result/header.php.

New message_stack: 'search_result' for displaying a message at the top of
  the search results page (for example 'try also doing xyz').
New notifier: NOTIFY_SEARCH_NO_RESULTS_MESSAGE, given $results, $message
  For reacting to the search results e.g. add to message stack (to appear
  on the 'search' form).

### DIFF
--- a/includes/auto_loaders/config.core.php
+++ b/includes/auto_loaders/config.core.php
@@ -73,6 +73,10 @@ if (!defined('USE_PCONNECT')) define('USE_PCONNECT', 'false');
   $autoLoadConfig[0][] = array('autoType'=>'class',
                                'loadFile'=>'Customer.php');
   $autoLoadConfig[0][] = [
+    'autoType'=>'class',
+    'loadFile'=>'class.search.php'
+  ];
+  $autoLoadConfig[0][] = [
     'autoType' => 'class',
     'loadFile' => 'zcDate.php'
   ];
@@ -170,6 +174,11 @@ $autoLoadConfig[55][] = [
                                 'objectName'=>'cart',
                                 'checkInstantiated'=>true,
                                 'classSession'=>true);
+  $autoLoadConfig[80][] = [
+    'autoType'=>'classInstantiate',
+    'className'=>'Zencart\Search\Search',
+    'objectName'=>'search'
+  ];
 /**
  * Breakpoint 90.
  *
--- a/includes/classes/class.search.php
+++ b/includes/classes/class.search.php
@@ -1,0 +1,461 @@
+<?php
+/**
+ * Product search SQL operations.
+ *
+ * @copyright Copyright 2003-2023 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: pRose on charmes 2023 Feb 03 Modified in v1.5.8a $
+ */
+
+namespace Zencart\Search;
+
+use Zencart\Exceptions\SearchException;
+
+/**
+ * Search Options for when searching the product catalogue.
+ * Constructor attempts to initialise from $_GET query parameters, or
+ * alternatively the fields may be set directly.
+ */
+class SearchOptions
+{
+    public string $keywords;
+    public string $dfrom;
+    public string $dto;
+    public float $pfrom;
+    public float $pto;
+    public int $categories_id;
+    public bool $inc_subcat;
+    public bool $search_in_description;
+    public ?int $manufacturers_id;
+    public int $alpha_filter_id;
+    public string $sort;
+
+    /**
+     * Attempt to initialise from $_GET.
+     * May throw SearchException if fields fail to parse.
+     */
+    public function __construct()
+    {
+        $this->keywords = $_GET['keyword'] ?? '';
+        $this->dfrom = $_GET['dfrom'] ?? '';
+        $this->dto = $_GET['dto'] ?? '';
+        $this->categories_id = (int)($_GET['categories_id'] ?? 0);
+        $this->inc_subcat = ($_GET['inc_subcat'] ?? '0') === '1';
+        $this->search_in_description = ($_GET['search_in_description'] ?? '0') === '1';
+        $this->manufacturers_id = (int)($_GET['manufacturers_id'] ?? 0);
+        $this->alpha_filter_id = (int)($_GET['alpha_filter_id'] ?? 0);
+        $this->sort = $_GET['sort'] ?? '';
+
+        // Parse inputs that might fail due to syntax.
+        if (!empty($_GET['pfrom']) && !is_numeric($_GET['pfrom'])) {
+            throw new SearchException(ERROR_PRICE_FROM_MUST_BE_NUM);
+        }
+        $this->pfrom = (float)($_GET['pfrom'] ?? '');
+
+        if (!empty($_GET['pto']) && !is_numeric($_GET['pto'])) {
+            throw new SearchException(ERROR_PRICE_TO_MUST_BE_NUM);
+        }
+        $this->pto = (float)($_GET['pto'] ?? '');
+    }
+}
+
+/**
+ * Helper class to perform searches of the product catalogue.
+ */
+class Search extends \base
+{
+    /** Options used for our building operations. */
+    protected SearchOptions $searchOptions;
+
+    /**
+     * Return the current SearchOptions, if any.
+     *
+     * @return SearchOptions
+     */
+    public function getSearchOptions(): SearchOptions
+    {
+        return $this->searchOptions;
+    }
+
+    /**
+     * Set the SearchOptions to be used in operations.
+     *
+     * @param SearchOptions $searchOptions
+     * @return void
+     */
+    public function setSearchOptions(SearchOptions $searchOptions)
+    {
+        $this->searchOptions = $searchOptions;
+    }
+
+    /**
+     * Builds SQL for searching the product catalogue, given SearchOptions.
+     * Note: Sets the global $column_list, needed by product_listing template.
+     *
+     * @param SearchOptions $searchOptions The options for the search.
+     * @return string The built SQL.
+     */
+    public function buildSearchSQL() {
+        global $db, $messageStack, $currencies, $column_list;
+
+        if (empty($this->searchOptions)) {
+            throw new SearchException(ERROR_MISSING_SEARCH_OPTIONS);
+        }
+
+        // -----
+        // Give an observer the chance to indicate that there's another element to the search
+        // that **is** provided, enabling the search to continue.
+        //
+        $search_additional_clause = false;
+        $this->notify('NOTIFY_ADVANCED_SEARCH_RESULTS_ADDL_CLAUSE', [], $search_additional_clause);
+
+        if ($search_additional_clause === false &&
+            empty($this->searchOptions->keywords) &&
+            (
+                (empty($this->searchOptions->dfrom) && empty($this->searchOptions->dto)) &&
+                (empty($this->searchOptions->pfrom) || $this->searchOptions->pfrom <= 0) &&
+                (empty($this->searchOptions->pto) || $this->searchOptions->pto <= 0)
+            )) {
+            throw new SearchException(ERROR_AT_LEAST_ONE_INPUT);
+        } else {
+            $dfrom_array = [];
+            $dto_array = [];
+
+            if (!empty($this->searchOptions->dfrom) &&
+                !zen_checkdate($this->searchOptions->dfrom, DOB_FORMAT_STRING, $dfrom_array)) {
+                throw new SearchException(ERROR_INVALID_FROM_DATE);
+            }
+
+            if (!empty($this->searchOptions->dto) &&
+                !zen_checkdate($this->searchOptions->dto, DOB_FORMAT_STRING, $dto_array)) {
+                throw new SearchException(ERROR_INVALID_TO_DATE);
+            }
+
+            if (!empty($this->searchOptions->dfrom) && !empty($this->searchOptions->dto)) {
+                if (mktime(0, 0, 0, $dfrom_array[1], $dfrom_array[2], $dfrom_array[0]) > mktime(0, 0, 0, $dto_array[1], $dto_array[2], $dto_array[0])) {
+                    throw new SearchException(ERROR_TO_DATE_LESS_THAN_FROM_DATE);
+                }
+            }
+
+            if ($this->searchOptions->pfrom > $this->searchOptions->pto) {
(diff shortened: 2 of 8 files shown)


includes/auto_loaders/config.core.php
```diff
@@ -72,6 +72,8 @@
                                'objectName'=>'zcPassword');
   $autoLoadConfig[0][] = array('autoType'=>'class',
                                'loadFile'=>'Customer.php');
+  $autoLoadConfig[0][] = array('autoType'=>'class',
```
proseLA, line 75: i see no reason why you should not use the short array syntax

includes/auto_loaders/config.core.php
```diff
@@ -170,6 +172,9 @@
                                 'objectName'=>'cart',
                                 'checkInstantiated'=>true,
                                 'classSession'=>true);
+  $autoLoadConfig[80][] = array('autoType'=>'classInstantiate',
```
proseLA, line 175: same comment as above with regards to the short array syntax

includes/modules/pages/search_result/header_php.php
```diff
@@ -2,11 +2,14 @@
 /**
  * Header code file for the Search Results page
  *
- * @copyright Copyright 2003-2023 Zen Cart Development Team
+ * @copyright Copyright 2003-2022 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: pRose on charmes 2023 Feb 03 Modified in v1.5.8a $
+ * @version $Id: DrByte 2020 Dec 25 New in v1.5.8-alpha $
```
proseLA, line 8: it looks like you are making a PR against the v1.5.8-alpha tag.  i believe all PRs should be going against the master branch.

IMO, changes in the stamp should only be done via the repo maintainers, and then when they are doing some sort of new tag or release.  otherwise it can get very confusing when doing upgrades.  if a PR changes the stamp, i would surmise something is a bit adrift with your process.

proseLA (changes requested):
big fan of moving this into a class file.

includes/classes/class.search.php
```diff
@@ -0,0 +1,434 @@
...
+            throw new SearchException(ERROR_PRICE_TO_MUST_BE_NUM);
+        }
+        $this->pto = (float)($_GET['pto'] ?? '');
+    }
+}
+
+/**
+ * Helper class to perform searches of the product catalogue.
+ */
+class Search
```
proseLA, line 60: this class file has some notifiers in it.  have you tested them? 

i believe this class file needs to extend base in order for those notifiers to work.

includes/auto_loaders/config.core.php
neekfenwick: Fair enough, I was trying to fit in with old code but I see zcDate's changes use the new syntax, and tighter indentation. I've updated my syntax.

includes/classes/class.search.php
neekfenwick: Good catch, I had rather stupidly imported the global `$zco_notifier` instead of deriving from `base` and using `$this->notify` .. this is now fixed.  I have tested with an observer and it is fired correctly.

includes/modules/pages/search_result/header_php.php
neekfenwick: I thought this work was development on v158, sorry.  I have rebased my change onto upstream/master and force pushed (I believe a force push is required in this case, a simple push was rejected by git).

neekfenwick:
Note to maintainer: This branch contained multiple commits; I ran into trouble earlier with my FontAwesome PR when you merged the branch plainly leaving multiple commits on `master`.  I have rebased this branch to squash my commits into a single one, I hope this helps.

includes/modules/pages/search_result/header_php.php
proseLA: no apologies necessary.

unfortunately, this still looks wrong.  i would refer to #5911 with regards to how PRs should work.

with regards to the `$Id` and other elements in the stamp, how they work is not important (at least to me).  i have opened issues and there have been discussions on them in the past, and i am comfortable with whatever process the maintainers use.  the point that i am making is that if your PR changes the stamp (and yours does), then there is a problem with your process of creating that PR.  again, i would refer to the above mentioned post.

however, in doing a cursory look, it does seem that your new class file, when using existing notifiers, those notifiers have the correct signature, ie they are the same as in the master branch; and not what is in the v1.5.8a tag.

as an example:

```
charmes: # git checkout v1.5.8a-alpha 
Note: switching to 'v1.5.8a-alpha'.

You are in 'detached HEAD' state
....
charmes: # git diff master includes/modules/pages/search_result/header_php.php
...
// Notifier Point
-$zco_notifier->notify('NOTIFY_SEARCH_COLUMNLIST_STRING', $select_column_list, $select_column_list);
+$zco_notifier->notify('NOTIFY_SEARCH_COLUMNLIST_STRING');

...
 
 // Notifier Point
-$zco_notifier->notify('NOTIFY_SEARCH_SELECT_STRING', $select_str, $select_str);
+$zco_notifier->notify('NOTIFY_SEARCH_SELECT_STRING');

```
in the above section of code, the added parameter elements to the notifiers are in the master branch, while the notifiers without parameters on in the v1.5.8a-alpha tag.  and when i look at your new class file for those notifiers, the parameters are there:

https://github.com/zencart/zencart/blob/86e2bd9690e46aaa9e61e26b88992ec03f592024/includes/classes/class.search.php#L198

https://github.com/zencart/zencart/blob/86e2bd9690e46aaa9e61e26b88992ec03f592024/includes/classes/class.search.php#L210

i'm not sure what more to say on it...

hopefully it makes sense to you on it.  and again, i am a fan of what you are doing here.

neekfenwick: I'm a little confused as to how those lines are wrong.  Github 'show files' on my PR shows, for example, the COLUMNLIST_STRING one looks like (from https://github.com/zencart/zencart/pull/6036/files) :

`        $this->notify('NOTIFY_SEARCH_COLUMNLIST_STRING', $select_column_list, $select_column_list);`

This is correct, no?  If you go back to the v1.5.8.a-alpha tag then the two parameters are missing, yes, but that's old code and was improved apparently by yourself at https://github.com/zencart/zencart/commit/fd0820894fbc1173e705bb515e625a9f29653524

Visible in latest code on `master`: https://github.com/zencart/zencart/blob/9ae0b145fff3d89bbeafdae83b1f9c00dbdf6788/includes/modules/pages/search_result/header_php.php#L211

Can you spell out what's wrong in a little more detail please?

proseLA: yes, the code looks correct.  (i have not tested it.)

you based your changes on the wrong branch; which is why the header comments have changed.  somehow you seem to have done the code correctly.  the commit you referenced, fd08208, was done 11/23/2022.  and yet the header that you are now changing is back to a release on 7/13/2022.

if you are confident that the code is correct, and no one else has a problem, i would not worry about it.  

if you still do no understand what i am saying, perhaps someone else can help clarify it.

neekfenwick: Well, "check twice, merge once" as someone somewhere probably says... Upon diffing against the original `search_results/php_header.php` I've found a couple of places `$_GET` was still used in the refactored code.  I've fixed these up to use the `$searchOptions` object as they should, but not committed or pushed yet, because of the next point.

I also note, the new class.search.php file has no Copyright comment at the top or place for the `$Id blah$` string. @drbyte could you please say what I should put in a brand new file?  Is it enough to have something like:

```
/**
 * @copyright Copyright 2003-2023 Zen Cart Development Team
 * @version $Id:  $
 */
```

and something will fill in the `$Id$` bit automatically?  I've always seen people's username, a date and some kind of tag id in there but never known where they come from or how they get updated; I presume it's not manually typed in by someone.

neekfenwick: Due to lack of response, I've copied the copyright header from `search_result/header_php.php` (where I refactored most of the code from to make `class.search.php`) and force pushed my branch, making sure it's rebased onto upstream/master.  I presume the `Id` tags will sort themselves out in the merge.

neekfenwick:
I think I want to refactor the code so the `Search` class instance owns its `$searchOptions` as a member, rather than being passed it as a parameter, i.e. from:
```
class SearchOptions {...};
class Search {
    public function buildSearchSQL($searchOptions) {...}
}

```
to

```
class SearchOptions {...};
class Search {
    protected SearchOptions $searchOptions;
    public function __construc(SearchOptions $searchOptions) { $this->$searchOptions = $searchOptions; }
    public function buildSearchSQL() {...}
}
```

In this way, the Observers reacting to notifier calls can access `$callingClass->searchOptions` without needing to be passed it in the notifier arguments.

Background: I've noticed that the old search system used to have the 'search in description' checkbox which was used within the code to add the match against the `pd.products_description` value.   This `$_GET['search_in_description']` value was not expressed as a local variable or passed into any of the notifiers.

Being a fairly straight port of the code, the `Search` class building the SQL also does not pass this value into the notifiers.  Observers need to know all the search options, such as `search_in_description`, to make informed decisions.  In my case, an addon I'm writing wants to add a match against `products_description` but doesn't know if that search option is enabled or not.

I'll update my PR (again! 😀) @proseLA you've gone a bit quiet, are you not happy with the situation with the `Id` tag or is there some other reason this wasn't merged yet?  It's a good thing, as it turns out, as I'm still tweaking it like this.

neekfenwick:
I think I want to refactor the code so the Search class instance owns its $searchOptions as a member, rather than being passed it as a parameter, i.e. from:

```
class SearchOptions {...};
class Search {
    public function buildSearchSQL($searchOptions) {...}
}
```

to

```
class SearchOptions {...};
class Search {
    protected SearchOptions $searchOptions;
    public function setSearchOptions(SearchOptions $searchOptions)
    {
        $this->searchOptions = $searchOptions;
    }
    public function buildSearchSQL() {...}
}
```

In this way, the Observers reacting to notifier calls can access $callingClass->searchOptions without needing to be passed it in the notifier arguments.  Note that we cannot pass `$searchOptions` into the constructor because `Search` is instantiated in the auto_load sequence.  We could pass it into the `buildSearchSQL()` function and immediate set `$this->searchOptions = $searchOptions` but that feels like an antipattern, better to set it in a setter and then have any functions in `Search` that would need it have it available.

I'm keeping the members of `SearchOptions` as `public` so that notifiers will be able to modify them if need be, which could be useful especially from some early stage notifier hooks that want to modify the search process.  Making them `protected` and adding getters/setters seems like overkill.

Background: I've noticed that the old search system used to have the 'search in description' checkbox which was used within the code to add the match against the pd.products_description value. This $_GET['search_in_description'] value was not expressed as a local variable or passed into any of the notifiers.

Being a fairly straight port of the code, the Search class building the SQL also does not pass this value into the notifiers. Observers need to know all the search options, such as search_in_description, to make informed decisions. In my case, an addon I'm writing wants to add a match against products_description but doesn't know if that search option is enabled or not.

I'll update my PR (again! 😀) @proseLA you've gone a bit quiet, are you not happy with the situation with the Id tag or is there some other reason this wasn't merged yet? It's a good thing, as it turns out, as I'm still tweaking it like this.

proseLA:
> I'll update my PR (again! 😀) @proseLA you've gone a bit quiet, are you not happy with the situation with the Id tag or is there some other reason this wasn't merged yet? It's a good thing, as it turns out, as I'm still tweaking it like this.

please do not mistake my lack of response for something that it is not.  i am following along, and when i have some time and energy, i will (hopefully) download and test the PR a bit more thoroughly.

again, i am a fan of what you are doing, as well as the energy that you bring to this project.  you are a most welcome addition.

with regards to the Id tag thing, please, i am NOT upset with that situation at all.  i was pointing that at, more for your own process, than anything else.  that stamp will get changed once the code is merged and a new release is done.  

as someone else has said, stamps are temporary, blame is forever.

finally, please remember, i am just another coder like you.  i try and support this project, but i am not a maintainer of this project.  i have no merge authority.  my opinion is just another voice in the wilderness...  so to speak...

best.

includes/languages/lang.english.php
```diff
@@ -187,6 +187,7 @@
     'ERROR_INVALID_KEYWORDS' => 'Invalid keywords.',
     'ERROR_INVALID_TO_DATE' => 'Invalid To Date.',
     'ERROR_MAXIMUM_QTY' => 'The quantity added to your cart has been adjusted because of a restriction on maximum you are allowed. See this item:<br>',
+    'ERROR_MISSING_SORT_OPTIONS' => 'Missing sort options',
```
drbyte, line 190: Should 'sort' be 'search' here? Both in the constant name and its defined value?

includes/modules/pages/search_result/header_php.php
drbyte: Yes, the basic copyright header which you copied over is fine. The `@version $Id: $` string gets updated when we run the script to stamp all files shortly before tagging a release. Until that stamping it's just the same as the previous stamping.

drbyte:
I like the approach you're taking here.
And I think the embedding of the options makes sense.

includes/languages/lang.english.php
neekfenwick: Whoops, you're right. Will fix.

neekfenwick:
> please do not mistake my lack of response for something that it is not. i am following along, and when i have some time and energy, i will (hopefully) download and test the PR a bit more thoroughly.

Cheers :) I'm doing all this alongside my day job and it can be easy to go down a psyche rabbit hole sometimes. I'm working on a Soundex Search addon that is a good testing bed for these `Search` changes, but it can be hard to get the time to work through it.  I have a bunch more critical path stuff to deal with this week and will go over the Search changes when possible.

> as someone else has said, stamps are temporary, blame is forever.

Heh :)

brittainmark:
Can anyone forsee a reason why the options would be needed outside the class? If so should a  public function getSearchOptions() be added?

neekfenwick:
> Can anyone forsee a reason why the options would be needed outside the class? If so should a public function getSearchOptions() be added?

Yes, from Observers who want to access the search options.  I can certainly add a public getter for it.

Looks like my mistake - `Search::searchOptions` is currently `protected`, yet as I said in earlier comment, I access `$callingClass->searchOptions->search_in_description` from my Observer which I now notice has been failing silently with a Warning logged.

drbyte:
Looks great!  Thanks for all your work on this!

PR #6038, opened 2023-11-02
Set a default locale in case the server's default is invalid or unconfigured

### LINKED ISSUE #6010: PHP Fatal error:  Uncaught IntlException: datefmt_create: invalid locale: U_ILLEGAL_ARGUMENT_ERROR
EDIT: Affects PHP 8.1.25 and 8.2.12 and 8.3-RC3/RC4/RC5+

---

Windows php 8.3.0 RC3/4 (RC2 is ok)

Here is admin login:

[11-Oct-2023 11:20:00 UTC] PHP Fatal error:  Uncaught IntlException: datefmt_create: invalid locale: U_ILLEGAL_ARGUMENT_ERROR in D:\GitHub\zencart\includes\classes\zcDate.php:87
Stack trace:
#0 D:\GitHub\zencart\includes\classes\zcDate.php(87): IntlDateFormatter->__construct('C', 3, -1)
#1 D:\GitHub\zencart\includes\classes\zcDate.php(39): zcDate->initializeConversionArrays()
#2 D:\GitHub\zencart\includes\autoload_func.php(47): zcDate->__construct()
#3 D:\GitHub\zencart\admin\includes\application_top.php(42): require('D:\\GitHub\\zenca...')
#4 D:\GitHub\zencart\admin\home.php(9): require('D:\\GitHub\\zenca...')
#5 D:\GitHub\zencart\admin\index.php(11): require('D:\\GitHub\\zenca...')
#6 {main}
  thrown in D:\GitHub\zencart\includes\classes\zcDate.php on line 87

drbyte:
Required since PHP 8.3.0 RC5 and expected in all 8.3+ versions. Also required in PHP 8.1.25 and 8.2.12 (but hopefully not in 8.1.26+ and 8.2.13+)

If not configured at the server/system level, then it defaults to `'C'` (un'C'onfigured), which is invalid when IntlDateFormatter is called, which now throws an exception in such cases.

Fixes #6010

### DIFF
--- a/admin/includes/application_bootstrap.php
+++ b/admin/includes/application_bootstrap.php
@@ -37,7 +37,16 @@ $_SERVER['SCRIPT_NAME'] = str_replace($serverScript, '', $_SERVER['SCRIPT_NAME']
 /*
  * Get time zone info from PHP config
 */
-@date_default_timezone_set(date_default_timezone_get());
+date_default_timezone_set(date_default_timezone_get());
+
+/*
+ * Check for a valid system locale, and override if invalid or set to 'C' which means 'unconfigured'
+ * It will be overridden later via language-selection operations anyway, but a valid default must be set for zcDate class methods to work
+ */
+$detected_locale = setlocale(LC_TIME, 0);
+if ($detected_locale === false || $detected_locale === 'C') {
+    setlocale(LC_TIME, ['en_US', 'en_US.UTF-8', 'en-US', 'en']);
+}
 
 if (!defined('DIR_FS_ADMIN')) define('DIR_FS_ADMIN', preg_replace('#/includes/$#', '/', realpath(__DIR__ . '/../') . '/'));
 
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -112,7 +112,17 @@ if (DEBUG_AUTOLOAD || (defined('STRICT_ERROR_REPORTING') && STRICT_ERROR_REPORTI
     error_reporting(0);
 }
 
-@date_default_timezone_set(date_default_timezone_get());
+date_default_timezone_set(date_default_timezone_get());
+
+/*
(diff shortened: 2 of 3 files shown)


torvista:
Unfortunately I don't find a single locale valid in both Windows and Unix, so you'll need
`setlocale(LC_TIME, ['en_US', 'en_US.UTF-8', 'en-US']);`

While I'm not requiring an explanation, it seems neater to put this single line in set_time_zone, and an easier fix for those all those site that are titsup now...

drbyte:
It's unfortunate that PHP isn't (yet) choosing to discern its own proper fallback.

drbyte:
Perhaps our implementation of zcDate which calls the affected IntlDateFormatter needs to just fail-silently (like PHP 'used to do').

drbyte:
> Unfortunately I don't find a single locale valid in both Windows and Unix, so you'll need `setlocale(LC_TIME, ['en_US', 'en_US.UTF-8', 'en-US']);`

Does that mean your local Windows setup has always required manually adding `en-US` to the language files?  Or was `'en'` actually working as a valid fallback?
(PHP iterates through the presented values, choosing the first valid one it finds from the provided list/array.)

torvista:
>Does that mean your local Windows setup has always required manually adding en-US to the language files

Yes, or a GB/ES equivalent: I had several in there to cover both OS, or so I thought until this cropped up.

'en' also works, but I just quoted en-US to match the others.

The tool I mentioned elsewhere when run locally and remotely makes it clear which locales are valid on which servers and there are none that coincide with both OS.

drbyte:
I'll add `'en'` back into the list, along with '`en-US'`. Gotta luv Windows.

drbyte:
Looks like the PHP team are finally reinstating support for the default `'C'` locale in PHP 8.2.15, which will hopefully also roll into 8.3.x

https://github.com /php/php-src/commit/6a447e7437b988062fda6529cc84c6595de09b54

PR #6047, opened 2023-11-09
Correcting auto-login on "Place Order"

### LINKED ISSUE #6046: hmac no longer validating
i am not sure if #6033 was properly tested.

when i now try to place an order as an admin, the method `zen_validate_hmac_login` returns false.

has anyone been able to get this to return true?

it seems that `'hmac', 'timestamp'` might be needed to be added to the array here:

https://github.com/zencart/zencart/blob/da0a8c6780890718480b77819b7cca129217a2ac/includes/functions/functions_customers.php#L296

but i am still not getting the method to validate.

lat9:
Fixes #6046

- The admin customer tool wasn't including the 'timestamp' in the generated hmac.
- For whatever reason, browsers update a posted 'timestamp' hidden field when there's also an `id=` parameter (go figure).
- The check in `zen_validate_hmac_login` needs to check its generated HMAC against that passed from the admin and ensure that the 'timestamp' value was posted.
- Includes a couple of refactorings, using multi-variable `isset` checks instead of looping.

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -1933,7 +1933,8 @@ if ($action === 'edit' || $action === 'update') {
                         $hmacpostdata = [
                             'cid' => $cInfo->customers_id,
                             'aid' => $_SESSION['admin_id'],
-                            'email_address' => $cInfo->customers_email_address
+                            'email_address' => $cInfo->customers_email_address,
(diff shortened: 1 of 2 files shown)


proseLA:
in my brief testing, this change works.  thank you @lat9!

PR #6048, opened 2023-11-10
Use AJAX to validate datepicker-entered dates

### LINKED ISSUE #6043: Featured, Specials and Banners: Directly setting end-date in the past results in a never-ending time span
In the admin's **_Catalog :: Specials_**, **_Catalog :: Featured Products_** and **_Tools :: Banner Manager_**, hand-entering (rather than using the date-picker) a date in the past results in a never-ending time span.

lat9:
Fixes #6043


proseLA:
in briefly looking at this code, it seems there is a ripe opportunity for a zen cart function.  it seems the following section of code is used 6 times in this PR.  why not do a function?

https://github.com/zencart/zencart/blob/fc593ca40fc11bf8ec9c579c0bcc6dbbede8a3f7/admin/banner_manager.php#L86-L104

also, after line 90, your indents have moved to 2 spaces as opposed to 4.

lat9:
> in briefly looking at this code, it seems there is a ripe opportunity for a zen cart function. it seems the following section of code is used 6 times in this PR. why not do a function?
> 
> https://github.com/zencart/zencart/blob/fc593ca40fc11bf8ec9c579c0bcc6dbbede8a3f7/admin/banner_manager.php#L86-L104
> 

Because it was enough of a pain to get the various parts working.  Good suggestion for a follow-on change.

> also, after line 90, your indents have moved to 2 spaces as opposed to 4.

For the most part, because I was trying real hard (my fingers were twitching) not to reformat the entire script!

lat9:
> For the most part, because I was trying real hard (my fingers were twitching) not to reformat the entire script!

Noting that anything I do to 'correct' the indentation in my additions is just going to make that section look more wonky than it already does.  Once this PR is merged, then the reformatting and refactering of these scripts can be a next step.

PR #6052, opened 2023-11-15
set timeouts for curl connection to version server

zcwilt:


admin/includes/classes/VersionServer.php
```diff
@@ -10,6 +10,10 @@ class VersionServer
     protected $projectVersionServer = 'https://ping.zen-cart.com/zcversioncheck';
     protected $pluginVersionServer = 'https://ping.zen-cart.com/plugincheck';
 
+    const TIMEOUT = 3;
+
+    const CONNECTTIMEOUT = 1;
+
```
lat9, line 16: Constant visibility should be provided for PHP versions 7.1+, per PSR-12.

PR #6061, opened 2023-11-22
Improve links between orders and customers screens.

### LINKED ISSUE #6054: Links between admin customer and order page for easier navigation.
The lack of hyperlinks from side panels in Admin give me grief.

When viewing an Order, you see the Customer's email address in the side panel but no easy way to navigate to their customer record (for example, to place another order as them).
When viewing a Customer, you see a number of orders, but no easy way to get to those orders.

- Customer email address viewed next to an order ought to be hyperlinked to the Customers page with that record selected.
- Order details viewed next to a Customer ought to list the last few orders (say, 5), in date order, hyperlinked to the order details page for each order.

I'm going to add these to my local shop just to ease things.

Anyone want to add anything to that?

neekfenwick:
Fixes #6054 - feel free to critique.

The more elements are turned into FontAwesome icons, the more the graphic icons such as `icon_edit.gif` and `icon_info.gif` on the orders page stand out.  I'd be happy to take a stab at converting them if there is general support from the core team to do so.

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -651,22 +651,25 @@ class Customer extends base
      * @param integer $limit Optional limit of the number of orders to return.
      * @return array
      */
-    public function getOrdersSummary(int $limit = 0): array {
+    public function getRecentOrderRecords(int $limit = 0): array 
+    {
         global $db;
         if (empty($this->customer_id)) {
             return [];
         }
-        $rs = $db->Execute("SELECT orders_id, date_purchased, orders_status_name
+        
+        $result = $db->Execute("SELECT orders_id, date_purchased, orders_status_name
             FROM " . TABLE_ORDERS . " o
             INNER JOIN " . TABLE_ORDERS_STATUS . " os
                 ON o.orders_status = os.orders_status_id AND os.language_id = {$_SESSION['languages_id']}
             WHERE customers_id = {$this->customer_id}
             ORDER BY date_purchased DESC" . (!empty($limit) ? " LIMIT $limit" : '') . ';');
-        $result = [];
-        foreach ($rs as $row) {
-            $result[] = $row;
+        
+        $orders = [];
+        foreach ($result as $row) {
+            $orders[] = $row;
         }
-        return $result;
+        return $orders;
     }
 
     public function setPassword(string $new_password)
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -645,6 +645,30 @@ class Customer extends base
         return $result->fields['total'];
     }
 
+    /**
+     * Return a simple summary of orders for this customer, ordered most recent first.
+     *
+     * @param integer $limit Optional limit of the number of orders to return.
+     * @return array
+     */
+    public function getOrdersSummary(int $limit = 0): array {
+        global $db;
+        if (empty($this->customer_id)) {
+            return [];
+        }
+        $rs = $db->Execute("SELECT orders_id, date_purchased, orders_status_name
+            FROM " . TABLE_ORDERS . " o
+            INNER JOIN " . TABLE_ORDERS_STATUS . " os
+                ON o.orders_status = os.orders_status_id AND os.language_id = {$_SESSION['languages_id']}
+            WHERE customers_id = {$this->customer_id}
+            ORDER BY date_purchased DESC" . (!empty($limit) ? " LIMIT $limit" : '') . ';');
+        $result = [];
+        foreach ($rs as $row) {
+            $result[] = $row;
+        }
+        return $result;
+    }
+
     public function setPassword(string $new_password)
     {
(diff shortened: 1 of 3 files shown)


neekfenwick:
I realised I'd missed some code when copying from my dev workspace, so I force pushed a squash merge of the fix.

includes/classes/Customer.php
```diff
@@ -645,6 +645,30 @@ public function getNumberOfOrders(): int
         return $result->fields['total'];
     }
 
+    /**
+     * Return a simple summary of orders for this customer, ordered most recent first.
+     *
+     * @param integer $limit Optional limit of the number of orders to return.
+     * @return array
+     */
+    public function getOrdersSummary(int $limit = 0): array {
```
drbyte, line 654: Given that `getOrderHistory()` already exists, I wonder if this is needed. Sure, that other function checks more tables and brings back more data, so might be slightly slower. 

Perhaps if this function is kept, naming it `getRecentOrderRecords()` might be slightly clearer about its intent?

includes/classes/Customer.php
```diff
@@ -645,6 +645,30 @@ public function getNumberOfOrders(): int
...
+        global $db;
+        if (empty($this->customer_id)) {
+            return [];
+        }
+        $rs = $db->Execute("SELECT orders_id, date_purchased, orders_status_name
+            FROM " . TABLE_ORDERS . " o
+            INNER JOIN " . TABLE_ORDERS_STATUS . " os
+                ON o.orders_status = os.orders_status_id AND os.language_id = {$_SESSION['languages_id']}
+            WHERE customers_id = {$this->customer_id}
+            ORDER BY date_purchased DESC" . (!empty($limit) ? " LIMIT $limit" : '') . ';');
```
drbyte, line 664: The `$limit` can be passed as the second param to `$db->Execute($sql, $limit);` where it is also checked for validity.

includes/classes/Customer.php
```diff
@@ -645,6 +645,30 @@ public function getNumberOfOrders(): int
...
+        $result = [];
+        foreach ($rs as $row) {
+            $result[] = $row;
+        }
+        return $result;
```
drbyte, line 669: Since this is returning an array of db records, there's no need to foreach over it. You can foreach over the results of the db query directly, so could return that directly.

Also, and sorry if this is super nitpicky: 
Instead of `$rs` for a variable, in Zen Cart we commonly use `$result`, since it is the query's result.
And then if a temporary value is needed for parsing results to be returned, commonly you'll see `$retVal` or `$response` used.

includes/classes/Customer.php
drbyte: Tiny point on formatting: while we used to keep the opening `{` for a function/method on the same line in years past, in recent years we're putting that opening brace on a new line when it's for a function declaration or class declaration. (We still keep it inline for `for`, `foreach`, `if` etc).

drbyte (commented):
A few thoughts and observations posted inline.

drbyte:
> The more elements are turned into FontAwesome icons, the more the graphic icons such as icon_edit.gif and icon_info.gif on the orders page stand out. I'd be happy to take a stab at converting them if there is general support from the core team to do so.


Fair point. Best as a separate PR.

Probably good to keep the ALT text rendered with it, probably as a `title=` tag. I'd wanna check what we did in the admin product-listing for that.

includes/classes/Customer.php
neekfenwick: You're right, I was scared off using `getOrderHistory()` because it looked like too much was going on, and it constructed a `splitPageResult`.  Taking a new look, I have optimised the query so the product count is done within the initial query, but I ran into trouble with the `splitPageResult` constructor in `admin` being different to the store front, i.e. https://github.com/zencart/zencart/blob/1da3aa3efc5fbdce79333c06b10bc738c223e230/admin/includes/classes/split_page_results.php#L92 vs https://github.com/zencart/zencart/blob/1da3aa3efc5fbdce79333c06b10bc738c223e230/includes/classes/split_page_results.php#L53

What's the deal with this code drift?  We cannot use all `Customer` methods from admin because of it.

On the formatting, that's cool.  I was afraid to run `cs-fixer` since it would probably create hundreds of modifications :)

includes/classes/Customer.php
neekfenwick: My aim was to avoid returning a database object (`mysqli_result`? not sure) from a data accessor, figured a plain array would be more standard.  Looks like I will squash this new method anyway, though, if I can get `getOrderHistory()` to play nice with `splitPageResult`. If not, will fix this as you suggest.

neekfenwick:
Thanks @drbyte - I see you've merged something a commit or two behind my intention, probably my fault.  Please see new PR #6081 which you can approve or chuck out :)

PR #6069, opened 2023-11-30
tax address entries reversed

zcwilt:
... introduced in #5297


dbltoe:
FYI
This did not fix the tax problem at https://www.zen-cart.com/showthread.php?229793-Incorrect-processing-of-Tax&p=1397915#post1397915
I am getting a consistent showing of item with tax now but, when going from checkout 3 back to the shopping cart, the tax amount drops off the product price.

zcwilt:
@dbltoe yes.
I found this bug while looking at your issue and wanted to fix it quickly 

your bug has a more complicated fix and I'll post more in your thread about this

PR #6073, opened 2023-12-04
find zone id from state where addressbook is broken

zcwilt:
Note from @scottcwilson - please note this is a confirmed bug found during the investigation of #6008.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -583,6 +583,7 @@ class order extends base
                 'telephone' => $customer_address->fields['customers_telephone'],
                 'email_address' => $customer_address->fields['customers_email_address'],
             ];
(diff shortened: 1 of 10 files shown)


zcwilt:
Note: this seems to work where there shipping tax is not set to store.
needs more work to cover that

zcwilt:
updated to fix shipping tax where zone id = 0 
Still needs a few more tests to cover tax basis = store

dbltoe:
As stated on the forum, we do not use a dropdown for state in the registration form.
Also, there are no entry_zone_id entries that are 0 except where the country has no states.  Then, the 0 is as it should be, correct?

zcwilt:
@dbltoe On A clean install of Zencart, if I create a customer account in store then it doesn't set the address book entry_zone_id, regardless of whether zones exist or not.

zcwilt:
@dbltoe as per my post on the thread. Contact me directly at ian@zen-cart.com to see if I can help more directly with your problem

PR #6081, opened 2023-12-11
Use Customer::getOrderHistory for recent order info.

neekfenwick:
This follows on from discussion on #6061 

After our discussion in your review comments, I agreed we could get rid of my new function `Customer::getRecentOrderRecords()`, and I re-wrote the original function `Customer::getOrderHistory()` so it can be used instead.  However I have hit a problem in its use of `splitPageResults` because that function does some basic string searching in the SQL for things like 'from' and splits the query so it can count rows, which fails when a subselect is used (anything with a 'from' or 'group by' clause, it seems).  The result is malformed SQL, probably an issue you've been long aware of with `splitPageResults`.

(Note, the only place I see `account_history` being linked to, and thus using this splitPageResults path, is from a "(show all orders)" link on the customer `account` page.)

So, I cannot optimise that function as I wanted to with a subselect, but I can at least remove my mostly unnecessary `getRecentOrderRecords()` function.

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -2085,7 +2085,7 @@ if ($action === 'edit' || $action === 'update') {
                     $cInfo->number_of_orders;
                 if ($cInfo->number_of_orders > 0) {
                     $text .= ' [ ';
-                    foreach ($customer->getRecentOrderRecords(5) as $order) {
+                    foreach ($customer->getOrderHistory(5) as $order) {
                         $text .= '<a href="' . zen_href_link(
                             FILENAME_ORDERS,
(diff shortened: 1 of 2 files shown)


drbyte:
Cool. Looks good.

Ya I agree with the challenges of SplitPageResults. There are many dozens of plugins and customized code that rely on the Admin parameters being one way and the Catalog parameters another way. So it's a non-starter for consolidating them.
Only option is a full new replacement that we refactor to, while keeping both the old ones in-place for compatibility.
Aside: The Laravel Eloquent infrastructure may actually be the best way to go there, since it already supports pagination, aggregating and querybuilding etc.

PR #6101, opened 2023-12-28
paypal_ipn.core.php: short array syntax, formatting

torvista:


proseLA:
[per the docs,](https://docs.zen-cart.com/dev/contributing/coding_standards/#arrays) arrays should have a trailing comma.

torvista:
In this specific instance, should not an override be used/there should be no edits to this file?

proseLA:
> In this specific instance, should not an override be used/there should be no edits to this file?

no.

lat9:
> In this specific instance, should not an override be used/there should be no edits to this file?

I don't understand the question.  I'll note that I'm not a big fan of using the /auto_loaders/overrides directory, if that's what you're referring to.

torvista:
I understand adding a trailing comma to the end of arrays facilitates modification of a core file and subsequent comparison.
But if there is a built-in mechanism for overriding that core file and avoiding modification (as in this case) is that not preferred?

lat9:
>.... But if there is a built-in mechanism for overriding that core file and avoiding modification (as in this case) is that not preferred?

I'm unclear as to how you think that `config.core.php` could be overridden in this case, but why would a core-file override a core-file?

torvista:
er...since there is an overrides directory, is that not meant for customised versions of those files?
image

lat9:
The /auto_loaders/overrides specifically overrides a file in the base /auto_loaders directory of the same name.

torvista:
I'm not getting why we appear to be at cross-purposes here:
2023-12-29 13_55_44-zencart  D__GitHub_zencart
This overrides that, so that does not need to be modified, no?

torvista:
I don't.
I just was pointing out that adding commas to a core file to allow easier modification seemed contradictory to the principal of always using an override where possible, precisely to avoid core file modification.
Much ado about nothing. I'll add the commas, and get my coat (british humor).

drbyte:
To clarify: the purpose of trailing-commas on the end of multi-line array declarations is to make git diffs easier to read, and (formerly to avoid potential syntax errors):

Say for example we may have had this declaration, without the trailing comma:
```php
$autoLoadConfig[100][] = [
    'autoType' => 'classInstantiate',
    'className' => 'template_func'
];
```
But then we wanted to add another element to the array ... we'd have to add the trailing comma on the preceding line, even though nothing else was changed on that line:
```diff
$autoLoadConfig[100][] = [
    'autoType' => 'classInstantiate',
-    'className' => 'template_func'
+    'className' => 'template_func',
+    'objectName' => 'template'
];
```

So, if we just always include the trailing comma, things are faster to review and comprehend:
```diff
$autoLoadConfig[100][] = [
    'autoType' => 'classInstantiate',
    'className' => 'template_func',
+    'objectName' => 'template',
];
```

Note: Unfortunately other languages like (most flavors of) javascript aren't able to handle the trailing comma after the last array element, so JS will throw a syntax error in such cases.

PR #6104, opened 2023-12-29
Adding Crowdin config file

syntaxerror:
Used for syncing English core language files from GitHub to Crowdin.

Note! Files in html_includes are added as txt files in Crowdin to make them translatable, since they isn't real language files with defines. On export they will be automatically renamed back to php.

_*I recommend moving flag icons out of language packs, and rather incorporate them in the core by default like other systems do._

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,6 +1,6 @@
-export_languages: ["english"]
+export_languages:
+  - english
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks!

PR #6108, opened 2023-12-30
headless api

zcwilt:


drbyte:
Was this intentional?
Screen Shot 2024-01-19 at 8 22 38 PM

zcwilt:
yes.  previous code only really used the Laravel/capsule database stuff.

This pulls in the whole Laravel backend to allow us to leverage routes/views/filament etc

drbyte:
Should we upgrade it to v9 again?

zcwilt:
@drbyte yeah we could do
only requires php8.0 and the upgrade guide seems ok 
I'll look to do this once I've finished some work on the cart classes, which is a dependancy for this

PR #6110, opened 2023-12-31
Require PHP 8.0 minimum

### LINKED ISSUE #6095: named arguments not allowed in php 7.x
#6066 added new functions.  a number of calls to the functions make use of named arguments.  named arguments were introduced in php 8.

[the docs need to be updated](https://docs.zen-cart.com/user/first_steps/server_requirements/#:~:text=Zen%20Cart%20v2.0.0,required%20for%20zc_install) or else the named calls need to be refactored.

one such example:
https://github.com/zencart/zencart/blob/a28f1e082c380590a013f50bfcc06dbf26519d7c/admin/orders.php#L1394

i would weigh in on this bug... but i'll just stick with reporting it...

drbyte:
Closes #6095


.github/workflows/zc_functional_test_suite.yml
```diff
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        php-version: ['8.0']
+        php-version: ['8.0', '8.3']
```
drbyte, line 11: Note: Here we are only testing "oldest supported" and "newest supported".

zc_install/index.php
```diff
@@ -9,7 +9,7 @@
   // Actual version check is more strict; this is just to start the program
   // For true minimum, see includes/systemChecks.yml under checkPhpVersionMin
   if (PHP_VERSION_ID < 70100) { // only checking for 7.1 here as a basic check. zc_install itself doesn't need higher than this at this time
```
drbyte, line 11: Note: This check in `zc_install/index.php` is merely for zc_install to "run".  We don't require super-modern PHP versions here in zc_install, which is good, so that the usual zc_install experience can still operate friendly until the PHP version inspection is run, at which point that and any other errors are displayed.

PR #6111, opened 2023-12-31
Consolidate zen_set_field_length duplicate functions

drbyte:


scottcwilson:
This looks great (surprised we missed this when we consolidated functions last time), but can you add comments explaining the derivation of 50 and 70 in includes/functions/functions_general_shared.php?  I agree this code is backwards compatible, I just don't know how those numbers were deduced.

drbyte:
>  I just don't know how those numbers were deduced.

Neither do I. After researching a bit, it appears those have been the defaults since version 1.1.0

scottcwilson:
LOL ok.

drbyte:
It probably should be standardized to one value in both places.
I can see how we may want to make sure the Admin doesn't enter values that are too long for display fields.
And I can see how we want to let customers enter as much as needed, but no more.

The value is only used when there's a database field length that's longer, in which case we use the shorter of the two values.

Requires a study of all places where the function is used, making a list of the cases where the db length is longer than the default, and "what" the value is being used for in those cases. And then decide which of the defaults to settle on, noting what the impact will be on customer and admin experience when we do so.

PR #6114, opened 2024-01-01
featured, header: simplify logic, strict checking

torvista:
Also this looks superfluous...could use $featured_products_split directly?
`$check_products_all = $db->Execute($featured_products_split->sql_query);`


torvista:
why bother.

drbyte:
Actually, I've spent several hours this week looking at this. 

I didn't merge it yet because the featured_products and products_all and products_new pages are near-clones of each other, and touching only this "featured" one would leave them out of sync. 

I was going to do those other two pages and add them to this PR, but I also know how much everyone loves to complain that these "cloned pages" exist ... and since that's such a hotly hated topic, I share your sentiment: "why bother".

But, despite the anticipated dissatisfaction with whatever is proposed, I'm hoping to present something soon.

PR #6124, opened 2024-01-06
Reformat zc_install code

drbyte:
Apply PSR-12 formatting, strict typing, and typehints.

Caveat: this makes zc_install no longer compatible with PHP 7, so even just to install and inspect the server requires PHP 8

### DIFF
--- a/zc_install/ajaxAdminSetup.php
+++ b/zc_install/ajaxAdminSetup.php
@@ -1,4 +1,6 @@
 <?php
+declare(strict_types=1);
+
 /**
  * ajaxAdminSetup.php
  * @copyright Copyright 2003-2022 Zen Cart Development Team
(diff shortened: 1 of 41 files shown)


brittainmark:
when can curlGetUrl return bool. it always sets 3 array values in $header which it returns
```
    public static function curlGetUrl($url, $follow_redirects = false): array|bool
    {
        $options = [
            CURLOPT_RETURNTRANSFER => true, // return web page
            CURLOPT_HEADER => false,        // don't return headers
            CURLOPT_FOLLOWLOCATION => $follow_redirects,    // follow redirects
            CURLOPT_ENCODING => "",         // handle all encodings
            CURLOPT_AUTOREFERER => true,    // set referer on redirect
            CURLOPT_CONNECTTIMEOUT => 3,    // timeout on connect
            CURLOPT_TIMEOUT => 3,           // timeout on response
            CURLOPT_MAXREDIRS => 10,        // stop after 10 redirects
        ];

        $ch = curl_init($url);
        curl_setopt_array($ch, $options);
        $content = curl_exec($ch);
        $err = curl_errno($ch);
        $errmsg = curl_error($ch);
        $header = curl_getinfo($ch);
        curl_close($ch);

        $header['errno'] = $err;
        $header['errmsg'] = $errmsg;
        $header['content'] = $content;
        return $header;
    }
```

brittainmark:
does` public function validateAdminCredentials($adminUser, $adminPassword): bool|string|int`
need string as it returns `false` or `$result->fields['admin_id']` which is defined as integer on the database.

brittainmark:
zc_install/includes/classes/class.zcConfigureFileWriter.php
` protected function processAllConfigureFiles($adminDir): int|string`
I do not think this can return a string as it returns a bit wise comparison which is an integer.

brittainmark:
zc_install/includes/classes/class.zcDatabaseInstaller.php

 ```
public function updateConfigKeys(): bool
    {
        if (isset($_POST['http_server_catalog']) && $_POST['http_server_catalog'] != '') {
            $email_stub = preg_replace('~.*\/\/(www.)*~', 'YOUR_EMAIL@', $_POST['http_server_catalog']);
            $sql = "UPDATE " . $this->dbPrefix . "configuration SET configuration_value=:emailStub: WHERE configuration_key IN ('STORE_OWNER_EMAIL_ADDRESS', 'EMAIL_FROM')";
            $sql = $this->db->bindVars($sql, ':emailStub:', $email_stub, 'string');
            $this->db->Execute($sql);
        }
        return false;
    }
```
This function never fails, is this correct?

**Comment by brittainmark** [comment]
zc_install/includes/functions/password_funcs.php
`function zen_pwd_rand(int $min = 0, int $max = 10): float|int`
does not need to be float as we are forcing int before return
 ```
$value = abs((int)$value);
return $value;
```

drbyte:
> does` public function validateAdminCredentials($adminUser, $adminPassword): bool|string|int` need string as it returns `false` or `$result->fields['admin_id']` which is defined as integer on the database.

Everything comes back from the database as a string. QueryFactoryResult doesn't do any typecasting.

drbyte:
Thanks @brittainmark ; updates pushed.

PR #6127, opened 2024-01-07
Enable 'retirement' of paypaldp/paypalwpp

### LINKED ISSUE #6126: Payment Methods: Adding switch to enable admin handling only
### Discussed in https://github.com/zencart/zencart/discussions/6088

<sup>Originally posted by **lat9** December 15, 2023</sup>
If a site's currently using payment-module `A`, but decides to switch to payment-module `B` ... payment-module `A` is disabled so that new orders can't be placed.  Unfortunately, that disablement results in payment-module `A` being disabled both storefront and admin.

This could pose problems for a store if they've still got orders that need to be either captured, refunded or voided that were placed using payment-module `A`.

Perhaps payment-modules should provide a tri-state "Enable this" setting to disable both storefront/admin, enable both storefront/admin or disable storefront-only?</div>

lat9:
Fixes #6126

### DIFF
--- a/includes/modules/payment/paypalwpp.php
+++ b/includes/modules/payment/paypalwpp.php
@@ -151,7 +151,7 @@ class paypalwpp extends base {
     $this->code = 'paypalwpp';
     $this->codeTitle = MODULE_PAYMENT_PAYPALWPP_TEXT_ADMIN_TITLE_EC;
     $this->codeVersion = '1.5.8';
-    $this->enabled = (defined('MODULE_PAYMENT_PAYPALWPP_STATUS') && MODULE_PAYMENT_PAYPALWPP_STATUS == 'True');
+
     // Set the title & description text based on the mode we're in ... EC vs US/UK vs admin
     if (IS_ADMIN_FLAG === true) {
       $this->description = sprintf(MODULE_PAYMENT_PAYPALWPP_TEXT_ADMIN_DESCRIPTION, ' (rev' . $this->codeVersion . ')');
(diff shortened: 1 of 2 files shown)


scottcwilson:
Does it make sense to wrap the "Retired" update in a check so we don't do it every time?

lat9:
No.  In order to see if the "Retired" option has been applied, you'd need to perform a database query.  It's simpler to just perform the update, which would need to be done anyway.

lat9:
... although it makes sense to limit that forced update to the admin's **_Modules :: Payment_** page; I'll get those updated.

includes/modules/payment/paypalwpp.php
```diff
@@ -177,6 +177,8 @@ function __construct() {
 
       if (null === $this->sort_order) return false;
 
+      $this->enabled = (MODULE_PAYMENT_PAYPALWPP_STATUS === 'True' || (IS_ADMIN_FLAG === true && MODULE_PAYMENT_PAYPALWPP_STATUS === 'Retired'));
```
torvista, line 180: $this->enabled is now inside the IS_ADMIN clause: this payment option will never show on the shopfront.

lat9: True, that.  I'll get this updated later today.

lat9: Sigh, the `paypaldp` has issues as well.  Both modules have some additional checks if enabled to set their title/description.

drbyte: @lat9 wouldn't just moving that $this->enabled check back to its original line be enough? (er, well, probably also restore the original `defined()` check as well).

I'm not seeing why it needed to be relocated.

lat9: That's me over-thinking again; you're right.

PR #6137, opened 2024-01-13
feature/installer_progress_updates

zcwilt:


zcwilt:
This covers off a couple of things

It adds a progress bar when upgrade happens as we originally only had progress bar on a clean install.

It also allows adding tags with in the install/upgrade sql files. 

These are their for sql upgrades that may take a significant amount of time. 

e.g we have thousands of orders and need to alter the order table. 

prompts can be added to the sql files using  a line like 

#PROGRESS_FEEDBACK:!TEXT=Altering Product table - may take some time

currently have only changed the v200 upgrade file and the main install sql file

zcwilt:
note also 

Probably needs some UI love and there is also a bit of duplication in the updateStatus code. 

Didn't want to spend too much time on this as have some more ideas around rebuilding the installer later on

drbyte:
Example screenshots:
Screen Shot 2024-01-19 at 8 27 46 PM
Screen Shot 2024-01-19 at 8 28 22 PM
Screen Shot 2024-01-19 at 8 30 28 PM

drbyte (approved):

scottcwilson:
Very nice improvements!

PR #6145, opened 2024-01-18
Correct tpl_ajax_checkout_confirmation_default.php Warnings

lat9:
As reported in [this](https://www.zen-cart.com/showthread.php?229896-PHP-Notice-ob_flush()-Failed-to-flush-buffer-zcAjaxPayment-php&p=1398629#post1398629) Zen Cart forum bug report.

### DIFF
--- a/includes/templates/responsive_classic/templates/tpl_ajax_checkout_confirmation_default.php
+++ b/includes/templates/responsive_classic/templates/tpl_ajax_checkout_confirmation_default.php
@@ -123,7 +123,7 @@
         <tr class="<?php echo $order->products[$i]['rowClass']; ?>">
           <td  class="cartQuantity"><?php echo $order->products[$i]['qty']; ?>&nbsp;x</td>
           <td class="cartProductDisplay"><?php echo $order->products[$i]['name']; ?>
-          <?php  echo $stock_check[$i]; ?>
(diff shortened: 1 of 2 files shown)


proseLA:
considering the master branch now requires php 8.0, should we not be using the null coalescing operator?

drbyte:
> Unfortunately the value can be either `null`, an empty string (`''`) or a non-empty string, precluding the use of the `??` operator.

Hmmm ... but ... if it's an empty string, it's okay to echo it out, right?

So, `??` actually does work.

```php
$stock_check = [];
$stock_check[] = null;
$stock_check[] = '';
$stock_check[] = 'foo';

foreach ($stock_check as $val) {
  echo var_export($val, true);
  echo ' becomes: ';
  echo var_export(!empty($val) ? $val :  '', true);
  echo ' or ';
  echo var_export($val ?? '', true);
  echo "\n";
}
```
Output:
```
NULL becomes: '' or ''
'' becomes: '' or ''
'foo' becomes: 'foo' or 'foo'
```

drbyte:
Oops. I didn't test for `0`. Using null-coalesce instead of empty() will produce a different result.

```
0 becomes: '' or 0
'0' becomes: '' or '0'
2 becomes: 2 or 2
'2' becomes: '2' or '2'
```

PR #6154, opened 2024-01-25
Shared product listing template and sort for All/New/Featured/Special pages

drbyte:
- S/N/A/F pages now use the same product-listing template as the category navigation does
- all these pages will also follow the setting for "columns per row", so if the product-listing uses rows, so will S/N/A/F page; likewise, if set to 3 columns, all pages will use 3 columns. (Example in screenshots below.)
- the "click column headings for sort" is replaced by the sort dropdown
- pagination is not changed

(ignore the "red" theme: that was just a way to remind myself of whether I was looking at "old" vs "new" layouts)
laptop-feat-grid
laptop-feat
laptop-linked
mobile-feat-grid
mobile-feat
mobile-linked
specials-grid
tablet-feat-grid
tablet-feat
tablet-linked

Closes #6107


drbyte:
There are a couple styling things to adjust yet.

drbyte:
Haven't decided whether to fully delete all the admin configuration switches that controlled all the per-page field placements. Doing that would certainly make the admin configuration menu cleaner.

lat9:
> Haven't decided whether to fully delete all the admin configuration switches that controlled all the per-page field placements. Doing that would certainly make the admin configuration menu cleaner.

Noting that some sites want all the featured products and specials displayed, but limit the pagination for all-products.

scottcwilson:
> Haven't decided whether to fully delete all the admin configuration switches that controlled all the per-page field placements. Doing that would certainly make the admin configuration menu cleaner.

This would also greatly impact the development of templates, which would have to be aware of the before and after behavior.

barco57:
...and templates would have to add their own configuration for these if they are removed...cause Chris likes less configuration.
I am not opposed to their removal eventually as I have discussed with lat9 that I would like to see per template configuration page(s) managed in a different spot where all of the current layout configs could go.

drbyte:
> Haven't decided whether to fully delete all the admin configuration switches that controlled all the per-page field placements. Doing that would certainly make the admin configuration menu cleaner.

I was referring to these kinds of settings, which do not control the Product Listing. They "did" control the old "separate pages for each" listings for S/N/A/F pages, but are no longer relevant (and IIRC, the Bootstrap template also ignores):
Screen Shot 2024-01-25 at 8 01 16 PM

drbyte:
Updated PR description:
- S/N/A/F pages now use the same product-listing template as the category navigation does
- all these pages will also follow the setting for "columns per row", so if the product-listing uses rows, so will S/N/A/F page; likewise, if set to 3 columns, all pages will use 3 columns. (Example in screenshots posted)
- the "click column headings for sort" is replaced by the sort dropdown

drbyte:
Note: Integration with BS4 template works out-of-the-box if you skip installing the header_php_zca_xx.php files for featured/new/all/specials ... but those files can be easily patched to happily co-exist without causing problems: just do an early-return if `!isset($xyz_query_raw)`.
Merging of the 4 `tpl_foo_default.php` template files will add the drop-down-sort if desired.

drbyte:
> OK - updates needed to https://docs.zen-cart.com/user/template/new_featured_all_listing_page_configuration/ (likely would have a pre-2.0.0 page with this content and a new page with less content.)

Right. Probably can just add a note at the top of that page saying that these pages are now all configured via the Product Listing settings since v2.0.0 and link them to https://docs.zen-cart.com/user/template/product_listing_page_configuration/

PR #6155, opened 2024-01-27
Add product "dimension" field stubs to shopping_cart.

drbyte:
These are used by some shipping modules which offer dimensional quoting capability. Including these here means those shipping modules don't need to overwrite the shopping_cart class file.


zcwilt (approved):

zcwilt:
my only comment would be that in the future would be better to have a way of adding variables like this without changing core code, either via a notifier or refactoring code to be more 'open' to changes

PR #6157, opened 2024-01-27
Updates for strict-types

### LINKED ISSUE #6082: functions_products: zen_get_prid(): Argument #1 ($uprid) must be of type string, int given
add 
declare(strict_types=1);
to the start of functions_products

Add a product to the cart
PHP Fatal error:  Uncaught TypeError: zen_get_prid(): Argument #1 ($uprid) must be of type string, int given, called in includes\functions\functions_products.php on line 550 and defined in \includes\functions\functions_products.php:458

I suggest adding the same (declare(strict_types=1);) to the shopping_cart class at the same time for related useages/as it is debateable exactly where fixes should be applied.

drbyte:
Fixes #6082

### DIFF
--- a/admin/includes/init_includes/init_special_funcs.php
+++ b/admin/includes/init_includes/init_special_funcs.php
@@ -10,7 +10,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 }
 
 // set a default time limit
-zen_set_time_limit(GLOBAL_SET_TIME_LIMIT);
+zen_set_time_limit((int)GLOBAL_SET_TIME_LIMIT);
 
 // -----
 // Load required sales/specials/etc function-files for core use, in preparation to
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -313,11 +313,11 @@ class shoppingCart extends base
                     //add htmlspecialchars processing.  This handles quotes and other special chars in the user input.
                     $attr_value = null;
                     $blank_value = false;
-                    if (strpos($option, TEXT_PREFIX) === 0) {
+                    if (is_string($option) && str_starts_with($option, TEXT_PREFIX)) {
                         if (trim($value) === '') {
                             $blank_value = true;
                         } else {
-                            $option = substr($option, strlen(TEXT_PREFIX));
+                            $option = substr((string)$option, strlen(TEXT_PREFIX));
                             $attr_value = stripslashes($value);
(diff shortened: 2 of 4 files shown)


torvista:
But then in strict types mode, **explode** on the next line complains when it is fed an integer...I've got functions_products and the shopping cart class using strict types mode and there are ("only") half a dozen places where casting is needed to clear all the errors (when a product is added to the cart).

drbyte:
Right.

Remember, `coercive typing` is one of the reasons people choose PHP over other languages like Java and C that are heavily strictly-typed.

torvista:
While I may the only one beating ZC with this stick, there are really very few places left where ZC falls foul of this/it's nearly ready for a stricter world. Be nice to clear them.

drbyte:
Test with several non-attribute products, with fractional quantities. Errors not thrown.

You probably have some specific products that warrant testing this with, though.

torvista:
Simple products give no errors.

Product with attributes: 
PHP Fatal error:  Uncaught TypeError: strpos(): Argument #1 ($haystack) must be of type string, int given in includes\classes\shopping_cart.php:316

https://github.com/zencart/zencart/blob/e6bce467a03a354704aac28ba9bb77634e73a403/includes/classes/shopping_cart.php#L316

drbyte:
Thanks. I'll reset my demo data later and check this

drbyte:
Lots more updates made.
There's a HUGE risk of introducing rounding errors due to the loss of precision that occurs when trying to implement strict-typing with numeric data when it could legitimately be an `integer` or a `float` interchangeably ... such as with fractional quantities and with taxes.
I really hope this didn't break anything.

torvista:
In my limited testing, with the shopping_cart class, functions_products and functions_general_shared all using strict types, I could checkout with simple and attributed products with no rounding errors.

Only two simple errors:

https://github.com/zencart/zencart/blob/e6bce467a03a354704aac28ba9bb77634e73a403/includes/functions/functions_general_shared.php#L20

$limit being passed as string instead of int.

https://github.com/zencart/zencart/blob/e6bce467a03a354704aac28ba9bb77634e73a403/includes/functions/functions_general_shared.php#L191-L197

$key being passed to stripslashes as an int instead of a string

drbyte:
> $key being passed to stripslashes as an int instead of a string

This is because elsewhere (some of) those values were cast to int for other reasons.

PR #6168, opened 2024-01-28
Admin order's details, display overall weight

lat9:
While the `orders::order_weight` handling was added in zc156, it's not displayed anywhere in the admin and could be a useful troubleshooting element.  Like when a product's weight hasn't been set in its definition.

### DIFF
--- a/admin/includes/extra_datafiles/dist-site_specific_admin_overrides.php
+++ b/admin/includes/extra_datafiles/dist-site_specific_admin_overrides.php
@@ -108,6 +108,18 @@
 // false .... hide IP address (default)
 // $show_registration_ip_in_listing = false;
 
+// Display an order's overall weight and the weight of each product?  Used
+// Used in: admin/orders.php
+//
+// Note: Orders placed on Zen Cart versions prior to 1.5.6 did not record
+// the order's overall weight.  If an order's overall weight isn't recorded,
+// this setting will automatically be set to (bool)false for that order.
+//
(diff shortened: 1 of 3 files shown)


lat9:
I'm (now) thinking perhaps to also add the individual products' weights to the products table.

scottcwilson:
Could be very powerful.  Make it switchable through `site_specific_admin_overrides.php` ?

lat9:
> Could be very powerful. Make it switchable through `site_specific_admin_overrides.php` ?

Sounds good; the default will be 'on'.

drbyte (approved):

asiksami:
Nice addon, thanks @lat9.

Just a thought. Could this product row weight be total weight of the selected products?

When one is processing the order and quicky checks the weight from the row, it might just use the weight of a one item and not add it with the quantity. This might also be problematic when quantity of a product is lower than 1 (eg. 0.7), and especially on heavy item weight is way too high.

asiksami:
> > Nice addon, thanks @lat9.
> > Just a thought. Could this product row weight be total weight of the selected products?
> > When one is processing the order and quicky checks the weight from the row, it might just use the weight of a one item and not add it with the quantity. This might also be problematic when quantity of a product is lower than 1 (eg. 0.7), and especially on heavy item weight is way too high.
> 
> Sounds like a good idea! Please open a separate issue (rather than having it buried in this already-merged PR) so that it's got visibility.

Great. Opened new issue #6180

PR #6173, opened 2024-01-29
update admin to Bootstrap 5.3.2

zenexpert:
update admin to Bootstrap 5.3.2


drbyte:
Screen Shot 2024-01-29 at 11 32 56 AM
Screen Shot 2024-01-29 at 11 33 21 AM
Screen Shot 2024-01-29 at 11 35 24 AM
Screen Shot 2024-01-29 at 11 34 41 AM

zenexpert:
dang... forgot to remove the local bootstrap css file (old one)

drbyte:
> dang... forgot to remove the local bootstrap css file (old one)

Oh. I didn't even think to check that. 

We might still want to keep that, but would have to ensure that people delete the old local file, or maybe drop the `.min` part from the filename that we look for. 

But, I'm not sure how many people will want to override their admin Bootstrap files with local ones. Way less admin traffic, plus browser caching, both make it kinda moot.

drbyte:
Screen Shot 2024-01-29 at 12 15 18 PM

drbyte:
> update admin to Bootstrap 5.3.2

Thanks. It's a lotta work to redo all that markup.

zenexpert:
> We might still want to keep that, but would have to ensure that people delete the old local file, or maybe drop the `.min` part from the filename that we look for.
> 
> But, I'm not sure how many people will want to override their admin Bootstrap files with local ones. Way less admin traffic, plus browser caching, both make it kinda moot.

I'll leave the decision to you, but personally I don't see a point of loading a local css file when it can be loaded from a CDN. If someone insists on overriding it with a local version for whatever reason, they'll find a way, I guess...

lat9:
I've not looked at the code, but I know that there are *major* Bootstrap class differences between 3 and 5.  Does Bootstrap 5 provide a polyfill?  I've got a ton of plugins that have been reworked to use the Bootstrap 3 classes and it's going to be a right PITA if not.

zenexpert:
No, BS5 does not provide a polyfill, but there is an active project to support IE11: https://github.com/coliff/bootstrap-ie11
Personally, I don't see the need to support that dinosaur, especially in the admin section so you folks can add it if you want.
As for BS3 to BS5 - yeah, there are some major differences, but nothing that would be too hard to update IMO. I'm not sure what your admin files have and what's the structure, but I'm hoping it won't be such a PITA to update...

scottcwilson:
Suggest deferring merge until after 2.0.0 release.

lat9:
> No, BS5 does not provide a polyfill, but there is an active project to support IE11: https://github.com/coliff/bootstrap-ie11 Personally, I don't see the need to support that dinosaur, especially in the admin section so you folks can add it if you want. As for BS3 to BS5 - yeah, there are some major differences, but nothing that would be too hard to update IMO. I'm not sure what your admin files have and what's the structure, but I'm hoping it won't be such a PITA to update...

It's not so much that it would be a PITA to update, but it will make it _very difficult_ for a plugin to support both the BS3 and BS5 requirements in a single admin directory structure.

drbyte:
Can BS3 markup (class names) be left in-place and BS5 classes be added alongside, without causing undesired side-effects?
ie: if the BS5 core CSS/JS isn't installed and both BS3 and BS5 class names are in a page's HTML, will there be a problem?

ie: instead of "replacing BS3 classes with BS5 classes", can the BS5 classes be "added", for compatibility with both, depending on whether BS3 or BS5 "core" is installed/loaded in the store?
That way plugins on an old site with only BS3 installed can work even though the plugin's HTML produces both BS3 and BS5 markup?

zenexpert:
In theory, they can coexist, but it would require loading both BS3 and BS5 libraries, and it could be challenging to avoid visual or javascript errors. It's definitely something worth testing - can you think of any plugins that have BS3 admin pages? However, in the long run, it would probably be best to transition to BS5 and squeeze out BS3...

drbyte:
> In theory, they can coexist, but it would require loading both BS3 and BS5 libraries

No, my point  is that the site would only load EITHER 3 or 5, never both.


> can you think of any plugins that have BS3 admin pages? 

lat9 was talking about numerous plugins she's built for current ZC using BS3. And having to maintain them to work on both old and new sites.

I'm trying to determine how safe it is to stuff both BS3 and BS5 markup into those plugins, for compatibility with both, so that if a store has BS3 it'll work, and if it instead has BS5 it'll work

zenexpert:
> No, my point is that the site would only load EITHER 3 or 5, never both.
But if the site doesn't load BS5, how will the core admin be updated to BS5? You need to load BS5 for admin core, and you would need to either load BS3 for the plugins or modify markup in the plugins for BS5. Or am I missing your point?

I was under the impression your idea was to allow plugins to use BS3 while core admin uses BS5, and then plugins could eventually be updated to BS5.

> I'm trying to determine how safe it is to stuff both BS3 and BS5 markup into those plugins, for compatibility with both, so that if a store has BS3 it'll work, and if it instead has BS5 it'll work
I think this would be OK, because "float-end pull-right" won't cause problems and the plugin will work on both BS3 and BS5.

lat9:
Some of the plugins I maintain that have been converted to BS3 mark-up are: Edit Orders, ZCA Bootstrap Template (its color-definition tool), PayPal Restful's admin notifications, Google Product Search Feeder II, SiteMap XML, Clone a Template, Image Handler ... and the list goes on.  Also various commercial plugins and site-specific admin customizations (sigh).

zenexpert:
> Some of the plugins I maintain that have been converted to BS3 mark-up are: Edit Orders, ZCA Bootstrap Template (its color-definition tool), PayPal Restful's admin notifications, Google Product Search Feeder II, SiteMap XML, Clone a Template, Image Handler ... and the list goes on. Also various commercial plugins and site-specific admin customizations (sigh).

Awesome, thanks! I'll play with it tomorrow - kinda like the idea of loading BS5 for admin core and BS3 for backward plugin compatibility. Just not sure if it'll work, hehe...

drbyte:
> But if the site doesn't load BS5, how will the core admin be updated to BS5? You need to load BS5 for admin core, and you would need to either load BS3 for the plugins or modify markup in the plugins for BS5. Or am I missing your point?

Ya, my point was that we could leave the ZC core admin using BS5 (when this is merged), but plugin authors would update their plugins to contain both BS3 and BS5 styles in their custom admin pages. That way whether the store is using BS3 or using BS5 (not both at the same time), the plugins would still display properly.

zenexpert:
I've tried my best, but the idea of running BS3 and BS5 at the same time just doesn't play well. There's always something that's conflicting and I couldn't find a way to allow current plugins to run normally without updating their markup. Sorry @lat9 :)

However, I did check Edit Orders and Image Handler and there's minimal markup to update, after which it could simultaneously work on either BS3 or BS5 (meaning easier to maintain future updates).

drbyte:
> the idea of running BS3 and BS5 at the same time just doesn't play well. There's always something that's conflicting

Agreed. That'll never be a possibility. (specifically, can never install and load BS3 and BS5 JS/CSS at same time).

> and there's minimal markup to update, after which it could simultaneously work on either BS3 or BS5 (meaning easier to maintain future updates)

Excellent. That'll make it easier on both the maintainer (add the markup and release) and on the storeowner (just drop it in and it works).

PR #6177, opened 2024-01-30
add label (accessibility) to currencies sidebox

### LINKED ISSUE #6016: Form in Currencies Sidebox Missing Aria Label (Accessibility)
Apologies for not catching this in one of our earlier sweeps but the currencies sidebox fails standard 4.1.2 Name, Role, Value.

That means the standard 
`$content .= '<label for="currency">Select Currency:</label>';`
will not clear the failure.

It needs a select and option field that I have not been successful in creating myself or I would submit it for consideration.

OldNGreyBMR:
fixes #6016

### DIFF
--- a/includes/templates/template_default/sideboxes/tpl_currencies.php
+++ b/includes/templates/template_default/sideboxes/tpl_currencies.php
@@ -9,8 +9,10 @@
  */
 $content = '';
 $content .= '<div id="' . str_replace('_', '-', $box_id . 'Content') . '" class="sideBoxContent centeredContent">';
+$content .= '<label for="Currency">';
 $content .= zen_draw_form('currencies_form', zen_href_link(basename(preg_replace('/.php/','', $PHP_SELF)), '', $request_type, false), 'get');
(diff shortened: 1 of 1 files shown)


includes/templates/template_default/sideboxes/tpl_currencies.php
```diff
@@ -9,8 +9,10 @@
  */
 $content = '';
 $content .= '<div id="' . str_replace('_', '-', $box_id . 'Content') . '" class="sideBoxContent centeredContent">';
+$content .= '<label for="Currency">';
```
lat9, line 12: For the `for=` attribute to be associated with an input/select tag, the attribute's value needs to match the input/select tag's `id=` attribute.  Since the call to `zen_draw_pull_down_menu` doesn't specify an `id=` attribute, that attribute's value will be `select-currency` (noting that case matters).

Note also that there's a Zen Cart function to draw those labels.  Instead of `<label for="Currency">`, use
```php
$content .= zen_draw_label(PLEASE_SELECT, 'select-currency', 'class="sr-only"');
```
_after_ line 13 (so it's in the form) and lose the `</label>` tag on line 17.

lat9 (changes requested):

lat9 (approved):

PR #6178, opened 2024-01-30
Auto apply coupon from affiliate link or query param.

neekfenwick:
See discussion at https://github.com/zencart/zencart/discussions/5856

Uses $_GET['coupon_code'], or HTTP_REFERER header, to resolve a coupon to apply to the current session.
Adds one 'referrer' field to `coupons` table with admin page support.


includes/init_includes/init_coupons.php
```diff
@@ -0,0 +1,93 @@
...
+function initCouponReferrerCheck(): ?string {
+    global $db;
+
+    if (empty($_SERVER['HTTP_REFERER'])) {
+        // There is no referer header, cannot do any lookup.
+        return null;
+    }
+    $referrer = $_SERVER['HTTP_REFERER'];
+
+    // Check for coupon that probably matches this referer.  More expensive validation done later.
```
scottcwilson, line 17: Extensive?

includes/init_includes/init_coupons.php
```diff
@@ -0,0 +1,93 @@
...
+    $matches = [];
+    $result = preg_match('/^(?:https?:\/\/)?(?:[^@\n]+@)?(?:www\.)?([^:\/\n?]+)/', $referrer, $matches);
+    if ($result !== 1) {
+        return null;
+    }
+    $domain = $matches[1];
+
+    $sql = "SELECT coupon_code
+        FROM " . TABLE_COUPONS . "
+        WHERE referrer LIKE '%:referrer:%'";
```
scottcwilson, line 28: Why not just `=`?  Would you please provide some explanatory examples?

includes/init_includes/init_coupons.php
neekfenwick: I refer to the loading of lang strings and instantiation of ot_coupon as being expensive.  Cheap validation is to directly query `coupons` table without regard to whether the ot_coupon module is enabled.

includes/init_includes/init_coupons.php
neekfenwick: It's to support a comma separated list of values in the `coupons.referrer` field (as mentioned in the admin page tooltip on the Referrer textbox), e.g. "a.com,b.com,c.com", so an incoming request with HTTP_REFERER "b.com" would match using `referrer like '%b.com%'`. This is a very lazy way to allow multiple values, and would fail for substrings, I was too lazy to think of a more robust way to do it :)

scottcwilson: Would it make sense to do a REGEXP match?   A bit of a pain since this has changed in MySQL 8 vs 5 but something to consider.  Or perhaps just use `=` but have a notifier?   Not strong on this idea, just looking ahead to the documentation.

neekfenwick: Yeah I thought about REGEXP too, but from a user's point of view, I feel they would want 1/ something easy to use, and REGEXP syntax is not easy to non-techies, and 2/ to name 1 or more well known domains on which links might be hosted, e.g. if content may appear on several of an affiliate's sites which may be similar or quite different e.g. case 1 "tomshardware.com OR tomsreviews.com" or case 2 "sofaking.com OR bedsandbeyond.com".  Perhaps wildcards would be useful for the first, but not the second. Seems best to allow them to "list" domains somehow, and the simplest way I thought of was to comma separate them.

One approach would be to accept CSV data on the admin UI but actually store it in a better format in the DB, for example "a.com,b.com" on the UI is saved as "~a.com~b.com~" (taking "~" as an irregular if not illegal char in URL) .. then the SQL could assume the referer value is to be found surrounded by "~" chars, say "LIKE '~a.com~'".

Does that get a thumbs up?
(I assume a notifier is generally a bad way to go so I'm not exploring that idea :) )

drbyte: @neekfenwick Do you envision the `referrer` feature being used with 10-20 coupons at any given time, vs 100, vs 1,000,000 ?
Obviously it's ideal to code for scalability, but if the volume of affected coupons and/or domains isn't astronomical, then perhaps it's safe to do some reasonable additional parsing beyond just the `LIKE` query.

I'm thinking of something where accepted domains are retrieved and then scanned for matches. This would help with the exact-match problem you mentioned, too.

Perhaps doing the LIKE query is fine, to limit down to which records need more parsing ... BUT ... 
Then `explode()` the `referrer` val from the db (explode on comma `,`) and check for an exact match against the incoming referrer.

This would require no complex changes to how you're storing the data, and not make it any more complex for the data-entry user putting in the domains.

drbyte:
@zcwilt are the tests failing simply because of the added field in the coupon table?

zcwilt:
@drbyte probably.

zcwilt:
There's definitely some pain in the way tests work at the moment which I need to address

zcwilt:
also @neekfenwick while this PR contains upgrade sql, it doesn't contain a install sql change.

includes/init_includes/init_coupons.php
```diff
@@ -0,0 +1,93 @@
...
+    // Check for coupon that probably matches this referer.  More expensive validation done later.
+    // Strip the domain name from the URL e.g. https://www.blah.com/page.html becomes blah.com
+    $matches = [];
+    $result = preg_match('/^(?:https?:\/\/)?(?:[^@\n]+@)?(?:www\.)?([^:\/\n?]+)/', $referrer, $matches);
+    if ($result !== 1) {
+        return null;
+    }
+    $domain = $matches[1];
+
+    $sql = "SELECT coupon_code
```
drbyte, line 26:
```suggestion
    $sql = "SELECT coupon_code, referrer
```

includes/init_includes/init_coupons.php
```diff
@@ -0,0 +1,93 @@
...
+        WHERE referrer LIKE '%:referrer:%'";
+    $sql = $db->bindVars($sql, ':referrer:', $domain, 'noquotestring');
+
+    $result = $db->Execute($sql, 1);
+
+    if ($result->EOF) {
+        return null;
+    }
+
+    return $result->fields['coupon_code'];
```
drbyte, line 37:
```suggestion
    $referrers = explode(',', $result->fields['referrer']);
    $referrers = array_map(fn ($val) => trim($val), $referrers);
    if (! in_array($domain, $referrers)) {
        return null;
    }

    return $result->fields['coupon_code'];
```

zcwilt:
I've updated the way the test suite works, so it should honour changes to the install sql, once this PR adds an install sql file change

includes/init_includes/init_coupons.php
neekfenwick: On getting up this morning I see the merge has now happened with your code changes :) personally I would sanitise the value entered by the admin when saving `coupons.referrer` (excise whitespace etc) instead of sanitising it in init_coupons.php every time.  But, your code works so we can let it be.  I hadn't considered scalability much, agree that if thousands of coupons exist we would want some optimisation, indexing on the `referrer` column, or something.  Can look at that if required.  I am actually building a system that assigns unique coupons to visitors which will result in thousands of coupons so I might hit that problem myself :)

neekfenwick:
Thank you for all the help @scottcwilson @drbyte @zcwilt 😘

drbyte:
> @neekfenwick index on the column

Good point. 
However I'm not sure that an index will help with a LIKE query :(

scottcwilson:
Most stores have a small number of coupons, so search time shouldn't be bad; my concern with LIKE would be false positive matching.

scottcwilson:
@neekfenwick Would you please write up instructions for the documentation?   Please them in here or PR against `user/order_total/coupons.md` in the doc repo.  I have an entry in what's new.

neekfenwick:
> @neekfenwick Would you please write up instructions for the documentation? Please them in here or PR against `user/order_total/coupons.md` in the doc repo. I have an entry in what's new.

Please see PR https://github.com/zencart/documentation/pull/1159

scottcwilson:
Super!  This has been merged, and the next update to Whats New for 2.0.0 will link to the new section.  Thank you.

PR #6179, opened 2024-01-30
harden inclusion of code for feature testing

zcwilt:


scottcwilson:
@zcwilt do you want to add `includes/application_testing.php` to the `.gitattributes` file so it's not part of the build?

drbyte:
> @zcwilt do you want to add `includes/application_testing.php` to the `.gitattributes` file so it's not part of the build?

It's been moved to the `not_for_release` dir, which is already excluded via `.gitattributes`

scottcwilson:
Oh sorry I didn't notice that.  TY!

PR #6182, opened 2024-01-30
Orders page - make customer lookup check by cID before email

scottcwilson:
Current customer search is by email address.  This fails if the customer has changed their email address, but will work if the customer id associated with the order is non-zero and is used in a search.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -621,7 +621,15 @@ if (!empty($action) && $order_exists === true) {
               </tr>
               <tr>
                 <td class="noprint"><strong><?php echo ENTRY_CUSTOMER; ?></strong></td>
-                <td class="noprint"><?php echo '<a href="' . zen_href_link(FILENAME_CUSTOMERS, 'search=' . $order->customer['email_address'], 'SSL') . '">' . TEXT_CUSTOMER_LOOKUP . '</a>'; ?></td>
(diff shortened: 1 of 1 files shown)


admin/customers.php
```diff
@@ -1574,6 +1574,8 @@ class="btn btn-default"><?php
             'a.entry_postcode',
         ];
         $search = zen_build_keyword_where_clause($keyword_search_fields, trim($keywords), true);
+    } else if (!empty($_GET['id'])) {
+        $search = " WHERE c.customers_id = " . (int)$_GET['id'];
```
drbyte, line 1578: This added check for `GET[id]` isn't necessary. `cID` is already built-in.

drbyte: It is already linked in the table-view of orders (see magnifying glass icon next to customer name) and in the info-sidebox (the email address is actually linked to a cID=(number) lookup).

drbyte: Screen Shot 2024-01-30 at 11 48 57 AM
Screen Shot 2024-01-30 at 11 49 15 AM

scottcwilson: Right but not on the order details page where I have it now.   Will make the other change above.

scottcwilson: I see your point.  Fixed.

PR #6186, opened 2024-01-31
Refine default-sort-order dropdown selection

### LINKED ISSUE #6185: Product Listing not honouring default order by PRODUCT_LISTING_DEFAULT_SORT_ORDER
ZC200 today, php 8.3

When loading an index listing the first time/the sorter has not been used, the sort order should be set in default_filter, I think.

It appears the $listing_sql has no order_by applied at all.

I've added debug code to default_filter and product_listing, which shows the progression and shows the order_by is correctly generated from the admin setting PRODUCT_LISTING_DEFAULT_SORT_ORDER,  but does not get appended.

image

When the sorter is used, the order_by is appended correctly.

image

drbyte:
Fixes #6185 

Most of it is fixed by the changes to `default_filter.php`.

Also added a new dropdown choice, "Recommended", which uses the store-owner-determined `product_sort_order`

### DIFF
--- a/includes/index_filters/default_filter.php
+++ b/includes/index_filters/default_filter.php
@@ -12,6 +12,10 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: DrByte 2024 Jan 27 Modified in v2.0.0-alpha1 $
  */
+/**
+ * @var queryFactory $db
+ * @var notifier $zco_notifier
+ */
 if (!defined('IS_ADMIN_FLAG')) {
     die('Illegal Access');
 }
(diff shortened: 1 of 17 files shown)


torvista:
ok. Works for me.
One point regarding default/first load, this will default to PRODUCT_LISTING_DEFAULT_SORT_ORDER.
If not set, it falls to hard-coded $order_by = " ORDER BY p.products_sort_order, pd.products_name";

I'll note that PRODUCT_LISTING_DEFAULT_SORT_ORDER can only reference one of the column numbers defined on the same admin page...so if you wanted a DESC sort order, you can't override it with this switch as sort order is not an available column.

Really just putting this here for a note/not wanting a solution as the whole file is easily overridden anyway.
(I use sort orders where a higher sort order is more important/should be shown first).

drbyte:
> I'll note that PRODUCT_LISTING_DEFAULT_SORT_ORDER can only reference one of the column numbers defined

Yes, that's been bugging me.

I've reworked it to treat that setting the same way the other sort-defaults are: choose a value of 1-8.  If it's left blank, it defaults to 8 behind the scenes because blank always meant products_sort_order which is what 8 does. If it's more than one character (ie: the legacy '2a' format) then it's ignored and treated as though it were '8'.

Issue #6183, opened 2024-01-30
Missing variable $_GET['gv_no'] causes log


scottcwilson:
Could be malicious.

[30-Jan-2024 04:34:49 UTC] Request URI: /index.php?main_page=gv_redeem, IP address: .
#0 /home/client/public_html/store/includes/modules/pages/gv_redeem/header_php.php(13): zen_debug_error_handler()
#1 /home/client/public_html/store/index.php(35): require('/home/client/p...')
--> PHP Warning: Undefined array key "gv_no" in /home/client/public_html/store/includes/modules/pages/gv_redeem/header_php.php on line 13.

zcwilt:
While not necessarily malicious (e.g. scanning software might hit this) still think is should be addressed.

I'll take a look.

PR #6187, opened 2024-01-31
Bug/missing gv no on gvredeem

zcwilt:


zcwilt:
fixes #6183

zcwilt:
there is probably more that could be done test wise here, for gv stuff, but will address this separately

PR #6190, opened 2024-01-31
Use zc_install database intaller rather than migrations

zcwilt:
using migrations misses PR's that update the sql


zcwilt:
this definitely needs some more love

PR #6192, opened 2024-02-01
Extend the 'fluid' layout mode from products to categories too

### LINKED ISSUE #6246: Lazy Load Product Listing Images
`loading="lazy"` is currently (only) used for the listing of category images: 

https://github.com/zencart/zencart/blob/444b7e45ae8acc84bfbbf12cb391b368c601b7b7/includes/modules/category_row.php#L44

I think it would be beneficial to add it to the product listing page in `\includes\modules\product_listing.php`. As that file is overriden in responsive_classic then I presume an edit to `\includes\modules\responsive_classic\product_listing.php` is also needed.

I can submit a PR if need be (probably not until tomorrow).

drbyte:
Pre-settable variables include:
- `$grid_category_classes_matrix`
- `$grid_category_cards_classes`
- `$grid_category_card_params`
- `$grid_category_wrap_classes`
- `$category_row_layout_style`

and
- `$grid_product_classes_matrix`
- `$grid_product_cards_classes`
- `$grid_product_card_params`
- `$grid_product_wrap_classes`
- `$product_listing_layout_style`

Inspired from https://github.com/lat9/ZCA-Bootstrap-Template/issues/305
Also adds lazy-loading of product images, and closes #6246

### DIFF
--- a/includes/modules/category_row.php
+++ b/includes/modules/category_row.php
@@ -14,40 +14,90 @@ if (!defined('IS_ADMIN_FLAG')) {
     die('Illegal Access');
 }
 
-$title = '';
 $num_categories = $categories->RecordCount();
+if (empty($num_categories)) {
+    return;
+}
 
-$row = 0;
-$col = 0;
+$rows = 0;
+$columns = 0;
 $list_box_contents = [];
-if ($num_categories > 0) {
-    if ($num_categories < MAX_DISPLAY_CATEGORIES_PER_ROW || MAX_DISPLAY_CATEGORIES_PER_ROW === 0) {
-        $col_width = floor(100/$num_categories);
-    } else {
-        $col_width = floor(100/MAX_DISPLAY_CATEGORIES_PER_ROW);
+$title = '';
+
+$columns_per_row = defined('MAX_DISPLAY_CATEGORIES_PER_ROW') ? (int)MAX_DISPLAY_CATEGORIES_PER_ROW : 0;
+if (empty($category_row_layout_style) || !in_array($category_row_layout_style, ['columns', 'fluid'])) {
+    $category_row_layout_style = $columns_per_row > 0 ? 'columns' : 'fluid';
+}
+
+// if in fixed-columns mode, calculate column width
+if ($category_row_layout_style === 'columns') {
+    $calc_value = $columns_per_row;
+    if ($num_categories < $columns_per_row || $columns_per_row === 0) {
+        $calc_value = $num_categories;
+    }
+    $col_width = floor(100 / $calc_value) - 0.5;
+}
+
+foreach ($categories as $next_category) {
+    $zco_notifier->notify('NOTIFY_CATEGORY_ROW_IMAGE', $next_category['categories_id'], $next_category['categories_image']);
+    if (empty($next_category['categories_image'])) {
+        $next_category['categories_image'] = 'pixel_trans.gif';
+    }
+    $cPath_new = zen_get_path($next_category['categories_id']);
+
+    // strip out 0_ from top level cats
+    $cPath_new = str_replace('=' . (int)TOPMOST_CATEGORY_PARENT_ID . '_', '=', $cPath_new);
+
+    //    $categories->fields['products_name'] = zen_get_products_name($categories->fields['products_id']);
+
+
+
+    // Set css classes for "row" wrapper, to allow for fluid grouping of cells based on viewport
+    // these defaults are inspired by Bootstrap4, but can be customized to suit your own framework
+    if ($category_row_layout_style === 'fluid') {
+        $grid_cards_classes = $grid_category_cards_classes ?? 'row row-clms-3';
+        if (!isset($grid_category_classes_matrix)) {
+            // this array is intentionally in reverse order, with largest index first
+            $grid_category_classes_matrix = [
+                '480' => 'row row-clms-1 row-clms-sm-2 row-clms-md-3 row-clms-lg-4 row-clms-xl-6',
+            ];
+        }
+
+        // determine classes to use based on number of grid-columns used by "center" column
+        if (isset($center_column_width)) {
+            foreach ($grid_category_classes_matrix as $width => $classes) {
+                if ($center_column_width >= $width) {
+                    $grid_cards_classes = $classes;
+                    break;
+                }
+            }
+        }
+        $list_box_contents[$rows]['params'] = 'class="' . $grid_cards_classes . ' text-center"';
     }
 
-    foreach ($categories as $next_category) {
-        $zco_notifier->notify('NOTIFY_CATEGORY_ROW_IMAGE', $next_category['categories_id'], $next_category['categories_image']); 
-        if (empty($next_category['categories_image'])) {
-            $next_category['categories_image'] = 'pixel_trans.gif';
-        }
-        $cPath_new = zen_get_path($next_category['categories_id']);
+    $style = '';
+    if ($category_row_layout_style === 'columns') {
+        $style = ' style="width:' . $col_width . '%;"';
+    }
+    $grid_category_card_params = $grid_category_card_params ?? 'categoryListBoxContents centeredContent back gridlayout';
+    $grid_category_wrap_classes = $grid_category_wrap_classes ?? '';
+    $list_box_contents[$rows][] = [
+        'params' => 'class="' . $grid_category_card_params . '"' . $style,
+        'text' =>
+            '<a href="' . zen_href_link(FILENAME_DEFAULT, $cPath_new) . '">' .
+            zen_image(DIR_WS_IMAGES . $next_category['categories_image'], $next_category['categories_name'], SUBCATEGORY_IMAGE_WIDTH, SUBCATEGORY_IMAGE_HEIGHT, 'loading="lazy"') .
+            '<br>' .
+            $next_category['categories_name'] .
+            '</a>',
+        'wrap_with_classes' => $grid_category_wrap_classes,
+        'card_type' => $category_row_layout_style,
+    ];
 
-        // strip out 0_ from top level cats
-        $cPath_new = str_replace('=0_', '=', $cPath_new);
-
-        //    $categories->fields['products_name'] = zen_get_products_name($categories->fields['products_id']);
-
-        $list_box_contents[$row][$col] = [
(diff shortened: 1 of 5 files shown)


includes/modules/category_row.php
```diff
@@ -14,40 +14,93 @@
...
-        $col_width = floor(100/MAX_DISPLAY_CATEGORIES_PER_ROW);
+$title = '';
+
+if (empty($num_categories)) {
+    return;
+}
+
+$columns_per_row = defined('MAX_DISPLAY_CATEGORIES_PER_ROW') ? MAX_DISPLAY_CATEGORIES_PER_ROW : 0;
+$product_listing_layout_style = (int)$columns_per_row > 1 ? 'columns' : 'fluid';
+if ($columns_per_row === 'fluid') $product_listing_layout_style = 'fluid';
```
scottcwilson, line 30: Should this be

if ($product_listing_layout_style === 'fluid') 
?   Because $columns_per_row is integer.

scottcwilson:
This doesn't seem to work when tested in latest master with RC even with the change I suggested above.  Looks like all categories are on the same row. 
test

scottcwilson:
Documentation note: Categories To List Per Row (MAX_DISPLAY_CATEGORIES_PER_ROW) now has the special value 0 just like Columns Per Row (PRODUCT_LISTING_COLUMNS_PER_ROW) does for products. 

https://docs.zen-cart.com/user/template/listing_columns/

drbyte:
> all categories are on the same row.

Actually that does mean it's "working" code-wise ... but you're right: the styling is missing.  Where'd I put that ...

includes/modules/category_row.php
drbyte: `$columns_per_row` is a string, which comes from the admin config setting. Yes, it should be numeric. And `'0'` means "fluid".
Originally I was accounting for the possibility of someone typing "fluid" instead of typing '0'.

However, I'd forgotten that we actually do have some numeric-only validation on that config setting.
I'll remove it.

drbyte:
> This doesn't seem to work when tested in latest master with RC even with the change I suggested above. Looks like all categories are on the same row. 

CSS updated: 
```diff
-.gridlayout.categoryListBoxContents { padding: 2em 0; }
+.gridlayout.categoryListBoxContents { padding: 2em; }
```

drbyte:
These are the changes to implement on the BS4 template. Pretty sure this is backward compatible to v158.
[fluid-cat-list-bs4-template.zip](https://github.com/zencart/zencart/files/14154719/fluid-cat-list-bs4-template.zip)

scottcwilson:
Couple of problems still: 

1. Doesn't seem to allow you to override the defaults using an externally defined `$grid_classes_matrix`.
The one I use is in `includes/extra_datafiles/grid_classes.php` and it's 

```
<?php
// From includes/modules/bootstrap/product_listing.php 
                $grid_classes_matrix = [
                    // '12' => 'row-cols-1 row-cols-sm-2 row-cols-md-3 row-cols-lg-4 row-cols-xl-6',
                    // '10' => 'row-cols-1 row-cols-md-2 row-cols-lg-4 row-cols-xl-5',
                    // '9' => 'row-cols-1 row-cols-md-3 row-cols-lg-4 row-cols-xl-5',
                    '8' => 'row-cols-1 row-cols-md-2 row-cols-lg-3 row-cols-xl-4',
                    // '6' => 'row-cols-1 row-cols-md-2 row-cols-lg-2 row-cols-xl-3',
                ];
```

1. I wonder if a different "category" grid_classes_matrix is needed, because you can have more than 1 category per row (if not more than one product per row). 

With what we have, here's what it looks like: 
mobile

scottcwilson:
Having said the above, I wonder if this should be a bootstrap only feature?  Setting PRODUCT_LISTING_COLUMNS_PER_ROW to 0 in responsive_classic looks horrible (particularly disasterous in mobile landscape mode); why would anyone expect MAX_DISPLAY_CATEGORIES_PER_ROW to support a fluid layout?

scottcwilson:
Bootstrap zip doesn't work at all; categories are smashed into a single line.   Some css must be missing here too.

includes/modules/category_row.php
```diff
@@ -14,40 +14,92 @@
...
-        $col++;
-        if ($col >= MAX_DISPLAY_CATEGORIES_PER_ROW) {
-            $col = 0;
-            $row++;
+
+    // set css classes for "row" wrapper, to allow for fluid grouping of cells based on viewport
+    // these defaults are based on Bootstrap4, but can be customized to suit your own framework
+    if ($product_listing_layout_style === 'fluid') {
+        $grid_cards_classes = 'row-cols-1 row-cols-md-2 row-cols-lg-2 row-cols-xl-3';
+        if (!isset($grid_classes_matrix)) {
```
drbyte, line 58: @scottcwilson This check for `!isset($grid_classes_matrix)` is the same as what the product_listing module does, and is the same as in the bootstrap class.

So if yours is not being picked up, maybe there's an order-of-operations issue?

drbyte:
With the BS4 template, the default setup has the `$center_column_width` calculated to 6 (12 - 3 - 3 = 6). So the `$grid_classes_matrix` having only the `8` option uncommented would mean it won't take effect.

drbyte:
Here is `responsive_classic` in the current implementation as this PR proposes:
Recording 2024-02-08 at 05 36 08 PM

includes/modules/category_row.php
drbyte: With the BS4 template, the default setup has the $center_column_width calculated to 6 (12 - 3 - 3 = 6). So the `$grid_classes_matrix` having only the 8 option uncommented would mean it won't take effect.

drbyte:
> I wonder if a different "category" grid_classes_matrix is needed, because you can have more than 1 category per row (if not more than one product per row).

I accept that suggestion. Updated to use `$grid_category_classes_matrix` instead.

scottcwilson:
I haven't pulled your new changes in, this is the old version.  When I use CATEGORIES_PER_ROW = 0 with RC, and set the grid class to yield a grid class with 3 across

```
<div class="row row-cols-1 row-cols-md-2 row-cols-lg-2 row-cols-xl-3 text-center">
```

The categories page shows 6 across.  

cats

scottcwilson:
Also, this logic: 
`                if ($center_column_width >= $width) {`
includes/modules/category_row.php line 72

Seems wrong - `$width` is in grid units but `$center_column_width` is in pixels.  So the comparison is like 670 vs 6. 


same for products:                   
`  if ($center_column_width >= $width) { ` 
on includes/modules/responsive_classic/product_listing.php line 195

drbyte:
> Seems wrong - $width is in grid units but $center_column_width is in pixels. So the comparison is like 670 vs 6.

True. 
But I guess if the keys in your `$grid_classes_matrix` are pixels instead of number of columns, then whatever classes you want to define them to would be applied.

I'm working on some adjustments to both modules to allow site-specific overrides to be easier to set without touching the modules as extensively. But I think those should be a separate PR after this one is done.

drbyte:
> I haven't pulled your new changes in, this is the old version. When I use CATEGORIES_PER_ROW = 0 with RC, and set the grid class to yield a grid class with 3 across
> 
> ```
> <div class="row row-cols-1 row-cols-md-2 row-cols-lg-2 row-cols-xl-3 text-center">
> ```
> 
> The categories page shows 6 across.
> 
> cats

Okay. I see your point. I kinda just ignored how many actual columns it was displaying because it dynamically just let them show up however many "fit" naturally ... like 'fluid' implies.  
Given the way RC's CSS works, I'm inclined to let fluid just do as-is ... unless someone wants to come up with a few specific CSS rules to add to the template to enforce a few variations of maximum numbers per row, and then we can set those into the `$grid_classes` vars for that template.

scottcwilson:
> True. But I guess if the keys in your `$grid_classes_matrix` are pixels instead of number of columns, then whatever classes you want to define them to would be applied.

OK - but that means that maybe the default keys for RC should be expressed in pixels too. 

I'm going to let this be and focus on how bootstrap comes out - it could be that people won't want to use fluid with RC anyway.

drbyte:
CSS classes added for Responsive Classic.  It's not perfect at every breakpoint (percentages aren't as one would expect) because of legacy things within the Responsive Classic CSS itself. But it delivers reasonably on the expectation when screens hit the defined sizes. 
The goal is not to make Responsive Classic to even use this `fluid` feature, but rather to accommodate it without serious gotchas, so that template authors can try it out and incorporate into their own if desired.

drbyte:
Pre-settable variables include:
- `$card_params`
- `$wrap_classes`
- `$grid_cards_classes`
- `$grid_category_classes_matrix`
- `$grid_classes_matrix`
- `$category_row_layout_style`
- `$product_listing_layout_style`

zcwilt:
Should code be honoring MAX_DISPLAY_CATEGORIES_PER_ROW
currently it seems to be using MAX_DISPLAY_CATEGORIES_PER_ROW -1

zcwilt:
bit more nuanced than just MAX_DISPLAY_CATEGORIES_PER_ROW -1 
set to 0 get a 4 column display - guess this is expected 
set to 1 is as above 
set to 2 just get 1 column 
set to 3 get 2 columns 
set to 4 get 3 columns
set to 5 get 3 columns

drbyte:
> Should code be honoring MAX_DISPLAY_CATEGORIES_PER_ROW
currently it seems to be using `MAX_DISPLAY_CATEGORIES_PER_ROW -1`

>bit more nuanced than just `MAX_DISPLAY_CATEGORIES_PER_ROW -1`
set to 0 get a 4 column display - guess this is expected
set to 1 is as above
set to 2 just get 1 column
set to 3 get 2 columns
set to 4 get 3 columns
set to 5 get 3 columns

Actually that wasn't intended. It was caused by a CSS padding issue that was throwing off the percentages.

Fixed.

drbyte:
Changed pre-settable variables to:
- `$grid_category_classes_matrix`
- `$grid_category_cards_classes`
- `$grid_category_card_params`
- `$grid_category_wrap_classes`
- `$category_row_layout_style`

and
- `$grid_product_classes_matrix`
- `$grid_product_cards_classes`
- `$grid_product_card_params`
- `$grid_product_wrap_classes`
- `$product_listing_layout_style`

Issue #6194, opened 2024-02-01
responsive classic: use of $show_table_header_row


torvista:
Is `$show_table_header_row` supposed to be there?

https://github.com/zencart/zencart/blob/a30bd090e0a3906034193e8127f83aa93e85763a/includes/modules/responsive_classic/product_listing.php#L42

I find no other reference to it.

drbyte:
Yes. It's an undocumented override one can use to re-enable the table header row if they really want it, such as for templates that are sticking with legacy row-based listing format and want to show the click-the-heading-to-sort row despite the sort dropdown already being present.
Undocumented because it's kindof an edge case. 

If you want to use it, stick it in site-specific-settings and set it to true.

scottcwilson:
Can this be closed?

torvista:
documented: https://github.com/zencart/documentation/pull/1170

PR #6202, opened 2024-02-02
add $show_table_header_row

torvista:
re: https://github.com/zencart/zencart/issues/6194#issuecomment-1922696805


torvista:
Is there a problem with this? I made the mistake of preparing the documentation too..

PR #6203, opened 2024-02-02
Admin orders show unit/total weight

### LINKED ISSUE #6180: Admin order's details | Replace individual product weight with row based total weight
Could admin order's `product` rows `weight` column be calculated as row based total weight?

When one is processing the order and quickly checks the weight from the row, it might wrongly just use the weight of a one item and not add it with the quantity. This might also be problematic when quantity of a product is lower than 1 (eg. 0.7), and especially on heavy item, weight is too high.

lat9:
Fixes #6180

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -860,7 +860,7 @@ if (!empty($action) && $order_exists === true) {
                         if ($products_weight_unit === null) {
                             $products_weight = '&mdash;';
                         } else {
-                            $products_weight_total = rtrim(number_format((float)($products_weight_unit * $order->products[$i]['qty']), 4), '0.');
+                            $products_weight_total = rtrim(number_format((float)($products_weight_unit * $order->products[$i]['qty']), 4, '.', ''), '0.');
                             $products_weight = "$products_weight_unit$weight_unit / $products_weight_total$weight_unit";
                         }
                 ?>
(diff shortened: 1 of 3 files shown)


admin/orders.php
```diff
@@ -847,11 +856,12 @@ function couponpopupWindow(url) {
...
                 <?php
                     if ($show_orders_weights === true) {
-                        $products_weight = $order->products[$i]['products_weight'];
-                        if ($products_weight === null) {
+                        $products_weight_unit = $order->products[$i]['products_weight'];
+                        if ($products_weight_unit === null) {
                             $products_weight = '&mdash;';
                         } else {
-                            $products_weight .= ' ' . ltrim(TEXT_PRODUCT_WEIGHT_UNIT, ' ');
+                            $products_weight_total = rtrim(number_format((float)($products_weight_unit * $order->products[$i]['qty']), 4), '0.');
```
drbyte, line 863: As opposed to `round()`, is `number_format()` intentionally used so as to apply American formatting rules? (The absence of 3rd, 4th params applies comma and period as separators).

lat9: `number_format` was chosen so that a value like `20.120` could be trimmed to `20.12` or a value like `20.000` could be trimmed to `20`.  I wasn't aware that formatting rules applied to weights, since that's not dealt with anywhere else in the code when handling weights.

lat9: Never mind, duh.  I'm (currently) using the default parameters like @drbyte indicated in his initial comment.

Issue #6193, opened 2024-02-01
Auto Apply Coupon | Ignore HTTP_REFERER check if it starts with HTTPS_SERVER


asiksami:
This is a nice addon, thanks @neekfenwick. In the future, this could work nicely on QR-codes / newsletters links.

Thought I didn't find a way to disable it in any way. It might not take that much resources on checking `$_GET['coupon_code']`, but if One is sure that it will not be used in its store or someone starts to exploit it somehow, you would like to have it disabled somewhere that it doesn't require code changes.

Would it be good to have this coupon auto apply to be disabled somewhere? Maybe soft configure to site-specific-overrides.php? Or maybe better as a database configure variable?

neekfenwick:
Hi @asiksami, I'd have to ask the likes of @drbyte about whether we should have a global kill-switch for the coupon auto apply feature and how best to implement it.  As you say, a `site_specific_overrides.php` flag or a new `configure` table database entry.  Because `ot_coupon` module may not be 'installed' i.e. the database entries it creates may not exist, then a `site_specific_overrides.php` flag seems wiser.

    $ot_coupon_coupon_code_auto_apply_disabled = true;

Bear in mind that two values are used in `init_coupons.php` which could theoretically be abused: `$_GET['coupon_code']`, and `$_SERVER['HTTP_REFERER']`, both of which are sanitized before use.

I am reluctant to implement further defenses such as this kill-switch, it seems over-cautious, and it's no more exploitable than any other query parameter that may be passed in.

What do you think @drbyte ?

I guess we don't need to consider specific sanitizing in `init_sanitize.php` for the `coupon_code` query param?  If it's provided, then the new code in `ini_coupons.php` will try to look it up, and it does use `zen_db_prepare_input` to sanitize it there.

scottcwilson:
> I am reluctant to implement further defenses such as this kill-switch, it seems over-cautious, and it's no more exploitable than any other query parameter that may be passed in.

I agree.  I think it's unlikely that this specific feature would be a target of attackers.  Most of what I see is URLs with embedded encoded queries, much of which is silently blocked by the latest releases.

neekfenwick:
So @asiksami it seems this is a case to close as WontFix, although one thought comes to me.

[...]

PR #6211, opened 2024-02-08
Add coupons.referrer edit UI and validation.

neekfenwick:
See conversation on Issue #6193 (which is closed, not sure how these PR's are supposed to work in this case).

Two improvements to the coupon-admin UI for the recently added `coupons.referrer` field:

- Search now includes `referrers` field.  Not perfect as it's just a substring search on the raw field, so if the field has `a.com,b.com` in it, you can search for "com,b" and it'll match, but this seems good enough for now.
- Edit UI now has a series of input boxes with +/- icons, and the update/preview handler pulls these values, validates that no other coupon has a specific referrer (displays error message and prevents save).

The actual values are still stored as a CSV list, so the catalog side does not need changing.

I wrote the +/- handling from scratch, adding new file `admin/includes/javascript/coupon_admin.js` with some jQuery code that's fairly abstract.  If it's useful elsewhere it could be re-employed, but I don't like the way the HTML hardcoded in there has to match the HTML used in `coupon_admin.php`.  Room for improvement, but good enough for now?

### DIFF
--- a/admin/coupon_admin.php
+++ b/admin/coupon_admin.php
@@ -722,12 +722,8 @@ switch ($_GET['action']) {
                 <td><?php echo COUPON_REFERRER; ?></td>
                 <td><?php
                 // Referrers inputs may be empty, sanitise here
-                $trimmed_referrers = array_map(function ($referrer) {
-                  return trim($referrer);
-                }, $_POST['coupon_referrer']);
-                $referrers = array_filter($trimmed_referrers, function ($referrer) {
-                  return !empty(trim($referrer));
-                });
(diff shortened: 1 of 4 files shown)


drbyte:
> Room for improvement, but good enough for now?

Agreed. 

> The actual values are still stored as a CSV list, so the catalog side does not need changing.

As discussed in #6193 comments there's probably value in splitting out the referrers into their own db table (eg: `coupon_referrers`) for performance reasons. It won't affect stores that have very few coupons. But for those that have lots and especially those with myriad referrers, the db query will be much more performant if it can be done on an indexable table.

PR #6215, opened 2024-02-10
Compatibility for pages without admin_html_head

### LINKED ISSUE #6214: Temporarily restore compatibilty for admin pages that don't use admin_html_head.php
[This PR](https://github.com/zencart/zencart/pull/5850) removed admin CSS and JS files that are used by current plugins and site-specific pages.  Other than those plugin/custom-pages non-use of `admin/includes/admin_html_head.php`, they would operate properly in a zc200 (and possibly PHP 8.3+) environment, but the pages are met with a 403-forbidden error due to a missing `admin/includes/stylesheet.css`.

Since it's possible that we'll be seeing upgrades from zc157 to zc200, I'm suggesting that we revert the removal of
- /admin/includes./menu.css
- /admin/includes/menu.js
- /admin/includes/stylesheet.css

This would be a _temporary_ measure (flagged for removal in zc210) to give site developers and plugin authors  notification of the deprecation rather than a 'failure to operate'.  Upon an admin-page's load, if the page is found **not to use** `admin_html_head.php`, a _one-time-per-admin-session_ message is displayed to the admin and a PHP deprecation log is written identifying the associated page. The message (both logged and displayed) would be something of the ilk:
```
This page requires updates for the next Zen Cart version.  Please let your site developer or plugin author know!
```

lat9:
Fixes #6214.

For each admin-session, the first time a page is loaded that doesn't use admin_html_head.php to perform the base CSS load, a message is displayed

image

... and a PHP Deprecated log is created.

```
[10-Feb-2024 14:32:17 UTC] Request URI: /zc200w/admin200/index.php?cmd=header_test, IP address: 127.0.0.1, Language id 1
#1 C:\xampp\htdocs\zc200w\admin200\includes\header.php(71): trigger_error()
#2 C:\xampp\htdocs\zc200w\admin200\header_test.php(58): require('C:\\xampp\\htdocs...')
#3 C:\xampp\htdocs\zc200w\admin200\index.php(11): require('C:\\xampp\\htdocs...')
--> PHP Deprecated: This page requires updates for the next Zen Cart version. Please let your site developer or plugin author know! in C:\xampp\htdocs\zc200w\admin200\includes\header.php on line 71.
```

### DIFF
--- a/admin/includes/admin_html_head.php
+++ b/admin/includes/admin_html_head.php
@@ -6,8 +6,16 @@
  * @version $Id: Nick Fenwick 2023 Jul 10 Modified in v2.0.0-alpha1 $
  */
 if (!defined('IS_ADMIN_FLAG')) {
-  die('Illegal Access');
+    die('Illegal Access');
 }
+
+// -----
+// Set a processing flag that indicates that this file has been loaded.  The
+// flag is used by /admin/includes/header.php to warn admins and developers
+// if the legacy stylesheet loading is currently in effect.  That legacy-loading
+// (and this section) will be removed in a subsequent release of Zen Cart.
+//
+$zen_admin_html_head_loaded = true;
 ?>
 <meta charset="<?php echo CHARSET; ?>">
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
--- a/admin/includes/header.php
+++ b/admin/includes/header.php
@@ -58,6 +58,18 @@ if (empty($action)) {
     $hide_languages = true;
 } // hide when other language dropdown is used
 
+// -----
+// If the current page-load did not use the admin_html_head.php for the CSS files'
+// loading, let the admin know via message and log a PHP Deprecated issue ... once for
+// each page during an admin's session.
+//
+// Note: This section will be removed in a future version of Zen Cart!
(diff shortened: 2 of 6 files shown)


scottcwilson:
I don't think this is a great idea.  Any file that doesn't use the new admin head likely also hasn't been upgraded to run under PHP 8 either.  We discussed and merged this change and shouldn't back it out.

lat9:
> I don't think this is a great idea. Any file that doesn't use the new admin head likely also hasn't been upgraded to run under PHP 8 either. We discussed and merged this change and shouldn't back it out.

I disagree.  I've got a plethora of (a) plugins and (b) site-specific tools that are PHP 8.3 compliant but I've just not had the chance to backport to use `admin_html_head.php`.  I'm pretty sure that I'm not the only site developer in this condition.

drbyte (approved):

drbyte:
While it's undesirable to back it out, I'm satisfied with keeping it until `v2.1.0`, since it's a breaking change that is being forced ... where there are no negative side-effects of keeping these files for another release cycle.
Obviously if there were complex interdependencies with core code, it wouldn't be this simple.

PR #6222, opened 2024-02-12
Admin, Option Values Sort: Add language handling for update by Option Name

torvista:
fixes: https://github.com/zencart/zencart/issues/5355

following on from [here](https://github.com/zencart/zencart/pull/5426)

Decided to work on one option at a time, to try and make it easier....so the rest will follow to complete the file overhaul.

For sites with more than one language, the selection is offered:

image

### DIFF
--- a/admin/option_values.php
+++ b/admin/option_values.php
@@ -109,13 +109,13 @@ switch ($_GET['action']) {
     <!-- body //-->
     <div class="container-fluid">
 
-        <h1><?= HEADING_TITLE; ?></h1>
+        <h1><?= HEADING_TITLE ?></h1>
         <!-- body_text //-->
         <?php
         if (empty($_GET['options_id'])) {
             ?>
             <div id="editOptionName">
-                <h2><?= TEXT_UPDATE_OPTION_VALUES; ?></h2>
+                <h2><?= TEXT_UPDATE_OPTION_VALUES ?></h2>
                 <?php
                 echo zen_draw_form('quick_jump', FILENAME_PRODUCTS_OPTIONS_VALUES, '', 'get', 'class="form-inline"');
                 echo zen_draw_label(TEXT_SELECT_OPTION, 'options_id', 'class="control-label"');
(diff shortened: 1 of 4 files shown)


torvista:
I don't see the error.....

drbyte:
> I don't see what the merge check error is.....

The tests are failing because you changed the language string that the test is hard-coded to look for:
Screen Shot 2024-02-15 at 3 52 13 PM

So the test will need to be updated to match what the language string is outputting, so that the test can confirm that the action is successful.

torvista:
Ah, I changed that to distinguish it from the other two, which will have their own messages.

torvista:
Is there something I need to do here?

drbyte:
> Is that error detail publicly viewable?

Yes. You can see the "Checks" tab in the top of a PR, or go to the "Actions" tab in this repository to see all the test run results.
eg: this is the one that was failing: https://github.com/zencart/zencart/actions/runs/7886767671/job/21520625066?pr=6222

drbyte:
In the test it says "... failed to assert that ... [big long HTML block] ... contains 'Option Values Default Sort Order'" ... which was the old value of HEADING_TITLE. 
Screen Shot 2024-02-19 at 10 33 32 PM
Screen Shot 2024-02-19 at 10 33 04 PM



The fix is to find the part in the test suite that has that old string hard-coded into it (tests often have things hardcoded in order to actually test that the expected result is accurate), and update that. eg: https://github.com/zencart/zencart/pull/6222/commits/9fad8e6e3104b5e60e877fc831260a47c9963c99

admin/option_values.php
```diff
@@ -86,112 +108,87 @@
 
     <!-- body //-->
     <div class="container-fluid">
-      <div class="table-responsive">
-        <h1><?php echo HEADING_TITLE; ?></h1>
+
+        <h1><?= HEADING_TITLE; ?></h1>
```
lat9, line 112: I'm liking the short-echo tags more and more, but when they're used, the final `;` in the associated PHP statement should not be included.  The same is true for the additional usages in this PR, please either revert the short-echo or remove the `;' in each place.

admin/option_values.php
```diff
@@ -86,112 +108,87 @@
...
+                <div class="row">
+                    <?php
+                    if (count($languages_array) > 1) { ?>
+                        <div class="radio"><label><?= zen_draw_radio_field('language', 'current_only') . TEXT_UPDATE_SORT_LANGUAGE_CURRENT; ?></label></div>
+                        <div class="radio"><label><?= zen_draw_radio_field('language', 'all', true) . TEXT_UPDATE_SORT_LANGUAGE_ALL; ?></label></div> <?php
+                    } ?>
+                    <button type="submit" class="btn btn-warning"><?= TEXT_UPDATE_SUBMIT; ?></button>
+                    <a href="<?= zen_href_link(FILENAME_PRODUCTS_OPTIONS_VALUES); ?>" class="btn btn-default" role="button"><?= IMAGE_CANCEL; ?></a>
+                </div>
+                <?php echo '</form>'; ?>
```
lat9, line 187: Since the rest of the changes are using short-echo (`<?='), this line should as well.

lat9 (changes requested):

admin/option_values.php
torvista: My tendency was to use them only when embedded in html, where the improvement is clear.
I actually think the full tag is clearer when in isolation, but I'll go with the flow.

lat9 (approved):
Thanks, @torvista.  The semi-colon removal request was based on a Google search of `php short echo tag with semicolon`  as well as [this](https://github.com/zencart/zencart/pull/6015#issuecomment-1902250885) issue-comment.  It appears that the global consensus is to drop the trailing `;` when using the short-echo tag.

PR #6227, opened 2024-02-15
Add template-specific settings support

drbyte:
This allows template-specific override capability.

### DIFF
--- a/includes/functions/functions_templates.php
+++ b/includes/functions/functions_templates.php
@@ -1,19 +1,29 @@
 <?php
 /**
- * @copyright Copyright 2003-2023 Zen Cart Development Team
+ * @copyright Copyright 2003-2024 Zen Cart Development Team
  * @license https://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: lat9 2022 Nov 27 Modified in v1.5.8a $
+ * @version $Id:  Modified in v2.0.0-alpha2 $
  */
 if (!defined('IS_ADMIN_FLAG')) {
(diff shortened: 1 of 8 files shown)


lat9 (commented):
It might be worth mentioning in the `template_settings.php` files' comments that database setting (i.e. constant definitions) cannot be overridden, only the soft settings.

drbyte:
FYI - ZC load order:
- extra-datafiles before breakpoint 0
- db constants at breakpoint 40
- non-db constants at breakpoint 45
- functions files at breakpoint 60
- template-settings at breakpoint 110, followed by language-defines and extra_definitions

drbyte:
For interoperability with the BS4 template, and to allow it to use the `template_settings.php` file on older ZC versions, since extra_definitions loads fairly soon after the core template_settings.php would get loaded, I'm pretty sure we can simply add the following to the `extra_definitions/zca_bootstrap_id.php` file, or a new file maybe named `zca_bootstrap_settings.php`:
```php
// For pre-v2.0.0 support, load the template_settings.php file if not yet loaded
if (!isset($tpl_settings['IS_ZCA_BOOTSTRAP_TEMPLATE'])) {
    require_once DIR_WS_TEMPLATE . 'template_settings.php';
}
```

Then later one could review whether the `extra_datafiles/site-specific-bootstrap-settings.php` needs to remain where it is or if its content could be relocated to `template_settings.php`. Load-order may be the primary determinant there.

drbyte:
Documentation: https://docs.zen-cart.com/dev/code/template_settings/

PR #6233, opened 2024-02-18
Add product dimension fields

drbyte:
Add product dimension fields. Primarily to be used by shipping modules to prepare more accurate shipping quotes.

Also addresses #6232 

Screen Shot 2024-02-17 at 10 23 52 PM

Screen Shot 2024-02-17 at 10 25 54 PM


scottcwilson:
Love it!  Looks like Canada Post. :)

scottcwilson:
New doc: https://docs.zen-cart.com/user/shipping/shipping_dimensions/ plus references in What's New and the Release Specific Upgrade Considerations pages.

torvista:
Couple of typos

2024-02-18 21_28_22-Shipping Dimensions _ Zen Cart Documentation

2024-02-18 21_29_12-Settings

zc_install/sql/updates/mysql_upgrade_zencart_200.sql
```diff
@@ -40,6 +40,14 @@ TRUNCATE TABLE db_cache;
 
 ALTER TABLE products ADD products_mpn varchar(32) DEFAULT NULL AFTER products_model;
 
+# Product Dimensions fields
+ALTER TABLE products ADD products_length DECIMAL(8,4) DEFAULT NULL AFTER products_weight;
+ALTER TABLE products ADD products_width DECIMAL(8,4) DEFAULT NULL AFTER products_length;
+ALTER TABLE products ADD products_height DECIMAL(8,4) DEFAULT NULL AFTER products_width;
+ALTER TABLE products ADD product_ships_in_own_box TINYINT DEFAULT NULL AFTER products_height;
```
lat9, line 47: Should the upgrade script also to an `ALTER TABLE products MODIFY` on the added fields so that the database schema matches the zc200 one, in case a site has installed an add-on that previously added these fields?

piloujp: Yes, definitely!

scottcwilson:
> Couple of typos
Yep, fixed the first one, but the second one is an image, not sure where that is coming from.  Would you please post a URL, @torvista.

zc_install/sql/updates/mysql_upgrade_zencart_200.sql
piloujp: Just to be sure, is ZC installer able to manage errors provoked by `ALTER ... ADD` if column already exists?
I was doing some testing and end with something like this to avoid MySQL errors:

```
SET @test_var_length = (SELECT NOT EXISTS(select * from information_schema.columns where table_schema = schema() and table_name = 'products' and column_name = 'products_length'));
SET @a = IF(@test_var_length = 1, 'ADD', 'MODIFY');
SET @st1 = CONCAT("ALTER TABLE products ", @a, " products_length DECIMAL(8,4) DEFAULT NULL AFTER products_weight;");
PREPARE stmt1 FROM @st1;
EXECUTE stmt1;
DEALLOCATE PREPARE stmt1;

SET @test_var_width = (SELECT NOT EXISTS(select * from information_schema.columns where table_schema = schema() and table_name = 'products' and column_name = 'products_width'));
SET @b = IF(@test_var_width = 1, 'ADD', 'MODIFY');
SET @st2 = CONCAT("ALTER TABLE products ", @b, " products_width DECIMAL(8,4) DEFAULT NULL AFTER products_length;");
PREPARE stmt2 FROM @st2;
EXECUTE stmt2;
DEALLOCATE PREPARE stmt2;

SET @test_var_height = (SELECT NOT EXISTS(select * from information_schema.columns where table_schema = schema() and table_name = 'products' and column_name = 'products_height'));
SET @c = IF(@test_var_height = 1, 'ADD', 'MODIFY');
SET @st3 = CONCAT("ALTER TABLE products ", @c, " products_height DECIMAL(8,4) DEFAULT NULL AFTER products_width;");
PREPARE stmt3 FROM @st3;
EXECUTE stmt3;
DEALLOCATE PREPARE stmt3;
```

drbyte: > Just to be sure, is ZC installer able to manage errors provoked by `ALTER ... ADD` if column already exists?

Yes. If those conditions are encountered, it silently logs a status note to the `upgrade_exceptions` table and carries on.
Same with the `Install SQL Patches` tool.

torvista: I usually paste these sql mods in the Patch tool....I note this time, it outputs error messages:
image

but the new columns **are** added correctly.

admin/includes/modules/collect_info.php
```diff
@@ -438,12 +442,43 @@
       ?>
     </div>
   </div>
+  <div class="well" style="color: #31708f;background-color: #d9edf7;border-color: #bce8f1;padding: 10px 10px 0 10px;">
```
torvista, line 445: I'm a bit surprised to see all this hard-coded styling...

drbyte: > I'm a bit surprised to see all this hard-coded styling...

For consistency, it was cloned from similar `well` styling implemented earlier in the same page.
A PR is welcome to adapt both to use whatever custom Bootstrap theme is active.

admin/includes/languages/lang.english.php
```diff
@@ -583,6 +583,10 @@
     'TEXT_SALEMAKER_IMMEDIATELY' => 'Immediately',
     'TEXT_SALEMAKER_NEVER' => 'Never',
     'TEXT_SET_DEFAULT' => 'Set as default',
+    'TEXT_SHIPPING_lbs' => '(lbs)',
```
brittainmark, line 586: Should these be UPPER case
`'TEXT_SHIPPING_LBS' => '(lbs)',`
etc

admin/includes/modules/collect_info.php
piloujp, line 1: If I understand well, this file is overridden by other `collect_info.php` in different products types folder. Shouldn't these be updated too?

drbyte:
In v2.0.0 they're consolidated into a parent directory unless you're customizing them, in which case you copy the original into that product type dir. 

I guess we could add it to the built in music one.

admin/includes/modules/collect_info.php
torvista: maybe best left until after https://github.com/zencart/zencart/pull/6173

admin/includes/modules/collect_info.php
drbyte: See PR #6245

admin/includes/languages/lang.english.php
drbyte: See PR #6245

admin/includes/modules/collect_info.php
drbyte: Done in PR #6244

neekfenwick:
I just tried to monkey patch my database using the SQL added here, but it fails because there's a MODIFY on `products_length` which does not exist:

https://github.com/zencart/zencart/blob/119d28c91282c7bbb9e239bcbed152893d29e8e0/zc_install/sql/updates/mysql_upgrade_zencart_200.sql#L44-L45

In fact, why does this SQL attempt to MODIFY and then ADD the same field?  The MODIFY line appears redundant.  Same goes for the width and height lines just below.

scottcwilson:
> I just tried to monkey patch my database ...

If you're doing that, remove the ALTER TABLE MODIFY statements.  These were put in place for people who already had these fields because of mods (like Numinix Product Fields).

neekfenwick:
> If you're doing that, remove the ALTER TABLE MODIFY statements. These were put in place for people who already had these fields because of mods (like Numinix Product Fields).

Ah, there's obviously some black magic going on in zc_install that I'm unaware of 👍 Thanks.

scottcwilson:
Please note: The guidance is to not do what you're doing.   That's why it says in red text in the Install SQL Patches tool:

**This tool should NOT be used to execute Zen Cart database-upgrade scripts: use the Zen Cart Installer as per the documentation.**

If you still want to do it, you have to dig in and understand the details.

drbyte:
> Ah, there's obviously some black magic going on in zc_install that I'm unaware of 👍 Thanks.

Yes, both `zc_install` and `Install SQL Patch` will check whether add/change/modify/drop operations, and some insert operations, would fail due to required dependencies, and simply skip running those statements. It logs it to `upgrade_exceptions` table and the `logs` directory, and the admin will display the alert ... but the script will not fail.
This is why Zen Cart upgrades can do all their SQL changes with just .sql files instead of having to write PHP code specific to every version upgrade. It also means one "can" copy/paste the statements into phpmyadmin or sqlpatch in most cases without side-effects.  (The reason that "the guidance" says "not" to do that, is because uninformed end-users tend to cherrypick things and assume they did enough, and then come on the forum asking why their site isn't working after "upgrading" when they didn't follow the complete upgrade.) zc_install is still important for most upgrades because it does use PHP code to check various dependencies and fix incompatible situations; and also upgrade configure.php if it's in an old format, etc.

PR #6238, opened 2024-02-20
Convert between weight/distance units

drbyte:
Helper functions for use with shipping modules

### DIFF
--- a/includes/classes/MeasurementUnits.php
+++ b/includes/classes/MeasurementUnits.php
@@ -1,0 +1,79 @@
+<?php
+
+namespace Zencart;
+
+use TypeError;
+
+class MeasurementUnits
+{
+    /**
+     * Convert weight from one unit to another
+     */
+    public static function convertWeight(float|int $incoming, string $from_unit, string $to_unit, ?int $precision): float|int
+    {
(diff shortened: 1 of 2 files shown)


includes/classes/MeasurementUnits.php
```diff
@@ -0,0 +1,79 @@
...
+        $ratio['g']['lbs'] = .00220462262;
+
+        if ($from_unit !== $to_unit) {
+            $result = $incoming * $ratio[$from_unit][$to_unit];
+        } else {
+            // if converting "to" same as "from", then use original value and carry on to rounding
+            $result = $incoming;
+        }
+
+        if (($precision === null) && in_array($to_unit, ['lbs', 'kgs'])) {
```
lat9, line 42: Just curious, why limit this precision override to lbs and kgs?

drbyte: The feeling was that grams/ounces might need extra precision, but converting to lbs/kgs likely doesn't need 0.0000001 accuracy because they're already such larger units. 
Also, it was only with lbs/kgs that this same thing was done in another mod I'd looked at for additional input.

PR #6239, opened 2024-02-20
IntlDateFormatter - show timezone offset

### LINKED ISSUE #6229: Date in Admin displays with mixed up character between time and AM/PM instead of a space when using PHP 8.3.
A mixed up character displays between time and AM/PM instead of a space when using PHP 8.3 on Admin page date.

Zen Cart version 1.5.8a, V2.0.0-alpha1 release and last GitHub files version too, all freshly installed.
PHP version: 8.3.2
Problem does not appear on PHP 8.1 and 8.2, only on PHP 8.3
System is Windows 11
Tested with different browsers with no influence.

Capture_ecran2024-02-15

Problem is coming from the hard coded conversion from ISO-8859-1 to UTF-8 in '..\admin\includes\header.php' at line 215, even if date provided is already UTF-8. Recent Windows systems (Windows 10, 11, Server 2022 and at least for PHP binaries Server 2019) support UTF-8 and don't need this conversion... Problem already mentioned in an old discussion about zcDate class.
`        echo((strtoupper(substr(PHP_OS, 0, 3)) == 'WIN') ? iconv('ISO-8859-1', 'UTF-8', $zcDate->output(ADMIN_NAV_DATE_TIME_FORMAT, time())) : $zcDate->output(ADMIN_NAV_DATE_TIME_FORMAT, time())); //windows does not "do" UTF-8...so a manual conversion is necessary`
An easy fix would be to remove this conversion that should only be applied on Windows Server 2016 and older using English language. The conversion will probably make problems on any system using encoding other than ISO-8859-1. With non-alphabetic characters sets, date will be surely mixed up.
It is possible to improve this conversion like this too:
`        echo((strtoupper(substr(PHP_OS, 0, 3)) == 'WIN') ? mb_convert_encoding($zcDate->output(ADMIN_NAV_DATE_TIME_FORMAT, time()), 'UTF-8') : $zcDate->output(ADMIN_NAV_DATE_TIME_FORMAT, time())); //windows does not "do" UTF-8...so a manual conversion is necessary`
Like this, it is not hard coded anymore.

drbyte:
`%Z` shows `ZZZZ` equivalent, which shows the timezone offset

Fixes #6229 

Refs: 
https://www.php.net/manual/en/function.strftime.php
https://www.php.net/manual/en/datetime.formats.php
https://unicode-org.github.io/icu/userguide/format_parse/datetime/#date-field-symbol-table

### DIFF
--- a/admin/includes/header.php
+++ b/admin/includes/header.php
@@ -212,8 +212,9 @@ if (defined('MODULE_ORDER_TOTAL_GV_SHOW_QUEUE_IN_ADMIN') && MODULE_ORDER_TOTAL_G
     </div>
     <div class="col-xs-12 col-sm-12 col-md-6 col-lg-6">
         <?php
-        echo((strtoupper(substr(PHP_OS, 0, 3)) == 'WIN') ? iconv('ISO-8859-1', 'UTF-8', $zcDate->output(ADMIN_NAV_DATE_TIME_FORMAT, time())) : $zcDate->output(ADMIN_NAV_DATE_TIME_FORMAT, time())); //windows does not "do" UTF-8...so a manual conversion is necessary
-        echo '&nbsp;' . date("O", time()) . ' GMT';  // time zone
+        /** @var zcDate $zcDate */
+        $date = $zcDate->output(ADMIN_NAV_DATE_TIME_FORMAT, time());
+        echo (function_exists('mb_convert_encoding')) ? mb_convert_encoding($date, 'UTF-8') : $date;
         echo '&nbsp;[' . $_SERVER['REMOTE_ADDR'] . ']'; // current admin user's IP address
         echo '<br>';
         echo gethostname();
--- a/admin/includes/languages/lang.english.php
+++ b/admin/includes/languages/lang.english.php
@@ -9,7 +9,7 @@
 @setlocale(LC_TIME, ['en_US', 'en_US.utf8', 'en', 'English_United States.1252']);
 
 $define = [
-    'ADMIN_NAV_DATE_TIME_FORMAT' => '%A %d %b %Y %X',
+    'ADMIN_NAV_DATE_TIME_FORMAT' => '%A %d %b %Y %X (%Z)',
     'ARIA_PAGINATION_' => '',
     'ARIA_PAGINATION_CURRENTLY_ON' => ', now on page %s',
     'ARIA_PAGINATION_CURRENT_PAGE' => 'Current Page',
--- a/includes/classes/zcDate.php
+++ b/includes/classes/zcDate.php
@@ -67,6 +67,8 @@ class zcDate extends base
             '%X' => 'H:i:s',
             '%y' => 'y',
             '%Y' => 'Y',
+            '%z' => 'ZZZZ',
+            '%Z' => 'ZZZZ',
         ];
         $this->strftime2date = [
             'from' => array_keys($strftime2date),
@@ -113,6 +115,8 @@ class zcDate extends base
                 '%X' => $time_short,
                 '%y' => 'yy',
                 '%Y' => 'y',
+                '%z' => 'ZZZZ',
+                '%Z' => 'ZZZZ',
             ];
             $this->strftime2intl = [
                 'from' => array_keys($strftime2intl),


drbyte:
@piloujp does this code change sufficiently address the problem you reported, and also accommodate what's needed for the japanese lang support?

piloujp:
> @piloujp does this code change sufficiently address the problem you reported, and also accommodate what's needed for the japanese lang support?

Yes it did. 
Capture_new_date-2024-02-23 131359

piloujp:
To display modern Japanese date, I just have to change `ADMIN_NAV_DATE_TIME_FORMAT` to `'%Y年 %m月 %d日 %A %X (%z)'` or this `y年MMMMd日 EEEE HH:mm (ZZZZ)`.
Capture_New_date_jp-2024-02-23 133335
It should work in any language and system now.
The only thing I can't display is the Era name in Japanese traditional calendar because it is not just a parameter in formatting date. Calendar type must be changed from Gregorian (default) to Traditional. I do it like this:
```
function zen_set_local_calendar ($pattern = 'r年（Gy年）MMMMd日 EEEE HH:mm ZZZZ') {
	$formatterJP = datefmt_create(
				'ja_JP@calendar=japanese',
				IntlDateFormatter::FULL,
				IntlDateFormatter::FULL,
				'Asia/Tokyo',
				IntlDateFormatter::TRADITIONAL,
			);
			$formatterJP->setPattern($pattern);
	return $formatterJP;
}
```
I then call this function with `time()` as parameter (and optionally a pattern string) from `header.php`. Then year is displayed as 令和６ (Reiwa 6=2024) with parameter `Gy`.
This is really not a common problem any way, and most of the other languages will do perfectly well with new code.

drbyte:
@piloujp is there any relatively simple adjustment we could do to the `zcDate` class to accommodate the non-Gregorian calendar type?

piloujp:
> @piloujp is there any relatively simple adjustment we could do to the `zcDate` class to accommodate the non-Gregorian calendar type?

Yes, there is, by adding an optional parameter to function output. I am still testing, but it works fine for now.
Change 'zcdate' code in function output from:
```
    public function output(string $format, int $timestamp = 0)
    {
        if ($timestamp === 0) {
            $timestamp = time();
        }
.................
..................
            $converted_format = $this->convertFormat($format, $this->strftime2intl);
            $this->dateObject = datefmt_create(
                $this->locale,
                IntlDateFormatter::FULL,
                IntlDateFormatter::FULL,
                date_default_timezone_get(),
                IntlDateFormatter::GREGORIAN,
                $converted_format
            );
            $output = $this->dateObject->format($timestamp);
.........................
........................
        return $output;
    }
```
To:
```
    public function output(string $format, int $timestamp = 0**, string $calendar_type = ''**)
    {
        if ($timestamp === 0) {
            $timestamp = time();
        }
.................
..................
            $converted_format = $this->convertFormat($format, $this->strftime2intl);
            **$loc_calendar = IntlCalendar::createInstance(NULL, $calendar_type);**
            $this->dateObject = datefmt_create(
                $this->locale,
                IntlDateFormatter::FULL,
                IntlDateFormatter::FULL,
                date_default_timezone_get(),
                **$loc_calendar,**
                $converted_format
            );
            $output = $this->dateObject->format($timestamp);
.........................
........................
        return $output;
    }
```
Then change `headr.php` from:
`        $date = $zcDate->output(ADMIN_NAV_DATE_TIME_FORMAT, time());`
To:
`        $date = $zcDate->output(ADMIN_NAV_DATE_TIME_FORMAT, time(), ADMIN_NAV_DATE_CALENDAR);`

And finally add a constant in language file:
`    'ADMIN_NAV_DATE_CALENDAR' => 'ja_JP@calendar=japanese',` For Japanese (admin lang.japanese.php),
`    'ADMIN_NAV_DATE_CALENDAR' => '',` For English (admin lang.english.php),

drbyte:
@piloujp Would you need the calendar override anywhere else besides the Admin Nav Menu time-display?

piloujp:
@drbyte 
For me, it is already good enough like this. Japanese era calendar is just a shift from Gregorian calendar, everything is count the same way and any event fixed in one calendar is fixed in the other one.

But for other countries or cultures, being able to set sales related to these even would be nice. Some of these events are becoming popular, like the Chinese new year. I remember seeing somebody on the forum trying to set some sales relative to Hebrew calendar events. Islam's Ramadan is pretty important for Muslim too.
If it is not too difficult, or too much change to zcDate class, it could give Zen Cart a new potential to increase its users.

PR #6241, opened 2024-02-22
Incorporate attribute filters to product pulldowns

### LINKED ISSUE #5359: Option Values Sorter: Update product dropdown is not filtered by product with attributes
..as is named the function call.
image
Test Three has no attributes.

This enters the world of pulldown classes for which I am not insured....

drbyte:
Fixes #5359
Closes #5817


drbyte:
@torvista Does this correctly update both dropdowns?

torvista:
I believe so!

PR #6243, opened 2024-02-23
Move coupon referrers to separate db table

drbyte:
Due to catalog-side query lookups that cannot use an index, this approach may be more performant.

Ref #6193
Ref #6211

### DIFF
--- a/admin/coupon_admin.php
+++ b/admin/coupon_admin.php
@@ -286,7 +286,6 @@ switch ($_GET['action']) {
       'coupon_type' => zen_db_prepare_input($coupon_type),
       'uses_per_coupon' => (int)$_POST['coupon_uses_coupon'],
       'uses_per_user' => (int)$_POST['coupon_uses_user'],
-      'referrer' => $_POST['coupon_referrer'] ?: 'null',
       'coupon_minimum_order' => (float)$_POST['coupon_min_order'],
       'restrict_to_products' => zen_db_prepare_input($_POST['coupon_products']),
       'restrict_to_categories' => zen_db_prepare_input($_POST['coupon_categories']),
@@ -318,6 +317,17 @@ switch ($_GET['action']) {
         ];
         zen_db_perform(TABLE_COUPONS_DESCRIPTION, $sql_data_desc_array, 'update', "coupon_id = " . (int)$_GET['cid'] . " and language_id = " . (int)$languages[$i]['id']);
       }
+      // referrers
+        $trimmed_referrers = array_map(static fn($referrer) => trim($referrer), explode(',', $_POST['coupon_referrer'] ?? []));
+        foreach ($trimmed_referrers as $referrer) {
+            if (empty(CouponValidation::referrer_already_assigned($referrer))) {
+                $sql_data_array = [
+                    'referrer_domain' => $referrer,
+                    'coupon_id' => (int)$_GET['cid'],
+                ];
+                zen_db_perform(TABLE_COUPON_REFERRERS, $sql_data_array);
+            }
+        }
     } else {
       zen_db_perform(TABLE_COUPONS, $sql_data_array);
       $cid = $db->insert_ID();
@@ -329,6 +339,17 @@ switch ($_GET['action']) {
         $sql_data_marray[$i]['language_id'] = (int)$language_id;
         zen_db_perform(TABLE_COUPONS_DESCRIPTION, $sql_data_marray[$i]);
       }
+      // referrers
+        $trimmed_referrers = array_map(static fn($referrer) => trim($referrer), explode(',', $_POST['coupon_referrer'] ?? []));
+        foreach ($trimmed_referrers as $referrer) {
+            if (empty(CouponValidation::referrer_already_assigned($referrer))) {
+                $sql_data_array = [
+                    'referrer_domain' => $referrer,
+                    'coupon_id' => (int)$_GET['cid'],
+                ];
+                zen_db_perform(TABLE_COUPON_REFERRERS, $sql_data_array);
+            }
+        }
     }
     zen_redirect(zen_href_link(FILENAME_COUPON_ADMIN, 'cid=' . $_GET['cid'] . (isset($_GET['status']) ? '&status=' . $_GET['status'] : '') . (isset($_GET['page']) ? '&page=' . $_GET['page'] : '')));
 }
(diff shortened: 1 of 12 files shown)


drbyte:
@neekfenwick What do you think of this approach?

neekfenwick:
@drbyte It would certainly be more better in terms of DB design.  What files would need modifying in the install/upgrade SQL files? I have minimal experience in that area, I missed out one of them and you had to patch it up.  If we're creating a new table with indexes, do we write duplicate SQL in both the install & upgrade files?

I guess since this is such a new feature there won't be existing data in the single `coupons.referrer` field that needs to be migrated into the new `coupons_referrers` table, but at least an old 1.5.x database would need the new table creating.

drbyte:
I've already done all of the SQL etc in this PR. You can see some of the answers to your questions in the committed code.

Actually the only thing this PR doesn't cover is deciding whether to keep the UI you built in coupon_admin for adding referrers, or to just delete that code and let the new `coupon_referrers` page (available on the menu) be the only place to handle that ... instead of being in both places.  Right now the coupon_admin page doesn't let you remove referrers if already in the database (it only adds new ones if not found). But the coupon_referrers page lets you add/remove/edit/reassign.

zc_install/sql/updates/mysql_upgrade_zencart_200.sql
```diff
@@ -76,7 +76,22 @@ ALTER TABLE products_attributes ADD options_values_price_w varchar(150) NOT NULL
...
+CREATE TABLE coupon_referrers (
+  referrer_id int(11) NOT NULL AUTO_INCREMENT,
+  referrer_domain varchar(64) NOT NULL,
+  coupon_id INT(11) NOT NULL,
+  date_added timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY  (referrer_id),
+  UNIQUE KEY idx_referrer_domain_zen (referrer_domain),
+  KEY idx_refcoupon_id_zen (coupon_id)
+) ENGINE=MyISAM;
```
lat9, line 91: Why is MyISAM being forced here?

admin/coupon_admin.php
```diff
@@ -318,6 +317,17 @@
         ];
         zen_db_perform(TABLE_COUPONS_DESCRIPTION, $sql_data_desc_array, 'update', "coupon_id = " . (int)$_GET['cid'] . " and language_id = " . (int)$languages[$i]['id']);
       }
+      // referrers
+      $trimmed_referrers = array_map(fn ($referrer) => trim($referrer), explode(',', $_POST['coupon_referrer'] ?? []));
```
lat9, line 321: PSR-12, 4.7: "...  there MUST NOT be a space between the method or function name and the opening parenthesis ..."

admin/coupon_admin.php
```diff
@@ -329,6 +339,17 @@
         $sql_data_marray[$i]['language_id'] = (int)$language_id;
         zen_db_perform(TABLE_COUPONS_DESCRIPTION, $sql_data_marray[$i]);
       }
+      // referrers
+      $trimmed_referrers = array_map(fn ($referrer) => trim($referrer), explode(',', $_POST['coupon_referrer'] ?? []));
```
lat9, line 343: As above.

lat9 (changes requested):

zc_install/sql/updates/mysql_upgrade_zencart_200.sql
drbyte: Artifact from copy/paste. Will remove the specific `ENGINE`

admin/coupon_admin.php
drbyte: There aren't any.

admin/coupon_admin.php
drbyte: ibid.

admin/coupon_admin.php
drbyte: Oh you meant between the `fn` keyword and the function parameter.

lat9 (approved):

drbyte:
TO DO as another PR (up for grabs to whomever wants it): 
- deciding whether to keep the widget UI @neekfenwick built in coupon_admin for adding referrers, or to just delete that code and let the new coupon_referrers page (available on the menu) be the only place to handle that ... instead of being in both places. 
- Right now the coupon_admin page doesn't let you actually remove referrers if already in the database (it only adds new ones if not found). You "can" delete them on-screen, BUT THE DELETE DOES NOT OCCUR IN THE DB.
- But the coupon_referrers page lets you add/remove/edit/reassign. Deletes do happen correctly there.

neekfenwick:
> * deciding whether to keep the widget UI @neekfenwick built in coupon_admin for adding referrers, or to just delete that code and let the new coupon_referrers page (available on the menu) be the only place to handle that ... instead of being in both places.

I was unaware of a new coupon_referrers page.  Why was this built if I'd handled the management of referrers in coupon_admin?

> * Right now the coupon_admin page doesn't let you actually remove referrers if already in the database (it only adds new ones if not found). You "can" delete them on-screen, BUT THE DELETE DOES NOT OCCUR IN THE DB.

This sounds impossible, the list of values provided on the UI is put together as a CSV string and overwrites the current `coupon.referrer` value in the database.  If one was removed, the string written to the DB doesn't include that referrer, i.e. it is deleted.

> * But the coupon_referrers page lets you add/remove/edit/reassign. Deletes do happen correctly there.

With this action going on in coupon_referrers, wiping out a bunch of work I've put into coupon_admin, I don't feel authoratative on any of this any more :) Do what you feel is best.

drbyte:
> Why was this built if I'd handled the management of referrers in coupon_admin?

@neekfenwick Apologies for the confusion.

The problem is that the db queries doing lookups in init_coupon on every page-visit were forced to use `LIKE %foo%` syntax which makes then non-indexable, requiring complete database table-scans. On a busy site that's a lot of slow queries. And on sites with a lot of coupons, the table-scans would take longer.
https://github.com/zencart/zencart/issues/6193

Reworking it to store the referrers in their own table fixes that problem, because the query can be an exact-match, making the lookup way more performant.

So, that's the problem that this was meant to solve. Sorry for not writing that out in the opening post; it was hidden behind the related issue link.

I didn't want to delete all the work you'd done already because I like what you did, so I merely adapted it, but didn't come up with a deletion strategy in that UI ... at the time. As for handling those deletes, I think I've got something that'll work. Renames will be a combination of delete-old-and-add-the-new.  Since they're always only related to a single coupon it's less complicated than it would be otherwise.

neekfenwick:
> I didn't want to delete all the work you'd done already because I like what you did, so I merely adapted it, but didn't come up with a deletion strategy in that UI ... at the time. As for handling those deletes, I think I've got something that'll work. Renames will be a combination of delete-old-and-add-the-new. Since they're always only related to a single coupon it's less complicated than it would be otherwise.

I thought my little UI I'd built on `coupon_admin` would do the whole job.  If your concern with deletes is figuring out which DB entity to delete, I'd use a bit of DB design that solves the problem since our data model is extremely simple.

```
CREATE TABLE coupon_referrers (
  coupon_id INT,
  referrer MEDIUMTEXT,
  PRIMARY KEY (coupon_id, referrer)
);
```

With the composite primary key you don't need a unique integer id for each row such as `coupon_referrers_id`.  When a coupon is saved on `coupon_admin` you can delete all entities in `coupon_referrers` with that `coupon_id` and insert new rows for each referrer present on the admin UI. So, you don't have to figure out which existing referrers have been added or deleted, you just stick them all in the table en mass.

I can't remember exactly how indexes work, you may need a primary index on (coupon_id, referrer) and another one just on (referrer) so that the `init_coupons` lookup works, or if that lookup could use the primary index then you'd just need the one.

I haven't looked at the new `admin_coupon_referrers` admin page at all, so I may be wrong, but it sounds rather redundant to me.

Does that help or am I barking up the wrong tree?

drbyte:
> With the composite primary key you don't need a unique integer id for each row such as coupon_referrers_id. When a coupon is saved on coupon_admin you can delete all entities in coupon_referrers with that coupon_id and insert new rows for each referrer present on the admin UI. So, you don't have to figure out which existing referrers have been added or deleted, you just stick them all in the table en mass.

Excellent! I'll have a go at adjusting things.

neekfenwick:
> > Why was this built if I'd handled the management of referrers in coupon_admin?
> 
> @neekfenwick Apologies for the confusion.
> 
> The problem is that the db queries doing lookups in init_coupon on every page-visit were forced to use `LIKE %foo%` syntax which makes then non-indexable, requiring complete database table-scans. On a busy site that's a lot of slow queries. And on sites with a lot of coupons, the table-scans would take longer. #6193

Maybe I've got the wrong end of the stick, then.  I thought you'd built a totally separate `admin_coupon_referrers` page, on which a list of referrers is handled instead of handling their UI on the existing `admin_coupon` page.  If we're keeping a single coupons admin page, and just improving the DB design, then it's all gravy.  Sorry if I got confused.

Should be fairly (!) trivial since I've already written the various bits of code that deal with the single `coupon.referrer` field, just take that out and replace it with interaction with the new `coupon_referrers` table.

PR #6254, opened 2024-02-25
zcDate->output() - allow calendar override

drbyte:
Ref: https://github.com/zencart/zencart/pull/6239#issuecomment-1962285093
Ref: #6229 

@piloujp 

If someone wants to override the calendar used with `$zcDate->output()`, they can pass the desired calendar-related locale.

### DIFF
--- a/includes/classes/zcDate.php
+++ b/includes/classes/zcDate.php
@@ -142,10 +142,11 @@ class zcDate extends base
     /**
      * @param string $format  output method should start with a strftime-format string
      * @param int    $timestamp
+     * @param string|null $calendar_locale Optional calendar-related locale. eg: 'ja_JP@calendar=japanese'
      *
      * @return false|string
      */
-    public function output(string $format, int $timestamp = 0)
+    public function output(string $format, int $timestamp = 0, ?string $calendar_locale = null)
     {
         if ($timestamp === 0) {
             $timestamp = time();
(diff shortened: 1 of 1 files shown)


drbyte:
> But doing so makes a permanent change, whatever the language is set to...

... and you WOULD want that for any specific store, right? 

Example: For your Japanese pack, you're already hard-coding changes based on locale, etc, and the entire site is using that locale, correct?

drbyte:
Okay, good points.

I think there's a workaround:
If you call this with `null`, it will use PHP's fallback to configured locale, instead of defaulting to Gregorian:
`$zcDate->setCalendarType(null);`

... but you have to call it after the language subsystem is loaded (or, at least after the locale is set when language files load), so you could probably test it by putting `$zcDate->setCalendarType(null);` into a file in the language's `extra_definitions` folder.

Does that allow it to work as expected?  Or is this something that truly needs to be called "every" time someone calls `$zcDate->output()`? (there are a lot of places where ->output() is used)

piloujp:
> ... but you have to call it after the language subsystem is loaded (or, at least after the locale is set when language files load), so you could probably test it by putting `$zcDate->setCalendarType(null);` into a file in the language's `extra_definitions` folder.

@drbyte 
I got 'Undefined variable $zcDate' when I put it in language file.
`null` parameter is generally same as `Gregorian` except if calendar has been set differently in php.ini.

piloujp:
@drbyte 
The more I look at it, the less I can see a good solution. Global calendar change, even language switchable, is not good. There are 17 files in ZC that use `$zcDate->output` (36 occurrences) according to developer tool. Some of them should not use traditional calendars like in payment modules. Others that just need formatting for calculations should not either...
Only dates displayed should be able to use different calendars, not 'timestamp type' dates used for export, saving or calculations.

drbyte:
Okay. I've reworked it to allow setting the calendar during calls to `->output()` instead. That way if one wants the override for a certain situation, they can use it there.

PR #6258, opened 2024-02-26
Coupon admin: referrer-help missing examples

lat9:
The referrer-help popover in the coupon admin is currently missing its examples:
image

This update to the associated language file provides that correction:
image


neekfenwick:
Thank you for this, however isn't this obviated by the newer UI that I added, in which I forgot to update the help text?  I believe that was merged.

image

It's all this gubbins (plus javascript support).
https://github.com/zencart/zencart/blob/82af1e1a04433941eea799416c2ebce271b23f51/admin/coupon_admin.php#L1010-L1029

Would you like me to make a PR with suitable help text?

Interesting that the fix here was to introduce HTML to the help text constant, and not html-escape the value where it is used at https://github.com/zencart/zencart/blob/82af1e1a04433941eea799416c2ebce271b23f51/admin/coupon_admin.php#L1010

I guess it doesn't matter a great deal and would be less readable with a function call around the use of a constant.

drbyte:
> Would you like me to make a PR with suitable help text?

Yes please.

neekfenwick:
> > Would you like me to make a PR with suitable help text?
> 
> Yes please.

Aye aye, cap'n, please see PR #6265.

PR #6262, opened 2024-02-28
Add option to copy specials to copied product.

neekfenwick:
Fixes Discussion #6257 

Add option to copy specials to copied product.
and...
Visibility of Copy To panels depend on radio buttons.

When the Link or Duplicate radio buttons are changed, show/hide the panels with link-only and duplicate-only classes by adding/removing the hiddenField class.
Would like to use an expando attribute like data-link-only="true" instead of non-functional classes link-only/duplicate-only, but infoBox does not allow this.


admin/includes/modules/copy_product.php
```diff
@@ -49,7 +51,18 @@
     $contents[] = array(
         'text' => '<h6>' . TEXT_COPY_ATTRIBUTES . '</h6>' .
             '<div class="radio"><label>' . zen_draw_radio_field('copy_attributes', 'copy_attributes_yes', true) . TEXT_YES . '</label></div>' .
-            '<div class="radio"><label>' . zen_draw_radio_field('copy_attributes', 'copy_attributes_no') . TEXT_NO . '</label></div>'
+            '<div class="radio"><label>' . zen_draw_radio_field('copy_attributes', 'copy_attributes_no') . TEXT_NO . '</label></div>',
+            'params' => 'infoBoxContent duplicate-only hiddenField'
```
proseLA, line 55: i will only comment on formatting; and this file could use some work.  short array tags would be great; but it also seems that added array element with key `params` is indented wrong.  it could also benefit from a comma at the end of the value of said element.  i would say the same for line 86 below.

includes/functions/functions_products.php
```diff
@@ -835,6 +835,21 @@ function zen_has_product_discounts($product_id)
...
+ * @param int $product_id
+ * @return boolean
+ */
+function zen_has_product_specials(int $product_id): bool
+{
+    global $db;
+    $result = $db->Execute('SELECT COUNT(*) AS cc
+        FROM ' . TABLE_SPECIALS . "
+        WHERE products_id = $product_id;");
+    return (int)$result->fields['cc'] !== 0;
```
proseLA, line 850: i can not comment on speed of execution, but it seems from a readability standpoint, `SELECT products_id...` and then checking for `$result->EOF` is far more legible to me than casting something to an integer and checking it's value against 0.  even when writing this comment i had to [review this link.](https://www.php.net/manual/en/language.operators.precedence.php)

neekfenwick: It's a `COUNT(*)` operation, which is more efficient in the DB than requesting actual fields from the table, and returns an integer value.  If I don't cast it to int, it will be implicitly cast anyway due to the `==` comparison operator.

admin/includes/modules/copy_product.php
neekfenwick: You're right on the indentation.  Not sure about the array syntax since the rest of the file uses the old format.  Should I convert the whole file to use `[]`? Not sure what the zencart consensus is on this.

neekfenwick: On the trailing comma, it's always been my style not to use one from my early C days, but I can certainly add one if that's the style we prefer.  I assume it's for clean diffs when lines are added after or reordered.

neekfenwick: I have force pushed formatting fixes.  Ran the psr-12 "format document" over it, some other lines have changed as a result. Not sure what github does when you force push on a PR, I read that it will "handle it" but YMMV.

includes/functions/functions_products.php
neekfenwick: I just noticed, I do it the implicit cast way further down the file around line 1006 😊
I hope you got the point about why `EOF` is not useful to check, since the `COUNT(*)` operation will always return one result row, and `EOF` will always be `false`.

proseLA: i think you have missed my point.  i fully understand that using a `count` function will always return 1 result row.  i was suggesting the following:
```
SELECT products_id 
        FROM ' . TABLE_SPECIALS . "
        WHERE products_id = $product_id;"
```
and now `EOF` will return either `true` or `false`.  

even taking your assertion that your method is faster (which it very well might be) as gospel, i would argue sacrificing that speed difference is worth it for enhanced code legibility.

with regards to strict comparisons, i 'think' SOP for PRs is to now use strict comparisons.  but, alas, i have found the rules with regards to code format and the like are very loosely defined in this repo, and selectively enforced.

lat9: But why do a `COUNT(*)` in the first place?  Why not just do a query for a single instance of the `$product_id` in the `specials` table and then, as @proseLA indicated, check for `$result->EOF` to see if the product has one or more associated specials?

neekfenwick: @lat9 because the question I'm asking the database is "do any rows exist that match my criteria" and I think the most direct operation that matches that intent is `COUNT`.  I don't query for a `$product_id` because I do not want to use the `products_id` value that is returned.

neekfenwick: > i think you have missed my point. i fully understand that using a `count` function will always return 1 result row. i was suggesting the following:

OK, I took your first question literally, you asked why I didn't test `EOF` and it was because the query I was using didn't make `EOF` a sensible thing to test, in my opinion.

While there is little eventual difference between `COUNT` and testing the result and `SELECT` and testing the result, the intent of the code is different, and I do think a `COUNT` approach is more readable, in that its intent is harder to misinterpret.  If you see a `COUNT`, you know that the objective is not to retrieve value data such as a `products_id` and the reader's mindset is different to if you see a `SELECT` and then look for where the returned values are inspected.  In this case, to do a `SELECT` and then immediately test if the resultset iterator is exhausted, by testing `EOF`, seems a misuse of the technology.

> even taking your assertion that your method is faster (which it very well might be) as gospel, i would argue sacrificing that speed difference is worth it for enhanced code legibility.

I'm not saying it's faster in that we'll notice the difference, but I believe it's slightly more efficient and in my opinion cleaner code.  The speed difference is neglible in this case but it's still hitting an index (on `products_id`) so in terms of utilising database resources it's not worse.

> with regards to strict comparisons, i 'think' SOP for PRs is to now use strict comparisons. but, alas, i have found the rules with regards to code format and the like are very loosely defined in this repo, and selectively enforced.

Yeah I should have used a strict comparison in the latter case I reference, which I will force push again to fix.  I stick by the general pattern that a `SELECT COUNT` and testing the result is a better way to solve the question of "how many rows match my criteria" rather than saying "SELECT products_id" and seeing if I immediately cannot iterate results.

proseLA: > OK, I took your first question literally, you asked why I didn't test `EOF` and it was because the query I was using didn't make `EOF` a sensible thing to test, in my opinion.
> 
i would suggest you reread what i initially wrote.  i started by stated rewriting the sql query;  and then testing `EOF`.  apparently @lat9 understood what i wrote.  so if you want to continue telling me what i did, feel free...  but it is right there for anyone to read.

> While there is little eventual difference between `COUNT` and testing the result and `SELECT` and testing the result, the intent of the code is different, and I do think a `COUNT` approach is more readable, in that its intent is harder to misinterpret. If you see a `COUNT`, you know that the objective is not to retrieve value data such as a `products_id` and the reader's mindset is different to if you see a `SELECT` and then look for where the returned values are inspected. In this case, to do a `SELECT` and then immediately test if the resultset iterator is exhausted, by testing `EOF`, seems a misuse of the technology.

the intent of the code?  misuse of the technology?  

here is the function signature:
```
function zen_has_product_specials(int $product_id): bool
```
i think the signature says the intent of the code pretty succinctly.

you think your way is more readable; i disagree.  big deal.  code it how you want...

neekfenwick: We do seem to run into arguments about meaning quite a lot :)

neekfenwick: When I talked about 'intent', I'm referring to how any code is written with a certain intent in mind.  One can examine what the code is designed to do, and how close the implementation is to that design.  A reader of the code will attempt to build a picture in their mind of what the code is trying to do, based on the language features it uses and the general approach taken.  I was partly trying to draw the distinction between the intent of code like this:

    $result = $db->Execute('SELECT products_id FROM specials WHERE <some criteria>;');

and

    $result = $db->Execute('SELECT COUNT(*) AS cc FROM specials WHERE <some criteria>;');

The first example fetches data from the database, in the `products_id` field of the resultset.  When I see a SELECT that fetches columns from a table, it is likely that the code below that statement will inspect the resultset for those fields, so the implicit intent is that the fields are going to be used by the following code.  In this case it's all rather redundant because the WHERE clauses actually says `products_id = <some value>`, so it's guaranteed that if a row is returned then the `products_id` value in it is predictable and one does not need to inspect it.

The second example clearly counts the number of rows found by the WHERE clause, no table data is retrieved.  So the implicit intent is that we do not actually require any data from the table, and the following code is not going to alter its behaviour based on data from the table, just the result of the COUNT.

It's a minor point, I know, and rather pedantic; not like this is super important.

As for misuse of the tech, maybe I was a bit flowery with my language.  I meant that when you fetch a resultset of table data and then do not actually use those values, only the metadata on the resultset (`EOF`), one may wonder what the point of getting the data was.  You've used SELECT, but not used the values that were selected.  On the other hand, if you use COUNT and then inspect the counted value that is returned, you're directly using the results of the COUNT, which seems proper usage.

neekfenwick: @lat9 ok, I get your point, if you want to know "do you have one special" and you ask "what's the id" and the answer is "none", yes, the answer satisfies the question.  You may think, job done.  Similarly, if you want to know "do you have one special" and you ask "how many specials do you have" (which is my `COUNT` approach) and the answer is zero, the answer also satisfies the question.  Also, job done.

As I said before, the difference between the two approaches isn't that that it's impossible one way or the other.  Of course it's possible.  We aren't arguing about that, or I hope we aren't, I hope the conversation has moved on to the points I made quite early on in this conversation about the best way to solve this problem.  Both you @lat9 and @proseLA have pointed out that it's possible to solve the problem using `SELECT`, and I've tried to argue that it's better to solve it using `COUNT`.

I'm a bit concerned at this point, that people with such an RDBMS heritage are still arguing about whether it's possible to answer the question using the `SELECT` method, as if if it's possible ends the debate.  I tried to move beyond that quite a few comments ago.

neekfenwick: > i can not comment on speed of execution, but it seems from a readability standpoint, `SELECT products_id...` and then checking for `$result->EOF` is far more legible to me than casting something to an integer and checking it's value against 0. even when writing this comment i had to [review this link.](https://www.php.net/manual/en/language.operators.precedence.php)

You're right, my response to this comment earlier was wrong. I spoke as if you were asking why I tested `EOF`, whereas you actually said that id I had `SELECT products_id...` and then tested `EOF.  My apologies :)

lat9: I agree that either approach "gets the job done".  For the `specials` table, since it _tends_ not to be humongous, there is most likely not a performance impact using `COUNT` vs. `SELECT`.  I'd be pushing harder on the use of `SELECT/LIMIT 1` for any query on the `customers`, `products` or `orders` type tables which can be huge.

neekfenwick: By the way, I ran into discussion online that `COUNT(*)` would necessitate examining all rows in a table, in order to count them, however in a query against an indexed column it seems that only 1 row is examined.  I wondered if it would be better to rewrite as `SELECT products_id .. LIMIT 1` to ensure the database stops trying to count rows after at least one match is found.  The explain plan for the following queries on my `specials` table with over 10K rows (excuse the lazy formatting):

    explain select COUNT(*) AS cc FROM specials WHERE products_id = 1234;

    id	select_type	`table`	`type`	possible_keys	`key`	key_len	`ref`	`rows`	Extra
    1	SIMPLE	specials	ref	idx_products_id_zen	idx_products_id_zen	4	const	1	Using index

    explain select EXISTS (SELECT products_id FROM specials WHERE products_id = 1234);

    id	select_type	`table`	`type`	possible_keys	`key`	key_len	`ref`	`rows`	Extra
    1	PRIMARY								No tables used
    2	SUBQUERY	specials	ref	idx_products_id_zen	idx_products_id_zen	4	const	1	Using index

    explain select products_id FROM specials WHERE products_id = 1234 LIMIT 1;

    id	select_type	`table`	`type`	possible_keys	`key`	key_len	`ref`	`rows`	Extra
    1	SIMPLE	specials	ref	idx_products_id_zen	idx_products_id_zen	4	const	1	Using index

So, only 1 row is examined both for the `COUNT(*)` and for the `WHERE products_id = 1234 LIMIT 1`.  I don't have in depth knowledge of exactly how indexes work and found it rather surprising that the `COUNT` approach only examined one row.  Perhaps in other cases where the BTREE is arranged differently it would not be so.

drbyte: > So, only 1 row is examined both for the COUNT(*) and for the WHERE products_id = 1234 LIMIT 1. I don't have in depth knowledge of exactly how indexes work and found it rather surprising that the COUNT approach only examined one row. Perhaps in other cases where the BTREE is arranged differently it would not be so.

I'm guessing that since it's using the index `idx_products_id_zen` (which is on `products_id`, which is the WHERE criteria), then it jumps directly to the first match, and then finds that the next record is not a match, so returns (and reports only on) one row.

The non-complex query hitting a non-complex index, combined with having only a single matching record in the db, is why the results are the same. They will probably vary if multiple matching WHERE records exist in the db, but as long as it's using a performant index for that column it will be almost negligible to use either COUNT or LIMIT 1.

I'm inclined to use LIMIT 1 myself, when I only care about the result being "zero" vs "one or more exist". Obviously if I actually want to know how many without retrieving any actual record details then COUNT is the best way.

Aside: I find that in a lot of cases we're actually using PHP to loop over a bunch of data, and in such cases doing the actual SELECT and asking the QueryFactoryResults to tell the "COUNT" (of records it got) gets it all done in one shot, because it knows the count already. It also allows cached results to be reused from RAM when an identical query is called before the cache expires.

I note that the function located above this new one, `zen_has_product_discounts()` does the `SELECT products_id ... LIMIT 1` (where the `1` is the 2nd parameter to `$db->Execute($sql, $limit)`).

We could go around in circles about it. 
In the case of these admin-only activities, saving a microsecond is moot because you're not having thousands of admins doing these operations simultaneously, unlike catalog-side where query optimization is far more critical.
Plus, this is only a simple lookup, not a complex loop of lookups-and-writes.

neekfenwick: Thank you @drbyte you make very good points.  Looking back, I was a bit silly to question why queries using a single indexed key would perform badly in pretty much any case, let alone this specific one where this is the case.

I'd be happy to change the code to use the `LIMIT 1` approach if it sits better with the zencart codebase, having taken into account that in this case, it's not actually a performance benefit either way.

Late here now, and a Friday, give me a minute.

drbyte: This whole discussion arose because of enforcing strict-types in `(int)$result->fields['cc'] !== 0`.

For this exact reason, when I'm only concerned about zero-vs-non-zero I'll use `empty()` (or `!empty()`) rather than casting ... because `empty()` handles 0, false, null, missing dependencies, etc ... and in the majority of cases it's safe to check this way; there are of course exceptions.

Sometimes, if the context dictates that checking for ` > 0` would make it more readable then I'll take that approach if there's no risk of a dependent array element being null where we'd need to check both for the null and then also the non-zero value.

As to COUNT vs LIMIT 1, for the sake of consistency with most other parts of the ZC codebase, let's go with `LIMIT 1`. 
The legacy `->EOF` or `->RecordCount()` approaches are the most common ways to check the result. 
(Aside: Since QueryFactoryResult is `countable` and `iterable`, you could also `count($queryResults)` (which is why we can `foreach()` through the results), although the most common approach is `->EOF` or `->RecordCount()`.)

drbyte (approved):

neekfenwick:
Thank you for your help @proseLA and others, I hope the discussion around code style was useful.  I'll try to stick with the approach of selecting data and testing empty resultset instead of counting rows.

torvista:
Should not this option also be hidden for a linked product?
2024-03-08 09_50_16-Admin Category Product Listing — Mozilla Firefox

PR #6265, opened 2024-02-28
Fix coupon referrer tooltip help text for new UI.

neekfenwick:
Goes with PR #6258, update help text for new UI.


drbyte:
Thanks!

neekfenwick:
> Should extra rows be allowed when the first row is empty?

Yeah, good point, the UI is a little weird, since we lack client-side validation etc. and I didn't want to spend too long implementing it as it's fairly niche, and may be thrown out if the general approach wasn't accepted.

I feel the lazy approach would be to allow blank textboxes, if the user chooses to do that, and as the values are collected on the server side discard any that are blank (or contain only whitespace, you'd `trim()` each one).

We could certainly block an 'add' attempt if any of the existing textboxes are empty.

PR #6283, opened 2024-03-04
Retire old netscape/IE patches from country pulldown function

drbyte:


drbyte:
Just tidying. 

Doesn't have to go into 2.0.0.

PR #6284, opened 2024-03-04
Hide empty zone field when appropriate, in setup wizard

### LINKED ISSUE #6282: Setup Wizard form displays an empty drop down field for store zone
**Describe the bug**
After ZC install and at first login to admin, setup wizard shows a form with empty drop down field for store zone when selecting a country with no zone.
FormWemptyField

**Version**
Zen Cart version: Last Github version (2.0.0-beta1) and probably older ones
PHP version: 8.3.3 and probably others
MySQL version: 8.2.0

**To Reproduce**
Choose a country with no zone in setup wizard form.

**Expected behavior**
An input text field or no field at all.
This behavior is controlled by JavaScript and I don't think it is possible to replace the dropdown field created with PHP/HTML by a text input field using only JavaScript.
But this field can be hidden when empty using this code on line 103:
```
		  // hide the field and its label if no zone
		  if (theForm.zone_id.options.length == 0)
		      $("#zone_id, label[for=zone_id]").hide();
		  else
                      $("#zone_id, label[for=zone_id]").show();
```
FormWhiddenField

 - Operating System: Windows 11
 - Browser: Firefox 123.0

drbyte:
Fixes #6282

### DIFF
--- a/admin/index_setup_wizard.php
+++ b/admin/index_setup_wizard.php
@@ -83,17 +83,19 @@ $zone_string = zen_draw_pull_down_menu('zone_id', zen_get_country_zones($store_c
     <script>
       function update_zone(theForm) {
           // if there is no zone_id field to update, or if it is hidden from display, then exit performing no updates
-          if (!theForm || !theForm.elements["zone_id"])
+          if (!theForm || !theForm.elements["zone_id"]) {
               return;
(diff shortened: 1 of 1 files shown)


drbyte:
Not urgent. 
UX improvement.

Doesn't have to go into 2.0.0

PR #6290, opened 2024-03-05
Add $tplSetting object for accessing Template Settings

drbyte:
(This essentially makes the `tpl()` function unnecessary, so it is deleted here.)

The global var `$tplSetting` added here (can choose a different name before merge) is a (class) object that (initially) loads all its properties from the global `$tpl_settings[]` array found in `template_settings.php`. 
(Other "soft settings" global vars in that file are unrelated, and can happily be there without conflict.)

Example:
(These are hypothetical, illustrative only; I'm not suggesting that these actual settings would commonly be used.) `template_settings.php` contains:
```
$tpl_settings['SET_COLUMN_LEFT_LAYOUT'] = '2';
$tpl_settings['SET_COLUMN_RIGHT_LAYOUT'] = [
    'value' => '2',
    'type' => 'int',
    ];
$tpl_settings['BS4_AJAX_SEARCH_ENABLE'] = [
    'value' => 'true',
    'type' => 'string',
    ];
$tpl_settings['header_bg_color'] = '#F0F0F0';
```

In `tpl_main_page.php` it could be used thus:
`$tplSetting->SET_COLUMN_LEFT_LAYOUT` (returns string value '2' , straight passthru) 
`$tplSetting->SET_COLUMN_RIGHT_LAYOUT` (returns integer value 2, due to 'type' cast) 
`$tplSetting->BS4_AJAX_SEARCH_ENABLE` (returns string 'true') 
`$tplSetting->header_bg_color` (returns string '#F0F0F0')

Additionally, one could subsequently override a setting by setting it to a new value: 
`$tplSetting->header_bg_color = '#C3C3C3';`
and/or optionally casting it to a new Type:
`$tplSetting->setType('header_bg_color', 'array');`  (why you'd want an array for this, I dunno, again, example only) So now referencing it reports the changes:
`echo $tplSetting->header_bg_color`  (returns: `[0 => "#C3C3C3"]` )

For a db constant that has no entry in `$tpl_settings` array or in the `$tplSetting` object: 
`echo $tplSetting->ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN;` gives string `'false'` because it's stored as string in configuration table. 
but call this first:
`$tplSetting->setType('ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN', 'bool');` 
and now 
`echo $tplSetting->ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN;` gives boolean `false`

### DIFF
--- a/includes/classes/Settings.php
+++ b/includes/classes/Settings.php
@@ -1,0 +1,194 @@
+<?php
+
+/**
+ * @copyright Copyright 2003-2024 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id:  New in v2.0.0 $
+ */
+
+abstract class Settings implements ArrayAccess, Countable
+{
+    /**
+     * If $this->settings[$key] not found, look for a defined($key) constant.
+     */
+    protected bool $includeConstants = false;
+
+    /**
+     * Stores all the "set" settings
+     */
+    protected array $settings = [];
+
+    /**
+     * Tracks requested type-casting instructions
+     */
+    protected array $types = [];
+
+    public function __construct(array $settings = [])
+    {
+        $this->setFromArray($settings);
+    }
+
+    /**
+     * Set multiple settings via an array.
+     * The array can be entries of $key=>$value and/or $key=>['value'=>$value, 'type'=>$type]
+     * If $overwrite is false then existing $keys will be ignored.
+     *
+     * @param array|null $settings_array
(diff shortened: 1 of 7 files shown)


drbyte:
Ideally, yes, but let's start on a custom template and see what issues one runs into, if any.

torvista:
Previously I've been using the template_init to set these values with queries, but of course these need to be duplicated in each template_init if you want to swap around when testing and have different values per template.
I just replaced the above today, and will do the rest presently.

PR #6301, opened 2024-03-08
copy-to: also hide category/discount/edit options

drbyte:
Per @torvista comment at https://github.com/zencart/zencart/pull/6262#issuecomment-1985295632

References #6257
Updates #6262

UI now does this: (depends on whether linked categories, discounts, attributes, specials exist)

Screen Shot 2024-03-08 at 11 45 15 AM
Screen Shot 2024-03-08 at 11 45 28 AM


drbyte:
You're right. It actually already does: I just didn't have any defined when I took the screenshot. Here's an update:
Screen Shot 2024-03-08 at 12 43 29 PM

torvista:
Ah, I thought **I'd** added it!
All good then!

PR #6305, opened 2024-03-09
Make links use underline decoration

### LINKED ISSUE #6304: Product info page (and others) in responsive_classic: Links rely on color to be distinguishable.
Simply using link decoration underline resolves this.

scottcwilson:
Fixes #6304


lat9 (commented):
Just note that browsers' _default_ styling for a link is with an underline (https://www.w3schools.com/html/html_links.asp).  Couldn't all those `text-decoration:underline` elements just be removed?

scottcwilson:
> Couldn't all those `text-decoration:underline` elements just be removed?

Perhaps ... but this is an easier diff for those who have modified their stylesheet.  I have already done testing with this change and really just want to merge it.

PR #6313, opened 2024-03-10
Don't offer GV payment if order only has GV's

lat9:
If an order has only Gift Certificates/Vouchers, don't offer Gift Vouchers as a possible payment for the order.

Processing currently displays on the `checkout_payment` page, but quietly disallows the GV payment if the order contains only GV products.

### DIFF
--- a/includes/modules/order_total/ot_gv.php
+++ b/includes/modules/order_total/ot_gv.php
@@ -261,27 +261,46 @@ class ot_gv {
       }
     }
   }
-  /**
-   * check system to see if GVs should be made available or not. If true, then supply GV-selection fields on checkout pages
-   */
-  function credit_selection() {
-    global $db, $currencies;
-    $selection = array();
-    $gv_query = $db->Execute("SELECT coupon_id FROM " . TABLE_COUPONS . " WHERE coupon_type = 'G' AND coupon_active='Y'");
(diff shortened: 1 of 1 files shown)


includes/modules/order_total/ot_gv.php
```diff
@@ -261,27 +261,46 @@ function update_credit_account($i) {
...
+            if (str_starts_with($next_product['model'], 'GIFT') === false) {
+                $only_gvs_in_order = false;
+                break;
+            }
+        }
+
+        $selection = [];
+        $gv_query = $db->Execute("SELECT coupon_id FROM " . TABLE_COUPONS . " WHERE coupon_type = 'G' AND coupon_active = 'Y' LIMIT 1");
+        // checks to see if any GVs are in the system and active or if the current customer has any GV balance
+        if ($only_gvs_in_order === false && (!$gv_query->EOF || $this->use_credit_amount())) {
```
drbyte, line 286: Instead of checking `=== false`, would it make the code simpler if we simply `if ($only_gvs_in_order === true) { return $selection;}` after `$selection = [];` above, instead of doing the DB query etc?

lat9: See update just pushed.  The only difference is to `return [];` if only GVs are in the order.

PR #6314, opened 2024-03-11
New date validation function for Date Of Birth field

piloujp:
Added a new function 'zen_valid_date' to validate date depending on DATE_FORMAT constant by default.
It was made for dob field validation, where a date-picker is not the best option, but can be use anywhere.
It can be overridden by simply adding a desired format as a second parameter.
The function 'zen_date_raw' has been modified to make it format independent.


includes/functions/functions_dates.php
```diff
@@ -22,29 +22,53 @@
...
 
+
+/**
+ * Validate a date in the selected locale date format
+ *
+ * @param string $date
+ * @param string $format (optional) needs to be a valid short date format for DateTimeImmutableObject using / or - or nothing as separators
+ * @return bool|true|false
+ */
+function zen_valid_date(string $date, $format = DATE_FORMAT)
```
drbyte, line 59:
```suggestion
function zen_valid_date(string $date, string $format = DATE_FORMAT): bool
```

And the docblock for `@return` can be just `@return bool` because `bool` already means "true or false".

includes/functions/functions_dates.php
```diff
@@ -22,29 +22,53 @@
...
     }
 }
 
+
+/**
+ * Validate a date in the selected locale date format
+ *
+ * @param string $date
+ * @param string $format (optional) needs to be a valid short date format for DateTimeImmutableObject using / or - or nothing as separators
+ * @return bool|true|false
```
drbyte, line 57:
```suggestion
 * @return bool
```

zc_install/sql/updates/mysql_upgrade_zencart_200.sql
```diff
@@ -178,6 +178,9 @@ UPDATE configuration SET configuration_description = 'What is the weight of typi
 # Add template_settings field
 ALTER TABLE template_select ADD template_settings LONGTEXT DEFAULT NULL;
 
+# Change minimum dob field length for new date VALIDATION
+UPDATE configuration SET configuration_value = 8 WHERE configuration_key = 'ENTRY_DOB_MIN_LENGTH';
```
drbyte, line 182: Question: Should we change it only if it's currently at the previous default of `10`?
```suggestion
UPDATE configuration SET configuration_value = 8 WHERE configuration_key = 'ENTRY_DOB_MIN_LENGTH' AND configuration_value=10;
```

piloujp:
Code review applied, everything looks good now. ;)

PR #6316, opened 2024-03-11
ipn-handler - catch undefined constant

### LINKED ISSUE #6315: Failed order when only PayPal Express defined under PHP 8.1
[11-Mar-2024 18:28:17 UTC] PHP Fatal error:  Uncaught Error: Undefined constant "MODULE_PAYMENT_PAYPAL_ORDER_STATUS_ID" in /home/account/domains/client/public_html/ipn_main_handler.php:360

Perhaps we should define this if !defined after inclusion of application_top.php?

drbyte:
Fixes #6315 - Failed order when PayPal Standard not installed, under PHP 8.1

### DIFF
--- a/ipn_main_handler.php
+++ b/ipn_main_handler.php
@@ -251,8 +251,14 @@ if (isset($_GET['type']) && $_GET['type'] === 'ec') {
         die();
     }
 
+    /**
+     * Catch undefined constants
+     */
+    zen_define_default('MODULE_PAYMENT_PAYPAL_ORDER_STATUS_ID', 2);
+
     // this is used to determine whether a record needs insertion. ie: original echeck notice failed, but now we have cleared, so need parent record established:
     $new_record_needed = ($txn_type === 'unique');
+
     /**
      * evaluate what type of transaction we're processing
      */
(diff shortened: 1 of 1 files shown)


ipn_main_handler.php
```diff
@@ -1,517 +1,584 @@
...
+                    //payment_status=Denied or failed
+                    $new_status = ($isECtransaction ? MODULE_PAYMENT_PAYPALWPP_REFUNDED_STATUS_ID : MODULE_PAYMENT_PAYPAL_REFUND_ORDER_STATUS_ID);
+                    break;
+                case 'echeck-cleared':
+                    $new_status = (int)MODULE_PAYMENT_PAYPAL_ORDER_STATUS_ID;
+                    break;
+                case ($txn_type === 'express-checkout-cleared' || str_starts_with($txn_type, 'cleared-')):
+                    //express-checkout-cleared
+                    $new_status = ($isECtransaction && defined('MODULE_PAYMENT_PAYPALWPP_ORDER_STATUS_ID') ? MODULE_PAYMENT_PAYPALWPP_ORDER_STATUS_ID : MODULE_PAYMENT_PAYPAL_ORDER_STATUS_ID);
+                    if (empty($new_status)) {
```
scottcwilson, line 527: Should `MODULE_PAYMENT_PAYPAL_ORDER_STATUS_ID` in above statement be casted to int?

scottcwilson (commented):
Looks good just one question.

ipn_main_handler.php
lat9: IMO, that entire conditional set into `$new_status` should be cast to int, not just that one constant.

drbyte: Okay, will do that for consistency and for clarity to the human reader.

Everywhere the value gets used already first casts it to `int`. eg:
https://github.com/zencart/zencart/blob/445529df14235b84547b5c35f3b2d8341380c859/includes/functions/functions_osh_update.php#L85

drbyte: Done.

scottcwilson (approved):

PR #6317, opened 2024-03-12
message in estimator when no shipping available

scottcwilson:
If all shipping methods have zones set and the selected address (or entered data) is outside the zones, the estimator simply shows a blank below available methods.

ship_before

This change adds the warning message from the checkout page. 

ship_after


proseLA:
moving forward, can we perhaps have language files just include text, and template files include the styling?

i think it is easier to maintain.

lat9:
> moving forward, can we perhaps have language files just include text, and template files include the styling?
> 
> i think it is easier to maintain.

Why not start with this change?

scottcwilson:
Done

PR #6319, opened 2024-03-12
Correct order tax calculations

### LINKED ISSUE #6318: More pricing woes
From [this](https://www.zen-cart.com/showthread.php?229752-Different-Total-Displayed&p=1399518#post1399518) Zen Cart forum thread, still a couple of wonky calculations:

From @todoonada:
> Currently a tax of 173.6363 is rounded to 173. (I have no decimals in the currency).
> I would like to have it rounded to 174 (half round up method).
> In the sense of "Total is 1910, a 10% tax of 174 is included".
> Now it displays 173 instead of 174.

> Sorry, some further information:
> 
> a) If I have a product price of 8280 and no shipping costs, the tax is displayed correctly as 753.
> b) If I have a product price of 8280 and shipping costs of 910 (total of 9190), the tax is displayed as 836, it should be 835.
> c) If I change the product price to 9190 (same like total in above) and free shipping, the tax is displayed correctly 835.
> 
> So it seems there is something wrong with the tax calculation of the shipping costs. 

From @OldNGreyBMR 

> Issue: currency conversion applied on item price in shopping_cart and reapplied on the sub-total for checkout_payment.
> 
> I updated /includes/classes/order.php to include changes from #6094 and #6141 to correct tax calculations.
> Prices are displayed including tax. Most orders are for domestic orders so they include tax. When the delivery address is overseas the items become tax exempt. This is determined by zones.
> 
> 
> Now with the new code:
> 
>  -  user selects a currency eg Euro and creates an account with a delivery address outside Australia.
>  -   The currency conversion is applied on the item:
>  -   item cost in AUD ex tax = $118.77 converted to EUR = €73.58 [ conversion rate of 0.61948508] = result is OK
> -    in checkout shipping, postage is calculated and returned in Euro = result is OK
>  -   in checkout payment, the sub-total before postage is now €45.93 = INCORRECT [ this is the items price in EURO reconverted to EURO by the same conversion factor ]

lat9:
Fixes #6318

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -900,11 +900,11 @@ class order extends base
             // currency.
             //
             if ($displaying_prices_with_tax === false) {
-                $tax_to_add = $currencies->value(zen_calculate_tax($tax_info['subtotal'], $tax_info['tax_rate']), false);
+                $tax_to_add = zen_calculate_tax($tax_info['subtotal'], $tax_info['tax_rate']);
             } else {
                 $tax_rate = $tax_info['tax_rate'];
                 $tax_rate_divisor = ($tax_rate >= 10) ? "1.$tax_rate" : "1.0$tax_rate";
-                $tax_to_add = $tax_info['subtotal'] - $currencies->value($tax_info['subtotal'] / $tax_rate_divisor, false);
+                $tax_to_add = $tax_info['subtotal'] - $tax_info['subtotal'] / $tax_rate_divisor;
             }
 
             $this->info['tax'] += $tax_to_add;
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -666,7 +666,7 @@ class order extends base
                 'price' => $products[$i]['price'],
                 'tax' => null, // calculated later
                 'tax_groups' => null, // calculated later
-                'final_price' => (DISPLAY_PRICE_WITH_TAX === 'true') ? $products_final_price_without_tax : zen_round($products_final_price_without_tax, $decimals),
+                'final_price' => $products_final_price_without_tax,
                 'onetime_charges' => $_SESSION['cart']->attributes_price_onetime_charges($products[$i]['id'], $products[$i]['quantity']),
                 'weight' => $products[$i]['weight'],
                 'length' => $products[$i]['length'] ?? null,
@@ -859,12 +859,12 @@ class order extends base
         //
         if (DISPLAY_PRICE_WITH_TAX === 'true') {
             $shown_price =
-                $currencies->value(zen_add_tax($product_final_price, $product_tax_rate), false) * $product_qty
-                    + $currencies->value(zen_add_tax($product_onetime_charges, $product_tax_rate), false);
+                zen_add_tax($product_final_price, $product_tax_rate) * $product_qty
+                    + zen_add_tax($product_onetime_charges, $product_tax_rate);
         } else {
             $shown_price =
-                $currencies->value($product_final_price * $product_qty, false)
-                    + $currencies->value($product_onetime_charges, false);
+                $product_final_price * $product_qty
+                    + $product_onetime_charges;
         }
 
         $this->info['subtotal'] += $shown_price;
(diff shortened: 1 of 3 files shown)


includes/classes/order.php
```diff
@@ -832,16 +842,11 @@ function setTaxRatesForProduct($products, $loop, $index, $taxCountryId, $taxZone
...
+     * and running-sub-total array for the final tax calculations.
      */
-    function calculateTaxForProduct($index, $taxRates, $taxCountryId, $taxZoneId)
+    protected function calculateTaxForProduct($index, $taxRates)
     {
-        $this->use_external_tax_handler_only = false;
-        $this->notify('NOTIFY_ORDER_CART_EXTERNAL_TAX_HANDLING', [], $index, $taxCountryId, $taxZoneId);
-        if ($this->use_external_tax_handler_only == true) {
-            return;
-        }
```
drbyte, line 843: Removal of the `NOTIFY_ORDER_CART_EXTERNAL_TAX_HANDLING` notifier prevents the integration of 3rd party tax-authority tools which could handle all the tax calculations for the order, such as applying per-county rules etc.

includes/classes/order.php
```diff
@@ -854,43 +859,56 @@ function calculateTaxForProduct($index, $taxRates, $taxCountryId, $taxZoneId)
...
             $shown_price =
-                $currencies->value(zen_add_tax($product_final_price, $product_tax_rate)) * $product_qty
-                    + $currencies->value(zen_add_tax($product_onetime_charges, $product_tax_rate));
-
-            // -----
-            // Calculate the amount of tax included in the price when tax-in pricing is enabled.
-            //
-            $tax_add =
-                $currencies->value(zen_calculate_tax($product_final_price, $product_tax_rate)) * $product_qty
-                    + $currencies->value(zen_calculate_tax($product_onetime_charges, $product_tax_rate));
```
drbyte, line 865: Was the dropping of all calls to `$currencies->value()` (which does exchange rate conversion) intentional?

lat9: Yes, since the order's base (as recorded in the database) values are kept in the site's DEFAULT_CURRENCY and converting them at this point caused various rounding errors.  The values are converted to the selected currency for display during customer and admin processing.

includes/classes/order.php
lat9: @scottcwilson checked all currently-posted plugins (and I checked on client's sites that have customized tax-handlers); none of those integrations made use of this notification.  For my clients, I've made use of the following notifications instead:
```
            /* From /includes/classes/order.php, during the cart->order conversion process */
            'NOTIFY_ORDER_CART_AFTER_ADDRESSES_SET',
            'NOTIFY_ORDER_CART_ADD_PRODUCT_LIST',

            /* From /includes/classes/order.php, after creating the base 'orders' entry */
            'NOTIFY_ORDER_DURING_CREATE_ADDED_ORDER_HEADER',

            /* From various functions in /includes/functions/functions_taxes.php */
            'NOTIFY_ZEN_GET_TAX_RATE_OVERRIDE',
            'NOTIFY_ZEN_GET_TAX_DESCRIPTION_OVERRIDE',
            'NOTIFY_ZEN_GET_MULTIPLE_TAX_RATES_OVERRIDE',

            /* From /includes/modules/order_total/ot_shipping.php */
            'NOTIFY_OT_SHIPPING_TAX_CALCS',
```

includes/classes/order.php
lat9: While there's still one remaining issue, as posted by @piloujp on the forums [here](https://www.zen-cart.com/showthread.php?229752-Different-Total-Displayed&p=1399601#post1399601), my opinion is that these changes are **_much closer_** to full correction of the penny-off calculations that have been haunting the ZC base for years.

I'll work to address that issue (and possibly others) once this gets a bit wider testing via incorporation into the next release candidate.

includes/classes/order.php
drbyte: I think it was TaxCloud or something like that which first used it. But they've not cooperated with any invitations to collaborate, for ages. So we can drop it for now.

PR #6321, opened 2024-03-13
Clarify "no shipping" msg for estimator

scottcwilson:
Unexpected issue with PR #6317: the first time the shopping cart page loads, when the state and post/zip have not been filled in, if the shipping modules are constrained to specific zones, the warning appears ("Sorry, we have no online options for shipping this order to the address shown. Please contact us for alternate arrangements!") which could be misleading.  It's hard to determine for non logged in users if sufficient data has been entered to know shipping is not possible (since zip code and state are both optional). 

If the person is logged in, they have entered an address which has a zone, and it can be definitively determined whether or not it is possible to ship to them.


drbyte:
But ... what about the case when the not-logged-in customer DOES provide address information but no quotes are available for that? Then it SHOULD show that message, shouldn't it? Otherwise doesn't it leave them with a blank space with no results, thus leaving them wondering what's going on?  Or is that the intent: if they see no results they'll be smart enough to go explore providing more/different address details? But what if they "do" provide as much information as is possible (without login), and still no results? Are we just leaving it blank?

Also ... is `TEXT_NO_SHIPPING_AVAILABLE_ESTIMATOR` used ONLY in the shipping estimator? (I would guess so based on the name of the constant, but haven't checked to be sure.) If it's only used by the estimator, then perhaps the wording could be revised?

scottcwilson:
> But ... what about the case when the not-logged-in customer DOES provide address information but no quotes are available for that? Then it SHOULD show that message, shouldn't it? 

Yes but this condition is very hard to detect, like I said.  If everyone used zip codes and states, you could ensure all three values were set.  But as it is, it's hard to differentiate between initial page load and bad data entry.

This is probably why they didn't use this message in the first place.

drbyte:
I'm inclined to lean towards giving the user more information rather than giving them nothing.

Thus I propose:
```diff
-    'TEXT_NO_SHIPPING_AVAILABLE_ESTIMATOR' => 'Sorry, we have no online options for shipping this order to the address shown.  Please contact us for alternate arrangements!',
+    'TEXT_NO_SHIPPING_AVAILABLE_ESTIMATOR' => 'Sorry, we have no online options for shipping this order to the address selected.<br><br>Please edit/update your desired shipping address to get updated quotes.<br><br>Also, please contact us to make alternate arrangements if necessary!',
```
... which renders as:
Screen Shot 2024-03-13 at 10 22 40 PM

scottcwilson:
I tweaked your language a bit, but this seems ok.

PR #6323, opened 2024-03-13
Update CURL Tester endpoints

drbyte:


scottcwilson:
UPSXML is going away in two months ... should we remove these?  Does @lat9 know the new endpoints for testing?

drbyte:
I checked the UPS REST module. It uses the same endpoints as the XML module. 
So I simply changed the display text from UPSXML to UPS for  simplicity.

PR #6325, opened 2024-03-14
Email failures: Don't log if message rejected for spam

scottcwilson:
Seeing more and more of these: 

```
[13-Mar-2024 21:40:32 UTC] PHP Notice:  Email Error: SMTP Error: data not accepted.SMTP server error: DATA END command failed Detail: Mail contain spam content BCCCF140117^M
 SMTP code: 550 Additional SMTP info: 5.7.1
 in /home/account/domains/client.com/public_html/includes/functions/functions_email.php on line 426
```

There's no value to dropping a log; this is just idiots pasting in junk to the contact_us page. 

I suspect the message above was written by a non-native speaker of English and will eventually be corrected to say "Mail contain**s** spam content" so I'm just grepping for the last two words.

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -506,7 +506,10 @@
             if ($ErrorInfo !== '') {
                 $mail_langs = $mail->getTranslations();
                 if (strpos($ErrorInfo, $mail_langs['recipients_failed']) === false) {
-                    trigger_error('Email Error: ' . $ErrorInfo);
+                   // Don't log SMTP rejected spam; log others
+                   if (!str_contains($ErrorInfo, 'spam content')) {
+                       trigger_error('Email Error: ' . $ErrorInfo);
+                   }
                 } else {
                     $log_prefix = (IS_ADMIN_FLAG === true) ? '/myDEBUG-bounced-email-adm-' : '/myDEBUG-bounced-email-';
                     $log_date = new DateTime();


includes/functions/functions_email.php
```diff
@@ -506,7 +506,9 @@ function zen_mail($to_name, $to_address, $email_subject, $email_text, $from_emai
             if ($ErrorInfo !== '') {
                 $mail_langs = $mail->getTranslations();
                 if (strpos($ErrorInfo, $mail_langs['recipients_failed']) === false) {
-                    trigger_error('Email Error: ' . $ErrorInfo);
+                   if (strpos($ErrorInfo, 'spam content') !== false) {
```
lat9, line 509: Isn't that going to log _only_ if that content **is** found?

scottcwilson: No.  If it's false means if it's not there.

lat9: Right, but you've used  `!== false` which means that (er) it's _not_ not there.  Shouldn't the conditional use `=== false`?

drbyte: @lat9 is correct.

But if we just use `str_contains()` instead, it's even clearer:

```php
if (!str_contains($ErrorInfo, 'spam content')) {
```

----

Examples:
```php
$ErrorInfo = 'Mail contain spam content';

if (!str_contains($ErrorInfo, 'spam content')) {
    echo 'Not spammy, so LOG IT';
    // trigger_error to do the logging
} else {
    // we don't use this ELSE clause, because we just skip logging
    echo 'Spammy, so do NOT log it.';
}

// RESULT: "Spammy, so do NOT log it."
```

scottcwilson: ARGH I was looking at the wrong line.  Yes, the new PHP8 functions are much easier to read, will switch to that, thanks for catching this.

scottcwilson:
Please note: this file is targeted at Zen Cart 2.0.0, which assumes PHP8. 
Users who need this fix in a PHP7 environment because they are using an older version of Zen Cart should change 

```
if (!str_contains($ErrorInfo, 'spam content')) {
```
to

```
if (strpos($ErrorInfo, 'spam content') === false) { 
```

PR #6329, opened 2024-03-16
allow arrays for NOTIFY_ADMIN_ORDERS_ADDRESS_FOOTERS

torvista:
once fettled, I'll add it to the other two instances in this file.
2024-03-16 10_18_18-Admin Orders — Mozilla Firefox

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -587,16 +587,25 @@ if (!empty($action) && $order_exists === true) {
                 <td>&nbsp;</td>
                 <td class="noprint"><a href="https://maps.google.com/maps/search/?api=1&amp;query=<?php echo urlencode($order->customer['street_address'] . ',' . $order->customer['city'] . ',' .  $order->customer['state'] . ',' . $order->customer['postcode']); ?>" rel="noreferrer" target="map"><i class="fa-regular fa-map">&nbsp;</i> <u><?php echo TEXT_MAP_CUSTOMER_ADDRESS; ?></u></a></td>
               </tr>
-<?php
-  $address_footer_suffix = '';
-  $zco_notifier->notify('NOTIFY_ADMIN_ORDERS_ADDRESS_FOOTERS', 'customer', $address_footer_suffix, $order->customer);
-  if (!empty($address_footer_suffix)) {
-  ?>
-                <tr><td>&nbsp;</td><td><?php echo $address_footer_suffix; ?></td></tr>
-<?php } ?>
-              <tr class="noprint">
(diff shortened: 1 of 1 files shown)


admin/orders.php
```diff
@@ -587,16 +587,19 @@ function couponpopupWindow(url) {
...
-  ?>
-                <tr><td>&nbsp;</td><td><?php echo $address_footer_suffix; ?></td></tr>
-<?php } ?>
-              <tr class="noprint">
-                <td colspan="2"><?php echo zen_draw_separator('pixel_trans.gif', '1', '5'); ?></td>
-              </tr>
+                <?php
+                $address_footer_suffix = '';
+                $zco_notifier->notify('NOTIFY_ADMIN_ORDERS_ADDRESS_FOOTERS', 'customer', $address_footer_suffix, $order->customer);
+                if (is_array($address_footer_suffix)) {
```
drbyte, line 593: Was this ever used as a non-array previously? 
If it could be returned as a non-empty string, then perhaps it should be cast to an array first?

The proposed code right now will just ignore it if it's a string.

torvista: You mean to make it backwards-compatible?

drbyte: Yes.

torvista: I'm not sure of the best approach.
While I'd like to be consistent, looking at other notifiers there is no consensus if the starting variable is a string or an array. Perhaps there should be a decision on the recommended method to use, going forward/for future rewrites.

drbyte: I don't know how any consensus applies in relation to "other" notifiers. Each notifier stands on its own in terms of the variables it responds to.

That said, `$address_footer_suffix = '';` clearly sets it to a string when passed "to" the observer. A purist would argue that it must therefore always be a string, but this is PHP. It can be returned as an array if you want to. Or we could set it to an array in the first place. We have choices.

BUT my point was: this notifier was previously released in at least v1.5.7 and v1.5.8 (I didn't check older versions) where it accepted and responded with a string, and if that string was returned as non-blank, it was echoed out as a string.

You've now changed that to only output if it's an array. That means anybody's code/plugins that previously used that notifier and returned a string ... is now broken. 

In fairness, the "symptom" of "broken" is simply that the string is skipped. No errors or logs are displayed or reported, and the application doesn't halt/fail/break because of it.

drbyte:
I'm satisfied with this.

Go ahead and replicate to the other instances where the notifier is called.

PR #6335, opened 2024-03-19
fix category_product observers to clear deprecation notices

proseLA:
```
PHP Deprecated: Automatic conversion of false to array is deprecated...
```

see [php 8.1 deprecations](https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.autovivification-false:~:text=Autovivification%20from%20false,is%20now%20deprecated.).

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -3,7 +3,7 @@
  * @copyright Copyright 2003-2024 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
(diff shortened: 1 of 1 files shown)


lat9:
On which line(s) is that deprecation thrown?  I'll note that changing the value initially submitted to the notification changes its signature to any watching observer.

proseLA:
> On which line(s) is that deprecation thrown? I'll note that changing the value initially submitted to the notification changes its signature to any watching observer.

it is thrown on the observer.  that is where we are asking people to change from `false` to an array.

lat9:
In the observers I've written that use those notifications, I specifically check for `false` and convert to any empty array before I add my additional fields.

proseLA:
ok

PR #6336, opened 2024-03-20
Enable the splitPageResults class to be overridden

lat9:


drbyte:
Instead of an observer hook (which must fire "very" early) followed by a conditionally-declared class, why not the following approach:

- replace the autoloader `loadFile` call with a call to `init_split_page_results.php`
- then `init_split_page_results.php` can be overridden like any other init file
- and `init_split_page_results.php` could check `if (!class_exists('splitPageResults')) { require DIR_WS_CLASSES . 'split_page_results.php'; }`

Then any plugin that wants to load a different file could either override via the `init_includes/overrides` dir, or in some other way ensure that the desired substitute class file be `require`d prior to the `init_split_page_results.php` file being called.

lat9:
> Instead of an observer hook (which must fire "very" early) followed by a conditionally-declared class, why not the following approach:
> 
>     * replace the autoloader `loadFile` call with a call to `init_split_page_results.php`
> 
>     * then `init_split_page_results.php` can be overridden like any other init file
> 
>     * and `init_split_page_results.php` could check `if (!class_exists('splitPageResults')) { require DIR_WS_CLASSES . 'split_page_results.php'; }`
> 
> 
> Then any plugin that wants to load a different file could either override via the `init_includes/overrides` dir, or in some other way ensure that the desired substitute class file be `require`d prior to the `init_split_page_results.php` file being called.

I _hate_ the `/init_includes/overrides` usage.  Whenever I'm doing a 'cold' upgrade (i.e. not a long-term client), it seems like I _always_ run into something that is using that type of override.  However, now that I've had my first sip of coffee and I can actually interpret what you said, I _like_ the idea of making an init_include that conditionally loads the class as you indicated.

drbyte (approved):

drbyte:
Should we do the same admin-side?

lat9:
> Should we do the same admin-side?

I'm not sure that it's required on the admin-side, but it _would_ be a fairly teeny update.  I'll defer to the rest of the community as to the need.

drbyte:
I can't recall which one it was, but I'm pretty sure I remember troubleshooting things in a replacement class admin-side.
Vaguely thinking it had to do with a search addon?
I can't recall why they'd overridden the class.

lat9:
I guess that since I'm "in there" anyway, keeping the override possibilities in-sync between admin/storefront will potentially reduce _pain_ in the future.  Update coming shortly.

PR #6338, opened 2024-03-20
Display "Single Column Settings" for responsive classic

lat9:
Force the "Layout Boxes Controller" to show the settings for "Single Column" mode for the `responsive_classic` template.


lat9:
Noting that without this change, it's right impossible (without database 'surgery') to get the search_header gone on the built-in templates (especially on a 'fresh' install), since the zc_install SQL sets `layout_boxes::layout_box_status_single` to 1 for all built-in's.

PR #6340, opened 2024-03-20
Admin, Orders, Edit: add "Copy to Clipboard" buttons to addresses

torvista:
Here's a starting point...


torvista:
I broke something again?

lat9:
> I broke something again?

I don't think so, @zcwilt has been making some updates to those unit tests; perhaps something's still not quite right?

zcwilt:
Looks like some of the PHP version repos are not responding properly. 
Probably just a temporary snafu.
Will check later and re-run.

zcwilt:
Yup. https://ppa.launchpadcontent.net/ondrej/php/ubuntu is foobar

zcwilt:
https://ubuntu.social/@launchpadstatus/112128869030316721
down for maintenance

zcwilt:
All back to normal. Have re-run the job

admin/includes/javascript/orders.php
```diff
@@ -0,0 +1,42 @@
...
+// Copy address details to clipboard
+if (isset($action) && $action === 'edit') { ?>
+<script>
+    function copyToClipboard(addressType) {
+        <?php
+        $address_types = ['customer', 'delivery', 'billing'];
+        foreach ($address_types as $address_type) {
+            $addressBlock[$address_type] =
+                $order->$address_type['name'] . "\\r\\n" .
+                (empty($order->$address_type['company']) ? '' : $order->$address_type['company'] . "\\r\\n") .
```
drbyte, line 11: Are these double-backslashes necessary? When they're inside double-quotes like this, PHP would already convert `\r` to carriage-return and `\n` as new-line.

If the reason for the double-backslash is to "escape" things for later use with the javascript, then probably using single-quotes in PHP would suffice.

admin/includes/javascript/orders.php
```diff
@@ -0,0 +1,42 @@
...
+                break;
+            case('billing') :
+                address = '<?= $addressBlock['billing'] ?>';
+                break;
+            default:
+                return;
+        }
+        navigator.clipboard.writeText(address).then(r => '');
+        // show the copied text
+        alert('"' + addressType + '":' + "\n" + address);
```
drbyte, line 38: Is this alert intentionally showing a confirmation pop-up that the user must acknowledge?
Thinking about UX here.

Perhaps updating the button text to say "Copied!" might be sufficient acknowledgement?

drbyte:
(FYI: I deleted discussion comments related to external test suite issues.)

drbyte:
I made some updates:
- in JS file, in the `$order->$address_type[]` entries I wrapped `$address_type` with braces like `{$address_type}` to prevent PHP from triggering warnings about ambiguity
- I relocated the bulk of the PHP processing to before the `<script>` output
- I added lang defines for the `Copy` button and the `Copied!` state
- Added ability for the button to self-update to say "Copied" after copy to clipboard is completed.

drbyte:
I was toying with simplifying the many `empty()` checks from the longer verbose form to use null-coalesce to make it simpler:
```diff
- (empty($order->{$address_type}['suburb']) ? '' : $order->{$address_type}['suburb'] . '\n') .
+ ($order->{$address_type}['suburb'] ?? '')  . '\n' .
```
... but that left blank lines (in the copied text) for those elements that were blank-but-not-null.
I wasn't sure if that would be the desired outcome or not.

eg: with blank-but-not-null/undefined "company" and "suburb";
```
Bill Smith

123 Any Avenue

Los Angeles
California
12345-6789
United States
805-555-1212
bill@example.com
```
vs without null-coalesce:
```Bill Smith
123 Any Avenue
Los Angeles
California
12345-6789
United States
805-555-1212
bill@example.com
```

torvista:
I said it was a starting point...at least I got the initial php tag correct!
I know it's messy with the tertiary operators, but better to have no blanks I think, you can't second guess where the text will be used.

drbyte:
> I said it was a starting point...at least I got the initial php tag correct!

Nah, u done good! 😄 Yes it was a good starting point.

drbyte:
I added some padding. 

Desktop:
Screen Shot 2024-03-22 at 2 03 18 PM

Mobile:
Screen Shot 2024-03-22 at 2 03 33 PM

Copied:
Screen Shot 2024-03-22 at 2 03 42 PM

torvista:
I liked the minimalist look, but think maybe it's better to remove the inline style completely and target them via css, since the containing tables have an id.

Issue #6341, opened 2024-03-20
Incorrect date displayed in admin header if `extension=intl` isn't loaded


lat9:
If running with a PHP base whose `php.ini` doesn't include the `intl` extension, the date in the admin header displays like:

![image](https://github.com/zencart/zencart/assets/2685585/777601e4-787b-41f9-b6d3-ac9158f68b76)

With that PHP extension loaded, the header date's display looks like:

![image](https://github.com/zencart/zencart/assets/2685585/dff7056f-3fcb-4f09-b158-04cb7493307b)

Credit: @proseLA for helping to figure that out!

drbyte:
Aside: I've updated the docs to reflect that `intl` is a required extension:
https://docs.zen-cart.com/user/first_steps/server_requirements/

lat9:
Should there be a zc_install check as well?

drbyte:
> Should there be a zc_install check as well?

Agreed.

PR #6342, opened 2024-03-20
Add zc_install check for `intl` extension

drbyte:
Ref #6341


drbyte:
I set this as a `WARN` priority, not a `FAIL` priority. 

So, if the extension is not installed, zc_install will display an alert, but will still allow the user to proceed.

lat9 (approved):

PR #6350, opened 2024-03-23
Admin MFA

drbyte:
Add email-driven 2-factor-auth to admin login, which shows after the Admin has successfully provided their password.

(OTP codes could be generated via a modified custom function to use something like Google Authenticator, etc)

Screen Shot 2024-03-23 at 4 11 52 AM


admin/includes/functions/admin_access.php
```diff
@@ -374,7 +374,7 @@ function zen_validate_user_login(string $admin_name, string $admin_pass): array
                 $response = zen_call_function(ZC_ADMIN_TWO_FACTOR_AUTHENTICATION_SERVICE, [$result['admin_id'], $result['admin_email'], $result['admin_name']]);
```
scottcwilson, line 374: This is where the enable/disable switch will go?  Can we also check for localhost or a private IP and skip if true?

drbyte: It's really hard to do that by relying on any IP-based conditions, because all servers have a localhost IP defined on themselves, and if they're on any network infrastructure other than directly on public internet, they'll be on a private IP range within their internal network behind the firewall.

And, then it would be super hard to test/verify the operation of MFA functionality in dev if it self-disabled.

So, no.

proseLA:
are we planning on offering numerous choices?  using google authenticator removes the need to check for expired tokens/codes as it is already a time based solution.

i have [the code already.](https://mxworks.cc/?product=2fa-for-zen-cart)

drbyte:
> are we planning on offering numerous choices? using google authenticator removes the need to check for expired tokens/codes as it is already a time based solution.
> 
> i have [the code already.](https://mxworks.cc/?product=2fa-for-zen-cart)

Yes. GA would be the preferred primary OTP option. Are you interested in contributing it to core?

proseLA:
> > are we planning on offering numerous choices? using google authenticator removes the need to check for expired tokens/codes as it is already a time based solution.
> > i have [the code already.](https://mxworks.cc/?product=2fa-for-zen-cart)
> 
> Yes. GA would be the preferred primary OTP option. Are you interested in contributing it to core?

i'll let you have a look.  if you want to add it to core, i have no problem with that.  i have sold 0 of them.

the repo does not have the changes i made to[ this file.](https://github.com/zencart/zencart/blob/master/admin/users.php)  it seemed that the amount of notifiers needed was not worth the effort.  on that file, i created a button for an admin to delete the 2FA seed for any user.  once done, that user would need to create a new seed afterwards.

proseLA:
in looking at my code, it seems that i based it on an outside library.  unfortunately that library has not been updated in a number of years, and newer versions of php will not work without modifications.

when i have more time, i will revisit that plugin and see about updating it.

drbyte:
> in looking at my code, it seems that i based it on an outside library. unfortunately that library has not been updated in a number of years, and newer versions of php will not work without modifications.

Yes, I observed the same: It's not been updated in awhile. But it's still recommended from a lot of less-older sources, rightly or wrongly. Most other libraries co-depend on multiple additional libraries, making it much more complex to integrate.

That said, I've got it working with minor updates to the library, albeit an extensive rewrite to the handler function, mostly to hybridize the email one in this PR with the GA lib.  Gotta refine it a bit before updating this PR with it.

scottcwilson:
If you merge this into master, we could add database changes to support MFA switchability per-user.

drbyte:
MultiFactorAuth class reworked to incorporate various features from more updated implementations, but still be a standalone class.

Store owner can enable MFA via a config switch in "Admin->My Store".
Store owner can exempt certain users (should only be for automated logins like ShipStation or similar plugins which can't do MFA automatically).
Storeowner can "delete" MFA from an Admin user account as a means of resetting access if it's not working. User will re-set-up on next login.
User has choice of OTP or code-by-email when setting up MFA for their profile.

PR #6354, opened 2024-03-23
fix mfa in testing

zcwilt:


scottcwilson:
I didn't see where it was looking for a blank value and skipping the process.

PR #6358, opened 2024-03-26
fix undefined variable debug log

scottcwilson:
--> PHP Warning: Undefined variable $date_separator_idx in /home/client/public_html/includes/functions/functions_dates.php on line 185.

### DIFF
--- a/includes/functions/functions_dates.php
+++ b/includes/functions/functions_dates.php
@@ -196,7 +196,7 @@ function zen_checkdate($date_to_check, $format_string, &$date_array)
         }
     }
 
-    if ($date_separator_idx != $format_separator_idx) {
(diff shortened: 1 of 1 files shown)


includes/functions/functions_dates.php
```diff
@@ -196,7 +196,7 @@ function zen_checkdate($date_to_check, $format_string, &$date_array)
         }
     }
 
-    if ($date_separator_idx != $format_separator_idx) {
+    if (!isset($date_separator_idx) || !isset($format_separator_idx) || $date_separator_idx != $format_separator_idx) {
```
lat9, line 199: Please combine those two `!isset()`'s into `!isset($date_separator_idx, $format_separator_idx)`.  If either variable is 'not set', that condition will be  `true`.

PR #6360, opened 2024-03-27
Category Tab: add css for selected category

torvista:
2024-03-27 11_07_52-Hardware _ Zen Cart!, The Art of E-commerce — Mozilla Firefox


drbyte:
Note that this will also affect the `categories` sidebox.

Also note that this same CSS selector being added here is already present in the old `classic` CSS too. 
So I'm okay with this addition.

torvista:
But, I see the mechanism for using that class is different/they are not in sync/consistent.
In the category tab, the class is applied based on the current cpath.
In the categories sidebox the class is applied to a category with no subcategories!

torvista:
just noticed this PR went awry, corrected now.

PR #6363, opened 2024-03-28
Show pass last changed date in admin users

scottcwilson:
pass_change


admin/users.php
brittainmark, line 1: line 135
`<th class="changed"><?php echo TEXT_PASS_LAST_CHANGED?></th>`
missing space before `?>`

brittainmark (commented):
great thanks.

PR #6369, opened 2024-04-02
Enable products' names to be overridden on listings

### LINKED ISSUE #6367: making product name consistent across the repo....
in looking at the code for v200, it seems that one can successfully override the `zen_get_products_name` function without touching any of the base code.  this is a change that i applaud. (thank you @torvista.  change was done in this commit 759b64d0af.)

the override is done when the `zen_get_products_name` function calls the `zen_get_product_details` function as seen here:

https://github.com/zencart/zencart/blob/c1b6aba2d61ffea86bc1d28bd964889b03fda415/includes/functions/functions_products.php#L32-L34

unfortunately, for the majority of times a customer is displayed a product, the `zen_get_products_name` function is not used.

one such instance is on line 219 here:

https://github.com/zencart/zencart/blob/c1b6aba2d61ffea86bc1d28bd964889b03fda415/includes/modules/product_listing.php#L219-L223

due to the multi-lingual aspect of ZC, one can very easily override the products description, as the method `zen_get_products_description` as seen on line 222 is consistently used whenever calling the products description. 

ideally if one were to want to change the construction of the product name as presented to the customer, one should be able to do it in 1 place; but and most importantly, do it WITHOUT changing any of the base zc code.

i see 2 potential solutions to this problem:

- change all uses of the product name to call the `zen_get_products_name` function.
- change all product listings to only retrieve the `products_id`, and then use the `zen_get_product_details` to retrieve all necessary information about the product.

i think the 2nd option affords more flexibility for more things.  but i'm open on that...  it seems the 1st option might be easier to implement.

does anyone have any thoughts?  and or willingness to accept a PR if one were attempted?

lat9:
Fixes #6367

### DIFF
--- a/includes/modules/product_listing.php
+++ b/includes/modules/product_listing.php
@@ -216,7 +216,7 @@ if ($num_products_count > 0) {
             $lc_text = '';
 
             $href = zen_href_link(zen_get_info_page($record['products_id']), 'cPath=' . zen_get_generated_category_path_rev($linkCpath) . '&products_id=' . $record['products_id']);
-            $listing_product_name = $record['products_name'] ?? '';
+            $listing_product_name = zen_get_products_name((int)$record['products_id']);
             $listing_description = '';
             if ((int)PRODUCT_LIST_DESCRIPTION > 0) {
                 $listing_description = zen_trunc_string(zen_clean_html(stripslashes(zen_get_products_description($record['products_id'], $_SESSION['languages_id']))), PRODUCT_LIST_DESCRIPTION);
--- a/includes/modules/product_listing.php
+++ b/includes/modules/product_listing.php
@@ -132,14 +132,14 @@ if ($num_products_count > 0) {
         $category_id = !empty($record['categories_id']) ? $record['categories_id'] : $record['master_categories_id'];
         $parent_category_name = trim(zen_get_categories_parent_name($category_id));
         $category_name = trim(zen_get_category_name($category_id, (int)$_SESSION['languages_id']));
+        $product_details = zen_get_product_details($record['products_id']);
         $records[] = array_merge($record,
             [
                 'parent_category_name' => (!empty($parent_category_name)) ? $parent_category_name : $category_name,
                 'category_name' => $category_name,
-//                'products_name' => $record['products_name'],
 //                'master_categories_id' => $record['master_categories_id'],
 //                'products_sort_order' => $record['products_sort_order'],
-            ]);
+            ], $product_details->fields ?? []);
     }
 
     if (!empty($_GET['keyword'])) $skip_sort = true;


drbyte:
Doesn't this cause another query to be run instead of using the already-retrieved value?

lat9:
> Doesn't this cause another query to be run instead of using the already-retrieved value?

It does, just trying to find a pivot-point in the code to satisfy the (what I consider to be valid) request in the referenced issue.

drbyte:
Would this work?

earlier in the file, around line 130:
```diff
    // Retrieve all records into an array to allow for sorting and insertion of additional data if needed
    $records = [];
    foreach ($listing as $record) {
        $category_id = !empty($record['categories_id']) ? $record['categories_id'] : $record['master_categories_id'];
        $parent_category_name = trim(zen_get_categories_parent_name($category_id));
        $category_name = trim(zen_get_category_name($category_id, (int)$_SESSION['languages_id']));
+       $product_details = zen_get_product_details($record['products_id']);
        $records[] = array_merge($record,
            [
                'parent_category_name' => (!empty($parent_category_name)) ? $parent_category_name : $category_name,
                'category_name' => $category_name,
-//                'products_name' => $record['products_name'],
//                'master_categories_id' => $record['master_categories_id'],
//                'products_sort_order' => $record['products_sort_order'],
-           ]);
+           ], $product_details->fields ?? []);
```

lat9:
> Would this work?
> 
> earlier in the file, around line 130:
> 
> ```diff
>     // Retrieve all records into an array to allow for sorting and insertion of additional data if needed
>     $records = [];
>     foreach ($listing as $record) {
>         $category_id = !empty($record['categories_id']) ? $record['categories_id'] : $record['master_categories_id'];
>         $parent_category_name = trim(zen_get_categories_parent_name($category_id));
>         $category_name = trim(zen_get_category_name($category_id, (int)$_SESSION['languages_id']));
> +       $product_details = zen_get_product_details($record['products_id']);
>         $records[] = array_merge($record,
>             [
>                 'parent_category_name' => (!empty($parent_category_name)) ? $parent_category_name : $category_name,
>                 'category_name' => $category_name,
> -//                'products_name' => $record['products_name'],
> //                'master_categories_id' => $record['master_categories_id'],
> //                'products_sort_order' => $record['products_sort_order'],
> -           ]);
> +           ], $product_details->fields ?? []);
> ```

I'm _pretty sure_ that that would "work".  I was just trying to keep the requested change on-point given that we're close to a zc200-final release.

Let's let @proseLA chime in, since he opened the to-be-closed issue.

proseLA:
with regards to @drbyte's proposed change, without question a step in the right direction.  in my limited testing this works on the product listing.

my issue gets to be a bit bigger.

this change works for all pages that make use of the product listing.  but what about the centerboxes?  the sideboxes?  the shopping cart?  some of these places make use of the method `zen_get_product_details`, ie the shopping cart, and some do not.  in addition, some places, the change is available to be made without making a change to the core code... ie, this notifier `NOTIFY_MODULES_FEATURED_PRODUCTS_B4_LIST_BOX` allows me to change the name there...  but why should i have to if it is already changed in the aforementioned function?

i suppose i'm looking to make the product more of an object, that i can change some aspect of it in 1 place, and it will be reflected throughout.

i think it is a bigger refactor.

but this single change is a step in the right direction.  and i thank you for that.

i would like to make a comment on the database query aspect of this issue.  the short answer is yes, but i am not that familiar with the caching that has been implemented and how that all works.  i will say that in my testing of other shopping cart solutions, ie prestacart, woocommerce, etc those carts are significantly slower in my experience.  furthermore when developing other modifications to the zen-cart code, i have noticed a significant number of what i believe to be repetitive queries when getting a products price.   while there is without question an additional query, i think the penalty is not that high, and that there are additional savings that could be made in other areas.

drbyte:
> i suppose i'm looking to make the product more of an object

Yes, and I agree that an active object is the ultimate direction to go. 
(As well some sort of extendability as you've mentioned in a couple other places, although a registry may be better than simply an empty class that inherits from a base/abstract. But that's another discussion.) 

When I last explored making the current product an object I ran into a couple snags where other dependencies needed addressing. I'll have to go find the notes I made about that to refresh my memory on those and see whether those ripple-effect issues have already been handled in the meantime.

drbyte:
> make a comment on the database query aspect of this issue. the short answer is yes, but i am not that familiar with the caching that has been implemented and how that all works

Fortunately, centralizing "many" functions to call the shared `zen_get_product_details()` function, which uses the same query string on every call, means that repeated calls to that function do indeed get a cached response, meaning the actual hit to the database engine is singular, with subsequent calls for the same page-hit coming from RAM.

scottcwilson:
@proseLA @lat9 @drbyte  please write up an example use case (or two!) for the documentation.   Thanks.

proseLA:
@scottcwilson i wish i had more time for the documentation.  unfortunately i do not.

personally, i think this change is part of a much bigger change that has not been merged as of yet. #6370.  

in looking at [this docs page](https://docs.zen-cart.com/dev/code/displaying_custom_fields/) on displaying a custom field; we are already advocating for changing a core file.  `includes/modules/pages/product_info/main_template_vars.php`  which is something i like to avoid at all costs.

with that pending change of #6370, one can use the notifier to include any information about the product and put it in the `$product_info` var; and then use that information in one's template files.  and completely bypass the `main_template_vars`.  and hopefully with that one notifier, the product object is now changed throughout its use in the storefront.

maybe this could have been done before; i do not have time to go back and look at the code.  i have always run into the problem with the product name field, and its lack of consistency across the storefront.

so now the documentation gets a bit trickier as how do we want to segment things out for v200 versus v15x? 

in addition, i like the docs to reflect what the code does; in my experience with you, it seems to me, you would like the docs to reflect what you want developers to do.  which is pretty apparent in the aforementioned page.

perhaps when that PR gets merged i will give some more thought.  but giving a single use scenario for changing the product name is really a minor element to the big picture.  the forum is loaded with users who want to know how to add a field to a product and then make use of that field in a template.  being able to do without changing any core ZC files is really the goal for me.  which it seems that we are on the way to doing that.

scottcwilson:
If you can tell me what this change will allow you to do - how it will change the user experience for the better - that would help.

proseLA:
you can change the product name based on any of the info in the products schema.

my use case has always been vintage.  i have some wine clients.  vintage needs to be a separate field.  IMO, it can not be part of the product name for a myriad of reasons.  but when the product name gets displayed it should always start with the vintage.

PR #6371, opened 2024-04-03
Skip "copy" button for shipping address when blank

### LINKED ISSUE #6368: admin, orders, edit: PHP Warning: Trying to access array offset on false \admin\includes\javascript\orders.php for store_pickup shipping
strict reporting,, make an order using store_pickup
Admin, Edit Order:

PHP Warning: Trying to access array offset on false in D:\GitHub\zencart\admin\includes\javascript\orders.php on line 18 etc,

as $order->delivery is not set

So perhaps delivery should be set with the store address?

drbyte:
Fixes #6368

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -648,7 +648,9 @@ if (!empty($action) && $order_exists === true) {
             <table class="table" id="addressDelivery">
               <tr>
                 <td><strong><?php echo ENTRY_SHIPPING_ADDRESS; ?></strong><br>
+                    <?php if (!empty($order->delivery)) { ?>
                     <button type="button" class="btn btn-xs btn-default mt-3" title="<?= TEXT_COPY ?>" onclick="copyToClipboard('delivery', this)"><?= TEXT_COPY ?></button>
+                    <?php } ?>
                 </td>
                 <td><?php echo (empty($order->delivery)) ? TEXT_NONE : zen_address_format($order->delivery['format_id'], $order->delivery, 1, '', '<br>'); ?></td>
               </tr>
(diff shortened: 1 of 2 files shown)


torvista:
Still get this
PHP Warning: Trying to access array offset on false in ADMIN/includes/javascript/orders.php on line 21.

WHy not change
$addressBlock[$address_type] = '';

torvista:
Still getting
PHP Warning: Trying to access array offset on false in ADMIN/includes/javascript/orders.php on line 21.

Why not change
$addressBlock[$address_type] = '';
to
continue;

drbyte:
Yes, I should add `continue`; I didn't want to "only" continue without setting the address to blank for that address_type in case the js were to be called "anyway".

That and other things updated.

admin/orders.php
```diff
@@ -648,7 +648,9 @@ function couponpopupWindow(url) {
             <table class="table" id="addressDelivery">
               <tr>
                 <td><strong><?php echo ENTRY_SHIPPING_ADDRESS; ?></strong><br>
+                    <?php if (!empty($order->delivery)) { ?>
                     <button type="button" class="btn btn-xs btn-default mt-3" title="<?= TEXT_COPY ?>" onclick="copyToClipboard('delivery', this)"><?= TEXT_COPY ?></button>
+                    <?php } ?>
                 </td>
                 <td><?php echo (empty($order->delivery)) ? TEXT_NONE : zen_address_format($order->delivery['format_id'], $order->delivery, 1, '', '<br>'); ?></td>
```
torvista, line 655: closing bracket of empty check should be at the end

drbyte: You mean the `(empty($order->delivery))` , right?

No, since the PHP statement is a simple ternary with a simple if-this-else-that logic that's not concatenating multiple strings or functions in the "this" or "else" parts, the true/false check can stand on its own before the opening `?`. Simpler to  read and maintain that way, because you don't have to worry about a dangling parenthesis at the end.

Technically it could be reduced to `empty($order->delivery)` without the wrapping parentheses because there's only one condition checked, but the wrapping parentheses does make it clear to the reader what the entire condition is.

PR #6376, opened 2024-04-04
Restoring NOTIFY_ADMIN_CUSTOMERS_LISTING_ELEMENT signature

### LINKED ISSUE #6375: NOTIFY_ADMIN_CUSTOMERS_LISTING_ELEMENT 'signature' change breaks pre-existing observers
... as merged for this commit: https://github.com/zencart/zencart/commit/6ce2a575ac2be575a067e365898c7754298db178,  has 'broken' the signature of that notification and now results in many PHP Warnings when used by an existing plugin [VAT for EU Countries](https://github.com/lat9/vat4eu/) that's been observing that notification for a couple of years.

lat9:
Fixes #6375

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -1700,7 +1700,7 @@ if ($action === 'edit' || $action === 'update') {
         $additional_columns = [];
         $zco_notifier->notify(
             'NOTIFY_ADMIN_CUSTOMERS_LISTING_ELEMENT',
-            $result,
+            array_merge($result, $customer),
             $additional_columns,
             $customer
         );


admin/customers.php
```diff
@@ -1700,7 +1700,7 @@ class="btn btn-default"><?php
         $additional_columns = [];
         $zco_notifier->notify(
             'NOTIFY_ADMIN_CUSTOMERS_LISTING_ELEMENT',
-            $result,
+            array_merge($result, $customer),
             $additional_columns,
             $customer
```
drbyte, line 1705: If we're merging $customer with $result, I guess we can remove the (recent) addition of $customer as well, right?

lat9: I kept that there, since there might be usage of that additional parameter ... no harm in keeping it, IMO.

PR #6391, opened 2024-04-11
Lost penny management

piloujp:
In case of a penny difference between invoice total and sum of numbers on the same invoice, this modification will adjust the total to equal the sum of other numbers.
The one penny difference is still visible between invoice with prices including tax and invoice with prices without tax, but one by one, they are coherent, all numbers add correctly.

See forum thread '[The Lost Penny of ZC](https://www.zen-cart.com/showthread.php?230029-The-Lost-Penny-of-ZC)'.

### DIFF
--- a/includes/modules/order_total/ot_cod_fee.php
+++ b/includes/modules/order_total/ot_cod_fee.php
@@ -122,7 +122,7 @@
               $cod_cost += zen_calculate_tax($cod_cost, $tax);
             }
           }
-          $order->info['option_modules']['cod_fee'] = $cod_cost;
+          $order->info['option_modules'][$this->code] = $cod_cost;
 
           $this->output[] = array('title' => $this->title . ':',
                                   'text' => $currencies->format($cod_cost, true,  $order->info['currency'], $order->info['currency_value']),
--- a/includes/modules/order_total/ot_cod_fee.php
+++ b/includes/modules/order_total/ot_cod_fee.php
@@ -122,7 +122,7 @@
               $cod_cost += zen_calculate_tax($cod_cost, $tax);
             }
           }
-		  $order->info['option_modules']['cod_fee'] = $cod_cost;
+          $order->info['option_modules']['cod_fee'] = $cod_cost;
 
           $this->output[] = array('title' => $this->title . ':',
                                   'text' => $currencies->format($cod_cost, true,  $order->info['currency'], $order->info['currency_value']),
--- a/includes/modules/order_total/ot_cod_fee.php
+++ b/includes/modules/order_total/ot_cod_fee.php
@@ -118,7 +118,7 @@
             $order->info['tax'] += zen_calculate_tax($cod_cost, $tax);
             $order->info['tax_groups'][$tax_description] += zen_calculate_tax($cod_cost, $tax);
             $order->info['total'] += zen_calculate_tax($cod_cost, $tax);
-            if (DISPLAY_PRICE_WITH_TAX == 'true') {
+            if (DISPLAY_PRICE_WITH_TAX === 'true') {
               $cod_cost += zen_calculate_tax($cod_cost, $tax);
             }
           }
--- a/includes/modules/order_total/ot_cod_fee.php
+++ b/includes/modules/order_total/ot_cod_fee.php
@@ -122,6 +122,7 @@
               $cod_cost += zen_calculate_tax($cod_cost, $tax);
             }
           }
+		  $order->info['option_modules']['cod_fee'] = $cod_cost;
 
           $this->output[] = array('title' => $this->title . ':',
                                   'text' => $currencies->format($cod_cost, true,  $order->info['currency'], $order->info['currency_value']),
--- a/includes/modules/order_total/ot_coupon.php
+++ b/includes/modules/order_total/ot_coupon.php
@@ -46,7 +46,7 @@ class ot_coupon extends base
      * $deduction amount of deduction calculated/afforded while being applied to an order
      * @var float|null
      */
-    public $deduction;
+    protected $deduction;
     /**
      * $description is a soft name for this order total method
      * @var string
@@ -162,7 +162,7 @@ class ot_coupon extends base
             }
 
             $order->info['total'] -= $od_amount['total'];
-            $order->info['option_modules']['coupon_amount'] = - $od_amount['total'];
+            $order->info['option_modules'][$this->code] = - $od_amount['total'];
 
             if (DISPLAY_PRICE_WITH_TAX != 'true') {
                 $order->info['total'] -= $tax;
--- a/includes/modules/order_total/ot_coupon.php
+++ b/includes/modules/order_total/ot_coupon.php
@@ -46,7 +46,7 @@ class ot_coupon extends base
      * $deduction amount of deduction calculated/afforded while being applied to an order
      * @var float|null
      */
-    protected $deduction;
+    public $deduction;
     /**
      * $description is a soft name for this order total method
      * @var string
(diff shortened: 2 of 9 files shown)


proseLA:
i will attempt to be brief.  as i stated [here](https://www.zen-cart.com/showthread.php?230029-The-Lost-Penny-of-ZC&p=1400000#post1400000), the problem resides in the need for the `ot_total` module.   

lets go over why i do not like your code:

- it is overly complex.
- it hard codes various elements into a new array in the `$order->info` object whose key is some new name that you think is somehow tied to the order total module.
- it does not account for other store owners developing their own order total modules where they run into the same problem that you have exposed here.

 and i would solve this problem completely differently.  i would look at this method:

https://github.com/zencart/zencart/blob/945701e29b6ae006c05c74570f33f8efc572ab77/includes/classes/order_total.php#L63

and potentially change it to something like this:

```php
function process()
    {
        global $order, $currencies;
        $order_total_array = [];
        if (is_array($this->modules)) {
            $this->notify('NOTIFY_ORDER_TOTAL_PROCESS_STARTS', ['order_info' => $order->info]);
            foreach ($this->modules as $value) {
                if (!isset($GLOBALS[$class])) continue;
                $GLOBALS[$class]->process();
                $this->notify('NOTIFY_ORDER_TOTAL_PROCESS_NEXT', ['class' => $class, 'order_info' => $order->info, 'ot_output' => $GLOBALS[$class]->output]);
                if (empty($GLOBALS[$class]->output)) {
                    continue;
                }
                for ($i = 0, $n = sizeof($GLOBALS[$class]->output); $i < $n; $i++) {
                    if (zen_not_null($GLOBALS[$class]->output[$i]['title']) && zen_not_null($GLOBALS[$class]->output[$i]['text'])) {
                        $order_total_array[] = [
                            'code' => $GLOBALS[$class]->code,
                            'title' => $GLOBALS[$class]->output[$i]['title'],
                            'text' => $GLOBALS[$class]->output[$i]['text'],
                            'value' => $GLOBALS[$class]->output[$i]['value'],
                            'sort_order' => $GLOBALS[$class]->sort_order,
                        ];
                    }
                }
            }
        }
        $orderTotal = 0;
        $notCalc = ['ot_total',];
        foreach ($order_total_array as $key => $totalItem) {
            if ($totalItem['code'] === 'ot_total') {
                $totalKey = $key;
            }
            if (in_array($totalItem['code'], $notCalc)) {
                continue;
            }
            $orderTotal += $currencies->value($totalItem['value'], false, $order->info['currency']);
        }
        if (empty($totalKey)) {
            return $order_total_array;
        }
        $order_total_array[$totalKey]['value'] = $orderTotal;
        $order_total_array[$totalKey]['text'] = $currencies->format($orderTotal, false, $order->info['currency'], $order->info['currency_value']);

        $order->info['total'] = $orderTotal;

        return $order_total_array;
    }
```

we are now recalculating the order total based on the addition of the all of the order total modules; have the ability to exclude certain modules if their total is included elsewhere and hopefully make address your `lost penny` issue.

the issues i see in my code are:
- it does not display the correct total on the checkout confirmation page (at least it did not, i tested it on a dev site with OPC).
- i rarely use foreign currencies, and in the 2 instances above i am not using the `$calculate_using_exchange_rate` as true.

perhaps you can solve those 2 issues.

and let us know if this solution works for you.

again, while i applaud your effort, i would hope we do not see code this confusing merged into the repo.

piloujp:
> i will attempt to be brief. as i stated [here](https://www.zen-cart.com/showthread.php?230029-The-Lost-Penny-of-ZC&p=1400000#post1400000), the problem resides in the need for the `ot_total` module.
> 
> lets go over why i do not like your code:
> 
>     * it is overly complex.
> 
>     * it hard codes various elements into a new array in the `$order->info` object whose key is some new name that you think is somehow tied to the order total module.
> 
>     * it does not account for other store owners developing their own order total modules where they run into the same problem that you have exposed here.
> 
> 
> and i would solve this problem completely differently. i would look at this method:

I must say I have a hard time understanding your comments:
* What is 'overly complex'? There is nothing complicated in this PR... Is it because it modifies many files? Or is it because of the line that determinates if a penny correction is necessary?
* When you speak of hard coded elements, I suppose you refer to keys associated with modules results like, for example, 'cod_fee' in $order->info['option_modules']['cod_fee'] = $cod_cost;.
If it is that, I agree it is not a good practice and I already changed it to: $order->info['option_modules'][$this->code] = $cod_cost;
* I don't see why this PR does not account for other owners developed ot modules, honestly...

This said your approach is interesting, although in the way you show it here it can't work as all modules have already done their calculations and displayed results. It happens on the line with this code: $GLOBALS[$class]->process(); just after the beginning of the foreach loop.
I did work on the idea though, and came out with this:
`    function process()
    {
        global $order, $currencies;
        $order_total_array = [];
        $orderTotal = 0;
        $mod_value = 0;
//        $notCalc = [];
        if (is_array($this->modules)) {
            $this->notify('NOTIFY_ORDER_TOTAL_PROCESS_STARTS', ['order_info' => $order->info]);
            foreach ($this->modules as $value) {
                if (!isset($GLOBALS[$class])) continue;
                if ($GLOBALS[$class]->code !== 'ot_total') {
                    $GLOBALS[$class]->process();
                } else {
                    $mod_value = $orderTotal;
                    $order->info['total'] = $orderTotal;
                    $GLOBALS[$class]->process();
                }
                $this->notify('NOTIFY_ORDER_TOTAL_PROCESS_NEXT', ['class' => $class, 'order_info' => $order->info, 'ot_output' => $GLOBALS[$class]->output]);
                if (empty($GLOBALS[$class]->output)) {
                    continue;
                }
                for ($i = 0, $n = sizeof($GLOBALS[$class]->output); $i < $n; $i++) {
                    if (zen_not_null($GLOBALS[$class]->output[$i]['title']) && zen_not_null($GLOBALS[$class]->output[$i]['text'])) {
                        if ($GLOBALS[$class]->code !== 'ot_total') {
                            $mod_value = $GLOBALS[$class]->output[$i]['value'];
//                            if (($GLOBALS[$class]->code !== 'ot_tax' || DISPLAY_PRICE_WITH_TAX !== 'true') && !in_array($GLOBALS[$class]->code, $notCalc)) {
                            if ($GLOBALS[$class]->code !== 'ot_tax' || DISPLAY_PRICE_WITH_TAX !== 'true') {
                                if(substr($GLOBALS[$class]->output[$i]['text'], 0, 1) === '-') {
                                    $orderTotal -= $currencies->value($GLOBALS[$class]->output[$i]['value'], true, $order->info['currency']);
                                } else {
                                    $orderTotal += $currencies->value($GLOBALS[$class]->output[$i]['value'], true, $order->info['currency']);
                                }
                            }
                        }
                        $order_total_array[] = [
                            'code' => $GLOBALS[$class]->code,
                            'title' => $GLOBALS[$class]->output[$i]['title'],
                            'text' => $GLOBALS[$class]->output[$i]['text'],
                            'value' => $mod_value,
                            'sort_order' => $GLOBALS[$class]->sort_order,
                        ];
                    }
                }
            }
        }

        return $order_total_array;
    }
`

It does the job although with actual ZC code, I don't think it was meant to be done here. Solution looks nicer in simpler too. But it is just an appearance. There are two potential problems that can occur:
First is the code to exclude some module from total, actually commented out.
It will only remove the module result from final total, but nothing would be changed in order->info and removed module will still be taken in account by other modules calculations. It might be possible to completely remove it from processing, like if it was not installed, if it is you meant.
Second is when a customer uses a currency different from the default one.
Currency conversion is normally done at display time, but with this solution it as to be done before each module total is added. If not, we are at square zero with two different calculation methods for total that will end up with a difference at the end. I had to set the currency conversion parameter to false in to_total.
Problem is that the converted total is saved in order->info in the displayed currency, but all other values are saved in default currency, and order->info is later used to save order data to database... Unfortunately, I can't see any solution for that.

piloujp:
Well, it seems ZC user must write their own ot_total module if they want to have invoices that with numbers that add together...
I just tried to provide a solution, actually two, with this PR but nobody seems to care.

Issue #6339, opened 2024-03-20
`ot_coupon`: Correct rounding errors


lat9:
@piloujp has provided a bunch of information and suggestions in [this](https://www.zen-cart.com/showthread.php?230003-Discount-coupon-value&p=1399705#post1399705) Zen Cart bug-report.

Since that correction is going to be complicated and require a fair amount of testing, I've suggested that it be deferred until zc201.

PR #6392, opened 2024-04-11
OT modules calculation correction

piloujp:
Mainly rewrite of ot_coupon.php to be consistent with other modules and different options settings.
Bug corrections on ot_gv and ot_group_pricing.
See forum thread '[Discount coupon value](https://www.zen-cart.com/showthread.php?230003-Discount-coupon-value/page3)'.
Related to #6339

### DIFF
--- a/includes/modules/order_total/ot_coupon.php
+++ b/includes/modules/order_total/ot_coupon.php
@@ -163,7 +163,7 @@ class ot_coupon extends base
             $order->info['total'] -= $od_amount['total'];
             $order->info['coupon_amount'] = $od_amount['total'];
 
-            if (DISPLAY_PRICE_WITH_TAX != 'true') {
+            if (DISPLAY_PRICE_WITH_TAX !== 'true') {
                 $order->info['total'] -= $tax;
             }
             $order->info['tax'] -= $tax;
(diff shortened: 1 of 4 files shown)


drbyte:
Sorry to be nitpicky, but your text editor is using tabs instead of spaces for indentation, and thus the alignment of the code is way off all over this PR.

One example:
Screen Shot 2024-04-14 at 4 16 07 PM

piloujp:
Sorry for that, I will try to change that. It is the automatic indentation, I think. Until I set that, I will use a regex to fix that and submit new commits.

includes/modules/order_total/ot_coupon.php
```diff
@@ -541,39 +543,86 @@ function calculate_deductions()
                         return $od_amount;
                         break;
                     case 'P': // percentage
-//                        $od_amount['total'] = zen_round($orderTotalDetails['orderTotal']*($coupon_details['coupon_amount']/100), $currencyDecimalPlaces);
-                        $od_amount['total'] = zen_round($orderAmountTotal * ($coupon_details['coupon_amount'] / 100), $currencyDecimalPlaces);
-//                        $ratio = $od_amount['total']/$orderTotalDetails['orderTotal'];
-                        $ratio = $od_amount['total'] / $orderAmountTotal;
+                        if ($this->calculate_tax == 'Credit Note') { // Calculate deduction from gross price
+                            $od_amount['total'] = (DISPLAY_PRICE_WITH_TAX == 'true' ? $orderAmountTotal - $orderTotalDetails['orderTax'] : $orderAmountTotal) * $coupon_details['coupon_amount'] / 100;
```
lat9, line 547: Here, and in other places, check for `DISPLAY_PRICE_WITH_TAX === 'true'`.

lat9 (changes requested):
See associated comment

piloujp:
After recently working on multiple tax rates calculation, I realized that files in this PR need more modification and these modifications can't be treated independently of my more recent code updates. I am not sure what is best, make a new PR or updating this one. I try to separate problems (and PRs) as much as possible, but It is not going to work in this case.

piloujp:
Closing this PR replaced by PR #6432 .

PR #6393, opened 2024-04-11
No tax should apply Tax basis = Store & store zone has no tax set

### LINKED ISSUE #6384: Tax calculation error when Basis of product tax is set to store and Display prices with tax is on.
In last GitHub version, with 'Basis of product tax' set to 'Store' and 'Display Prices with Tax' is set to 'true', and if store zone has not tax rate set.
When client zone (different from shop) has a tax rate set it will be used which is questionable... No tax should be applied in this case.
Then, if displaying prices with tax, tax is calculated as a 'gross price'. Displayed value is like 'actual tax - tax of the tax'...

For example:
Set shop zone to Alabama and 'Basis of Product Tax' to 'Shop', then 'Display prices with Tax' to true and customer address to Florida (where 7% tax apply).
Add product 25 'Microsoft keyboard' in cart and go through checkout.
Tax will be 4.58 instead of 4.90.
Strangely, 4.58 + (4.58 * .07) = 4.90...

When 'Display prices with Tax' is set to false, then tax is 4.90. But anyway in both case there should be no tax at all

piloujp:
Fixes #6384


drbyte:
I don't think this fixes the root problem.

This PR will simply apply no tax, instead of reverting to the store's tax zone.

piloujp:
ZC is actually programmed to apply store tax (When tax basis is 'Store') only if client is in same zone. Then if client is in another zone, it applies client's zone tax (actually with error in calculation). If it is meant to be like this, then setting tax basis to store is exactly the same as setting it to shipping.
I did this PR so it applies no tax when client zone is different than store one. Another possibility would be to apply store tax whatever client zone is.
When store has no tax zone, no tax is applied either.

Sorry, but I don't understand when you write ` instead of reverting to the store's tax zone.`, reverting from which situation are you referring to?

piloujp:
I thought this PR was an acceptable solution, the only argument against it was cleared in issue #6384 or was it?
To be clear here is what it does depending on zone settings, it is as stated on explanation text for this setting (apply store tax only if client is in same zone):
store zone ---- client zone ----- applied tax
not set               not set                 none
set                     not set                 none
not set               set                       none
set                     set, diff. zone        none
set                     set, same zone      yes, store tax

Of course it could be modified if requested. Actually it does this;
store zone ---- client zone ----- applied tax
not set               not set                 none
set                     not set                 none
not set               set                       yes, client tax
set                     set, diff. zone        yes, client tax
set                     set, same zone      yes, store/client tax
Which is exactly identical to 'Basis of Shipping Tax' set to 'Shipping'.

piloujp:
After few months, I reviewed this PR with new ZC code and it is not relevant anymore. It is time to close.
Although there are still some problems with actual code displaying tax when no tax is applied...

PR #6394, opened 2024-04-11
Switch terminology of Gross/Net prices to indicate Tax "included" vs "excluded"

piloujp:
In admin product edit page, gross and net prices are reversed. It should be:
Gross ---> Tax excluded
Net ---> Tax included


dbltoe:
Is it different in your area?  In the US, gross would represent all the proceeds from a sale, and net would represent the money left after all fees are removed from the sale(s).  Neither actually involves taxes.  Total sales is where tax finally appears, and it would be the net plus taxes.

i.e., 
Gross Sales - $22,815.95*
Returns - $320.53
Discounts - $116.82
Net Sales - $22,378.60*
Taxes - $1848.61
Total Sales - $24,227.21#

* Neither figure contains tax
# Tax is included.

piloujp:
Speaking of sales, I agree with you except about taxes that I have seen included in net sales very often, but both are possible depending where you are.
But here we speak of product prices, a gross price and a net price for a product. In French we say prix brut (gross?) and prix net (tax included).
In ZC product edit page, gross price includes tax and net price does not, which seems completely reversed to me.
How would you define gross price and net price?

torvista:
>How would you define gross price and net price?

Net is what is left at the end after adding or subtracting whatever...your net salary is what you actually get to spend.
Hence gross is the starting/base amount before doing anything.

scottcwilson:
The terms "gross" and "net" are from the American perspective, where taxes are not included.   You "gross up" an amount to deal with taxes.

dbltoe:
Gross is always equal to or greater than net.

IMHO, the shopping cart is not using gross or net.  It's uses sub-total and total.  The sub-total does not include tax or shipping, while the total contains the product cost, shipping, and taxes (if applicable).  That total becomes the gross for accounting purposes because it is the largest amount received in the sale.  After fees are paid, taxes are removed, and the actual cost of shipping is applied, the amount finally going to the store owner is the net income.  Even that net can be further defined.  When the cost of the product is deducted from net income, the remainder is net profit.

torvista:
>You "gross up" an amount to deal with taxes.

I'd never heard of that, but from what I've read, that term is only applied to a **salary** where the net/final payment (money handed over) has been agreed, so the "gross up" is the backwards calculation made to cover the taxes/social security etc. and arrive at the starting "gross" payment.

I concur with the original post, specifically:
Gross ---> Amount before taxes deducted
Net ---> Amount after taxes deducted

dbltoe:
> > You "gross up" an amount to deal with taxes.
> 
> I'd never heard of that, but from what I've read, that term is only applied to a **salary** where the net/final payment (money handed over) has been agreed, so the "gross up" is the backwards calculation made to cover the taxes/social security etc. and arrive at the starting "gross" payment.
> 
> I concur with the original post, specifically: Gross ---> Amount before taxes deducted Net ---> Amount after taxes deducted

But the original post is the opposite.

>Gross ---> Tax excluded
>Net ---> Tax included

piloujp:
I know there are differences in the way things are seen between continents, but I just try to focus on product prices in a shop.
The American way is kind of a buyer perspective, you buy from a wholesaler (net) then you fix (gross up) your new price to cop with charges, tax and whatever to be sure to make profit. Right?
But when you sell, your price become a net price for your buyer. If you include a sale tax, would you not do it at this time? Tax actually depends on your buyer. This sale price might not be actually the price you fixed (gross). Won't you call this price a net price which actually correspond to the money you received?
When I was a kid in France, many shops were not displaying prices with taxes, which when added to a 30% sale tax did not make customers happy when it is time to pay. Law changed and shops had to display prices with tax included. As this is not what they really get, they although still displayed prices without taxes together. Since then (more than 30 years ago) these prices were clearly labeled gross prices (the one without tax) and net prices (the one included tax). Then Europe came and there were different currencies, but this is another story.
Same things appended in Japan more recently, but language is so different that I can't tell you sure that prices are called gross or net.
Then whatever it is about, sales on tax return, salary or sales in a shop, net prices are always what is left at the end and gross prices what you start with.
In ZC product edit page there are two prices, one without tax and one with tax included. Whatever the system you are in, gross price can't be including sale tax. Sale tax is a kind of special tax, you receive it from your customer and send it to tax authority later, like a proxy. It should not change your profit (not directly), how could you account for it when deciding (grossing up) your product price?
Anyway, a good solution for ZC would be not to use 'gross' and 'net' terms for products prices, and just replace them by tax excluded and tax included.

torvista:
I excelled myself with late-night gibberish again. Apologies.
I meant to say
Gross: base/unmodified amount/**EX**cluding tax e.g. 100
Net: final/modified amount/**IN**cluding tax e.g. 120

> The American way is kind of a buyer perspective, you buy from a wholesaler (net) then you fix (gross up) your new price to cop with charges, tax and whatever to be sure to make profit. Right?

I think it is incorrect to use that term "gross up" with respect to anything other than salaries since the amount you get in your pocket is **less** than the salary. e.g Gross salary 100, net payment 70.
This "gross up" is purely to derive a starting figure that will produce an agreed money-in-pocket amount.

With a product the gross price is 100 and the net price is **more**, e.g. 120.

So there is a gross and net figure per product, and you could also say gross and net on the totals of entire order, but it's always before and after tax.

So this is correct:
image

But this is incorrect:
image

and this discussion illustrates why it would be better to rename them incl/excl tax.

piloujp:
I did a new commit, removing all 'gross' and 'net' wording.
I hesitated, but renamed constant too, as this pr is for V2.0.0.

dbltoe:
What's wrong with just saying Sub Total for excluded and Total for included?

If not, then technically it should be Product's Price vs Products Price.  It's the OCD in me.  😃

drbyte:
Note: Zen Cart only stores the pre-taxed price of the product, even with tax-included-pricing enabled.
The higher price with "tax included" is simply present as an aid to back-calculate the price to store in the database.

Then during shopping it adds the tax into prices if tax-included-pricing is enabled. It also calculates and displays tax as a subtotal during checkout, whether in tax-included-pricing mode or not.

PR #6395, opened 2024-04-11
Ensure variable is set in checkout shipping

scottcwilson:
```
[11-Apr-2024 19:55:40 UTC] Request URI: /client/index.php?main_page=checkout_shipping, IP address: ::1, Language id 1
#0 /Users/scott/sites/client/includes/modules/pages/checkout_shipping/header_php.php(191): zen_debug_error_handler()
#1 /Users/scott/sites/client/index.php(35): require('/Users/scott/si...')
--> PHP Warning: Undefined array key "methods" in /Users/scott/sites/client/includes/modules/pages/checkout_shipping/header_php.php on line 191.

```

### DIFF
--- a/includes/modules/pages/checkout_shipping/header_php.php
+++ b/includes/modules/pages/checkout_shipping/header_php.php
@@ -188,7 +188,7 @@ if (isset($_SESSION['cart']->cartID)) {
   if (isset($_SESSION['shipping']['id'])) {
     $checklist = [];
     foreach ($quotes as $key=>$val) {
-      if (is_array($val['methods'])) {
+      if (isset($val['methods']) && is_array($val['methods'])) {
         foreach($val['methods'] as $key2=>$method) {
           $checklist[] = $val['id'] . '_' . $method['id'];
         }


scottcwilson:
(This is an error path; happens when USPS credentials are wrong.)

proseLA:
usps is not part of the base zen-cart core.

i would suggest that the fix should be there in that plugin.

[the docs say that the quote method](https://docs.zen-cart.com/dev/code/modules/shipping_modules/#anatomy-of-a-shipping-module) should return an array where the method element is also an array.

the docs also allow for a return of false.  which is captured by the `is_array`.

so clearly to me, the USPS module needs to be corrected, as opposed to further checks in the base code.

scottcwilson:
I think the code should be robust enough to handle a bad return value without falling over.   This is just a warning now but will likely be a fatal error in a future version of PHP.  We should ensure that checkout doesn't crash even in the event of a problem with one module.

proseLA:
yeah, i disagree.  i code to specs.  

i would think modules/plugins should not get approved if they do adhere to the specs.  and, in this case, the specs seem pretty clear.

scottcwilson:
> i would think modules/plugins should not get approved if they do adhere to the specs. and, in this case, the specs seem pretty clear.

a) Your understanding of the approval process is not accurate.  It doesn't work that way. 

I look for malware and violations of the rules.  I don't test plugins or do code inspections.  If I had to do that, nothing would ever get approved. 

b) Like I said, it's crazy to crash just because a module doesn't do the right thing. 

Anyway, you've expressed your opinion; Chris will decide what to do.

torvista:
My opinion is that external code should behave/follow the specs.
But, if it doesn't, core code should error_log it, but not fail completely.
I remember the spec about the array is a relatively recent definition, so perhaps belt and braces to prevent a shop breaking should be the priority.

lat9:
> ```
> [11-Apr-2024 19:55:40 UTC] Request URI: /client/index.php?main_page=checkout_shipping, IP address: ::1, Language id 1
> #0 /Users/scott/sites/client/includes/modules/pages/checkout_shipping/header_php.php(191): zen_debug_error_handler()
> #1 /Users/scott/sites/client/index.php(35): require('/Users/scott/si...')
> --> PHP Warning: Undefined array key "methods" in /Users/scott/sites/client/includes/modules/pages/checkout_shipping/header_php.php on line 191.
> ```

I'm not seeing such an error logged when I add an `&` to the account password.  I _am_ seeing the USPS response.

scottcwilson:
Failing password sent by Skype.

scottcwilson:
You did this test with the password I sent and no call to `htmlspecialchars` on the password? 
If so then go ahead and close until someone else can reliably repro.

lat9:
> You did this test with the password I sent and no call to `htmlspecialchars` on the password? If so then go ahead and close until someone else can reliably repro.

Yes and yes.

PR #6396, opened 2024-04-12
Fix for forgotten password email

scottcwilson:
Forgotten password doesn't pass in an array because no HTML email is sent.  So this log happens:

```
--> PHP Warning: Undefined array key "EMAIL_TO_ADDRESS" in /home/client/public_html/store/includes/functions/functions_email.php on line 626.
```

This PR fixes.

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -604,6 +604,9 @@
         if (is_array($content)) {
             $block = $content;
         } else {
+            if ($content === '' || $content === 'none') {
+               return '';
+            }
             $block['EMAIL_MESSAGE_HTML'] = $content;
         }
 


scottcwilson:
Same issue would happen with any email where "" or 'none' was passed in for the HTML block to `zen_mail`.

drbyte:
> `Undefined array key "EMAIL_TO_ADDRESS" in /includes/functions/functions_email.php on line 626.`

Which version of Zen Cart is this error quoted from?

scottcwilson:
I think 1.5.8a most likely.  I can check when I get home if it's important.

drbyte:
Please. 
Because I can't see where `EMAIL_TO_ADDRESS` is getting called strangely

drbyte:
While core code shouldn't trigger the situation you describe, we do make a comment in the password_forgotten module to optionally pass a blank string instead of an array for $html_msg.

But what you are replacing is the fallback for templates that support many additional fields that the proposed early-return could break.

IMO a safer solution is more like: 
```diff
 if (is_array($content)) {
            $block = $content;
        } else {
+           if ($content === '' || $content === 'none') return '';
            $block['EMAIL_MESSAGE_HTML'] = $content;
        }
```

scottcwilson:
ok will get this done when I get back.

scottcwilson:
Updated.

PR #6399, opened 2024-04-13
Correct new products query in 2.0.0

### LINKED ISSUE #6398: New Products not Displaying
**Describe the bug**
No new products were displaying anywhere, even though the time setting in New Product Listing - Limited to ... is 0

Upon searching I found

Code:
// $display_limit = zen_get_products_new_timelimit();
  $display_limit = zen_get_new_date_range();

**Version**
ZC 2.0
PHP 8.3
Boot 3.6.4

Switching to the commented out code brings all the new products to their parts (sidebox, centerbox, page, etc)

scottcwilson:
SHOW_NEW_PRODUCTS_LIMIT is an int, not a string #6399

Fixes #6398.

Removal of old `zen_get_products_new_timelimit` to be done as a follow on PR.

### DIFF
--- a/includes/functions/functions_products.php
+++ b/includes/functions/functions_products.php
@@ -179,10 +179,10 @@ function zen_get_new_date_range($time_limit = false)
 
     $zc_new_date = date('Ymd', $date_range);
     switch (true) {
-        case (SHOW_NEW_PRODUCTS_LIMIT === '0'):
+        case (SHOW_NEW_PRODUCTS_LIMIT === 0):
             $new_range = '';
             break;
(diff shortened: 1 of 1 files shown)


lat9:
My bad; I need to make a list of those Min/Max Values constants that are converted to `int` values for future use.  They bite me every time.

scottcwilson:
> My bad; I need to make a list of those Min/Max Values constants that are converted to `int` values for future use. They bite me every time.

Hang on ... this means we need documentation in that area!

scottcwilson:
https://docs.zen-cart.com/dev/code/constants/

PR #6403, opened 2024-04-16
Fix ANSI mode issues

scottcwilson:
As reported in https://www.zen-cart.com/showthread.php?230070-Testing-v2-0-0-Upgrade-Issues-Errors-When-Loading-Order-Details

### DIFF
--- a/admin/includes/attributes_preview.php
+++ b/admin/includes/attributes_preview.php
@@ -27,7 +27,7 @@ if (!defined('IS_ADMIN_FLAG')) {
 
     if ($pr_attr->fields['total'] > 0) {
       if (PRODUCTS_OPTIONS_SORT_ORDER=='0') {
-        $options_order_by= ' order by LPAD(popt.products_options_sort_order,11,"0")';
+        $options_order_by= " ORDER BY LPAD(popt.products_options_sort_order,11,'0')";
       } else {
         $options_order_by= ' order by popt.products_options_name';
       }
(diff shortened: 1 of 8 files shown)


lat9:
Here's a screenshot of all the hits on the `ORDER BY LPAD` usage.  You can, of course, safely ignore the hit on the `shopping_cart - Copy.php`:
image

scottcwilson:
ok updated.

scottcwilson:
The second argument is an integer; neither single nor double quotes are required.

proseLA:
> The second argument is an integer; neither single nor double quotes are required.

that's really interesting.  but you are changing all of the 3rd arguments.

IMO, all of the 3rd arguments should be done in single quotes.

scottcwilson:
Apparently counting is not a strength of mine today.  Thanks for finding this!

lat9 (changes requested):
`/includes/functions/functions_products.php` (in the `zen_products_sort_order` function) needs this change as well.

lat9 (approved):
Thanks, @scottcwilson, I'm confident that all 'bases' are now covered.

scottcwilson:
Did you mean to merge into master? I thought this was a bug fix.

lat9:
Crud, I missed the `2.0.x` label.

lat9:
Should I revert and then merge into `2.0`?

scottcwilson:
Let's sit tight and see what @drbyte says.  There is a 100% chance this is not the only time one of us will make a mistake, and we should have a recipe for fixing it.

PR #6404, opened 2024-04-16
Handle case of uptime command not available

scottcwilson:
As reported here: https://www.zen-cart.com/showthread.php?230071-RE-PHP-Warning-Undefined-array-key-0


lat9:
Holy crow!  Some webhosts _still_ allow `exec`?

PR #6407, opened 2024-04-20
bug fix for in cart check

proseLA:
such good var names, and yet errors occur...

this error started here 636d4a3acd8


scottcwilson:
Thanks for tracking this down.  Added to known bugs for 2.0.0 list.

PR #6409, opened 2024-04-20
refactor shipping modules

proseLA:
if approved i can reformat the modified shipping files.

### DIFF
--- a/includes/classes/ZenShipping.php
+++ b/includes/classes/ZenShipping.php
@@ -1,0 +1,79 @@
+<?php
+
+abstract class ZenShipping extends base
+{
+    /**
+     * $_check is used to check the configuration key set up
+     * @var int
+     */
+    protected $_check;
+    /**
+     * $code determines the internal 'code' name used to designate "this" shipping module
+     *
+     * @var string
+     */
+    public $code;
+    /**
+     * $description is a soft name for this shipping method
+     * @var string
+     */
+    public $description;
+    /**
+     * $enabled determines whether this module shows or not... during checkout.
+     * @var boolean
+     */
+    public $enabled;
+    /**
+     * $debug is an array containing debug information
+     * @var array
+     */
+    public $debug = [];
+    /**
+     * $icon is the file name containing the Shipping method icon
+     * @var string
+     */
+    public $icon;
+    /**
+     * $quotes is an array containing all the quote information for this shipping module
+     * @var array
+     */
+    public $quotes;
+    /**
+     * $sort_order is the order priority of this shipping module when displayed
+     * @var int
+     */
+    public $sort_order;
+    /**
+     * $tax_basis is used to indicate if tax is based on shipping, billing or store address.
+     * @var string
+     */
+    public $tax_basis;
+    /**
+     * $tax_class is the  Tax class to be applied to the shipping cost
+     * @var string
+     */
+    public $tax_class;
+    /**
+     * $title is the displayed name for this shipping method
+     * @var string
+     */
+    public $title;
+
+    protected $class_key;
+
+    abstract public function quote($method = ''): array;
+
+    abstract public function keys(): array;
+
+    abstract public function install(): void;
+
+    /**
+     * Remove the module's settings
+     *
+     */
+    public function remove() {
+        global $db;
+        $db->Execute("delete from " . TABLE_CONFIGURATION . " where configuration_key like '" . $this->class_key . "_%'");
+    }
+
+}
--- a/includes/classes/ZenShipping.php
+++ b/includes/classes/ZenShipping.php
@@ -59,8 +59,6 @@ abstract class ZenShipping extends base
      */
     public $title;
 
-    protected $class_key;
-
     abstract public function quote($method = ''): array;
 
     abstract public function keys(): array;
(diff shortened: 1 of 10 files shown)


includes/modules/shipping/perweightunit.php
```diff
@@ -75,6 +22,7 @@ function __construct() {
     $this->title = MODULE_SHIPPING_PERWEIGHTUNIT_TEXT_TITLE;
     $this->description = MODULE_SHIPPING_PERWEIGHTUNIT_TEXT_DESCRIPTION;
     $this->sort_order = defined('MODULE_SHIPPING_PERWEIGHTUNIT_SORT_ORDER') ? MODULE_SHIPPING_PERWEIGHTUNIT_SORT_ORDER : null;
+    $this->class_key = 'MODULE_SHIPPING_ITEM_PERWEIGHTUNIT';
```
lat9, line 25: Should be `MODULE_SHIPPING_PERWEIGHTUNIT`.

includes/modules/shipping/freeshipper.php
```diff
@@ -107,7 +56,7 @@ function update_status() {
       }
     }
 
-    function quote($method = '') {
+    function quote($method = ''):array {
```
lat9, line 59: Space between `:` and `array` needed.

includes/modules/shipping/freeoptions.php
```diff
@@ -243,11 +187,11 @@ public function check()
     public function get_configuration_errors()
     {
         if (!zen_check_for_misconfigured_downloads()) {
-            return TEXT_DOWNLOADABLE_PRODUCTS_MISCONFIGURED; 
+            return TEXT_DOWNLOADABLE_PRODUCTS_MISCONFIGURED;
         }
     }
 
-    public function install()
+    public function install():void
```
lat9, line 194: Space needed between `:` and `void`.

includes/modules/shipping/freeoptions.php
```diff
@@ -201,10 +145,10 @@ protected function checkForFreeOptions()
     // -----
     // Return the "Free Options" quote, as requested.
     //
-    public function quote($method = '')
+    public function quote($method = ''):array
```
lat9, line 148: Space needed between `:` and `array`.

includes/modules/shipping/flat.php
```diff
@@ -107,7 +56,7 @@ function update_status() {
       }
     }
 
-    function quote($method = '') {
+    function quote($method = ''):array {
```
lat9, line 59: Space needed between `:` and `array`.

includes/modules/shipping/storepickup.php
```diff
@@ -47,32 +20,7 @@ class storepickup extends base {
...
-    /**
-     * $title is the displayed name for this shipping method
-     * @var string
-     */
-    public $title;
-    
+
   /**
    * constructor
    *
```
lat9, line 26: Doesn't set the `class_key` to `MODULE_SHIPPING_STOREPICKUP`.

lat9 (changes requested):
Looking good; a couple of minor changes requested.

includes/classes/ZenShipping.php
```diff
@@ -71,7 +71,8 @@ abstract public function install(): void;
      * Remove the module's settings
      *
      */
-    public function remove() {
+    public function remove(): void
+    {
         global $db;
         $db->Execute("delete from " . TABLE_CONFIGURATION . " where configuration_key like '" . $this->class_key . "_%'");
```
lat9, line 77: One more thing, I'm _pretty sure_ that the `_` needs to be escaped: `"\_%'.  I ran into this when recently.

lat9 (changes requested):
Previous changes look good, just that escape of `_`.

includes/classes/ZenShipping.php
proseLA: can you provide a mysql mode setting where this is required?  i would like to test it to see how it would fail and then fix accordingly.

`$this->class_key` would probably need to get escaped as well.  and IMO, i would think this should get done in the `db` class. 

in looking at the code, it seems that `bind_vars` escapes as necessary.  

would this be more appropriate:

```php
global $db;
        $sql = "DELETE FROM " . TABLE_CONFIGURATION . " WHERE configuration_key LIKE :key";
        $sql = $db->bindVars($sql, ':key', $this->class_key . '_%', 'string');
        $db->Execute($sql);
```

lat9: I'd run across this when developing the `spam_cleanup_check` for v2.0.0.  When searching for `%_SRC_%`, I was getting _false positives_ for an email address like `########@src-zone.com`.

The use of `bindVars` looks like a great solution.

drbyte: It is true: with `LIKE` clauses, you must escape `_` characters with a preceding `\`, else the `_` is treated as a single-char wildcard.

drbyte: `bindVars` doesn't know when you're using a `LIKE` clause, so doesn't escape the `_`.

proseLA: ok, i now have a better understanding of the `_` in mysql situation.  as to whether a key may exist such as:

`MODULETSHIPPING8STOREPICKUP` that would fit the un-escaped clause is up for debate....  but it would now be addressed, if it did in fact exist.

lat9 (approved):

lat9:
That last update will make it even easier for shipping modules to use `ZenShipping`, one less variable to worry about.

PR #6411, opened 2024-04-20
Correct 'multiple taxes' calculation error

### LINKED ISSUE #6410: Multiple taxes calculation error when prices are displayed with tax.
In last github version of ZC, when multiple taxes apply to a product and option to display prices with tax is set to true, calculated taxes are wrong. When displaying prices without tax, they are OK.

**To Reproduce**
Add a new tax rate with Florida as zone.
Set 'Display prices with Tax' to true.
Set 'Show Split Tax Lines' to true (not necessary, but easier to check).
Log in client side and use an address in Florida.
Add a product to cart and go to last checkout screen.
Check taxes

I have a solution I will PR soon.

piloujp:
fixes #6410 
Correct calculation error when prices are displayed with tax and multiple taxes apply.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -877,7 +877,7 @@ class order extends base
                     'subtotal' => 0,
                 ];
             }
-            $this->info['tax_subtotals'][$tax_description]['subtotal'] += $shown_price;
+            $this->info['tax_subtotals'][$tax_description]['subtotal'] += $product_final_price * $product_qty + $product_onetime_charges;
         }
     }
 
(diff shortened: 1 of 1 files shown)


drbyte:
Upon initial review, I don't believe this "fixes" any calculation error.

This PR just treats prices as always without tax, regardless of the DISPLAY_PRICES_WITH_TAX setting.

Remember: DISPLAY_PRICES_WITH_TAX means "tax-must-be-added-back-in-so-it-is-included", because product prices are always stored in the database *without* tax.

PR #6417, opened 2024-04-25
Enable `template_settings.php` to be viewed in admin

lat9:
If a template has a `template_settings.php`, enable it to be viewed in the admin.

### DIFF
--- a/admin/template_select.php
+++ b/admin/template_select.php
@@ -5,269 +5,330 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: neekfenwick 2023 Dec 09 Modified in v2.0.0-alpha1 $
  */
-require('includes/application_top.php');
+require 'includes/application_top.php';
 
-if (isset($_GET['tID'])) $selected_template = (int)$_GET['tID'];
-$action = (isset($_GET['action']) ? $_GET['action'] : '');
+if (isset($_GET['tID'])) {
+    $selected_template = (int)$_GET['tID'];
+}
+$action = $_GET['action'] ?? '';
 $template_info = zen_get_catalog_template_directories();
 
 if (!empty($action)) {
-  switch ($action) {
-    case 'insert':
-      $selected_template = zen_register_new_template($_POST['ln'], $_POST['lang']);
-      $action = '';
-      break;
+    switch ($action) {
+        case 'insert':
+            $selected_template = zen_register_new_template($_POST['ln'], $_POST['lang']);
+            $action = '';
+            break;
 
-    case 'save':
-      zen_update_template_name_for_id($selected_template, $_POST['ln']);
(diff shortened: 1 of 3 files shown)


admin/template_select.php
```diff
@@ -188,8 +188,42 @@
...
                  $contents[] = ['text' => TEXT_INFO_TEMPLATE_VERSION . $template_info[$tInfo->template_dir]['version']];
                  $contents[] = ['text' => TEXT_INFO_TEMPLATE_DESCRIPTION . '<br>' . $template_info[$tInfo->template_dir]['description']];
-                 $contents[] = ['align' => 'text-center',
-                    'text' => '<a href="' . zen_href_link(FILENAME_TEMPLATE_SELECT, 'page=' . $_GET['page'] . '&tID=' . $tInfo->template_id . '&action=edit') . '" class="btn btn-primary" role="button">' . TEXT_INFO_EDIT_INTRO . '</a>' .
+                 if ($template_info[$tInfo->template_dir]['has_template_settings'] === true) {
+                     $template_settings_button =
+                        '<button type="button" class="btn btn-info" data-toggle="modal" data-target="#view-settings">' .
+                            TEXT_VIEW_TEMPLATE_SETTINGS .
+                        '</button> ';
+                     $template_settings = file_get_contents($template_info[$tInfo->template_dir]['template_path'] . '/template_settings.php');
```
drbyte, line 196: Is there any chance of this coming back as boolean `false`? (eg: fail to open/load file?)

lat9: I guess there's a possibility; the update to the template-function checks that the file exists but it could be unreadable.

lat9:
If that base update is now acceptable, I'll follow up with some reformatting/refactoring prior to merge.

drbyte:
I wonder if it makes sense to wrap the output in `<code>` tags so it presents as monospaced, like it is IRL?

lat9:
I tried that, but `<code>` tags in the admin display as that ugly red/pink color.

drbyte:
> I tried that, but `<code>` tags in the admin display as that ugly red/pink color.

Oh, yuck.   Maybe `<pre>`? 

I was just feeling like without monospaced font it's less apparent that these are code-related settings, and not intended to be super "readable", etc.

lat9:
Oooh!  `<samp>` works!

> Also, depending on screen-width, the buttons show up like below. I wonder if there should just be an intentional `<br>` between them, so they're forced to new line and a specific bit of padding?
> Screen Shot 2024-04-25 at 5 52 24 PM Screen Shot 2024-04-25 at 5 52 32 PM

There are many, many more aberrant displays in the admin if you choose to do your business on a phone.

drbyte:
> I tried that, but `<code>` tags in the admin display as that ugly red/pink color.

Comes from Bootstrap!

drbyte:
> There are many, many more aberrant displays in the admin if you choose to do your business on a phone.

True

drbyte:
> If that base update is now acceptable, I'll follow up with some reformatting/refactoring prior to merge.

👍

PR #6431, opened 2024-05-02
Common site map building code

scottcwilson:
Some pages were missing from the "page not found" template.  Let's use a single source of content for both the sitemap and page not found to fix this.


drbyte:
Do storeowners typically want a complete site-map on the 404-page-not-found page?

dbltoe:
We still have the not found page but add a 404.shtml to the root of hosted sites. 

It's not a short page codewise, but we add a quick link to home and to report an error.  Just customize it to the site's colors.  One local CSS error, needs to be upgraded to the latest HTML, but gets a green light on accessibility.

I've been thinking about adding the Ajax search to get the customer to a product's add-to-cart button a little quicker. 😃 

haredoDOTcom/nothing

scottcwilson:
> Do storeowners typically want a complete site-map on the 404-page-not-found page?

What's the point of picking and choosing?  Show them what's available.   If a storeowner doesn't like the entire list, they can change it.

drbyte:
I'd rather we re-think what the page-not-found does, and make it leaner instead of bigger: 
it would probably be better to have it simply offer a search box instead of pulling in the site map.

If someone arrives at a "not found" URL, it's primarily because they're following a link posted externally ... so offering a site map of all pages and products and categories isn't giving a speedy solution for them. 
We could examine the URL they visited and perhaps offer that as search terms pre-filled-in in the search box.

scottcwilson:
OK, then this is not a good starting point.  Closing.

barco57:
https://searchengineland.com/404-pages-best-practices-examples-436618

Issue #6425, opened 2024-04-27
OT modules and taxes calculations


piloujp:
In ZC last V2.0.0, after doing few PRs about Order Total modules, especially discount modules, it appears to me that with actual logic of these modules, it is impossible to apply all possible options and get correct results. There are different problems I identified:

1- Discount modules (discount coupons, GV and group pricing) are treated as 'charges' (fee) modules. They have their own tax class option like they were taxable items, which in my opinion is a non-sens. The items/fees they apply to have tax classes, but the discount in itself can't! There should not be any tax class to choose in discount modules.

2- When applying a discount, there are three options to re-calculate tax, 'None', 'Standard' and 'Credit Note'.
The first one, 'None' does not make any sens, especially with percentage discounts. How do you do to not re-calculate tax when applying a 10% discount, for example? When you apply it to total without tax, you get a result without tax. If you apply it to a total with tax included, your result includes tax, reduced from 10%. It is automatic ! This option should not exist in discount modules.
For other options, 'Standard' and 'Credit Note', I decided to interpret them as 'discount amount is price incl.' and 'discount amount is price excl.' when discount is a fixed amount. With percentage, it does not change a thing.

3- Option 'Include tax' was to be use only when no tax recalculation is done... See above. This one should not be there either.

4- When using multiple tax rates or even multiple tax classes with multiple tax rates, details for shipping tax rates are missing, making it impossible to use multiple tax rates as soon as shipping has two or more taxes applied. Shipping taxes for different tax rates should be saved.

I concentrated on all these points and did a rewrite of OT coupons. Discount ones have been heavily modified.
Now I get good results with multiples taxes, even compound, (tested with 4) and possibles options left in discount modules. Additional charge modules although perform well.
Files for now are available on [branch Multiple-taxes-support-in-ot-_modules](https://github.com/piloujp/zencart/tree/Multiple-taxes-support-in-ot-_modules) of my GitHub repo.
I am not sure of what to do with this: Make a new PR and deleting PR#6392 or updating this old PR with new files. Another problem is that this files won't pass some of ZC validation tests and these test conditions are kind of obscure which makes it hard to detect if error comes from new code or was in the test already.

piloujp:
Here are some test results I got:
First setting I used:
There are 2 tax class with bot 2 compound tax rates each.
Tax class 1:  Tax rate 1 → 7%,  tax rate 2 → 14%. Equivalent to one tax rate at 21.98%.
[...]

PR #6432, opened 2024-05-05
Multiple taxes support in ot modules

piloujp:
Rewrite of Order Total modules, mainly discount modules, to support different modules options and multiple tax classes and rates.
Replace PR #6392 
See issue #6425 for more details.
This PR needs PR #6411 and is complementary to PRs #6391 and #6393 and optionally #6394 .
All options concerning calculation have been tested thoroughly. Options for validation are partially tested.

It can't pass actual tests because there is a new array in 'order->info' and some options have been suppressed too.


zc_install/sql/install/mysql_zencart.sql
```diff
@@ -2880,12 +2876,10 @@ INSERT INTO configuration (configuration_title, configuration_key, configuration
...
 INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, last_modified, date_added, use_function, set_function) VALUES ('This module is installed', 'MODULE_ORDER_TOTAL_TOTAL_STATUS', 'true', '', 6, 1, NULL, '2003-10-30 22:16:55', NULL, 'zen_cfg_select_option(array(\'true\'),');
 INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, last_modified, date_added, use_function, set_function) VALUES ('Sort Order', 'MODULE_ORDER_TOTAL_TOTAL_SORT_ORDER', '999', 'Sort order of display.', 6, 2, NULL, '2003-10-30 22:16:55', NULL, NULL);
-INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, last_modified, date_added, use_function, set_function) VALUES ('Tax Class', 'MODULE_ORDER_TOTAL_COUPON_TAX_CLASS', '0', 'Use the following tax class when treating Discount Coupon as Credit Note.', 6, 0, NULL, '2003-10-30 22:16:36', 'zen_get_tax_class_title', 'zen_cfg_pull_down_tax_classes(');
```
drbyte, line 2883: I'm trying to get my mind around the removal of these settings.

But if we do indeed remove them, then the `DELETE` statements for them should be added to the (recently committed) `mysql_upgrade_zencart_210.sql` file.

drbyte:
Looking at the test failure for Low Order Fee, it appears as though this new code is simply not charging any tax. The price mismatch is the amount of the tax: $43.49 instead of $46.01 ... the difference is $2.52
Screen Shot 2024-05-05 at 4 39 21 PM
https://github.com/zencart/zencart/blob/dac1dbc10fb300b891b49b4486fa16636ae38159/not_for_release/testFramework/FeatureStore/LowOrderFees/LowOrderFeeTest.php#L249-L254

drbyte:
@piloujp Thanks for all your work on this. .... lots to review!

piloujp:
> Looking at the test failure for Low Order Fee, it appears as though this new code is simply not charging any tax. The price mismatch is the amount of the tax: $43.49 instead of $46.01 ... the difference is $2.52 Screen Shot 2024-05-05 at 4 39 21 PM
> 
> https://github.com/zencart/zencart/blob/dac1dbc10fb300b891b49b4486fa16636ae38159/not_for_release/testFramework/FeatureStore/LowOrderFees/LowOrderFeeTest.php#L249-L254

I really don't know, I get 46.01 in my test cart, which is the actual GitHub ZC with only this PR applied.
Settings are as follows:
Shipping(2.5) and low order fee(5) have no tax class.
Group discount of 10% does not include shipping ->-4$
tax is 7%. Florida zone with store and customer in Florida and tax basis set to 'shipping'.
If I apply a GV of -46.01, result is 0$.

piloujp:
I found the problem. It because of the reversed calculation of taxes 'Credit Note' <=> 'Standard' between original code and new one. Problem is probably the same with other test failing. When test is using 'Standard' recalculation, I get the same results if I use 'Credit Note'.

includes/modules/order_total/ot_loworderfee.php
```diff
@@ -106,25 +111,38 @@ function process()
...
+                    $tax_multiple = zen_get_multiple_tax_rates(MODULE_ORDER_TOTAL_LOWORDERFEE_TAX_CLASS, $tax_address['country_id'], $tax_address['zone_id']);
+                    $tax_description = zen_get_tax_description(MODULE_ORDER_TOTAL_LOWORDERFEE_TAX_CLASS, $tax_address['country_id'], $tax_address['zone_id'], true);
+
+                    $low_order_tax = zen_calculate_tax($low_order_fee, $tax);
+                    $order->info['tax'] += $low_order_tax;
+                    foreach ($tax_description as $key => $value) {
+                        if (!isset($order->info['tax_groups'][$value])) {
+                            $order->info['tax_groups'][$value] = 0;
+                        }
+                        if (!isset($order->info['tax_groups'][$value])) {
```
drbyte, line 133: Did you mean this instead?
```suggestion
                        if (!isset($order->info['tax_subtotals'][$value]['subtotal'])) {
```

includes/modules/order_total/ot_shipping.php
```diff
@@ -109,32 +114,49 @@ public function process()
...
+                            $shipping_tax_description = zen_get_tax_description($GLOBALS[$module]->tax_class, $order->billing['country']['id'], $order->billing['zone_id'], true);
                         } elseif (STORE_ZONE == $order->delivery['zone_id']) {
                             $shipping_tax = zen_get_tax_rate($GLOBALS[$module]->tax_class, $order->delivery['country']['id'], $order->delivery['zone_id']);
-                            $shipping_tax_description = zen_get_tax_description($GLOBALS[$module]->tax_class, $order->delivery['country']['id'], $order->delivery['zone_id']);
+                            $shipping_tax_multiple = zen_get_multiple_tax_rates($GLOBALS[$module]->tax_class, $order->delivery['country']['id'], $order->delivery['zone_id']);
+                            $shipping_tax_description = zen_get_tax_description($GLOBALS[$module]->tax_class, $order->delivery['country']['id'], $order->delivery['zone_id'], true);
                         } else {
-                            $shipping_tax = 0;
+                            $shipping_tax_multiple[] = 0;
```
drbyte, line 137: Am I reading correctly that `$shipping_tax_multiple` should probably be initialized to an empty array up around line 101-102? Otherwise if it hits this "else" case it will be undefined and calling it with `[]` here will throw a warning.

includes/modules/order_total/ot_coupon.php
```diff
@@ -148,28 +131,41 @@ function process()
...
+                $order->info['shipping_tax'] -= $od_amount['ShippingTax'];
+            }
             $tax = 0;
+            // Update all tax rates in order object
             foreach ($order->info['tax_groups'] as $key => $value) {
                 if (isset($od_amount['tax_groups'][$key])) {
                     $order->info['tax_groups'][$key] -= $od_amount['tax_groups'][$key];
-                    $order->info['tax_groups'][$key] = zen_round($order->info['tax_groups'][$key], $currencies->get_decimal_places($_SESSION['currency']));
                     $tax += $od_amount['tax_groups'][$key];
+                    $order->info['tax_subtotals'][$key]['subtotal'] -= $od_amount['total'];
```
drbyte, line 144: I'm not understanding this.
Why reduce `tax-subtotals` by `$od_amount['total']`, which is the entire discount amount... instead of only by the amount of tax relevant to that specific tax_subtotal?

includes/modules/order_total/ot_shipping.php
piloujp: Yes, you are right, it should be initialized. I will add `$shipping_tax_multiple = [];` on line 102.

drbyte:
> I found the problem. It is because of the reversed calculation of taxes 'Credit Note' <=> 'Standard' between original code and new one. The problem is probably the same with other test failing. When test is using 'Standard' recalculation, I get the same results if I use 'Credit Note'.

Checking: do you have a potential fix in mind?

includes/modules/order_total/ot_coupon.php
piloujp: This subtotal is not a tax. Array info['tax_subtotals'] contain the pair 'Tax rate'-'Order subtotal on which tax apply' for each tax class. With these pairs it is possible to recalculate tax value for each tax rate. Array tax_groups is similar and contain tax rate names and values. The second one is used a lot in modules codes, but I am not sure tax_subtotals array is used much.

drbyte: Oh. Yes. Thanks. I knew I was missing something :D

drbyte:
Several configuration settings have been removed which previously allowed specific tax handling for GV/Coupon/GroupPricing, including the ability to turn tax off altogether for them. 
(Albeit, I think they were imperfectly implemented, and most stores rarely touched those settings.)

Is it accurate to say that now whatever tax rates apply to a customer's order, are also being applied to their GV's and coupons? 
And that tax rules are forced by ot discount modules, without option to bypass.

The description now reads `Discount amount is tax included -> Standard, if tax excluded -> Credit Note`. But I'm not sure I know how to explain that to a storeowner. 
1 - does "amount is tax included" mean "if my store uses tax-included pricing I must choose Standard"? Or does it mean "the discount that will be shown to customer already includes taxes"?
2 - similarly, "if tax excluded->Credit Note" ... does that mean "if I want to exclude tax handling on the discount, choose Credit Note"? or does it mean "the discount shown will not yet have taxes applied yet, but they will be assessed in subsequent calculations"?

drbyte:
I've updated the Unit Tests to make them show the reason why the assertions are failing:
Screen Shot 2024-05-06 at 7 46 22 PM

drbyte:
I've also shortened the scanned HTML for (several of) the Feature Tests to make it easier to find the relevant part of the HTML that's being asserted upon.

piloujp:
> Is it accurate to say that now whatever tax rates apply to a customer's order, are also being applied to their GV's and coupons? And that tax rules are forced by ot discount modules, without option to bypass.

Roughly yes, but your wording is not accurate, see below.

> The description now reads `Discount amount is tax included -> Standard, if tax excluded -> Credit Note`. But I'm not sure I know how to explain that to a storeowner. 1 - does "amount is tax included" mean "if my store uses tax-included pricing I must choose Standard"? Or does it mean "the discount that will be shown to customer already includes taxes"?

Both.

> 2 - similarly, "if tax excluded->Credit Note" ... does that mean "if I want to exclude tax handling on the discount, choose Credit Note"? or does it mean "the discount shown will not yet have taxes applied yet, but they will be assessed in subsequent calculations"?

The second one.

Well, I think the problem is that discounts are being seen as another item like those in cart or like charges added after, but in my opinion it is wrong.
For a programmer, it would be like mixing property and method in an object.
Let say the invoice is the parent object. Inside you have children objects like items in cart, shipping module, c.o.d. module and low order fee module. They all have their properties like amounts (totals, subtotals), tax class, tax rates, values and descriptions, and methods to calculate these values (shipping cost, tax value...).
Invoice object has its methods too, like discount modules, tax module and total module. These methods just modify different objects properties. Discounts modules reduce values in a predefined way, tax module retrieves taxes descriptions and values to display them and total just displays the final total.

It is why I removed options like tax class, include tax and re-calculate 'none'.
Plus, applying a tax to a discount would mean having negative tax! Never heard of it... Tax authority given away money... it would be nice!
Taxes don't apply to discounts, they apply to items and charges which are discounted (or not). It is why taxes should always be re-calculated when a discount is applied.

Actually, in ZC, tax class in discount modules were used to recalculate taxes (in Credit Note option) in a way they replace taxes already applied. Big problem here, if everything was at 7% tax rate and then a 20% rate is used, you get a discount very different from what was announced.

Finally, option 'include tax' is still partially a mystery. It was only possible to activate it when re-calculate was set to 'none'.

For options 'Credit Note' and 'Standard', wording could be better. To be precise, 'Credit Note' is when discount is applied before tax and 'Standard' is when discount is applied after tax of each item they apply to.

I hope this clarifies things a bit.

piloujp:
> I've updated the Unit Tests to make them show the reason why the assertions are failing: Screen Shot 2024-05-06 at 7 46 22 PM

Very good idea. Looking at test results, I could not see what was the actual error.

piloujp:
> I've updated the Unit Tests to make them show the reason why the assertions are failing: Screen Shot 2024-05-06 at 7 46 22 PM

Test 1 and 2 easily fixed with a rounding: `round($result[0]['value'], 2)`

Test 3, I think, is wrong. Shipping cost has been discounted 100% and should be 0. Before I modified the code, shipping cost stayed untouched until the end, but what do you do when multiple discounts including shipping are applied? The second discount should be applied on shipping cost discounted from the first, if not you might end up with a discount on shipping higher than original shipping cost.

Test 4 has the same problem as 3. The no tax calculation makes it looks like when set to Credit Note, but there might be some differences.

Test 5 has a reversed option problem. I got 501 for discount if set to Credit Note, but 468.3767.... when set to Standard.

Test 6 -> reversed option problem too.

Test 7 free shipping calculation is a little different. Free shipping discount is added at the end to discount without shipping. Before, they were mixed together. I don't know where the test is taking discount amount, but value correspond at discount before adding shipping discount.

Tests with fixed discount are mixed up by reversed options, but tests with no tax calculation option that does not exist are ok because treated as Credit Note option.

drbyte:
> Well, I think the problem is that discounts are being seen as another item like those in cart or like charges added after, but in my opinion it is wrong.

I see what you're saying, but I don't think it's wrong.

In accounting, it's important to (at least be able to) treat discounts as separate values, because they "could" (if the storeowner chose) be treated as marketing/advertising/promotion expenses.

---

Aside:
The current ot infrastructure is quite rigid because it's still mostly based on the osC roots from which it was first derived.

We've rewritten (offline) it in a way that makes orders super flexible and manageable, but haven't put it into core yet because every plugin that deals with orders and totals, possibly even payment, will be affected. 
It treats orders as a bunch of line-items, consisting of products, surcharges, discounts, shipping, taxes, etc, and then payments against it, and even returns/exchanges/refunds and additional payments, etc, are all easily incorporated. Accounting departments like it because every number affecting the total has an explanation. 
And never any penny rounding issues, because there's only one source of truth and one set of calculations.
It pretty much does away with all the pains we're discussing in this PR.
But there's a big cost to such a change. Including incompatibility of old order data.

drbyte:
> For options 'Credit Note' and 'Standard', wording could be better. To be precise, 'Credit Note' is when discount is applied before tax and 'Standard' is when discount is applied after tax of each item they apply to.

So, you're saying with this PR "Credit Note" and "Standard" have new meanings: now they just mean "(apply discount) before tax" and "(apply discount) after tax"?

piloujp:
> > Well, I think the problem is that discounts are being seen as another item like those in cart or like charges added after, but in my opinion it is wrong.
> 
> I see what you're saying, but I don't think it's wrong.
> 
More than the way discount are seen, I meant the consequences of it are wrong, looking at all these options actually available but not usable by store owners.
 
> In accounting, it's important to (at least be able to) treat discounts as separate values, because they "could" (if the storeowner chose) be treated as marketing/advertising/promotion expenses.
> 
I completely agree with that and don't see any incompatibility here... ZC lacks a little bit of traceability for accounting, but I think my PRs go in the right direction.

> Aside: The current ot infrastructure is quite rigid because it's still mostly based on the osC roots from which it was first derived.
> 
> We've rewritten (offline) it in a way that makes orders super flexible and manageable, but haven't put it into core yet because every plugin that deals with orders and totals, possibly even payment, will be affected. It treats orders as a bunch of line-items, consisting of products, surcharges, discounts, shipping, taxes, etc, and then payments against it, and even returns/exchanges/refunds and additional payments, etc, are all easily incorporated. Accounting departments like it because every number affecting the total has an explanation. And never any penny rounding issues, because there's only one source of truth and one set of calculations. It pretty much does away with all the pains we're discussing in this PR. But there's a big cost to such a change. Including incompatibility of old order data.

It sounds too good to be true.

piloujp:
> > For options 'Credit Note' and 'Standard', wording could be better. To be precise, 'Credit Note' is when discount is applied before tax and 'Standard' is when discount is applied after tax of each item they apply to.
> 
> So, you're saying with this PR "Credit Note" and "Standard" have new meanings: now they just mean "(apply discount) before tax" and "(apply discount) after tax"?

Yes, and it is a meaning easy to understand for any store owner. After using ZC for years as a store owner and more recently digging deeply into its code, I still can't see any meaning in the following discount modules options:
Include tax
Tax class
Re-calculate set to 'None'
Re-calculate set to 'Credit Note'

And I am pretty sure that I am not alone.
It seems to me that some of these options are there as a quick fix to deal with complicated situations that normally ZC can not deal with, but I won't be surprise to be wrong on this.

torvista:
> We've rewritten (offline) it in a way that makes orders super flexible and manageable...It pretty much does away with all the pains we're discussing in this PR....but there's a big cost to such a change

Better that time is invested in progress than all the time I see invested/wasted in trying to polish an oSC t*rd that seems to be  unfixable with all the price/discount/tax options available.

Break plugins....listen to people complaining about what functionality they have lost....put it on a roadmap....ask people to take on those tasks. Organised progress.

piloujp:
I do not remember closing this PR. Perhaps it happened when merging with ZC new commits?

PR #6435, opened 2024-05-07
Coupon notify

scottcwilson:
Add notifier to coupon product check.

### DIFF
--- a/includes/classes/CouponValidation.php
+++ b/includes/classes/CouponValidation.php
@@ -14,6 +14,7 @@ class CouponValidation
     public static function is_product_valid(int $product_id, int $coupon_id): bool
     {
         global $db;
+        global $zco_notifier; 
 
         $product_id = (int)$product_id;
 
@@ -23,7 +24,7 @@ class CouponValidation
 
         $coupons = $db->Execute($coupons_query);
 
-        $product_query = "SELECT products_model FROM " . TABLE_PRODUCTS . "
+        $product_query = "SELECT * FROM " . TABLE_PRODUCTS . "
                           WHERE products_id = $product_id";
 
         $product = $db->Execute($product_query);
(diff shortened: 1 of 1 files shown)


drbyte:
The test is suggesting that there's no response from the browser, which might mean a blank page.
I wonder if there's a problem with referencing `$product->fields` maybe?
I guess technically since the query only brings back `products_model` we could (at the query, before the notifier stuff altogether) set `$product_model` to the result of the query, or to blank if not found. And then use that in the subsequent `if` clause, and then the notifier could just use that string instead of `$product->fields`.

drbyte:
Another observation: 
Should the $coupon_id be passed to the observer too? 
Or were you only expecting this notifier to check whether or not to allow coupons at all for the product, instead of the "is this product valid for this coupon" that the wrapping function is doing?

(Completely separate, and will handle separately, I think the $coupons var obscures meaning, and would like to add: `$restrictions = $coupons->fields;` and then change all the `$coupons->fields[]` references to `$restrictions[]` in all the rest of this is_product_valid() function we're discussing here.)

scottcwilson:
> Should the $coupon_id be passed to the observer too?

Good idea.  Done.

includes/classes/CouponValidation.php
```diff
@@ -32,6 +32,12 @@ public static function is_product_valid(int $product_id, int $coupon_id): bool
             return false;
         }
 
+        $product_can_use_coupon = true; 
+        $this->notify('NOTIFY_COUPON_ADDITIONAL_CHECKS', $product->fields, $coupon_id, $product_can_use_coupon);
```
drbyte, line 36: `$this->notify` is not available in this class.

Will have to use $zco_notifier instead:
`global $zco_notifier;` at the top of the function
and use `$zco_notifier->notify` instead of `$this->notify`

scottcwilson: geez I thought i had fixed that.  Sorry.

drbyte: Pretty sure that's why the test is failing: syntax error causing blank page, so test is looking for a product price in a blank string.

scottcwilson: Yep.  I implemented this in a 2.0 upgrade and thought I had copied my work to the github branch, but I think I fell behind. 

Fixed now.

PR #6442, opened 2024-05-09
deduction is  needed by local sales tax mod

scottcwilson:


lat9:
I've got no clients that use that mod, in what manner does the "Local Sales Tax" _need_ access to that value?

scottcwilson:
I reduce the tax by the amount of tax that would have been charged on the discount.  (So the local tax is charged on the subtotal less discounts.)

lat9:
Noting that this change should also be applied to the 2.0.x branch; that variable's defined as `protected` there, too.

scottcwilson:
Yep, sorry, fat fingered the destination.

lat9:
One more (I promise!) comment: This change is required due to the changes introduced [here](https://github.com/zencart/zencart/commit/c8ca5b03f856e70b3922f82e15f1c065e56a7894) to pre-define all class variables in zc158.

scottcwilson:
Yes!  The specification of visibility is a new thing and we're figuring it out as we go.

PR #6447, opened 2024-05-12
Fix attempted INSERT IGNORE on every customer login.

### LINKED ISSUE #6444: Very minor: customers_info_date_account_created should be date_account_created in Customer.php
Very minor issue, I noticed the field name is wrong in Customer.php.  I'm not sure if this has any actual impact on the system.

https://github.com/zencart/zencart/blob/8eb0b099d98ec053954694cb21cd31d33831790b/includes/classes/Customer.php#L190-L193

While the database field is indeed `customers_info_date_account_created`, the `data` element being referenced here is `date_account_created`.  The result is that the if always tests true, so the `INSERT IGNORE` is always attempted.

Anyone making mods to `Customer.php` might want to include this little fix.

neekfenwick:
Fixes #6444 
Simple bug mixing up internal data array names with database field names.  Tested OK by stepping through in VS Code.

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -187,7 +187,7 @@ class Customer extends base
 
         // @TODO - delete this if we collapse the Info table
         // enforce db integrity: make sure related record exists
-        if (empty($this->data['customers_info_date_account_created'])) {
+        if (empty($this->data['date_account_created'])) {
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks!

lat9:
Should this have gone to branch 2.0 as well?

PR #6450, opened 2024-05-14
Add explicit null allowed for function type definitions.

brittainmark:
Prevent depreciation warning in PHP8.4

### DIFF
--- a/includes/functions/php_polyfills.php
+++ b/includes/functions/php_polyfills.php
@@ -86,7 +86,7 @@ if (!function_exists('json_validate')) {
 }
 
 if (!function_exists('mb_str_pad') && function_exists('mb_substr')) {
-    function mb_str_pad(string $string, int $length, string $pad_string = ' ', int $pad_type = \STR_PAD_RIGHT, string $encoding = null): string
+    function mb_str_pad(string $string, int $length, string $pad_string = ' ', int $pad_type = \STR_PAD_RIGHT, ?string $encoding = null): string
     {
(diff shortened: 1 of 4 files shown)


lat9:
> Prevent depreciation warning in PHP8.4

Would you post the contents of one of these PHP Deprecated logs, so someone running into this condition has a better clue?

includes/functions/php_polyfills.php
```diff
@@ -86,7 +86,7 @@ function json_validate(string $json, int $depth = 512, int $flags = 0): bool
 }
 
 if (!function_exists('mb_str_pad') && function_exists('mb_substr')) {
-    function mb_str_pad(string $string, int $length, string $pad_string = ' ', int $pad_type = \STR_PAD_RIGHT, string $encoding = null): string
+    function mb_str_pad(string $string, int $length, string $pad_string = ' ', int $pad_type = \STR_PAD_RIGHT, ?string $encoding = null): string
```
drbyte, line 89: This line will never be hit under PHP 8.4 due to the version check on line 62.

lat9:
> > Prevent depreciation warning in PHP8.4
> 
> Would you post the contents of one of these PHP Deprecated logs, so someone running into this condition has a better clue?

Found this, based on lack of response: https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated

brittainmark:
Apologies I have only just seen your request @lat9. Yes, you found the correct notification. I was trying to pre-empt the issue for the future.

PR #6453, opened 2024-05-15
Add notifier hook to admin dashboard to allow custom widgets

### LINKED ISSUE #6448: Add notify to index_dash.php
Can we add
$zco_notifier->notify('NOTIFY_ADMIN_DISPLAY_WIDGET' , colNum);
to the end of each column so that plugins can add boxes to the dashboard easily

drbyte:
Notifier hook `'NOTIFY_ADMIN_DASHBOARD_WIDGETS'` allows widget list to be altered and appended-to. 
Can specify `'column'`, `'sort'`, `'visible'` and `'path'` properties for each widget.

Fixes #6448

An encapsulated plugin will need to register its widget by specifying the complete widget directory path, including version number.

### DIFF
--- a/admin/index_dashboard.php
+++ b/admin/index_dashboard.php
@@ -1,15 +1,52 @@
 <?php
 /**
- * @copyright Copyright 2003-2022 Zen Cart Development Team
+ * @copyright Copyright 2003-2024 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
- * @version $Id: DrByte 2020 Jul 10 Modified in v1.5.8-alpha $
+ * @version $Id: DrByte  Modified in v2.0.1 $
+ *
+ * @var notifier $zco_notifier
  */
 
 if (empty($currencies)) {
     require_once DIR_WS_CLASSES . 'currencies.php';
     $currencies = new currencies();
 }
+
+$widgets = [];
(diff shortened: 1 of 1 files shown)


lat9:
Cool, since that will not only enable additional widgets to be inserted but also to have the "built-in" ones removed if desired.

admin/index_dashboard.php
```diff
@@ -39,41 +65,30 @@
...
-        ?>
-
-        <?php
-        include DIR_WS_MODULES . 'dashboard_widgets/OrderStatusDashboardWidget.php';
-        ?>
-
+    <?php
+    foreach ($widgets as $widget) {
+        if ($widget['column'] === 1) {
+            include $widget['path'];
```
lat9, line 71: Why `include` instead of `require`?

lat9 (commented):
See comment re: `include` vs. `require`.

scottcwilson:
> Cool, since that will not only enable additional widgets to be inserted but also to have the "built-in" ones removed if desired.

Were you thinking "Set `column` field to -1 to hide a widget" - or perhaps something different?

admin/index_dashboard.php
drbyte: Mainly just because that's what it was already using.

I tend to use `include` when I don't care if the loading of the file fails. Here I'm content to let the script continue even if the file fails to `include`; whereas calling `require` would emit a hard failure error and halt execution.

scottcwilson:
> An observer could `unset` a widget if they wanted it excluded/hidden.
> 
> Or we could add a `'visibility'` property.

I like `visibility` or `display` - I think it's more self-documenting.

drbyte:
Updated with `'visible'` as a `boolean`.
Also updated to set a default `'sort'` if one not provided, so that the multisort doesn't fail.

PR #6456, opened 2024-05-16
Provide additional script helpers for encapsulated plugin mgmt

drbyte:
Built as a trait for now. Could incorporate directly into the parent class.

Any `ScriptedInstaller.php` class that an encapsulated plugin runs has access to these helpers:

```
public function getConfigurationKeyDetails(string $key_name, bool $only_check_existence = false): array|bool
public function addConfigurationKey(string $key_name, array $properties): int
public function updateConfigurationKey(string $key_name, array $properties): int
public function deleteConfigurationKeys(array $key_names): int
public function addConfigurationGroup(string $group_title, array $properties): int
public function updateConfigurationGroup(int $group_id, array $properties): int
public function deleteConfigurationGroup(int $group_id): int
public function getConfigurationGroupDetails(int|string $group, bool $only_check_existence = false): array|bool
```

Example use shown in included DisplayLogs v3.0.2 installer class.

### DIFF
--- a/includes/classes/PluginSupport/ScriptedInstallHelpers.php
+++ b/includes/classes/PluginSupport/ScriptedInstallHelpers.php
@@ -1,0 +1,234 @@
+<?php
+/**
+ * @copyright Copyright 2003-2024 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id:      $
+ *
+ */
+
+namespace Zencart\PluginSupport;
+
+use queryFactory;
+
+trait ScriptedInstallHelpers
+{
+    protected queryFactory $dbConn;
+
+    /**
+     * Get details of current configuration record entry, false if not found.
+     * Optional: when $only_check_existence is true, will simply return true/false.
+     */
+    public function getConfigurationKeyDetails(string $key_name, bool $only_check_existence = false): array|bool
+    {
+        $sql = "SELECT * FROM " . TABLE_CONFIGURATION . " WHERE configuration_key = '" . zen_db_input($key_name) . "'";
+        $result = $this->dbConn->Execute($sql, 1);
+
+        // false if not found, or if existence-check fails
+        if ($only_check_existence || $result->EOF) {
+            return !$result->EOF;
+        }
+
+        return $result->fields;
+    }
+
+    public function addConfigurationKey(string $key_name, array $properties): int
+    {
+        $exists = $this->getConfigurationKeyDetails($key_name, true);
+        if ($exists) {
+            return 0;
+        }
+
+        $fields = [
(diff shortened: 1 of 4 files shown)


drbyte:
Updated to ignore duplicates when adding new configuration keys, and fixed sql error calling delete of config key.

includes/classes/PluginSupport/ScriptedInstallHelpers.php
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        return $result->fields;
+    }
+
+    public function addConfigurationKey(string $key_name, array $properties): int
+    {
+        $exists = $this->getConfigurationKeyDetails($key_name, true);
+        if ($exists) {
+            return 0;
```
lat9, line 38: Shouldn't this return the `configuration_id` if the configuration-key exists (i.e. `$exists` is an array)?

lat9:
> > With that modification to `DisplayLogs`, shouldn't its version number -- and thus its directory name -- be changed as well?
> 
> I wasn't planning to. There's no functional change to the actual plugin. Also, the ScriptedInstaller class for the plugin was identical to its prior version, before applying these alterations. The plugin itself is identical.

While there's no `functional` change to the plugin, it's been refactored which, IMO, should lead to a different version number.

lat9:
Sometimes, when a plugin is converted to an `encapsulated` one, the author wants to keep existing configuration settings from the plugin's current installation, as I've done for [Printable Price List](https://github.com/lat9/printable_price_list/tree/4.0.0/zc_plugins/PrintablePriceList/v4.0.0).

Here's what I've currently got for the additions to that plugin's `ScriptedInstaller`:
https://github.com/lat9/printable_price_list/blob/77b388091a277182ba16f7f97d031772cbcdb32a/zc_plugins/PrintablePriceList/v4.0.0/Installer/ScriptedInstaller.php#L154-L180

Those added methods enable the plugin's installer to:
- Retrieve the results of a MySQL "SELECT" query
- Use a single method to either install or update the plugin's "Configuration Group", retrieving the `configuration_group_id` for use in follow-on configuration "INSERT IGNORE" operations.

Note, too, the use of "INSERT IGNORE" instead of "INSERT" on those configuration settings so that the existing configuration settings are kept.

includes/classes/PluginSupport/ScriptedInstallHelpers.php
drbyte: `exists` will only be a `bool` in this case because of the 2nd param sent on line 36.
My thinking was that if it already exists, there won't be an insert, so returning 0 is an easy indicator of that (and `INSERT IGNORE` wouldn't return the id either).

But yes we could change it up and return the found ID without reporting any clue that it already existed.

PR #6462, opened 2024-05-18
Enable zc_plugins to provide AJAX content

lat9:
Currently, only files present in `/includes/classes/ajax` are loaded to handle AJAX requests.

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -44,8 +44,6 @@ function ajaxAbort($status = 400, $msg = null)
 }
 // --- support functions ------------------
 
-
-
 if (!isset($_GET['act']) || !isset($_GET['method'])) {
     ajaxAbort();
 }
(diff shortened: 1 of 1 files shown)


lat9:
Quick (and sneaky) way to validate:
1. Copy `/includes/classes/ajax/zcAjaxAdminDatePickerDateCheck.php` to `/zc_plugins/DisplayLogs/v3.0.2/catalog/includes/classes/ajax/zcAjaxAdminDatePickerDateCheck.php`
2. Rename `/includes/classes/ajax/zcAjaxAdminDatePickerDateCheck.php` to `/includes/classes/ajax/zcAjaxAdminDatePickerDateCheck.php.notused`
3. Log into the admin, navigate to **Tools :: Banner Manager**, edit a banner and supply an invalid date and see that the 'invalid date' message still appears: 
image
4. Remove the file from the DisplayLogs plugin's directory.
5. Rename the base file back to  `/includes/classes/ajax/zcAjaxAdminDatePickerDateCheck.php`

Note: The DisplayLogs plugin, v3.0.2, must be installed via **Modules :: Plugin Manager** for the above to operate correctly!

ajax.php
```diff
@@ -44,23 +44,25 @@ function ajaxAbort($status = 400, $msg = null)
...
 $className = 'zc' . ucfirst($_GET['act']);
-$classFile = $className . '.php';
-$basePath  = DIR_FS_CATALOG . DIR_WS_CLASSES;
-
-if (!file_exists(realpath($basePath . 'ajax/' . basename($classFile)))) {
-    ajaxAbort();
+$classFile = basename($className . '.php');
+$classPath = DIR_WS_CLASSES . 'ajax/';
+$basePath  = DIR_FS_CATALOG;
+if (file_exists(realpath($basePath . $classPath . $classFile))) {
```
torvista, line 57: file_exists demands a string, and realpath may return false.

drbyte: Ref #6500

PR #6469, opened 2024-05-22
Add category description to notifier

### LINKED ISSUE #6467: index/category listing, final notifier for categories_description correct?
https://github.com/zencart/zencart/blob/198b15002b53a534fc9fbd3907415f4612ded6a7/includes/modules/pages/index/main_template_vars.php#L230-L235

I was looking at modifying the categories description and see this notifier is after the template is loaded/after the category description is used.

scottcwilson:
Fixes #6467

### DIFF
--- a/includes/modules/pages/index/main_template_vars.php
+++ b/includes/modules/pages/index/main_template_vars.php
@@ -227,7 +227,7 @@ if ($current_categories_name == '' && isset($_GET['music_genre_id'])) {
                            WHERE music_genre_id = " . (int)$_GET['music_genre_id'] . " LIMIT 1");
   if (!$result->EOF) $current_categories_name = $result->fields['music_genre_name'];
 }
-$zco_notifier->notify('NOTIFY_HEADER_INDEX_MAIN_TEMPLATE_VARS_PAGE_BODY', NULL, $tpl_page_body, $current_categories_name);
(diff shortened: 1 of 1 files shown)


torvista:
Should not also $current_categories_description be removed from
$zco_notifier->notify('NOTIFY_HEADER_END_INDEX_MAIN_TEMPLATE_VARS', NULL, $current_categories_description);
since it has no effect if modified?

scottcwilson:
No.  We want to avoid not-upwardly-compatible changes to notifers.

torvista:
Understood as a general rule, but given a previous comment to make the function of a notifier self-explanatory, leaving this here implies it can be changed, when it can't, so it is misleading.

scottcwilson:
It *can* be changed.  The change just won't show up in the template.  It could be used somewhere else, who knows.

drbyte:
I wonder if the placement of the original notifier is the real problem here.
Yes, our standard practice is to leave most final notifier calls at the bottom of these header_php and main_template_vars files .... but in this sort of case where tpl_page_body is called manually (as is the case in all main_template_vars files), there's little point in the notifier hook coming after it. 
I haven't looked at the overall code-base to see how drastic this sort of pattern correction might be.

scottcwilson:
Let's not overcorrect a non-problem based on a speculation.  The original issue is resolved.  Unless there's an actual problem with the last notifier, this should be sufficient.

PR #6471, opened 2024-05-22
Make allowance for null products model

scottcwilson:
Products model is a nullable field.

--> PHP Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /Users/scott/sites/client/includes/functions/functions_prices.php on line 32.

### DIFF
--- a/admin/products_price_manager.php
+++ b/admin/products_price_manager.php
@@ -817,6 +817,10 @@ if (!empty($action)) {
                   <div class="col-sm-9 col-md-6 text-center">
                     <?php
 // Specials cannot be added to Gift Vouchers when false
+                    // prevent log on null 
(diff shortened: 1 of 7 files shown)


lat9:
Should this be going to 'master' to '2.0'?  It feels like it should be '2.0'.

scottcwilson:
As discussed, we will merge to master but label 2.0.2 just in case we go back and do another 2.0 build.

lat9:
Noting that each of these changed files will need to be touched again for https://github.com/zencart/zencart/pull/6370 (or that PR will need change once it's rebased).

scottcwilson:
Please merge this first so the backport to 2.0.2 (if needed) is trivial.

PR #6472, opened 2024-05-22
Products model could be null

scottcwilson:
--> PHP Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /home/account/catalog/includes/functions/functions_products.php on line 733.


lat9:
Should this be going to 'master' to '2.0'? It feels like it should be '2.0'.

scottcwilson:
Closing, will put all these changes in #6471.

PR #6473, opened 2024-05-22
Allow Notifier deprecations to be registered

drbyte:
Adds the following methods to Notifier and Observer base traits:
`registerDeprecatedEvent($old, $new)`
`registerObserverAlias($old, $new)`

### DIFF
--- a/includes/classes/traits/NotifierManager.php
+++ b/includes/classes/traits/NotifierManager.php
@@ -7,47 +7,66 @@
 
 namespace Zencart\Traits;
 
-use  Zencart\Events\EventDto;
+use Zencart\Events\EventDto;
 
 trait NotifierManager
 {
     /**
      * @var array of aliases
      */
-    private $observerAliases = ['NOTIFIY_ORDER_CART_SUBTOTAL_CALCULATE' => 'NOTIFY_ORDER_CART_SUBTOTAL_CALCULATE'];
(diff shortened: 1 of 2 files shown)


includes/classes/traits/NotifierManager.php
```diff
@@ -126,16 +147,26 @@ protected function logNotifier($eventID, $param1, $param2, $param3, $param4, $pa
         error_log($zcDate->output("%Y-%m-%d %H:%M:%S") . ' [main_page=' . $main_page . '] ' . $eventID . $output . "\n", 3, $file);
     }
 
-    private function eventIdHasAlias($eventId)
+    private function eventIdHasAlias($eventId): bool
     {
         if (array_key_exists($eventId, $this->observerAliases)) {
             return true;
         }
+
```
lat9, line 155: Why not just `return array_key_exists($eventId, $this->observerAliases);` directly?

lat9 (commented):
Looks good, just one small (and optional) refinement.

includes/classes/traits/NotifierManager.php
drbyte: Ya, I was thinking the same. It was an existing method so I didn't change it. But I agree. Will update.

drbyte: Updated.

PR #6475, opened 2024-05-23
function_prices: cast constant/string to int for number_format

torvista:
this file is close to working in strict mode

### DIFF
--- a/includes/functions/functions_prices.php
+++ b/includes/functions/functions_prices.php
@@ -216,7 +216,7 @@ function zen_get_products_display_price($product_id)
             if ($display_sale_price) {
                 if (SHOW_SALE_DISCOUNT === '1') {
                     if ($display_normal_price != 0) {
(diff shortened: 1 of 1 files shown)


scottcwilson:
Nice.  I will update https://docs.zen-cart.com/dev/code/constants/
on next checkin.

includes/functions/functions_prices.php
```diff
@@ -216,7 +216,7 @@ function zen_get_products_display_price($product_id)
             if ($display_sale_price) {
                 if (SHOW_SALE_DISCOUNT === '1') {
                     if ($display_normal_price != 0) {
-                        $show_discount_amount = number_format(100 - (($display_sale_price / $display_normal_price) * 100), SHOW_SALE_DISCOUNT_DECIMALS);
```
drbyte, line 219: Should this also be done on line 245?

PR #6476, opened 2024-05-23
fix for PHP Fatal error:  Uncaught TypeError: number_format(): Argument #1 ($num) must be of type int|float, string given

torvista:
last fix to allow strict mode


drbyte:
This doesn't look like a complete fix to me.
There are many other places where `$special_price` are used in this function, without being converted this way. As well as many other price-related variables.

And touching "all" of those price-related variables in this way is super-messy. 

And all because strict-type mode disables PHP's natural safe coercive type conversion from string to numeric.

torvista:
ok, live and learn, back to my onions.

PR #6477, opened 2024-05-24
cast SHOW_SALE_DISCOUNT_DECIMALS to int

torvista:


scottcwilson:
There are two uses of this constant in this file.

lat9:
What's the problem being solved here?  A log that shows the resulting PHP warning/error before the correction would be appreciated.

torvista:
error message was along the lines of number_format parameter #2 should be an integer.

PR #6487, opened 2024-05-25
Fix PHP 8.1 syntax in ipncheck.php utility

scottcwilson:
From PHP 8,1 and later, certain changes to the $GLOBALS array throws fatal errors. In general, modifying individual array elements is allowed, but making mass changes are not allowed.

Current code throws a fatal error which shows as an HTTP 500.


drbyte:
What's the fatal error that's thrown?

scottcwilson:
```
PHP Fatal error:  $GLOBALS can only be modified using the $GLOBALS[$name] = $value syntax in ipncheck.php on line 37
```

PR #6488, opened 2024-05-25
Add switch for legacy "_" additional_image matching bypass

drbyte:
For issue discussed at https://www.zen-cart.com/showthread.php?230077-Problem-with-Product-Images&p=1401219

additional-images-legacy


drbyte:
Documentation:

> Prior to Zen Cart 2.1.0, image matching did not enforce the check of a `'_'` suffix when looking for additional product images in the "images/" directory (but it DID require the '_' in any "images/whatever/" subdirectories). Since v2.1.0 the "_" is required in both locations. Stores that are upgrading and wish to retain the old functionality can enable the legacy mode via the `site_specific_settings.php` file by setting `$use_legacy_additional_image_matching` to `true`.

includes/extra_datafiles/dist-site_specific_overrides.php
```diff
@@ -86,8 +86,17 @@
...
+//
 // true .... Display the legacy clickable column headings to sort the product listing (in addition to the sorter dropdown)
 // false ... Do not show the column headings; this is the default.
 //
 //$show_table_header_row = false;
+
+// -----
+// Product "Additional" Images: whether to enforce "_" underscore pattern matching when not in subdirectories
+//
+// true .... Use legacy (pre v2.1.0) mode of matching "{base}{suffix}.{ext}" in main images dir but force the underscore in all subdirs
```
lat9, line 98: So, we're going to penalize all the sites that have "followed the rules" regarding additional-image naming for the past decades when product images are present in the `/images` "root" directory?

This setting's name and default should, IMO, be reversed since those preexisting "rule-followers" are going to be wondering where the heck all their additional images disappeared to.

includes/modules/additional_images.php
```diff
@@ -36,6 +36,14 @@
         $products_image_base = $products_image_match;
     }
 
+    // Unless legacy mode is turned on, force the use of a '_' suffix when detecting additional images NOT in a subdirectory
+    if (!empty($use_legacy_additional_image_matching)) {
+        $products_image_base .= '_';
+    }
```
lat9, line 42: See commentary above.

lat9 (dismissed):
This is a serious backward-compatibility nightmare.  See comments.

includes/extra_datafiles/dist-site_specific_overrides.php
torvista: Was there a rule? This bug has only just been noticed after decades!

When they notice their images have gone, it's only a one-line edit, and it flags them to sort it out.

IMHO this "inconsistency" is really a bug and so the correction should be the default.

Just put this item first on the list of the breaking changes in the update docs, maybe with a blink tag.

lat9: > Was there a rule? This bug has only just been noticed after decades!
> 

https://github.com/zencart/documentation/blob/f680cfa9e14d3e3d151138cb8aad7b3014c7d0ae/content/user/images/image_filename_conventions.md

> When they notice their images have gone, it's only a one-line edit, and it flags them to sort it out.
> 

It's going to be a support _nightmare_.

> IMHO this "inconsistency" is really a bug and so the correction should be the default.
> 

You're entitled to your opinion.

> Just put this item first on the list of the breaking changes in the update docs, maybe with a blink tag.

scottcwilson: I agree with Cindy.  The default should be the current loosey-goosey, and the option should be strict.   This way a very small number of people are affected.

drbyte:
IMO it's a bug of inconsistency. And the fix should be built-in for all new installs.
And if we never draw a line in the sand, then the smarter/better standard will never become "the norm".

Upgraders can easily flip the switch.

barco57:
"And if we never draw a line in the sand, then the smarter/better standard will never become "the norm"."
drbyte - May 27, 2024

Lat9 can see the smile on my face as she knows I will repeat this.... but drbyte said....

dbltoe:
Just another view here.  Just like the grocery store has no idea how many times someone stopped at the shelf for an item and found it missing, we have no real idea of how many store owners are or have been victims of the "loosey-goosey" settings.

We have counseled our clients on image naming conventions since 1.2.x.  I also cannot think of a single site we've upgraded that didn't have to be adjusted in some way to not have an image problem.

I can only imagine what a grocery chain would pay to know about the empty shelf scenario.  They rely so heavily on the data for stocking.  If I can get something developed, we'll all be rich.

With the advent of IH and other products, the need for _med and _lrg doesn't really exist.  Even bootstrap does a really decent job 
without IH or any other product.  That puts more emphasis on the main image and more danger if the owner puts the image in the root image folder.

As it currently stands, adding car.jpg to the images folder opens you up to 672 possible "loosey-goosey" matches. (Thank you Scrabble Words)  We have no way of knowing how many have said, "If the additional images are going to be this much trouble, I'm gone."

I can only imagine what a grocery chain would pay to know about the empty shelf scenario.  They rely so heavily on the data for stocking.  If I can get something developed, we'll all be rich.

Through the efforts of some of you great minds ouut there, I am really seeing unprecedented growth in the operation of Zen Cart.  Emphasis on performance and compliance can do nothing but help.

Even with a "vested interest", I'm not sure having a feature work in all but one of the possible areas is worth keeping.

lat9:
> IMO it's a bug of inconsistency. And the fix should be built-in for all new installs. And if we never draw a line in the sand, then the smarter/better standard will never become "the norm".
> 
> Upgraders can easily flip the switch.

I accede, although if this was a "proper" configuration setting the initial install could set the "new" way and an upgrade could set the "old" way.  Here's hoping that this won't become an "I told you so" decision.

drbyte:
> if this was a "proper" configuration setting the initial install could set the "new" way and an upgrade could set the "old" way.

True. We "could" add an "upgrade-only" configuration switch (that wouldn't exist if the store were never upgraded from an older version), and honor that. 
I don't love the idea of adding configuration switches to turn on a bug though, so prefer not to make such a thing part of the main install script.

> Here's hoping that this won't become an "I told you so" decision.

Agreed.

scottcwilson:
If we add a config, let's do upgrade and new install please - otherwise just too confusing.  New install would be "strict;" upgrade would be "loosey goosey."

drbyte:
After considering a bunch of scenarios and trying various approaches, I've acceded to the suggestions of NOT making this a `site-specific-settings.php` flag, but instead to make it an Admin Configuration switch.
Fresh installs will default to "strict" matching mode, and upgrades will default to "legacy" matching mode.
additional-images-legacy

PR #6489, opened 2024-05-26
Improve curltester output

scottcwilson:
The first piece of output is 

```
Error Connection uses TLS 1.2
```

This is confusing.  The error is not that the connection uses TLS 1.2.

Changed to 
```
Checking server against howsmytls.com
Error (Rerun with "?d=1" for details)
Connection uses TLS 1.2
```

### DIFF
--- a/extras/curltester.php
+++ b/extras/curltester.php
@@ -23,7 +23,7 @@ ini_set('display_errors', 1);
 set_time_limit(500);
 
 $showDetails = (isset($_GET['d']) && $_GET['d'] != '0') || (isset($_GET['details']) && $_GET['details'] != '0');
-$errorMessage = '<span style="color:red;font-weight:bold">Error </span>';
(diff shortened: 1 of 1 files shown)


drbyte:
> The error is not that the connection uses TLS 1.2.

So what is the error?

scottcwilson:
That's what's confusing!  The error is totally separate from the TLS version.
You get an error if old ciphers are allowed.  (But you need to run with the parameter 'd' to get this detail). 
That's why I wanted to separate out things.

extras/curltester.php
```diff
@@ -41,7 +41,7 @@
         $data = curl_exec($ch);
         curl_close($ch);
         $json = json_decode($data, false);
-        echo (stristr($json->rating, 'Okay')) ? $goodMessage : $errorMessage;
+        echo (stripos($json->rating, 'Okay') !== false ? $goodMessage : $errorMessage) . ' Rating: ' . $json->rating;
```
scottcwilson, line 44: Better!

PR #6491, opened 2024-05-26
Refactor ipncheck

drbyte:
Hasn't been touched in a long time ... this does some cleanup


scottcwilson:
Great!  I will add to the release notes that we have cleaned up and improved the several of the extras files.

PR #6498, opened 2024-05-28
Pass `$oldVersion` to zc_plugins' `executeUpgrade` method

### LINKED ISSUE #6497: `ScriptedInstaller` class doesn't pass `$oldVersion` down
When writing an encapsulated plugin, on a plugin upgrade it's _normally_ helpful to know what version is currently installed.  Unfortunately, the `$oldVersion` doesn't currently get passed to a `zc_plugin`'s `executeUpgrade` function.

1. The progression starts with `admin/plugin_manager.php`'s call to `PluginManagerController->processActionDoUpgrade`.
2. That method determines and passes a plugin's old version to `BasePluginInstaller->processUpgrade`.
3. That method passes the `$oldVersion` to `Installer->executeUpgraders`, which passes the value to its `executeScriptUpgraders` method.
4. That method passes the `$oldVersion` to `ScriptedInstaller->doUpgrade`.  Unfortunately, that method doesn't indicate that it takes any inputs and subsequently invokes `ScriptedInstaller->executeUpgrade` which also doesn't take any parameters.

Adding that parameter to both of those `ScriptedInstaller` methods:
https://github.com/zencart/zencart/blob/5b9f527b016cc22355966e8d14673608c4c3b972/includes/classes/PluginSupport/ScriptedInstaller.php#L37-L41
https://github.com/zencart/zencart/blob/5b9f527b016cc22355966e8d14673608c4c3b972/includes/classes/PluginSupport/ScriptedInstaller.php#L53-L56

would enable a `zc_plugin` to "know" what version it was being upgraded from, but that breaks the function signature so that any (**are there any?**) existing plugins that use the existing method signatures will "break":
```
28-May-2024 07:41:08 America/New_York] PHP Fatal error:  Declaration of ScriptedInstaller::executeUpgrade() must be compatible with Zencart\PluginSupport\ScriptedInstaller::executeUpgrade($oldVersion) in C:\xampp\htdocs\zc200w\zc_plugins\POSM\v5.0.1\Installer\ScriptedInstaller.php on line 174
```

lat9:
Fixes #6497

### DIFF
--- a/includes/classes/PluginSupport/ScriptedInstaller.php
+++ b/includes/classes/PluginSupport/ScriptedInstaller.php
@@ -34,9 +34,9 @@ class ScriptedInstaller
         return $uninstalled;
     }
 
-    public function doUpgrade()
+    public function doUpgrade($oldVersion)
     {
-        $upgraded = $this->executeUpgrade();
+        $upgraded = $this->executeUpgrade($oldVersion);
         return $upgraded;
     }
 
@@ -50,7 +50,7 @@ class ScriptedInstaller
         return true;
     }
 
-    protected function executeUpgrade()
+    protected function executeUpgrade($oldVersion)
     {
         return true;
     }


drbyte:
Yes: any older encapsulated plugins will need to update the function signature (even if the variable isn't used) if they've declared it in the old syntax.
There's no way around it. And PHP will throw a fatal error (blank screen, log generated) as soon as the file is loaded ... which I think only occurs when firing Install or Upgrade via one of the Plugin Manager buttons.

PR #6509, opened 2024-05-31
`install_suggested` and `phpupgrade`, correct missing constant warning/error

scottcwilson:
In PHP8, we get a warning about this constant because the base language file isn't pulled in. 
Let's just remove the constant.

```
[31-May-2024 11:53:23 UTC] PHP Warning:  Use of undefined constant HTML_PARAMS - assumed 'HTML_PARAMS' (this will throw an Error in a future version of PHP) in /home/client/public_html/store/includes/templates/template_default/templates/tpl_zc_phpupgrade_default.php on line 19
```


includes/templates/template_default/templates/tpl_zc_install_suggested_default.php
```diff
@@ -14,7 +14,7 @@
 $docsPath = (file_exists('docs/index.html')) ? 'docs/index.html' : (file_exists('../docs/index.html') ? '../docs/index.html' : '');
 ?>
 <!DOCTYPE html>
-<html <?php echo defined('HTML_PARAMS') ? HTML_PARAMS : '';?>>
```
lat9, line 17: Instead of removing, why not just use the default `dir="ltr" lang="en"` since the hard-coded language is English?

PR #6518, opened 2024-06-05
Fix Undefined array key log

scottcwilson:


drbyte:
Is there a particular typical circumstance where that `'id'` element intentionally doesn't exist?

scottcwilson:
I don't know, but if it doesn't exist, it sure won't match some other value so there's no need to go into that branch.  This is PHP8 behavior; in PHP7, it would have just skipped over this.  Now to get the behavior we want to you have to ensure it's set or provide a blank value.

PR #6519, opened 2024-06-06
Product class - linked products count

### LINKED ISSUE #6516: admin, category-product listing: all products marked as linked
since this is always >0
https://github.com/zencart/zencart/blob/8c52b72bbeceb1c39233443aafe43e72a141612c/includes/classes/Product.php#L244

drbyte:
Fixes #6516 by treating master_categories_id as "not" a category to which the product is linked

### DIFF
--- a/includes/classes/Product.php
+++ b/includes/classes/Product.php
@@ -243,7 +243,11 @@ class Product
         $results = $db->Execute($sql, null, true, 900);
         $data['linked_categories_count'] = $results->RecordCount();
         $data['linked_categories'] = [];
-        foreach($results as $result) {
+        foreach ($results as $result) {
+            if ($result['categories_id'] === $data['master_categories_id']) {
+                $data['linked_categories_count']--;
+                continue;
+            }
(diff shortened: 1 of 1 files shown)


includes/classes/Product.php
```diff
@@ -244,6 +244,10 @@ protected function loadProductDetails(int $product_id, ?int $language_id = null)
         $data['linked_categories_count'] = $results->RecordCount();
         $data['linked_categories'] = [];
         foreach($results as $result) {
+            if ($result['categories_id'] === $data['master_categories_id']) {
+                $data['linked_categories_count'] --;
```
lat9, line 248: No space between a unary operator and its operand.

lat9 (changes requested):
PSR-12 change requested

includes/classes/Product.php
drbyte: Space removed before the `--`

lat9 (approved):

PR #6525, opened 2024-06-06
QUANTITY_DECIMALS: Look for qty that *doesn't* include '.'

### LINKED ISSUE #6517: Product quantity with 0 as the last digit not being accepted
ZC 2.0
PHP 8.3
Store with fractional quantities. 
QUANTITY_DECIMALS constant is set to 2.
Shopping cart quantity field trims all integer values ending in zero(s) to the number without the last zeros, e.g. 10 goes to 1 when updating the field. The issue seems to arise from the preg_replace function on lines 1370 and 2395 in includes/classes/shopping_cart.php.

lat9:
Fixes #6517

### DIFF
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -1361,7 +1361,7 @@ class shoppingCart extends base
 
             // convert quantity to proper decimals
             $precision = QUANTITY_DECIMALS > 0 ? (int)QUANTITY_DECIMALS : 0;
-            if ($precision === 0 || str_contains($data['qty'], '.')) {
+            if ($precision === 0 || !str_contains($data['qty'], '.')) {
                 $new_qty = $data['qty'];
             } else {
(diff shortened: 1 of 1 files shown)


lat9:
While I'm "in there", what's the purpose of the `preg_replace`?
https://github.com/zencart/zencart/blob/ace59217e87876cf2d8a3697de497f2a025b1796/includes/classes/shopping_cart.php#L1367

If it's just to lop off trailing zeros from the decimal portion of the quantity, why not simplify to just use `rtrim`?

drbyte (approved):

lat9:
Any comment on the `rtrim`?

drbyte:
Yes, it would appear that `rtrim()` should be suitable.

lat9:
For clarity, I'll create a separate PR for the `shopping_cart` and `order` class to make that `rtrim` simplification and bring the `order` handling to look like that in the `shopping_cart`.

PR #6526, opened 2024-06-06
shopping_cart/order: Simplification of QUANTITY_DECIMALS calculations

lat9:
A corollary to #6525.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -294,18 +294,9 @@ class order extends base
 
         while (!$orders_products->EOF) {
             // convert quantity to proper decimals - account history
-            if ($precision !== 0) {
-                $fix_qty = $orders_products->fields['products_quantity'];
-                switch (true) {
-                    case (false === strpos($fix_qty, '.')):
-                        $new_qty = $fix_qty;
-                        break;
-                    default:
-                        $new_qty = preg_replace('/[0]+$/', '', $orders_products->fields['products_quantity']);
-                        break;
-                }
-            } else {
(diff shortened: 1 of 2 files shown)


piloujp:
I am a little bit late (as usually), but testing for a decimal by looking for a dot won't work when the decimal separator is a comma.
Instead of `str_contains($new_qty, '.')` something like this would be better: `is_numeric($new_qty) && floor($new_qty) != $new_qty`

drbyte:
This is regarding database data. This dot-detection has also been in core code for many years and hasn't been a problem.

Is your database actually storing the decimal separator as a comma?

piloujp:
Your are right... at least for this code, But there are problems in ZC when you try to input a decimal using comma. Most of the forms are sanitized (to convert comma to dot) but not all.
Try to input a number like 1,5 in a product quantity box and click 'add to cart'. You will have a http 500 error. But this is not related directly to this PR I think.

piloujp:
`Try to input a number like 1,5 in a product quantity box and click 'add to cart'. You will have a http 500 error. But this is not related directly to this PR I think.`
My bad, it has been corrected by last updates...

piloujp:
`Is your database actually storing the decimal separator as a comma?`
No, In Japan, we use the dot. But few years ago I had a French language pack of ZC in use and I had can remember that I had some problems. I was with ZC 1.5.5 I think, not relevant anymore... PHP seems to use only dot as decimal separator anyway, which means everything should be stored with a dot format.

PR #6531, opened 2024-06-11
Fixes undefined array key log when sessions config uses Check User agent

scottcwilson:
```
--> PHP Warning: Undefined array key "HTTP_USER_AGENT" in /home/client/catalog/includes/init_includes/init_sessions.php on line 150.
```

### DIFF
--- a/includes/init_includes/init_sessions.php
+++ b/includes/init_includes/init_sessions.php
@@ -147,7 +147,7 @@ if ($request_type === 'SSL' && SESSION_CHECK_SSL_SESSION_ID === 'True' && ENABLE
  * verify the browser user agent if the feature is enabled
  */
 if (SESSION_CHECK_USER_AGENT === 'True') {
-    $http_user_agent = $_SERVER['HTTP_USER_AGENT'];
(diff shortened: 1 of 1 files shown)


includes/init_includes/init_sessions.php
```diff
@@ -147,7 +147,7 @@
  * verify the browser user agent if the feature is enabled
  */
 if (SESSION_CHECK_USER_AGENT === 'True') {
-    $http_user_agent = $_SERVER['HTTP_USER_AGENT'];
+    $http_user_agent = $_SERVER['HTTP_USER_AGENT'] ?? '';
     if (empty($_SESSION['SESSION_USER_AGENT'])) {
```
drbyte, line 151: Should the next line also check that `$http_user_agent` is not empty, since that is likely now?

scottcwilson: I don't think so - setting `$_SESSION['SESSION_USER_AGENT']` to blank means they will match, so no session destroy.  I think it's ok as is.

PR #6540, opened 2024-06-15
Load `zc_plugin` directories in a predictable order

lat9:
Currently, `zc_plugin` files are loaded in the order in which they were installed, leading to an unpredictable load order.

This PR loads them, instead, by sorting the installed plugins' `unique_key`.

### DIFF
--- a/includes/classes/PluginManager.php
+++ b/includes/classes/PluginManager.php
@@ -47,14 +47,18 @@ class PluginManager
 
     public function isUpgradeAvailable($uniqueKey, $currentVersion)
     {
-        if (empty($currentVersion)) return false;
+        if (empty($currentVersion)) {
+            return false;
+        }
         $versionList = $this->getVersionsForUpgrade($uniqueKey, $currentVersion);
         return count($versionList);
     }
 
     public function getVersionsForUpgrade($uniqueKey, $currentVersion)
     {
-        if (empty($currentVersion)) return [];
+        if (empty($currentVersion)) {
+            return [];
+        }
         $versions = $this->getPluginVersions($uniqueKey);
         $versionList = [];
         foreach ($versions as $version) {
@@ -86,7 +90,9 @@ class PluginManager
 
         // if no results or invalid format, abort
         // @TODO - is this the right return type? or should we return the unaltered $plugins array?
-        if (empty($results)) return false;
+        if (empty($results)) {
+            return false;
+        }
 
         // make sure $results is the actual array we want to iterate over, and not a sub-array
         if (is_array($results) && !isset($results[0]['id']) && isset($results[0][0]['id'])) {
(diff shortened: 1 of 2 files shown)


includes/classes/PluginManager.php
```diff
@@ -28,7 +28,7 @@ public function inspectAndUpdate()
 
     public function getInstalledPlugins()
     {
-        $results = $this->pluginControl->where(['status' => 1])->get();
+        $results = $this->pluginControl->where(['status' => 1])->orderBy('unique_key')->get();
```
drbyte, line 31: Do we want it to be sorted "first" by `unique_key`, which might be very different than alpha-by-name?

Or would `->orderBy('name')->orderBy('unique_key')` (first alpha by name, then by key for duplicate names) be better?

lat9: Done; see commit below.

drbyte:
I gather that the intent of this load-order change is specific to "booting up" the (installed) plugins on page-load.

The Admin UI for Plugin Manager would require a different change for actually sorting the list of displayed "available, whether 'installed' or not" plugins:
@zcwilt ... would changing `PluginManagerDataSource.php` be safe (no undesired side-effects in other lookups?), done this way?:
```diff
    protected function buildInitialQuery() : Builder
    {
-        return (new PluginControl())->query();
+        return (new PluginControl())->query()->orderBy('name')->orderBy('unique_key');
    }
```

zcwilt:
@drbyte I have no issue with that change, although see no benefit in it, had always wanted to get back to the ViewBuiolderrs stuff to tidy up and add sort filters etc.

zcwilt:
as for the overall change, I would ask, what problem is this solving. What use cases mean we need to load plugins in a this specific order.

Just to be clear. I agree that there maybe specific reason you want a plugin to load before another plugin, 
I can imagine a plugin that provides a set of utility classes that another plugin might want to use and not have to re-implement them.

I did consider writing code to do some kind of dependency check based on a manifest entry, but at that point you are pretty much writing a custom Composer clone  when why not just use Composer. 
(Well apart from then having to have Service Provider classes, DI containers blah blah)

lat9:
Mostly thinking about language-file load order.  Granted, plugins shouldn't be supplying the same filenames, but just in case there's a rogue this change ensures that the result is the same from site to site, regardless the order in which the plugins were installed.

lat9:
> I gather that the intent of this load-order change is specific to "booting up" the (installed) plugins on page-load.
> 
> The Admin UI for Plugin Manager would require a different change for actually sorting the list of displayed "available, whether 'installed' or not" plugins: @zcwilt ... would changing `PluginManagerDataSource.php` be safe (no undesired side-effects in other lookups?), done this way?:
> 
> ```diff
>     protected function buildInitialQuery() : Builder
>     {
> -        return (new PluginControl())->query();
> +        return (new PluginControl())->query()->orderBy('name')->orderBy('unique_key');
>     }
> ```

That was also what I was looking for (too many buried classes).  Without that change, the Plugin Manager displays the plugins in the order in which they were installed:
image

With that change, they're now displayed alphabetically:
image

drbyte:
> That was also what I was looking for (too many buried classes). Without that change, the Plugin Manager displays the "plugins in the order in which they were installed":

Clarification, because "installed" is a multifaceted word; so I think what you meant was: "... displays the plugins in the order they were added to the filesystem"

scottcwilson:
Is this ready to be merged?

lat9:
> Is this ready to be merged?

In my opinion, Yes.

PR #6541, opened 2024-06-16
Use `$psr4Autoload` instead of `config.core.php` to handle the base class' loading

lat9:
... thus enabling ~~that class~~ those classes to be overridden if needed.


drbyte:
Strictly speaking, no plugins should be directly modifying these files, because they should be providing their own autoload file for their own additional requirements.
In the case where they were intentionally altering the contents of these files, obviously this will be a breaking change for them, and they will need to revise their plugin accordingly.

drbyte (approved):

PR #6546, opened 2024-06-19
Realign Icons for Downloads Manager

dbltoe:
As it is now, closing the col-sm-6 and re-opening it between the misconfigured and missing listings forces them to be separated and confuse the store owner. https://www.zen-cart.com/showthread.php?230203-Downloadable-Product-is-misconfigured.

before
image

after
image


scottcwilson:
Good catch!

PR #6551, opened 2024-06-22
zc_install - fix parsing of optional COLUMN keyword

drbyte:


lat9 (approved):

zc_install/includes/classes/class.zcDatabaseInstaller.php
```diff
@@ -493,8 +493,15 @@ public function parserAlterTable(): void
                 case 'MODIFY':
                     // Check to see if the column / index already exists
                     // we treat it falsey in this case because we cannot perform the change if the column is not present
-                    $exists = !$this->tableColumnExists($this->lineSplit[2], $this->lineSplit[4]);
-                    $result = sprintf(REASON_COLUMN_DOESNT_EXIST_TO_CHANGE, $this->lineSplit[4]);
+
+                    // we check for semi-optional COLUMN keyword
+                    if (strtoupper($this->lineSplit[4]) === 'COLUMN') {
+                        $exists = !$this->tableColumnExists($this->lineSplit[2], $this->lineSplit[5]);
+                        $result = sprintf(REASON_COLUMN_DOESNT_EXIST_TO_CHANGE, $this->lineSplit[5]);
```
CeonSupport, line 500: Why, if the column actually does exist, is a log entry made indicating that it does not? Shouldn't there be some sort of check against `$exists` such that if now `truthy` that should make this log entry?

All sorts of ways to approach, e.g. if want to always log an action has occurred or been taken, there is a path. If wanting to only make a log when a problem occurs, this is logging the incorrect information as the "nonexistent column" condition is not reported based on its existence/absence of the column even though the requested action will occur against the column when it does exist.

PR #6553, opened 2024-06-23
Language in admin stuck to deleted language

piloujp:
In Admin pages, when using 2 languages and deleting the one (not default) that you are actually using through languages page, you get stuck in that language.
The language drop down menu is not there anymore and although in languages page there is only the default language English, you can't change back to it...
With this modification, language will change back to default using GET parameter when deleting the language that is actually in use.

### DIFF
--- a/admin/languages.php
+++ b/admin/languages.php
@@ -247,10 +247,9 @@ if (!empty($action)) {
       $db->Execute("DELETE FROM " . TABLE_EZPAGES_CONTENT . " WHERE languages_id = " . (int)$lID);
 
       // if we just deleted our currently-selected language, need to switch to default lang:
-      if ((int)$_SESSION['languages_id'] == (int)$_POST['lID']) {
+      $getlang = '';
+      if ((int)$_SESSION['languages_id'] === (int)$_POST['lID']) {
           $getlang = '&language=' . DEFAULT_LANGUAGE;
-      } else {
-          $getlang = '';
       }
 
       $zco_notifier->notify('NOTIFY_ADMIN_LANGUAGE_DELETE', (int)$lID);
--- a/admin/languages.php
+++ b/admin/languages.php
@@ -250,12 +250,16 @@ if (!empty($action)) {
       $lng = $db->Execute("SELECT languages_id
                            FROM " . TABLE_LANGUAGES . "
                            WHERE code = '" . zen_db_input(DEFAULT_LANGUAGE) . "'");
-      if ((int)$_SESSION['languages_id'] == (int)$_POST['lID'])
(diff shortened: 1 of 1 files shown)


drbyte:
Hmm. Haven't tested it yet.
In reading the code, setting it via the session variable should have been enough, since that's what's checked if no GET param is provided.

piloujp:
I checked that and changing only $_SESSION['languages_id'] do not change language at all, but if I add $_SESSION['language'] then language changes. 
actually looking in language init these 3 session variables are set with a new language object.
 ```
 $_SESSION['language']
  $_SESSION['languages_id']
  $_SESSION['languages_code']
```

Replacing $_SESSION['languages_id'] by $_SESSION['language'] did the job!

There are 2 solutions for this PR, updating SESSION variable or just use GET parameter:
```
      // if we just deleted our currently-selected language, need to switch to default lang:
      $lng = $db->Execute("SELECT languages_id, name, code
                           FROM " . TABLE_LANGUAGES . "
                           WHERE code = '" . zen_db_input(DEFAULT_LANGUAGE) . "'");
      if ((int)$_SESSION['languages_id'] == (int)$_POST['lID']) {
          $_SESSION['language'] = $lng->fields['name'];
          $_SESSION['languages_id'] = $lng->fields['languages_id']; // Without this one we got some log files
          $_SESSION['languages_code'] = $lng->fields['code']; // Not sure this one is necessary, it seems to work without it
      }

      $zco_notifier->notify('NOTIFY_ADMIN_LANGUAGE_DELETE', (int)$lID);

      zen_redirect(zen_href_link(FILENAME_LANGUAGES, 'page=' . $_GET['page']));
```
Or with GET parameter:
```
      // if we just deleted our currently-selected language, need to switch to default lang:
      if ((int)$_SESSION['languages_id'] == (int)$_POST['lID']) {
          $getlang = '&language=' . DEFAULT_LANGUAGE;
      } else {
          $getlang = '';
      }

      $zco_notifier->notify('NOTIFY_ADMIN_LANGUAGE_DELETE', (int)$lID);

      zen_redirect(zen_href_link(FILENAME_LANGUAGES, 'page=' . $_GET['page'] . $getlang));
```
I did few test with these two different codes.
With GET parameter everything seems perfect but with Session variables, when I set back French language (the second language I used for testing) the language drop down menu is set to French but admin is in English. It seems language is not completly reinitialized. I could not find any other problem but it does not fell good...
The GET solution seems safer to me.

piloujp:
I updated the PR to use GET parameter only.
Using Session variables does not initiate language completely, as you can see in init_languages.php code. Language directory is not reset and notifier for language change is not fired, making it useless for any plugin.

drbyte:
Thank you for investigating this!

PR #6556, opened 2024-06-26
Protects against non-existent order statuses

scottcwilson:
If a plugin writes to the order_status_history table and uses a status of 0 (or any non-existent status), this will produce a log. 

```
--> PHP Warning: Undefined array key 0 in /home/client/public_html/store/admin/orders.php on line 1057.
```

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -1064,7 +1064,7 @@ if ($show_orders_weights === true) {
                         }
                         ?>
                     </td>
-                    <td><?php echo $orders_status_array[$item['orders_status_id']]; ?></td>
+                    <td><?php echo $orders_status_array[$item['orders_status_id']] ?? ''; ?></td>
 <?php
                     // -----
                     // A watching observer can provide an associative array in the form:


lat9:
Would a similar issue be present on the storefront where an order's history is displayed, e.g. `account_history_info`?

scottcwilson:
Good question - let me double check.

scottcwilson:
OK in 2.0, the order class does a query on TABLE_ORDERS_STATUS LEFT JOIN TABLE_ORDERS_STATUS_HISTORY to build the list of order statuses, so this entry is not included in the list of OSH entries.

scottcwilson:
Note to self: MBG 45361 is a test order.

lat9:
> OK in 2.0, the order class does a query on TABLE_ORDERS_STATUS LEFT JOIN TABLE_ORDERS_STATUS_HISTORY to build the list of order statuses, so this entry is not included in the list of OSH entries.

Interesting.

PR #6557, opened 2024-07-01
Admin submenus translation scripts

piloujp:
Admin submenus translation scripts using an observer. When language changes, a file in 'admin\includes\languages\english\admin_submenus\' (replace 'english' by desired language) is loaded and import in database where necessary.
These language files use a new format as there is no constant here, and they don't need to be loaded in advance.

It is based on an associative array $admin_submenus containing four associative arrays, one for each table in ZC that have fields to translate. These arrays use for key the configuration key or whatever unique index of tables in the database. The value is an associative array with pairs of fields/column names and their translated content.
Please, take a look at 'admin_submenus.php' which is the English version necessary to translate back to English when desired.

If the translation file is missing or part of it, missing menus will just stay in the language they were, no error occurs and no error 500 like with constants. It is safe to use and there is no compatibility issue as there is no modification of actual ZC code.

Adding translation for plugins is easy, just create a file starting by 'admin_submenus' and containing an array like described above.
Here is an example for Image Handler:
'admin_submenus_IH_v5.3.5_fr.php'  in 'admin\includes\languages\french\admin_submenus\'
containing this:
```
<?php

// Traductions additionnelles pour le plugin Image Handler 5.3.5
$admin_submenus['configuration'] += ['IH_RESIZE' => ['title' => 'IH redimensionne les images', 'description' => 'Sélectionnez soit -no- qui est l\'ancien comportement de Zen-Cart, soit -yes- pour activer le redimensionnement automatique et la mise en cache des images. --Remarque : Si vous sélectionnez -no-, tous les paramètres d\'image spécifiques au gestionnaire d\'images ne seront pas disponibles, notamment : sélection du type de fichier image, couleurs d\'arrière-plan, compression, survol de l\'image et filigrane - Si vous souhaitez utiliser ImageMagick, vous devez spécifier l\'emplacement de l\'exécutable <strong>de conversion</strong> dans <em>includes/extra_configures/bmz_image_handler_conf.php</em>.'],
                                        'IH_VERSION' => ['title' => 'Version d\'IH', 'description' => 'Affiche la version actuellement installée d\'<em>Image Handler</em>.'],
                                        'IH_CACHE_NAMING' => ['title' => 'Convention de dénomination des fichiers de cache d\'IH', 'description' => '<br>Choisissez la méthode utilisée par <em>Image Handler</em> pour nommer les images redimensionnées dans le répertoire <code>bmz_cache</code>.<br><br><em>Hashed</em> : Utilise un hache &quot;MD5&quot; pour produire les noms de fichiers. Cela peut être « difficile » pour identifier visuellement le fichier original en utilisant cette méthode.<br><br><em>Readable</em> : C\'est un bon choix pour les nouvelles installations de <em>IH</em> ou pour les installations mises à niveau qui n\'ont pas de liens d\'images codés en dur.<br><br><em>Mirrored</em> : Similaire à <em>Lisible</em>, mais la structure des répertoires dans <code>bmz_cache</code> reflète la structure des sous-répertoires des images d\'origine.'],
                                    ];
```
English version is needed too in 'admin\includes\languages\english\admin_submenus\'.

To test this new functionality, you can use the French language pack from my GitHub repo [here](https://github.com/piloujp/French_Language_Pack_for_ZC). It is quick and easy to install.
Configuration table translation is still in progress (40%), but it is enough to test.


drbyte:
This functionality already exists in core code, and is easily activated in existing language files simply by defining the constants with the necessary prefix:

https://github.com/zencart/zencart/blob/2fa3d15b255e26fb45b6a7b3984d527b3d40327c/admin/configuration.php#L60-L68

https://github.com/zencart/zencart/blob/2fa3d15b255e26fb45b6a7b3984d527b3d40327c/admin/configuration.php#L163-L172

piloujp:
I found this code a while ago. It is great but limited to two tables, configuration and configuration_group. I can't remember exactly, but I think it was not usable for plugins and there were few small problems with developer tool. I try to extend its use without success.
If you really want to translate everything, there is although product_types, product_type_layout and plugin_control tables that need some translation. This PR takes care of this.

drbyte:
Also, separate issue about the code: this observer, while using prepared SQL statements, doesn't use Zen Cart's database queryFactory, and this would be the only place in the code where that is done.

piloujp:
> Right, but what I don't like about this PR is that you're directly changing the name/description of all those settings in the database every time the language is swapped.

I agree and tried other methods to do this translation without touching database, but there were always limitations or worse drawbacks. And when you think of it, it does not happen very often. It is in admin where language is not change very often...

piloujp:
> Also, separate issue about the code: this observer, while using prepared SQL statements, doesn't use Zen Cart's database queryFactory, and this would be the only place in the code where that is done.

I did not find any function for prepared statements in queryFactory. There is a function that mimics prepared statement for parameters by accepting parameters that are replaced using subtrstring function, then query is parsed as any other query.
To have the performance advantage of prepared statement (query parsed only one time then only parameters are sent at each iteration), I had to use directly PHP functions for that.
If I missed a prepared statement function in queryFactory, I would be happy to rewrite the code using it.

admin/includes/classes/observers/class.admin_submenus.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $this->attach($this, ['NOTIFY_LANGUAGE_CHANGE_REQUESTED_BY_ADMIN_VISITOR']);
+    }
+
+    public function update(&$class, $eventID, $lang, &$olan)
+    {
+        $admin_submenus = [];
+        $table_type = [];
+
+        $conn = new mysqli(constant('DB_SERVER'), DB_SERVER_USERNAME, DB_SERVER_PASSWORD, constant('DB_DATABASE'));
```
drbyte, line 19: I think you could replace this with:
```php
global $db;
$conn = $db->link;
```

admin/includes/classes/observers/class.admin_submenus.php
```diff
@@ -0,0 +1,84 @@
...
+        if (is_dir($lang_dir_name)) {
+            $dir_content = array_diff(scandir($lang_dir_name), array('..', '.'));
+            foreach($dir_content as $key => $filename) {
+                if (is_file($lang_dir_name . '/' . $filename) && strpos($filename, 'admin_menus') === 0) { // checking for files starting by 'admin_menus' to include them as they should contain '$admin_submenus' table data.
+                    include($lang_dir_name . '/' . $filename);
+                }
+            }
+        }
+        
+        $plugin_infos = $conn->query('SELECT unique_key, version FROM plugin_control WHERE 1'); // check fo installed encapsulated plugins
```
drbyte, line 31: This could be a regular `$result = $db->Execute()` call, which then iterates `foreach ($result as $plugin)`

admin/includes/classes/observers/class.admin_submenus.php
```diff
@@ -0,0 +1,84 @@
...
+                        foreach($menus as $type_id => $type_name) {
+                            $query_plugin->bind_param("ss", $type_name, $type_id);
+                            $query_plugin->Execute();
+                        }
+                        $query_plugin->close();
+                        break;
+                }
+            }
+        }
+        $conn->close();
```
drbyte, line 82: This wouldn't be necessary if `$conn` comes from the `$db` class object, because the class registers a call to `close` automatically.

drbyte:
> And when you think of it, it does not happen very often. It is in admin where language is not change very often.

Perhaps. But if someone's editing define pages, etc, they will switch languages multiple times to load the file for editing. In that case it resets things a lot.

drbyte:
One thing I haven't had time enough to explore is how often any of these values are used on the storefront, and not merely in the Admin.

admin/includes/classes/observers/class.admin_submenus.php
piloujp: Done!

admin/includes/classes/observers/class.admin_submenus.php
piloujp: Done!

admin/includes/classes/observers/class.admin_submenus.php
piloujp: Done!

piloujp:
> Perhaps. But if someone's editing define pages, etc, they will switch languages multiple times to load the file for editing. In that case it resets things a lot.
I think database can easily handle it, but problems might occur if you have multiple admins doing this at the same time and different languages... Very low probability but possible.
Even 2 admins connected at the same time using different languages is a problem, the last one to change language will impose it (database part only) to the other. This is really the weak point of this PR.

torvista:
I've been repeatedly vocal in promoting the idea that all efforts from anyone willing to contribute should be encouraged.

But, given my post-holiday exhaustion and subsequent grumpiness...and apologies to the well-meaning contributor.

What a lot of effort and discussion to solve a non-problem and maybe create more. We change languages all the time, I don't want a load of db changes every time. 
Whose admin user is complaining about this?
On whose wish list is this?
Where are the users complaining they are not using ZC because of this?

If there was a roadmap, I doubt this would be at the top.

Some focus on the real problem / a roadmap please....

piloujp:
> I've been repeatedly vocal in promoting the idea that all efforts from anyone willing to contribute should be encouraged.
> 
> But, given my post-holiday exhaustion and subsequent grumpiness...and apologies to the well-meaning contributor.
> 
> What a lot of effort and discussion to solve a non-problem and maybe create more. We change languages all the time, I don't want a load of db changes every time. Whose admin user is complaining about this? On whose wish list is this? Where are the users complaining they are not using ZC because of this?
> 
> If there was a roadmap, I doubt this would be at the top.
> 
> Some focus on the real problem / a roadmap please....

Wow... First line that says something, then ten that say the opposite!
I prefer not to comment on this...

PR #6560, opened 2024-07-02
Adding Products' Options' Stock Manager

lat9:
See https://github.com/lat9/options_stock_support/wiki for additional information.


scottcwilson:
WOW!  Great stuff. 

When merged I will update https://docs.zen-cart.com/release/whatsnext/ to move this feature to 2.1.

torvista:
WOW! Indeed...that may be the biggest **feature** catapult forward that Zen Cart has ever had.
The elephant in the room has filled the hole, to mix metaphors.

A new home page is needed **more than ever** now there is so much to shout about.

drbyte:
@lat9 would you be open to renaming this slightly, to reduce the plurals so that extra escaping of apostrophes can be avoided?

ie: `private string $configGroupTitle = 'Products\\\' Options\\\' Stock Manager';`

If we rename this to `Product Options Stock Manager` (ie: products have options, and this is the "stock manager for product-options") then none of the apostrophes are needed, nor the associated char-escaping.

lat9:
That's going to be a pain for sites that purchased the plugin when it was in the commercial 'space', since that's the config-group-title that was/is in use.  With the change in name, non-encapsulated versions of the plugin (when updated) won't find the pre-existing configuration settings.

Note, too, that the apostrophes denote possessive rather than plural.

scottcwilson:
Couldn't we fix that with a SQL UPDATE in the 2.1 upgrade script?  Similar to the ALTER TABLE statements we did in case dimensions were already in the products table?

lat9:
> Couldn't we fix that with a SQL UPDATE in the 2.1 upgrade script? Similar to the ALTER TABLE statements we did in case dimensions were already in the products table?

I'm not clear what we're fixing; it seems to me an added complication.  The configuration-group-title has been that way for 10+ years.

scottcwilson:
The concern that "non-encapsulated versions of the plugin ... won't find the pre-existing configuration settings."

lat9:
I meant why are we having a discussion surrounding two apostrophes?

PR #6565, opened 2024-07-04
Corrects attribute pricing display

### LINKED ISSUE #6564: Attributes' +/- pricing display includes product's base price
When a product is priced by attributes, its base price is adjusted by the attribute(s)' pricing when displayed.  In the current Zen Cart repo, the attributes' pricing displayed now includes the product's base price.

To see using the demo products:
1. Create a copy of the "Gift Certificates by attributes" product (id=47), also copying the associated attributes, enable the copied product and set its quantity to some value > 0.
2. Edit the copied product, changing its base price to 5.00.
3. Edit the copied product's attributes, changing the price of the `$5.00` 'Amount' to have a price_prefix of '+' and a price of 0.00.
4. Navigate to the storefront to view the now-active copied product.

The product's attribute selection section's pricing incorrectly includes the $5.00 product base price:
image

lat9:
Fixes #6564

There was, for whatever reason, an already present parameter on `zen_get_attributes_price_final` to indicate whether/not to include the product's base price, it was just never used.

Additional parameters needed for the admin's Attributes' Controller display, since that _does_ include the product's base pricing.

Noting that the issue's been around since at least zc158a.

### DIFF
--- a/admin/attributes_controller.php
+++ b/admin/attributes_controller.php
@@ -1520,7 +1520,7 @@ function zen_js_option_values_list($selectedName, $fieldName)
 // calculate current total attribute price
 // $attributes_values
 
-                  $attributes_price_final = zen_get_attributes_price_final($attributes_value['products_attributes_id'], 1, $attributes_values, false);
+                  $attributes_price_final = zen_get_attributes_price_final($attributes_value['products_attributes_id'], 1, $attributes_values, false, false, 0, true);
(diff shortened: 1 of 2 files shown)


scottcwilson:
Would you please take a screenshot to match the one in #6564 ?  Thanks.

lat9:
It's using Bootstrap instead of responsive-classic, but you'll get the drift:
image

scottcwilson:
Great catch!

PR #6573, opened 2024-07-11
Add notifier to ot_group_pricing

### LINKED ISSUE #6572: Add notifier to `ot_group_pricing.php`
### Discussed in https://github.com/zencart/zencart/discussions/6569

<sup>Originally posted by **daphilli224** July 11, 2024</sup>
Please add a new notifier in ot_group_pricing.php, calculate_deductions function, in the next zencart release to allow an observer to override the discount amount for guest customers. See One-Page Checkout Support forum thread post #2876. Thank you!</div>

lat9:
Fixes #6572

### DIFF
--- a/includes/modules/order_total/ot_group_pricing.php
+++ b/includes/modules/order_total/ot_group_pricing.php
@@ -138,10 +138,9 @@ class ot_group_pricing {
     return $order_total;
   }
   function calculate_deductions($order_total) {
-    global $db, $order;
+    global $db, $order, $zco_notifier;
     $od_amount = array();
     if ($order_total == 0 || !zen_is_logged_in() || zen_in_guest_checkout()) {
-        global $zco_notifier;
         $zco_notifier->notify('NOTIFY_OT_GROUP_PRICING_DEDUCTION_OVERRIDE', ['order_total' => $order_total], $od_amount);
         return $od_amount;
     }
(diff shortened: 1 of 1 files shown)


proseLA:
my opinion, has not changed.

when adding a notifier, always allow it to be more inclusive as opposed to less inclusive.

in this situation, the notifier ONLY gets called if those conditionals are met.  

i firmly believe that the notifier should be OUTSIDE of those conditionals.  that allows the notifier to be used for situations that perhaps we have not thought of.

the observer can then determine if it wants to change the `$od_amount` based on those conditionals.

now perhaps those conditionals may need to get called twice or perhaps a new var in the notifier signature is necessary to indicate whether to do an early return of the `$od_amount`.  

but if we are putting an observer in, make it more useful by not only calling it within a conditional.

lat9:
I'll consider adding another notification within the 'normal' processing path.

lat9:
Noting that the additional notification enables an observer to modify the deduction for a logged-in customer in an _existing_ pricing group.  That would enable, for instance, a graduated deduction based on the customer's order value or products in the cart.

There's not a notification when a non-guest customer is logged in and they're not in an existing pricing group because that didn't make sense to me.  If the customer should get a group-pricing discount, they should be set as such via the admin.

scottcwilson:
Suggestion: make the keys 'order_total' and 'orderTotal' the same.

lat9:
> Suggestion: make the keys 'order_total' and 'orderTotal' the same.
They're different value types (and reflect the associated variable being passed).  In the first notification, it's the _scalar_ value of the order; in the second it's a copy of the `$orderTotal` _array_.

PR #6575, opened 2024-07-12
Issue viewing older orders from front page

### LINKED ISSUE #6574: Issue viewing orders not on first page
Same as #6468 but on orders page.  Repro requires a cart with > 20 orders (assuming default settings). 

Repro: go to home page, in the list of new orders on the right hand side, click an order which is more than 20 orders ago.  I will open the page 

admin/index.php?cmd=orders&oID=old-order-number&origin=index

with the right infoBox blank.  Same fix as for customers page required on orders page.

scottcwilson:
Fixes #6574.

Normally in order navigation you have a `page` parameter, which obscures this issue, but if you click from the front page new orders widget, you don't, so the issue appears. 

Credit to @piloujp for the original fix in #6481.


lat9:
Fix verified!

PR #6581, opened 2024-07-13
Correct invalid HTML yielding wonky display on POSM view-all

### LINKED ISSUE #6580: Bottom of POSM View All screen is mixed up
Zen Cart logo should be at the very bottom. 

Screen Shot 2024-07-13 at 2 25 39 PM

lat9:
Fixes #6580.

### DIFF
--- a/zc_plugins/POSM/v6.0.0/admin/products_options_stock_view_all.php
+++ b/zc_plugins/POSM/v6.0.0/admin/products_options_stock_view_all.php
@@ -345,6 +345,9 @@ if ($products_list->EOF) {
                 <tr>
                     <td colspan="<?= STATIC_FIELD_COUNT ?>" class="text-center"><?= POSM_VIEW_ALL_NO_PRODUCTS_TO_LIST ?></td>
                 </tr>
+            </tbody>
+        </table>
+    <?= '</form>' ?>
 <?php
 } else {
     $posm_view_all = new PosmViewAll($posm_stock_reorder_level, $sort_by);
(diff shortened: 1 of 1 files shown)


zc_plugins/POSM/v6.0.0/admin/products_options_stock_view_all.php
```diff
@@ -345,6 +345,9 @@
                 <tr>
                     <td colspan="<?= STATIC_FIELD_COUNT ?>" class="text-center"><?= POSM_VIEW_ALL_NO_PRODUCTS_TO_LIST ?></td>
                 </tr>
+            </tbody>
+        </table>
+    <?= '</form>' ?>
```
proseLA, line 350: this is not what i envisioned the use of the short echo tag.

the short echo tag is to echo php vars within a html block.

good coding standards would remove the short echo and just have the `</form>` within the html.

zc_plugins/POSM/v6.0.0/admin/products_options_stock_view_all.php
```diff
@@ -532,8 +535,8 @@
 <?php
 }  // One or more products was found
 ?>
-<!-- body_eof //-->
 </div>
+<!-- body_eof //-->
 
 <!-- footer //-->
 <?php require DIR_WS_INCLUDES . 'footer.php'; ?>
```
proseLA, line 542: i would remove the end php `?>` tag.

zc_plugins/POSM/v6.0.0/admin/products_options_stock_view_all.php
lat9: But, since the form is rendered via `zen_draw_form`, the end-form tag should also (IMO) be rendered via PHP.

zc_plugins/POSM/v6.0.0/admin/products_options_stock_view_all.php
lat9: I'm `assuming` that you mean the one on line 542?

zc_plugins/POSM/v6.0.0/admin/products_options_stock_view_all.php
proseLA: i will never make a comment on formatting again.

while i see your reasoning, i do not agree with it.  it strikes me as something someone else would say in order that their IDE does not complain.

PR #6584, opened 2024-07-16
Admin submenus translation

piloujp:
# Admin Multi-Language Support
**Prior to v2.1.0 very little of the Admin Interface would "display" in a language other than English.**

**Admin Configuration Menu Translations**
Since v1.5.5 it was possible to define translations in language packs for the Admin "Configuration" menu
using a few override conventions:
```
//For example, in admin/includes/languages/spanish/lang.configuration.php

define('CFGTITLE_STORE_NAME', 'Nombre de la Tienda');
define('CFGDESC_STORE_NAME', 'El nombre de mi tienda');
define('CFG_GRP_TITLE_MY_STORE', 'Mi Tienda');
```

### Now with v2.1.0 this is expanded to more areas:

**For product_types:**

The `product_type` `type_name` can be defined using 
`'PRODUCT_TYPE_NAME_FOR_HANDLER_'` followed by the product_type's uppercase `type_name`

**Example for French:**
```
    'PRODUCT_TYPE_NAME_FOR_HANDLER_PRODUCT' => 'Produit - Général',
    'PRODUCT_TYPE_NAME_FOR_HANDLER_PRODUCT_MUSIC' => 'Produit - Musique',
    'PRODUCT_TYPE_NAME_FOR_HANDLER_DOCUMENT_GENERAL' => 'Document - Général',
    'PRODUCT_TYPE_NAME_FOR_HANDLER_DOCUMENT_PRODUCT' => 'Document - Produit',
    'PRODUCT_TYPE_NAME_FOR_HANDLER_PRODUCT_FREE_SHIPPING' => 'Produit - Envoi Gratuit',
```

**And for Product Type layout/configuration settings** (found in product_type_layout table's configuration_title and configuration_description fields):
`'PRODUCT_TYPE_LAYOUT_TITLE_FOR_'` followed by uppercase `configuration_key`
`'PRODUCT_TYPE_LAYOUT_DESC_FOR_'` followed by uppercase `configuration_key`

French examples:
```
    'PRODUCT_TYPE_LAYOUT_TITLE_FOR_SHOW_PRODUCT_INFO_MODEL' => 'Afficher le modèle',
    'PRODUCT_TYPE_LAYOUT_DESC_FOR_SHOW_PRODUCT_INFO_MODEL' => 'Afficher le numéro de modèle dans les informations de produit 0= off 1= on',
    'PRODUCT_TYPE_LAYOUT_TITLE_FOR_SHOW_PRODUCT_INFO_WEIGHT' => 'Afficher le poids',
    'PRODUCT_TYPE_LAYOUT_DESC_FOR_SHOW_PRODUCT_INFO_WEIGHT' => 'Afficher le poids dans les informations de produit 0= off 1= on',
    'PRODUCT_TYPE_LAYOUT_TITLE_FOR_SHOW_PRODUCT_INFO_MANUFACTURER' => 'Afficher le fabricant',
    'PRODUCT_TYPE_LAYOUT_DESC_FOR_SHOW_PRODUCT_INFO_MANUFACTURER' => 'Afficher le nom du fabricant dans les informations de produit 0= off 1= on',
(and there are a few dozen more possible entries that can be translated this way)
```

**For Plugins listed in the Plugin Manager**
In your plugin's language file, add a define:
`'ADMIN_PLUGIN_MANAGER_DESCRIPTION_FOR_'` followed by the uppercase english `Description`

French examples:
```
    'ADMIN_PLUGIN_MANAGER_DESCRIPTION_FOR_DISPLAYLOGS' => 'Affiche et gère les fichiers journaux de Zen Cart.',
    'ADMIN_PLUGIN_MANAGER_DESCRIPTION_FOR_POSM' => 'Ce plugin permet à votre site d\'attribuer des niveaux de stock et des numéros de modèle à vos produits en fonction de combinaisons produit-attributs (également appelées « variantes de produit » ou « stock par attributs »).',
```

---

~First, this PR is meant to replace PR #6557 .~
~Before this PR few paths have been explored to translate admin submenus:~
- ~Already embedded code to use constant to replace configuration table title and description: Works great but limited to two tables.~
- ~Adding table with translation data and a language ID to link them to original table: Too many modifications of core code to apply for configuration table making it impossible to maintain backward plugins' compatibility.~
- ~Adding new translation columns in existing tables: Less modification of ZC code but same problem as above.~
- ~PR 6557 that uses an observer to detect language changes and update tables that need translation with new language data: At first it looks good, there was no modification of ZC code but modifying the database each time language is changed was a bigger problem than I thought.~

~For this PR, I used a mix of these possibilities. It uses original code that read constants in lang.configuration.php file for configuration and configuration group tables and for other tables a column is added using same name as original plus a suffix using language code. These columns are created/deleted each time a language is added or deleted, and lang.configuration.php file is populated at the same time.~
~Like this, when changing language, there is no writing anywhere, only reading translated data.
Translation data are provided by two different kind of files, one with constants for configuration table and another with an array for other tables. These files can be multiple, which allows using files in plugins too, both standard and embedded.~

I added these special languages files (be careful, they are different from previous PR) to French language pack on [Github here](https://github.com/piloujp/French_Language_Pack_for_ZC), if you need to test...

For this PR, few ZC core files need modifications but not only it provides language change to core admin submenus, but although to plugin added ones. French language files have data for POSM, DisplayLogs and IH5.


admin/configuration.php
```diff
@@ -58,13 +58,14 @@
...
-    // For example, in admin/includes/languages/spanish/configuration.php
+    // For example, in admin/includes/languages/spanish/lang.configuration.php
     // define('CFG_GRP_TITLE_MY_STORE', 'Mi Tienda');
     $str = $cfg_group->fields['configuration_group_title'];
-    $str = preg_replace('/[^a-zA-Z0-9_\x80-\xff]/', '_', $str);
-    $const = 'CFG_GRP_TITLE_' . strtoupper($str);
-    if (defined($const)) {
-        $cfg_group->fields['configuration_group_title'] = constant($const);
+    $str = str_replace(' ', '_', $str);
+    $str = strtoupper(preg_replace('/[^a-zA-Z0-9_\x80-\xff]/', '', $str));
```
drbyte, line 65: Why did you replace the `_` substitution with a blank?

admin/languages.php
```diff
@@ -139,15 +139,16 @@
...
 
 // create additional orders_status records
-        $orders_status = $db->Execute("SELECT orders_status_id, orders_status_name
+        $orders_status = $db->Execute("SELECT orders_status_id, orders_status_name, sort_order
                                        FROM " . TABLE_ORDERS_STATUS . "
                                        WHERE language_id = " . (int)$_SESSION['languages_id']);
 
         foreach ($orders_status as $status) {
-          $db->Execute("INSERT INTO " . TABLE_ORDERS_STATUS . " (orders_status_id, language_id, orders_status_name)
+          $db->Execute("INSERT INTO " . TABLE_ORDERS_STATUS . " (orders_status_id, language_id, orders_status_name, sort_order)
```
drbyte, line 147: Do you mind moving these 3 changes for including `sort_order` to a separate PR?  I'd merge that fix right away.

admin/languages.php
```diff
@@ -176,6 +177,145 @@
...
+        $products_option_stock_names = $db->Execute("SELECT pos_name_id, pos_name
+                                       FROM " . TABLE_PRODUCTS_OPTIONS_STOCK_NAMES . "
+                                       WHERE language_id = " . (int)$_SESSION['languages_id']);
+
+        foreach ($products_option_stock_names as $option_stock_name) {
+          $db->Execute("INSERT INTO " . TABLE_PRODUCTS_OPTIONS_STOCK_NAMES . " (pos_name_id, language_id, pos_name)
+                        VALUES ('" . (int)$option_stock_name['pos_name_id'] . "',
+                                '" . (int)$insert_id . "',
+                                '" . zen_db_input($option_stock_name['pos_name']) . "')");
+        }
```
drbyte, line 190: And this POSM fix, as another separate PR, could probably be merged by @lat9 quickly as well.

admin/languages.php
```diff
@@ -245,6 +388,28 @@
       $db->Execute("DELETE FROM " . TABLE_META_TAGS_PRODUCTS_DESCRIPTION . " WHERE language_id = " . (int)$lID);
       $db->Execute("DELETE FROM " . TABLE_METATAGS_CATEGORIES_DESCRIPTION . " WHERE language_id = " . (int)$lID);
       $db->Execute("DELETE FROM " . TABLE_EZPAGES_CONTENT . " WHERE languages_id = " . (int)$lID);
+      $db->Execute("DELETE FROM " . TABLE_PRODUCTS_OPTIONS_STOCK_NAMES . " WHERE language_id = " . (int)$lID);
```
drbyte, line 391: part of POSM PR too

admin/configuration.php
piloujp: In POSM, some text/title use quote ' which was replaced by _ making constant NAMES__LIKE_THIS (2 _) which does not work well.

admin/languages.php
piloujp: No problem!

admin/languages.php
piloujp: Yes but I realized there is a problem here. If POSM had never been installed, constant TABLE_PRODUCTS_OPTIONS_STOCK_NAMES does not exists.
A test must be added before running this code like this:
`if (defined('TABLE_PRODUCTS_OPTIONS_STOCK_NAMES')) { ...}`
I will submit the corrected version.

admin/languages.php
piloujp: See above. PRs have been submitted.

piloujp:
I removed the mentioned above correction and submitted them in separate PR #6616 and PR #6617 .

admin/languages.php
drbyte: Thanks

admin/configuration.php
drbyte: Hmmm. Good point. Could also run another preg_replace afterward, converting `_+` (means 1 or more in a row) to `_` (single).

drbyte:
I appreciate the effort you've spent on this, and the associated prior, PR.

One concern I have with this approach is the interactive editing of PHP files. We've gone to great lengths in Zen Cart core to avoid doing that, so that its potential security risks are limited.

Still, I appreciate that you've endeavored to make these changes be compatible with all other parts of core code without breaking things ... which is valuable for both core code and upgrades, as well as mods/plugins.

One thing that's very apparent in this PR is the pattern of lookups for these places in the Admin where multilanguage support is lacking. First, it's exposed the dozen (or so) places that need addressing.
In the pattern (which you've used to implement your own pattern of adding columns for additional languages), there exists a potential of perhaps a new db table for translated strings, instead of adding new columns and creating PHP files.

admin/configuration.php
piloujp: Yes, I thought of it, but the logic of changing any 'special' character into _ was kind of reversed. Only space that separate words should be changed and other special characters removed.
Still, in case some title contains two or more space in a row, it might be safer to add this line:
$str = preg_replace('/_+/', '_', $str);

Or perhaps better replace the first line (str_replace) by this:
$str = preg_replace('/[\s ]+/', '_', $str);

piloujp:
> One concern I have with this approach is the interactive editing of PHP files. We've gone to great lengths in Zen Cart core to avoid doing that, so that its potential security risks are limited.

I was afraid of that...
 
> One thing that's very apparent in this PR is the pattern of lookups for these places in the Admin where multilanguage support is lacking. First, it's exposed the dozen (or so) places that need addressing. In the pattern (which you've used to implement your own pattern of adding columns for additional languages), there exists a potential of perhaps a new db table for translated strings, instead of adding new columns and creating PHP files.

That is an interesting idea, although it will certainly need a lot more modification of ZC core files to use this new table, especially in existing queries...

What I am trying to do with this PR is to provide a full translation of admin which is missing for too long, even imperfect or temporary. Imaging when you click on an admin menu link and submenu that appears is written in Chinese, Russian or what ever character you cannot read. For non-alphabetic languages users, it is like this. Even if main admin is English Speaking but want to have an employee that do admin tasks it becomes a problem, at least an increase risk of error and discomfort for the employee.

drbyte:
> What I am trying to do with this PR is to provide a full translation of admin which is missing

Fully understood. I'm in agreement with the objective.

I need to set aside some time to look at this PR more.

admin/category_product_listing.php
```diff
@@ -1264,7 +1264,8 @@
 
               if (empty($product_types)) {
                 // There are no restricted product types so offer all types instead
-                $sql = "SELECT * FROM " . TABLE_PRODUCT_TYPES;
+                $lang_suffix = (!empty($_SESSION['languages_code']) && $_SESSION['languages_code'] != 'en') ? '_' . $_SESSION['languages_code'] : '';
+                $sql = "SELECT type_id, type_name" . $lang_suffix . " AS type_name, type_handler, type_master_type, allow_add_to_cart, default_image, date_added, last_modified FROM " . TABLE_PRODUCT_TYPES;
```
drbyte, line 1268: I think you can continue to use `SELECT *`, without having to list all column names.
I'm pretty sure that a later-declared column replaces the former.
So `SELECT *, type_name$lang_suffix AS type_name` would probably work fine. Have to test, to be sure it doesn't come back as `type_name(1)`.

drbyte: Edit: actually, in this case most of the fields aren't needed. It's only using the `type_id` and `type_name` from this query anyway, because it's only populating a dropdown.

admin/configuration.php
```diff
@@ -161,7 +162,7 @@
                   ?>
                   <?php
                   // multilanguage support:
-                  // For example, in admin/includes/languages/spanish/configuration.php
+                  // For example, in admin/includes/languages/spanish/lang.configuration.php
```
drbyte, line 165: Since this PR is probably going to be open for awhile longer, you are welcome to move this typo-correction (and the other one on line 61) to a separate PR, which we'll merge right away.
I don't mind doing it myself, but it will create a merge-conflict for you to have to fix anyway.

piloujp: No problem, it is done.

admin/category_product_listing.php
piloujp: OK, I am going to test that and see if `type_id` and `type_name` are enough.

piloujp: `type_id` and `type_name` are definitely enough, but just for information, I although tested `SELECT *, type_name$lang_suffix AS type_name` and it worked fine. The last declaration takes over.
I will update that query.

piloujp: I took time to although check other queries and optimized them.

drbyte:
I'm interested in your thoughts about the changes I've committed.

I've created an if-defined-then-use-constant lookup function, and incorporated it in most of the places you had previously updated.
This negates the need for most, if not all, of the `language.php` file changes.

I didn't implement it in the plugin_manager.php file yet, but the function is ready to be used there.

I also didn't touch the `developers_toolkit.php` file. Not sure it even makes sense to edit it in this PR.

drbyte:
I'm concerned that the `product_types.php` changes may be confusing: I think all the edits should be treated as "english only", and translations left to be in constants. I guess if someone's editing or adding product types, they can sort out their language issues themselves. If their site is multilingual (not just "one" language), then perhaps they should consider only making edits in English, and then all the rest be handled via constants.

includes/functions/functions_lookups.php
```diff
@@ -290,3 +290,43 @@ function zen_get_order_status_name(int $order_status_id, int $language_id = 0)
...
+            $lookup = strtoupper('CFGTITLE_' . $lookup_key);
+            break;
+        case 'configuration_key_description':
+            $lookup = strtoupper('CFGDESC_' . $lookup_key);
+            break;
+        case 'configuration_group_title':
+            $str = $lookup_key;
+            $str = preg_replace('/[\s ]+/', '_', $str);
+//            $str = preg_replace('/[^a-zA-Z0-9_\x80-\xff]/', '', $str);
+            $str = preg_replace('/[^a-zA-Z0-9_\x80-\xff]/', '_', $str);
```
drbyte, line 319: I haven't come to a final conclusion on these regex replacements yet. We can test and determine what's simplest for translators to implement. It would be ideal not to break "existing" constants that people may already be using.

drbyte:
Here are some language strings I tested with:
```php
    'PRODUCT_TYPE_NAME_FOR_HANDLER_PRODUCT' => 'Produit - Général',
    'PRODUCT_TYPE_NAME_FOR_HANDLER_PRODUCT_MUSIC' => 'Produit - Musique',
    'PRODUCT_TYPE_NAME_FOR_HANDLER_DOCUMENT_GENERAL' => 'Document - Général',
    'PRODUCT_TYPE_NAME_FOR_HANDLER_DOCUMENT_PRODUCT' => 'Document - Produit',
    'PRODUCT_TYPE_NAME_FOR_HANDLER_PRODUCT_FREE_SHIPPING' => 'Produit - Envoi Gratuit',
    'CFG_GROUP_TITLE_MY_STORE' => 'Mi Tienda',
```

piloujp:
> Here are some language strings I tested with:
> 
> ```
>     'PRODUCT_TYPE_NAME_FOR_HANDLER_PRODUCT' => 'Produit - Général',
>     'PRODUCT_TYPE_NAME_FOR_HANDLER_PRODUCT_MUSIC' => 'Produit - Musique',
>     'PRODUCT_TYPE_NAME_FOR_HANDLER_DOCUMENT_GENERAL' => 'Document - Général',
>     'PRODUCT_TYPE_NAME_FOR_HANDLER_DOCUMENT_PRODUCT' => 'Document - Produit',
>     'PRODUCT_TYPE_NAME_FOR_HANDLER_PRODUCT_FREE_SHIPPING' => 'Produit - Envoi Gratuit',
>     'CFG_GROUP_TITLE_MY_STORE' => 'Mi Tienda',
> ```

Where do you put them?
Where ever I tried they don't seem to be recognized as defined constant by lookup function and only English titles display.
I had another small problem with configuration group which constants are originally prefixed 'CFG_GRP_TITLE_' where prefixed 'CFG_GROUP_TITLE_' in lookup function. This was easily fixed.

piloujp:
> I'm concerned that the `product_types.php` changes may be confusing: I think all the edits should be treated as "english only", and translations left to be in constants. I guess if someone's editing or adding product types, they can sort out their language issues themselves. If their site is multilingual (not just "one" language), then perhaps they should consider only making edits in English, and then all the rest be handled via constants.

I agree, if another language is used when editing, it will be saved in database...
Removing the last 2 calls to function `zen_lookup_admin_menu_language_override` on lines 276 and 337 should do the trick.

Same for developers_toolkit.php, better looking at it, I don't think modification are necessary. Concerned code is used for displaying settings for a key when found but not used in search.

drbyte:
> originally prefixed 'CFG_GRP_TITLE_' where prefixed 'CFG_GROUP_TITLE_' in lookup function. This was easily fixed

Oops. Yes. My mistake.

piloujp:
After few modifications, I was able use the new function to display the right language for `configuration_group`, `configuration`, `product_types`, `product_type_layout` and `plugin_control` tables. All this without modifying database at all.
For tables that have a language_id, like `orders_status` and `products_options_stock_names` , I don't think constants should be used as it won't go very well with user edit that are possible through admin menus.
Another part that is not covered by this function is plugin configurations.
I am still testing and working on possible options.

drbyte:
In relation to plugins, I think ultimately we need to enhance some of that infrastructure to allow multilingual stuff via the plugin's own files. We can maybe treat the topic of plugin configurations in a separate PR.

Also: `PLUGIN_DESCRIPTION_FOR_`... as a prefix can probably be improved. For example: if the "define" for that is going to go into a plugin's own language files, then perhaps renaming it to `ADMIN_PLUGIN_MANAGER_DESCRIPTION_FOR_` might be more intuitive/contextual?

drbyte:
> For tables that have a language_id, like `orders_status` and `products_options_stock_names` , I don't think constants should be used as it won't go very well with user edit that are possible through admin menus.

If it seems necessary to keep these in the database, then I'm open to adding a db table for those text strings, like we have in other places. Doing it for orders_status is straightforward. (But not within this current PR).

Doing it for POSM should be in consultation with @lat9, and also not within this current PR.

includes/functions/functions_lookups.php
```diff
@@ -290,3 +290,43 @@ function zen_get_order_status_name(int $order_status_id, int $language_id = 0)
...
 }
 
+
+function zen_lookup_admin_menu_language_override(string $lookup_type, string $lookup_key, string $fallback): string
+{
+    switch ($lookup_type) {
+        case 'order_status_name':
+            $lookup = strtoupper('ORDER_STATUS_NAME_FOR_' . $lookup_key);
+            break;
+        case 'product_type':
```
drbyte, line 300: @piloujp After looking at this more, I think we should rename this one lookup_type from `product_type` to `product_type_name`. It feels more logical/intuitive. Unfortunately it means updating the places where it's been called.

piloujp: @drbyte I have just sent a simplified version without all plugin stuffs and few modifications per your remarks.
I agree something should be done for internationalization on plugin side rather than in ZC files.

piloujp:
> > For tables that have a language_id, like `orders_status` and `products_options_stock_names` , I don't think constants should be used as it won't go very well with user edit that are possible through admin menus.
> 
> If it seems necessary to keep these in the database, then I'm open to adding a db table for those text strings, like we have in other places. Doing it for orders_status is straightforward. (But not within this current PR).
> 
> Doing it for POSM should be in consultation with @lat9, and also not within this current PR.

I would be happy to work on it if you could precise a little bit how you see it.

piloujp:
I forgot, languages files for testing are available in French language pack.
lang.configuration.php
lang.product_type.php
lang.plugin_manager.php

drbyte:
I've updated the original PR description to include basic documentation for the features added by this PR.

//cc @scottcwilson

lat9:
> > For tables that have a language_id, like `orders_status` and `products_options_stock_names` , I don't think constants should be used as it won't go very well with user edit that are possible through admin menus.
> 
> If it seems necessary to keep these in the database, then I'm open to adding a db table for those text strings, like we have in other places. Doing it for orders_status is straightforward. (But not within this current PR).
> 
> Doing it for POSM should be in consultation with @lat9, and also not within this current PR.

@piloujp, I don't understand the comment regarding these already language-based tables.  What won't "go very well"?

piloujp:
> @piloujp, I don't understand the comment regarding these already language-based tables. What won't "go very well"?

@lat9 There is some double definition for one translation. If constants are used and database values for non-English languages have been updated through admin editing (for example admin::localization::Orders Status), then which one should be the current one? If constants override database fields, then database editing (through admin interface) are useless. Just using constants like done in other place is not enough, database needs to be modified and queries... Which is not necessary the best choice.
There are more choices/modifications to do than actual lookup function update.

drbyte:
Oh. Yes, it's true.
When adding a new language, the `orders_status` table records are copied to the new `language_id`.
And any subsequent admin-UI edits to each order-status record, allows customizing the per-language `name` of each status.

I guess we could allow the "copy" process to read some language-defines for pre-populating those copied records with the translated value (if the constant exists) instead of just copying the english value. This would be used only one-time, during the "add new language" process.

Maybe in `lang.orders_status.php`: 
```php
'INSTALL_ORDER_STATUS_PENDING' => 'En attente',
'INSTALL_ORDER_STATUS_PROCESSING' => 'En cours',
'INSTALL_ORDER_STATUS_DELIVERED' => 'Livré',
'INSTALL_ORDER_STATUS_UPDATE' => 'Mise à jour',
```

piloujp:
> I don't understand the "double definition for one translation". Each of the two tables referenced have PRIMARY key defined as an id in conjunction with a `language_id`.

@lat9 When I say double definition, it is for translated string. If using constants with lookup function as with other part of admin menus, there will be a translation of English string in the constant and another one in database (copy of English one created when adding new language).

piloujp:
> Oh. Yes, it's true. When adding a new language, the `orders_status` table records are copied to the new `language_id`. And any subsequent admin-UI edits to each order-status record, allows customizing the per-language `name` of each status.
> 
> I guess we could allow the "copy" process to read some language-defines for pre-populating those copied records with the translated value (if the constant exists) instead of just copying the english value. This would be used only one-time, during the "add new language" process.
> 
> Maybe in `lang.orders_status.php`:
> 
> ```
> 'INSTALL_ORDER_STATUS_PENDING' => 'En attente',
> 'INSTALL_ORDER_STATUS_PROCESSING' => 'En cours',
> 'INSTALL_ORDER_STATUS_DELIVERED' => 'Livré',
> 'INSTALL_ORDER_STATUS_UPDATE' => 'Mise à jour',
> ```

Yes, exactly! It is what I did in language.php except I did not use language constants but arrays instead.

lat9:
> Okay. That should work for POSM as well.
> 
> Actually, I think those INSTALL_xxxx defines need to be in lang.languages.php for both order-status and POSM, so that the copy query can lookup those constants when creating the new lang-specific cloned records.

It's best to plunk that file into `/admin/includes/languages/{language}/extra_definitions` to enable the definitions to be loaded for any page that requires those pre-fills.

drbyte:
Okay, since they'll be maybe only a dozen defines, extra_definitions would be fine.
It just feels "off" to put the orders-status ones into RAM on every page despite only being utilized on one page. 
But for POSM and other plugins, I agree: they need to be accessible from an unpredictable page location.

OR the plugin-related language defines could go into the plugin's languages subdirs. That'd be ideal, but harder for language-pack-creators to prepare files for. And nearly impossible for Crowdin use.

lat9:
The extra-language definitions file could check the `$current_page` and include those installation defaults IFF the associated page is current.

drbyte:
... and for Encapsulated plugins, I would expect that the plugin's own code would be responsible for searching for said constants when it does its own copying/cloning of English records to support other languages.

If it's deemed something that all plugins should share code for, we can consider incorporating something in core for that.

piloujp:
I did some tests with this code, placing at first the constants above in `../admin/includes/languages/french/lang.languages.php` or a new file `../admin/includes/languages/french/extra_definitions/lang.install_orders_satus.php`, and it did not work, it was like constants were not defined.
Problem was that when setting a new language (French), I was using English. Putting constants in one of the above file but in `english` folder fixed it.
But what then if there are 3 or more languages installed? It would be better if these languages constants could be loaded independently of language used. I don't know if it is possible with actual code.

drbyte:
@piloujp Good point. Can you re-post this same detail in PR 6712. That's a significant complication, and maybe we ought to revert this order-status change and find another approach. But let's discuss it over there, because it's different than the (working) improvements in the PR on this present page.

Issue #6600, opened 2024-07-20
Sales Weekly reports display


piloujp:
When looking at sales weekly reports, at first it is fine but when clicking few times on 'Previous' button (you need to have data for that), weeks start to shift. Instead of starting with the first week of the month, it starts with one, two, three or more weeks before depending on how far you went back, same problem with last week.
The number of weeks displayed does not correspond to the number of week in the month either.
The first week displayed for each page should be the first week of the month (containing the first day) and the last one should be the last week of the month (containing the last day).

PR #6601, opened 2024-07-20
Weekly sales reports display months by months

piloujp:
Fixes issue #6600


scottcwilson:
It goes back by groups of 5 weeks at a time.  What's there now looks fine to me, I don't think it makes sense to change.

piloujp:
I think these weekly reports are made to be viewed one month of data at a time with week by week details like on a table calendar going from Sunday to Saturday for columns and first week of the month on first line and last line with last week.
Which means between 4 (rarely) and 6 lines per month.
That said, try this:
Click four times on Previous:  You should have five weeks per page each time, as you said.
Then click on Next until you are back to last data of this month:  I get an alternance of three and four lines pages and need to click six times.
Then, if I click on Previous again, at second time I get a nine lines page!

scottcwilson:
If you want to change it, please make it jump back (and forward) by groups of 5 weeks.  It doesn't need to be month by month, people are looking for weekly numbers, not monthly.

piloujp:
Why 5 weeks then? There is more room on the page.
Plus:
Hourly reports --> 1 day
Daily reports --> 1 week
Weekly reports --> 5 weeks ???
Monthly reports --> 1 year of data

Sorry to insist like that, but before I do something I like to understand why.

scottcwilson:
Pick a number 5,6,8. Enough to give context.  You're trying to see your weekly performance.

scottcwilson:
I appreciate the work you have done on this but it seems like any change we make is going to make someone (who's used to the old format) unhappy.  I think we should put this to bed.

PR #6604, opened 2024-07-21
Child template prep: Split out CSS/JS loading from html_header.php

lat9:
Part of [this](https://github.com/zencart/zencart/discussions/6428) discussion.

### DIFF
--- a/includes/templates/responsive_classic/common/html_header.php
+++ b/includes/templates/responsive_classic/common/html_header.php
@@ -90,72 +90,9 @@ if (count($languages = $lng->get_language_list()) > 1) {
 
 <?php
 /**
- * load all template-specific stylesheets, named like "style*.css", alphabetically
+ * Load all template-specific stylesheets, via the common CSS loader.
  */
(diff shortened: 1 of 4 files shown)


scottcwilson:
Love it, happy to merge it, but one question - should we add some notifiers to allow people to do any custom loading needed?

lat9:
Not sure it's needed as the template can provide an override in its `common` sub-directory.  I'll be handling this in the bootstrap override as
```php

// -----
// If the 'common' CSS loader is present, use that.  Note this module can be removed
// from the template once support for Zen Cart versions less than v2.1.0 is dropped!
//
if (file_exists(DIR_WS_TEMPLATES . 'template_default/common/html_header_css_loader.php')) {
    require DIR_WS_TEMPLATES . 'template_default/common/html_header_css_loader.php';
    return;
}
//- Otherwise, the file's not present; functionality repeated below
```

scottcwilson:
Fair.  If someone asks for it, we can add it later.

PR #6614, opened 2024-07-24
Undefined array key "coupon_free_ship" in coupon_admin.php

scottcwilson:
```
--> PHP Warning: Undefined array key "coupon_free_ship" in /home/client/public_html/admin/coupon_admin.php on line 377.
```

(Line number from 1.5.8a)

### DIFF
--- a/admin/coupon_admin.php
+++ b/admin/coupon_admin.php
@@ -225,7 +225,7 @@ switch ($_GET['action']) {
       $update_errors = 1;
       $messageStack->add(ERROR_NO_COUPON_NAME, 'error');
     }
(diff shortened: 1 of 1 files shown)


drbyte:
Looks fine. But perhaps most of those similar checks for `!` on array elements should actually be `!empty()`. 
Must account for cases where `0` is actually desired, but even then the original logic is probably meaning "not zero", so "not empty" is acceptable since it also accommodates null.

scottcwilson:
I think the checkbox `coupon_free_ship` is the only one that's going to produce a log like this.   Let's just stop at the one known log.

PR #6616, opened 2024-07-25
Table Order status' field sort_order missing when adding new language

piloujp:
When adding a new language, field sort_order in table order_status is not copied/updated for the new language.

### DIFF
--- a/admin/languages.php
+++ b/admin/languages.php
@@ -139,7 +139,7 @@ if (!empty($action)) {
         }
 
 // create additional orders_status records
-        $orders_status = $db->Execute("SELECT orders_status_id, orders_status_name
+        $orders_status = $db->Execute("SELECT orders_status_id, orders_status_name, sort_order
                                        FROM " . TABLE_ORDERS_STATUS . "
                                        WHERE language_id = " . (int)$_SESSION['languages_id']);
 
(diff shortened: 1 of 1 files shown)


admin/languages.php
```diff
@@ -139,15 +139,16 @@
...
                                        WHERE language_id = " . (int)$_SESSION['languages_id']);
 
         foreach ($orders_status as $status) {
-          $db->Execute("INSERT INTO " . TABLE_ORDERS_STATUS . " (orders_status_id, language_id, orders_status_name)
+          $db->Execute("INSERT INTO " . TABLE_ORDERS_STATUS . " (orders_status_id, language_id, orders_status_name, sort_order)
                         VALUES ('" . (int)$status['orders_status_id'] . "',
                                 '" . (int)$insert_id . "',
-                                '" . zen_db_input($status['orders_status_name']) . "')");
+                                '" . zen_db_input($status['orders_status_name']) . "',
+                                '" . (int)$status['sort_order'] . "')");
```
lat9, line 151: Single quotes around integer values for a database field are neither needed nor wanted and the casting is redundant as you've just pulled the information from the database.

lat9 (changes requested):
Just that update on line 151.

admin/languages.php
piloujp: You are right, I stupidly reproduced original formatting... I changed that for the 3 integer values.

lat9 (approved):

PR #6628, opened 2024-08-04
Make banner-image upload more intuitive

### LINKED ISSUE #6624: Admin, Banner Manager, Image file definition
2024-07-31 18_10_00-Admin Banner Manager — Mozilla Firefox

I find this block of GUI and texts quite confusing.

Browsing to a file (with target empty or entering the full path) and Update does not work unless the (old) image referenced in the input field is deleted.

Image Target is the optional subdirectory.

I've not come up with an easy change to make this more intuitive...maybe it needs a rethink to be more like the product image edit section?

2024-07-31 18_09_22-Admin Product — Mozilla Firefox

drbyte:
- shows paths inline with the input field, for improved clarity
- added tiny group border around fields
- made banner-group dropdown field (above image fields) narrower, as it doesn't need to be full-width, and this width-change helps distinguish it as "different" from the image fields.
- no longer need to manually "delete" the old image filename in order for an uploaded image to replace it
- existing/local image is used if present; but if an upload is POSTed and validates, it overrides the local/prior image
- target subdirectory now also defaults to "banners/"

BEFORE:
Screen Shot 2024-08-03 at 11 04 35 PM

AFTER:
Screen Shot 2024-08-03 at 11 15 05 PM

Revised:
Screen Shot 2024-08-07 at 1 52 06 AM

Fixes #6624


torvista:
Much improved.

I think the current image should be the first item.
How about keeping the target next to the browse.
Since the user-selected folder can only be below the pre-set images folder, just offer that as the suggestion.
I think "local" can be misinterpreted.

My fuddled brain finds this demands the minimum thought:
banner_manager

drbyte:
Updated and reordered fields:
Screen Shot 2024-08-07 at 1 52 06 AM

torvista:
Fine for me!

PR #6630, opened 2024-08-04
Update layout_controller to be more UI-friendly

lat9:
Instead of going through the sideboxes one-by-one, this update makes use of jQuery UI's `sortable` widget to enable the sideboxes to drag-and-drop in the various areas for display.

See the on-screen notes for additional information.

Noting that this update came to life in collaboration with @barco57


torvista:
A big step in the right direction.

I had a missing sidebox notice. I ignored it, moved an active sidebox and saved: got a messagestack about a placeholder, and storefont did not change.

After deleting the missing sidebox, things worked as advertised.

lat9:
> A big step in the right direction.
> 
> I had a missing sidebox notice. I ignored it, moved an active sidebox and saved: got a messagestack about a placeholder, and storefont did not change.
> 
> After deleting the missing sidebox, things worked as advertised.

Missing a `</form` tag in that missing boxes' section; should be ok-fine now.

torvista:
image

Clicking on Remove Boxes has no effect (Firefox(Chrome).

torvista:
I guess the use and display of this lot is outside scope for the moment. But it is confusing...."why can't I see these active single sideboxes??"
image

lat9:
> I guess the use and display of this lot is outside scope for the moment. But it is confusing...."why can't I see these active single sideboxes??"

Somewhat outside of the scope, since the single-column usage is _very_ template-dependent.  Perhaps a `?` type bubble popover that indicates the display of the information can vary based on the template in use.

drbyte:
I strongly desire to delete all the single-column sideboxes except for the 3 that aren't really even column-sideboxes ... they're header-related: currency, language, search.

That would make the presence of them very intuitive.

I'm not aware of any template that (correctly and intentionally) uses any of the non-header single-column sideboxes.

torvista:
I've done stuff to re-use Category and Brands sideboxes code as single-column items in a header, as drop-down lists. That's a common layout. Needs work to allow that.

lat9:
How about that a template, in addition to identifying that it supports single-column settings, identifies _which_ single-column boxes it *does* support?

drbyte:
> How about that a template, in addition to identifying that it supports single-column settings, identifies which single-column boxes it does support?

The second-most common complaint I've heard about the Layout Controller is that the "single-column" stuff makes the whole concept confusing.

And they're completely right. Because no ZC template has ever used single-column.

drbyte:
(Apart from `search_header`, which really isn't a "side" box anyway), a sidebox module is just that: a module.
A sidebox has no knowledge of whether it's assigned to left-col, right-col, or single-col.
The idea that a sidebox can be single-column (never used ever) or dual-column has nothing to do with the sidebox itself.
(It's determined by the active template, which hard-codes calls to `tpl_box_default_left.php`/`..._right.php`)

No ZC templates I've seen have ever used single-column layout. Thus there's not really any such thing as a "single column sidebox".
Yes, some have (perhaps even at my behest years ago to avoid having to add more UI's) monkey-patched the use of the otherwise-unused single-column "feature" to leverage an admin UI to enable/disable some of the sideboxes when used in "other" places, such as "header".

I think a more intuitive UI would be one that offered only LEFT/RIGHT. And if a template ever "wanted" to be SINGLE, it should just push all sideboxes to LEFT to enable/disable and set sort order.

And then the single-column UI sorter should become a "header modules" sorter. And offer only the modules whose filename ends with `_header`. (Common ones would be the built-in `search_header`, and the commonly-added `languages_header` and `currencies_header`.)

drbyte:
(It's been discussed elsewhere that currencies and languages are often used in a template header. True; however, they're different from the sideboxes by similar name, as they render differently and with a different template.)

We could ADD those two header "sideboxes" to core, to make it easier for them to be enabled in the header by default.

drbyte:
Apologies: I'm not trying to hijack this PR at all.

Given that there's much discussion about the single-column "column", I wanted to clarify that we have an opportunity here (even if it's via a separate PR) to make that portion of this page even more intuitive to all who use it.

lat9:
> How about that a template, in addition to identifying that it supports single-column settings, identifies _which_ single-column boxes it _does_ support?

@drbyte, was this suggestion not acceptable?  That would be accompanied by changing the "Single-Column" elements to read "Template-Specific".

TMCSherpa:
jscript_responsive_framework.php, responsive_default.css and responsive_tablet.css hide the sideboxes by default even when they are enabled for single column. Most of them anyway so make sure whatever you want to show doesn't have it's css set to {display:none;visibility:hidden;}
The whole thing is kinda weird and (other than header_search which should get moved out of sideboxes) the fix isn't horrible
Also I'm finishing code that adds brands/manufacturers, currencies and languages to the menu if that helps?
It's like half useful half derailment so I feel kind of ok.

lat9:
OK, let's try this functional description on for size.

The tool will be updated to have four sortable/draggable areas:

1. Main-Page Boxes, containing three sections.  All sideboxes whose name doesn't end in either `_header` or `_footer` will show in this section.
   1. Active Left-Column Boxes
   2. Active Right-Column Boxes
   3. Inactive Main-Page Boxes

2. Header Boxes, containing two sections. If a sidebox file's name ends in `_header`, it'll show in this section
   1. Active Boxes
   2. Inactive Boxes

3. Footer Boxes, containing two sections.  If a sidebox file's name ends in `_footer`, it'll show in this section
   1. Active Boxes
   2. Inactive Boxes

4. Mobile-Menu Boxes, containing two sections. All sideboxes whose name doesn't end in either `_header` or `_footer` will show in this section.  This section will be collapsible and the template can "opt-out" of its display by including `$does_not_use_mobile_settings = true` in its `template_info.php` file.
   1. Active Boxes
   2. Inactive Boxes

   The "Header Boxes", "Footer Boxes" and "Mobile-Menu Boxes" areas are displayed *only if* the template includes `$uses_single_column_layout_settings = true;` in its template_info.php` file.

TMCSherpa:
Adding a tag is getting pretty complicated but I do like the idea of adding top/bottom header/footer to the database.
Personally I think having a working center column is important for info and banner boxes.
The fix to get sideboxes to display left/right/up/down properly isn't that difficult but it's almost all template files.

torvista:
That functional description seems logical/understandable, unlike the current schema.

Possibly the same /modules code could be used for all four, with just the template styling/wrapping being different.

TMCSherpa:
Currently single column is confusing because doesn't work the way it says it does.
layout_box_status_single is only used for search_header and 
layout_box_sort_order_single does nothing.

lat9:
> Currently single column is confusing because doesn't work the way it says it does. layout_box_status_single is only used for search_header and layout_box_sort_order_single does nothing.

Right, the 'single-box' settings are *very* template-dependent.  There will be some (hopefully) helpful popup text on that page that indicates as such.

torvista:
Looks good, can it not go in the next build?

lat9:
> Looks good, can it not go in the next build?

I believe so; let's see what @drbyte has to say.

admin/includes/css/layout_controller.css
```diff
@@ -0,0 +1,94 @@
...
+}
+.d-flex {
+    display: flex;
+}
+.justify-content-around {
+    justify-content: space-around;
+}
+.mb-0 {
+    margin-bottom: 0;
+}
```
drbyte, line 36: Should these BS4/5 styles be moved to the core stylesheet?
There are several other bridging styles there already.

drbyte:
Thanks @lat9 and @barco57 for all your work on this!

admin/includes/css/layout_controller.css
drbyte: We can do that in a separate PR.

torvista:
Is there documentation on the new naming of sideboxes for the side/for the center?

PR #6632, opened 2024-08-04
Version stamping 2.1.0-alpha1

scottcwilson:


torvista:
[MOAN]
For the bleeding-edgers keeping their site up to date with development, these are 156 files with no changes from yesterday to ignore for the time being, or review, followed by the same again on release. All requiring manual comparison, "just in case".

Is a hard-coded version number really necessary given GitHub Blame? Can you not just use a date-last-updated and forget all this? The rest of the year is trying to **not** update files for no good reason.
[/MOAN]

Issue #6247, opened 2024-02-24
Enhancement: Featured Categories


drbyte:
And, looking at the above, I also found something many of my customers could use.  We already have Featured Products, but what about adding [Featured Categories](https://www.semrush.com/blog/seo-ecommerce-category-pages/)?

_Originally posted by @dbltoe in https://github.com/zencart/zencart/discussions/5992#discussioncomment-8180925_

scottcwilson:
Interesting idea.  Centerbox and page for this?

lat9:
Wouldn't this (not small) change go against the guidance just posted on the Zen Cart forums: https://www.zen-cart.com/showthread.php?229962-Feedback-on-2-0-0-beta1

drbyte:
> Wouldn't this (not small) change go against the guidance just posted on the Zen Cart forums: https://www.zen-cart.com/showthread.php?229962-Feedback-on-2-0-0-beta1

If one tried to get it into v2.0.0 release, yes. But there's no version target in this discussion yet.

scottcwilson:
Could be a "next release after 2.0.0" feature.  We need to start building our backlog! :)

scottcwilson:
@drbyte If you are agreeable, please label this 2.1.x.  Thanks!

drbyte:
I'm agreeable to 2.1, but not sure who will code this feature or when that will happen, so untagging it for now.

TMCSherpa:
I have a local copy that 'works' but some idiot nuked the SQL file.
You can add categories, delete them and display them (yay) but the breadcrumbs aren't working properly (parent ID I think) and I know there should be a hook so that deleting categories removes them from featured.
I'll get the SQL file rebuilt so someone can play with it

scottcwilson:
John, you should be able to rebuild the SQL file by looking at your current database schema and comparing it to the default schema.  Thanks for sharing your work.

TMCSherpa:
I made a branch but I don't think it's ready for a pull yet. I need to work on hiding the sidebox for mobile and a couple other things first.

PR #6633, opened 2024-08-06
 Enhancement: Featured Categories #6247 

TMCSherpa:
It's mostly made from featured (products) files that have been reworked for use with categories and it works the same way. Everything is turned on by default (category sidebox and the index page is set to group 5)


TMCSherpa:
Are you sure your test scripts are working? I get the same errors locally on a stock version of zencart.

zcwilt:
Couple of notes.

If you want to see the logs behind why the tests are failing see 

https://github.com/zencart/zencart/actions/runs/10275975225?pr=6633

At the bottom of this page you can download the artifact file which contains error logs.

The reason the tests are failing is that your FeaturedCategory.sql file is never loaded by the tests system, and would also never be loaded in a Zen Cart install.

scottcwilson:
@TMCSherpa Are you updating the dates in the file headers by hand?  Please don't do that.  The dates will get set correctly when the next official build is run.

TMCSherpa:
I'll put it back. Let me fix the SQL stuff and resubmit

PR #6594, opened 2024-07-18
Update class.zcDatabaseInstaller.php

brittainmark:
fixes #6593

line 73 move update ignore  before update to allow 
line 223 efficiency change once you have  found the command don't check the remaining commands.

PR #6650, opened 2024-08-11
Convert zc_install to Bootstrap 5

drbyte:
Remove Zurb Foundation framework, and replace with Bootstrap 5.
Alerts with no additional help text are no longer clickable.

install 1
modal 1
install 2
install 3
progress 1
progress 2
install 4
install 5
upgrade 1


scottcwilson:
WOW!  Beautiful and much more clear.

dbltoe:
As we Texans say, "Mucho more gooderer!"

lat9:
Why were we "not able to change the name of your admin folder automatically"?

scottcwilson:
Probably had the `./includes/extra_configures/dev-override_admin_rename_dialog.php` file in his test environment (it's in Github in the project, just not deployed).

drbyte:
> Why were we "not able to change the name of your admin folder automatically"?

Yes, that's just an artifact of my local dev overrides. When those overrides aren't in place, the message is different, accordingly.

lat9:
> > Why were we "not able to change the name of your admin folder automatically"?
> 
> Yes, that's just an artifact of my local dev overrides. When those overrides aren't in place, the message is different, accordingly.

Whew, I felt like we were going "back to the future" as I remember having to manually change the directory name back when dinosaurs roamed.

drbyte:
> Whew, I felt like we were going "back to the future" ...

I wondered if you had a DeLorean stashed away somewhere!

lat9:
Performing an initial install, I created an empty database using `utf8_general_ci` collation.  The `zc_install` process completed successfully, but all tables/table-fields were created with `utf8mb4_general_ci` collation, which will lead to potential collation-mismatch errors in the future, since any additional tables added will inherit the database's default `utf8_general_ci` collation.

drbyte:
Noteworthy. 
But I don't think that's anything that got changed in this conversion to bootstrap. 


> On Aug 17, 2024, at 10:20 AM, lat9 wrote:
> Performing an initial install, I created an empty database using utf8_general_ci collation. The zc_install process completed successfully, but all tables/table-fields were created with utf8mb4_general_ci collation, which will lead to potential collation-mismatch errors in the future, since any additional tables added will inherit the database's default utf8_general_ci collation.

lat9:
Upgrade from zc139h (with some _very_ wonky `configure.php` files) went without issue (other than the above).

scottcwilson:
Is this actually a documentation issue?  Should we give guidance to use `utf8mb4_general_ci` as the collation when creating a new database?

lat9:
Upgraded a zc157 site without issue, too, but I'm seeing some common zcInstallLog*.log files that contain the following (which doesn't make sense to me since those table-fields are present):
```
---------------
Aug 17 2024 15:48 -- systemChecker::findCurrentDbVersion
2.0.0

---------------
Aug 17 2024 15:48 -- upgradeException
ALTER TABLE address_book MODIFY entry_postcode varchar(64) NOT NULL default '';  Cannot CHANGE column entry_postcode because it does not exist.  C:\xampp\htdocs\install\zc_install/sql/updates/mysql_upgrade_zencart_210.sql

---------------
Aug 17 2024 15:48 -- upgradeException
ALTER TABLE orders MODIFY customers_postcode varchar(64) NOT NULL default '';  Cannot CHANGE column customers_postcode because it does not exist.  C:\xampp\htdocs\install\zc_install/sql/updates/mysql_upgrade_zencart_210.sql

---------------
Aug 17 2024 15:48 -- upgradeException
ALTER TABLE orders MODIFY delivery_postcode varchar(64) NOT NULL default '';  Cannot CHANGE column delivery_postcode because it does not exist.  C:\xampp\htdocs\install\zc_install/sql/updates/mysql_upgrade_zencart_210.sql

---------------
Aug 17 2024 15:48 -- upgradeException
ALTER TABLE orders MODIFY billing_postcode varchar(64) NOT NULL default '';  Cannot CHANGE column billing_postcode because it does not exist.  C:\xampp\htdocs\install\zc_install/sql/updates/mysql_upgrade_zencart_210.sql

---------------
Aug 17 2024 15:48 -- upgradeException
ALTER TABLE paypal MODIFY address_zip varchar(64) default NULL;  Cannot CHANGE column address_zip because it does not exist.  C:\xampp\htdocs\install\zc_install/sql/updates/mysql_upgrade_zencart_210.sql

---------------
Aug 17 2024 15:48 -- upgradeException
ALTER TABLE paypal_testing MODIFY address_zip varchar(64) default NULL;  Cannot CHANGE column address_zip because it does not exist.  C:\xampp\htdocs\install\zc_install/sql/updates/mysql_upgrade_zencart_210.sql
```

lat9:
> Is this actually a documentation issue? Should we give guidance to use `utf8mb4_general_ci` as the collation when creating a new database?

Kind of (see #6664), since the issue also appears in upgrades.

lat9:
zc157c upgrade; `project_version_history` table missing the version update from 1.5.8->2.0.0:

image

scottcwilson:
This is a known bug, fixed in 2.1.0. 
https://github.com/zencart/zencart/pull/6594

lat9:
> This is a known bug, fixed in 2.1.0. #6594
> 
> You can pull the fix back into your zc_install folder, reload the 1.3.9 db and rerun if it's not too late.

I'm using the `zc_install` update for this testing, so that update _should_ already be baked in.

scottcwilson:
You're right - I was thinking about a recent conversation where you were talking about a 2.0.1 upgrade.  

I can't test right now but should be able to do some next week.

lat9:
If an upgrade is run as multiple steps, e.g. an upgrade from zc157 to zc210, the `zero_dates_cleanup.sql` is run at the start of **every** step.  Isn't it sufficient to run that once?

For large databases, this can be a time-consumer.

scottcwilson:
We could just migrate it to the latest upgrade file (and remove it from the older files) on every release.

drbyte:
> For large databases, this can be a time-consumer.

Agreed.

> We could just migrate it to the latest upgrade file (and remove it from the older files) on every release.

Well, it needs to be done as early as possible, else subsequent table alterations may fail.

And if it's done before some tables or columns are created then those won't be caught by the script (granted, that should be moot since those won't (shouldn't) contain any invalid data).

Can be addressed in a separate Issue.

drbyte:
Thanks for the thorough inspections!

I've made separate issues for the non-UI-related concerns.

Given that no mention has been made of glitchy UI details, nor glitchy ajax calls, I'm inclined to merge. Any objections?

lat9:
> We could just migrate it to the latest upgrade file (and remove it from the older files) on every release.

It's already a separate file, so it's something programmatic that's bringing it in on each and every update 'bump'.

PR #6652, opened 2024-08-12
Additional sanitization

lat9:
Protecting against URLs like `/index.php?cPath=244_249&main_page=product_info&products_id=12345'"\'\");|]*%00{%0d%0a<%00>%bf%27'💡`

1. Initially all $_GET variables should be strings, protecting against arrays and `NUL` values.
2. Adding 'end' string for products-id matching REGEX.

### DIFF
--- a/includes/init_includes/init_sanitize.php
+++ b/includes/init_includes/init_sanitize.php
@@ -19,10 +19,10 @@ if (!defined('IS_ADMIN_FLAG')) {
 $zco_notifier->notify('NOTIFY_INIT_SANITIZE_STARTS');
 
 foreach ($_GET as $varname => $varvalue) {
-    if (is_array($varvalue)) {
-        $site_array_override = false;
-        $zco_notifier->notify('NOTIFY_INIT_SANITIZE_GET_VAR_CHECK', ['name' => $varname, 'value' => $varvalue,], $site_array_override);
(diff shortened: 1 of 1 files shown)


includes/init_includes/init_sanitize.php
```diff
@@ -56,18 +56,19 @@
 // value or a uprid (dddd:xxxx), where xxxx is the 32-hexadecimal character md5 hash of the currently-selected
 // attributes.
 //
+// Noting that if an id-value is found to be invalid, there's no since
```
drbyte, line 59:
```suggestion
// Noting that if an id-value is found to be invalid, there's no sense
```

drbyte (approved):
Review note: In our present architecture, we are indeed relying on GET values arriving as text. Therefore numeric-only product-id's are still strings at this stage of execution. It's only post-init_sanitization that they're treated as (or assessed for being) integers.

torvista:
Breaks DPU

https://www.zen-cart.com/showthread.php?70577-Dynamic-Price-Updater&p=1404116#post1404116

lat9:
I'm sure I've asked before, but rather than adding a comment to a merged PR (thus reducing its visibility), please either open a new issue or a PR that addresses the solution.

torvista:
True, but I didn't this was a problem with the core code.

PR #6653, opened 2024-08-12
Restore auto-focus to admin search

lat9:
It was lost in the zc157->zc158 transition.  Used to be only on Categories/Product, but now since the search-box is used multiple places, it'll be widespread.


drbyte:
Looks like the only conflicting admin page would be the customers page, and only when using the password-reset feature.

lat9:
Perhaps a PHP variable to set that instructs the search-form to "not" autofocus?

drbyte (approved):

PR #6658, opened 2024-08-14
Fix unusual zero-date values during upgrade

drbyte:
Discovered these in an upgrade of old data, which were not getting fixed by the current script:
```
0000-00-00 00:00:02
0000-00-00 00:00:03
0000-00-00 00:00:04
0000-00-00 00:00:05
thru
0000-00-00 00:00:58
```


drbyte:
@lat9 commented elsewhere: 
> Ugh, that's going to additionally extend the upgrade time required for sites with a _large_ number of products.

drbyte:
Changed the strategy. Using `LIKE` with wildcards captures the most likely scenarios.
Still scans every record, so expecting little difference in performance penalty.

Could be more "accurate" by doing a triple set of LIKE comparisons: one for years, one for days, one for months. But I'm not convinced that's necessary. And it would take three times longer.

PR #6661, opened 2024-08-15
Shipping estimator display errors

### LINKED ISSUE #6659: Shipping estimator doesn't display shipping-module error message
See similar issue/correction for the Bootstrap template: https://github.com/lat9/ZCA-Bootstrap-Template/issues/419

lat9:
Fixes #6659

### DIFF
--- a/includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
+++ b/includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
@@ -7,6 +7,15 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: Steve 2024 May 31 Modified in v2.1.0-alpha1 $
  */
+if ($_SESSION['cart']->count_contents() === 0) {
+    return;
+}
+
+if (empty($extra)) {
+    $extra = '';
+} else {
+    $extra = ' class="' . $extra . '"';
+}
 ?>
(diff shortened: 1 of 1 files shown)


includes/templates/template_default/templates/tpl_modules_shipping_estimator.php
```diff
@@ -48,9 +48,8 @@
     </address>
     <br class="clearBoth">
 <?php
-} else {
-    if ($_SESSION['cart']->get_content_type() !== 'virtual') {
-        $flag_show_pulldown_states = (ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN === 'true');
+} elseif ($_SESSION['cart']->get_content_type() !== 'virtual') {
```
proseLA, line 51: i frequently do multiple commits; first correct the issue, and then reformat/refactor.  here you have done the reverse...  of which i thought about; and of which i can be converted to be a fan.

that said, this section of code looks to be a refactor/reformat; and not the part of the actual correction.

it would be great to just look at the actual correction and not superfluous stuff in a commit labeled as such.  

because for the rest of this commit, i have no idea where the correction now is...  nor am i now willing to now commit the time to finding it.

lat9: Yep, me too, I just "got on a roll" and saw those couple of other refinements when I made the 'actual' change.

Issue #6204, opened 2024-02-03
product, add to cart form tags


torvista:
1) The enclosing tags for the form seem far wider than necessary/very detached from the form fields.
The start tag is right after the breadcrumb, encompassing the category image, prev-next, product name, Ask a Question.
The end tag is...at the end, including description, reviews, Also Purchased.

Doesn't seem right and not an error, but to add another form like Back in Stock/to put AAQ on the same page/merge the two, requires moving that closing tag up to where it actually should be.
Why not put the form tags only around the input elements?

2) The form id "AddToCartForm" and name "cart_quantity" are not the same.
Not a bug either, but it seems convention is to keep them the same.

While logically AddtoCartForm would be the better one to use, it is not used anywhere and so easy to replace, unlike cart_quantity which would cause much wailing and gnashing of teeth with plugin authors if touched.

torvista:
addressed here: https://github.com/zencart/zencart/pull/6663

drbyte:
You are 100% correct that a form ought to enclose all the needed input fields, and ideally not more than that.

Historically we did have the form covering a much narrower focus. But people often added more inputs, locating them higher/lower on the page, not considering where the form started/ended, and thus all their data wasn't getting submitted.

So we widened the scope and positioned it to cover a broader scope, allowing for incorporation of additional input fields with less complication. And of course also instructing those who posted having problems with it, to move the form start/end up/down where necessary to accommodate all the other differences unique to their template.

The proposed PR addresses only one of 5 product-types, and only one of 2 installed-by-default templates. And the other installed-by-default template requires the form start/end tags to be in different locations than proposed here.

I'm inclined to leave this as-is.

scottcwilson:
There are often inputs after that, like cross sells.   Let's leave this alone please.

drbyte:
I think cross-sell is just a grid of clickable things (like also-purchased), not form-fields. 

I can appreciate that something like BISN would require the form-close tag to be relocated. 
Either that, or it could use a javascript widget that takes care of itself independent of anything else on the page.

That said, I'm gonna leave it to plugin authors to work out how they change form elements, because each template is different in how/where it positions things on the page.

drbyte:
Another thing that could be done:  If BISN's form appears only when something's "not in stock" (ie: 'notify me when back in stock'), it could change/intercept the current form's destination action, in which case no changes are required to the form-close tag.

That change could be done via JS or within the core template.

PR #6663, opened 2024-08-16
fit form tags around inputs instead of entire file

torvista:


lat9:
Doesn't that cause the form-fields associated with attributes to be lost/unsubmitte?

torvista:
True. Fixed. Next...

drbyte:
Closing per discussion in #6204

PR #6673, opened 2024-08-18
Correct/reformat/refactor `Catalog :: Option Name Sorter`

### LINKED ISSUE #6671: `Catalog :: Option Name Sorter`: Empty dropdown when single-language admin
If a site's admin uses a single language, the language-selection dropdown for the subject tool is empty/unselectable:

image

That's because the `options_name.php` script determines its `$language_array` _before_ bringing in the admin's header script which _also_ creates that variable **only if** the store has more than one language!

lat9:
Fixes #6671 

Also
- Reformats to PSR-12
- Sanitizes input, ensuring it's set and an array
- Refactors repetitive code

### DIFF
--- a/admin/option_name.php
+++ b/admin/option_name.php
@@ -57,7 +57,7 @@ if ($action === 'update_sort_order') {
         <table class="table table-condensed table-striped">
           <thead>
             <tr class="dataTableHeadingRow">
-              <th class="dataTableHeadingContent text-center" colspan="2"><?php echo TEXT_EDIT_ALL; ?></th>
+              <th class="dataTableHeadingContent text-center" colspan="2"><?= TEXT_EDIT_ALL ?></th>
             </tr>
 <?php
 $languages_array = [];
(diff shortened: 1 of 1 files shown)


torvista:
As is
2024-08-18 12_43_22-Admin Option Name — Mozilla Firefox

Leveraging my advanced GUI design skills...

Personally I'd centre the id and sort

2024-08-18 12_44_09-Settings

Why show the language select if there is only one language?

Can the language-select and sort input be fitted to the content?

Why show the Flag column at all if the selected-language is already indicated above?

Why are the types in brackets?

lat9:
> 
> Why show the language select if there is only one language?
> 
> Can the language-select and sort input be fitted to the content?
> 
> Why show the Flag column at all if the selected-language is already indicated above?
> 
> Why are the types in brackets?

All valid comments, but since this PR is/was simply to get the existing behavior working properly and provide some code-structure refinements, those can certainly be submitted via a different PR/issue.

PR #6676, opened 2024-08-18
Update `order_status` page to use OPC-like version

### LINKED ISSUE #6586: Include COWOA/OPC update to `order_status` page?
Right now, the page simply redirects to `account_history`.

lat9:
Fixes #6586


lat9:
Bumping this, since if it's merged I've got follow-on work to do in both OPC and Bootstrap to special-case zc210 and later.

drbyte:
I've no objection.

PR #6679, opened 2024-08-19
Banner Manager: add position column

torvista:
Before:
BM default

After:
BM locations


drbyte:
Interesting idea. And interesting approach to accomplish it.

Perhaps it's a minor semantic point, but I prefer the word "Position" over "Location", because it relates to the position on the template/page.

drbyte:
Observation: If someone translates the Admin config settings to another language (note, open PR's regarding same), this "strip out a common prefix" approach will break.

drbyte:
Question: can the same end-goal be accomplished by simply (re-)naming the Banner Group (which is freeform text entered by the Admin) to explain the page-position, instead of describing the banner-dimension/orientation?

ie: `Wide-Banners` could be renamed to `Header Banners`, and `Sidebox-Banners` to `Sidebox 2`.

torvista:
The idea is to reduce the WTF feeling when I go into Banners every 5 years or so.
It's always seemed the inverse of "don't make me think", so I'm trying to make it more complete/understandable at first look.

If it prompts a re-think of what's underneath, others need to chip in, who are more users of this functionality....it should also be borne in mind multi-language banners, for which I have another "interesting" solution.

But one thing at a time.

torvista:
> Observation: If someone translates the Admin config settings to another language (note, open PR's regarding same), this "strip out a common prefix" approach will break.

Pushed a better effort, stripping anything before the hyphen.

PR #6682, opened 2024-08-22
Enable order-total/payment/shipping modules to be encapsulated

lat9:
Many, many commits so that the progress can be seen.


torvista:
So this enforces the use of lang. files/does not allow the older format:

2024-08-24 20_17_35-Admin Modules — Mozilla Firefox

I understand this is a generic message format, but it would be better to correctly name the expected file and create a debug...I just noticed Advanced shipper was not working and started poking about in the module code before noticing this message, as I would not have expected any message in this section.

I'm working on Advanced Shipper...not trivial.

This should be highlighted as a breaking change.

PR #6693, opened 2024-08-26
`pulldown` class: Align name with usage

scottcwilson:
All callers use lower case 'd' (`generatePulldownHtml`).

### DIFF
--- a/includes/classes/pulldown.php
+++ b/includes/classes/pulldown.php
@@ -182,7 +182,7 @@
          *
          * @return string
          */
-        public function generatePullDownHtml(string $name, string $parameters = '', bool $required = false)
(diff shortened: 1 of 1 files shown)


lat9:
That's weird, since the method name has been `generatePullDownHtml` since the class was initially created.

scottcwilson:
We could change the callers if you prefer but this was less change so I did it this way.   It just makes searching easier if the case is consistent.

lat9:
I'm just wondering why this "all of a sudden" became an issue.

scottcwilson:
Because I was doing a search for calls to this function and couldn't find them since the case was wrong.

PR #6696, opened 2024-08-26
Enable "both" encapsulated and legacy module language files

### LINKED ISSUE #6683: Encapsulated module update requires lang.-prefixed language files
So this enforces the use of lang. files/does not allow the older format:

2024-08-24 20_17_35-Admin Modules — Mozilla Firefox

I understand this is a generic message format, but it would be better to correctly name the expected file and create a debug...I just noticed Advanced shipper was not working and started poking about in the module code before noticing this message, as I would not have expected any message in this section.

I'm working on Advanced Shipper...not trivial.

This should be highlighted as a breaking change.

_Originally posted by @torvista in https://github.com/zencart/zencart/issues/6682#issuecomment-2308484125_

lat9:
Fixes #6683

### DIFF
--- a/admin/modules.php
+++ b/admin/modules.php
@@ -117,7 +117,7 @@ if (!empty($action)) {
             }
 
             if (in_array($class_file, array_keys($modules_found))) {
-                if ($languageLoader->loadCatalogLanguageFile($_SESSION['language'], $class_file, '/modules/' . $module_type)) {
+                if ($languageLoader->loadModuleLanguageFile($_SESSION['language'], $class_file, $module_type)) {
                     require DIR_FS_CATALOG . $modules_found[$class_file] . $class_file;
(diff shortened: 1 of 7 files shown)


scottcwilson:
Does this mean we should NOT make this statement: 

- Storefront: All shipping/payment/order-total modules *must* now have an associated [array based language file](/dev/languages/158_language_files/).  The use of the older `define` based language files is no longer supported for these modules.

lat9:
... although, any shipping/payment/order-total module included in a `zc_plugin` **must** use an array-based language file.

scottcwilson:
Ah!  I will add that detail, thank you.

scottcwilson:
You can't put new wine in old wineskins! :)

torvista (commented):
works for me.

PR #6697, opened 2024-08-27
EZ-Pages - add controls for links to show on a template's mobile-menu

drbyte:
Before this, the Responsive-Classic template would force all ezpages-header-enabled links into the mobile menu. 
But that meant a store had to use header links, and any footer links were ignored on mobile. 
This allows greater granularity over what links can be added to the mobile menu, via the admin UI.

Demo data updated to include a few defaulted to Mobile by default (see "Important Links" section):
Screen Shot 2024-08-26 at 10 11 13 PM

Column added to Admin EZ-Pages editor:
Screen Shot 2024-08-26 at 10 11 59 PM


drbyte:
For another template to integrate this, simply call `/includes/modules/ezpages_mobile.php` to generate an array of HREFs to output in whatever part of your template renders such links on mobile.

drbyte:
@lat9 I guess for BS4 template, the offcanvas template, instead of only querying the sidebox status/sort columns, could 
co-query the mobile fields (whose status defaults to 1 on install/upgrade).

lat9:
Noting that the BS4 implementation will lag since those settings aren't available until zc210 and the template currently supports zc158/zc200 installations as well.

PR #6698, opened 2024-08-27
Options Values Manager: Reformat/Refactor and correct search capability

### LINKED ISSUE #6694: Option Values search seems not to work
Admin > Catalog > Option Value Manager
Enter a search string, press return
Observe the page refreshes but the list is unchanged.

lat9:
Fixes #6694


scottcwilson:
Works great!

PR #6699, opened 2024-08-27
Add delivery vs billing postcode check

scottcwilson:
Increase robustness of mismatch check by adding postcode as a compared field.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -1346,7 +1346,7 @@ if ($show_orders_weights === true) {
                       ];
                       $search = zen_build_keyword_where_clause($keyword_search_fields, trim($keywords), true);
                   }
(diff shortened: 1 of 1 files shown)


admin/orders.php
```diff
@@ -1410,6 +1410,9 @@ function couponpopupWindow(url) {
                     if (!empty($orders->fields['delivery_street_address']) && (strtoupper($orders->fields['delivery_street_address']) !== strtoupper($orders->fields['billing_street_address']))) {
                       $show_difference = zen_icon('status-red', TEXT_BILLING_SHIPPING_MISMATCH) . '&nbsp;';
                     }
+                    if ((strtoupper($orders->fields['delivery_postcode']) != strtoupper($orders->fields['billing_postcode']) and trim($orders->fields['delivery_postcode']) != '')) {
```
lat9, line 1413: Use `!==` and `&&` instead of `and` in the conditional, please.

lat9 (changes requested):

lat9 (approved):

lat9:
Note that I'm being 'generous' about the additional parentheses and the use of loosey comparison for the trim.

lat9 (changes requested):

PR #6706, opened 2024-08-28
Admin language-loads, fallback to english

lat9:
Currently, the admin language-loading falls back to 'english' _only for_ the base language files.  This has (I think) been a problem since the language-arrays were introduced for multi-language sites.

This will 'protect' upgrading sites when new language-constants are defined in the ZC 'core' (e.g. `lang.ezpages.php` and `lang.layout_controller.php`) or when english-only encapsulated plugins are installed.

### DIFF
--- a/includes/classes/ResourceLoaders/AdminArraysLanguageLoader.php
+++ b/includes/classes/ResourceLoaders/AdminArraysLanguageLoader.php
@@ -5,50 +5,97 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: Zcwilt 2020 Jun 02 New in v1.5.8-alpha $
  */
-
 namespace Zencart\LanguageLoader;
 
 class AdminArraysLanguageLoader extends ArraysLanguageLoader
 {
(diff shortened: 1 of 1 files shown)


includes/classes/ResourceLoaders/AdminArraysLanguageLoader.php
```diff
@@ -5,50 +5,97 @@
...
         $this->addLanguageDefines($defineList);
+
+        if ($_SESSION['language'] !== $this->fallback) {
+            $defineList = $this->pluginLoadArraysFromDirectory($_SESSION['language'], '/extra_definitions');
+            $this->addLanguageDefines($defineList);
+        }
     }
 
-    protected function loadBaseLanguageFile()
+    protected function loadBaseLanguageFiles()
```
drbyte, line 58:
```suggestion
    protected function loadBaseLanguageFiles(): void
```
Or not?

drbyte: Oh, if it's extending from another class, then all need it...

PR #6712, opened 2024-08-29
Translate copied order-status names when installing new language

drbyte:
Expects admin `lang.languages.php` or a file in `extra_definitions` to contain order status defines for each pre-existing english status name. 
Any statuses for which a matching define is not found, will just use the English value, as before, and can be manually translated via the Admin UI as before.

French example for pending, processing, delivered, update:
```
'INSTALL_ORDER_STATUS_PENDING' => 'En attente',
'INSTALL_ORDER_STATUS_PROCESSING' => 'En cours',
'INSTALL_ORDER_STATUS_DELIVERED' => 'Livré',
'INSTALL_ORDER_STATUS_UPDATE' => 'Mise à jour',
```

Addendum to #6584

### DIFF
--- a/admin/languages.php
+++ b/admin/languages.php
@@ -147,7 +147,7 @@ if (!empty($action)) {
           $db->Execute("INSERT INTO " . TABLE_ORDERS_STATUS . " (orders_status_id, language_id, orders_status_name, sort_order)
                         VALUES (" . $status['orders_status_id'] . ",
                                 " . (int)$insert_id . ",
-                                '" . zen_db_input($status['orders_status_name']) . "',
+                                '" . zen_db_input(zen_lookup_admin_menu_language_override('install_order_status', $status['orders_status_name'], $status['orders_status_name'])) . "',
                                 " . $status['sort_order'] . ")");
         }
 
--- a/includes/functions/functions_lookups.php
+++ b/includes/functions/functions_lookups.php
@@ -318,6 +318,9 @@ function zen_lookup_admin_menu_language_override(string $lookup_type, string $lo
         case 'plugin_description':
             $lookup = strtoupper('ADMIN_PLUGIN_MANAGER_DESCRIPTION_FOR_' . $lookup_key);
             break;
+        case 'install_order_status':
+            $lookup = strtoupper('INSTALL_ORDER_STATUS_' . $lookup_key);
+            break;
     }
 
     if (isset($lookup) && defined($lookup)) {


lat9:
So, for a plugin, what would these value-override constants be named?

drbyte:
> So, for a plugin, what would these value-override constants be named?

In envision using a `INSTALL_` convention of some sort. Maybe `TRANSLATE_FOR_COPYING_` ... but that's really verbose.

And then the plugin can use similar code logic to what this PR does. (We can find someplace in core to put a similar function if we want.)

lat9:
If I've got a multi-language site, greater than 2 languages supported, how does this work?

drbyte:
> If I've got a multi-language site, greater than 2 languages supported, how does this work?

This particular feature does nothing "after" you've installed the languages.
But if you've put these constants in place "before" adding the language via the Admin, then when the Admin UI duplicates all the English language strings (in this PR it's addressing the Order Status records) into the newly-added language, it will lookup these constants for their defined value and use that defined value instead of the English value, when inserting those newly-duplicated records into the db for the new language.

If you've already installed the language to the Admin, then you've already inserted those language specific records into the db, and so the features of this PR have no effect at all.

piloujp:
Copied from PR 6584:

I did some tests with this code, placing at first the constants above in ../admin/includes/languages/french/lang.languages.php or a new file ../admin/includes/languages/french/extra_definitions/lang.install_orders_satus.php, and it did not work, it was like constants were not defined.
Problem was that when setting a new language (French), I was using English. Putting constants in one of the above file but in english folder fixed it.
But what then if there are 3 or more languages installed? It would be better if these languages constants could be loaded independently of language used.
Putting a file in admin/includes/extra_datafiles with below content allow using language install as intended from any language:

<?php
  die('Illegal Access');
}

define('INSTALL_ORDERS_STATUS_PENDING', 'En attente');
define('INSTALL_ORDERS_STATUS_PROCESSING', 'En cours');
define('INSTALL_ORDERS_STATUS_DELIVERED', 'Livré');
define('INSTALL_ORDERS_STATUS_UPDATE', 'Mise à jour');
?>

Last problem is if these constants need to be defined for 2 or more languages and to avoid constant double definition, language identifier should be in constant name like define('INSTALL_ORDERS_STATUS_PENDING_FR', 'En attente');.

lat9:
> I was able to make it work with few modifications: File in `admin/includes/extra_datafiles` has this content:
> 

Might I suggest instead that a file named something like `admin/includes/languages/english/extra_definitions/lang.install_order_status_fr.php`?

Placing the file in the `english` language sub-directory guarantees that it will be loaded on *all* admin pages and for *all* languages.

> 
> Like this, more than 2 languages can be used. The only thing I do not like is that constants are loaded when not necessary. Loading them only in language admin page would be more efficient.

I'll note that the memory consumed by the function that occasionally handles the language-specific look-up is most likely greater than that consumed by the language constants themselves!

piloujp:
> Questions:
> 
> a) The constants INSTALL_ORDER_STATUS_PENDING etc are new for 2.1.0, correct?
> 
> b) Prior to this, the table `orders_status` was populated with English strings when a new language was added in the admin, and the admin had to update the strings in Zen admin, correct?

Answer is yes to both questions.

lat9:
> > Might I suggest instead that a file named something like `admin/includes/languages/english/extra_definitions/lang.install_order_status_fr.php`?
> 
> I think it will work only if one file at a time is used/loaded. If all files are loaded and constants inside have same names, it's no good. Actually, I like the idea of loading/including only one file at a time, but there is more coding necessary. EDIT: There might be a misunderstanding here, when you suggested using `lang.install_order_status_fr.php` as filename I thought you mean also removing `_FR` from constant name. If it is not the case, then of course adding `_fr` to filename is not a problem and will allow to separate language files.
> 

What you said in your EDIT is what I meant.  The language file that ends in `_fr` would hold the `_FR`-suffixed language values.

> > Placing the file in the english language sub-directory guarantees that it will be loaded on all admin pages and for all languages.
> 
> Yes, but there is a case I thought of although did not test yet... If an admin user delete English language, what would happen?

If the "English" language is removed from the site's usage, the language fall-back is still `english` and the files in the `english` sub-directories will still be loaded.

piloujp:
OK, I moved files to `admin/english/extra_definitions` (and change defined back to array type) without any problem.
I although did some tests deleting English.
A problem appeared with order_status constant definition that uses `orders_status_name` which is an English name. When using another language to add back English, this name does not correspond anymore with constant name. The problem was resolved using `orders_status_id` instead.
Although, I added an English version language file to be able to put back English names in database.
Code in languages.php was simplified a little bit and adapted to POSM in observer function `class.products_options_stock_admin_observer.php`.
I now have something that works well in any case I could though of, both for orders status and POSM's 'back ordered' label.
As this PR is closed, perhaps I should make a new one to show the code, at least for orders status part? Please tell me what you prefer.

lat9:
An additional PR sounds like a good plan.

drbyte:
From the start, I didn't love the idea of re-using/extending `zen_lookup_admin_menu_language_override()` to be used for this "translate-only-upon-install", because it's different from directly overriding a thing for a current language, by using defines in real-time.

In the new PR, perhaps we can use a similar, but differently-named, function for this purpose?

piloujp:
OK, done. See PR #6713

PR #6713, opened 2024-08-30
Updating database with translated local strings when adding language

piloujp:
Added languages files and a new lookup function for translation of database local records when adding a new language.
Actually used to automatically update orders status strings and POSM back ordered label name when adding a new language.
The function uses language constants in files `admin\includes\languages\english\extra_definitions\lang.install_orders_satus_(language code here).php` and `admin\includes\languages\english\extra_definitions\lang.install_pos_name_(language code here).php`.

Language files for French are available in [French language pack on Github](https://github.com/piloujp/French_Language_Pack_for_ZC).

### DIFF
--- a/admin/includes/languages/english/extra_definitions/lang.install_orders_satus_en.php
+++ b/admin/includes/languages/english/extra_definitions/lang.install_orders_satus_en.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * @copyright Copyright 2003-2024 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: piloujp 2024 Aug 30 Modified in v2.1.0-alpha1 $
+*/
+
+$define = [
+    'INSTALL_ORDERS_STATUS_1_EN' => 'Pending', // Pending
+    'INSTALL_ORDERS_STATUS_2_EN' => 'Processing', // Processing
+    'INSTALL_ORDERS_STATUS_3_EN' => 'Delivered', // Delivered
+    'INSTALL_ORDERS_STATUS_4_EN' => 'Update', //Update
+];
+
+return $define;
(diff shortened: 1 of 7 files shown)


lat9:
A couple of things:
1. For a semblance of consistency, a plugin's language-override should (IMO) use its 'key' as part of the language file and constants' names.  For example `lang.install_posm_name_*`, since POSM is the plugin's key.
2. While you've handled the possibility of an additional language being installed _after_ POSM has been installed, the change doesn't cover the possibility of POSM being installed _after_ additional languages.

drbyte:
> the change doesn't cover the possibility of POSM being installed after additional languages.

That might be better as a separate PR.

There are a few ways it could be handled:
- Plugin provides a SQL patch that simply does the necessary "UPDATE" statements.
- Admin UI (perhaps Languages page?)  provides a "button" (language-specific, perhaps via a dropdown? or only affects currently-selected admin language?), which goes through orders-status and all plugins, looking for possible constants that could be used to replace db values. 
- Admin UI could report a list of situations where the db value matches the english (or default language) value?
- Admin UI could take such a report and offer a page full of input fields for someone to type in their translations, potentially showing the value of available constants that could be copied over?

drbyte:
But for 4 order-status records and 1 posm record ... **why is any of this necessary at all?** 

I'm considering reverting the PR I merged, as it feels like a whole lot of code that will rarely be used (I say rarely, not just because probably nobody will know about it to use, but also because "after" a language is installed, all those constants serve no purpose anyway.

drbyte:
@piloujp I appreciate the collaboration in this and multiple other PRs related to these language matters, and other topics.

In talking with @lat9 in more detail about this translate-during-language-install "feature", we've come to the conclusion that this is an unnecessary feature.
For less than a half-dozen strings to translate after installing a language, this is moot. And most language-translators aren't going to bother with this. AND end-users aren't going to understand what to even do with it. 
Anyone adding a language already can (and needs to) translate many database-stored strings by hand. It's not difficult or hard-to-find for them to do that for order-status and a plugin string.

I'm going to close this PR, and revert my last one on this topic.

Thanks again for your engagement and creativity with the various angles we've explored in relation to this topic. It is appreciated.

PR #6719, opened 2024-09-02
category tabs: minor rework

torvista:
Idea is to add mega menu functionality to this...

### DIFF
--- a/includes/modules/categories_tabs.php
+++ b/includes/modules/categories_tabs.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * categories_tabs.php module
  *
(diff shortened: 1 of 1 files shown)


torvista:
maybe better to use BS as reference for this

drbyte:
@torvista Are you building megamenu stuff bespoke? Or from an existing plugin?

torvista:
Still making up my mind. I cleaned up an existing plugin a while ago as a start, and forgot about it.  
Then changed tack to a flyout, that I could never get to work properly, then back to a mega. Going nowhere at a snails pace, wanting to stay as Bootstrappy as possible to contribute back, but it's not going well/at all. Design trial and error is so unproductive and I don't have the time or patience with the witchcraft of css rule-mountains.

PR #6721, opened 2024-09-02
[test] Experimenting with workflow updates

drbyte:


drbyte:
This was necessary because today Github Actions updated `upload-artifact@v3` with a breaking change that skips hidden files by default. And some of our tests were set to fail if no artifacts were found. (Previously was only uploading a hidden-by-coincidence .gitignore file, which is not a sensitive file, but was implicitly bypassing any such failures previously.)

PR #6722, opened 2024-09-02
Add mb_string polyfill

drbyte:
Since we already have it on the Laravel side, may as well make use of it elsewhere in core.


drbyte:
FYI - the `php_polyfills.php` updates are included here because they depend on the other polyfill (or the original extension) being present. It's not that their code needs extensive review (it comes from other packages). That said, if this PR is rejected, then those polyfills will have to be put in anyway, but their conditional wrappers changed to detect `extension_loaded('mb_string')` instead.

lat9:
Verified (on the admin-side anyway) via the Database I/O Manager (DbIo), which uses the mb-extensions if present on PHP 8.3.

Commented-out the `extension=mbstring` directive, saw that `mbstring` was disabled via **_Tools :: Server/Version Information_**.  DbIo loaded the multi-byte version of its string functions and exported a `products` table without error.

lat9 (approved):

drbyte:
I'll note that it's always more performant (and more reliable) to use the actual `mb_string` extension.
But this will at least allow us to leverage those functions even if a storeowner hasn't (or can't) install the extension.

PR #6725, opened 2024-09-04
Correct 'table' shipping module PHP Warnings when rates are based on a percentage

lat9:
As reported in [this](https://www.zen-cart.com/showthread.php?230325-2-01-log-file-question&p=1403269) Zen Cart forum posting.

### DIFF
--- a/includes/modules/shipping/table.php
+++ b/includes/modules/shipping/table.php
@@ -12,7 +12,6 @@
  */
 class table extends ZenShipping
 {
-
     /**
      * constructor
      *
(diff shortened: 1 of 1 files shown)


scottcwilson:
@lat9 Doesn't `zones` also need this fix?

scottcwilson:
Maybe low order fee too?

lat9:
I don't know; if you're seeing such an issue, please fix it.  I just corrected what was reported on the forums.

drbyte:
Done in b9ea5a8

PR #6728, opened 2024-09-04
admin, category/product listing, metatags on: change icon colour from yellow to green

torvista:
Yellow implies warning to me.

before
2024-09-04 20_12_26-Admin Category Product Listing — Mozilla Firefox

after
2024-09-04 20_14_38-Admin Category Product Listing — Mozilla Firefox


drbyte:
Perhaps changing the background instead would provide more contrast?

torvista:
The green is the same as the Enabled icon, and my thinking is that the green should **not** stand out, as green means "done"...not demanding attention.
Similarly I considered perhaps no-tags-defined should be red and some-tags-defined should be yellow, but that could be argued more as not all products may need tags at all, so I concluded to white as a neutral and green as something done, but certainly not yellow.

torvista:
To summarise the default indicators:

nothing defined: white star on black background

with extra meta title: no change/white star on black background

with meta keywords: yellow star on black background

with meta description: yellow star on black background

I would change that to 

nothing defined: red star on black background.

with meta keywords only: ignore, as keywords are pointless in 2025, so red star on black background

with extra meta title: yellow on black background.  
Implies something custom has been done for that product, so assume some attention has already been made to the title construction using the radio buttons.

with meta description: green star on black background. Possibly the only one that's actually important.

PR #6734, opened 2024-09-05
admin, product, collect-info: enforce a value for product name

torvista:


drbyte:
At first glance I agree: it seems sensible that a product's name never be allowed to be blank, whether accidental or intentional.

But what about in a multilanguage situation where the user isn't ready to translate it for every language?

torvista:
I can only imagine that, for a new product that is not going on sale yet...you could put a space.

I just submitted one by accident and thought it should have been trapped by the browser.

scottcwilson:
> But what about in a multilanguage situation where the user isn't ready to translate it for every language?

You'd put the English in as a placeholder.  It would be a nightmare to manage dozens of products with a blank name - it would take you way longer to figure out each one. 

I agree the field should be required.

PR #6736, opened 2024-09-05
implode 2nd argument requires array; string errors out

proseLA:
while i have not been able to reproduce, i have come across this error:
```
PHP Fatal error:  Uncaught TypeError: implode(): Argument #2 ($array) must be of type ?array, string given in /worsthost/worstadminrename/includes/modules/update_product.php:21
```
so if these `$_POST` vars are not set, we are passing a string.  which causes a FATAL error.

this can easily be reproduced by executing this 1 line script:
```php
// good
$test2 = implode(' ', $test ?? []);
//bad
$test2 = implode(' ', $test ?? '');
```


drbyte:
Good catch. Thanks.
Ya, those should fallback to empty array, not string.

(Those POST vars would be arrays because they support multi-language entry. 
Might not trigger if you didn't have a missing POST['manufacturer'] element.)

Issue #6543, opened 2024-06-16
problems with multi-lingual websites.


proseLA:
according to [the docs](https://docs.zen-cart.com/user/admin_pages/localization/languages/#:~:text=You%20can%20create%20a%20multi%2Dlanguage%20store%20by%20installing%20language%20packs%20and%20then%20adding%20Languages%20as%20described%20above.), one can create a multi-lingual website.

however, the current implementation of multi-lingual sites will create all kinds of SEO problems, as the implementation of the `hreflang` looks to be wrong.

right now, if one has a multi-lingual site, one can set the language by using a `$_GET` var which in turn sets a `SESSION` var so that from that point on, all pages are now in the chosen language.  for google, that is a problem.  depending on the `SESSION` var, ZC will display the chosen language.  which means the same URL could have different languages, ie mysite.com may be in english, french, italian, etc. depending on the `SESSION`.

in order to properly implement the `hreflang` tag; the tags must be bi-directional, and right now that is not the case.

for example, assuming a default language of `en`; one is at mysite.com; in the header there is an alternate tag for french which is mysite.com/?language=fr.  when one goes to that link, because tags need to be bi-directional, there should be a link back to mysite.com for english.  but of course there is not (there could be mysite.com/?language=en for english; but that does NOT satisfy the bi-directional requirement); when one does go back to mysite.com, it will now be in french, as ZC implements languages based on `SESSION` vars and not `GET` vars.

i believe that unless each language has its own unique url, zc will not be able to conform to successfully implementing the `hreflang` tag.

drbyte:
> there could be mysite.com/?language=en for english; but that does NOT satisfy the bi-directional requirement); when one does go back to mysite.com, it will now be in french, as ZC implements languages based on SESSION vars and not GET vars.

As long as the `GET` parameter is present, the site will render in the specified language, because the session will switch over as long as the GET parameter is part of the URL.

proseLA:
i am not saying the site will not render in the specified language.

what i am saying is that the same URL could be in many different languages, and from an SEO standpoint that is a problem.

there is plenty of good data on the web as to how to successfully implement the `hreflang` tag.  [this site](https://www.weglot.com/guides/hreflang-tag) for example.

the bottom line is that the language must be in the URL, and not in a `SESSION` var.

if we do not want to support the `hreflang` tag, i would totally understand that.  but right now it is in the code, and i still believe it is incorrectly implemented.

drbyte:
[...]

PR #6739, opened 2024-09-06
Extract `hreflang` logic to separate module, adding notifier

drbyte:
- relocated hreflang-generation logic to separate module file
- added notifier to allow bypass (ie: to intercept and output different format)
- added x-default "fallback" link, which simply points to the no-language-specified version of the page (ie: default language)

Retains bi-directional linking.
Retains former behavior of only outputting hreflang links if more than 1 language is installed.

Retains prior language-choosing via $_GET parameter provided in URL:
> (eg: If a visitor is directed to a page with `language=` in the URL, that page will display using that language, if still installed (and will also set the language in the session, for subsequent clicks to stay in that language). 
**Further**, if visitor "first" hits the x-default or canonical (eg: without a GET param or a session setting), their language will be auto-selected via the browser's language-default if resolvable, or presented in the store's default language.)

If one wanted to change the entire way that the URLs are generated (such as using subdomains-per-language), they could use the notifier hook and bypass this default operation in conjunction with any other notifier hooks that relate to URL generation.

Change in output:
```diff
    <link rel="alternate" hreflang="en" href="https://example.com/index.php?main_page=product_info&amp;products_id=34&amp;language=en"/>
    <link rel="alternate" hreflang="fr" href="https://example.com/index.php?main_page=product_info&amp;products_id=34&amp;language=fr"/>
+    <link rel="alternate" hreflang="x-default" href="https://example.com/index.php?main_page=product_info&amp;products_id=34"/>
```

Easily integrated into any existing template.

Related to #6543

### DIFF
--- a/includes/modules/hreflang.php
+++ b/includes/modules/hreflang.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * hreflang module
+ *
+ * @copyright Copyright 2003-2024 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: New in 2.1.0 $
+ *
+ * @var notifier $zco_notifier
+ */
+if (!defined('IS_ADMIN_FLAG')) {
+  die('Illegal Access');
+}
+// BOF hreflang for multilingual sites
+if (!isset($lng) || !$lng instanceof language) {
+    $lng = new language;
+}
+if (method_exists($lng, 'get_language_list')) {
+    $languages = $lng->get_language_list();
+} else {
+    // fallback for pre-v2.0.0 with old language class
+    $languages = array_keys($lng->catalog_languages);
+}
+
+$bypass = false;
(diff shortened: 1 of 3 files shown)


torvista:
Just caught me....
I tried this tool
https://technicalseo.com/tools/hreflang/
Before mod
hreflang PRE

After mod

post

drbyte:
Well, the "return link" exists, as you can see in the list of links.
They're just not treating the full URI (including GET params) as a URL, and thus calling it "not indexable".

Entirely functional for the actual website human user. Just not for bots.

Nothing we can do about that without using `.htaccess` rewrites to inject either subdirectories or subdomains, both for generating site-wide URLs and parsing them back out again.

drbyte:
This PR (intentionally) changes nothing in regard to URI formatting.
It does "add" the recommended `x-default` link.

More importantly, this at least adds a hook to intercept the default processing in conjunction with any URL-rewriter which might want to enhance this output.

PR #6744, opened 2024-09-08
Product-class: New notifier when requested product isn't found

lat9:
Needed for the Edit Orders update, in case a product in an order no longer exists on the site.

The `$data_override` variable, if updated, _must_ be an array type or a PHP Warning/Error will occur.

### DIFF
--- a/includes/classes/Product.php
+++ b/includes/classes/Product.php
@@ -225,7 +225,9 @@ class Product
         $product = $db->Execute($sql, 1, true, 900);
 
         if ($product->EOF) {
-            return [];
+            $data_override = [];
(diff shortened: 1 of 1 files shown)


drbyte:
I'm okay with this, but will ask: there's another notifier later near the end of the function, which allows further override. Doing this early-return skips that so it will never fire. Is that "material" in this situation?

lat9:
Yep, that 'later' notification is on my EO list as well ... for products that are still present in the site's products table.  This added one gives EO a chance to fill-in-the-blanks for an order being edited with a product that's subsequently been removed from the site.

drbyte (approved):

PR #6752, opened 2024-09-16
Change queryTime return type from int to float

webotron:


drbyte:
Thanks. Probably correct. Would you be willing to share how you arrived at the need for this? What was triggering errors about it?

proseLA:
i have just tested it and i'm a fan.

i have not been getting any errors, but on my test sites, i was getting 0 seconds for total time.  now, the data looks much more accurate:

```
Parse Time: 0.357 - Number of Queries: 196 - Query Time: 0.04349
```

we are using `microtime` when calculating.

drbyte:
Agreed. I'm gonna update queryFactory to obtain microtime AS a float in the first place.

drbyte:
But I still want to know what @webotron was doing to expose this ... so that related factors can be addressed as well.

PR #6753, opened 2024-09-17
Update Plugin Manager translations support

drbyte:
If a plugin contains language defines for
 `ADMIN_PLUGIN_MANAGER_NAME_FOR_xyz`
or
`ADMIN_PLUGIN_MANAGER_DESCRIPTION_FOR_xyz`
... then they will be used when displaying details of **installed and enabled** plugins in the Plugin Manager, 
according to whatever language is actively selected in the Admin interface.

### DIFF
--- a/admin/plugin_manager.php
+++ b/admin/plugin_manager.php
@@ -37,7 +37,13 @@ $tableDefinition = [
     'maxRowCount' => 20,
     'defaultRowAction' => '',
     'columns' => [
-        'name' => ['title' => TABLE_HEADING_NAME],
+        'name' => [
+            'title' => TABLE_HEADING_NAME,
+            'derivedItem' => [
+                'type' => 'local',
+                'method' => 'getLanguageTranslationForName'
+                ],
+            ],
         'unique_key' => ['title' => TABLE_HEADING_KEY],
         'filespace' => [
             'title' => TABLE_HEADING_FILE_SPACE,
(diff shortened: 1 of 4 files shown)


drbyte:
Example: MAP Pricing encapsulated plugin

French language defines within the plugin itself:
`zc_plugins/...../admin/includes/languages/french/extra_definitions/lang.map_pricing.php`:
```php
$define = [
    'ADMIN_PLUGIN_MANAGER_NAME_FOR_MAP_PRICING' => "PMA Prix Minimum Annoncé",
    'ADMIN_PLUGIN_MANAGER_DESCRIPTION_FOR_MAP_PRICING' => "Ce plugin permet à votre site de masquer les prix à moins que le client ne se connecte ou d'afficher uniquement des prix meilleurs que ceux annoncés après avoir ajouté un produit au panier.",
];
```

Screen Shot 2024-09-17 at 2 50 46 AM
(Pardon the lack of using a complete French language pack for all the UI components. It actually accentuates exactly which display components are affected by this PR. Naturally all translated strings would take effect in all the buttons and column headings and field descriptions, etc, if present.)

piloujp:
After few testing, I was only able to get translation when plugin is installed and active. Before installation, when deactivated or after uninstall I got only English.

drbyte:
Yes. That is correct.

Before the plugin is "installed" or "active", none of the language files for it are "loaded", so it is unaware of their content.

PR #6758, opened 2024-09-18
Various corrections to the admin's Customer-edit

### LINKED ISSUE #6754: Editing a customer in admin, no way to enter a state
If a customer's address starts out with a country, e.g. Canada, which includes zones and the customer's country changes to one that doesn't, there's no way to make that change in the admin since the state dropdown is "sticky":

On entry to the customer's edit:
image

Changing the customer's country to Australia makes no change to the dropdown states:
image

### LINKED ISSUE #6757: admin customer edit of address does not work
i am editing a customer address in the admin area.  i am changing a customer address from US to Canada.  both of these countries have state/province pulldowns.  

when clicking update, the country gets moved to afganistan (1st entry) and zc generates no error logs.

lat9:
Fixes #6757 
Fixes #6754

### DIFF
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -879,7 +879,7 @@ if ($action === 'edit' || $action === 'update') {
                     echo zen_draw_pull_down_menu(
                         'entry_country_id',
                         zen_get_countries_for_admin_pulldown(),
-                        $cInfo->country_id,
+                        $cInfo->entry_country_id,
                         'class="form-control" id="entry_country_id"'
                     ); ?>
                 </div>
--- a/admin/customers.php
+++ b/admin/customers.php
@@ -17,10 +17,8 @@ if (!isset($show_registration_ip_in_listing)) {
     $show_registration_ip_in_listing = false;
 }
 $action = $_GET['action'] ?? '';
-$customers_id = (int)($_GET['cID'] ?? 0);
-if (isset($_POST['cID'])) {
-    $customers_id = (int)$_POST['cID'];
-}
+$customers_id = (int)($_POST['cID'] ?? $_GET['cID'] ?? 0);
+
 if (!isset($_GET['page'])) {
     $_GET['page'] = '';
 }
@@ -82,6 +80,10 @@ if (!empty($action)) {
             $action = '';
             break;
         case 'update':
+            if ($customers_id === 0) {
+                zen_redirect(zen_href_link(FILENAME_CUSTOMERS, zen_get_all_get_params(['cID', 'action'])));
+            }
+
             $customers_firstname = zen_db_prepare_input(zen_sanitize_string($_POST['customers_firstname']));
             $customers_lastname = zen_db_prepare_input(zen_sanitize_string($_POST['customers_lastname']));
             $customers_email_address = zen_db_prepare_input($_POST['customers_email_address']);
@@ -394,6 +396,7 @@ if (!empty($action)) {
                 $cInfo->postcode = $cInfo->entry_postcode;
                 $cInfo->city =  $cInfo->entry_city;
                 $cInfo->state = $cInfo->entry_state;
+                $cInfo->customers_default_address_id = $default_address_id;
                 $processed = true;
             }
             break;
@@ -467,7 +470,11 @@ if (!empty($action)) {
             $customer->delete($delete_reviews, $forget_only);
             zen_redirect(zen_href_link(FILENAME_CUSTOMERS, zen_get_all_get_params(['cID', 'action']), 'NONSSL'));
             break;
-        default:
+        case 'edit':
+            if ($customers_id === 0) {
+                zen_redirect(zen_href_link(FILENAME_CUSTOMERS, zen_get_all_get_params(['cID', 'action'])));
+            }
+        default:    //- Fall-through from above if $customers_id appears to be valid.
             $customer = new Customer($customers_id);
             $cInfo = new objectInfo($customer->getData());
             break;
@@ -542,6 +549,7 @@ if ($action === 'edit' || $action === 'update') {
         true
     );
     echo zen_draw_hidden_field('default_address_id', $cInfo->customers_default_address_id);
+    echo zen_draw_hidden_field('cID', $customers_id);
     echo zen_hide_session_id();
 ?>
         <div class="row formAreaTitle"><?php echo CATEGORY_PERSONAL; ?></div>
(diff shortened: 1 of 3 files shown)


proseLA (commented):
given the fact that i used the following command:
```
git pull origin pull/6758/head:tempAdminCustomer
```
i can say that i am looking at **_ALL_** of the changes made in this PR.  i verified it by then looking at all of the commits:
```
e645babf8 (HEAD -> tempAdminCustomer, master) Finally, account for address record not updated ...
da5ff2043 Next, address #6754
e64a59e27 First, address #6757
0c52189b1 (upstream/master, origin/master) Merge pull request #6755 from scottcwilson/null_description
```
which would suggest that i am now looking at your code and not making some (adjective withheld) comments on some partial bit of your code.

this is an improvement, but in my limited testing i have still found a problem.

if i edit an address that goes from a country with a state/province pulldown to another country with a state/province pulldown, it seems to work fine.  in fact, as long as i can going to country with a state/province pulldown, all is good.

however, if i go from a country with a state/province pulldown, to a country without one, and i manually enter the state/province name, that field does not get saved.  i have to go back into the customer and reenter the state/province name.  then it gets saved.

this situation does not happen when going from country to country with no pulldowns.  

so i see 4 types of changes with regards to country address.  lets define a country with a pulldown as PULL; and a country without one as NONE.  and the situations are:

- from PULL to PULL => GOOD
- from NONE to NONE => GOOD
- from NONE to PULL => GOOD
- from PULL to NONE => BAD

in the bad situation, the state field does not get saved.

none of the other fields seem to be affected.

piloujp:
Problem is that when JavaScript updates form in browser, PHP variables are not updates, especially `$entry_zone_id` which is used in testing if a dropdown or text field is used...
In admin/customers.php, removing lines 269 to 271 (`$entry_state` reset to empty string) and changing test logic after line 321 fixed the problem above. See changes below:
```
                if (ACCOUNT_STATE === 'true') {
                    if ($entry_zone_id > 0 && $entry_state == '') {
                        $sql_data_array[] = [
                            'fieldName' => 'entry_zone_id',
                            'value' => $entry_zone_id,
                            'type' => 'integer'
                        ];
                    } else {
                        $sql_data_array[] = [
                            'fieldName' => 'entry_zone_id',
                            'value' => 0,
                            'type' => 'integer'
                        ];
                    }
                    $sql_data_array[] = [
                        'fieldName' => 'entry_state',
                        'value' => $entry_state,
                        'type' => 'string'
                    ];
                }
```
If you don't have already your own solution, please try it. I hope this will help.

lat9:
Thank you both for the review.  The underlying issue was that if a country with zones was selected _prior to_ one without zones, the `select` tag for the previous country's zones was still present (only hidden from display) so that non-zero `zone_id` was being transmitted on the form's submittal.

The jQuery update essentially toggles the `disabled` property of the `zone_id` select tag so that it's not sent with the form if the currently-selected country doesn't have zones.

piloujp:
> Thank you both for the review. The underlying issue was that if a country with zones was selected _prior to_ one without zones, the `select` tag for the previous country's zones was still present (only hidden from display) so that non-zero `zone_id` was being transmitted on the form's submittal.
> 
> The jQuery update essentially toggles the `disabled` property of the `zone_id` select tag so that it's not sent with the form if the currently-selected country doesn't have zones.

Good, you took it at the source. I tried to do the same, but my JavaScript skills are a little bit rusty and it did not work.

piloujp:
@lat9 Just two questions:
What about language constant `TYPE_BELOW`? It is not used anywhere.
Then would it not be better to have some highlight or message on state field when submitting the form without selecting a state from the dropdown list (for countries with zones)? Actually, page just reload as is and it looks like it just blinked.

lat9:
> @lat9 Just two questions: What about language constant `TYPE_BELOW`? It is not used anywhere. Then would it not be better to have some highlight or message on state field when submitting the form without selecting a state from the dropdown list (for countries with zones)? Actually, page just reload as is and it looks like it just blinked.

That constant's used by `zen_prepare_country_zones_pull_down`.  The jQuery update for the admin address-entry is just mimicking the storefront.

lat9:
> > however, i see a different problem; using my definitions from above going from NONE to NONE erases the state. i am not a fan of that approach. if someone selected the wrong country, why should they manually have to re-enter the state?
> 
> Good point. I can imagine, if a client entered the wrong country and you want to correct that quickly... If you did not memorize the state, you might even be in trouble and not able to re-enter it.

Point taken, another update forthcoming.

piloujp:
> That constant's used by `zen_prepare_country_zones_pull_down`. The jQuery update for the admin address-entry is just mimicking the storefront.

Sorry, I should have been more precise or direct in my question (I have probably lived too long in Japan). Constant `TYPE_BELOW` is already defined on line 30 of `lang.customers.php` and I was wondering if it was a mistake or if there was a good reason for that which escaped me.
My original question was a very Japanese way of getting your attention on a potential mistake without really mentioning the mistake... ;))

lat9:
> > That constant's used by `zen_prepare_country_zones_pull_down`. The jQuery update for the admin address-entry is just mimicking the storefront.
> 
> Sorry, I should have been more precise or direct in my question (I have probably lived too long in Japan). Constant `TYPE_BELOW` is already defined on line 30 of `lang.customers.php` and I was wondering if it was a mistake or if there was a good reason for that which escaped me. My original question was a very Japanese way of getting your attention on a potential mistake without really mentioning the mistake... ;))
> 
> I tested your last modification and everything work perfectly.

Duplicate removed, domo arigato.

lat9:
@proseLA, do you agree with @piloujp that this update is ready to go?

proseLA:
i can not comment on the code; but as far as the functionality, it definitely addresses the issue i raised in #6757, as well as the other issues i brought up in these comments.

proseLA:
as i previously stated, i did not look at the code; now that i am glancing at it, i think one should change the stamp on this file, as it is clearly a new file and the stamp is misleading:

https://github.com/zencart/zencart/blob/09351bbedffc5b1efef6315f7a768c0e1155e0ba/admin/includes/javascript/customers.php#L2-L7

drbyte:
@proseLA wrote:
>  i think one should change the stamp on this file, as it is clearly a new file and the stamp is misleading

Agreed. 

The version-stamping process which is run as part of building a release zip will take care of updating the @version tags of all edited files.

PR #6760, opened 2024-09-19
Admin's product-music observer doesn't load

lat9:
... since the name of the file doesn't confirm to the auto-loading observer's requirements.


drbyte:
I'm unable to replicate any problem here. Both naming conventions allow it to correctly load, and the notifier events "fire".

If I rename by removing the `auto.` prefix to intentionally force it to not load, and then edit a music product by changing its company name, it actually triggers a debug log because the observer isn't intercepting to change `manufacturers_id` to `record_company_id`. And the products_music_extra table doesn't get updated.

But when I put back the `auto.` prefix again, the debug log doesn't get created, and the record is updated.

Tested with both `auto.ProductMusicObserver.php` and `auto.product_music_observer.php`, and both work correctly.

So, unless I'm missing something, I don't think this change is needed.

lat9:
I must have been on another planet when I posted this.  I can also verify that the observer's additions to a `product_music` product's data-collection are there.

PR #6761, opened 2024-09-20
Delete zc_plugins/DisplayLogs/v3.0.1 directory

drbyte:
No need for the duplication in core.

Caveat: as with any zc_plugin whose files go missing, if a storeowner is still using 3.0.1 (most are using 3.0.2), and also deletes this v3.0.1 directory from their server, the module will (silently) cease to function until the storeowner clicks Upgrade in the Plugin Manager.


lat9:
The caveat should probably be added to the zc210 upgrade notes (_assuming_ that people will actually be reading them).

PR #6762, opened 2024-09-20
Encap plugin installer logic error?

drbyte:
@zcwilt

It would seem that it checks here for `$versionDir/installer/Installer.php` and then tries to load `$versionDir/Installer.php` Plus the use of `new Installer()` missing the `\` as in `new \Installer()` potentially clashes with the main `Installer` class, which has none of the methods on it that are expected by the code using this Factory to generate an "installer" object for.

So, I'm not convinced that this PR is even a proper "fix", but it does point out where there are mismatches that need attention.

What was the "Installer/Installer.php" fallback intended to be used for?  Is there a use-case for other than `BasePluginInstaller`, which `ScriptedPluginInstaller` uses?

### DIFF
--- a/includes/classes/PluginSupport/InstallerFactory.php
+++ b/includes/classes/PluginSupport/InstallerFactory.php
@@ -30,12 +30,14 @@ class InstallerFactory
         if (!file_exists($versionDir . 'manifest.php')) {
             throw new PluginInstallerException('NO VERSION MANIFEST');
         }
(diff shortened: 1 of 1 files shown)


drbyte:
No errors encountered. But don't have an example plugin using the naming convention therein. 

Just saw it while trying to follow the code in order to pass version details down into the installer class.

PR #6763, opened 2024-09-21
Include mod_list in core, as encapsulated plugin, named "System Inspection"

drbyte:
Screen Shot 2024-09-21 at 2 32 01 AM


dbltoe:
This has always been a valuable tool for me and has been on my wish list for core inclusion for some time.

I would like to propose changing line 44 of System Inspection/1.5.0/admin/system_inspection.php to `$pages_query_raw = " SELECT * FROM " . TABLE_ADMIN_PAGES . " ORDER BY menu_key ASC ";`

Adding that sort order results in the output following the order of the main menu items in the admin.  This makes it easier for the user to see which menu section the information applies to.  I find it much more useful than the current result.

drbyte:
Actually the admin menu is sorted first by `sort_order`, and then in "some" (configured by a define()) menus the entries are then sorted alphabetically (er, well, via a custom function). See `zen_get_admin_pages()` function for the more complex logic.

drbyte:
Given that `zen_get_menu_titles()` uses `ORDER BY sort_order`, I think I'll use a combination of `menu_key, sort_order`

PR #6769, opened 2024-09-23
Correct best-sellers' sidebox index

lat9:
... as identified in [this](https://www.zen-cart.com/showthread.php?223910-ZCA-Bootstrap-4-Template-Support-Thread&p=1403427#post1403427) forum-post and introduced in [this](https://github.com/zencart/zencart/pull/5599) PR.

Templates are _obviously_ expecting that the array returned has an index that starts with 1.

### DIFF
--- a/includes/templates/template_default/sideboxes/tpl_best_sellers.php
+++ b/includes/templates/template_default/sideboxes/tpl_best_sellers.php
@@ -7,12 +7,16 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: DrByte 2020 Jul 10 Modified in v1.5.8-alpha $
  */
-  $content = '';
-  $content .= '<div id="' . str_replace('_', '-', $box_id . 'Content') . '" class="sideBoxContent">' . "\n";
-  $content .= '<div class="wrapper">' . "\n" . '<ul class="list-links">' . "\n";
-  for ($i=1, $j=sizeof($bestsellers_list); $i<=$j; $i++) {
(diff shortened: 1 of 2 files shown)


lat9:
Yep, that portion of this multitude of errors has been around for a while; see [this](https://github.com/zencart/zencart/pull/1571) PR.

includes/templates/template_default/sideboxes/tpl_best_sellers.php
```diff
@@ -10,7 +10,7 @@
   $content = '';
   $content .= '<div id="' . str_replace('_', '-', $box_id . 'Content') . '" class="sideBoxContent">' . "\n";
   $content .= '<div class="wrapper">' . "\n" . '<ul class="list-links">' . "\n";
-  for ($i=1, $j=sizeof($bestsellers_list); $i<$j; $i++) {
+  for ($i=1, $j=sizeof($bestsellers_list); $i<=$j; $i++) {
```
drbyte: Changing the `for()` to `foreach()` would simplify all of this.  Simpler fix to the template, and no changes required on the module file.

lat9: Many templates 'in the wild', Bootstrap included, _presume_ that the ordering of that array starts with `1`.

drbyte: Okay, then we can keep the proposed module changes. 
But this can still be changed to a foreach moving forward, and is probably the better advice to give to users, because it's the most-compatible "fix", both backward and forward.

PR #6773, opened 2024-09-24
Add `InteractsWithPlugins` trait for Observers in Encapsulated Plugins

drbyte:
This will allow Observers in Encapsulated Plugins to query details about the plugin version/directory in which it is located. Helps with:
- knowing what version is in effect
- knowing what context we're in (admin/catalog/Installer)
- loading a catalog stylesheet via a plugin, first from plugin directory, and then from active template directory (as an override), without having to publish files to the core catalog directory.

To use:
```diff
<?php
+use Zencart\Traits\InteractsWithPlugins;
use Zencart\Traits\NotifierManager;
use Zencart\Traits\ObserverManager;

class MyEncapsulatedPluginObserverClass
{
+    use InteractsWithPlugins;
    use NotifierManager;
    use ObserverManager;
```

and then that observer class will have access to:
```php
    protected bool $isAZcPlugin;
    protected string $zcPluginDirName;
    protected string $zcPluginVersionDir;
    protected string $zcPluginPath;

    /** @var string catalog, admin, or Installer */
    protected string $zcPluginContext;

    /** @var string working directory of currently installed version */
    protected string $pluginManagerInstalledVersionDirectory;

    /** @var string will be null if no 'catalog' dir present (no catalog features) */
    protected string $zcPluginCatalogPath;
    /** @var string will be null if no 'admin' dir present (no admin features) */
    protected string $zcPluginAdminPath;
    /** @var string will be null if no 'Installer' dir present (should never be) */
    protected string $zcPluginInstallerPath;
```

and 
```php
    protected function linkCatalogStylesheet(string $stylesheet_filename, ?string $current_page): bool
```

### DIFF
--- a/includes/classes/ResourceLoaders/PageLoader.php
+++ b/includes/classes/ResourceLoaders/PageLoader.php
@@ -93,9 +93,13 @@ class PageLoader
         return $path;
     }
 
-    public function getTemplatePluginDir($templateCode, $templateDir)
+    public function getTemplatePluginDir(string $templateCode, string $templateDir, ?string $whichPlugin = '')
     {
         foreach ($this->installedPlugins as $plugin) {
+            if (!empty($whichPlugin) && $plugin['unique_key'] !== $whichPlugin) {
+                continue;
+            }
+
             $checkDir = 'zc_plugins/' . $plugin['unique_key'] . '/' . $plugin['version'] . '/catalog/includes/templates/default/' . $templateDir . '/';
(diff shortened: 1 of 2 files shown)


drbyte:
> So, you're proposing to extend an encapsulated plugin's catalog/includes/templates directory to contain a 'base' /css subdirectory (catalog/includes/templates/css)? Why not use a plugin's templates/**default**/css so that the plugin's directory structure is in-line with the base directory?

While I do strongly feel that the "`default`" subdir in the middle is significantly redundant, serving no purpose, I see that we've documented that as a "standard", so I've updated it to use the shared function which does that.

PR #6774, opened 2024-09-24
Update to admin/modules.php to use admin lookup function for translation

piloujp:
Update to admin/modules.php to use the zen_lookup_admin_menu_language_override function for translation using constants placed in admin's lang.modules.php.
Complements PR #6584

### DIFF
--- a/admin/modules.php
+++ b/admin/modules.php
@@ -394,7 +394,7 @@ switch ($action) {
             if (ADMIN_CONFIGURATION_KEY_ON === '1') {
                 $displayKey = 'Key: ' . $key . '<br>';
             }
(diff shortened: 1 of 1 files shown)


drbyte:
I haven't had a chance to test whether the following is also necessary: There are 2 places where `->description` and 5 where `->title` are used without apparent translation. Do these need addressing also? Or are they already translated correctly?

piloujp:
I checked both admin display where everything was translated and in code too. The `$module->title` (and description) and `$mInfo->title` (and description) use the constants defined in language files.
They are already translated and don't need anything more to add.

PR #6781, opened 2024-09-27
Check for null entity

scottcwilson:
Seeing more of these.  Happens when people go directly to the URL 

`index.php?main_page=ask_a_question&action=send&pid=NNN`

This prevents the log; the check done later will bounce them back to the AAQ page.


drbyte:
You only addressed the one line that's calling strip_tags(), so I presume the logs were saying that `strip_tags()` was complaining about a null?  
(And not complaining that this and the other POST values were null when being passed to zen_db_prepare_input()?)

scottcwilson:
Correct - the log was from strip tags.

PR #6783, opened 2024-09-28
Add notifications for OSH insert in account_history_info

lat9:
... think "Ty Package Tracker".

Includes some minor alignment changes so that I could 'see' what I was dealing with!

### DIFF
--- a/includes/templates/template_default/templates/tpl_account_history_info_default.php
+++ b/includes/templates/template_default/templates/tpl_account_history_info_default.php
@@ -137,7 +137,7 @@ if (!empty($order->statuses)) {
         <th scope="col" id="myAccountStatus"><?php echo TABLE_HEADING_STATUS_ORDER_STATUS; ?></th>
 <?php
     $extra_headings = [];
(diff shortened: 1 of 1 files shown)


includes/templates/template_default/templates/tpl_account_history_info_default.php
```diff
@@ -130,39 +130,61 @@
...
 ?>
-
 <h2 id="orderHistoryStatus"><?php echo HEADING_ORDER_HISTORY; ?></h2>
 <table id="myAccountOrdersStatus">
     <tr class="tableHeading">
         <th scope="col" id="myAccountStatusDate"><?php echo TABLE_HEADING_STATUS_DATE; ?></th>
         <th scope="col" id="myAccountStatus"><?php echo TABLE_HEADING_STATUS_ORDER_STATUS; ?></th>
+<?php
+    $extra_headings = [];
+    $zco_notifier->notify('NOTIFY_ACCOUNT_HISTORY_INFO_OSH_HEADINGS', ['order' => $order], $extra_headings);
```
drbyte, line 140: Instead of passing an array of `['order' => $order]` why not just pass `$order` directly? Easier to use directly without having to de-reference an array on the receiving end.

PR #6785, opened 2024-09-28
zc_install: block proceeding if cannot convert db default to uft8mb4

### LINKED ISSUE #6664: `zc_install` can result in "mixed collations"
1. Performing an initial install, I created an empty database using `utf8_general_ci` collation. The `zc_install` process completed successfully, but all tables/table-fields were created with `utf8mb4_general_ci` collation.
2. Performing an upgrade from a zc139h database (no 'DB_CHARSET' in the original `configure.php` files, since that was introduced in zc150).  The database's base collation is `latin_general_ci` and all tables added by the `zc_install` process use `utf8mb4_general_ci` collation.  The 'DB_CHARSET' is successfully added to the updated `configure.php` files, with a value of 'utf8mb4'.

This will lead to potential collation-mismatch errors in the future, since additional tables added will inherit the database's default collation, which is not necessary reflective of the 'DB_CHARSET' value.

drbyte:
- `queryFactory` sets 2 publicly-queryable properties to note the detected **default* db charset and collation
- `zc_install system inspection` checks whether those detected defaults are of the `utf8mb4` family, and if not, then attempts to set the database's defaults, inspired from current collation; if that fails, it halts install/upgrade.
- `sniffer` class adds 2 convenience helpers to query collation on tables and columns

Attempts to resolve #6664


lat9:
At which point is the zc_install supposed to report?  I've got a very foobared zc154 database whose default collation is latin1_swedish_ci and an assortment of latin1-collated tables.

The zc_install allowed the upgrade to proceed until it came to a screeching halt at the 155 to 156 upgrade (for some reason).  No mention was made through the process about the errant collation.

lat9:
I've also got a base install of zc155 (using `latin1_general_ci` throughout).  No collation message on the zc_install upgrade.

When the upgrade was completed, the newly-added tables (e.g. `plugin_groups`) have a `latin1_general_ci` collation, but the database's overall collation has been changed to `utf8mb4_general_ci`.

drbyte:
Hmm. This does its inspection right up front when it checks all the prerequisites and before it detects the ZC version.

drbyte:
> the database's root collation has been changed to utf8mb4_swedish_ci
> database's overall collation has been changed to utf8mb4_general_ci

Okay, those are expected: it does attempt to set the database's default, if it can, up front.

But that's also supposed to affect the new tables. I will double check the latin1 detection...

drbyte:
This now has enough error messaging and blocks in place that it should prevent users from getting into a bad state.

They'll probably wish that the db-converter was built-in though.

drbyte:
Screenshots added to OP.

torvista:
>They'll probably wish that the db-converter was built-in though.

Could that not be done (making a backup and converting)?...it's another inevitable step no?

lat9:
Looking good for various databases that 'identify' as `latin1`, but I've found an issue with a 'utf8' installation.

All collations in the database are `utf8_unicode_ci` and the configure.php files report DB_CHARSET as 'utf8' and DB_PREFIX as an empty string.  When attempting an upgrade, I receive the following messages and the overall db-collation has been changed to `utf8mb4_general_ci` after running the zc_install process.

1. There is no prefix in use in any of the database's table definitions.
2. There is no usage of a `latin1_` collation type in the database.

image

lat9 (changes requested):
See base comment posted

PR #6792, opened 2024-10-01
Add $params to zen_back_link() and add zen_datetime_without_seconds()

zenexpert:


drbyte:
Can you explain why you named this function `..._frontend()`?

I can see the value in passing classes/etc to the backlink. Are you wanting to do that with any built-in backlinks? Or more in added/custom code?

zenexpert:
Because I'm terrible at naming functions? Dunno, the idea was to use it in frontend  so it's just the first thing I thought of...

Backlinks - it's just something I noticed because I wanted to style them on all inner pages (ie. conditions, shippinginfo etc). I know you can always do it with `zen_back_link(true)` but that's just ugly...

drbyte:
> naming functions

Haha, ya, naming is hard!  

No worries.

How about the following. Does it work the same on your end?

```php
/**
 * Return date and time without seconds (ie. 2024/10/01 9:54)
 */
function zen_datetime_without_seconds (string $raw_datetime): string
{
    if (empty($raw_datetime) || $raw_datetime <= '0001-01-01 00:00:00') return false;

    $year = (int)substr($raw_datetime, 0, 4);
    $month = (int)substr($raw_datetime, 5, 2);
    $day = (int)substr($raw_datetime, 8, 2);
    $hour = (int)substr($raw_datetime, 11, 2);
    $minute = (int)substr($raw_datetime, 14, 2);

    zen_define_default('DATE_TIME_FORMAT_WITHOUT_SECONDS', '%m/%d/%Y %H:%M');

    global $zcDate;
    return $zcDate->output(DATE_TIME_FORMAT_WITHOUT_SECONDS, mktime($hour, $minute, 0, $month, $day, $year));
}
```
And that would allow the locale to still be controlled via the language file, by adding a new definition:
```php
    'DATE_TIME_FORMAT_WITHOUT_SECONDS' => '%m/%d/%Y %H:%M',
```

zenexpert:
Works like a charm!
I was trying to avoid the new definition (I hate them when doing upgrades) - it's just the date formatting that differs around the world, but time is pretty much always the same so I figured an existing definition for date would be enough. I do agree, this is a more streamlined solution and makes sense to have it separate.

End result with `zen_datetime_without_seconds($statuses['date_added'])`:


UK date formatting so - jolly good, sir!

PR #6793, opened 2024-10-02
Changing admin image display to size via CSS instead of HTML attribute

retched:
This change allows the image to be resized dynamically via CSS instead of hardcoded via HTML  attributes.

For example: the pictures in the category listing would be stretched to fit both the height and width of the image settings as designated in the admin backend. This change moves the setting of the image's height and width to CSS where appropriate (ie. where the image's height or width are set in the function call).

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -694,7 +694,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                       <a href="<?= zen_catalog_href_link('index', zen_get_path($category['categories_id'])) ?>" rel="noopener" target="_blank" title="<?= BOX_HEADING_CATALOG ?>"><?= zen_icon('popup', BOX_HEADING_CATALOG, '') ?></a>
                   <a href="<?= zen_href_link(FILENAME_CATEGORY_PRODUCT_LISTING, zen_get_path($category['categories_id'])) ?>" class="folder"><i class="fa-solid fa-lg fa-folder"></i>&nbsp;<strong><?= $category['categories_name'] ?></strong></a>
                 </td>
-                  <td class="hidden-sm hidden-xs imageView <?= $additionalClass ?>"><?= zen_image(DIR_WS_CATALOG_IMAGES . $category['categories_image'], $category['categories_name'], IMAGE_SHOPPING_CART_WIDTH, IMAGE_SHOPPING_CART_HEIGHT) ?></td>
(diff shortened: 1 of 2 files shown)


drbyte:
I see your point, and like it. 
One problem I see caused by this change is the loss of the ability to continue using a single-pixel gif to draw a divider line, which is still in use around a lot of the admin pages (due for conversion to an HR or some other divider at some point, probably in the Spring for v3.0)

Before:
before

After:
after

retched:
Dang it! I was on the lookout for that thing!

The only other thing I could think of would be to create a new function solely for the listing grid (call it like `zen_product_image` or something) and change the category_listing calls for images to that new one. But I feel there is a better way to go about it.

drbyte:
You had used `style="width="22px;""`, where the `=` should have been `:`, and the extra quotes were wrong too.

I reworked it to sus out whether `style=` had already been passed, and to conditionally add height/width where not already present, and to account for already being set as `%` or not.

This appears to work harmoniously with `zen_black_line()` and `zen_draw_separator()`.

drbyte:
But does that let the proportional stuff work that you were after initially?

retched:
> But does that let the proportional stuff work that you were after initially?

Close to the desired effect. Added in a `, 'style="object-fit: contain;"'` to the call params from category_product_listing.php file. This way it forces the listing to use the proportional fit.

These are the results:

image

lat9:
> You had used `style="width="22px;""`, where the `=` should have been `:`, and the extra quotes were wrong too.
> 
> I reworked it to sus out whether `style=` had already been passed, and to conditionally add height/width where not already present, and to account for already being set as `%` or not.
> 
> This appears to work harmoniously with `zen_black_line()` and `zen_draw_separator()`.

Don't the `width=` and `height=` just want to be numbers?  That is, no `px` added.

retched:
> > You had used `style="width="22px;""`, where the `=` should have been `:`, and the extra quotes were wrong too.
> > I reworked it to sus out whether `style=` had already been passed, and to conditionally add height/width where not already present, and to account for already being set as `%` or not.
> > This appears to work harmoniously with `zen_black_line()` and `zen_draw_separator()`.
> 
> Don't the `width=` and `height=` just want to be numbers? That is, no `px` added.

Effectively yes. The numbers used in `category_display` are the `SHOPPING_CART_HEIGHT` and `SHOPPING_CART_WIDTH` defines. Those are normally used in HTML attributes of the image tag and only accept the number of pixels and not percentages.

drbyte:
> Don't the `width=` and `height=` just want to be numbers?  That is, no `px` added.

Yes. When used as ``. But this PR is dropping those and replacing them with CSS syntax via a `style="width: 20px; height=40px;"` attribute. CSS syntax wants the units, because they could be `px` or `%` or `rem` etc.

PR #6794, opened 2024-10-02
remove 0 count categories from `zen_get_categories()` when requesting active categories

proseLA:
if the request is made to only include active categories, it seems appropriate to NOT include categories where the count of items in the category is 0.

### DIFF
--- a/includes/functions/functions_categories.php
+++ b/includes/functions/functions_categories.php
@@ -160,6 +160,12 @@ function zen_get_categories($categories_array = array(), $parent_id = TOPMOST_CA
     $results = $db->Execute($categories_query);
 
     foreach ($results as $result) {
+        if ($status_flag !== null) {
+            $count = zen_products_in_category_count($result['categories_id']);
+            if ($count === 0) {
+                continue;
+            }
+        }
         $categories_array[] = [
             'id' => $result['categories_id'],
             'text' => $indent . $result['categories_name'],


scottcwilson:
Empty and disabled are different things though.

drbyte:
Is there any reason for the subject function `zen_get_categories()` to be called outside of `tpl_search_default.php`?

In "core", that's the only place that calls it. And if it's only ever used in the context of "search", one "could" argue that empty categories don't need to be in the dropdown for search criteria. Although some (including the storeowner) may be confused when empty categories don't show up in the list and they don't understand why.

Granted, then ideally the function name should reflect that context for clarity.

scottcwilson:
What if we add a parameter to indicate show/don't show empty categories?
I'm thinking a plugin may want to use a category, even if it's empty.

drbyte:
Sure. `bool $exclude_empty = true` would require no changes to existing templates.

proseLA:
i will not be in front of my computer until next week but i did want to briefly respond to the comments. @scottcwilson i do not really understand your first comment. this PR is about active and empty categories. how can an empty category truly be considered active?  with regards to your 2nd comment, a plugin could always call all categories if it wanted to see them all. @drbyte i can see the use of the method outside of the search function. i know on a bootstrap site i am doing, the active and non empty categories are the only ones listed. i think that was accomplished by doing its own query of the database. that said, i still can not see when you would want an empty category when calling a method to get only active categories. is an empty category active? i think that is the question.

proseLA:
i have no problem with the adding of the parameter.

torvista:
I would have thought customers have no interest in empty categories, but an admin may...so a parameter seems appropriate.

torvista:
I commented on this but I now find that the comment and reply are not displayed (with console errors) when logged in:

https://github.com/zencart/zencart/commit/30ea100193d9d273a91da6ea6ad721972876bff8#commitcomment-147974044

anyone else see that problem?

PR #6795, opened 2024-10-03
Update collect_info.php fixed image size

shot-guun:
Fixed product image size stops big images when editing product


barco57:
I don't particularly like that declaring both small height and width messes with proportion of the displayed image. declaring just width works for me but I used MEDIUM_IMAGE_WIDTH so it displayed at the same size as it is going to be displayed on the front end product page

shot-guun:
I see what you mean. I use a carousel on products page which shows small images until clicked but I do agree with keeping proportions, thanks.

scottcwilson:
@shot-guun Would you please attach images showing before and after your change for documentation?  Thank you and thanks for doing this.

drbyte:
Something to keep in mind:
- the "main" product image is also the one used for the thumbnail, even if the image itself is higher resolution
- the "small" dimensions are the thumbnail size, and are best applied to the "main"/thumbnail image
- if using MEDIUM, then you "could" still display the "main" image, but you'd be scaling it up to medium dimensions, unless you also tell it to render the _MED image.
- and for LARGE, the same scenario if you don't render the _LRG image.

Summary: if you've actually optimized your images for proper display at each resolution, then scaling up the main image to "medium" is likely gonna look a bit grainy.

Isn't the "real" problem that you're getting bizarre image sizes "because" you've not actually created images for each resolution in the first place?

barco57:
"Isn't the "real" problem that you're getting bizarre image sizes "because" you've not actually created images for each resolution in the first place?" - who does this anymore? I haven't seen it in years.... image handler does the image resizing.

torvista:
> you've not actually created images for each resolution in the first place?" - **who does this anymore**?

Good question. 

https://github.com/zencart/zencart/discussions/6380

admin/includes/modules/collect_info.php
```diff
@@ -469,7 +469,7 @@
     if (!empty($pInfo->products_image)) { ?>
         <div class="form-group">
             <div class="col-sm-offset-3 col-sm-9 col-md-6">
-                <?php echo zen_info_image($pInfo->products_image, (is_array($pInfo->products_name) ? $pInfo->products_name[$_SESSION['languages_id']] : $pInfo->products_name)); ?>
+                <?php echo zen_info_image($pInfo->products_image, (is_array($pInfo->products_name) ? $pInfo->products_name[$_SESSION['languages_id']] : $pInfo->products_name),SMALL_IMAGE_WIDTH,SMALL_IMAGE_HEIGHT); ?>
```
drbyte, line 472:
```suggestion
                <?php echo zen_info_image($pInfo->products_image, (is_array($pInfo->products_name) ? $pInfo->products_name[$_SESSION['languages_id']] : $pInfo->products_name), MEDIUM_IMAGE_WIDTH); ?>
```

drbyte (changes requested):
Let's go with medium.

drbyte:
I can't edit the PR or contribute to it due to blocks set against it in your branch. 
I'll just merge and edit it afterward.

PR #6797, opened 2024-10-03
Fatal error when paying credits | Update payment.php

asiksami:
In latest `v2.1.0-beta1` with `PHP 8.3.12 (Zend: 4.3.12)`, getting fatal error when paying with credits.

```
PHP Fatal error:  Uncaught TypeError: Cannot assign string to property payment::$modules of type array
```

`Payment`-class property `modules` is initiated in the class as an `public array`, but in the function `checkCreditCovered` the property is "nulled" to as an empty string.


lat9:
Thanks!

PR #6803, opened 2024-10-07
admin, orders, paypal extra admin: remove hard-coded text

torvista:
2024-10-07 20_19_14-Admin Orders — Mozilla Firefox

Initially I noticed it was text in the input field instead of a placeholder, so **typing into the field did not remove the text.**
Then I found it was hard-coded in english.
Then after seeing the already-excessive amount of constants of fragmented/un-sprintf-ed texts already defined, and the already-existing explanatory text, I concluded it doesn't need any placeholder at all.


torvista:
Could argue the same here:
2024-10-07 20_54_44-Admin Orders — Mozilla Firefox
(and the field is too small)

torvista:
Bump.

PR #6811, opened 2024-10-13
Admin, copy_product_confirm: notifier to prevent redirect

torvista:
I'm on a plugin that copies multiple products in one operation, so the redirect after each copy needs to be prevented, as this file is called in a foreach.
Maybe there's a better way...

### DIFF
--- a/admin/includes/modules/copy_product_confirm.php
+++ b/admin/includes/modules/copy_product_confirm.php
@@ -245,8 +245,15 @@ if (isset($_POST['products_id'], $_POST['categories_id'])) {
     // reset products_price_sorter for searches etc.
     zen_update_products_price_sorter($products_id);
 }
-if ($_POST['copy_as'] === 'duplicate' && !empty($_POST['edit_duplicate'])) {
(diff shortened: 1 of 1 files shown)


drbyte:
> Maybe there's a better way...

It raises the point that all of these modules do redirects.

I guess the redirects might be better relocated into `category_product_listing.php` immediately following the call to each module. 
Not sure if that would break any/many plugins that have been making calls directly to these modules and would therefore need to add the redirect themselves. It's also possible that plugins have cloned these modules solely to exclude the redirect, and would be grateful for the relocation of it so the plugin's distro could be simpler.

torvista:
My interest aside, although the product types all allow for cloning/overriding the core/base modules, anything that can reduce the necessity of extra files sounds like progress...avoids the multiple comparisons when a core file is updated.

drbyte:
Need to assess the ripple-effect of relocating the `zen_redirect()` calls into `category_product_listing.php`.

If a plugin has modified that file and a user doesn't merge the change, the no redirects would happen at all, meaning not only would the page not reload as usually intended, but no status update messages will display either, since the messageStack won't fire(er, well, it'll display on the "next" page load, which will be confusing).

torvista:
Just another breaking change to document in blink tags.

PR #6818, opened 2024-10-16
Fix checkAll() button on profile-edit page

drbyte:
Multiple CSS classes are now applied to checkbox elements, so we can't check for `element.className == val`; instead, must use `element.classList.contains(val)`


scottcwilson:
Broken in 1.5.8, 2.0.0, 2.0.1.  I guess people don't use this feature very much - but thanks for fixing!

PR #6821, opened 2024-10-16
Retire obsolete rand() in favor of newer methods

drbyte:
`rand()` was retired in PHP 7.2, being made an alias of `mt_rand()`, which has since been superceded by `random_int()` when specifying a range. mt_rand() is still valid for the ways in which we use it.

### DIFF
--- a/includes/classes/Coupon.php
+++ b/includes/classes/Coupon.php
@@ -198,11 +198,10 @@ class Coupon extends base
             return '';
         }
         $random_string = bin2hex(random_bytes(128));
-        mt_srand((int)microtime() * 1000000); // seed the random number generator
         $good_result = 0;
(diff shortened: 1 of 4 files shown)


includes/classes/Coupon.php
```diff
@@ -202,7 +202,7 @@ public static function generateRandomCouponCode(string $salt = "secret", $length
         $good_result = 0;
         $new_code = '';
         while ($good_result === 0) {
-            $random_start = @rand(0, (128 - $length));
+            $random_start = @random_int(0, (128 - $length));
```
lat9, line 205: Why continuing with the error-suppression?

drbyte: Removed.

PR #6825, opened 2024-10-17
Enable zc_plugins to provide extra-cart-actions

### LINKED ISSUE #6824: Enable zc_plugins to provide `extra_cart_actions`
... as identified/requested in [this](https://www.zen-cart.com/showthread.php?230387-Extra-Cart) Zen Cart forum posting.

lat9:
Fixes #6824.

### DIFF
--- a/includes/main_cart_actions.php
+++ b/includes/main_cart_actions.php
@@ -21,33 +21,24 @@ if (!defined('IS_ADMIN_FLAG')) {
  */
 
 /**
- * Gather all PHP files present in the extra_cart_actions subdirectory.
+ * Load all PHP files present in the extra_cart_actions subdirectory.
  */
-$baseDir = DIR_WS_INCLUDES . 'extra_cart_actions/';
-$main_cart_actions = [];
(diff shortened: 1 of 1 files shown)


drbyte:
>  * Gather all PHP files present in enabled zc_plugins, noting that any like-named file
> * in one of these directories **overrides** any file present in the 'base' directory!

Do we really want to override (completely ignore) both core and other plugins' versions of (these) files, based on filename clash?

lat9:
> > * Gather all PHP files present in enabled zc_plugins, noting that any like-named file
> > * in one of these directories **overrides** any file present in the 'base' directory!
> 
> Do we really want to override (completely ignore) both core and other plugins' versions of (these) files, based on filename clash?

Just emulating how the order-total/shipping/payment modules overrides work ... for consistency. I've got no skin in the game and would be happy to just load _everyone_, if that's what's desired.

drbyte:
(Edit: we don't distribute any "extra" cart actions in "core", so my mention of core was redundant.)

Yes, I think we should load all. While this extra-cart-actions feature is seldom-used, we don't want to stomp on someone's inclusion because another plugin used the same filename.

PR #6831, opened 2024-10-18
Admin product listing: include products which are missing a description record

### LINKED ISSUE #6828: Missing product description record handling
I have seen databases in the wild that have had a record for a product in the `products` table but not the `products_description` table. 

The behavior produced by this situation in the latest Github is the following: 

- the product can be seen in the catalog (for example on the listing page)
- the product *cannot* be seen in the admin panel if a lookup by product ID is done

This makes the product harder to delete.  As a low priority improvement, we should show the product in admin when a search is done.

drbyte:
Admin product listing: include products which are missing a description record

Fixes #6828

- requires the search to include missing-lang products, in order to "find" such products
- requires all the fallback-lookups used when editing/previewing/updating to allow for missing data, in order to prevent filled-in-data from being blanked-out due to failed joins
- requires checking whether the "update" record exists, and fallback to "insert" mode if not found, instead of trying to update the missing record

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -882,10 +882,10 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
             $products_query_raw .= " FROM " . TABLE_PRODUCTS . " p";
             $products_query_raw .= $extra_from;
 
-            $products_query_raw .= " LEFT JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (pd.products_id = p.products_id)";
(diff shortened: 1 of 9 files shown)


drbyte:
Hmmm. There's a ripple-effect problem with this: Editing the product (ie: to "fix" the description) will blank out even more data.

scottcwilson:
Ugh I thought my Audit tool handled this situation but it doesn't.  I'm going to add this.  I'm also going to make the "Fix" links open in a new window so it's faster to use if there are a lot of problems.

drbyte:
- requires the search to include missing-lang products, in order to "find" such products
- requires all the fallback-lookups used when editing/previewing/updating to allow for missing data, in order to prevent filled-in-data from being blanked-out due to failed joins
- requires checking whether the "update" record exists, and fallback to "insert" mode if not found, instead of trying to update the missing record

scottcwilson (approved):
Nice.

PR #6832, opened 2024-10-18
Relocate zen_redirect() calls in admin product copy/move/delete modules

drbyte:
Allows these modules to be called directly from other modules/plugins, particularly in a batch/loop, without triggering the redirects.

- move zen_redirect() to category_product_listing.php
- reduce nested indents by using early-return in copy_product_confirm.php (no other changes, although indentation touches the whole file)

Replaces and closes #6811

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -221,12 +221,19 @@ if (!empty($action)) {
         $delete_linked = 'false';
       }
       require zen_get_admin_module_from_directory($product_type, 'delete_product_confirm.php');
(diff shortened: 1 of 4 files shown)


torvista (approved):
I dropped these files into my site (disabling the custom overrides) and did not see any differences in function from the previous versions.
I also used it with multiple copy, move and delete.
So, no problems from my observations.

torvista:
Is there any reason to not replace the deprecated calls now too?
image

lat9:
I'll note that additional comments on merged PRs tend to get 'lost in the fog', better to open a new issue/discussion.

drbyte:
> Is there any reason to not replace the deprecated calls now too?

We're nearly ready to do a release, so not now. Perhaps in next major version.

PR #6833, opened 2024-10-20
functions_categories: first pass of unambiguous type hints and return types

torvista:
No code changes.

Added type hints to those function parameters where there appeared to be no doubt, and similarly for return types.
Clicked through admin and storefront to check/remove any that caused a problem, so should not affect anything in core code.

It flagged up a couple of inconsistencies in my plugins though, so I guess that's progress.


torvista:
I've had this in production for three months (using strict_mode).
It's highlighted one issue to resolve subsequently: https://github.com/zencart/zencart/pull/6833/commits/d1fe0b46ef4841decc0b015ffbe13facb3dba132

torvista:
Bump.

PR #6834, opened 2024-10-21
Trim key in case there are trailing blanks

scottcwilson:
As reported in 

https://www.zen-cart.com/showthread.php?36126-Track-Package-from-FedEx-UPS-Postal-DHL-etc&p=1403811#post1403811

### DIFF
--- a/laravel/app/Models/Configuration.php
+++ b/laravel/app/Models/Configuration.php
@@ -35,7 +35,7 @@ class Configuration extends Eloquent
                 $configValue = (int)$configValue;
             }
             if (!defined(strtoupper($config['configuration_key']))) {
-                define(strtoupper($config['configuration_key']), $configValue);
+                define(trim(strtoupper($config['configuration_key'])), $configValue);
             }
         }
     }


laravel/app/Models/Configuration.php
```diff
@@ -35,7 +35,7 @@ public function loadConfigSettings()
                 $configValue = (int)$configValue;
             }
             if (!defined(strtoupper($config['configuration_key']))) {
-                define(strtoupper($config['configuration_key']), $configValue);
+                define(trim(strtoupper($config['configuration_key'])), $configValue);
```
drbyte, line 38: Probably should be trimmed up at line 27 before used in the multiple additional places in this function.

drbyte:
I'm inclined to think that a check for this in another area of the code would be better than trimming here.
Probably not so much an "inspection", but an automated cleaner (which trims and re-saves if no duplicate exists, or deletes the untrimmed version if the trimmed version does already exist).

scottcwilson:
I wanted to get something in as a placeholder so I didn't forget.  If you see a better way please make the change - I'm not super familiar with Laravel.

scottcwilson:
Note that this issue is present in 1.5.8+, not just 2.0.x  (i.e. when we transitioned to having Laravel build the defines from the configuration table).

drbyte:
> Note that this issue is present in 1.5.8+, not just 2.0.x (i.e. when we transitioned to having Laravel build the defines from the configuration table).

And in v1.5.7 too, so not related to the Laravel integration:
```php
  $configuration = $db->Execute('SELECT configuration_key AS cfgKey, configuration_value AS cfgValue
                                 FROM ' . TABLE_CONFIGURATION);
  while (!$configuration->EOF) {
    define(strtoupper($configuration->fields['cfgKey']), $configuration->fields['cfgValue']);
    $configuration->MoveNext();
  }
```

drbyte:
After considering this more, I'm inclined (instead of this PR) to put something in zc_install upgrade script to clean any constants that have trailing spaces.
And suggest that the plugin clean itself up in its own future upgrade.

drbyte:
Inspection: `SELECT * FROM configuration WHERE configuration_key RLIKE ' ';`

Fix: 
`UPDATE configuration SET configuration_key = RTRIM( configuration_key ) WHERE configuration_key RLIKE ' ';`

drbyte:
The danger in "fixing" this in an automated way is twofold:
1. "activating" things that have been "dormant"/(broken) forever, and would suddenly "start working" unexpectedly, with potentially confusing side-effects.
2. The need to catch the case where the "fixed" constant is actually also already in the db without the trailing space, and in such case the "bad" one should actually be deleted instead of "fixed".

So, in that regard, I'm inclined to leave it as-is.

Perhaps in the open Discussion for repairing db anomalies we can add a helper to expose such offending config keys and offer the option to delete them, one-off.

scottcwilson:
No argument.  This is a detection that could be added to an audit tool in the future.

PR #6844, opened 2024-11-04
`zen_get_categories_image`, ensure return is a string

### LINKED ISSUE #6843: function zen_get_categories_image can return null or string
Since the db default is null, if the category exists but no image has been defined, it returns null, not string.

https://github.com/zencart/zencart/blob/417f0347558a13768af625cba0596cfff9a7812e/includes/functions/functions_categories.php#L790-L805

lat9:
Fixes #6843

Note: Actual correction on line 807; remaining changes are reformatting/refactoring.

### DIFF
--- a/includes/functions/functions_categories.php
+++ b/includes/functions/functions_categories.php
@@ -793,15 +793,18 @@ function zen_get_category_description($category_id, $language_id = null): string
  * @param $category_id
  * @return string
  */
-function zen_get_categories_image($category_id) {
+function zen_get_categories_image($category_id): string
+{
     global $db;
 
-    $sql = "SELECT categories_image FROM " . TABLE_CATEGORIES . " WHERE categories_id= " . (int)$category_id;
-    $result = $db->Execute($sql);
+    $sql = "SELECT categories_image FROM " . TABLE_CATEGORIES . " WHERE categories_id = " . (int)$category_id;
+    $result = $db->Execute($sql, 1);
 
-    if ($result->EOF) return '';
+    if ($result->EOF) {
+        return '';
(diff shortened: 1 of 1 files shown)


torvista (commented):
Since there is an effort to handle
$result->EOF
I take that to mean the category id was invalid...should some debug notice be issued for that?

lat9:
> Since there is an effort to handle $result->EOF I take that to mean the category id was invalid...should some debug notice be issued for that?

Probably, but that's outside of the scope of the PR and the associated issue.

includes/functions/functions_categories.php
```diff
@@ -793,15 +793,18 @@ function zen_get_category_description($category_id, $language_id = null): string
...
+    $sql = "SELECT categories_image FROM " . TABLE_CATEGORIES . " WHERE categories_id = " . (int)$category_id;
+    $result = $db->Execute($sql, 1);
 
-    if ($result->EOF) return '';
+    if ($result->EOF) {
+        return '';
+    }
 
-    return $result->fields['categories_image'];
+    return (string)$result->fields['categories_image'];
```
drbyte, line 807: Given that the situation we're "preventing" is a `null` being returned, would it be more human-intuitive (for future looking at the code) to use null-coalesce here?

ie: 
```php
return $result->fields['categories_image'] ?? '';
```

(technically that could allow the deletion of the ->EOF check too, but I'm not opposed to keeping that)

lat9: I'd considered that, but felt that simply casting to a string was more 'obvious' and potentially more performant.

PR #6863, opened 2024-11-26
Changing admin image display to size via CSS instead of HTML attribute for products

retched:
In my other edit (#6793), I made this change to allow the images displayed to keep their proportions. I failed to realize I had only changed the category icon in that edit. This change also changes the images of the product listing rows.


retched:
Closing this to split into three separate PR's.

PR #6866, opened 2024-11-27
Changing admin image display to size via CSS instead of HTML attribute for products

retched:
In my other edit (#6793), I made this change to allow the images displayed to keep their proportions. I failed to realize I had only changed the category icon in that edit. This change now makes the same change for the images of the product listing rows.

(Duplicate of #6863, since other edits were added. This PR is the correct one to use since it only contains just the one edit.)

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -951,7 +951,7 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
                         <?= $product['products_name'] ?>
                     </a>
                 </td>
-                <td class="hidden-sm hidden-xs imageView <?= $additionalClass ?>"><?= zen_image(DIR_WS_CATALOG_IMAGES . zen_get_products_image($product['products_id']),'', IMAGE_SHOPPING_CART_WIDTH, IMAGE_SHOPPING_CART_HEIGHT) ?></td>
+                <td class="hidden-sm hidden-xs imageView <?= $additionalClass ?>"><?= zen_image(DIR_WS_CATALOG_IMAGES . zen_get_products_image($product['products_id']),'', IMAGE_SHOPPING_CART_WIDTH, IMAGE_SHOPPING_CART_HEIGHT, 'style="object-fit: contain;"') ?></td>
                 <td class="hidden-sm hidden-xs"><?= $product['products_model'] ?></td>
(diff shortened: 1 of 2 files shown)


admin/category_product_listing.php
```diff
@@ -951,7 +951,7 @@
                         <?= $product['products_name'] ?>
                     </a>
                 </td>
-                <td class="hidden-sm hidden-xs imageView <?= $additionalClass ?>"><?= zen_image(DIR_WS_CATALOG_IMAGES . zen_get_products_image($product['products_id']),'', IMAGE_SHOPPING_CART_WIDTH, IMAGE_SHOPPING_CART_HEIGHT) ?></td>
+                <td class="hidden-sm hidden-xs imageView <?= $additionalClass ?>"><?= zen_image(DIR_WS_CATALOG_IMAGES . zen_get_products_image($product['products_id']),'', IMAGE_SHOPPING_CART_WIDTH, IMAGE_SHOPPING_CART_HEIGHT, 'style="object-fit: contain;"') ?></td>
```
lat9, line 954: Instead of an inline-style, how about adopting the Bootstrap 5 class itself: https://getbootstrap.com/docs/5.3/utilities/object-fit/?

Note that many of the bootstrap 4/5 classes have already been included in the admin's includes/css/stylesheet.css ... alphabetically:
https://github.com/zencart/zencart/blob/b14f1bf2e5643ecbd8c2ee92bb184d6f5f2373c7/admin/includes/css/stylesheet.css#L900-L1181

That'll help to pave the way for an admin bootstrap version update, too.

retched: Would this just be creating the class like such?

``` css
  .object-fit-contain {
    object-fit: contain;
  }
```

For the sake of example? I wouldn't mind making the change but besides the category icons or the product listing, not sure if it would have any usage else.

lat9: Yep, that's it.  If you look at those other bootstrap 4/5 classes that have been added, there's not a lot of usage there either.

Having those CSS classes available makes core and plugin code less vulnerable to additional styling changes once the admin eventually adopts a Bootstrap 5 base.

retched: Done. Added the other `object-fit` classes and changed the category icon and product rows to the same.

PR #6867, opened 2024-11-27
Change description of MFA to add "Zen Cart" suffix, for disambiguation

retched:
Since a person can have multiple services installed on their domain which may use MFA, this edit changes the description of the MFA Handle to better represent what it entails.

In this case, it is the store's name (retrieved from the DB Constant `STORE_NAME`) and the name of the software (ZenCart). For example: if the store's name is "Blockbuster Video", when you scan the MFA QR Code, it will show as "Blockbuster Video - ZenCart" in the MFA tool. This change does NOT affect any already issued MFA tokens as it is only changing the Issuer field and does not touch anything of the issued Secret.

### DIFF
--- a/admin/includes/functions/functions_mfa.php
+++ b/admin/includes/functions/functions_mfa.php
@@ -89,7 +89,8 @@ function zen_mfa_by_totp(array $admin_info = []): bool
         $_SESSION['mfa'] = [];
     }
 
-    $domain = str_replace(['http'.'://', 'https://'], '', HTTP_SERVER);
+    $domain = str_replace(['http'.'://', 'https://'], '', HTTP_SERVER) . ' - Zen Cart';
+    $domain = defined('MFA_DESCRIPTIVE_NAME') ? MFA_DESCRIPTIVE_NAME : $domain;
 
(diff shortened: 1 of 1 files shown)


drbyte:
@retched I like the suggestion. I've found that a lot of stores can have quite long STORE_NAME values, so I adjusted this slightly to still use the domain name, but add the ` - Zen Cart` suffix to it.

Also added an optional new constant `MFA_DESCRIPTIVE_NAME` which one could use to be extra specific if desired, either as a constant added in `extra_datafiles` or `extra_configures`, etc, or as a language-define. (And if one does add it as a language-define (doesn't have to be, but if it is, then the following applies) it could re-use other constants like this: `'MFA_DESCRIPTIVE_NAME' => '%%STORE_NAME%% - Zen Cart'` to make it that little bit more dynamic.)

retched:
To be fair that was a side concern of mine. I just didn't know how long some store names would be. 

Thanks for reviewing.

scottcwilson:
Thanks @retched for helping make Zen Cart better!

PR #6868, opened 2024-11-27
Displays MFA secret during MFA setup

retched:
In the possibility that a person is unable to directly scan in a QR MFA setup barcode, there may be a chance that they can set-up the token by entering the secret instead. This change will display the secret so that it can be copied and pasted (or carefully typed) into the MFA app.

This is done by retrieving the value used in the QR Code generation by displaying the `$_SESSION['mfa']['secret']` value below the QR Code.

This code will NOT be displayed again once it is confirmed.

This change does not affect already issued token secrets and already setup MFA accounts.


retched:
An example of what this change will do:

image

drbyte:
Thanks!

PR #6878, opened 2024-12-03
Changing admin image display to size via CSS instead of HTML attribute for products in preview

retched:
Changing the category image and the product image in preview_info.php

Uses the same class as the edit from #6866.


retched:
Sorry about that, squashed two commits into one. One change actually broke the script and would've generated an error.

PR #6889, opened 2024-12-12
MobileDetect 4.8.09

drbyte:
Patch with latest updates


drbyte:
Note: this is a partial patch of the latest updates to 4.8.09, bypassing the new caching and extra classes that have been added which make it harder to use in procedural code bases like ours.

PR #6894, opened 2024-12-23
Creates a Notifier Hook on Config Change

retched:
This may be possible in other ways, but the creates a hook when a setting in the Configuration Table has been changed by the admin area. It won't pick up if the change is performed by script, SQL import, or other external means).

Some ideas might be if the measuring weight or store's measurement standard has been allowed for all items in the store to be changed quickly. Or if the location itself has changed as well.

### DIFF
--- a/admin/configuration.php
+++ b/admin/configuration.php
@@ -35,6 +35,9 @@ if (!empty($action)) {
                               LIMIT 1");
       zen_record_admin_activity('Configuration setting changed for ' . $result->fields['configuration_key'] . ': ' . $configuration_value, 'warning');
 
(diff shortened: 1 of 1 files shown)


proseLA:
i am all for adding new notifiers to help get the job done without modifying core code; but i am not sure why `$cID` is in position 2?  that would allow one to change the value of that var and then the redirect on line 49 would, in theory, go to that place.

is that what you want?

is not , i would create an array in the first var, and put both values there.

retched:
It might have been an oversight on my part to include the cID#. I think it would be OK to just monitor and report which `configuration_key` was being modified should sufficient. I'll recommit with a change.

PR #6897, opened 2024-12-30
rename file uploads to unix format

zenexpert:
uploaded files are renamed to unix-style names - all lowercase, spaces replaces with dashes, double dots removed. avoids issues with JPG vs jpg files, using user-uploaded images with spaces in srcset tags and similar issues

### DIFF
--- a/includes/classes/upload.php
+++ b/includes/classes/upload.php
@@ -260,10 +260,18 @@ class upload extends base
         }
     }
 
-    function sanitizeFileName($filename) {
+    protected function sanitizeFileName(string $filename): string
+    {
+        // Convert to lowercase
         $filename = strtolower($filename);
+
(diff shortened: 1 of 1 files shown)


includes/classes/upload.php
```diff
@@ -259,4 +259,13 @@ function message_stack($msg = '', $type = '')
             }
         }
     }
+
+    function sanitizeFileName($filename) {
```
lat9, line 263: - Add type hints and function scope, i.e. `protected function sanitizeFilename(string $filename): string`
- Opening brace on a new line

lat9 (changes requested):

zenexpert:
Sorry, keep forgetting about that... :)

lat9 (approved):

PR #6900, opened 2025-01-02
add override ability for upper menu items

proseLA:
this PR does 2 things:

- removes duplicate code so that items do not have to match every time we want to change the upper menu items.
- adds the ability to override those links via a notifier.

admins want quick links to get where they want to go.  this PR will allow people to customize those links to their choosing as opposed to the maintainers choosing.


torvista:
Great stuff. Another two files reverted to vanilla.
I use this to modify the storefront link so it goes to whatever category/product that's being viewed in the admin, and show the current admin name next to logoff.

proseLA:
glad someone else thinks so!

best.

Issue #6902, opened 2025-01-03
admin catalog search parameter double quotes handling


zenexpert:
**Describe the bug**
admin catalog search replaces double quotes and doesn't carry the search term through the pages

**Version**
Zen Cart version: 2.1.0 vanilla install
PHP version: 8.3

**To Reproduce**
Steps to reproduce/demonstrate the behavior:
admin->Catalog->search for a product (already exists in database):
10" rubber duck
Search filter is showing 10" rubber duck
URL is showing index.php?cmd=category_product_listing&search=10"+rubber+duck
click on the product found - URL search parameter changes to search=10&quot; rubber duck
update the product, search parameter changes to search=10

**Expected behavior**
keep the search parameter

drbyte:
While looking into this further, other characters may also need preservation, such as single-quote/apostrophe, and other things that would otherwise get converted to html-special-characters.

So I've added a new function for preserving those characters, just within the search string (not elsewhere, so as to not introduce security risks).

Submitted via PR #6962. Testing and feedback welcomed.

PR #6903, opened 2025-01-03
Special handling for search parameter

zenexpert:
Fixes zencart/zencart#6902


drbyte:
I wonder if hard-coding `search=` into the very generic html_output function is too wide-reaching?

Could this be handled within the controller file/s instead of in the zen_href_link function?

zenexpert:
I'm not sure I understand "too wide-reaching"... Do you see any problems or potential issues with it? IMHO, it's an admin-restricted function so I'm not overly concerned about SQL injections. Although I currently don't see where else this search parameter could be used with double quotes, I don't see a reason for it to not be allowed everywhere in the admin, possibly for some use we can't think of right now. But hey, I've been wrong before (probably more times than I'm willing to admit) so it IS worth considering limiting this to the controller file... ;)

drbyte:
Argh, I'd meant to also mention that there are separate html_output functions files for admin and storefront, and we try to maintain parity between the two, despite a number of notable differences. 
I'd rather not include this one catalog-side, so, that was another reason for wishing to explore whether an alternative is practical or not.

Haven't had time to look at code in more detail myself yet.

PR #6904, opened 2025-01-03
Update header_php.php

retched:
Various whitespace fixing and alignment.


lat9:
You need to rebase your PR; see above.

retched:
Got ya. I think mine would result in the same net sum here. Leaving this one closed.

PR #6907, opened 2025-01-04
Correct PHP warning for empty shipping quote returned

lat9:
If a shipping module returns an empty array for its quotes, a PHP warning is issued:
```
[03-Jan-2025 23:17:03 America/Chicago] Request URI: /index.php?main_page=checkout_shipping, IP address: xx.xx.xx.xx, Language id 1
#0 /includes/templates/template_default/templates/tpl_checkout_shipping_default.php(66): zen_debug_error_handler()
#1 /includes/templates/responsive_classic/common/tpl_main_page.php(181): require('/home/mysite/...')
#2 /index.php(94): require('/home/mysite/...')
--> PHP Warning: Undefined array key "module" in /includes/templates/template_default/templates/tpl_checkout_shipping_default.php on line 66.
```
This can be triggered by configuring the `zones.php` shipping module with a single zone to offer a fixed-price to any non-US-destined order:

image

### DIFF
--- a/includes/classes/shipping.php
+++ b/includes/classes/shipping.php
@@ -259,7 +259,7 @@ class shipping
                     $quotes['tax'] = 0;
                 }
                 $shipping_weight = $save_shipping_weight;
-                if (is_array($quotes)) {
+                if (!empty($quotes) && is_array($quotes)) {
                     $quotes_array[] = $quotes;
                 }
             }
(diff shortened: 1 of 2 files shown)


scottcwilson:
I have seen logs like this too - thanks for taking the time to fix.

PR #6908, opened 2025-01-04
Record/gather order's wholesale status

lat9:
While the `order` table field `is_wholesale` was created in zc200, the `order.php` class neither records that value when an order is placed nor gathers it into a queried order.

This PR corrects that oversight.

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -235,6 +235,7 @@ class order extends base
             'ip_address' => $order->fields['ip_address'],
             'language_code' => $order->fields['language_code'],
             'order_weight' => $order->fields['order_weight'],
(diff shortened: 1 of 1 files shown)


drbyte:
Should null-coalesce be used, ie: `?? 0` in case someone uses this updated class against an old db?

Or maybe there's enough other things that would make it obvious that an upgrade was needed, without needing to accommodate this way.

lat9:
> Should null-coalesce be used, ie: `?? 0` in case someone uses this updated class against an old db?
> 
> Or maybe there's enough other things that would make it obvious that an upgrade was needed, without needing to accommodate this way.

No, because the `null` value indicates that the field has not been set.

PR #6910, opened 2025-01-06
Correct developer tool kit display for files path on Windows

piloujp:
Windows uses backslashes in file paths, but results in Developers Tool Kit are displayed with slashes.


proseLA:
zc does not support windows environments. as such i am not a fan of adding this code.

drbyte:
I don't see the value in this change. And if it gets changed here, someone's gonna want to change it in a thousand other places afterward.

What's the actual "problem" being solved here? Is it merely visual aesthetics? Or is there an actual "problem"? 
Why is the linux-style slash posing a problem? Is it just a visual difference? or is there an actual technical handicap problem caused by it?

piloujp:
It is not much of a problem. What made me do this PR is that admin path is autodetected and displayed using system standard, which means backslash on Windows, but all other paths use slash. You get something like this on Windows system depending on where the file is (admin or catalog):
```
C:\folder\folder\folder\admin\file.php
C:\folder\folder\folder\admin\folder/folder/folder/file.php
C:/folder/folder/folder/folder/folder/file.php
```
It is only visual, but there is not much to do to fix it. The advantage of fixing it would be that you can copy and paste the path directly in Windows Explorer to access the file quicker.
Visually, I don't mind the path using slash, but both backslash and slash together is a little bit disturbing.
I did not see this problem anywhere else, and it is so easy to fix that I did not think there would be so much discussion.

brittainmark:
On Windows 11 you can use either the Windows or the Linux slashes, so you don't need to changes the slashes to paste into Windows 11. (I suspect the same is true of Windows 10).

piloujp:
I was concerned with the mix of slash and backslash in the same path. But I was surprised that Windows (11) accept it too, at least in Explorer and editor I use! Contrary to what I thought, there is no problem here, it is purely visual...

PR #6923, opened 2025-01-20
jscript_sidebox_select_form.php: escape forward slashes for validator

torvista:
2025-01-20 10_18_07-Zen Cart!, The Art of E-commerce

...however I don't see how this code comes into play to test it!


drbyte:
The file was added in #4932 by @lat9 ... but I don't think the feature in that specific function is used in our templates.

PR #6930, opened 2025-01-24
Retire use of E_USER_ERROR for PHP 8.4 compatibility

### LINKED ISSUE #6876: php 8.4: Passing E_USER_ERROR to trigger_error() is deprecated
php 8.4.1, strict reporting
Cropped up in query_factory, but there are multiple instances to change.

drbyte:
Fixes #6876

Adds `exit()`/`die()` to treat as fatal.

And, in order to retain former `logging` as `error` (not "warning"), the word `'FATAL'` is added as a prefix to the error string, when combined with E_USER_WARNING.

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -80,7 +80,7 @@ function zen_href_link($page = '', $parameters = '', $connection = 'SSL', $add_s
         $link = HTTP_CATALOG_SERVER . DIR_WS_CATALOG;
       }
     } else {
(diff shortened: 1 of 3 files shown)


scottcwilson:
This is a compromise that preserves the desired log naming functionality.  Thank you. 

Curses on the PHP core team member who thought this was a good idea.

torvista:
https://github.com/zencart/zencart/blob/9ba4386cdf499c1b00f978dc2a6cd3cb89d54025/laravel/vendor/autoload.php#L17-L19

lat9:
> https://github.com/zencart/zencart/blob/9ba4386cdf499c1b00f978dc2a6cd3cb89d54025/laravel/vendor/autoload.php#L17-L19

There's going to be more than a lousy deprecated log if the site's running a version of PHP < 5.0.

PR #6931, opened 2025-01-24
Avoid use of SID constant with PHP 8.4.1+

### LINKED ISSUE #6906: PHP 8.4 Constant SID is depreciated
Found in functions/html_output.php

// Add the session ID when moving from different HTTP and HTTPS servers, or when SID is defined
    if ($add_session_id == true && $session_started == true && SESSION_FORCE_COOKIE_USE == 'False') {
      if (defined('SID') && !empty(constant('SID'))) {
        $sid = constant('SID');
      } elseif ( ($request_type == 'NONSSL' && $connection == 'SSL' && ENABLE_SSL == 'true') || ($request_type == 'SSL' && $connection == 'NONSSL') ) {
        if ($http_domain != $https_domain) {
          $sid = zen_session_name() . '=' . zen_session_id();
        }
      }
    }

drbyte:
Closes #6906

### DIFF
--- a/admin/includes/classes/split_page_results.php
+++ b/admin/includes/classes/split_page_results.php
@@ -336,7 +336,7 @@ class splitPageResults
                     }
                 }
 
-                if (defined('SID') && !empty(SID)) $display_links .= zen_draw_hidden_field(zen_session_name(), zen_session_id());
+                if (!empty(zen_session_id())) $display_links .= zen_draw_hidden_field(zen_session_name(), zen_session_id());
 
                 $display_links .= '</form>';
             }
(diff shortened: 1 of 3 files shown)


brittainmark (commented):
You took the PHP version check out of the admin version of html_output.php, but not the storefront version. Was this intended?

drbyte:
> You took the PHP version check out of the admin version of html_output.php, but not the storefront version. Was this intended?

?
I intentionally added it in both.

piloujp:
Are you sure this code works?
`if (PHP_VERSION_ID < 80401`
On my PHP console it is always true whatever the PHP version. Shouldn't it be `if (PHP_VERSION_ID < '8.4.1'` ?

lat9:
Unit test suites need updating to remove the check for the zenid/zenAdminID parameters (beyond my ken).

PR #6944, opened 2025-01-27
Remove outdated error-logging configuration settings

### LINKED ISSUE #6942: Remove no longer needed Logging configuration settings
There are a couple of settings in `Configuration :: Logging` that are no longer needed

Image

The first two (the report all errors ones) were needed for older versions of PHP that didn't enable various types of errors to be unconditionally logged.  Now extant versions of PHP no longer honor the change.

Similarly, including the backtrace on notice errors should not be optional as without the backtrace, it's sometimes impossible to determine where a PHP Notice issue is sourced.

lat9:
Fixes #6942


barco57:
I also tested all of @mc12345678 test cases above on a fresh install of master on server used by customers using PHP 8.3.19, and in each case it generated a deprecated.log

PR #6950, opened 2025-02-01
zc_install - rework email confirmation validation

### LINKED ISSUE #6943: installer admin email address validation problem?
**Describe the bug**
When going through installation process and getting to the page where you enter a admin username and enter your email address twice, the second email address field is wrapped in red border with exclamation mark even though the email addresses match between the fields. ignoring and clicking the continue button lets you see quickly the green check mark that it passed validation before moving on. 

**Version**
Zen Cart version:  2.10
PHP version: 8.3
MySQL version: 8
Plugins that may be related: n/a
What older/newer versions were also tested?

**To Reproduce**
go through install to the point you need to enter the superuser admin email address twice

**Expected behavior**
should not appear as if there is a validation error while adding the second email entry unless there is an actual error in the entry

**Screenshots**

Image

**Debug/Error Logs**
N/A
 
N/A

drbyte:
Fixes #6943

### DIFF
--- a/zc_install/includes/template/templates/admin_setup_default.php
+++ b/zc_install/includes/template/templates/admin_setup_default.php
@@ -128,15 +128,21 @@ if (!empty($errors)) { ?>
 <script>
     (() => {
         'use strict'
-        document.getElementById('admin_email2').addEventListener('change', el => {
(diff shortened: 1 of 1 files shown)


drbyte:
@barco57 Do you mind testing this out?

PR #6952, opened 2025-02-02
Send password-reset URL instead of new password

drbyte:
OLD: When a customer requests a password reset, the old behavior was to generate a new password for them, reset it to that password, and email that password to the customer. These emails often got lost in spam traps, and if the email never arrived then the customer could never login again even if they remembered their old password because their old password had been replaced.

NEW: This updated logic now generates a temporary reset token and emails that token in the form of a URL to click to take them to the password_reset page where they can choose their own new password. Once the token is validated and the user selects a (suitable-length) password, their account is updated to use that new password, and the token is deleted.

Tokens also auto-expire after an admin-configurable timeframe, so that old links in old emails don't present a security risk.

NOTE: Retires notifier `NOTIFY_PASSWORD_FORGOTTEN_CHANGED` which is superceded by `NOTIFY_PASSWORD_RESET_URL_SENT` But the same former notifier hooks used with the account_password page (when user chooses their own new password) still fire, for the benefit of 3rd party integrations.

### DIFF
--- a/zc_install/sql/install/mysql_zencart.sql
+++ b/zc_install/sql/install/mysql_zencart.sql
@@ -718,6 +718,19 @@ CREATE TABLE customer_groups (
 );
 
 # --------------------------------------------------------
+#
+# Table structure for table 'customer_password_reset_tokens'
+#
+
+DROP TABLE IF EXISTS customer_password_reset_tokens;
+CREATE TABLE customer_password_reset_tokens (
+  customer_id int(11) NOT NULL default 0,
+  token varchar(100) NOT NULL default '',
+  created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY  (token, customer_id)
+);
+
+# --------------------------------------------------------
 
 #
 # Table structure for table customers_to_groups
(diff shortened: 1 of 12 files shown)


lat9:
Why use a separate table for the login token instead of simply adding the token and timestamp to the `customers` table?

drbyte:
> Why use a separate table for the login token instead of simply adding the token and timestamp to the `customers` table?

Mostly just to keep the disparate data separate. 
It also makes it real easy to look at what pending unused tokens exist.
And avoids schema differences with existing `customers` table which might cause fatal errors in a badly-done upgrade.
But mostly just cuz I preferred to keep it separate.

zc_install/sql/install/mysql_zencart.sql
```diff
@@ -717,6 +717,19 @@ CREATE TABLE customer_groups (
   UNIQUE KEY idx_groupname_zen (group_name)
 );
 
+# --------------------------------------------------------
+#
+# Table structure for table 'customer_password_reset_tokens'
+#
+
+DROP TABLE IF EXISTS customer_password_reset_tokens;
+CREATE TABLE customer_password_reset_tokens (
```
lat9, line 726: Instead of `DROP` and `CREATE`, how about just `CREATE IF NOT EXISTS` on the table (in case there are early adopters)?

drbyte: Fair question.

I opted for DROP/CREATE because that's the standard we use in the rest of the file.
And tokens are only valid for 1 hour, so the impact is essentially nil. Even during an upgrade.

PR #6953, opened 2025-02-02
developer email override; fixes #6905

### LINKED ISSUE #6905: problem with DEVELOPER_OVERRIDE_EMAIL_ADDRESS
IMO, the whole point of having a developer email override is to TEST what emails are going out to which customers and in what situations.  the logic should not change.  unfortunately this is not the case.

we start with the zen_mail function and very early on, we change the to address to the developer address:
https://github.com/zencart/zencart/blob/038d695959ac4c3b6b426c6b907f783bc15cd5f5/includes/functions/functions_email.php#L67-L69
this next section of code changes the `$to_address` to the `$to_email_address`:
https://github.com/zencart/zencart/blob/038d695959ac4c3b6b426c6b907f783bc15cd5f5/includes/functions/functions_email.php#L99-L109
now later on, we test to see what the customer wants:
https://github.com/zencart/zencart/blob/038d695959ac4c3b6b426c6b907f783bc15cd5f5/includes/functions/functions_email.php#L215-L219

which is now using the `DEVELOPER_OVERRIDE_EMAIL_ADDRESS` as the email address.

IMO, this is wrong.  the format for the customer should be determined based on the customer email address NOT the DEVELOPER_OVERRIDE_EMAIL_ADDRESS.

now despite what line 215 says, if there is no record found; the email goes out as text. 

IMO, the changing of the email address to the `DEVELOPER_OVERRIDE_EMAIL_ADDRESS` should be done AFTER the determination of the type of email.

proseLA:
i was on the fence about validating the dev email override; but i feel it is not a bad idea, with minimal processing time.

i'm a fan of logging that error; but not including the notifier `NOTIFY_EMAIL_ADDRESS_VALIDATION_FAILURE`. if you are a developer, learn to read logs, and enter a valid email address.

in addition, if there are multiple email addresses in the `$to_address`, you will now get 1 email for each of those addresses in the format requested by said customer.  which i think is a better testing methodology for ensuring that all of those emails are getting created and getting sent.  which i think mimics the behavior of all of you fancy developers with your own custom email server doing your own fancy overrides there as opposed to within ZC.

Fixes #6905

### DIFF
--- a/includes/functions/functions_email.php
+++ b/includes/functions/functions_email.php
@@ -64,9 +64,6 @@
         if (defined('DEVELOPER_OVERRIDE_EMAIL_STATUS') && DEVELOPER_OVERRIDE_EMAIL_STATUS === 'false') {
             return false;
         }  // disable email sending when in developer mode
-        if (defined('DEVELOPER_OVERRIDE_EMAIL_ADDRESS') && DEVELOPER_OVERRIDE_EMAIL_ADDRESS !== '') {
-            $to_address = DEVELOPER_OVERRIDE_EMAIL_ADDRESS;
-        }
 
         // ignore sending emails for any of the following pages
         // (The EMAIL_MODULES_TO_SKIP constant can be defined in a new file in the "extra_configures" folder)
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks for the edits. I agree with them.

Note to devs using this constant: the defined value must be a single email address now.
I suspect that's how people would have used it previously too; but now it's required else validation will fail (and without validation, sending would fail because its use is in a spot where only a single email address is allowed).

IMO this is better than previous.

PR #6957, opened 2025-02-03
Paypalwpp Admin Notifications: disable refund buttons until confirmed

torvista:
Currently the confirm button is before the confirm checkbox, clicking on Do Full Refund without marking the checkbox refreshes the page with an error.

PP 01

Seems to make more sense to put the confirm checkbox first and button second and disable the button until Confirm is checked.

PP 02


torvista:
Is this contentious?

PR #6974, opened 2025-02-12
Remove unwanted `my_broken_box` sidebox

### LINKED ISSUE #6963: Install SQL: orphan sidebox "my_broken_box"?
https://github.com/zencart/zencart/blob/97aef4d312a776e74deb26475eb433fa424d96ae/zc_install/sql/install/mysql_zencart.sql#L3238-L3240

lat9:
Fixes #6963


drbyte:
I'm pretty sure this was originally created as a testing tool. Faster than manually creating and adding a dummy sidebox and then deleting it to test that the admin UI properly notated a deleted box. Would only appear when forcing use of template_default template, which no end-user would ever do.

But I guess if it's causing confusion we can delete it.

lat9:
No skin in the game here; if that's a testing tool, it's not doing any damage by leaving that record as-is.

PR #6978, opened 2025-02-13
Show true order status, not last public OSH status

### LINKED ISSUE #6977: Customer class getOrderHistory() method issue
If the last update to the order status history table changes the order status but uses customer_notified = -1, the Account page shows the order in the status it was at the last osh update with customer_notified > 0.  This is wrong - the order status is whatever orders.order_status is, irrespective of what the osh record's customer_notifed flag was.

To test, change the status of an order in admin, and set customer_notified = -1.  Now login as the customer who placed the order, and look at the order under My Account.  It will show the order's status as the old status, not the current status. 

Root cause: `getOrderHistory` uses the clause 

```                           
AND osh.customer_notified >= 0
```

scottcwilson:
Fixes #6977

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -619,23 +619,17 @@ class Customer extends base
         $language = $_SESSION['languages_id'];
         global $db, $currencies;
         $sql =
-            "SELECT o.orders_id, o.date_purchased, o.delivery_name,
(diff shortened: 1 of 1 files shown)


scottcwilson:
I am open to this - I will get this PR updated ASAP.

scottcwilson:
Done

Issue #6983, opened 2025-02-16
problem with stats sales report with graphs...


proseLA:
v2.1.0

it seems somewhere along the way, we have lost the ability to filter items on and off based on the status code in this report.

if i were to just select the report from the reports menu, i get something that looks like this:

![Image](https://github.com/user-attachments/assets/c1167982-41f8-4bce-9659-3cc2a649eb4f)

previously one could turn these things on and off by clicking on the icons.

now, the only way i see to manipulate this data is by manually manipulating the url and specifically the `$_GET['filter']` var.  see:

![Image](https://github.com/user-attachments/assets/0ff2f7ca-b464-4c47-a686-94ced6506c8d)

now, i am not sure how this report really works, and whether the data is correct, but it does seem like something got lost in 8906c67.

in addition to losing the icons, it seems that we have lost something in the construction of the url that the icon gets linked to.  as all of the `filter` get vars seem to be equal to 0.

piloujp:
I found multiple problems, 3 actually.
1- Some bad HTML tags.
2- Some bad variable type conditional testing that prevents $sales_report_filter to initialize properly.
3- Light green status icon missing/not mapped in Font Awesome class. 

I can fix problems 1 and 2 but did not find a solution for problem 3.

scottcwilson:
Please make 3.  Thanks for diving into this.

piloujp:
Done.
Problem 1, see PR #6985 
Problem 2, see PR #6986 
Problem 3, see PR #6987

scottcwilson:
All three of the above PRs have now been merged. Thanks everyone!

PR #6986, opened 2025-02-17
Conditional test error on variable $sales_report_filter in stats_sale…

piloujp:
Conditional test error on variable $sales_report_filter in stats_sales_report_graphs.php file
Partial fix for issue #6983

### DIFF
--- a/admin/stats_sales_report_graphs.php
+++ b/admin/stats_sales_report_graphs.php
@@ -292,7 +292,7 @@ for ($i = 0; $i < $report->size; $i++) {
             <td class="dataTableContent text-right"><?php echo FILTER_VALUE ?></td>
           </tr>
           <?php
-          if (($sales_report_filter) === '') {
+          if (empty($sales_report_filter)) {
             for ($i = 0; $i < $report->status_available_size; $i++) {
               $sales_report_filter .= "0";
             }
--- a/admin/stats_sales_report_graphs.php
+++ b/admin/stats_sales_report_graphs.php
@@ -292,7 +292,7 @@ for ($i = 0; $i < $report->size; $i++) {
             <td class="dataTableContent text-right"><?php echo FILTER_VALUE ?></td>
           </tr>
           <?php
-          if (($sales_report_filter) == 0) {
+          if (($sales_report_filter) === '') {
             for ($i = 0; $i < $report->status_available_size; $i++) {
               $sales_report_filter .= "0";
             }


scottcwilson:
Would you please post the log that this fixes (or repro instructions for seeing the log)?  Thanks.

piloujp:
I am not sure what you are speaking of, there is no log to fix here. Anyway here are few points to check:
First, you need to have some sales data.
The value for the GET parameter `filter` is 0 when you click on a red dot on the right of orders status, which is wrong. You can see it in the URL. Before clicking on a red dot, there is nothing which is normal.

This parameter should be some zeros (by default four of them, as many as order status) with a 1 in place of the order status number you click on, something like this `0100` if you clicked on the second red dot (order processing) for example. This is what you get after applying this PR.
Each time you click on a red dot, a new 1 replaces a 0 in the GET parameter filter. There is still the problem with green dots not displayed, which is fixed by PR #6987 .
This GET parameter value is conveyed by `$sales_report_filter variable`. This variable is set to an empty string at line 64 then populated by GET parameter (filter=) value if it exists.
Later in the file, if it has been populated, then it is updated on each click on a dot. If not, it is populated with the default value `0000` before any click is registered. This is where the bug is.

admin/stats_sales_report_graphs.php
```diff
@@ -292,7 +292,7 @@ function drawChart() {
             <td class="dataTableContent text-right"><?php echo FILTER_VALUE ?></td>
           </tr>
           <?php
-          if (($sales_report_filter) == 0) {
+          if (($sales_report_filter) === '') {
```
drbyte, line 295: Would using `empty()` suit? That would accommodate both the blank string and the old check for `0`.

If not, then let's remove the extra set of redundant parentheses, to remove ambiguity.

piloujp: `empty()` is good too, probably safer. I think the `0` check and parentheses I forgot to remove come from a copy and paste from line 75 `if (strlen($sales_report_filter) == 0) { ` that went wrong and are not necessary at all.
`$sales_report_filter` is initialized to an empty string and testing this (`=== ''`) should be enough, but `empty()` has a wider range and might be safer in case of future code modification.

piloujp:
Modified test to use `empty()`.

proseLA:
i have tested empty, and this seems to resolve the last of the errors with regards to the issue i posted.

piloujp:
Yes, tested here too and problem is solved.
As long as the condition is true for an empty string (initial value of `$sales_report_filter`), the problem is solved.

scottcwilson:
Thanks guys!

PR #6999, opened 2025-02-24
Unset's $_SESSION['payment_method_messages'] earlier

retched:
In short, this moves the unsetting of `$_SESSION['payment_method_messages']` to earlier in the checkout process.

Right now, at least by quickly glancing at a few modules and the ZenCart defaults, only AuthorizeAIM uses $_SESSION['payment_method_messages'] in a way to set a post-checkout message to the end user. That same variable is unset at the same time as part of the header_php.php process of checkout_success.

My rationale is that if something happens and that page is refreshed, that content is lost as that variable is unset immediately after assigning the value to `$additional_payment_messages`. Rather than kill it there, I figured might as well leave that SESSION variable alone until either moving away from the page (index) or restarting the checkout process. I chose the latter in this case and had the variable unset as part of the `checkout_confirmation` page. This way if someone checks out again through the normal process, ie. not using PayPal Express, the variable is cleared at the penultimate step and can be reset again by another method.


drbyte:
Thanks for the technical explanation. Now I find myself trying to understand the background behind how this came up.

> and can be reset again by another method.

Such as? 

What's the business case that precipitated the need for this?

retched:
Right now I use three payment modules, each having a situation of leaving instructions for the customer to follow up: moneyorder, zelle, and venmo. To leave instructions for the customer, I pass through a value to be set under `$_SESSION['additional_payment_messages']` and from there, it's assigned to `$additional_payment_messages`. The idea being each time the checkout process is completed, a different message is included.

drbyte:
The original purpose for `$_SESSION['additional_payment_messages']` was to allow a payment module to provide a one-time message to the customer. Usually it's the payment confirmation details, similar to what might show on a printed receipt or a payment-terminal message. I think that was initially in an old Interac module.

It was only made session-bound because of the 3-stage checkout flow where the message needed to be shown only at a certain point post-payment. And the intent was that any refresh of the page would intentionally not re-display that one-time message.

Not sure whether there are any other current-ish modules that use it.

drbyte:
I'm trying to determine whether it's important that it be a one-time-only thing, vs allowed to be retained across page-refreshes.

retched:
Right now, the only (stock) module that uses it is AuthorizeAIM and it's only used for a debugging message. In particular, here:

https://github.com/zencart/zencart/blob/02175e98149ed057fdaf283167fcf3c585ed95d7/includes/modules/payment/authorizenet_aim.php#L489-L491

I'm not sure of other modules that use the variable as a whole. 

If it was supposed to be a "one time message", that message should still persist on the final checkout success screen despite refreshes as it may contain important info. For example, Money Order could print the "send your payment here" messages. For the Zelle/Venmo screen, it contains instructions on how to send the payment through.

drbyte:
After much digging, it appears that that AIM `$response[88]` case is for handling an edge case related to attempting to use gift cards. In the scenario of credit-card-slamming I can see where displaying that only once would make sense. 

But it's also an undocumented scenario. So probably moot.

retched:
> .

I keep forgetting I didn't submit those changes yet to where it would be placed in as well. That would be another repo change.

drbyte:
Most other checkout-related session vars are unset in the checkout_process.php module. But if this needs to be set so it's viewable on checkout_success then resetting with the others at the end of checkout_process would be counterproductive.

drbyte:
Must also consider here that whatever relocation is done may affect other checkout flows such as OPC.

drbyte:
There's no guarantee what page the shopper will go to next after completing a purchase.

Unsetting the session value immediately upon extraction for display to the template is the most efficient code-wise, since that information will no longer be persisted in the session on all subsequent clicks.

I guess if the message is always short, it won't bloat session storage badly.

retched:
Theoretically, it should only be a paragraph big in plain/HTML text. 

The only other thing I could think of, and I'll admit I don't know where or how it's processed, is to hook into the `after_process()` method of the payment method and set $additional_payment_message at that point but I'm not certain if it would hold.

drbyte:
You mentioned checkout_confirmation.
I suppose anywhere that `$_SESSION['payment']` is `unset()` or set `= ''` is a place where (maybe) this var should be cleared too. 

Ajax handler too.

My concern is ensuring it's cleared adequately so that it doesn't falsely carry over when no longer relevant.

retched:
Generally order instructions like mailing address information for Money Orders. I guess tacking those instructions into the order comments could work?

drbyte:
I'm not opposed to your PR; I'm concerned about ripple effect with one-page checkout or other checkout approaches, which might not clear the var. And haven't dug into known addon/plugin code to expose what necessary changes would apply. Don't want to create one problem while solving another.

Hence the exploration of initial objective and contemplating ways to accomplish same.

drbyte:
Probably the simplest for all is to add it to status history.

Can still one-time-flash it with this payment_method_messages var.
But if it's added into order-status-history, then it's always accessible to the customer for subsequent reference.

drbyte:
Recently did similar with PayPal RESTful module: https://github.com/lat9/paypalr/pull/62/files

retched:
I thought I got that when I rebased it on master. Guess I'll try it again.

retched:
Yup. Going to kill this and rebase it.

PR #7001, opened 2025-02-25
Allow search to find recs with missing descs

### LINKED ISSUE #6828: Missing product description record handling

scottcwilson:
Right now if a database has a damaged product record with a missing products description, it's not possible to find and fix (or delete) it in the admin interface.  This change allows the use of the "Restrict Search to Product/Category ID" checkbox to find such records.

Fixes #6828.

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -885,7 +885,11 @@ if (is_dir(DIR_FS_CATALOG_IMAGES)) {
             $products_query_raw .= " LEFT JOIN " . TABLE_PRODUCTS_DESCRIPTION . " pd ON (pd.products_id = p.products_id)";
             $products_query_raw .= $extra_joins;
 
-            $where = " WHERE pd.language_id = " . (int)$_SESSION['languages_id'];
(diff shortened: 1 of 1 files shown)


drbyte:
Yes this lets the user "find" and "open to edit" said product. However, actually editing the product destroys the rest of the product information upon save. 
Same issue preventing https://github.com/zencart/zencart/pull/6831#issuecomment-2423348265 from merging.

So, the "update/save" process needs to also generate the needed record as part of the submission process. The reason it doesn't do that already is because it (largely legitimately) expects the description-table record to exist if the base product-table record exists.

scottcwilson:
LOL I can't even detect that the UPDATE for product_description fails and try an INSERT instead.  This may have to be a utility.

scottcwilson:
I'm going to go ahead and close mine; you can decide if yours is salvageable.

PR #7003, opened 2025-02-27
Add zenDoCurlRequest() function

drbyte:
This will allow making external calls from a shared function without having to build out all the usual setup.

### DIFF
--- a/includes/functions/functions_communications.php
+++ b/includes/functions/functions_communications.php
@@ -1,0 +1,114 @@
+<?php
+
+use http\Exception\BadQueryStringException;
+
+/*
+ * Communication-related functions, such as for making CURL requests
+ *
+ * @copyright Copyright 2003-2025 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id:  New in v2.2.0 $
+ */
+
+/**
+ * Make connection to $url and get a response back.
+ * For GET requests only the $url is needed.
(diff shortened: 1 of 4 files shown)


drbyte:
(The TinyMCE implementation already makes use of this, if present (ie: after this is merged)).

includes/functions/functions_communications.php
```diff
@@ -0,0 +1,81 @@
...
+      curl_setopt($ch, CURLOPT_HEADER, false);
+      curl_setopt($ch, CURLOPT_USERAGENT, $userAgent);
+      curl_setopt($ch, CURLOPT_REFERER, $referrer);
+      curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
+      curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+
+      if (!empty($payload) && strtoupper($method) === 'POST') {
+          curl_setopt($ch, CURLOPT_POST, true);
+          curl_setopt($ch, CURLOPT_POSTFIELDS, $payload);
+      }
```
lat9, line 44: Why are the CURL options 'hard-coded' instead of accepting an array of to-be-applied ones using `curl_setopt_array`?  Those hard-coded ones could be the default if the caller supplies an empty array of options.

drbyte:
Good point.

includes/functions/functions_communications.php
drbyte: Updated.

PR #7004, opened 2025-02-27
allow encapsulated payment modules on admin orders page fixes #6993

### LINKED ISSUE #6993: Admin orders' tool "can't find" an encapsulated payment module
payment modules are expected to NOT be in the zc_plugins directory.  as one can see in the code below, they are expected to be under `DIR_FS_CATALOG_MODULES`

https://github.com/zencart/zencart/blob/a6016b7a750a0e899c2f270cba90e133e0318c7c/admin/orders.php#L508-L517

i would propose that there be an alternative load, going to the `zc_plugins` directory if the file does not exist in the conditional on 509.  

my proposal would be for the zen-cart base to look for a plugin key that is the same name as the payment module key and to load from that directory if found.

this does place a requirement that the plugin key be the same name as the payment module code, but i do not think that to be so burdensome.

the loading of the language file for said module would need to change as well as `zen_include_language_file` also does not look at any plugin directories.

i am open to other solutions, but this is the only one that came to me as of now.

proseLA:
fixes #6993

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -7,6 +7,9 @@
  */
 require('includes/application_top.php');
 
+    use Zencart\FileSystem\FileSystem;
+    use Zencart\ResourceLoaders\ModuleFinder;
+
 // unset variable which is sometimes tainted by bad plugins like magneticOne tools
 if (isset($module)) {
   unset($module);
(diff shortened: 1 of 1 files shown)


lat9:
Would you include an example (for documentation) of how an encapsulated plugin would make use of that newly-added method?

lat9:
Under private discussion, turning on "do not merge".

admin/orders.php
```diff
@@ -506,11 +509,23 @@ function couponpopupWindow(url) {
...
           if (!file_exists($payment_module)) {
               $messageStack->add(sprintf(WARNING_PAYMENT_MODULE_DOESNT_EXIST, $order->info['payment_module_code']), 'warning');
           } else {
             require $payment_module;
-            zen_include_language_file($order->info['payment_module_code'] . '.php', '/modules/payment/','inline');
+              if (dirname($modules_found[$payment], 3) !== '.') {
+                  $languageLoader->loadModuleLanguageFile($payment, 'payment');
+              } else {
+                  zen_include_language_file($order->info['payment_module_code'] . '.php', '/modules/payment/', 'inline');
+              }
```
lat9, line 528: You don't need the `dirname` conditional check; the languageLoader method will load the language file based on the standard search order (zc_plugin, then base language then template-override).

lat9 (changes requested):
See comment

admin/orders.php
proseLA: the method `zen_include_language_file` includes logic to display something in the message stack in the event of a missing file.  if you want to remove that logic, feel free to do so.  alternatively, feel free to change the aforementioned method to just use the `$languageLoader` the way it is used here, and then display the message if the return is false. i really am not interested in doing a bunch more commits to get things exactly the way you want them.  i expressed a problem; you lead me on a path to a solution; i have in turn implemented that so that an encapsulated payment plugin can now be used on this page without an error; and at the same time preserved the existing logic for non-encapsulated payment modules.  if you want to change that, feel free.  there is a reason i checked this box.  please take advantage of it:

image

PR #7018, opened 2025-03-04
Verify current product has a description record

### LINKED ISSUE #7012: ZC 2.1.0 logs when product description record doesn't exist
If there is a database record for a specific product_id but no matching record in the products_description table, several logs are produced in ZC 210.  It's not clear to me what the best resolution is:

- redirect to product not found
- use a default name ("Product nnn" perhaps)
-  dynamically create the products description record (with a default name). 

[02-Mar-2025 19:55:51 America/Chicago] Request URI: /index.php?cPath=3&main_page=product_info&products_id=168, IP address: 20.171.207.206, Language id 1
#0 /home/client/public_html/includes/modules/pages/product_info/main_template_vars.php(66): zen_debug_error_handler()
#1 /home/client/public_html/includes/templates/bootstrap/common/tpl_main_page.php(233): require('/home/client/...')
#2 /home/client/public_html/index.php(94): require('/home/client/...')
--> PHP Warning: Undefined array key "lang" in /home/client/public_html/includes/modules/pages/product_info/main_template_vars.php on line 66.

[02-Mar-2025 19:55:51 America/Chicago] Request URI: /index.php?cPath=3&main_page=product_info&products_id=168, IP address: 20.171.207.206, Language id 1
#0 /home/client/public_html/includes/modules/pages/product_info/main_template_vars.php(66): zen_debug_error_handler()
#1 /home/client/public_html/includes/templates/bootstrap/common/tpl_main_page.php(233): require('/home/client/...')
#2 /home/client/public_html/index.php(94): require('/home/client/...')
--> PHP Warning: Trying to access array offset on null in /home/client/public_html/includes/modules/pages/product_info/main_template_vars.php on line 66.

scottcwilson:
Fixes #7012

### DIFF
--- a/includes/classes/Product.php
+++ b/includes/classes/Product.php
@@ -25,6 +25,8 @@ class Product
     /** @deprecated use !exists()  */
     public bool $EOF = true;
 
+    public bool $has_description = false; 
+
     public function __construct(protected ?int $product_id = null)
     {
         $this->initLanguages();
@@ -35,6 +37,7 @@ class Product
             // set some backward compatibility properties
             $this->fields = $this->data;
             $this->EOF = empty($this->data);
+            $this->has_description = (!empty($this->data['products_name'] ?? '') || $this->hasProductDescription($this->product_id)); 
         }
     }
 
@@ -99,7 +102,12 @@ class Product
 
     public function exists(): bool
     {
-        return !empty($this->product_id) && !empty($this->data);
+        if (!empty($this->product_id)) { 
+            if (!empty($this->data['products_name']) || $this->has_description) {
+                return true; 
+            }
+        }
+        return false;
     }
     public function isValid(): bool
     {
@@ -179,6 +187,15 @@ class Product
         $sql = "SELECT products_id FROM " . TABLE_PRODUCTS_DISCOUNT_QUANTITY . " WHERE products_id=" . (int)$this->product_id;
         $results = $db->Execute($sql, 1);
         return !$results->EOF;
+    }
+
+    public function hasProductDescription(int $product_id)
+    {
+        // Remember constructor has not yet completed running.  No $this.
+        global $db;
+        $sql = "SELECT count(products_id) AS found FROM " . TABLE_PRODUCTS_DESCRIPTION . " WHERE products_id = " . (int)$product_id . " AND language_id = " . (int)$_SESSION['languages_id'];
+        $result = $db->Execute($sql, 1);
+        return !empty($result->fields['found']);
     }
 
     public function hasPriceSpecials()


includes/classes/Product.php
```diff
@@ -99,7 +102,12 @@ public function getId(): ?int
 
     public function exists(): bool
     {
-        return !empty($this->product_id) && !empty($this->data);
+        if (!empty($this->product_id)) { 
+            if (!empty($this->data['products_name']) || $this->has_description) {
+                return true;
```
proseLA, line 107: there is no need for a nested `if` here.

in addition, i am not sure about the logic.  we have a product object which we are now saying it does not exist.  but it does exist, but it is missing a description.  how could one differentiate between a product_id which has no data and one which has data but no product description?  i would think the product object should be able to provide that information as opposed to saying they are the same; which this code does.

handling a product with no description by saying it does not exist just seems wrong to me.  it makes it ambiguous.

includes/classes/Product.php
```diff
@@ -181,6 +189,15 @@ public function hasPriceQuantityDiscounts(): bool
...
     }
 
+    public function hasProductDescription(int $product_id)
+    {
+        // Remember constructor has not yet completed running.  No $this.
+        global $db;
+        $sql = "SELECT count(products_id) AS found FROM " . TABLE_PRODUCTS_DESCRIPTION . " WHERE products_id = " . (int)$product_id . " AND language_id = " . (int)$_SESSION['languages_id'];
+        $result = $db->Execute($sql, 1);
+        return !empty($result->fields['found']);
+    }
```
proseLA, line 199: - i think the comment line is unnecessary as well as untrue.  i do not know why it is here.  it is attempting to justify the code as opposed to providing any insight as to what the code does.
- i suppose it is a question of style (maybe), but i do not understand why we need to use the mysql `count` function.
- i have no idea if our `db` class returns the string 0 or the integer 0, and i would have to look that up.
- casting `$product_id` to `int` is completely unnecessary.
- it just seems that removing the `count` method and using the `db` `EOF` indicator would make the code far more legible and clear to me.

proseLA (commented):
i am not a fan.

`exists` is now ambiguous.

while there is no per se excepted code style, i see no need for nesting ifs.

includes/classes/Product.php
scottcwilson: This code is from Dr. Byte's commit to fix this issue on the admin side.

proseLA: > This code is from Dr. Byte's commit to fix this issue on the admin side.

and your point is?

perhaps if Dr. Byte had made the commit, i would be less inclined to comment on it; but my feeling on it is still the same.

in addition, perhaps next time, you can specifically reference the commit as opposed to making me look through the code for it?  i suppose it is this commit 8151713db?  because when i look at that, i am not sure why we are not using the product class as opposed to coding all of the db calls in the product listing script?  why not:

```
$product_id = new Product($_GET['pID']);
```

but for whatever reason, he has chosen not to do that.

and in his code, he needs to cast `product_id` to int; here it is superfluous.

your response does not address any of the points i bring up.  it has your name on the code.  i still stand by my comments.

the product object is now ambiguous.  ideally it should not be.

drbyte:
So, if `exists()` is ambiguous, then given that the ONLY place in our code base where `exists()` is used is to determine whether a product-page should be shown to the shopper (vs redirecting to a 404 etc), 
then perhaps we don't need an `exists()` method at all, and should simply rename it to `isIntegrousEnoughToShowOnStorefrontSoShoppersAndBotsCanViewItForBuyingOrIndexing()`?

proseLA:
> So, if `exists()` is ambiguous, then given that the ONLY place in our code base where `exists()` is used is to determine whether a product-page should be shown to the shopper (vs redirecting to a 404 etc), then perhaps we don't need an `exists()` method at all, and should simply rename it to `isIntegrousEnoughToShowOnStorefrontSoShoppersAndBotsCanViewItForBuyingOrIndexing()`?

i don't think sarcasm from the maintainer of this repo is a good look.

lat9:
How about, as an alternate solution, we just plunk the following code
```php
        if (IS_ADMIN_FLAG === false && !isset($data['lang'][$_SESSION['languages_code']])) {
            $data['lang'][$_SESSION['languages_code']] = [
                'language_id' => $_SESSION['languages_id'],
                'products_name' => '',
                'products_description' => '',
                'products_url' => null,
                'products_viewed' => 0,
                'description_record_missing' => true,
            ];
            $this->notify('NOTIFY_PRODUCT_DETAILS_NO_DESCRIPTION', (int)$product_id, $data);
        }
```

... right after this section:
https://github.com/zencart/zencart/blob/429635238c41e4c9becf2838ddab18f96cb209e7/includes/classes/Product.php#L239-L250

That'll
- Ensure that the base `products_description` fields are present even if there's no entry in the database for the current product
- Give an observer the chance to inject any additional default fields from that table or take a site-specific action to notify someone of that missing data
- Add an entry in the array to indicate that the description-table record is missing

drbyte:
@lat9 I support your proposed alternative. It's cleaner insomuch as it simply prevents the empty-array situation that was triggering the PHP warnings.

It does make it possible that the product will return blank description if a language-specific description/name is missing, despite being present in other languages.

Neither solution directly triggers a 404 to the shopper despite the product being undescribed and therefore shouldn't be purchasable. Although, Scott's code combined it to the exists() check so the combined effect was a not-found response from calling code; but upon further reflection I do agree with Paul's concern that exists() doesn't need to mean "valid in every way".

lat9:
> @lat9 I support your proposed alternative. It's cleaner insomuch as it simply prevents the empty-array situation that was triggering the PHP warnings.
> 
> It does make it possible that the product will return blank description if a language-specific description/name is missing, despite being present in other languages.
> 
> Neither solution directly triggers a 404 to the shopper despite the product being undescribed and therefore shouldn't be purchasable. Although, Scott's code combined it to the exists() check so the combined effect was a not-found response from calling code; but upon further reflection I do agree with Paul's concern that exists() doesn't need to mean "valid in every way".

As much as I hate what I'm saying ... what if we had yet-another soft-configuration setting that identifies whether/not a product with no description record is purchasable?

PR #7019, opened 2025-03-05
Fix errors in banner manager when Group names don't conform to pattern

scottcwilson:
```
[05-Mar-2025 11:52:02 UTC] Request URI: /client/admin/index.php?cmd=banner_manager&page=1&bID=6, IP address: ::1, Language id 1
#0 /Users/scott/Sites/client/admin/banner_manager.php(554): zen_debug_error_handler()
#1 /Users/scott/Sites/client/admin/index.php(16): require('/Users/scott/Si...')
--> PHP Warning: Undefined array key 1 in /Users/scott/Sites/client/admin/banner_manager.php on line 554.

```


lat9:
What change to the default `configuration_title` values for the various banner groups caused this PHP Warning to be issued?

scottcwilson:
They have added banner groups which don't have dashes in their configuration text.

drbyte:
Appears to have been caused by https://github.com/zencart/zencart/pull/6679

scottcwilson:
Most of the time that logic would work fine.  We just have to prevent a log from happening when the string value is different from the English language default, which it could easily be in many cases.

PR #7022, opened 2025-03-06
Ensure methods array key is set before reference

scottcwilson:
It is possible that the `$quotes` array will not have the `methods` key set.  For example, in zones shipping if MODULE_SHIPPING_ZONES_SKIPPED is the current zone (see includes/modules/shipping/zones.php line 316).

Associated log:
```
[06-Mar-2025 22:31:20 UTC] Request URI: /client/index.php?main_page=checkout_shipping, IP address: ::1, Language id 1
#0 /Users/scott/Sites/client/includes/modules/pages/checkout_shipping/header_php.php(191): zen_debug_error_handler()
#1 /Users/scott/Sites/client/index.php(35): require('/Users/scott/Si...')
--> PHP Warning: Undefined array key "methods" in /Users/scott/Sites/client/includes/modules/pages/checkout_shipping/header_php.php on line 191.

```

### DIFF
--- a/includes/modules/pages/checkout_shipping/header_php.php
+++ b/includes/modules/pages/checkout_shipping/header_php.php
@@ -188,7 +188,7 @@ if (isset($_SESSION['cart']->cartID)) {
   if (isset($_SESSION['shipping']['id'])) {
     $checklist = [];
     foreach ($quotes as $key=>$val) {
-      if (is_array($val['methods'])) {
+      if (isset($val['methods']) && is_array($val['methods'])) {
         foreach($val['methods'] as $key2=>$method) {
           $checklist[] = $val['id'] . '_' . $method['id'];
         }


drbyte:
I have no problem with this specific fix. 
But I'm puzzled that it hasn't previously come up as an issue over the years.
And are there any other places where it could be empty but we're not accounting for that?

scottcwilson:
> But I'm puzzled that it hasn't previously come up as an issue over the years.

Warning on key not present is new in PHP 8.  Didn't used to be an warning condition. 

> And are there any other places where it could be empty but we're not accounting for that?

Sure, also includes/modules/shipping/zones.php line 319.  It's not common but it could happen.   And it could certainly happen in a shipping module using an API where there's the possibility of errors, etc.

lat9:
> It is possible that the `$quotes` array will not have the `methods` key set. For example, in zones shipping if MODULE_SHIPPING_ZONES_SKIPPED is the current zone (see includes/modules/shipping/zones.php line 316).
> 
> Associated log:
> 
> ```
> [06-Mar-2025 22:31:20 UTC] Request URI: /client/index.php?main_page=checkout_shipping, IP address: ::1, Language id 1
> #0 /Users/scott/Sites/client/includes/modules/pages/checkout_shipping/header_php.php(191): zen_debug_error_handler()
> #1 /Users/scott/Sites/client/index.php(35): require('/Users/scott/Si...')
> --> PHP Warning: Undefined array key "methods" in /Users/scott/Sites/client/includes/modules/pages/checkout_shipping/header_php.php on line 191.
> ```

In my mind, that PHP Warning indicates that a shipping module doesn't conform to the documented interface.  See this doc-page for details: https://docs.zen-cart.com/dev/code/modules/shipping_modules/#quote

scottcwilson:
That document says an empty array is acceptable, and this is what's being returned.  

"If no quote(s) is/are available ... the shipping module sets its quotes property to either (bool)false or an empty array ([]) and returns that value."

lat9:
> That document says an empty array is acceptable, and this is what's being returned.
> 
> "If no quote(s) is/are available ... the shipping module sets its quotes property to either (bool)false or an empty array ([]) and returns that value."

And the `shipping.php` class handles that by not including an empty array of quotes in the quotes' array returned:
https://github.com/zencart/zencart/blob/dc32e4105a118fa56226f3cc0ff6a4e36c5a8547/includes/classes/shipping.php#L257-L264

scottcwilson:
That fix didn't make it into 2.1.0.

As shipped version 2.1.0 of includes/classes/shipping.php has line 262 as 

```
 if (is_array($quotes)) { 
```

Can we please just add this fix as an easy anti-bugging feature and move on?

lat9:
> That fix didn't make it into 2.1.0.
> 
> As shipped version 2.1.0 of includes/classes/shipping.php has line 262 as
> 
> ```
>  if (is_array($quotes)) { 
> ```
> 
> Can we please just add this fix as an easy anti-bugging feature and move on?

This change is redundant; it didn't get into the zc210 release either.  It's better to correct an issue at the source rather than providing sticking-plasters.

drbyte:
So, just to confirm ...
This only happens when the Zones module has misconfigured settings.
So this PR is essentially a duplicate of #6907 ?
While the logged warnings show different line numbers, they're caused by the same root issue.
And either "fix" could be used to solve the same reported problem if someone ran into it up to v2.1.0 ?

lat9:
> So, just to confirm ... This only happens when the Zones module has misconfigured settings. So this PR is essentially a duplicate of #6907 ? While the logged warnings show different line numbers, they're caused by the same root issue. And either "fix" could be used to solve the same reported problem if someone ran into it up to v2.1.0 ?

Yes, it's essentially a duplicate of #6907.

scottcwilson:
a) I don't think it's a misconfiguration. 
b) Without this fix you will get a log in a bunch of cases, like when someone just returns an array with the error element set. 
Also, this fix is closer to the log they will likely see and thus easier for novices to grok and apply.   

Check before reference is part of life in PHP8.

drbyte:
> Also, this fix is closer to the log they will likely see and thus easier for novices to grok and apply.

Noted.

> PHP8

Noted

That said, doesn't the #6907 fix to `/includes/classes/shipping.php` fix the issue way better and more efficiently than this location in `checkout_shipping/header_php.php`?

What I'm getting at is: yes, while we can (and probably will) patch this spot with an isset() call, the advice given to anyone "today" should probably still be primarily to patch the shipping.php class, regardless of this header_php loop.

drbyte:
Oops. Duplicated and fixed in #7202

PR #7023, opened 2025-03-07
Handle gender radio button not selected

scottcwilson:
For the new shipping address and new billing address screens, the gender is not preselected.  We must anticipate someone not selecting a gender when the form is submitted. 

Sample log created: 
```
[06-Mar-2025 18:32:25 UTC] Request URI: /store/index.php?main_page=checkout_shipping_address, IP address: xxx.xxx.xx.xxx, Language id 1
#0 /home/client/public_html/store/includes/modules/checkout_new_address.php(32): store_debug_error_handler()
#1 /home/client/public_html/store/includes/modules/pages/checkout_shipping_address/header_php.php(46): require('/home/client/publi...')
#2 /home/client/public_html/store/index.php(35): require('/home/client/publi...')
--> PHP Warning: Undefined array key "gender" in /home/client/public_html/store/includes/modules/checkout_new_address.php on line 32.

```

### DIFF
--- a/includes/modules/checkout_new_address.php
+++ b/includes/modules/checkout_new_address.php
@@ -29,7 +29,7 @@ if (isset($_POST['action']) && ($_POST['action'] === 'submit')) {
     if (!empty($_POST['firstname']) && !empty($_POST['lastname']) && !empty($_POST['street_address'])) {
         $process = true;
         if (ACCOUNT_GENDER === 'true') {
-            $gender = zen_db_prepare_input($_POST['gender']);
+            $gender = zen_db_prepare_input($_POST['gender'] ?? '');
         }
         if (ACCOUNT_COMPANY === 'true') {
             $company = zen_db_prepare_input($_POST['company']);


lat9:
Isn't the root-cause of the issue the fact that the template_default/responsive_classic templates leave both radio buttons unchecked?
https://github.com/zencart/zencart/blob/dc32e4105a118fa56226f3cc0ff6a4e36c5a8547/includes/templates/template_default/templates/tpl_modules_checkout_new_address.php#L19-L26

scottcwilson:
If you want to track that down and open a new PR I will happily close this PR.

drbyte:
One could argue:
- current implementation only uses gender for customizing greeting/salutation/welcome text onscreen and in emails, and (should) silently accommodate blank value by just leaving out any associated greeting text. I believe this is why both buttons are left unchecked on new addresses
- why is a gender needed for shipping/billing?

Based on the above I'd accept this PR.

But if `ACCOUNT_GENDER === 'true'` means "we DO want to collect that info for ALL addresses" (especially since any address could later be selected as "default"), then:
- could the previously-chosen gender for the "default" address be used as pre-selected?

scottcwilson:
It's not clear to me that pre-setting the gender is the right thing to do, and the current code handles the case where it's not set (see includes/modules/checkout_new_address.php line 55), except for the new PHP 8 requirement that the for null coalescing if no value is present).  This change satisfies the new PHP 8 requirement.

PR #7024, opened 2025-03-09
`freeoptions` shipping can be invalidly enabled

lat9:
See [this](https://www.zen-cart.com/showthread.php?230639-Free-Shipping-Issue&p=1406530) Zen Cart forum posting for the initial report of the problem.

Essentially, the checks in the `checkForFreeOptions` method are too strict, resulting in the shipping module being enabled even though the configuration indicates that it should not be.

With this configuration, for example, on an order with a weight of 7 the "Free Shipping Options" is still enabled due to MODULE_SHIPPING_FREEOPTONS_TOTAL_MIN being set to '0.00':

image

### DIFF
--- a/includes/modules/shipping/freeoptions.php
+++ b/includes/modules/shipping/freeoptions.php
@@ -76,14 +76,24 @@ class freeoptions extends ZenShipping
     protected function checkForFreeOptions()
     {
         global $order;
+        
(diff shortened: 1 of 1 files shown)


scottcwilson:
LOL I got stung by this very issue recently.

PR #7029, opened 2025-03-15
Changed log entry for adding configuration key

retched:
It read as if it was deleting keys for admin pages instead of adding a new value.


drbyte:
Thanks!

PR #7030, opened 2025-03-16
Add cron for activating enabled products

drbyte:
This cron specifically activates products as-of the current minute (actually to the "second") when the cron job runs, by passing `false` to new `$useMidnight` parameter.

`php /path/to/admin-dir/products_cron.php` with optional timezone being passed. Call with `-help` for details.

If one wanted to bypass using the cron job but set activations to happen "at the current minute", passing `false` to the new `$useMidnight` parameter where this function is called in `/includes/init_includes/init_special_funcs.php` would accomplish the same.

These changes are relatively straightforward because the `products_date_available` field is a `datetime` field already, supporting not just a date but also time. So, as long as that field is set to the desired minute (or second) for an activation time, the next time this function runs with `$useMidnight` set to `false`, the product will be activated. (Previously it would activate "all products whose date matches today's date, regardless of minutes/seconds", meaning "even products which activate 2 hours from now would be activated", which might be undesired in some cases.)

Closes #7027 discussion/request.

### DIFF
--- a/admin/includes/auto_loaders/products_cron.core.php
+++ b/admin/includes/auto_loaders/products_cron.core.php
@@ -1,0 +1,82 @@
+<?php
+
+/**
+ * @copyright Copyright 2003-2025 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id:  New in v2.2.0 $
+ */
+
+/**
+ * autoloader array for admin products_cron.php
+ */
+$autoLoadConfig[0][] = [
+    'autoType' => 'require',
+    'loadFile' => DIR_FS_CATALOG . DIR_WS_INCLUDES . 'version.php',
+];
+$autoLoadConfig[0][] = [
+    'autoType' => 'class',
+    'loadFile' => 'class.notifier.php',
+];
+$autoLoadConfig[0][] = [
+    'autoType' => 'classInstantiate',
+    'className' => 'notifier',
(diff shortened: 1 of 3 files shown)


drbyte:
@wcbutler feel free to try out these changes per your discussion in #7027

piloujp:
Perhaps I am missing something, but I can't see how you can set time. Well, you can set your crown job running time, but date available time in database is always `00:00:00`. Problem is that function `zen_enable_disabled_upcoming` will not only be called by cron job but also each time a user session is reset (as it seems to do before).
On a busy site, you end up with your product activated soon after midnight...

piloujp:
These options would cover all cases, I think. But whatever the solution, we need to offer users an easy way to set and manage products release time... It is why I am not too keen on cron jobs. You can't set or check them in ZC admin interface, which is not very user-friendly. And if it is not user-friendly, it is not going to be used much.
It is why I was thinking more of MySQL events.
In this case, the user would just input time (and date) in products' edit page, which would be saved in database. Then when function `zen_enable_disabled_upcoming` is call by existing ZC triggers/autoload (or a cron job every day after midnight), it could create a MySQL event for today's products' release. If the `jQuery datepicker` had a time option, it would be very easy to do.

piloujp:
If you are interested in the idea of using MySQL scheduled events, you can take a look at [this code](https://github.com/piloujp/zencart/tree/Products_available_time) on my GitHub repo. I tested it and it performs well.

PR #7032, opened 2025-03-17
Correct issues

### LINKED ISSUE #4432: Why is FileSystem->isAdminDir response backwards?

mc12345678:
Fixes #4432, v210 fixes #4435, fixes #5810

### DIFF
--- a/includes/classes/FileSystem.php
+++ b/includes/classes/FileSystem.php
@@ -80,16 +80,50 @@ class FileSystem extends IlluminateFilesystem
         return $found;
     }
 
+    public function isAdminRootDir(string $filePath): bool
+    {
+        if (!defined('DIR_FS_ADMIN')) {
+            return false;
+        }
+        $filePath = $filePath . DIRECTORY_SEPARATOR;
+        $test = str_replace(DIR_FS_ADMIN, '', $filePath);
+        if ($test !== $filePath && ($test === '' || $test === DIRECTORY_SEPARATOR)) {
+            return true;
+        }
+        return false;
+    }
+
     public function isAdminDir(string $filePath): bool
     {
         if (!defined('DIR_FS_ADMIN')) {
             return false;
         }
+        if ($this->isAdminRootDir($filePath)) {
+            return true;
+        }
+        $filePath = $filePath . DIRECTORY_SEPARATOR;
         $test = str_replace(DIR_FS_ADMIN, '', $filePath);
-        if ($test != $filePath) {
+        if ($test !== $filePath) {
+            return true;
+        }
+        return false;
+    }
+
+    public function isCatalogRootDir(string $filePath): bool
+    {
+        if ($this->isAdminDir($filePath)) {
             return false;
         }
-        return true;
+        if (!defined('DIR_FS_CATALOG')) {
+            return false;
+        }
+        $filePath = $filePath . DIRECTORY_SEPARATOR;
+        $test = str_replace(DIR_FS_CATALOG, '', $filePath);
+        if ($test !== $filePath && ($test === '' || $test === DIRECTORY_SEPARATOR)) {
+            return true;
+        }
+        return false;
+
     }
 
     public function isCatalogDir(string $filePath): bool
@@ -100,21 +134,29 @@ class FileSystem extends IlluminateFilesystem
         if (!defined('DIR_FS_CATALOG')) {
             return false;
         }
+        if ($this->isCatalogRootDir($filePath)) {
+            return true;
+        }
+        $filePath = $filePath . DIRECTORY_SEPARATOR;
         $test = str_replace(DIR_FS_CATALOG, '', $filePath);
         if ($test !== $filePath) {
-            return false;
+            return true;
         }
-        return true;
+        return false;
 
     }
 
     public function getRelativeDir(string $filePath): string
     {
         if ($this->isAdminDir($filePath)) {
-            return str_replace(DIR_FS_ADMIN, '', $filePath);
+            $returnPath = str_replace(substr(DIR_FS_ADMIN, 0, -1), '', $filePath);
+            $returnPath = substr($returnPath, 1);
+            return $returnPath;
         }
         if ($this->isCatalogDir($filePath)) {
-            return str_replace(DIR_FS_CATALOG, '', $filePath);
+            $returnPath = str_replace(substr(DIR_FS_CATALOG, 0, -1), '', $filePath);
+            $returnPath = substr($returnPath, 1);
+            return $returnPath;
         }
         return $filePath;
     }
--- a/includes/classes/InitSystem.php
+++ b/includes/classes/InitSystem.php
@@ -88,16 +88,21 @@ class InitSystem
         if (isset($entry['classPath'])) {
             $filePath = $entry['classPath'];
         }
-        if ($entry['loaderType'] === 'plugin') {
-            $filePath = $this->findPluginDirectory($entry['classPath'] ?? DIR_WS_CLASSES, $entry['pluginInfo']['unique_key']);
+        $classResult = 'ZC';
+        if ($entry['loaderType'] == 'plugin') {
+            $pluginFilePath = $this->findPluginDirectory($filePath, $entry['pluginInfo']['unique_key']);
+            if (file_exists($pluginFilePath . $entry['loadFile'])) {
+                $filePath = $pluginFilePath;
+                $classResult = 'PLUGIN';
+            }
         }
-        $this->debugList[] = 'processing class - ' . $filePath  . $entry['loadFile'];
+        $this->debugList[] = 'processing ' . $classResult . ' class - ' . $filePath  . $entry['loadFile'];
         $result = 'FAILED';
         if (file_exists($filePath . $entry['loadFile'])) {
             $result = 'SUCCESS';
             $this->actionList[] = ['type' => 'include', 'filePath' => $filePath . $entry['loadFile'], 'forceLoad' => $entry['forceLoad']];
         }
-        $this->debugList[] = 'loading class - ' . $filePath . $entry['loadFile'] . ' - ' . $result;
+        $this->debugList[] = 'loading ' . $classResult . ' class - ' . $filePath . $entry['loadFile'] . ' - ' . $result;
     }
 
     /**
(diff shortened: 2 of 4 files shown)


lat9:
If this PR were modified, removing the changes to the `InitSystem` and `FileSystem` classes, I'd be happy to merge those other two file changes.

I'd do it myself, but maintainers aren't allowed to modify this PR.

includes/classes/FileSystem.php
```diff
@@ -100,21 +130,25 @@ public function isCatalogDir(string $filePath): bool
...
+        return false;
 
     }
 
     public function getRelativeDir(string $filePath): string
     {
         if ($this->isAdminDir($filePath)) {
-            return str_replace(DIR_FS_ADMIN, '', $filePath);
+            $returnPath = str_replace(dirname(DIR_FS_ADMIN), '', $filePath);
+            $returnPath = substr($returnPath, 1);
```
drbyte, line 145: Is this `substr(x, 1)` just intended to strip a leading slash? Would `ltrim()` be more clear about the intent?

drbyte:
> I'd do it myself, but maintainers aren't allowed to modify this PR.

Ya, most projects would close and reject the PR when that flag is set, since it is only set by the person submitting the PR.

piloujp:
I tested this PR, but I got an error 500 as soon as I try to install plugins with class auto loader like POSM.
```
[20-Mar-2025 12:35:59 Asia/Tokyo] PHP Fatal error:  Uncaught Error: Class "products_options_stock_observer" not found in ..\includes\autoload_func.php:47
Stack trace:
#0 ..\includes\application_top.php(329): require()
#1 ..\index.php(25): require('...')
#2 {main}
  thrown in ..\includes\autoload_func.php on line 47
```

includes/classes/FileSystem.php
mc12345678: Yes, basically to remove a left directory divider in whatever OS was involved, with attempts to replicate what seemed to be the desired method of optimization rather than logical or convenient. Also the idea was to remove only one instance of, even if incorrectly populated with a duplicate of the character. 

Ltrim would take off however many of the character existed on the left side until it wasn't present, not just one. 

I'm otherwise not hard pressed on which one or the other method to use.

drbyte: > Ltrim would take off however many of the character existed on the left side until it wasn't present, not just one.

Good point.

lat9:
> I tested this PR, but I got an error 500 as soon as I try to install plugins with class auto loader like POSM.
> 
> ```
> [20-Mar-2025 12:35:59 Asia/Tokyo] PHP Fatal error:  Uncaught Error: Class "products_options_stock_observer" not found in ..\includes\autoload_func.php:47
> Stack trace:
> #0 ..\includes\application_top.php(329): require()
> #1 ..\index.php(25): require('...')
> #2 {main}
>   thrown in ..\includes\autoload_func.php on line 47
> ```
> 

Yep, I got the same result.  That's based on the `initSystem.php` failing to load a zc_plugin's class file:
```
    [89] => ##################################################################
    [90] => Action Point - 78
    [91] => Auto Type Method - processAutoTypeClass
    [92] => processing ZC class - C:/xampp/htdocs/zc210/includes/classes/observers/class.products_options_stock_observer.php
    [93] => loading ZC class - C:/xampp/htdocs/zc210/includes/classes/observers/class.products_options_stock_observer.php - FAILED
    [94] => =================================================================
    [95] => Auto Type Method - processAutoTypeClassInstantiate
    [96] => processing class instantiate - class = products_options_stock_observer object name = posObserver
    [97] => instantiating normal class - products_options_stock_observer as posObserver
    [98] => =================================================================
    [99] => ##################################################################
```

Without the changes to initSystem.php and FileSystem.php, that plugin-based class file **is** found: 
```
 [89] => ##################################################################
    [90] => Action Point - 78
    [91] => Auto Type Method - processAutoTypeClass
    [92] => processing class - C:/xampp/htdocs/zc210/zc_plugins/POSM/v6.1.0/catalog/includes/classes/observers/class.products_options_stock_observer.php
    [93] => loading class - C:/xampp/htdocs/zc210/zc_plugins/POSM/v6.1.0/catalog/includes/classes/observers/class.products_options_stock_observer.php - SUCCESS
    [94] => =================================================================
    [95] => Auto Type Method - processAutoTypeClassInstantiate
    [96] => processing class instantiate - class = products_options_stock_observer object name = posObserver
    [97] => instantiating normal class - products_options_stock_observer as posObserver
    [98] => =================================================================
```

includes/classes/FileSystem.php
mc12345678: One thing about changes and methods here is that in some cases it is not necessarily about obtaining a final appropriate string but that the provided string tells the story needed.

E.g. Instead of always trimming the final character off of `DIR_FS_ADMIN`, I've instead chosen to add a character to the string being used for comparison. In this way, if the path being tested were to include the admin foldername plus some, it wouldn't falsely result in true. 

Unfortunately, `is_file()` or `is_dir()` doesn't really work here because each only returns true if the item really exists on the path to detect if the provided path includes a file or is simply a path to a folder.

There's perhaps a number of things to be done to narrow down the condition; however, it also seems knowing whether it is/is not an actual folder or perhaps a file doesn't really matter. The goal is to identify if the provided information represents being in the applicable folder or not.

piloujp:
I am afraid that your last modification of code using `DIRECTORY_SEPARATOR` constant is not good. On Windows, only part of the path build with PHP constants like `__DIR__` or `__FILE__` will use system separator "\". 

`DIR_FS_ADMIN` which, in recent versions of ZC, is build using `__DIR__` is like this: `C:\path\to\ZC\root\admin/`
The trailing slash is added by some ZC installer script to PHP constant `__DIR__` (`C:\path\to\ZC\root\admin`) when in admin folder.

`DIR_FS_CATALOG` is defined in configure.php and is like this: `C:/path/to/ZC/root/`

Your code was OK without using `DIRECTORY_SEPARATOR` constant.
In related discussion thread, your logic and explanation was not taking in account Windows backslash. But anyway in actual ZC it concerns only admin path and where constant `__DIR__` is used which is only one place (after applying this PR) where there is no problem anyway and not related to this PR.

piloujp:
There is some part of code that seems inefficient to me, or I might be missing something.
In function `isAdminDir` I don't see the benefit of this code:
```
        if ($this->isAdminRootDir($filePath)) {
            return true;
        }
```
In case you are in admin root folder, it is one function call plus two condition tests (IF). Without this code, next IF test would get the same result in one test.
In case you are in an admin subfolder, then it is one function call plus three tests. Again, without this code, next IF test would get the same result in one test.

Same logic goes for function `isCatalogDir` calling `isCatalogRootDir` and `isCatalogRootDir` calling `isAdminDir`.

piloujp:
OK, I was missing something...
Still, this code could be removed and test change like this:
`if ($test !== $filePath || $test === '' || $test === '/') {`

mc12345678:
> I am afraid that your last modification of code using `DIRECTORY_SEPARATOR` constant is not good. On Windows, only part of the path build with PHP constants like `__DIR__` or `__FILE__` will use system separator "".

That's unfortunate that in coding to match the described/expected filename convention I have caused a problem.

> 
> `DIR_FS_ADMIN` which, in recent versions of ZC, is build using `__DIR__` is like this: `C:\path\to\ZC\root\admin/` The trailing slash is added by some ZC installer script to PHP constant `__DIR__` (`C:\path\to\ZC\root\admin`) when in admin folder.

This is not directly associated with "ZC installer script" but instead routine code in all such systems where ZC no longer requires coding the admin's folder path and has been similar to this since ZC 1.5.0.

Specifically (ZC 2.1.0 related): https://github.com/zencart/zencart/blob/ebc589515e448f67ed04a53d24134dc08bc8e91c/admin/includes/defined_paths.php#L47

Which contains:
````
if (!defined('DIR_FS_CATALOG')) define('DIR_FS_CATALOG', realpath(DIR_FS_ADMIN . '/../') . '/');
````

and would be "fixed" by the following:
````diff
- if (!defined('DIR_FS_CATALOG')) define('DIR_FS_CATALOG', realpath(DIR_FS_ADMIN . '/../') . '/');
+ if (!defined('DIR_FS_CATALOG')) define('DIR_FS_CATALOG', realpath(DIR_FS_ADMIN . '/../') . DIRECTORY_SEPARATOR);
````

> 
> `DIR_FS_CATALOG` is defined in configure.php and is like this: `C:/path/to/ZC/root/`
> 
> Your code was OK without using `DIRECTORY_SEPARATOR` constant. 

Here is where those supporting incorporation make their path forward determination. Unfortunately, the above change would have been better in transition from 1.5.x to 2.x as part of that significant change, but here we are. Alternate suggestions are to regress the new/revised code to another form (thereby hiding this issue to be found/corrected in a future attempt to further Windows compatibility): 

1. hard-code the forward slash in `FileSystem.php` as it was in the previous commit push, 
2. use some `real`xxxx command on both sides of the comparison(s) (allowing proper use of `DIRECTORY_SEPARATOR` but the `real`xxxx command to eventually be found redundant), 
3. some further unique evaluation on each use/run (even if just a one-time per execution run with temporary, held variable), 
4. etc...

> In related discussion thread, your logic and explanation was not taking in account Windows backslash. But anyway in actual ZC it concerns only admin path and where constant `__DIR__` is used which is only one place (after applying this PR) where there is no problem anyway and not related to this PR.

As to lack of Windows filenaming consideration, as suggested, history has been:

1. ZC is not designed to run on a Windows server, there are known incompatibilities, 
2. over the last few years use of `DIRECTORY_SEPARATOR` has been encouraged/advocated, 
3. there is other code that has been incorporated into recent ZC code specifically targeting Windows compatibility, 
4. I haven't found an online sandbox supporting Windows style PHP operations to test simple commands. (Welcome one be pointed out)
5. in the previous post there was no description of what assertion was in question/an issue and no demonstration of characteristics of consideration, so to align with previous direction it only appeared appropriate to incorporate code expected to be Windows compatible.

I stopped using Windows server style development many, many years ago when my suggested compatibility improvements were rejected, ZC code didn't completely/consistently work on a windows server and I had/made ready access to a *nix development platform.

PR #7035, opened 2025-03-23
tpl_image_additional.php: minor changes

torvista:
Invalid html: Replaced span as it was wrapping block-level divs.

Removed br from commented code:
before
add 1

after
add 2

Can squash this lot if preferred.


torvista:
bump

torvista:
Why is this still gathering dust here?
Don't you realise that if a non-event like this gets ignored, why would I bother to make any effort offering up the real improvements I have backed up?
Maybe that's the idea.

Plain-speaking english without insults would be appreciated, and I'll call it a day.

lat9:
> Why is this still gathering dust here? Don't you realise that if a non-event like this gets ignored, why would I bother to make any effort offering up the real improvements I have backed up? Maybe that's the idea.
> 
> Plain-speaking english without insults would be appreciated, and I'll call it a day.

FWIW, I'm deferring to one of the other maintainers that has been paying attention to responsive_classic.

drbyte:
> Why is this still gathering dust here?

The inconvenient truth is that the reason I repeatedly put this one off until later is because the PR only addressed one template, not both. And in the past there has regularly been contributor resistance to giving such feedback, and I didn't want to give feedback that would be perceived as insulting or nagging. 

So, knowing that I'd probably have to "just do it myself", and that the changes were mostly trivial, it was fine to just trivially leave it until a more convenient time.

Sorry if you don't like that explanation. But you did ask for plain english.

drbyte:
With the explanation out of the way, now I can say: Thanks for submitting this. It's cleaner. 👍

torvista:
My intention is always to equate things where duplicated in tempalates....but just to do one first to get that approved and not have to redo things in the others...so yes I'll mirror this in the others when I have time.

PR #7041, opened 2025-03-28
move order_total->process to pages and outside of template

proseLA:
persuant to this discussion #7038 (with myself), we should not be doing any processing in templates.  it makes little sense to me, in addition to screwing up totals for any payment modules.

there is a trend in payment modules to use javascript to isolate the store owner from touching any sensitive cardholder data.  in that regard, the payment module should NOT have to figure out what the order total is.  the payment module module needs to rely on the `$order->info['total']` to be correct at the time of the page rendering.

this PR only refactors when that happens.  it moves it from the template rendering to the page processing part of the page rendering.

here is [a forum posting](https://www.zen-cart.com/showthread.php?222449-Square-Payment-Module-for-Zen-Cart-Support-Thread&p=1406720#post1406720) that documents the issue.

in [a follow up post](https://www.zen-cart.com/showthread.php?222449-Square-Payment-Module-for-Zen-Cart-Support-Thread&p=1406822#post1406822), there are some screencasts that show the results.

### DIFF
--- a/includes/modules/pages/checkout_confirmation/header_php.php
+++ b/includes/modules/pages/checkout_confirmation/header_php.php
@@ -70,6 +70,9 @@ require(DIR_WS_CLASSES . 'order_total.php');
 $order_total_modules = new order_total;
 $order_total_modules->collect_posts();
 $order_total_modules->pre_confirmation_check();
+    if (MODULE_ORDER_TOTAL_INSTALLED) {
+        $order_totals = $order_total_modules->process();
+    }
 
 // load the selected payment module
 require(DIR_WS_CLASSES . 'payment.php');
--- a/includes/modules/pages/checkout_payment/header_php.php
+++ b/includes/modules/pages/checkout_payment/header_php.php
@@ -99,6 +99,9 @@ require(DIR_WS_CLASSES . 'order_total.php');
 $order_total_modules = new order_total;
 $order_total_modules->collect_posts();
 $order_total_modules->pre_confirmation_check();
+    if (MODULE_ORDER_TOTAL_INSTALLED) {
+        $order_totals = $order_total_modules->process();
+    }
 
 //  $_SESSION['comments'] = '';
 $comments = !empty($_SESSION['comments']) ? $_SESSION['comments'] : '';
--- a/includes/templates/responsive_classic/templates/tpl_checkout_confirmation_default.php
+++ b/includes/templates/responsive_classic/templates/tpl_checkout_confirmation_default.php
@@ -155,7 +155,7 @@
 
 <?php
   if (MODULE_ORDER_TOTAL_INSTALLED) {
-    $order_totals = $order_total_modules->process();
+//    $order_totals = $order_total_modules->process();
 ?>
 <div id="orderTotals"><?php $order_total_modules->output(); ?></div>
 <?php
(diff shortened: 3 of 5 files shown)


drbyte:
I have no objection to the `->process()` call being relocated from template file to module file.

One thing to test: is there any negative side-effect if the storeowner doesn't update their template file, and therefore ends up calling `->process()` twice?

proseLA:
in my experience i have to yet to do an upgrade that does not involve also upgrading/changing a template.

it is a grand goal to do an upgrade without the need to modify a template; unfortunately i have never come close to experiencing it.

drbyte:
> in my experience i have to yet to do an upgrade that does not involve also upgrading/changing a template.
> 
> it is a grand goal to do an upgrade without the need to modify a template; unfortunately i have never come close to experiencing it.

I completely agree. 

Unfortunately I've seen that many storeowners are scared of template updates and don't touch them.
Or worse, they use plugins that overwrite template files, and then re-introduce thing that were intended to be removed but the plugin doesn't keep step.

That's why I'm known for being reluctant to make big-impact template changes if they can be avoided.

Anyway, I'm not opposing this change; but I'm curious whether there's a way to minimize the risk of very bad negative side-effects.

drbyte:
I'm also curious "why" we originally put these calls in the template, when the module file already has so many calls to order-total "steps" ... in case there's a chicken/egg situation we're staring at here.

proseLA:
> I'm also curious "why" we originally put these calls in the template, when the module file already has so many calls to order-total "steps" ... in case there's a chicken/egg situation we're staring at here.

i have looked at this a bunch, and i have no good reasons.  

as far as history, this method call from the template goes back to the initial commit on github:

https://github.com/zencart/zencart/blame/1f1f2945358ed8b302ed6bf7d4e579e68d5bdd31/includes/templates/template_default/templates/tpl_checkout_payment_default.php#L40-L50

unfortunately i can not dig deeper than that.  

we start with this `pre_confirmation_check` whose sole job is to check is there are coupon or gift vouchers that cover the order total:

https://github.com/zencart/zencart/blob/355bf985a8c0ccda120baa19447425fade8c371c/includes/classes/order_total.php#L215-L242

during this process, it calls the order total `process` method for each enabled module.

and then resets the order info object.

but because we do not have the process called at the time i am proposing, we now have these 2 new notifiers in this method that were added in 22b6b04.  those were added for use here:

https://github.com/lat9/paypalr/blob/main/includes/classes/observers/auto.paypalrestful.php

and that seems to be [the recommendation for paypal users is to move to this new payment module.](https://docs.zen-cart.com/user/payment/paypal/#other-paypal-services:~:text=zen%20cart%20recommends%20using%20express%20checkout%20instead%20of%20paypal%20standard.%20another%20option%20is%20a%20newer%20module%2C%20called%20paypal%20restful%2C%20which%20will%20be%20available%20soon)

**now, my question is why would a payment module for a shopping cart require an observer?**  

and yet it requires 2 observers.

coding an integrating payment modules is hard enough.  now we need an observer to figure out the amount we need to charge the consumer?

i understand this is not an easy change, as it has potential negative side effects.  but how far down this path of making things more complex do we want to go?

---

in writing this comment (and looking at the code), i have come up with a far easier solution.  

what i want to do can be accomplished with:
```php
$calculatedTotal = $order_total_modules->pre_confirmation_check(true);
```
i see no negative side effects from calling `pre_confirmation_check` multiple times.

the only real potential downside i see with that, is that it would be nice to get the whole `$order->info` object back as opposed to just the order total.  with that whole object, one would have the actual tax and shipping data (level 2 data) which one could provide to the payment processor to get a lower interchange data.

i will close this and no pain, ie negative side effects, will be inflicted.

PR #7042, opened 2025-03-28
add third option for return of preconfirmation check

proseLA:
see discussion on #7041

### DIFF
--- a/includes/classes/order_total.php
+++ b/includes/classes/order_total.php
@@ -217,7 +217,7 @@ class order_total
     // true. This is used to bypass the payment method. In other words if the Gift Voucher is more than the order
     // total, we don't want to go to paypal etc.
     //
-    public function pre_confirmation_check(bool $returnOrderTotalOnly = false)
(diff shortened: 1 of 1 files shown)


lat9:
Would you provide an example of usage where the input parameter is a string?

proseLA:
```php
public function initPayment(bool $customerExists): Psr\Http\Message\ResponseInterface
        {
            global $order, $order_total_modules;
            $calculatedTotal = $order_total_modules->pre_confirmation_check('string');
            $body = [
                'paymentType' => 'preauth',
                'amount' => zen_round($calculatedTotal['total'], 2),
                'currency' => $calculatedTotal['currency'],
                'google-pay' => 1,
            ];
            if ($calculatedTotal['tax'] > 0) {
                $body['taxAmount'] = zen_round($calculatedTotal['tax'], 2);
            }
...
            return $this->guzzleCall('https://some.endpoint.cc', json_encode($body));
        }
```

lat9:
Much more elegant than my added notifications!

PR #7044, opened 2025-04-02
Refactor & deduplicate payment module instantiation in /admin/orders.php

proseLA:
building on 87a228c and #7004 and #6993, this refactor allows an encapsulated payment module to make use of the ZC common payment admin functions.  

it further reduces code duplication, while maintaining the same functionality.

the only functional change is that voids and refunds are ONLY recorded in the zen admin activity log if said function is executed by the module.  i see no reason to record that activity if the payment module does not contain the method.

that said, i do not feel strongly about when to record things in the admin activity log.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -107,6 +107,16 @@ switch (NOTIFY_CUSTOMER_DEFAULT) {
 
 if (!empty($action) && $order_exists === true) {
     $order = new order($oID);
+    $module = false;
+    global $installedPlugins;
+    $moduleFinder = new ModuleFinder('payment', new Filesystem());
+    $modules_found = $moduleFinder->findFromFilesystem($installedPlugins);
+    $payment = ($order->info['payment_module_code'] ?? 'nothing') . '.php';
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks for this.

admin/orders.php
```diff
@@ -107,6 +107,19 @@
...
     $order = new order($oID);
+    $module = false;
+    global $installedPlugins;
+    $moduleFinder = new ModuleFinder('payment', new Filesystem());
+    $modules_found = $moduleFinder->findFromFilesystem($installedPlugins);
+    $payment = ($order->info['payment_module_code'] ?? 'nothing') . '.php';
+    if (array_key_exists($payment, $modules_found) && file_exists(DIR_FS_CATALOG . $modules_found[$payment] . $payment)) {
+        $payment_module = DIR_FS_CATALOG . $modules_found[$payment] . $payment;
+        $languageLoader->loadModuleLanguageFile($payment, 'payment');
+        if (file_exists($payment_module)) {
```
drbyte, line 118: Am I correct that this `file_exists()` is redundant because it's already checked 3 lines earlier?

proseLA: yes.  you are correct.  i'm just the worst coder ever!

i will fix and force push.

PR #7053, opened 2025-04-12
Correct 'statecomma' handling in zen_address_format

### LINKED ISSUE #7052: Missing 'state' from `zen_address_label` addresses for some countries
Some address formats (like Canada, which uses address_format_id 1) use a `$statecomma` value in the formatting.  When such a country, like Canada, also uses zone-based states the `zen_address_label` formatting of any such address 'loses' the address' state:

Image

That's because `zen_address_format` sets the 'statecomma' value based on the incoming address' 'state' value which (for the CA case) isn't yet set.  See the following code section for more information:
https://github.com/zencart/zencart/blob/eda91b0eb2a88deb46e01afbb79df070bd610a6e/includes/functions/functions_addresses.php#L274-L294

lat9:
Fixes #7052

### DIFF
--- a/includes/functions/functions_addresses.php
+++ b/includes/functions/functions_addresses.php
@@ -276,7 +276,6 @@ function zen_address_format($address_format_id = 1, $incoming = array(), $html =
     $address['zip'] = $address['postcode'];
 
     $address['streets'] = !empty($address['suburb']) ? $address['street'] . $address['cr'] . $address['suburb'] : $address['street'];
-    $address['statecomma'] = !empty($address['state']) ? $address['state'] . ', ' : '';
 
     $country = '';
     if (!empty($incoming['country_id'])) {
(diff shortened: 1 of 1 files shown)


drbyte:
Do you have an "after" image? (I'm assuming the image in the Issue ticket is a "before" image.)
Can you set the "city" to something other than "Ontario"? (since Ontario is a Province (akin to a State), not a city.)

drbyte:
Part of my reason for asking is: the City or State should not be on the same line as the country.

eg: 
INVALID: 
123 Main St
Toronto, Ontario M1M1M1
~Ontario~ Canada

VALID:
123 Main St
Toronto, Ontario M1M 1M1
Canada

(or with M1M 1M1 on its own line)

lat9:
There you go, from the `address_book` page

Before:
image

After:
image

drbyte:
After digging ... actually, Canada uses Address Format 7 (same as USA), not 1. 

And, from what I can tell in the demo database, no countries use format 1.

Nevertheless, I agree with the proposed fix.
And, since it's probably reasonable to expect that format 1 is not normally used, there should be no backward compatibility breaks.

PR #7058, opened 2025-04-17
fix tax exempt customers to not get tax

proseLA:
this bit of code is not enough:

https://github.com/zencart/zencart/blob/02b6efdd442a0ba89d28511e90d62c42a5ac53c9/includes/functions/functions_taxes.php#L22-L27

most US clients show taxes separate from the line item and utilize what in ZC we call an `order_total` module.  without this PR, when a store is set up with wholesale customers being tax exempt, and taxes are not shown in the price, ie `(DISPLAY_PRICE_WITH_TAX !== 'true')`, the tax amount is not calculated as 0.

this PR addresses that issue.

### DIFF
--- a/includes/functions/functions_taxes.php
+++ b/includes/functions/functions_taxes.php
@@ -173,6 +173,11 @@ function zen_get_multiple_tax_rates($class_id, $country_id = -1, $zone_id = -1, 
     if (is_array($rates_array)) {
         return $rates_array;
     }
+    if (Customer::isTaxExempt() === true) {
+        return [
+            TAX_EXEMPT_DESCRIPTION => 0,
+        ];
+    }
 
     $rates_array = [];
 
@@ -434,6 +439,11 @@ function zen_get_all_tax_descriptions($country_id = -1, $zone_id = -1)
     );
     if (is_array($tax_descriptions)) {
         return $tax_descriptions;
+    }
+    if (Customer::isTaxExempt() === true) {
+        return [
+            TAX_EXEMPT_DESCRIPTION => 0,
+        ];
     }
 
     if ($country_id == -1 && $zone_id == -1) {
--- a/includes/languages/lang.english.php
+++ b/includes/languages/lang.english.php
@@ -367,6 +367,7 @@ $define = [
     'TABLE_HEADING_UPCOMING_PRODUCTS' => 'Upcoming Products',
     'TABLE_HEADING_WEIGHT' => 'Weight',
     'TABLE_HEADING_FEATURED_CATEGORIES' => 'Featured Categories',
+    'TAX_EXEMPT_DESCRIPTION' => 'Tax Exempt',
     'TEXT_ADMIN_DOWN_FOR_MAINTENANCE' => 'NOTICE: The website is currently down for maintenance to the public',
     'TEXT_ALL_CATEGORIES' => 'All Categories',
     'TEXT_ALL_MANUFACTURERS' => 'All Manufacturers',


includes/functions/functions_taxes.php
```diff
@@ -173,6 +173,12 @@ function zen_get_multiple_tax_rates($class_id, $country_id = -1, $zone_id = -1,
     if (is_array($rates_array)) {
         return $rates_array;
     }
+    // tax-exempt get no taxes
+    if (Customer::isTaxExempt() === true) {
+        return [
+            WHOLESALE_TAX_EXEMPT_DESCRIPTION => 0,
```
lat9, line 179: Since `Customer::isTaxExempt` can return `true` for reasons other than a wholesale customer, please create another `Customer` class static method to return the customer's tax-exempt description, e.g.
```php
    public static function getTaxExemptDescription(): string
    {
        $tax_exempt_description = WHOLESALE_TAX_EXEMPT_DESCRIPTION;

        global $zco_notifier;
        $zco_notifier->notify('NOTIFY_CUSTOMER_TAX_EXEMPT_DESCRIPTION', [], $tax_exempt_description);

        return $tax_exempt_description ;
    }
```

... and then use that here (on line 178-180):
```php
        return [
            Customer::getTaxExemptDescription() => 0,
        ];
```

includes/languages/lang.english.php
```diff
@@ -529,6 +529,7 @@
     'WARNING_NO_FILE_UPLOADED' => 'Warning:  no file uploaded.',
     'WARNING_PRODUCT_QUANTITY_ADJUSTED' => 'Quantity has been adjusted to what is in stock. ',
     'WARNING_SHOPPING_CART_COMBINED' => 'NOTICE: For your convenience, your current shopping cart has been combined with your shopping cart from your last visit. Please review your shopping cart before checking out.',
+    'WHOLESALE_TAX_EXEMPT_DESCRIPTION' => 'Wholesale Customer: Tax Exempt',
```
lat9, line 532: Since this could appear as a tax-description in an order's order-totals' display, please simplify to remove the intervening `:` since the display would wind up looking like
```
Wholesale Customer: Tax Exempt:  $0.00
```

I'll suggest something akin to `Tax Exempt (Wholesale)`.

lat9 (dismissed):

includes/functions/functions_taxes.php
proseLA: no.  

i am not sure what other reasons a customer may be tax exempt.

feel free to put your own name on a function and notifier for which i see no use.

includes/languages/lang.english.php
proseLA: i have renamed and removed colon.

you do not like my name, feel free to change it.

proseLA:
i have a very differing opinion about what a code review is.  if you want to play dictator, you will need to find a new fiefdom for which to get up their nose.

you want to ask questions, point out potential bugs in my code, great.  the rest of it is crap IMO.

i am not your underling.

9a47cf2 and #6017 have your name all over it.
 
and there is a bug.  a pretty obvious one.  your bug, not mine.

i found it; i fixed it; thanks are in order.

personally, i always start there.  makes for a better interaction.

includes/functions/functions_taxes.php
lat9: FWIW, I'll point out that TaxJar makes a distinction between wholesale tax-exempt and government tax-exempt ... but I can live with the simple generalization of the language constant.

PR #7063, opened 2025-04-25
Debug logs when a non- `product_info` type product is "not found"

scottcwilson:
```
[25-Apr-2025 10:50:54 UTC] Request URI: /zen/index.php?main_page=document_general_info&cPath=63&products_id=5132, IP address: **.***.***.***, Language id 1
#0 /home/client/public_html/zen/includes/modules/products_quantity_discounts.php(92): zen_debug_error_handler()
#1 /home/client/public_html/zen/includes/modules/pages/document_general_info/main_template_vars.php(139): require('/home/client/publi...')
#2 /home/client/public_html/zen/includes/templates/responsive_classic/common/tpl_main_page.php(181): require('/home/client/publi...')
#3 /home/client/public_html/zen/index.php(94): require('/home/client/publi...')
--> PHP Warning: Undefined variable $products_discount_type_from in /home/client/public_html/zen/includes/modules/products_quantity_discounts.php on line 92.
```

This is happening because a bunch of things are being done even if product not found.  Compare to the product_info `main_template_vars.php` and you'll see that the close bracket for product found was set too early in this file.

### DIFF
--- a/includes/modules/pages/document_general_info/main_template_vars.php
+++ b/includes/modules/pages/document_general_info/main_template_vars.php
@@ -93,53 +93,52 @@
 
     $products_discount_type = $product_info->fields['products_discount_type'];
     $products_discount_type_from = $product_info->fields['products_discount_type_from'];
+
+    require(DIR_WS_MODULES . zen_get_module_directory('product_prev_next.php'));
+  
+    $module_show_categories = PRODUCT_INFO_CATEGORIES;
+    $module_next_previous = PRODUCT_INFO_PREVIOUS_NEXT;
+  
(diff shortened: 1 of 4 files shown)


scottcwilson:
If you use "hide whitespace" on the diff on the files tab, you can see the close bracket was moved from line 96 to 141.

lat9:
Note that the same issue arise for the `document_product_info`, `product_free_shipping_info` and `product_music_info` pages.  Please apply the same correction to each of those pages `main_template_vars.php` as well.

scottcwilson:
good catch

PR #7073, opened 2025-05-02
Correct "greedy" search via `get_template_dir`

lat9:
The `$template->template_dir` method calls `$pageLoader->getTemplateDirectory` which uses a `regex` search to locate a collection of template files.

Since the current regex on the search strings when loading named CSS, JS and PHP files in a template's `/css` and/or `/jscript` sub-directories is 'loose' this results in false-positive location of files when using a template's page-override feature:
https://github.com/zencart/zencart/blob/02b6efdd442a0ba89d28511e90d62c42a5ac53c9/includes/templates/template_default/common/tpl_main_page.php#L7-L10

This unwanted behavior can be viewed most easily via a page-override using the `bootstrap` template.  Simply create the sub-directory `/includes/templates/bootstrap/privacy` (or your clone's name) and add a file with a `.php` extension in that directory.  You'll note that the coloring of the `privacy` page goes wonky since the `css/stylesheet_zca_colors.php` file isn't loaded.

This PR provides more specific search parameters for the `template_dir` requests for the common CSS/JS loading.

### DIFF
--- a/includes/templates/template_default/common/html_header_css_loader.php
+++ b/includes/templates/template_default/common/html_header_css_loader.php
@@ -16,9 +16,9 @@ if (!defined('IS_ADMIN_FLAG')) {
 /**
  * load all template-specific stylesheets, named like "style*.css", alphabetically
  */
-$directory_array = $template->get_template_part($template->get_template_dir('.css', DIR_WS_TEMPLATE, $current_page_base, 'css'), '/^style/', '.css');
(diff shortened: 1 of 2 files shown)


scottcwilson:
Thanks for the work on this, @lat9 and @proseLA!

scottcwilson:
This might be worth a writeup in the Known Bugs thread: 
https://www.zen-cart.com/showthread.php?230454-Known-Bugs-in-2-1-0

PR #7081, opened 2025-05-06
Remove comma from figure in gv

scottcwilson:
```
[05-May-2025 16:01:31 America/Los_Angeles] PHP Fatal error:  Uncaught TypeError: number_format(): Argument #1 ($num) must be of type int|float, string given in /home/client/public_html/includes/modules/order_total/ot_gv.php:125
Stack trace:
#0 /home/client/public_html/includes/modules/order_total/ot_gv.php(125): number_format('1,011.94', 2)
#1 /home/client/public_html/includes/classes/order_total.php(75): ot_gv->__construct()
#2 /home/client/public_html/includes/modules/pages/checkout_payment/header_php.php(117): order_total->__construct()
#3 /home/client/public_html/index.php(35): require('/home/client/p...')
#4 {main}
  thrown in /home/client/public_html/includes/modules/order_total/ot_gv.php on line 125
```


scottcwilson:
Yep, you're right.  Code in haste, repent at leisure.

drbyte:
The tests are failing because the GV amounts being output into the OT module fields are being stripped of the expected comma separator used in the specified currency.

lat9:
Perhaps, rather that patching the issue in `ot_gv.php`, a change should be made to the currencies' class' `normalizeValue` method:
https://github.com/zencart/zencart/blob/cd24471b65981010a37ca9c4964dff03428a45aa/includes/classes/currencies.php#L126-L130

Instead of just 'converting the decimal-point to a '.' to also change the thousands' separator to an empty string.  Note that some currencies might use the same value for both, so the decimal-point change should be the last occurrence and the thousands' separator any remaining.

scottcwilson:
TY - like the idea, just need to get some time to explore.

lat9:
> TY - like the idea, just need to get some time to explore.

No problem; I'll note that the same amount using Costa Rican Colones (CRC) would be displayed as '1`.`011`,`94'

PR #7085, opened 2025-05-07
Additional error-logging updates

lat9:
Consider this as an extension to PR #6944 and issue #6942.

1. Remove the no-longer-used 'report-all-errors' init_includes files and their auto-loaders.
2. Recognize that `E_USER_ERROR` is to be deprecated in PHP 8.4.
3. Default logging-levels to `E_ALL` -- was removing `E_NOTICE`.


scottcwilson:
Note that for 2.2.0, the configs REPORT_ALL_ERRORS_ADMIN, REPORT_ALL_ERRORS_STORE and REPORT_ALL_ERRORS_NOTICE_BACKTRACE have already been removed.

PR #7088, opened 2025-05-10
Correct multiple loads of a module's language file

### LINKED ISSUE #7087: 2.1.0 problem with language loader....
i have some payment modules that extend other payment modules.  as such they require the language file of the payment file being extended.

given that situation i have encountered what i would call a bug.  and i am curious if the following is intended behavior?

create a file called `test.php` in one's store root with the following code:

```php
<?php

    $test = $languageLoader->loadModuleLanguageFile('moneyorder.php', 'payment');
    $test2 = $languageLoader->loadModuleLanguageFile('moneyorder.php', 'payment');

```
`$test` is `true`; while `$test2` is `false`.

i find that odd, as this behavior will return an error in the `messageStack` that says the language file can not be found.  even though the language file clearly is there.

is this intended behavior?

lat9:
Fixes #7087


proseLA:
well, at this point, I just do not know.

- I still think this is a bug.
- this PR does address the bug.
- BUT, I think it is adding more code on top of other code.
- there already is a method to determine if a file is loaded:

https://github.com/zencart/zencart/blob/2dc6aaa3463227ab1f1e4075aa71bf70774fd524/includes/classes/ResourceLoaders/LanguageLoader.php#L110

but coming up with the `$defineFile` is another whole big thing (at least for me, as i do not see an easy way to it with the existing code).

the question to me is whether it is worth adding this code on top of all of this other code.

and the next question would be whether to leave the existing code, move the `array_merge` above the existing code, and then remove the `return false`.

but at this point, I have stared at it enough and I’m not sure what the best answer is.

lat9:
> well, at this point, I just do not know.
> 
>     * I still think this is a bug.
> 
>     * this PR does address the bug.
> 
>     * BUT, I think it is adding more code on top of other code.
> 
>     * there already is a method to determine if a file is loaded:
> 
> 
> https://github.com/zencart/zencart/blob/2dc6aaa3463227ab1f1e4075aa71bf70774fd524/includes/classes/ResourceLoaders/LanguageLoader.php#L110
> 
> but coming up with the `$defineFile` is another whole big thing (at least for me, as i do not see an easy way to it with the existing code).
> 
> the question to me is whether it is worth adding this code on top of all of this other code.
> 
> and the next question would be whether to leave the existing code, move the `array_merge` above the existing code, and then remove the `return false`.
> 
> but at this point, I have stared at it enough and I’m not sure what the best answer is.

Right, there's not a good way to use the existing `isFileAlreadyLoaded` method, since `loadModuleLanguageFile` has no clue as to where the language file might have been loaded (template-override, different languages, zc_plugins); that's why I chose to just add that code to `loadModuleLanguageFile` since it needs to return an indication as to whether/not a language file for a module was found.

drbyte:
In `FilesLanguageLoader`, is there an important reason why we're returning `false` when `isFileAlreadyLoaded` returns `true`?

https://github.com/zencart/zencart/blob/e924713e98474240beacb5b6957149f5238d70e2/includes/classes/ResourceLoaders/FilesLanguageLoader.php#L46:L48

It would seem that the `loadModuleLanguageFile` method just above is returning `true` if a template-specific override is found and loaded, but `false` if no override is found but the base file is previously loaded (or `true` as a fallback). That seems inconsistent, and not intuitive. And IMO is what's causing the anomaly in the reported issue.

Perhaps just returning true there would suffice? What would the side-effect be?

PR #7100, opened 2025-05-14
Handle reviews with no text (only stars)

scottcwilson:


lat9:
Is this due to the database _somehow_ missing an associated entry in the `reviews_description` table?  Possibly due to an older Zen Cart version?

scottcwilson:
it is an upgrade from a much older version; perhaps the template allowed a no-text entry at the time.

PR #7111, opened 2025-05-19
Add multilanguage support for tax rates descriptions

piloujp:
This PR add the possibility to have multilingual description for any tax rate set in admin. This description is visible on invoices with tax details.
The principle for that is similar to order status one. Different languages versions are saved in a separate table `tax_rates_description` and linked to `tax_rates` table through `language_id` and `taxe_rates_id` parameters.
Update by admin user is made through actual admin tax rates setting page, depending on which language is used at this time.
This new table is although updated automatically when a language is added or deleted.


drbyte:
@piloujp I think the index on the tax_description field is unnecessary.
Lookups are always done based on language_id and tax_rate_id, so the index is moot.
And, unless there are thousands of records, indexing on such a tiny table has very little benefit.

The technical reason for my concern is utf8mb4 vs a field length of more than 191 chars, since max index byte length is 768 in many implementations of MySQL (and mb4 means 4 byte, so 4x191=764).  4x250 (or 4x255 earlier schema) is too long.

So, either the index needs to be on at most 191 chars, not 250. Or the extra index needs to be deleted.
The table already has both a primary key and a unique-index which doubles as a lookup index for the essential criteria.

Any objection to deleting that line?
`KEY idx_tax_rates_description_zen (tax_description(250))`

PR #7117, opened 2025-05-28
Add missing storefront DATE_TIME_FORMAT definition

lat9:
Needed for storefront calls to `zen_datetime_short`; reported [here](https://www.zen-cart.com/showthread.php?230744-shipstation-error&p=1407656#post1407656) on the Zen Cart forums.

### DIFF
--- a/includes/languages/lang.english.php
+++ b/includes/languages/lang.english.php
@@ -94,6 +94,7 @@ $define = [
     'CHARSET' => 'utf-8',
     'DATE_FORMAT' => 'm/d/Y',
     'DATE_FORMAT_LONG' => '%A %d %B, %Y',
+    'DATE_TIME_FORMAT' => '%m/%d/%Y %H:%M:%S',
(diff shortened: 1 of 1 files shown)


lat9:
Who knows what other plugins are out "in the wild" that make use of the storefront call to `zen_datetime_short`.  Shipstation could be the tip of the iceberg.

lat9:
In what way does this 'break' the shipstation update?  Did you choose to `define` that constant instead of using the `zen_define_default`?

drbyte:
While I agree that the correct solution is for the end-user to update to the updated plugin.

That said, I think the plugin should define its own independent (ideally non-conflicting/non-competing) date constants because those are specific to the shipstation's API's dependency requirements. It never should have relied on a ZC-provided output format.
Strictly speaking, since ZC has abandoned that constant, the shipstation module is free to use it ... but I'd suggest it be renamed, or just set to a variable used only by the module.

As such, closing this.

PR #7121, opened 2025-05-31
Correcting coupon tax calculation

lat9:
As reported on the Zen Cart forum's [Bug Reports](https://www.zen-cart.com/showthread.php?230751-Discount-Coupon-Tax-issue&p=1407710#post1407710) by ryk.

### DIFF
--- a/includes/modules/order_total/ot_coupon.php
+++ b/includes/modules/order_total/ot_coupon.php
@@ -594,7 +594,7 @@ class ot_coupon extends base
                             $od_amount['tax_groups'][$key] = zen_round($this_tax * $ratio, $currencyDecimalPlaces);
                             $od_amount['tax'] += $od_amount['tax_groups'][$key];
                         }
-                        if (DISPLAY_PRICE_WITH_TAX == 'true' && $coupon_details['coupon_type'] == 'F') {
+                        if (DISPLAY_PRICE_WITH_TAX === 'true' && $coupon_details['coupon_type'] === 'F' && $this->include_tax === 'true') {
                             $od_amount['total'] += $od_amount['tax'];
                         }
                         break;
@@ -680,7 +680,7 @@ class ot_coupon extends base
                 $orderTotalTax -= $order->info['shipping_tax'] ?? 0;
             }
         }
-        if (DISPLAY_PRICE_WITH_TAX !== 'true') {
+        if (DISPLAY_PRICE_WITH_TAX !== 'true' || $this->include_tax !== 'true') {
             $orderTotal -= $orderTotalTax;
         }
 


piloujp:
This correction only works if:
Include Tax is set to true and Re-calculate Tax is set to None
or
Include Tax is set to false and Re-calculate Tax is set to Standard
Any other combination and tax calculation is wrong.

lat9:
> This correction only works if: Include Tax is set to true and Re-calculate Tax is set to None or Include Tax is set to false and Re-calculate Tax is set to Standard Any other combination and tax calculation is wrong.

Could you provide a bit more detail besides the calculation is "wrong"?

piloujp:
ZC settings:
Display Prices with Tax -> true
Sales Tax Display Status -> 1 (On)
Tax 10%
Coupon 10%
Shipping with no tax

One Die Hard With A Vengeance $39.99(excl.) $43.99(incl.) in cart

Depending on the options set in Discount Coupon module, here are some results:
Module Discount Coupon setting:
Include Tax -> false
Re-calculate Tax -> Standard
    Sub-Total: $43.99
    Store Pickup: $2.50
    Discount Coupon: Coupon10 : -$4.00
    **CA TAX: $3.60**
    Total: $42.49

Module Discount Coupon setting:
Include Tax -> false
Re-calculate Tax -> None
    Sub-Total: $43.99
    Store Pickup: $2.50
    Discount Coupon: Coupon10 : -$4.00
    **CA TAX: $4.00**
   Total: $42.49

Only tax change between these two settings. You got the same with next two settings:

Module Discount Coupon setting:
Include Tax -> true
Re-calculate Tax -> None
    Sub-Total: $43.99
    Store Pickup: $2.50
    Discount Coupon: Coupon10 : -$4.40
    **CA TAX: $4.00**
    Total: $42.09

Module Discount Coupon setting:
Include Tax -> true
Re-calculate Tax -> Standard  _There is a message in ZC that this combination shoulf not be done_
    Sub-Total: $43.99
    Store Pickup: $2.50
    Discount Coupon: Coupon10 : -$4.40
    **CA TAX: $3.60**
    Total: $42.09

Finally, a bonus bug when choosing store pickup and everything display and is calculated without tax, although it is not the setting :

Module Discount Coupon setting:
Include Tax -> false
Re-calculate Tax -> Standard
    Sub-Total: $39.99
    Store Pickup: $0.00
    Discount Coupon: Coupon10 : -$4.00
    **Sales Tax + CA TAX: $0.00   !!!!!!!!!!!!!**
    Total: $35.99

piloujp:
I think the tax calculation problem might have nothing to do with this PR, well not directly, and was there before.
What I can remember from the time already spent on this, is that the 3 options Include Tax, Re-calculate Tax and Tax Class, or in fact their logic was the problem. Something like half the possible combinations were leading to impossible situations. Fixing one case always ends up with troubles in another one...

lat9:
Withdrawn.

PR #7123, opened 2025-06-11
Adding multibyte support using mb strings functions

### LINKED ISSUE #7098: Multibyte characters cut in half
In all Zen cart versions.

String functions do not take in account multibyte characters. With standard PHP string functions, mainly counting strings' number of characters and extracting sub-strings, fails when the string contains multibyte characters.

In ZC, the most obvious consequence is in products listing where products descriptions are cut shorter than they should and sometimes a character itself is cut, resulting in the display of a weird character `�` at the end of the description's extract.

It is not new and in Japanese Language pack, to deal with it, I have to modify core files.
I realized only recently that the same problem occurs with any language that use non ASCII characters, which is nearly all languages other than English.
In non-alphabetic languages like Japanese, it occurs all the time, but with languages like French where only few of these non-ASCII characters are used, it is not so visible. For example, products descriptions are shortened only a little bit and if the last character from the extract is an ASCII one, then no problem.

Just for information, French letter like `éèêùô` generally count for 2 ASCII characters. Japanese ones will count for 3.

Replacing PHP functions` strlen` or `substr` by their equivalent multibyte version in "strategic points" in ZC like in `functions_strings.php` will fix this problem.

piloujp:
This PR is a possible improvement on multibyte characters support.
Fixes #7098   
It replaces some `strlen()` and `substr()` functions by their equivalent in `mbstring` extension, where they might have to deal with multibyte characters,


haredo:
@piloujp 
I have notice there are 2 strlen left on this php page instead of mb_strlen is this an error?
line 57 if (strlen($dob) >10 || zen_valid_date($dob) === false) {
line 95 if (strlen($telephone) < ENTRY_TELEPHONE_MIN_LENGTH) {

PR #7131, opened 2025-06-23
Correct attribs' qty-discount check

### LINKED ISSUE #7130: product_info/attributes: module code comparisons not working as intended
Due to this commit:
https://github.com/zencart/zencart/commit/a8aaa9a1dfcda83e0fb5a28e56ff1bb0b532c769

Pre commit:

Image

Post commit:

Image

Here
https://github.com/zencart/zencart/blob/a8aaa9a1dfcda83e0fb5a28e56ff1bb0b532c769/includes/modules/attributes.php#L209-L212
if the array values are null, the test passes and the flags are set to true.

lat9:
Fixes #7130

### DIFF
--- a/includes/modules/attributes.php
+++ b/includes/modules/attributes.php
@@ -206,7 +206,7 @@ foreach ($products_options_names as $next_option_name) {
                 $price_onetime = '';
             }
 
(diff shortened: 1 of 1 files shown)


torvista (approved):
ok for me, I did the same solution.

PR #7132, opened 2025-06-24
Remove password-reset-tokens when customer deleted

lat9:
Remove any records in the `customer_password_reset_tokens` table when a customer is deleted.


torvista:
I assume this means that there will be orphan entries for previously-deleted customers....perhaps should be cleaned up on an upgrade?

PR #7135, opened 2025-06-25
Fix situation where attributes don't exist

scottcwilson:
If the zen_products_attributes table has an entry with options_id = 0 and options_values_id = 0 (or any other non-existent values), logs are produced on the storefront side (although not on the admin side).  Checking the attribute query fixes this.

Prevents these two logs: 

```
[23-Jun-2025 00:17:05 UTC] Request URI: /index.php?main_page=shopping_cart, IP address: 134.238.195.201, Language id 1
#0 /home/client/public_html/shop/includes/modules/pages/shopping_cart/header_php.php(79): zen_debug_error_handler()
#1 /home/client/public_html/shop/index.php(35): require('/home/client/...')
--> PHP Warning: Uninitialized string offset 0 in /home/client/public_html/shop/includes/modules/pages/shopping_cart/header_php.php on line 79.

[23-Jun-2025 00:17:05 UTC] Request URI: /index.php?main_page=shopping_cart, IP address: 134.238.195.201, Language id 1
#0 /home/client/public_html/shop/includes/modules/pages/shopping_cart/header_php.php(80): zen_debug_error_handler()
#1 /home/client/public_html/shop/index.php(35): require('/home/client/...')
--> PHP Warning: Uninitialized string offset 0 in /home/client/public_html/shop/includes/modules/pages/shopping_cart/header_php.php on line 80.

[23-Jun-2025 00:17:05 UTC] Request URI: /index.php?main_page=shopping_cart, IP address: 134.238.195.201, Language id 1
#0 /home/client/public_html/shop/includes/modules/pages/shopping_cart/header_php.php(86): zen_debug_error_handler()
#1 /home/client/public_html/shop/index.php(35): require('/home/client/...')
--> PHP Warning: Undefined array key "products_options_name" in /home/client/public_html/shop/includes/modules/pages/shopping_cart/header_php.php on line 86.

[23-Jun-2025 00:17:05 UTC] Request URI: /index.php?main_page=shopping_cart, IP address: 134.238.195.201, Language id 1
#0 /home/client/public_html/shop/includes/modules/pages/shopping_cart/header_php.php(89): zen_debug_error_handler()
#1 /home/client/public_html/shop/index.php(35): require('/home/client/...')
--> PHP Warning: Undefined array key "options_values_price" in /home/client/public_html/shop/includes/modules/pages/shopping_cart/header_php.php on line 89.

[...]

PR #7136, opened 2025-06-25
Correcting storefront logs when an option is missing from the db

lat9:
An alternate solution to #7135.

Correcting the issue at the source, essentially ensuring that any attribute-related function request validates that the 'options_id' associated with a `products_attributes` record exists in the `products_options` table.

Using `INNER JOIN` instead of `LEFT JOIN` to assure the above.


scottcwilson:
Works when tested with my database.  TY!

PR #7142, opened 2025-06-28
Prevent log during who's online

scottcwilson:
```
[28-Jun-2025 12:06:18 UTC] PHP Fatal error:  Uncaught Error: Undefined constant "ERROR_PRODUCT_OPTION_SELECTION" in /Users/scott/Sites/gh_demo_200/includes/classes/shopping_cart.php:1098
Stack trace:
#0 /Users/scott/Sites/gh_demo_200/includes/classes/shopping_cart.php(1393): shoppingCart->attributes_price('1:edff669e5da95...')
#1 /Users/scott/Sites/gh_demo_200/admin/includes/classes/WhosOnline.php(361): shoppingCart->get_products()
#2 /Users/scott/Sites/gh_demo_200/admin/includes/classes/WhosOnline.php(249): WhosOnline->inspectSessionCart('', 'securityToken|s...')
#3 /Users/scott/Sites/gh_demo_200/admin/includes/classes/WhosOnline.php(157): WhosOnline->getStatusCode(Array)
#4 /Users/scott/Sites/gh_demo_200/admin/whos_online.php(52): WhosOnline->retrieve('', '423a3f87ea98313...', true, false)
#5 /Users/scott/Sites/gh_demo_200/admin/index.php(16): require('/Users/scott/Si...')
#6 {main}
  thrown in /Users/scott/Sites/gh_demo_200/includes/classes/shopping_cart.php on line 1098
```

### DIFF
--- a/admin/includes/languages/lang.english.php
+++ b/admin/includes/languages/lang.english.php
@@ -235,6 +235,7 @@ $define = [
     'ERROR_PASSWORDS_NOT_MATCHING' => 'Password and confirmation must match',
     'ERROR_PASSWORD_RULES' => 'Passwords must contain both letters and numbers, must be at least %s characters long, and must not be the same as the last 4 passwords used. Passwords expire every 90 days, after which you will be prompted to choose a new password.',
     'ERROR_PAYMENT_MODULES_NOT_DEFINED' => 'NOTE: You have no payment modules activated. Please go to Modules->Payment to configure.',
+    'ERROR_PRODUCT_OPTION_SELECTION' => '<br> ... Invalid Option Values Selected ',
     'ERROR_SHIPPING_CONFIGURATION' => '<strong>Shipping Configuration errors!</strong>',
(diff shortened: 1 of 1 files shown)


lat9:
I guess my question would be "How did a product with a display-only attribute get added to the cart in the first place?"

scottcwilson:
The only thing I can think of is that the attribute was changed to read-only after the item was added to the cart.  That's the only way I could dup it.   But it's in the code, so I guess it's something someone thought should be checked.   It's just that the check fails when it's done from admin in who's online.

lat9:
> The only thing I can think of is that the attribute was changed to read-only after the item was added to the cart. That's the only way I could dup it. But it's in the code, so I guess it's something someone thought should be checked. It's just that the check fails when it's done from admin in who's online.

OK, makes some sense!  Is there a reason for not putting the language constant into lang.whos_online.php?

scottcwilson:
This spot could be hit by a plugin place since it involves price calculation.   I'd just put it here since there's where it is on the storefront side.

lat9:
OK, then a minor update, please, to put that constant into its alphabetic ordering like the rest of the language-constants in that file.

PR #7144, opened 2025-06-29
Refinements to JS hiding of state input field on country change

### LINKED ISSUE #7143: Misleading Zone Pulldown Required Indicator
**Describe the bug**

This is a minor issue but has bugged me for some time. When the create account State field is set to display as a pulldown when possible (ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN set to true in admin) and the selected country has zones, the "required" indicator is shown. But if the country is changed to one without zones, the required indicator remains, even if the minimum length of the state field entry is set to zero. The same thing happens when the pulldown is not displayed with the placeholder "required" indicator. This behavior is confusing to foreign visitors from countries without states or provinces who want to create an account on a site where the default country has zones.

When the pulldown is displayed, the problem is in function update_zones in includes/templates/template_default/jscript/zen_addr_pulldowns.php, specifically in two lines near the end of the file (111 and 115 in zc1.5.8a):
`jQuery('#stateZone').next('span.alert').show();`
and
`jQuery('#stateZone').next('span.alert').hide();`

The selectors for the lines that show or hide the required indicator [jQuery('#stateZone').next('span.alert')] do not exist. The correct element id is #zoneLabel for only those two lines. #stateZone is correct for other lines in the function. The code should be:
`jQuery('#zoneLabel').next('span.alert').show();`
and
`jQuery('**#zoneLabel**').next('span.alert').hide();`

I have not investigated a solution for when the pulldown is not displayed, nor addressed the issue of minimum state length. I have just set the minimum state length to zero.

**Version**
Zen Cart version: 1.5.8a
PHP version: 8.2.26
MySQL version: 8.0.40
What older/newer versions were also tested? None, but 2.1.0 has the same two lines with the incorrect element id.

**To Reproduce**
1. Set the State field to display a pulldown if possible in the admin
2. on the store side, click login, then create account (New? Please provide your billing information in an unmodified zen cart)
3. Scroll down to the state pulldown. Observe the required indicator next to State/Province
4. Change the country to any that do not have zones, such as United Kingdom
5. Notice the required indicator remains next to State/Province

**Expected behavior**

I expect the "required" indicator to appear only if the selected country has zones and not be displayed if the country has no zones. Further, the minimum state length criteria should only apply if the selected country has zones.

**Screenshots**

**Debug/Error Logs**

No error logs are generated
 
 - Operating System: Mac OS 15.5
 - Browser: Firefox 140.0.2, Chrome 138.0.7204.50 

**Additional context**

drbyte:
Might fix #7143

### DIFF
--- a/includes/templates/template_default/jscript/zen_addr_pulldowns.php
+++ b/includes/templates/template_default/jscript/zen_addr_pulldowns.php
@@ -76,10 +76,12 @@ jQuery(document).ready(function() {
     }
     if (jQuery('#stateZone > option').length > 1) {
         jQuery('#state').hide();
+        jQuery('#stateLabel').hide();
         jQuery('#stateZone').show();
         jQuery('#stateZone').next('span.alert').show();
     } else {
         jQuery('#state').show();
+        jQuery('#stateLabel').show();
         jQuery('#stateZone').hide();
         jQuery('#stateZone').next('span.alert').hide();
     }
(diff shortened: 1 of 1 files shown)


drbyte:
Works okay with BS4 template; but RC template may need to adapt for a spacer as well. 

Template updates (incl BS4) may be required to set the actual non-pulldown State field label.

lat9:
I believe that this PR can be closed, given @piloujp's work on those other state-field PRs/issues.

PR #7145, opened 2025-06-30
Function zen_break_string use

### LINKED ISSUE #7113: Function zen_break_string use in ZC
Function `zen_break_string($string, $len, $break_char = '-')` is a function that will cut long words (longer than `$len`) and add a break character (default `-`) like you will do when at the end of a line, but a word is too long and has to be written on two lines.
In ZC, this function is used only 3 times and I think none of this use is a proper one.

First in `\admin\reviews.php` (line 210), where there seems to be a swap with function `zen_trunc_string` that shorten a string to create an extract of it.
On the admin page to approve reviews, if the review is very long, it is displayed in full, even continuing after the next field (rating), which does not look nice. Plus, it will break any word longer than 15 characters, adding a `-` each time.

It is although used in two templates files that are used in catalog to display reviews:
`includes/templates/responsive_classic/templates/tpl_product_reviews_info_default.php` on line 53 and
`includes/templates/template_default/templates/tpl_product_reviews_info_default.php` on line 52.
In both cases, string length is set to 60. I don't know any word that long in English, I suppose it is more to avoid code injection?
Anyway, it will cause problems in any language that don't use spaces (that are considered words separator). Adding that these languages use multibyte characters, and you can end up with a break character `-` every 20 characters...

Is this function really necessary/useful?

piloujp:
Fixes #7113 
Removes use of function zen_break_string in templates reviews' display that leads to random line breaks when multibyte characters are used in languages with no spaces between words.

### DIFF
--- a/admin/reviews.php
+++ b/admin/reviews.php
@@ -207,7 +207,7 @@ if (!empty($action)) {
                 <p class="control-label"><?php echo ENTRY_REVIEW; ?></p>
               </div>
               <div class="col-sm-9 col-md-6">
-                <span class="form-control" style="border:none; -webkit-box-shadow: none"><?php echo nl2br(zen_output_string_protected(zen_break_string($rInfo->reviews_text, 15))); ?></span>
+                <span class="form-control" style="border:none; -webkit-box-shadow: none"><?php echo nl2br(zen_output_string_protected(zen_trunc_string($rInfo->reviews_text, 15))); ?></span>
               </div>
             </div>
             <div class="form-group">
--- a/includes/templates/responsive_classic/templates/tpl_product_reviews_info_default.php
+++ b/includes/templates/responsive_classic/templates/tpl_product_reviews_info_default.php
@@ -12,7 +12,7 @@
 
 <?php
   if (!empty($products_image)) {
-   	/**
+    /**
      * require the image display code
      */
 ?>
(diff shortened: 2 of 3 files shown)


piloujp:
> @piloujp Can you explore whether adding this CSS will have negative effects in your multibyte reviews examples? It's needed in english if someone posts a review that has no-word-break-text longer than the display-width of the div.
> 
> ```css
> #reviewsInfoDefault .reviews-wrapper {
>   overflow-wrap: break-word;
> }
> ```

@drbyte There is no problem with css overflow wrapper in Japanese. Japanese characters will always stay inside the wrapper as all characters can be break one by one even in a word. It is true with all possible parameters for overflow-wrap.
No problem at all.

PHP problem is due to the fact that standard string functions count bytes, not characters.

drbyte:
@piloujp I don't think I said well what I meant. 

When I added a very long string of text, it overflows the div invisibly.
But it wraps properly with overflow-wrap: break-word; or overflow-wrap: any

I wanted to confirm whether Japanese will benefit from this CSS directive, or if it will cause a complication, in this situation. If it causes no problem, that's good.

Before/After examples:
Screen Shot 2025-07-05 at 5 07 11 PM
Screen Shot 2025-07-05 at 5 07 52 PM

drbyte:
Thanks. CSS updated

PR #7146, opened 2025-06-30
Correcting paypaldp/paypalwpp payment capture

lat9:
Introduced in [this](https://github.com/zencart/zencart/commit/1ad558b9134f1d7f96fcd5e141ccbf6721fe3d72) commit post zc200


scottcwilson:
Thanks for catching!

PR #7155, opened 2025-07-16
Unnecessary Multibyte function provoking bug

piloujp:
A bug was introduced in PR #7123 in file admin/define_page_editor.php line 72 where using mb_strlen() function gives wrong results as the variable is coming from POST data and is encoded differently from a standard PHP multibyte string.
Just replacing back `mb_strlen()` by `strlen()` function fixes the problem.

### DIFF
--- a/admin/define_pages_editor.php
+++ b/admin/define_pages_editor.php
@@ -69,7 +69,7 @@ switch ($_GET['action']) {
         @rename($file, 'bak' . $file);
         $new_file = fopen($file, 'w');
         $file_contents = stripslashes($_POST['file_contents']);
-        fwrite($new_file, $file_contents, mb_strlen($file_contents));
+        fwrite($new_file, $file_contents, strlen($file_contents));
         fclose($new_file);
       }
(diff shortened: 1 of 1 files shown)


lat9:
@piloujp, could you give an example of some POST data that displays this bug?

piloujp:
Just add some Japanese text like below in a define page using TinyEditor and save. It can't be read again by TinyEditor and in ZC catalog, the text is shortened a lot and ends with a question mark on a black diamond background.

東京に拠点を置くショップで、ドローンやヘリコプターなどの無線操縦飛行機械を専門としていますが、飛行機、グライダー、パラグライダーも取り扱っています。

I think the problem is due to the POST data having a special invisible character at the end of the string. This character is ignored by strlen() but taken in account by mb_strlen() which mess up the multibyte string.
Plus, it seems POST data are already treated/formated (encoded?) at byte level and don't need use of PHP multibyte functions.

dennisns7d:
Noting that the third argument to fwrite is the length of the data stream in bytes, not characters.
https://www.php.net/manual/en/function.fwrite.php

PR #7157, opened 2025-07-22
function zen_lookup_admin_menu_language_override update

piloujp:
Add slash character use in names for configuration groups titles and plugins title/description.
Actually, admin configuration::Shipping / Packaging menu is not translated because the language override constant name is not properly recognized due to the slash (/) character.

This PR updates function zen_lookup_admin_menu_language_override to recognize slashes in group titles and replace them by underscore in constants names.
To be coherent, characters space, hyphen and slash are replaced by underscore in language constants name for configuration groups titles, plugins names and description.

### DIFF
--- a/includes/functions/functions_lookups.php
+++ b/includes/functions/functions_lookups.php
@@ -311,20 +311,20 @@ function zen_lookup_admin_menu_language_override(string $lookup_type, ?string $l
             break;
         case 'configuration_group_title':
             $str = $lookup_key;
-            $str = preg_replace('/[\s ]+/', '_', $str);
+            $str = preg_replace('/[\s -\/]+/', '_', $str);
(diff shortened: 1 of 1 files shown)


includes/functions/functions_lookups.php
```diff
@@ -311,20 +311,20 @@ function zen_lookup_admin_menu_language_override(string $lookup_type, ?string $l
             break;
         case 'configuration_group_title':
             $str = $lookup_key;
-            $str = preg_replace('/[\s ]+/', '_', $str);
```
drbyte, line 314: Note to self: The `[\s ]` pattern was introduced in #6584 with v2.1.0
So backward-compatibility of including the `-` is only impacted between v2.1.0 and v2.2.0

PR #7158, opened 2025-07-22
Yearly sales report display bug when more than 5 years of reports

piloujp:
Yearly sales report display bug fix to display year last 2 digits only when there are more than 5 years of reports.

When there are more than five years of report, year's first digit only (2 for 2025) is displayed instead of last 2 digits (25 for 2025).
This due to some bad `substring` parameter, '`0,1`' that should be '`2,2`'.
You need more than five years of data for testing. If not, you can change test on line 113 from '`> 5`' to a lower number.


drbyte:
Note to self: the `2,2` is correct because `date('Y')` is 4 digits.
```php
    case self::YEARLY_VIEW:
          $this->info[$i]['text'] = date('Y', $this->startDates[$i]);
```

PR #7159, opened 2025-07-24
tax_rates_description table update query failing

piloujp:
When updating database to v2.2.0, the MySQL query to populate new tax_rates_description table after its creation is returning an error due to wrong number of column inserted.
This query needs to be adapted to the new auto increment index added after some modifications.


piloujp:
@drbyte Did you resolve the compatibility problem between the SQL query and installer?
I had this to submit, a query that works with the installer, but you had already merged:
```
INSERT INTO project_version_history (project_version_key, project_version_major, project_version_minor, project_version_patch, project_version_date_applied, project_version_comment)
SELECT project_version_key, project_version_major, project_version_minor, project_version_patch1 as project_version_patch, project_version_date_applied, project_version_comment
FROM project_version;
```

PR #7167, opened 2025-07-30
Message stack reformat and sync function signatures

drbyte:
Upcoming updates for PayPal, which uses messageStack (by extending the class for visual control), needs the `output()` method to be compatible with both admin and catalog versions (because webhook handlers only have catalog awareness).

This update, apart from 2 reformatting/refactoring commits, has a commit that simply syncs the `output()` method signature for compatibility.

This is NOT bringing exact parity between the admin/catalog classes, because "use" of the messageStack features are quite different between catalog/admin implementations.

### DIFF
--- a/admin/includes/classes/message_stack.php
+++ b/admin/includes/classes/message_stack.php
@@ -67,7 +67,7 @@ class messageStack extends boxTableBlock
         $this->size = 0;
     }
 
-    public function output(): string
+    public function output(string $class='')
     {
         $this->table_data_parameters = 'class="messageBox"';
         return $this->tableBlock($this->errors);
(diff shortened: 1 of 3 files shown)


drbyte:
Note: `zca_messageStack.php` should be updated, in the Bootstrap template.

```diff
-    function add($class, $message, $type = 'error') 
+    function add(string $class, string $message, string $type = 'error') 
```

drbyte:
> Any suggestions on how to make that compatible with prior versions of Zen Cart where the type-hinting isn't included? 

Two approaches we could take:
- These changes are backward-compatible to v1.5.8, so could just have v158 users grab this updated messageStack.php class and carry on
- Or we could skip the `string` type declarations on the `add()` function in this PR, relaxing the strictness for broader compatibility.

drbyte:
Sure, I'll update the PR.

PR #7168, opened 2025-07-31
Modifying currencies table to support four letter codes

retched:
Some currency services (ie, OpenExchangeRates) have support for four-digit currency codes—namely CryptoCurrencies like DOGE (DogeCoin), DASH, others.

Some currencies are longer, but those are mostly "Black Market Rates" not widely in circulation or use.

### DIFF
--- a/zc_install/sql/updates/mysql_upgrade_zencart_220.sql
+++ b/zc_install/sql/updates/mysql_upgrade_zencart_220.sql
@@ -47,7 +47,8 @@ CREATE TABLE customer_password_reset_tokens (
 ALTER TABLE orders_products_attributes MODIFY products_options varchar(191) NOT NULL default '';
 ALTER TABLE products_options MODIFY products_options_name varchar(191) NOT NULL default '';
 ALTER TABLE products_options_values MODIFY products_options_values_name varchar(191) NOT NULL default '';
-
+ALTER TABLE currencies MODIFY code char(4) NOT NULL default '';
+ALTER TABLE orders MODIFY currency char(4) default NULL;
 
 #PROGRESS_FEEDBACK:!TEXT=Updating configuration settings...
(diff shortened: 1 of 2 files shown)


zc_install/sql/updates/mysql_upgrade_zencart_220.sql
```diff
@@ -47,7 +47,7 @@ CREATE TABLE customer_password_reset_tokens (
 ALTER TABLE orders_products_attributes MODIFY products_options varchar(191) NOT NULL default '';
 ALTER TABLE products_options MODIFY products_options_name varchar(191) NOT NULL default '';
 ALTER TABLE products_options_values MODIFY products_options_values_name varchar(191) NOT NULL default '';
-
+ALTER TABLE currencies MODIFY code varchar(4) NOT NULL default '';
```
lat9: Shouldn't this be simply `char`, like the install-SQL?  There should be a similar `ALTER` for orders::currency`.

retched: YES. I thought I included that...

retched: Also I left it as `VARCHAR()`  as that was what the original column was. I only incresed the size to fit four characters instead.

drbyte: > Also I left it as VARCHAR() as that was what the original column was.

Perhaps in your own db you changed them to varchar, but in the original install SQL they're `char` (see the diff)
So, in the upgrade, both `modify` statements should use `char`

retched: That may have been the case. That or I defaulted to VARCHAR() instead. No worries, just changed the ALTER lines back to char(4) instead.

drbyte: 👍

PR #7171, opened 2025-08-08
Strftime_parameter_mismatch_in zcDate.php

### LINKED ISSUE #6840: Lots of numbers display in admin date in place of timezone when PHP extension 'intl' is not installed.
If you deactivate PHP extension `intl`, date in admin page will display lots of numbers instead of timezone.

This is due to `zcDate` conversion that replace `%Z` by `ZZZZ` on line 71 of `zcDate.php` file. For `date()` function, `Z` is actually the timezone shift in seconds. There is an identical problem on line 70 with `%z`, although it is not used in standard Zen Cart.
Perhaps a better conversion would be:
```
'%z' => 'O',
'%Z' => 'T',
```

Conversion from `strftime` to `intl` format is not exact too on lines 118 and 119. I suppose it is a trick to display GMT in front of timezone shift, which is not possible with `strftime` parameters alone.

piloujp:
Corrects conversion of some `strftime` parameters to date() and `intlDate` format parameters.
Fixes #6840  and adds parameter `%k` conversion, useful for some plugins.

### DIFF
--- a/admin/includes/languages/lang.english.php
+++ b/admin/includes/languages/lang.english.php
@@ -9,7 +9,7 @@
 @setlocale(LC_TIME, ['en_US', 'en_US.utf8', 'en', 'English_United States.1252']);
 
 $define = [
-    'ADMIN_NAV_DATE_TIME_FORMAT' => '%A %d %b %Y %X (%Z)',
+    'ADMIN_NAV_DATE_TIME_FORMAT' => '%A %d %b %Y %X (%z)',
     'ARIA_PAGINATION_' => '',
     'ARIA_PAGINATION_CURRENTLY_ON' => ', now on page %s',
     'ARIA_PAGINATION_CURRENT_PAGE' => 'Current Page',
--- a/includes/classes/zcDate.php
+++ b/includes/classes/zcDate.php
@@ -59,6 +59,7 @@ class zcDate extends base
             '%B' => 'F',
             '%d' => 'd',
             '%H' => 'H',
+            '%k' => 'G',
             '%m' => 'm',
             '%M' => 'i',
             '%S' => 's',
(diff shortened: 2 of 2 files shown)


includes/classes/zcDate.php
```diff
@@ -116,7 +118,7 @@ protected function initializeConversionArrays()
                 '%y' => 'yy',
                 '%Y' => 'y',
                 '%z' => 'ZZZZ',
-                '%Z' => 'ZZZZ',
+                '%Z' => 'zzzz',
```
drbyte, line 125: Most of the time the pattern is to map lowercase to lowercase and uppercase to uppercase. 
Was the opposite intentional here?

piloujp: This is probably the origin of the mismatch.
For `strftime`:
`%z` --> The time zone offset. Example: -0500 for US Eastern Time
`%Z` --> The time zone abbreviation. Example: EST for Eastern Time
For `intlDateFormat`:
`zzzz` --> The time zone long specific non-location format. Example: Pacific Daylight Time (falls back to PDT)
`ZZZZ` --> The time zone long localized GMT format. Example: GMT-8:00

Correspondence is not perfect but close enough, Perhaps, instead of `zzzz`, `z` would be enough, anyway when there is no long format for `zzzz`, it falls back to `z`.
Here are useful links if you want to check:
[strftime format](https://www.php.net/manual/en/function.strftime.php).
[IntlDafFormat](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table).

drbyte: Right. I see what you mean.

admin/includes/languages/lang.english.php
```diff
@@ -9,7 +9,7 @@
 @setlocale(LC_TIME, ['en_US', 'en_US.utf8', 'en', 'English_United States.1252']);
 
 $define = [
-    'ADMIN_NAV_DATE_TIME_FORMAT' => '%A %d %b %Y %X (%Z)',
+    'ADMIN_NAV_DATE_TIME_FORMAT' => '%A %d %b %Y %X (%z)',
```
drbyte, line 12: I think this is the part that I worry about, just for upgraders. It's one more thing to have to touch that looks either insignificant or is really confusing. Just something they have to think about whether they agree with the change or not.
But, there are enough changes to this file in the upgrade that I guess it's negligible.

PR #7172, opened 2025-08-08
Updating category image to be proportionally fit

retched:
Added the "object-fit-contain" class to the category icon display in collect_info.php


retched:
I completely forgot about making that change to the other collect_info.php files (I thought they were all harmonized.) But in running 2.1.0 with an uneven image (sized 223 x 100). In looking at that image, in the "collect_info.php" backend, I noticed that the image was stretched horizontally upward to fill the square in. I thought it too would be properly fit but evidently I missed something?

PR #7173, opened 2025-08-09
Fix Move Category and some incomplete cPaths

torvista:
Offered as a starting point or interim measure.

Corrects the error in the move_category action.

But it highlighted processes where a complete cPath: 1_3_33 becomes a truncated cPath 33.
Some have been corrected.

I gave up when I found that move_product also drops the full cPath, and correcting that becomes non-trivial with cascading edits required that possibly require a review/different approach to ensuring a full cPath is always maintained.

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -208,7 +208,7 @@ if (!empty($action)) {
             zen_update_products_price_sorter($product_id);
           }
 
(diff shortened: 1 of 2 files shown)


drbyte:
Okay, I see where you were going with this. But yes, there are some wrinkles that need sorting.

Adding it to my list...

PR #7175, opened 2025-08-12
Admin: Group modules by status and sort by sort_order

drbyte:
This makes it faster to distinguish amongst enabled vs disabled payment modules, and shows them in the order the customer will encounter them.

OLD:
Screen Shot 2025-08-11 at 10 45 39 PM
NEW:
Screen Shot 2025-08-11 at 10 46 20 PM

--

OLD:
Screen Shot 2025-08-11 at 10 45 52 PM
NEW:
Screen Shot 2025-08-11 at 10 46 38 PM

--
OLD:
Screen Shot 2025-08-11 at 10 46 01 PM
NEW:
Screen Shot 2025-08-11 at 10 46 46 PM


admin/modules.php
```diff
@@ -413,7 +450,7 @@
                 '<button type="submit" class="btn btn-danger" id="saveButton">' .
                     IMAGE_UPDATE .
                 '</button>&nbsp;' .
-                '<a href="' . zen_href_link(FILENAME_MODULES, 'set=' . $set . ($_GET['module'] != '' ? '&module=' . $_GET['module'] : ''), 'SSL') . '" class="btn btn-default" role="button" id="cancelButton">' .
+                '<a href="' . zen_href_link(FILENAME_MODULES, 'set=' . $set . ($_GET['module'] !== '' ? '&module=' . $_GET['module'] : ''), 'SSL') . '" class="btn btn-default" role="button" id="cancelButton">' .
```
scottcwilson, line 453: Should this be `!empty($_GET['module'])` ?

drbyte: Perhaps. I only touched that line to use strict comparison.

drbyte:
Updated screenshot showing fix to columns being same width in both top and bottom tables:
Screen Shot 2025-08-12 at 1 01 07 AM

admin/modules.php
lat9: There are, unfortunately, a slew of spots within `modules.php` where `$_GET['module']` is _presumed_ to be present.

drbyte: Are you saying `!== ''` should be reverted to `!= ''`? 
I don't see the difference, but I can do that.

lat9: No, that was my misread of the code for page accesses (like on the initial entry) where the 'module' variable isn't part of the link.

lat9:
Where do modules that are installed but not enabled (yellow status on Sort Order) fit into the display?

lat9:
> Where do modules that are installed but not enabled (yellow status on Sort Order) fit into the display?

Ah, I see,  They're grouped with the "Available" modules.
image

lat9 (approved):

PR #7182, opened 2025-08-15
Multiple format for available date

piloujp:
Actual ZC needs products' available/expire date to be in the format `YYYY-mm-dd`. This PR allows any date format to be used, or more precisely, any date format set by constant `DATE_FORMAT_DATE_PICKER` in `lang.english.php` file.
This applies to start and end dates for specials, featured products and featured categories too.


admin/includes/modules/preview_info.php
```diff
@@ -22,6 +22,14 @@
       $url = str_replace(['http://', 'https://'], '', $url);
   }
   unset ($url);
+  $p_date_available = $pInfo->products_date_available;
+  if (DATE_FORMAT_DATE_PICKER !== 'yy-mm-dd' && !empty($pInfo->products_date_available)) {
```
drbyte, line 26: Is the `!== 'yy-mm-dd'` because of a default db date format?

admin/includes/languages/lang.english.php
```diff
@@ -147,7 +147,7 @@
     'CONFIGURATION_MENU_ENTRIES_TO_SORT_BY_NAME' => '1',
     'CONNECTION_TYPE_UNKNOWN' => '\'%1$s\' is not a valid connection type for generating URLs' . PHP_EOL . '%2$s' . PHP_EOL,
     'DATE_FORMAT' => 'm/d/Y',
-    'DATE_FORMAT_DATE_PICKER' => 'yy-mm-dd',
+    'DATE_FORMAT_DATE_PICKER' => 'mm-dd-yy',
```
drbyte: I'm trying to understand this change.

admin/featured.php
```diff
@@ -213,7 +213,8 @@
           $form_action = 'update';
 
           $product = $db->Execute("SELECT p.products_id, p.products_model, pd.products_name, p.products_price, p.products_priced_by_attribute,
-                                          f.expires_date, f.featured_date_available
+                                          date_format(f.expires_date, '" .  zen_datepicker_format_forsql() . "') as expires_date,
```
drbyte: In SQL queries, we're trying to use UPPERCASE for SQL keywords like `DATE_FORMAT` and `AS`.

admin/includes/modules/preview_info.php
drbyte: Oh. I think I get it: this converts it TO y/m/d if not already in that format, to match MySQL standard date storage?

piloujp: Yes exactly.
This code already existed in ZC somewhere else.

admin/includes/languages/lang.english.php
piloujp: This is just to display in the standard English (US) date format, but any format is OK here.

piloujp: In French you would have `dd-mm-yy`.

admin/featured.php
piloujp: This is normally what I do too...

piloujp: Done!

admin/includes/languages/lang.english.php
drbyte: A downside of making this change is that the "empty date" check, which is hard-coded to yyyy-mm-dd format, won't blank out a zero/empty date.

eg: `($sInfo->specials_date_available == '0001-01-01')`

Therefore when DATE_FORMAT_DATE_PICKER is changed from yy-mm-dd to (for example) mm-dd-yy, the display changes from:
yy-mm-dd
to:
mm-dd-yy

drbyte: Let's revert this line back to yy-mm-dd
Then we can merge this PR, addressing what this PR is about.

The empty/zero-date issue is separate.

piloujp: I missed that! As you say, it is a different problem. Just reverting this line is OK.

PR #7185, opened 2025-08-16
Fixes #7174 Notifier bypassed in ot_group_pricing

### LINKED ISSUE #7174: Notifier bypassed in ot_group_pricing
**Describe the bug**
Notifier NOTIFY_OT_GROUP_PRICING_DEDUCTION_OVERRIDE_FINAL in includes/modules/order_total/ot_group_pricing.php is bypassed if Standard Re-calculate Tax option is set and tax is zero.

**Version**
Zen Cart version: 2.1.0
PHP version: 8.2
MySQL version: 8.0
What older/newer versions were also tested? None

**To Reproduce**
1. Set up observer for the above-mentioned notifier
2. Set up pricing group in admin Customers | Group Pricing
3. Set Re-calculate tax option to Standard (the default) in admin Modules | Order Total | Group Discount
4. Checkout with product
5. Note that observer is not executed

**Expected behavior**
The observer should be executed.

**Screenshots**
Not applicable

**Debug/Error Logs**
None generated
 
 - Operating System: Mac OS 15.6
 - Browser [e.g. chrome, firefox] and version number: Firefox v141.0.3

**Additional context**
The cause for this behavior is the "return" located on line 175 of the file mentioned above. Replacing "return" with "break" should solve the problem. I want the observer to correct the discount (and tax) for any products that should not be discounted for the pricing group, and to tailor the discount title and description on the checkout page to the specific pricing group (I have more than one pricing group), and do it without modifying core code.

daphilli224:
Replaces return statement with break so notifier is reached if tax is zero and tax recalculate option is set to Standard (the default).

Fixes #7174

### DIFF
--- a/includes/modules/order_total/ot_group_pricing.php
+++ b/includes/modules/order_total/ot_group_pricing.php
@@ -172,7 +172,7 @@ class ot_group_pricing {
             $ratio = 1;
           }
           $adjustedTax = $orderTotalTax * $ratio;
-          if ($order->info['tax'] == 0) return $od_amount;
+          if ($order->info['tax'] == 0) break;
           $ratioTax = ($orderTotalTax != 0 ) ? $adjustedTax/$orderTotalTax : 0;
           $tax_deduct = 0;
           foreach ($taxGroups as $key=>$value) {


drbyte (approved):
Thanks!

PR #7186, opened 2025-08-17
Special/featured active & expires empty dates displaying as 0001-01-01…

piloujp:
… when DATE_FORMAT_DATE_PICKER is not set to `'yy-mm-dd'` (like `'dd-mm-yy'` or `'mm-dd-yy'`).
When no date is set in form for Specials, featured products and featured categories, to display an empty field instead of the value saved in database (0001-01-01), this PR replaces database value by an empty string as soon as it is red from DB and do the opposite in case of an update just before writing to DB. All tests in between for this string (0001-01-01) have been removed too.

Complementary to #7182 .


drbyte:
@piloujp Thanks for this.

While testing I discovered 2 bugs in products_price_manager related to this, specifically when datepicker constant is not yy-mm-dd:
1. (I fixed this) - string `'null'` was being set when empty date submitted for features/specials, causing a SQL error.
2. When editing a Featured start/end date, if I set either to blank, it blanks out both. But only when not yy-mm-dd. Still need to figure out why.
3.

PR #7190, opened 2025-08-20
TinyMCE localization using community language packs

### LINKED ISSUE #7188: TinyMCE editor language files not loading
**Describe the bug**
When using TinyMCE in admin pages in another language than English, TinyMCE menu should change to that language, but it is always in English.

**Version**
Zen Cart version: 2.2.0-alpha and precedent ones
PHP version: 8.3~8.4
MySQL version: 8.4.6
Plugins that may be related: none
What older/newer versions were also tested? ZC v2.1.0

**To Reproduce**
Editing a define page with TinyMCE editor in ZC admin.

**Debug/Error Logs**
None, but browser debug shows an error 404 for the link to the localized JavaScript file.
 
 - Operating System: Windows
 - Browser Firefox 141.0.3 

It seems there are 2 problems. First, the language code is used to create the link, but TinyMCE actually uses locale code for file language file naming. For example, the French version of the file is `fr_FR.js` but ZC creates a link with `fr.js`.
Second problem is, even if correcting file name, the language file does not seem to exist on TinyMCE CDN.

piloujp:
Corrects language codes to fit TinyMCE ones.
Add TinyMCE languages pack for common Zen Cart languages.
Fixes #7188

### DIFF
--- a/admin/includes/tinymce.php
+++ b/admin/includes/tinymce.php
@@ -83,8 +83,8 @@ if (strtoupper(TINYMCE_EDITOR_API_KEY) === 'GPL' || empty(TINYMCE_EDITOR_API_KEY
 
 // Language Support Setup
 $lng ??= new language;
-$localesDirectory = $editor_assets_path . '/langs';
-$tinyLanguagesUrl = $editor_assets_url . '/langs';
+$localesDirectory = $editor_assets_path . 'langs';
+$tinyLanguagesUrl = $editor_assets_url . 'langs';
 $tinyLanguageCode = $_SESSION['languages_code'];
 $tinyLanguageFiles = (new FileSystem)->listFilesFromDirectory($localesDirectory, '~^([a-z]{2})([-_][A-Z]{2,4})?\.js$~', false);
 foreach ($tinyLanguageFiles as $key => $tinyLanguageFile) {


admin/includes/tinymce.php
```diff
@@ -123,12 +123,27 @@ function tinymce_config_setup() {
 
         }
         let languagesConfig = {
-            language: '<?= zen_output_string_protected($_SESSION['languages_code']) ?>',
+            <?php
+            $lancode = ['bg' => 'bg_BG', 'bn' => 'bn_BD', 'fr' => 'fr_FR', 'he' => 'he_IL', 'hu' => 'hu_HU', 'is' => 'is_IS', 'ko' => 'ko_KR', 'nb' => 'nb_NO', 'pt' => 'pt_PT', 'sl' => 'sl_SL', 'sv' => 'sv_SE', 'th' => 'yh_TH', 'zh' => 'zh_CN'];
```
drbyte: I'd like us to use something like `$localesArray` instead of `$lancode`.

admin/includes/tinymce.php
```diff
@@ -123,12 +123,27 @@ function tinymce_config_setup() {
...
+            language_load: false,
+            language: '<?= zen_output_string_protected($tinylangcode) ?>',
             content_langs: [
 
-            <?php
+            <?php // For spellchecker language with paid premium account
             foreach ($lng->get_languages_by_code() as $lang) {
-                echo "    { title: '" . zen_output_string_protected($lang['name']) . "', code: '" . zen_output_string_protected($lang['code']) . "' },\n";
+                if (array_key_exists($lang['code'], $lancode)) {
+                    $spelchecklangcode = $lancode[$lang['code']];
```
drbyte: Can we change spel to spell

admin/includes/tinymce.php
drbyte: Aside: I notice that for future Tiny v8 they're changing to the newer locale standard of `-` separators instead of `_` separators, but safely falling back to `_` if `-` not found.
We're not integrating v8 here yet, but that'll be a distinct incompatibility since we're hardcoding things here.

I suppose we could scan the filesystem to look for installed lang files and then pick the best match found.

piloujp: I was aware of this problem. As long as file names are not updated, it should be OK, but after that...
Your idea of checking for files sounds good. I will look at it and come back to you.

piloujp:
> I've committed some updates to use complete filesystem path (to prevent malicious manipulation), and built-in directory scanning methods. Also outputting all admin languages in content_langs. They're only used in spellcheck (paid) mode, but are probably useful to be able to access in that case.

Code looks good and perform nearly as good. There is a small bug in the regex which eliminates files with only 2 letters code like ja.js. After correcting this, all is good.

admin/includes/tinymce.php
```diff
@@ -81,6 +80,19 @@ function zenGetLatestTinyMceReleaseTag() {
 } else {
     $editor_js_src = "https://cdn.tiny.cloud/1/" . TINYMCE_EDITOR_API_KEY . "/tinymce/7/tinymce.min.js";
 }
+
+// Language Support Setup
+$lng ??= new language;
+$localesDirectory = $editor_assets_path . '/langs';
+$tinyLanguagesUrl = $editor_assets_url . '/langs';
+$tinyLanguageCode = $_SESSION['languages_code'];
+$tinyLanguageFiles = (new FileSystem)->listFilesFromDirectory($localesDirectory, '~^[a-zA-Z]{2}[-_]?[a-zA-Z]{0,4}\.js$~i', false);
```
drbyte: Good point. 
I wonder if this is an even better regex? `([a-zA-Z]{2})([-_][a-zA-Z]{2,4})?\.js`
... "2-letters", optionally followed by "- or _, and then 2-4 letters"

piloujp: Yes, it is better like this, it eliminates few cases that would not correspond to a valid code like `frFR` or `fr-`.

piloujp: The `i` at the end could also be eliminated and regex rewrite like this: `'~^([a-z]{2})([-_][A-Z]{2,4})?\.js$~'`.
The RFC whatever that defines these codes specifies that language code should always be lower case and country/region codes upper case.

drbyte: Ya, the official distributed files will match the pattern.

PR #7200, opened 2025-08-23
Add notifier ZEN_UPDATE_ORDERS_HISTORY_AFTER_INSERT

drbyte:
... which can be observed by PayPal module if needed

### DIFF
--- a/includes/functions/functions_osh_update.php
+++ b/includes/functions/functions_osh_update.php
@@ -97,6 +97,15 @@ function zen_update_orders_history($orders_id, $message = '', $updated_by = null
                   LIMIT 1"
             );
 
+            // PayPal Trans ID, if any
+            $paypalLookup = $GLOBALS['db']->Execute(
+                "SELECT *
+                 FROM " . TABLE_PAYPAL . "
+                 WHERE order_id = $orders_id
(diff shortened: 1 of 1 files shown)


includes/functions/functions_osh_update.php
```diff
@@ -97,6 +97,15 @@ function zen_update_orders_history($orders_id, $message = '', $updated_by = null
                   LIMIT 1"
             );
 
+            // PayPal Trans ID, if any
+            $result = $GLOBALS['db']->Execute(
+                "SELECT txn_id, parent_txn_id
```
lat9: Perhaps, `SELECT *` to grab all fields for the matching record for future use?

drbyte: Sure. Also: I limited it to the first record. Is there any reason to need "all" records? I'm thinking not. Just looking for second opinion.

lat9: I don't see any reason for *all* records, but the order-by clause is picking the most recent not the first record.

drbyte: Hmmm ... testing this with PayPal reveals that it's pulling the "REFUND" record, not the "CAPTURE" record, so it can't link up properly for the tracking id. Tracking must link to a capture record whose status is 'COMPLETED".

I guess passing "all" records would allow a re-ordering and selective picking.

PR #7203, opened 2025-08-26
Customer login/password-forgotten throttling

lat9:
Extend the `password_forgotten` throttling to the `login` page, using a common (and override-able) maximum number of failures value.

The session-based count increments each time either a `login` page error is detected or an unknown email_address is submitted to the `password_forgotten` page.  If that session-based count exceeds the maximum value above, both pages' form-processing respond with an HTTP 406 response code until the session times out.

The `password_forgotten` page is also updated to throttle the sending of the password-reset-link email.  If a password-reset email has been previously sent for an email address and the link's still valid, simply display the "check your email" message without resending that email.  This makes it more difficult for a site to reach its per-hour-email-limit in the presence of repeated bashing on the page.

### DIFF
--- a/includes/classes/Customer.php
+++ b/includes/classes/Customer.php
@@ -127,10 +127,7 @@ class Customer extends base
     {
         global $db;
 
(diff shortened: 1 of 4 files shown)


scottcwilson:
Great idea!

PR #7204, opened 2025-08-27
Products to Categories: remove p inside span and remove use of document.getElementById .innerHTML

torvista:


admin/products_to_categories.php
```diff
@@ -644,10 +644,10 @@
                         $selected_categories[] = (int)$product_linked_category['categories_id'];
                     }
                     ?>
-                    <span id="toggleCheckbox"></span><?php // placeholder for toggle checkbox: no content when javascript disabled ?>
+                    <p id="toggleCheckbox">
+                        <label><input type="checkbox" onClick="toggle(this)"> <?= TEXT_LABEL_SELECT_ALL_OR_NONE ?></label>
```
lat9: HTML attributes should be lowercase, i.e. `onclick` instead of `onClick`.

PR #7208, opened 2025-08-28
Plugin title & description language constants

piloujp:
Actually, when a plugin is not installed or is deactivated, its name and description are always in English whatever the admin UI language settings. This PR adds a function in `PluginManager.php` that retrieve these two constants from the language file that is normally loaded when a plugin is activated and allows language override.
There is one constraint, the file name must follow this pattern; `lang.uniquekey.php`, where 'unique key' corresponds to the plugin unique key/ folder name and is case-sensitive.
In case the file is not found or there is no defined constant in it, nothing happens and plugin information is displayed is in English.
This code was tested with 19 plugins installed and 3 languages. As long as the file name is correct, plugin information are displayed in the language set in admin.

### DIFF
--- a/includes/classes/PluginManager.php
+++ b/includes/classes/PluginManager.php
@@ -315,7 +315,7 @@ class PluginManager
     protected function loadPluginLanguageConstants($pluginpath)
     {
         $pluginpath = str_replace('\\', '/', $pluginpath);
-        if (!is_dir($pluginpath) || !str_contains($pluginpath, DIR_FS_CATALOG . 'zc_plugins/')) {
+        if (!is_dir($pluginpath) || !str_starts_with($pluginpath, DIR_FS_CATALOG . 'zc_plugins/')) {
             return;
         }
         $filePath = [];
--- a/includes/classes/PluginManager.php
+++ b/includes/classes/PluginManager.php
@@ -192,7 +192,9 @@ class PluginManager
             }
             $manifest = require $fileinfo->getPathname() . '/manifest.php';
             $versionList[$fileinfo->getFilename()] = $manifest;
-            $this->loadPluginLanguageConstants($fileinfo);
+            if ($_SESSION['languages_code'] !== 'en') {
+                $this->loadPluginLanguageConstants($fileinfo->getPathname());
+            }
         }
         return $versionList;
     }
@@ -310,24 +312,16 @@ class PluginManager
         return $this->pluginControl;
     }
 
-    protected function loadPluginLanguageConstants($parent)
+    protected function loadPluginLanguageConstants($pluginpath)
     {
         $filePath = [];
         foreach ($this->getInstalledPlugins() as $plugin) {
             $filePath[] = DIR_FS_CATALOG . 'zc_plugins/' . $plugin['unique_key'] . '/' . $plugin['version'] . '/';
         }
-
-        $dir = new \DirectoryIterator($parent->getPathName());
-        foreach ($dir as $fileinfo) {
-            $univ_path_name = rtrim(str_replace('\\', '/', $fileinfo->getPathname()), '.');
-            if ($fileinfo == '.' && !in_array($univ_path_name, $filePath)) {
-                $lang_file = $univ_path_name . '/lang.manifest_' . $_SESSION['languages_code'] . '.php';
-                if (file_exists($lang_file)) {
-                    include $lang_file;
-                    foreach ($define as $key => $value) {
-                        define($key, $value);
-                    }
-                }
+        if (!in_array($pluginpath, $filePath)) {
+            $lang_file = $pluginpath . '/lang.manifest_' . $_SESSION['languages_code'] . '.php';
+            if (file_exists($lang_file)) {
+                require_once $lang_file;
             }
         }
     }
--- a/includes/classes/PluginManager.php
+++ b/includes/classes/PluginManager.php
@@ -314,14 +314,29 @@ class PluginManager
 
     protected function loadPluginLanguageConstants($pluginpath)
     {
+        $pluginpath = str_replace('\\', '/', $pluginpath);
+        if (!is_dir($pluginpath) || !str_contains($pluginpath, DIR_FS_CATALOG . 'zc_plugins/')) {
+            return;
+        }
         $filePath = [];
-        foreach ($this->getInstalledPlugins() as $plugin) {
-            $filePath[] = DIR_FS_CATALOG . 'zc_plugins/' . $plugin['unique_key'] . '/' . $plugin['version'] . '/';
+        foreach ($this->getInstalledPlugins() as $plugin) { // make an array of all installed plugins paths
+            $filePath[$plugin['unique_key']] = DIR_FS_CATALOG . 'zc_plugins/' . $plugin['unique_key'] . '/' . $plugin['version'];
         }
         if (!in_array($pluginpath, $filePath)) {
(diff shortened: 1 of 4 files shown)


includes/classes/PluginManager.php
```diff
@@ -308,4 +311,33 @@ public function getPluginControl()
     {
         return $this->pluginControl;
     }
+
+    protected function loadPluginLanguageConstants($pluginpath)
+    {
+        $pluginpath = str_replace('\\', '/', $pluginpath);
+        if (!is_dir($pluginpath) || !str_contains($pluginpath, DIR_FS_CATALOG . 'zc_plugins/')) {
```
drbyte: Suggested security improvement:
```suggestion
        if (!is_dir($pluginpath) || !str_starts_with($pluginpath, DIR_FS_CATALOG . 'zc_plugins/')) {
```

drbyte:
Re: `lang.uniquekey.php` 

I wonder if it's even necessary to enforce the use of the plugin key name.
Why not just something much more generic? Maybe `lang.manifest.php`, since it's an override to the `manifest` file's content?
That would save a lot of parsing for uniquekey etc.

piloujp:
> The naming of those language files, IMO, shouldn't start with `lang.` since they're in the older, non-array form.

Perhaps I should have squash everything before doing the PR. The final version uses files that are already in language packs and are actually loaded automatically when the plugin is installed. They are in the array format and should begin with `lang.`.

piloujp:
> Re: `lang.uniquekey.php`
> 
> I wonder if it's even necessary to enforce the use of the plugin key name. Why not just something much more generic? Maybe `lang.manifest.php`, since it's an override to the `manifest` file's content? That would save a lot of parsing for uniquekey etc.

It is not really necessary, these files were already named like this (except for the case that was not always respected due to my case-insensitive Windows system) in the language packs... Plus, in the code, the unique key had to be extracted for constants names anyway.
These files contain not only the name and description for the plugin, but also all plugins constants that are defined in the database.
The only inconvenient is that some plugin already had a file with this name in the `extra_definitions` folder, well actually I found only one.
That said, it is no problem to rename those to `lang.manifest.php` as the only language packs that actually provide those database constants files are probably the two packs I maintain...

drbyte:
You bring up a good point. 
And that's partially what I was targeting: a naming convention that's straightforward for plugin creators/maintainers to implement, and as generic as possible.

The fact that database constants are in that file seems logical too.

Maybe `lang.plugin_core.php`? Or just `lang.plugin_details.php`? I suggested `lang.manifest.php` because the manifest file has some similar data in it, but your point about database constants is good.

It doesn't really need the plugin's "name" in the filename, because it's already located in the plugin's subdirectory structure, and installing it to a store doesn't require uniqueness because it's not copied into the main store subdirectories (zc_plugins is separate).

That's where my thoughts were going.

piloujp:
I changed the name for languages constants file and did few tests without any problems.

includes/classes/PluginManager.php
```diff
@@ -308,4 +311,32 @@ public function getPluginControl()
...
+        if (!is_dir($pluginpath) || !str_starts_with($pluginpath, DIR_FS_CATALOG . 'zc_plugins/')) {
+            return;
+        }
+        $filePath = [];
+        foreach ($this->getInstalledPlugins() as $plugin) { // make an array of all installed plugins paths
+            $filePath[$plugin['unique_key']] = DIR_FS_CATALOG . 'zc_plugins/' . $plugin['unique_key'] . '/' . $plugin['version'];
+        }
+        if (!in_array($pluginpath, $filePath)) {
+            $pluginuniquekey = preg_replace('/\/v?\d+\.\d+\.\d+$/', '', $pluginpath);
+            $pluginuniquekey = strtoupper(substr($pluginuniquekey, strrpos($pluginuniquekey, '/') + 1)); // retrieve plugin's unique key
```
lat9: 1. There's no guarantee that DIR_FS_CATALOG doesn't also have a backslash (`\`), so the str_replace on line 317 could negate the comparison on line 318.  I also believe that the check is unneeded since its a protected method called from a protected method.
2. The input `$pluginpath` is already of the form `DIR_FS_CATALOG . 'zc_plugins/$unique_key`, so the current plugin key can simply be derived as `$pluginuniquekey = end(explode('/', $pluginpath));`.  Note that the regex currently used on line 326 _assumes_ a version numbering that might not be true, i.e. all plugin versions start with 'v' and have no '-beta1' or similar suffix.

piloujp: > 1. There's no guarantee that DIR_FS_CATALOG doesn't also have a backslash (`\`), so the str_replace on line 317 could negate the comparison on line 318.  I also believe that the check is unneeded since its a protected method called from a protected method.
> 
You are right, the function that calls this one already checked for that. This test can just be dropped, I suppose.

>     2. The input `$pluginpath` is already of the form `DIR_FS_CATALOG . 'zc_plugins/$unique_key`, so the current plugin key can simply be derived as `$pluginuniquekey = end(explode('/', $pluginpath));`.  Note that the regex currently used on line 326 _assumes_ a version numbering that might not be true, i.e. all plugin versions start with 'v' and have no '-beta1' or similar suffix.

Well, I don't follow here... The end of `$pluginpath` before the regex is the plugin version that begins or not by a v (`v?` equals 0 or 1 v). For the end part, adding `.*$` would let possibilities for any sub-versioning which is better.
About the `end/explode` functions, I must say I do not see any advantage for this over `substr/strrpos`. I actually think it would be slower to convert a string to an array rather than using `substr`, but I could be wrong on this.
If there was a `end-1()` function It would be nice...

piloujp: What about this?
```
$explodedpath = explode('/', $pluginpath);
$pluginuniquekey = strtoupper($explodedpath[count($explodedpath) - 2]);
```
In place of lines 326 & 327?

piloujp: To take in account backslashes, perhaps this would be better and line 317 (str_replace...) would not be needed anymore:
```
$explodedpath = preg_split('/[\\/]+/', $pluginpath);
$pluginuniquekey = strtoupper($explodedpath[count($explodedpath) - 2]);
```

lat9: The caller of the new method has already started with `DIR_FS_CATALOG . 'zc_plugins/`, so any backslashes present in DIR_FS_CATALOG don't matter for the use of `explode`.

piloujp: Well, I am on a Windows system and as soon as I remove line 317 that replaces backslashes, I got a blank page. `$plugingpath` is actually formatted like this:
D:/wamp/www/zencartJP/zc_plugins\DisplayLogs\v3.0.3

piloujp: I have just sent the last code that works for me using `explode` function, but still replacing backslashes at the beginning. It has to be done before the `in_array` test.

lat9 (approved):

drbyte:
@piloujp For this language-constant detection, would you consider changing the name from lang.database_constants.php to `lang.menu.php`?

... cuz it's not really database stuff that you're looking up here. And since it's related to the plugin's admin menus, the defines for both the Plugin Manager and any other admin menus can go in this `lang.menu.php` file that's language-specific.

And separately I'd like to explore an additional "standard" for zc_plugins where we encourage plugin authors to put their db defines into a file in the root of the plugins dir, named database_defines.php

piloujp:
No problem at all!

drbyte:
@piloujp Yes please.

piloujp:
@drbyte Done.

PR #6994, opened 2025-02-22
Display sold out as an image if so configured

scottcwilson:
In response to this forum post:
https://www.zen-cart.com/showthread.php?230618-Sold-Out-Image

PR #7209, opened 2025-08-28
CSS-images_buttons

piloujp:
This PR is a solution to, and closes #6994 .
As the discussion was dropped off 3 months ago without a final decision and I found out I still had this code in one off my test cart, here it is.
It adds a third option to the setting `Images use CSS buttons` to allow use of CSS button only if there is no image file. This actually allows overriding the use of CSS buttons only for a specific button by just adding its image.
This third option is called 'If' but if you have an idea for a more meaningful name, it is easy to change.


zc_install/sql/install/mysql_zencart.sql
```diff
@@ -3074,7 +3074,7 @@ INSERT INTO configuration (configuration_title, configuration_key, configuration
...
 
 # CSS Buttons switch
-INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('CSS Buttons', 'IMAGE_USE_CSS_BUTTONS', 'Yes', 'CSS Buttons<br />Use CSS buttons instead of images (GIF/JPG)?<br />Button styles must be configured in the stylesheet if you enable this option.', '19', '147', 'zen_cfg_select_option(array(\'No\', \'Yes\'), ', now());
+INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('CSS Buttons', 'IMAGE_USE_CSS_BUTTONS', 'Yes', 'CSS Buttons<br>Use CSS buttons instead of images (GIF/JPG)?<br>Button styles must be configured in the stylesheet if you enable this option.<br>Yes - Use css buttons<br>No - Use images buttons<br>If - Use images if exist, if not use css buttons', '19', '147', 'zen_cfg_select_option(array(\'No\', \'Yes\', \'If\'), ', now());
```
drbyte:
```suggestion
INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('CSS Buttons', 'IMAGE_USE_CSS_BUTTONS', 'Yes', 'CSS Buttons<br>Use CSS buttons instead of images (GIF/JPG)?<br>Button styles must be configured in the stylesheet if you enable this option.<br>Yes - Use CSS buttons<br>No - Use images buttons<br>Found - Use images if exist, else use CSS buttons', '19', '147', 'zen_cfg_select_option(array(\'No\', \'Yes\', \'Found\'), ', now());
```

includes/templates/template_default/sideboxes/tpl_search_header.php
```diff
@@ -16,7 +16,7 @@
 
 $content .= '&nbsp;';
 
-if (strtolower(IMAGE_USE_CSS_BUTTONS) == 'yes') {
+if (strtolower(IMAGE_USE_CSS_BUTTONS) === 'yes' || (strtolower(IMAGE_USE_CSS_BUTTONS) === 'if' && !file_exists(DIR_FS_CATALOG . DIR_WS_TEMPLATE . 'buttons/' . $_SESSION['language'] . '/' . BUTTON_IMAGE_SEARCH))) {
```
drbyte: Since `zen_image_submit` already checks for the file, this is probably redundant. 
Same for `tpl_search.php`.

drbyte:
This seems like a reasonable approach. 

I'd like to use `Found` instead of `If`.

includes/templates/template_default/sideboxes/tpl_search_header.php
piloujp: Yes, I removed them.

torvista:
Could this not be viewed as a template-specific option? i.e. should not be a global setting

torvista:
Should not the first clause be 

if (strtolower(IMAGE_USE_CSS_BUTTONS) === 'no'...: i.e use an image.

https://github.com/zencart/zencart/blob/c649b2344e4f356f78d9640ac913184c61733ce4/includes/templates/template_default/sideboxes/tpl_search_header.php#L19-L23

and since the values yes/no/found are fixed/in the database, is there any need to convert to lowercase?

drbyte:
> if (strtolower(IMAGE_USE_CSS_BUTTONS) === 'no'...: i.e use an image.

I understand your question. I too wondered the same. 
This header template logic is strange, and always has been; here's why: in the `no` case, no image is used at all, ever. Just a regular HTML browser-generated form-submit "button".
And since `zen_image_submit()` is where the actual logic is for generating a CSS button or not, the `yes/found` case is correct.

Your point about lowercase checking is valid. I was going to propose we provide a 3-way-choice dropdown where the displayed text is customizable; however that requires that the value stored in the db is numeric, which then breaks all existing templates where the check is for 'yes' or 'no', and this topic isn't one that warrants such an aggressive change.

PR #7216, opened 2025-09-01
Products Available Time setting

piloujp:
This PR adds possibility to set product availability date and time. It uses a JavaScript `DateTimePicker` and MySQL scheduled events.
The date and time picker is used in product update page and price manager instead of the actual date picker.
Scheduled events are managed in `functions_products.php` where they are created for each product that is going to be available the same day. They can also be created, deleted or modified when updating a product.
Similar functionality was developed in PR #7030 ,but it was not working as the time saved in database was always `00:00:00`.

I have tested this for a few weeks now and it performs well. There are some limitations:
- MySQL scheduled events need the Event Scheduler to be active in MySQL Server, which is the default setting in MySQL 8 but was not in MySQL 7. This is checked and an error message is sent if necessary.
- MySQL user needs to have `EVENT` rights, which is generally the case, but some hosting company might not give it by default. Here too, a message is sent if the requirement is not met.
- The free `DateTimePicker` does not scale down well on mobile, but I don't think the admin interface is made to be used on mobile anyway.
- There is only one slider for the hours and minutes setting. Separate ones would be better. There are some `DateTimePicker` that do that, but they are all paid products. The default setting for minute is now set to every 10 minutes but can be changed easily.

Except that it performs very well and filter dates too, and user can still do manual settings in the field if necessary.

`DateTimePicker` references [here](https://xdsoft.net/jqplugins/datetimepicker/).


torvista:
How does Specials do it/is this the same/should they not use the same method?

piloujp:
> How does Specials do it/is this the same/should they not use the same method?

The database (and few other files) needs to be modified to apply this for specials, featured products and categories. For those, database fields are `Date` type only...
If it goes well with products available date, extending this functionality is definitely possible.

piloujp:
> I really meant, if it works for Specials already, why not extend that method to the product?
> 
> Or is this new method preferable for some reason and hence could/should be extended to Specials etc., once approved?

There is no time setting for Specials, there has never been.

piloujp:
> Could a simple textbox be added for the time to be typed in manually? Defaulting to `00:00:00` as it does today?
> 
Yes, it is another possibility, but this new field would have to be combined or separated when writing/reading database. This is also a ton of code to add, and I am not sure it looks so good.

> Time-pickers are a usability nuisance on desktop where full keyboard is readily available: faster to type than to mouse-select 3 time components.
I agree, but generally you can do both. This PR `DateTimePicker` lets you manually adjust the field, but still filters it.

PR #7217, opened 2025-09-01
Correct PHP warnings on `customers_authorization` page

lat9:
Corrects PHP warnings on the `customers_authorization` page when the page is "hit" and a customer isn't logged in.

Also
- Correctly displays the page when the customer's current authorization status is '3' (Pending, can browse with prices but cannot purchase until authorized).
- Simplifies the header/footer/left/right enable/disable settings.

### DIFF
--- a/includes/modules/pages/customers_authorization/header_php.php
+++ b/includes/modules/pages/customers_authorization/header_php.php
@@ -7,27 +7,31 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: DrByte 2020 Jul 10 Modified in v1.5.8-alpha $
  */
+if (!zen_is_logged_in()) {
+    zen_redirect(zen_href_link(FILENAME_DEFAULT));
+}
 
-
-$sql = "SELECT customers_authorization 
-        FROM " . TABLE_CUSTOMERS . " 
-        WHERE customers_id = :customersID";
+$sql =
+    "SELECT customers_authorization
+       FROM " . TABLE_CUSTOMERS . "
+      WHERE customers_id = :customersID";
 
 $sql = $db->bindVars($sql, ':customersID', $_SESSION['customer_id'], 'integer');
-$check_customer = $db->Execute($sql);
+$check_customer = $db->Execute($sql, 1);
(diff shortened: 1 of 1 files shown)


includes/modules/pages/customers_authorization/header_php.php
```diff
@@ -7,27 +7,31 @@
...
-if (CUSTOMERS_AUTHORIZATION_COLUMN_LEFT_OFF == 'true') $flag_disable_left = true;
-if (CUSTOMERS_AUTHORIZATION_FOOTER_OFF == 'true') $flag_disable_footer = true;
-if (CUSTOMERS_AUTHORIZATION_HEADER_OFF == 'true') $flag_disable_header = true;
-
-?>
\ No newline at end of file
+$flag_disable_right = (CUSTOMERS_AUTHORIZATION_COLUMN_RIGHT_OFF === 'true');
+$flag_disable_left = (CUSTOMERS_AUTHORIZATION_COLUMN_LEFT_OFF === 'true');
+$flag_disable_footer = (CUSTOMERS_AUTHORIZATION_FOOTER_OFF === 'true');
+$flag_disable_header = (CUSTOMERS_AUTHORIZATION_HEADER_OFF === 'true');
```
drbyte, line 37: For "this" page, this approach is probably fine.
However, the "normal" practice is to assume all these are `false`, or have been already set to `true` at an earlier point in the template prep process, and only set them to `true` if some setting for the page intends such. 
That way any earlier "overrides" aren't re-overridden here. Unless truly intended.

drbyte: Actually, I take that back, sorta: the/this `header_php.php` file is loaded before all template-system overrides/defaults.

lat9: Although this has been merged, how about using the null-coalescing operation (`??=`) so that the conditional applies only if the variable's not yet been set?

drbyte: Yes, good point.

lat9: I'll do a follow-on PR ...

PR #7226, opened 2025-09-03
Attributes Controller: minor things

torvista:
I'll squash when approved.

Aside there is a great deal of uneeded type casting in sql queries...I assume this should stay as belt and braces security?

### DIFF
--- a/admin/attributes_controller.php
+++ b/admin/attributes_controller.php
@@ -864,7 +864,7 @@ function zen_js_option_values_list($selectedName, $fieldName)
           <div class="form-group">
             <div class="col-sm-6 text-center">
               <span class="text-danger"><strong><?= TEXT_INFO_ATTRIBUTES_FEATURE_CATEGORIES_COPY_TO ?></strong></span><br>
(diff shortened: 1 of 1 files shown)


admin/attributes_controller.php
```diff
@@ -2157,17 +2157,17 @@ function update_option(theForm) {
...
-        var Opts = theField.options.length;
+        let Opts = theField.options.length;
         while (Opts > 0) {
-          Opts = Opts - 1;
+          Opts -= 1;
           theField.options[Opts] = null;
         }
 
-<?= zen_js_option_values_list('SelectedOption', 'theField') ?>
+<?php echo zen_js_option_values_list('SelectedOption', 'theField'); ?>
```
drbyte, line 2170: Why the change to full `echo` syntax?

torvista: poor little IDE couldn't cope with that in the middle of javascript so was flagging two errors.

Also looks odd/orphaned.

drbyte: Not opposed.

drbyte (approved):

PR #7227, opened 2025-09-03
Adding instructions to the `moneyorder` status history

retched:
When a customer completes a payment using `moneyorder`, it will add a note in the `order_status_history` table providing instructions to the customer where and how to prepare payment. (Done by tacking the EMAIL_FOOTER note and one additional define.)


retched:
This is the repaired #6999 (issue). (Canceled that one due to a botched catch-up merge.)

The end result will provide the instructions as noted. Here's how they appear on a sample run:

image

(The first line is the customer notes line... so it was blanked.)

drbyte:
I can't recall if we may already have such a flag, to just piggyback from (er, to use)?

drbyte:
I'm in another project right now without my IDE open. I'd like to search for wherever that zen_osh_update() function is called, look at how that updated_by is set in all those cases, and whether there's a predeterminer done before the call, as well as also any fallbacks within the function itself. THat'd be what I'd wanna review.

drbyte:
Of course, your "added" use-case has a unique objective that we've never had cause/demand to accommodate, so that's an additional wrinkle to consider.

PR #7230, opened 2025-09-04
Correcting 'undefined array key' log thrown by the cache.php class

lat9:
... as reported [here](https://www.zen-cart.com/showthread.php/230855-quot-Undefined-array-key-quot-warning-in-includes-classes-cache-php) on the Zen Cart forums.

Also 'refreshes' the class' processing.

### DIFF
--- a/includes/classes/cache.php
+++ b/includes/classes/cache.php
@@ -8,212 +8,206 @@
  * @version $Id: DrByte 2024 Oct 16 Modified in v2.1.0 $
  */
 if (!defined('IS_ADMIN_FLAG')) {
-  die('Illegal Access');
+    die('Illegal Access');
 }
+
 /**
  * cache Class.
  * handles query caching
  *
  */
-class cache extends base {
-
-  function sql_cache_exists($zf_query, $zf_cachetime=null) {
-    global $db;
-    $zp_cache_name = $this->cache_generate_cache_name($zf_query);
-    switch (SQL_CACHE_METHOD) {
-      case 'file':
-      // where using a single directory at the moment. Need to look at splitting into subdirectories
-      // like adodb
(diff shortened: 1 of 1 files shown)


includes/classes/cache.php
```diff
@@ -57,7 +57,8 @@ public function sql_cache_is_expired($zf_query, $zf_cachetime): bool
         $zp_cache_name = $this->cache_generate_cache_name($zf_query);
         switch (SQL_CACHE_METHOD) {
             case 'file':
-                if (@filemtime(DIR_FS_SQL_CACHE . '/' . $zp_cache_name . '.sql') > (time() - $zf_cachetime)) {
+                $filename = DIR_FS_SQL_CACHE . '/' . $zp_cache_name . '.sql';
+                if (file_exists($filename) && filemtime($filename) > (time() - $zf_cachetime)) {
```
drbyte: I'm pretty sure that the `@` suppressor was originally used to accommodate cases where the server OS has forbidden making calls to filemtime causing PHP to croak.
I'm not sure if there's a straightforward way to programmatically pre-determine whether it will fail.
We "could" use `try/catch` exception handling if we want to avoid the `@`.

lat9: Perhaps, but I'm more of the mind that the suppressor's use was that `filemtime` will throw a warning if the file doesn't exist.

drbyte: That would also be true; however, I do recall it was a filesystems permissions issue, where it was denying the PHP user the permission to read the file attributes.

lat9: So, just add the suppressor back?

drbyte: IMO that's the simplest way. And the code previously failed gracefully with the suppressor, so we're not changing what's worked before.

PR #7232, opened 2025-09-04
Add another notifier below comments box.

scottcwilson:
For an additional content block.


torvista:
image

Just out of interest, what was the driver for adding this extra block as opposed to using one of the other observers...looks a bit lonely down there!

scottcwilson:
Moving Admin Notes below the order comments.

PR #7233, opened 2025-09-05
additional images database table

zenexpert:
All additional images are stored in the database just like the main image - this allows you to easily add additional images when you're creating a new product or when you're editing the product - it all runs on the admin edit product page.

Here's a quick overview and benefits of a DB driven approach:

- Faster lookups with indexing: a simple SELECT query is much faster than scanning directories.
- Current approach relies on preg_match() to find matches, which can be slow, especially when scanning a large number of files.
- Reduced server load: eliminates unnecessary filesystem operations and reduces disk I/O, especially on busy websites where multiple users could trigger the same directory scan simultaneously.
- Easier to integrate with admin panels or external tools.

Scanning directories with thousands of files can be slow, especially on shared hosting... A database query is significantly faster than scanning a large folder. Specific file naming conventions are no longer required (ie. image.jpg, image_01.jpg, image_02.jpg) and you can use different file extensions for additional images (no longer limited to same file type as main image).

And last but not least - it's just simpler.


piloujp:
A query to create the new database table would be nice...

zenexpert:
> A query to create the new database table would be nice...

What?I Are you saying I forgot to include some files? No way, that's impossible, that never happens... Hehe, whooops...

torvista:
I assume this is an **optional** way to assign an additional images from any location that 
a) does not break the current Zen Cart method (whatever that is, me being an Image Handler user from day one, ) of adding additional images.
b) does not break the current Image Handler method where correctly_named images automatically become additional.

zenexpert:
No, not optional. It's a new method that replaces the old one. IMHO, a much better method that wants to replace a dinosaur.
It probably could be done so you get to choose which method you want to use, but I believe that would be extremely confusing for the end user.

Images could still be manually uploaded the old way (FTP) and then you run the convert tool (which is added to Tools menu). Same with Image Handler - you could in theory use it to add additional images, but would need to run the convert tool (or IH would need a minor modification to update the database). If you were to choose that, you'd want to keep the processed_products_ids.txt meaning the convert tool shouldn't delete it. 
However, if it gets adopted and we turn the additional images feature into a drag-and-drop field where you can add all additional images at once, I don't think IH interface will be much fun for anyone. IH still works in the frontend, of course.

zenexpert:
You actually had me thinking there - why not go for the drag-and-drop straight away?

add1

add2

lat9:
> No, not optional. It's a new method that replaces the old one. IMHO, a much better method that wants to replace a dinosaur. It probably could be done so you get to choose which method you want to use, but I believe that would be extremely confusing for the end user.
> 
> Images could still be manually uploaded the old way (FTP) and then you run the convert tool (which is added to Tools menu). Same with Image Handler - you could in theory use it to add additional images, but would need to run the convert tool (or IH would need a minor modification to update the database). If you were to choose that, you'd want to keep the processed_products_ids.txt meaning the convert tool shouldn't delete it. However, if it gets adopted and we turn the additional images feature into a drag-and-drop field where you can add all additional images at once, I don't think IH interface will be much fun for anyone. IH still works in the frontend, of course.

Not optional is, IMO, a non-starter.

zenexpert:
So, you want to confuse users with an option for THEM to choose from how the system would handle additional images? This will cause so many headaches down the road, especially if users decide to switch back and forth...

But sure, here you go.

admin/includes/init_includes/init_aidba_install.php
```diff
@@ -0,0 +1,25 @@
...
+    $db->Execute("ALTER TABLE " . TABLE_PRODUCTS ." ENGINE=InnoDB;");
+
+    // create products_additional_images table
+    $db->Execute("CREATE TABLE " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " (
+            id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+            products_id INT(11) NOT NULL,
+            additional_image VARCHAR(255) NOT NULL,
+            sort_order INT(11) DEFAULT 0,
+            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+            FOREIGN KEY (products_id) REFERENCES " . TABLE_PRODUCTS ."(products_id) ON DELETE CASCADE
```
drbyte: Should there be a space between the table name and the `(products_id) ON DELETE CASCADE`?

includes/modules/additional_images.php
```diff
@@ -24,75 +24,87 @@
...
-    // if in a subdirectory
-    if (strrpos($products_image, '/')) {
-        $products_image_match = substr($products_image, strrpos($products_image, '/') + 1);
-        //echo 'TEST 1: I match ' . $products_image_match . ' - ' . $file . ' -  base ' . $products_image_base . '<br>';
-        $products_image_match = str_replace($products_image_extension, '', $products_image_match) . '_';
-        $products_image_base = $products_image_match;
-    }
+    if (ADDITIONAL_IMAGES_APPROACH === 'modern') {
+        $images_array = [];
+        $sql = "SELECT additional_image FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = :productsID";
```
drbyte: Should this include sort_order?
```suggestion
        $sql = "SELECT additional_image FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = :productsID ORDER BY sort_order";
```

includes/modules/additional_images.php
```diff
@@ -24,75 +24,87 @@
...
+        $sql = $db->bindVars($sql, ':productsID', (int)$_GET['products_id'], 'integer');
+        $images_query = $db->Execute($sql);
+
+        while (!$images_query->EOF) {
+            $images_array[] = DIR_WS_IMAGES . $images_query->fields['additional_image'];
+            $images_query->MoveNext();
```
drbyte: Using `foreach` instead of `while ! EOF` would eliminate the need for the `->MoveNext()` call.

In lots of places.

drbyte (commented):
Posted a few code comments spotted while reviewing.

I haven't reviewed the admin "converter" tool. I wonder if it'd be best as an encapsulated plugin that someone could temporarily install and then remove when no longer needed, since in theory it'd be only used once?

lat9:
> So, you want to confuse users with an option for THEM to choose from how the system would handle additional images? This will cause so many headaches down the road, especially if users decide to switch back and forth...
> 
> But sure, here you go.

It's called downward compatibility.  It's one thing for _new_ users to have this processing provided, it's yet another to re-train admins that have worked with the existing method for the past decades.

admin/includes/init_includes/init_aidba_install.php
zenexpert: No, I don't think so. According to MySQL [docs](https://dev.mysql.com/doc/refman/8.4/en/create-table-foreign-keys.html) and the example provided, there's no space between the table name and field name that it references to:
`CREATE TABLE parent (
    id INT NOT NULL,
    PRIMARY KEY (id)
) ENGINE=INNODB;

CREATE TABLE child (
    id INT,
    parent_id INT,
    INDEX par_ind (parent_id),
    FOREIGN KEY (parent_id)
        REFERENCES parent(id)
        ON DELETE CASCADE
) ENGINE=INNODB;`

torvista:
Forget my comments on Image Handler that is nothing to do with naming or additional images, in my use.

As I now have checked, ZC core method is that additional images are auto-found and displayed just by using a _suffixOfWhatever with the base filename. No clicking or selecting or thinking involved and forces logic. Could not be simpler.
Dumbos who dump everything in the root have to deal with their laziness at some point...all us grown-ups have to correct our past mistakes sometimes. 

I am all for allowing almost any contribution to the ZC code while it improves or at least does not degrade anything....beggars can't be choosers. If someone only wants to correct a missing semicolon, encourage them!

If the world and his dog now demands the ability to manually select randomly-named images to add to a product and someone is willing to code it, I don't see the problem **as long as it's an option**, as the current method is a no-brainer.

dbltoe:
An estimated 3.5 million devices worldwide are still using Windows XP.

If a company like Microsoft is not willing to draw the line, ...

Optional for at least a few versions is the better choice.

drbyte:
Alright, the banter has become unproductive.
As far as I'm concerned, storing the image filenames into the database is beneficial. So I support the proposed move towards that.
Plus having a friendlier UI for product images will make it easier for both existing and new merchants alike, where they're in the pattern of using the Admin to upload their product images. (For those using FTP as their workflow, they can stick with the old, and encounter no negative impact.)
It also potentially lets merchants sort the display of images from the browser without having to go rename files on the filesystem.

For the sake of compatibility to not break existing stores, we'll retain the old functionality in v2.x.

We may (strongly) consider in ZC v3.x to make the database approach the default, and merchants can decide whether to change from the new default to go back to the old, or to migrate their images into the database for faster performance.

drbyte:
EDIT/Update to my last post.
Two things: first, my comment about 2.x was intended to mean "we'll add this, but not make it the default" ... not "we won't add this until 3.x"

Thinking further, I think we should add this in 2.2.x AND make it the default in 2.2.x for NEW installs only. 
ie: The upgrade script just would default to 'legacy' instead of 'modern'.

piloujp:
I think I found two bugs in catalog side, when displaying products details page:

First, the new table definition is missing in catalog side, producing a log.

Second one, variables  `$products_image_directory` and `$products_image_extension` are undefined near line 132 (when using new method). Replacing line 132 by this:
```
        $products_image_extension = substr($file, strrpos($file, '.'));
        $products_image_large = str_replace(DIR_WS_IMAGES, DIR_WS_IMAGES . 'large/', $file) . str_replace($products_image_extension, '', $file) . IMAGE_SUFFIX_LARGE . $products_image_extension;
```
And deleting the 2 other occurrences of `$products_image_directory` few lines under (around 144) fixed it.

admin/includes/init_includes/init_aidba_install.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+if (!defined('IS_ADMIN_FLAG')) {
+    die('Illegal Access');
+}
+if(!defined('ADDITIONAL_IMAGES_APPROACH')) {
+    $db->Execute("INSERT INTO " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Additional Images Approach', 'ADDITIONAL_IMAGES_APPROACH', 'old', 'Enable Additional Images Module', '4', '100', 'zen_cfg_select_option(array(\'old\', \'modern\'), ', NOW())");
+}
```
lat9: Just use INSERT IGNORE INTO, then you don't need the PHP check.  Use short-array syntax for the `zen_cfg_select_option` parameter.

admin/includes/init_includes/init_aidba_install.php
```diff
@@ -0,0 +1,25 @@
...
+            id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+            products_id INT(11) NOT NULL,
+            additional_image VARCHAR(255) NOT NULL,
+            sort_order INT(11) DEFAULT 0,
+            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+            FOREIGN KEY (products_id) REFERENCES " . TABLE_PRODUCTS ."(products_id) ON DELETE CASCADE
+            ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;");
+
+    zen_register_admin_page('toolsAidba', 'BOX_TOOLS_CONVERT_AIDBA', 'FILENAME_AIDBA', '', 'tools', 'Y');
+}
```
lat9: Since this is going into core, shouldn't this initialization be part of the zc_install for zc220?  The `CHARSET` specification on the create-table SQL should be dropped &mdash; while *all* stores should have their databases set to use a utf8mb4 variant by default, having the specification on the SQL could result in mixed collations for stores that are behind-the-times.

Q: Why the forcing of the database engine to InnoDB?

lat9 (changes requested):

admin/includes/init_includes/init_aidba_install.php
zenexpert: aye aye captain. also changing old/modern to legacy/modern

admin/includes/init_includes/init_aidba_install.php
zenexpert: uhm... to be honest, I'm not sure how to get it into the installer. Sorry.
But now that you mentioned it, we should probably move the extra definitions into the core files as well...

InnoDB - because of ON DELETE CASCADE. It requires InnoDB, and is a pretty safe way to ensure additional images are deleted from the database when a product is deleted so you don't have orphaned records.

lat9: Hmm, I'm not sure how to get it into the installer either, since the `REFERENCES` needs to include the table-prefix if set.  Perhaps @drbyte can chime in on this.

zenexpert: I'm stuck on the converter - as mentioned earlier by drbyte, not sure if it should be moved into an encapsulated plugin. If so, it might be quite counter-intuitive for the user to switch to the new style and then has to install the converter - this way it installs automatically and is straight-forward. On the other hand it is pretty much one-time use (unless it times out or fails so you need to re-run) so there's no point in leaving it in the Tools menu... Any suggestions?

lat9: Sneaky, but ...

Upon successful completion, the tool could change its `admin_pages::display_on_menu` value to 'N'.  That way, the tool is still accessible via direct browser link but doesn't display anymore within the menu structure.

lat9: > Hmm, I'm not sure how to get it into the installer either, since the `REFERENCES` needs to include the table-prefix if set. Perhaps @drbyte can chime in on this.

Noting that it'll be easier/possible to get this into zc_install if the `products_additional_images` table were to 'lose' that auto-remove-when-the-product-is-removed clause.  That removal would be accompanied by a specific `DELETE FROM` action in the `zen_remove_product` function for that table.

piloujp:
I had time to look into it this weekend and found some problems with logic which messes up additional images file path both in converter `admin/convert_additional_images_for_database_approach.php` and product display page `includes/modules/additional_images.php`.
As there were too many changes to do (mainly in the two files above), I cannot share it here easily and put it in a branch on [my GitHub ZC clone repo](https://github.com/piloujp/zencart/tree/Additional-images-database-approach).
I also added the code for the installer modifications.
Please have a look.

lat9:
> I had time to look into it this weekend and found some problems with logic which messes up additional images file path both in converter `admin/convert_additional_images_for_database_approach.php` and product display page `includes/modules/additional_images.php`. As there were too many changes to do (mainly in the two files above), I cannot share it here easily and put it in a branch on [my GitHub ZC clone repo](https://github.com/piloujp/zencart/tree/Additional-images-database-approach). I also added the code for the installer modifications. Please have a look.

Could we also (pretty please) shorten the name of that conversion tool?

zenexpert:
> > I had time to look into it this weekend and found some problems with logic which messes up additional images file path both in converter `admin/convert_additional_images_for_database_approach.php` and product display page `includes/modules/additional_images.php`. As there were too many changes to do (mainly in the two files above), I cannot share it here easily and put it in a branch on [my GitHub ZC clone repo](https://github.com/piloujp/zencart/tree/Additional-images-database-approach). I also added the code for the installer modifications. Please have a look.
> 
> Could we also (pretty please) shorten the name of that conversion tool?

Hahaha, yeah, been poking me in the eye as well, just wanted to make it perfectly clear what the file is. Any suggestions? aidba_converter.php perhaps?

admin/includes/init_includes/init_aidba_install.php
zenexpert: Agreed. Was thinking of this as a bulletproof method of not having orphaned records in some wild case scenarios, but if it causes more trouble than it's potentially worth, it's better to just stick with the standard stuff. Will make the change (also noting that the `zen_remove_product` function needs a minor update as it will fail on delete from products_additional_images if still in legacy mode)

lat9: If the `products_additional_images` table is **always** present, then I'm not seeing how the legacy vs. modern mode would affect that failure.

Noting that once the install is converted to a zc_install, that table's going to "always" be present.

lat9: If you're in agreement to stick with the "standard stuff", I'll send you the zc_install `.sql` updates to your direct email address.

zenexpert: Oh, OK, gotcha. The way it's set now in the installer is that the table doesn't exist until you switch to modern mode. But since it's going into zc_install and will always be there, yes, no need to do that check. Wasn't thinking that far ahead...

zenexpert: And yes, of course I'm in agreement to stick with whatever is deemed best. You're more than welcome to email me, or just add the command yourself, whatever you feel is easier (my vote is on the latter).
I've updated the installer to this:
`$db->Execute("CREATE TABLE " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " (
            id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
            products_id INT(11) NOT NULL,
            additional_image VARCHAR(255) NOT NULL,
            sort_order INT(11) DEFAULT 0,
            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
            INDEX idx_products_id (products_id)
            ) ENGINE=MyISAM;");`
switched to MyISAM and added an index. Hopefully that would do it.

piloujp:
> > I had time to look into it this weekend and found some problems with logic which messes up additional images file path both in converter `admin/convert_additional_images_for_database_approach.php` and product display page `includes/modules/additional_images.php`. As there were too many changes to do (mainly in the two files above), I cannot share it here easily and put it in a branch on [my GitHub ZC clone repo](https://github.com/piloujp/zencart/tree/Additional-images-database-approach). I also added the code for the installer modifications. Please have a look.
> 
> I guess I'm missing something obvious here - you've made adjustments to the original legacy code, which I literally copy/pasted. The only change I had made was to pull those two lines out of the if-else statement to avoid warnings that you had reported.
> 
> Could you please elaborate on how it messes up the file path and how come this wasn't an issue until now?

There were few problems:
- When changing option `ADDITIONAL_IMAGES_APPROACH` between `legacy` and `modern`, additional images were not displaying in `legacy` mode but, were OK in `modern`.
- Another problem came out when main and additional images extensions are different, like `.gif` for main and `.jpg` for additional. Then additional images would not display.
- The next one was with large images (_LRG) that would not display. When clicking on thumbnails you always got standard image even if large one exists.
- The last one was with converter tool that missed half of the images (in subfolders) that needed to be added to database.
Actually, one of these problems might already have been there before your PR... I will check when I have time.

lat9:
Sure, if there are issues with the current base code, please submit a separate PR and I'll get those updates merged.

piloujp:
> > > > I had time to look into it this weekend and found some problems with logic which messes up additional images file path both in converter `admin/convert_additional_images_for_database_approach.php` and product display page `includes/modules/additional_images.php`. As there were too many changes to do (mainly in the two files above), I cannot share it here easily and put it in a branch on [my GitHub ZC clone repo](https://github.com/piloujp/zencart/tree/Additional-images-database-approach). I also added the code for the installer modifications. Please have a look.
> > > 
> > > 
> > > I guess I'm missing something obvious here - you've made adjustments to the original legacy code, which I literally copy/pasted. The only change I had made was to pull those two lines out of the if-else statement to avoid warnings that you had reported.
> > > Could you please elaborate on how it messes up the file path and how come this wasn't an issue until now?
> > 
> > 
> > There were few problems:
> > ```
> > * When changing option `ADDITIONAL_IMAGES_APPROACH` between `legacy` and `modern`, additional images were not displaying in `legacy` mode but, were OK in `modern`.
> > 
> > * Another problem came out when main and additional images extensions are different, like `.gif` for main and `.jpg` for additional. Then additional images would not display.
> > 
> > * The next one was with large images (_LRG) that would not display. When clicking on thumbnails you always got standard image even if large one exists.
> > 
> > * The last one was with converter tool that missed half of the images (in subfolders) that needed to be added to database.
> >   Actually, one of these problems might already have been there before your PR... I will check when I have time.
> > ```
> 
> Uhm... Between legacy and modern? So, switch from legacy to modern, you add some additional images in modern and then switch back to legacy? Yeah, that will never work because modern doesn't rename images to comply with legacy. There's even a warning it's not recommended to do it in the config entry. gif and jpg in modern or legacy? If in legacy, that has always been the case. You were forced to match the extension. In modern, this should work without issues. _LRG - honestly didn't check. Does anyone still actually use that? Converter - in all of my tests, it never missed a single additional image, including subfolders, assuming the main image was there as well. But again, the converter's detection is based on the original legacy code so if that code misses anything, the converter will miss it as well.
> 
> So yeah, if you found errors in the legacy code, these will certainly be in this code as well because I didn't go into that at all. Cheers for picking it up, it will make the transition to modern more robust!
> 
> May I ask @lat9 to verify these changes and we merge it here?

Well, I did some new test with ZC last GitHub version with demo data and the PR freshly downloaded after last commit. Results are as I already said, nothing changed.

Converter is only grabbing files in `images` folder, not in subdirectories. In Demo data, there are 3 DVDs and a mouse that have additional images, none of them are visible after installing and activating this PR. When going back to legacy mode, they are there again (which is kind of opposite of what I wrote in last message, my mistake).
The code I proposed is not perfect either and might grab some images because of their names, like the gift certificates in demo data which are like this, `gv.gif`, `gv_5.gif`, `gv_10.gif`, `gv_25.gif`, `gv_50.gif`, `gv_100.gif`, although each image file is for a different certificate...

About the extension problem, it seems I digged a fossilized bug, but It has some influence on this PR, especially if you copy/ mimic the code. After verification, in actual ZC, extensions must be identical for additional images but can be different for large ones... weird!

admin/includes/init_includes/init_aidba_install.php
```diff
@@ -3,9 +3,7 @@
...
-}
+$db->Execute("INSERT IGNORE INTO " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Additional Images Approach', 'ADDITIONAL_IMAGES_APPROACH', 'old', 'Use Legacy mode for old-school directory scanning and matching filenames. Additional images must be uploaded manually via FTP or using Image Handler.<br>Use Modern mode to add additional images directly from the admin product page. Filenames and extensions do not need to match.<br><strong>NOTE:</strong> if you are switching from Legacy to Modern for the first time, you can use the converter tool found in your Tools menu to update all existing product data with existing additional images. Switching back and forth is not recommended.', '4', '100', 'zen_cfg_select_option([\'legacy\', \'modern\'], ', NOW())");
```
drbyte:
```suggestion
$db->Execute("INSERT IGNORE INTO " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Additional Images Approach', 'ADDITIONAL_IMAGES_APPROACH', 'legacy', 'Use Legacy mode for old-school directory scanning and matching filenames. Additional images must be uploaded manually via FTP or using Image Handler.<br>Use Modern mode to add additional images directly from the admin product page. Filenames and extensions do not need to match.<br><strong>NOTE:</strong> if you are switching from Legacy to Modern for the first time, you can use the converter tool found in your Tools menu to update all existing product data with existing additional images. Switching back and forth is not recommended.', '4', '100', 'zen_cfg_select_option([\'legacy\', \'modern\'], ', NOW())");
```

zenexpert:
> #7239 should allow the SQL for CREATE TABLE with FOREIGN KEY in this PR to be put into the installer, and work in SQL Patch.
> 
> (It'll still fail if the referenced table isn't InnoDB yet, but that's fixable by doing the conversion first.)

Uhm... Just switched it over to MyISAM and no Foreign Key... So, back to InnoDB with Foreign Key? As lat9 suggested, it's not the end of the world if we're not using it, the `zen_remove_product` function should handle it anyway with the `DELETE` statement. I was just thinking of some rogue scenarios where possibly some 3rd party plugins weren't using that function but deleting "manually". Probably just overthinking?

drbyte:
As for the "fossil" of different file extensions on _LRG, I don't see any point in "fixing" it.  
Since same-type-extension is enforced for small/med, most sites will have followed same convention for large.
This is the way the legacy implementation has always worked. 
As long as the converter accommodates properly detecting matches that the legacy template will recognize and render, that should suffice.

drbyte:
Thinking further about the Encapsulated Plugin approach for the converter.  I envision it being a plugin that comes pre-installed in the zc_plugins dir.

How much value is there in the converter being run periodically if a storeowner uploads images using the legacy approach and then runs the converter to mass-add them to the database?
I'm thinking of the case where a merchant has scripts to mass-name files according to the legacy pattern, and uses that script when vendors send new product images. It would be easiest for them to keep using the same script, instead of writing a new one and then having to find a way to insert them into the db. 

So, whether the converter remains as a core tool, or an encapsulated tool, if it can be run anytime to refresh and ensure everything that needs to be in the db "is", that's probably ideal.

That said, if it's going to hang around, calling it something like 'Sync Product Images To Database' might be more meaningful in the long-term than 'Convert Images to Database Approach'. (ie: a tool called "convert to the new approach" might make sense the first time, but will lose its meaningfulness long-term).

A future iteration could be to rework it into an image-importer that treats the "legacy" logic as one possible import formula/strategy, and offers other strategies that might fit with supplier-feeds.

admin/aidba_converter.php
```diff
@@ -16,7 +16,7 @@
 
     $counter = 0;
 
-    $products_query = $db->Execute("SELECT products_id, products_image FROM " . TABLE_PRODUCTS);
+    $products_query = $db->Execute("SELECT products_id, products_image FROM " . TABLE_PRODUCTS . " WHERE products_image IS NOT NULL");
```
piloujp: I have some code improvement for this:
`$products_query = $db->Execute("SELECT products_id, products_image FROM " . TABLE_PRODUCTS . " WHERE NOT COALESCE(products_image, '') = ''");`
It will eliminate NULL as now and empty strings too. It might make a difference in big old databases where lots of images references have been deleted.

includes/modules/additional_images.php
```diff
@@ -24,75 +24,88 @@
...
+    $products_image_extension = substr($products_image, strrpos($products_image, '.'));
+
+    if (ADDITIONAL_IMAGES_APPROACH === 'modern') {
+        $images_array = [];
+        $sql = "SELECT additional_image FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = :productsID ORDER BY sort_order";
+        $sql = $db->bindVars($sql, ':productsID', (int)$_GET['products_id'], 'integer');
+        $images_query = $db->Execute($sql);
+
+        foreach ($images_query as $image) {
+            $images_array[] = DIR_WS_IMAGES . $image['additional_image'];
```
piloujp: Using this code:
`$images_array[] = substr($image['additional_image'], (strpos($image['additional_image'], '/') ?? -1) + 1);`
will remove subfolder like in legacy mode. Additional images are then displayed properly in both modes.

piloujp: Adding this line of code between lines 132 and 133 will allow the use of larges images in both mode:
`        $products_image_extension = substr($file, strrpos($file, '.'));`

With these 2 minimal corrections, letting aside old code problems, the display of additional images should now work.

drbyte:
To my comment earlier about running the converter multiple times ... when I ran it 3 times (using just demo products/images) it inserted database records 3 times for each.
Probably should set a unique index on `products_id, additional_image` for safety, and then have the converter skip re-inserting existing entries.

admin/aidba_converter.php
```diff
@@ -0,0 +1,140 @@
...
+                }
+            }
+            $dir->close();
+        }
+
+        // Insert matches into products_additional_images table
+        foreach ($matches as $sort_order => $additional_image) {
+            // Check if already exists
+            $exists_query = $db->Execute(
+                "SELECT id FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = " . $products_id . " AND additional_image = '" . zen_db_input($additional_image) . "'"
```
drbyte: To prevent the duplicate that's happening on line 70:
```suggestion
                "SELECT id FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = " . $products_id . " AND additional_image = '" . zen_db_input($subdir . $additional_image) . "'"
```

admin/includes/init_includes/init_aidba_install.php
```diff
@@ -0,0 +1,25 @@
...
+
+if(!$sniffer->table_exists(TABLE_PRODUCTS_ADDITIONAL_IMAGES) && defined('ADDITIONAL_IMAGES_APPROACH') && ADDITIONAL_IMAGES_APPROACH == 'modern') {
+    // create products_additional_images table
+    $db->Execute("CREATE TABLE " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " (
+            id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+            products_id INT(11) NOT NULL,
+            additional_image VARCHAR(255) NOT NULL,
+            sort_order INT(11) DEFAULT 0,
+            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+            INDEX idx_products_id (products_id)
```
drbyte: Make index unique
```suggestion
            INDEX idx_products_id (products_id)
            UNIQUE KEY idx_pid_img_zen (products_id, additional_image)
```

includes/modules/additional_images.php
zenexpert: Using ` $images_array[] = substr($image['additional_image'], (strpos($image['additional_image'], '/') ?? -1) + 1);` actually strips the first character of the image name when it's not in a subdirectory. So, for example, 1_small_00.jpg becomes _small_00.jpg
I've went for a somewhat different solution and it seems to work for both subdirs and main dir.

 As for `$products_image_extension = substr($file, strrpos($file, '.'));` - OK, taking your word for it.

zc_plugins/SyncAdditionalImages/v1.0.0/Installer/ScriptedInstaller.php
```diff
@@ -0,0 +1,18 @@
...
+
+use Zencart\PluginSupport\ScriptedInstaller as ScriptedInstallBase;
+
+class ScriptedInstaller extends ScriptedInstallBase
+{
+    protected function executeInstall()
+    {
+        zen_deregister_admin_pages(['toolsAidba']);
+        zen_register_admin_page('toolsAidba', 'BOX_TOOLS_CONVERT_AIDBA', 'FILENAME_AIDBA', '', 'tools', 'Y', 20);
+    }
```
drbyte, line 56: We'll put the table structure changes into zc_install. 
But if you also put them here, we can delete the admin-install files (config.aidba.php, init_aidba_install.php) so they're not re-installing upon every admin page reload. Then we rely on the plugin manager install/uninstall to look after the rest.
```suggestion

        // THE FOLLOWING SHOULD ALREADY BE PART OF CORE ZEN CART, SO THIS IS DUPLICATION

        $fields = [
            'configuration_title' => 'Additional Images Approach',
            'configuration_value' => 'legacy',
            'configuration_description' => 'Use Legacy mode for old-school directory scanning and matching filenames. Additional images must be uploaded manually via FTP or using Image Handler.<br>Use Modern mode to add additional images directly from the admin product page. Filenames and extensions do not need to match.<br><strong>NOTE:</strong> if you are switching from Legacy to Modern for the first time, you can use the converter tool found in your Tools menu to update all existing product data with existing additional images. Switching back and forth is not recommended.',
            'configuration_group_id' => 4,
            'sort_order' => 100,
            'set_function' => 'zen_cfg_select_option([\'legacy\', \'modern\'], ',
        ];
        $this->addConfigurationKey('ADDITIONAL_IMAGES_APPROACH', $fields);


//        // alter product table to InnoDB if not already done
//        $sql = "ALTER TABLE " . TABLE_PRODUCTS . " ENGINE=InnoDB";
//        $this->executeInstallerSql($sql);
//
//        // create products_additional_images table
//        $sql = "CREATE TABLE IF NOT EXISTS " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " (
//            id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
//            products_id INT(11) NOT NULL,
//            additional_image VARCHAR(255) NOT NULL,
//            sort_order INT(11) DEFAULT 0,
//            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
//            FOREIGN KEY (products_id) REFERENCES " . TABLE_PRODUCTS . "(products_id) ON DELETE CASCADE
//        ) ENGINE=InnoDB";
//        $this->executeInstallerSql($sql);

        // create products_additional_images table
        $sql = "CREATE TABLE IF NOT EXISTS " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " (
            id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
            products_id INT(11) NOT NULL,
            additional_image VARCHAR(255) NOT NULL,
            sort_order INT(11) DEFAULT 0,
            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
            INDEX idx_products_id (products_id)
        ) ENGINE=MyISAM";
    }
```

drbyte: Putting it in here also in theory lets users of older ZC versions potentially install this plugin there too, and not need to do the SQL updates by hand.  They'd need to manually update their core files to "use" the db data though.

zenexpert: was `$this->executeInstallerSql($sql);` omitted on purpose (line 48-49)? I've added just in case

drbyte: Oops. You're right. Bad copy-paste. Good catch.

includes/modules/additional_images.php
piloujp: > Using ` $images_array[] = substr($image['additional_image'], (strpos($image['additional_image'], '/') ?? -1) + 1);` actually strips the first character of the image name when it's not in a subdirectory. So, for example, 1_small_00.jpg becomes _small_00.jpg I've went for a somewhat different solution and it seems to work for both subdirs and main dir.

No, it does not because the -1 value which is set in that case. It's a little trick and I have tested it before posting. :)
But if you have your own working solution, then it is as good or better.

> As for `$products_image_extension = substr($file, strrpos($file, '.'));` - OK, taking your word for it.

Thanks!

drbyte:
I switched from `@dir()` iteration to calling `zen_get_files_in_directory()` which uses `glob()` which inherently does an alpha sort.
This is because the dir->read() was inserting images to db in some random-ish order that didn't match the filename-sort that was part of the legacy image pattern.

drbyte:
@piloujp If you have time to retry things using today's latest updates, that'd be helpful.

piloujp:
> @piloujp If you have time to retry things using today's latest updates, that'd be helpful.

The first run of tests is done. Plugin installs without any problems, conversion too and plugin uninstall too.
On products page, additional images are all displaying both in legacy and modern mode.
I just found a problem in modern mode where large images won't display when clicking on thumbnails. In legacy mode they display well.
The fix is easy (I already suggested it), adding this line after line 135 of `includes/modules/additional_images.php` is the best way.
`        $products_image_extension = substr($file, strrpos($file, '.'));`
Then lines 30 and 31 can be removed.
There are other ways to fix this, but it can add limitations on files extension one can use.

Apart that, I wonder if other products types `collect_info.php` should not be modified like the main `collect_info.php` file?

I will do a few more tests and come back here if I found something.

I saw the InnoDB foreign key is not used. It could certainly avoid messing the database with external tools or bad plugins. Is it still experimental?

drbyte:
Apologies. I might have accidentally removed that line if zenexpert had already committed it.

I'm not strongly opposed to the foreign key.

zenexpert:
> Apologies. I might have accidentally removed that line if zenexpert had already committed it.
> 
> I'm not strongly opposed to the foreign key.

No, I actually removed it following lat9's suggestion on keeping it standard MyISAM because the `DELETE` statement would pick it up from `zen_remove_product`. It's up to you to decide which is better. I can't really think of any 3rd party plugin that runs the `DELETE` commands on it's own, but you never know... Your call.

piloujp:
> > Apologies. I might have accidentally removed that line if zenexpert had already committed it.
> > I'm not strongly opposed to the foreign key.
> 
> No, I actually removed it following lat9's suggestion on keeping it standard MyISAM because the `DELETE` statement would pick it up from `zen_remove_product`. It's up to you to decide which is better. I can't really think of any 3rd party plugin that runs the `DELETE` commands on it's own, but you never know... Your call.

I was more thinking of using tools like PhpMyAdmin to delete some rows that would make orphans in other linked tables. This happens easily with country zones, for example. But this is another subject. Anyway, I will certainly experiment with this.

zc_install/sql/install/mysql_zencart.sql
```diff
@@ -2676,6 +2690,7 @@ INSERT INTO configuration (configuration_title, configuration_key, configuration
...
+INSERT INTO configuration (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Additional Images Approach', 'ADDITIONAL_IMAGES_APPROACH', 'modern', 'Use Legacy mode for old-school directory scanning and matching filenames. Additional images must be uploaded manually via FTP or using Image Handler.<br>Use Modern mode to add additional images directly from the admin product page. Filenames and extensions do not need to match.<br><strong>NOTE:</strong> if you are switching from Legacy to Modern for the first time, you can use the converter tool found in your Tools menu to update all existing product data with existing additional images. Switching back and forth is not recommended.', '4', '80', 'zen_cfg_select_option([\'legacy\', \'modern\'], ', NOW())");
```
piloujp: I found an error/a typo at the end of the line, `")` should be removed.

zc_install/sql/install/mysql_zencart.sql
```diff
@@ -1748,6 +1748,20 @@ CREATE TABLE products (
...
+# Table structure for table 'products_additional_images'
+#
+
+CREATE TABLE IF NOT EXISTS products_additional_images (
+  id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  products_id INT(11) NOT NULL,
+  additional_image VARCHAR(255) NOT NULL,
+  sort_order INT(11) DEFAULT 0,
+  created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+  UNIQUE KEY idx_pid_img_zen (products_id, additional_image)
```
piloujp, line 1762: The index is too long, the field length for additional image should be reduced.190 is ok, but could be smaller too, probably 128 is more than enough.

piloujp (commented):
Field size should be modified in SQL upgrade file too.
For the INT types, setting a length (11) is deprecated in actual MySQL version.

zenexpert:
> I was more thinking of using tools like PhpMyAdmin to delete some rows that would make orphans in other linked tables. This happens easily with country zones, for example. But this is another subject. Anyway, I will certainly experiment with this.

If you're using phpMyAdmin to delete records, there's an assumption you know what you're doing. Delete a record from products table and you'll leave products_description and products_to_categories with orphans straight away, and possibly many other tables. I was thinking more like 3rd party software for example handling product feeds and not using the `zen_remove_product` function, but they ARE handling the other tables "internally" because those have been there since forever... But yeah, not the end of the world if it remains MyISAM.

drbyte:
@piloujp - I've read through your feedback and made adjustments.  If you have some time to re-test things to be sure I haven't missed something, that would be helpful.

piloujp:
> @piloujp - I've read through your feedback and made adjustments. If you have some time to re-test things to be sure I haven't missed something, that would be helpful.

@drbyte - I am already on it and for now everything goes well. There is still one thing that I found confusing, it is the `Add Additional Image(s)` button under the drop zone when editing a product. Clicking on the drop zone will do the same as clicking on this button. Is this duplication in case the drop & drag functionality fells? Is it really necessary?

admin/includes/modules/collect_info.php
```diff
@@ -507,18 +518,135 @@
...
+        <h3><?= TEXT_PRODUCTS_ADDITIONAL_IMAGES ?></h3>
+        <?php if(!empty($additional_images)) { ?>
+            <div class="form-group">
+                <div class="col-sm-offset-3 col-sm-9 col-md-6">
+                    <?php foreach($additional_images as $img) { ?>
+                        <div class="col-sm-3 col-md-3 col-lg-3">
+                            <?= zen_info_image($img['additional_image'], (is_array($pInfo->products_name) ? $pInfo->products_name[$_SESSION['languages_id']] : $pInfo->products_name), '', '', 'class="img-responsive"') ?>
+                            <br>
+                            <?= $img['additional_image'] ?><br>
+                            <?= zen_draw_hidden_field('previous_additional_images[]', $img['additional_image'], 'data-img-id="' . $img['id'] . '"') ?>
```
piloujp: Adding the image id here will allow identifying images to be deleted and not displaying them in preview.
`                            <?= zen_draw_hidden_field('previous_additional_images[' . $img['id'] . ']', $img['additional_image'], 'data-img-id="' . $img['id'] . '"') ?>`

admin/includes/modules/preview_info.php
```diff
@@ -102,6 +102,24 @@
...
+
+        // BOF Additional Images
+        if (!empty($additional_images_names) || !empty($pInfo->previous_additional_images)) {
+            echo '<div class="clearfix"></div>';
+            echo '<h4 class="pull-right">' . TEXT_PRODUCTS_ADDITIONAL_IMAGES . '</h4><br>';
+            echo '<div class="clearfix"></div>';
+            echo '<div class="row">';
+            if (!empty($pInfo->previous_additional_images)) {
+                foreach ($pInfo->previous_additional_images as $img) {
+                    echo zen_image(DIR_WS_CATALOG_IMAGES . $img, '', SMALL_IMAGE_WIDTH, SMALL_IMAGE_HEIGHT, 'class="img-thumbnail pull-right"');
```
piloujp: Using the image index here to not display images to be deleted.
```
                foreach ($pInfo->previous_additional_images as $imgindex => $img) {
                    if (empty($pInfo->additional_image_delete) || !array_key_exists($imgindex, $pInfo->additional_image_delete)) {
                        echo zen_image(DIR_WS_CATALOG_IMAGES . $img, '', SMALL_IMAGE_WIDTH, SMALL_IMAGE_HEIGHT, 'class="img-thumbnail pull-right"');
                    }
                }
```

piloujp (commented):
The only problem I could find is that additional images to be deleted are still displaying in preview page.
Please see suggestions in files `admin/includes/modules/collect_info.php` and `admin/includes/modules/preview_info.php` to correct this.
The same corrections need to be applied to other products types.

admin/includes/modules/collect_info.php
drbyte: Would there also be value in specifying a heading like `"The following images will be deleted:"`, and display them below?

piloujp: I think it would. It might give a chance to someone that did not click on the right image delete button to see it before it is actually deleted.

drbyte:
How about something like this:
Screen Shot 2025-09-17 at 1 11 18 PM

and on hover:
Screen Shot 2025-09-17 at 1 11 24 PM

barco57:
ok, I am out of the conversation then

barco57:
Think about the new layout boxes controller with the ability to either drag or click the arrows to sort. Now the ladies I have to deal with never go into the layout boxes controller, so I don't have any feedback on any issues they would have with it. I think they could figure out sorting by dragging and dropping. I would make the primary image for the product the first one and maybe border it specially.

drbyte:
> Think about the new layout boxes controller with the ability to either drag or click the arrows to sort. Now the ladies I have to deal with never go into the layout boxes controller, so I don't have any feedback on any issues they would have with it. I think they could figure out sorting by dragging and dropping. I would make the primary image for the product the first one and maybe border it specially.

Ya, that's sorta what I had in mind. Might explore that for "round two".

drbyte:
I'm gonna merge this. We can refine it from any feedback post-merge.

drbyte:
Argh. Will re-sync the collect_info.php files separately.

PR #7234, opened 2025-09-05
Apply `strip_tags` to protected string

retched:
If a string is coming from an "unprotected" source (example: the customer's order notes), HTML should be stripped out. If it's coming from an admin or the system, for example, extra order notes, delivery confirmations, and so on, then it should be OK for the HTML to pass through.

### DIFF
--- a/includes/functions/functions_strings.php
+++ b/includes/functions/functions_strings.php
@@ -10,7 +10,7 @@
  * Returns a string with conversions for security.
  * @param ?string $string The string to be parsed
  * @param array|bool $translate contains a string to be translated, otherwise just quote is translated
- * @param bool $protected Do we run htmlspecialchars over the string
+ * @param bool $protected Do we run htmlspecialchars and strip_tags over the string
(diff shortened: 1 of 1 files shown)


retched:
> Or simply trying to remove any markup so that just the "text" is shown without tags? (ie: `<a href="link">Click here</a>` converts to just `Click here`)?

More of this than the former. When I ran  status history and posted the comment, it ended up outputting the raw html. (Oddly, it was also a hyperlink.) I figured strip_tags before sending it to htmlspecialchars was the better approach for this.

drbyte:
Okay. I understand your use-case, and agree.

Here's my concern. This is a very heavily used function, used in almost every template file.

Everywhere, both admin and catalog, where we output customer-provided content, we typically call `zen_output_string_protected()`, which in turn calls `zen_output_string()` with the `$protect` param set to `true`.

Due to this function's prolific use, I'm concerned about the unexpected negative side-effects.

PR #7237, opened 2025-09-07
Adding Edit Category Button

retched:
Will add an edit (current) category button to the bottom of a category listing. Will not show up on the root category. (As there are no details to really edit there.) Line 1289-1293


torvista:
How about adding it as the ID at the front of the title to match the category listing:

2025-09-07 09_15_41-Admin Category Product Listing — Firefox Developer Edition

retched:
Wait. Do you mean to add the ID next to the `<h1>` header at the top and use that as a link?

torvista:
Yes.
Although could be some icon:
just to be next to the thing it references and take up little space instead of a big button.

retched:
I'm of the mindset it could be both. I'm also 99% sure I can also use Font Awesome.

drbyte:
One thing to note is we have a max of 12 columns. Currently it's using 3+3+6 = 12, but you've added 1 before the first 3, which could potentially force the last one (the 6) to bump to a new line (but only on small screens in this case).

Could the first of the two `col-sm-3` be reduced to `col-sm-2`  without its content being two wide to display?

Screen Shot 2025-09-07 at 3 38 00 PM

PR #7244, opened 2025-09-08
Notifiers for install/uninstall

### LINKED ISSUE #6847: payment/shipping modules: add notifier post install to update with store-specific parameters
Since sometimes the advice is to uninstall/install modules on an upgrade, perhaps a notifier at the end of the install method could remove the tedium of entering those store-specific values again.

drbyte:
Possible solution to close #6847

### DIFF
--- a/admin/modules.php
+++ b/admin/modules.php
@@ -131,7 +131,7 @@ if (!empty($action)) {
                     $result = $module->install();
                 }
             }
-            $zco_notifier->notify('NOTIFY_ADMIN_MODULES_DO_INSTALL', ['module_name' => $class], $result);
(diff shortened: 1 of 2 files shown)


torvista:
Testing with a payment module, php 8.4.12.

Install:  ok.

Uninstall (with/without an observer attached): 
--> PHP Fatal error: Uncaught Error: base::notify(): Argument #3 ($param2) could not be passed by reference in admin\modules.php:164

drbyte:
Good point. Notifiers params 2-9 are one place where only variables can be used, not fallback values ... because they are passed-by-reference even if the observer class doesn't intentionally "receive"-by-reference.

PR #7246, opened 2025-09-09
Moving `zen_remove_order` to `order.php` class

retched:
Found the function referenced once in admin\includes\functions\general.php. This is a breaking change, I guess. (But I know of only one other module that used it, fraudlabspro, and I converted that to an observer class.)

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -372,6 +372,17 @@ function zen_get_system_information($privacy = false): array
     return $systemInfo;
 }
 
+/**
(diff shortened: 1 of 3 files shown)


retched:
I saw this as an @todo. Figured, while I was about to suggest another change to order.php, let me try to do this one too.

drbyte:
To minimize the backward-compatibility issue, what about keeping `zen_remove_order()` but having it call the order class to do the deletion?  Treat it as a passthru/alias. 

And add an `/** @deprecated v2.2.0 - Use $order->delete() instead. */`, or such, docblock.

PR #7251, opened 2025-09-10
Matching deletion logic to new value of checkbox

retched:
Since the checkbox is now valued as '1', this should check for true.


retched:
> confusion

Changing it to:

``` php
            $order->delete(!empty($_POST['restock']) && $_POST['restock'] === '1');
```

and
``` php
            zen_draw_checkbox_field('restock', '1')
````

Issue #7223, opened 2025-09-03
Pricing performance reduced when using `Product` class


lat9:
As identified in the following postings on the **_Google Product Search Feeder II_**'s Zen Cart support-thread:

- https://www.zen-cart.com/showthread.php/229785-Google-Product-Search-Feeder-II-Support-Thread?p=1408772
- https://www.zen-cart.com/showthread.php/229785-Google-Product-Search-Feeder-II-Support-Thread?p=1408782

The performance of the various product-pricing functions has taken a hit when we converted to using the `Product.php` class (which is called by the  zen_get_product_details).  Perhaps, as suggested in those posts, we revert some of the pricing-related functions to simply grab the `products` fields needed to perform their required action?

proseLA:
product is an object.  reverting to hand coding sql in this spot is a mistake IMO.

lets backup a bit; you are referencing a plugin that creates a feed of products; and now you do not want to use the product object?

how does that make sense?

i am all for increased speed.  but the solution here is NOT to revert back to not using the product object.

lat9:
I'm all for using a product-object, but not at a reduction in performance.  While the issue was originally raised due to the performance degradation in a plugin, the same's going to be true for a site's `products_all` page.

Perhaps a means within the `Product.php` class to statically cache the last product requested and only go through the time-consuming gathering of *all* product information if the requested product's ID has changed?

lat9:
The `products_all` (pretty much any product-listing page) page does bounce between products, and the number displayed depends on a specific site's configuration.  Not as much as GPSF where every qualifying product is interrogated.

Note, too, that declaring the class' `data` array as static and simply checking to see if (a) the array exists and (b) the products_id present in that array matches the one submitted in the constructor isn't (IMO) exceedingly complex.

proseLA:
> The `products_all` (pretty much any product-listing page) page does bounce between products, and the number displayed depends on a specific site's configuration. Not as much as GPSF where every qualifying product is interrogated.
> 
you are missing my point.  if the product page does NOT have any sideboxes or centerboxes and it is just retrieving 1 and only 1 product on the page, the `zen_get_product_details` method is called multiple times for that page on the same product_id.

so on the `products_all` page, when we are dealing with the first product, we are calling that method multiple times while getting the data for that first product.  and the same for the second product, the third, etc.

in my limited testing on a reasonably generic dataset, i count 3 times prior to moving the next `product_id`.

that is ripe for a refactor.

[...]

PR #7252, opened 2025-09-10
add global productLoaded array to products class #7223

proseLA:
drastic performance increase with minimal/no disruption to any old templates.

Ref #7223

### DIFF
--- a/includes/classes/Product.php
+++ b/includes/classes/Product.php
@@ -44,7 +44,8 @@ class Product
 
     public function forLanguage(?int $language_id): self
     {
-        $this->fields = $this->getDataForLanguage($language_id);
+        self::$data = $this->getDataForLanguage($language_id);
+        $this->fields = self::$data;
         unset($this->fields['lang']);
 
         return $this;
@@ -52,7 +53,8 @@ class Product
 
     public function withDefaultLanguage(): self
     {
-        $this->fields = $this->getDataForLanguage();
+        self::$data = $this->getDataForLanguage();
+        $this->fields = self::$data;
         unset($this->fields['lang']);
 
         return $this;


lat9 (commented):
Rather than using a `global` variable, why not declare 
```php
protected static array $data;
``
and then in the constructor check to see that (a) $this->data isn't empty and (b) the submitted product_id matches that previously registered in the class 'data'?

File: includes/classes/Product.php [hunk 1]
```diff
@@ -27,47 +27,43 @@ class Product
...
         $this->initLanguages();
 
         if ($this->product_id !== null) {
-            if (!empty($productLoaded) && $this->product_id === (int)$productLoaded['products_id']) {
-                $this->data = $productLoaded;
-            } else {
-                $this->data = $this->loadProductDetails($this->product_id);
-                $productLoaded = $this->data;
-            }
+            if (empty(self::$data) || (int)self::$data['products_id'] !== $this->products_id) {
```
proseLA: `$this->products_id` does not exist.  should be `$this->product_id`.

includes/classes/Product.php
```diff
@@ -27,47 +27,43 @@ class Product
...
+                self::$data = $this->loadProductDetails($this->product_id);
 
-            // set some backward compatibility properties
-            $this->fields = $this->data;
-            $this->EOF = empty($this->data);
+                // set some backward compatibility properties
+                $this->fields = self::$data;
+                $this->EOF = empty(self::$data);
+            }
```
proseLA, line 37: lines 37 and 38 need to be outside of the conditional, as `$this->fields` and `$this->EOF` are not static.  and they will be reset upon every substantiation of the class.

proseLA:
this is my testing:

myadmin/index.php?cmd=category_product_listing&search=red
core: 
Parse Time: 2.041 - Number of Queries: 7837 - Query Time: 1.51892
Parse Time: 1.768 - Number of Queries: 7837 - Query Time: 1.28968

Commit 1
Parse Time: 0.855 - Number of Queries: 1837 - Query Time: 0.43194
Parse Time: 0.853 - Number of Queries: 1837 - Query Time: 0.4378


products_all (50 products, page 1)
core:
Parse Time: 0.586 - Number of Queries: 546 - Query Time: 0.15643
Parse Time: 0.633 - Number of Queries: 546 - Query Time: 0.15277
Parse Time: 0.623 - Number of Queries: 545 - Query Time: 0.15715

commit 1:
Parse Time: 0.504 - Number of Queries: 546 - Query Time: 0.10454
Parse Time: 0.503 - Number of Queries: 546 - Query Time: 0.10588

i have not tested @lat9 commits using the static data, as i have not been able to reproduce the error that you 2 say you see with regards to the wrong name/description.  i do not know what page you are seeing that on.

if someone could tell me how to reproduce that, i am now curious.

with regards to testing the static data approach, i am more game to see this error.  the difference between the static approach and my approach from a time standpoint is neglible.

with regards to the gpsf feeder, my testing consistently showed a 75% reduction with both the static approach as well as my approach.

@torvista i can not replicate your numbers at all.  i emailed you earlier (i think) to get a copy of your test db; you can find my email at my website: mxworks.cc

includes/classes/Product.php
proseLA: this has not been done properly. you have moved the code outside of both conditionals, and added a null coalesce.  the null coalesce is NOT necessary if these lines are placed within the conditional of `!== null` as was done in the original code.

includes/classes/Product.php
```diff
@@ -25,43 +26,48 @@ class Product
...
         return $this;
     }
 
     public function withDefaultLanguage(): self
     {
-        $this->data = $this->getDataForLanguage();
-        $this->fields = $this->data;
+        self::$data = $this->getDataForLanguage();
+        $this->fields = self::$data;
+        unset($this->fields['lang']);
```
proseLA, line 58: both of these return values are now different from what they were in the original code because of lines 49 and 58.  both of these methods are getting the language values for product name and description.  once the methods  retrieve these values, you now unset them prior to returning them.  anyone making use of  the array `$this->fields['lang']` in the return from either of these methods is in for a rude awakening as the values are no longer there.  i see that the unset was previously done somewhere else (line 90), but until one can provide details on the specifics of this bug, i am not for implementing the change.

proseLA (commented):
i am not a fan, especially considering i presented 2 working viable solutions that limited changes to this class to 1 method and 1 new var (not counting the memcache solution).  now we have changing return values.  

considering all of the refactoring suggestions presented in #7223, i'm not sure about how much more effort i can put into what this PR has now become.

includes/classes/Product.php
lat9: If you'll inspect the original version of the class, the return value from `getDataForLanguage` was removing the 'lang' element of the array.  Rather than removing all language variations from the protected array, the two additional methods simply remove the 'lang' array from the publicly-available 'fields' array.

proseLA:
done.

PR #7255, opened 2025-09-11
Enable "account activation"

### LINKED ISSUE #7235: Add authorization link-back to aid in the prevention of fraudulent accounts
Nefarious actors will create an account with a bogus, non-existent, email address with the intention to determine whether credit-cards they've stolen are still valid.  One way to address this issue is to send a verification email to the address submitted and not enable checkout until the link in that email has been clicked.  Note that a similar problem is present when a customer mistypes their email address in the form entry, since they'll never receive communication from the store.

This feature is somewhat based on the storefront password-reset handling introduced post-zc210 release in [this](https://github.com/zencart/zencart/commit/9806d3078114e4ab1f4e7a09c92a8b2dcaf56424) GitHub commit.

Following is a description of how I envision this feature working; comments welcomed.  I'll come back and update this description as needed so that it's usable as documentation.

## High-Level Description

*When the feature is enabled*, a valid `create_account` submission (i.e. all form fields' values are acceptable) requires activation prior to checkout.  The operation is similar to Zen Cart's current handling of customers' password-forgotten requests.

Upon the validation of the fields in a `create_account` form, the information is stored into the database, creating a customer record.  The customer's current authorization status is set to '3' (Pending, can browse with prices but cannot purchase until authorized) and the create-account email contains an account-authorization link which the customer must click within a configured number of minutes to activate their account for purchasing. That link includes a database token, tied to the just-created customer's ID, which is used to validate/verify that the customer "owns" the email address supplied.

##### Customer Details :: Account Activation Token Length

This setting, using the key `CUSTOMERS_ACTIVATION_TOKEN_LENGTH`, identifies the number of characters in a generated account-activation token, defaulting to 24.  Allowed values are 12 to 100.

**Note**: The maximum value (100) is the length of the newly-added `customers_auth_tokens::token` field.

##### Customer Details :: Account Activation Token Valid For

This setting, using the key `CUSTOMERS_ACTIVATION_TOKEN_MINUTES_VALID`, identifies the number of minutes that an account-activation token is valid, defaulting to 60 minutes (1 hour). Allowed values are 1 to 1440, with the 'best' values in the range 60 to 120.

##### Customer Details :: Require Account Activation

This setting, using the key `CUSTOMERS_ACTIVATION_REQUIRED`, identifies whether ('true') or not ('false', *the default*) a customer's account creation requires the click of an activation link sent to the email address used in the account creation.  When this setting's value is 'true', the value for ***Customer Approval Status - Authorization Pending*** (`CUSTOMERS_APPROVAL_AUTHORIZATION`) is forced to '3' (customer must be authorized to buy) and ***Customer Authorization Filename*** (`CUSTOMERS_AUTHORIZATION_FILENAME`) is forced to `customers_authorization`.

#### `customers_auth_tokens` (New Table)

This new database table stores the ***most recent*** token for a customer's account-activation.  It's created via the following SQL:

```
CREATE TABLE customers_auth_tokens (
    customers_id int(11) NOT NULL,
    email_address varchar(96) NOT NULL,
    token varchar(100) NOT NULL default '',
    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY (customers_id)
);
```

##### `activation_required`

This field, added to the `customers` table after the `customers_authorization` field, is a binary `tinyint(1)` value indicating whether (1) or not (0, the default) this customer's account requires activation.  The default value of 0 applies to any pre-existing account records.

##### `welcome_email_sent`
This field, added to the `customers` table after the `activation_required` field is a **nullable** binary `tinyint(1)` value indicating whether (1) or not (0, the default for *new* accounts) this customer has received an account-creation "welcome" email. 

The default value (`NULL`) is applied to any pre-existing customer records on an upgrade and indicates that it's not-known whether the welcome email was sent.

#### Configuration (`configuration.php`)

The admin **Configuration** tool is updated to force `CUSTOMERS_APPROVAL_AUTHORIZATION` to '3' and `CUSTOMERS_AUTHORIZATION_FILENAME` to 'customers_authorization' if `CUSTOMERS_ACTIVATION_REQUIRED` is set to 'true'.

#### Customers :: Customers (`customers.php`)

On the customers' listing, an additional color (using `text-warning`) is added to the **Authorized** column if the customer's account is awaiting email validation.  The hover-text will show as 'Awaiting Verification'.

When editing a customer's record, a checkbox is added underneath the **E-Mail Address** to indicate whether or not the account has been activated (i.e. the email address has been verified). An admin can, thus, override a known customer's lack of response to the email sent when they created their account or changed their email address.

### Storefront Changes

The storefront changes affect the processing and template-rendering for the account-creation, email-address change and customers' authorization pages.

#### Account Creation (`/modules/create_account.php`)

If the account-verification feature is enabled:

1.  The welcome-email normally sent during the create-account processing is deferred until the customer has clicked the verification link sent to the account's email address.  As such, the email-sending section of the current `create_account.php` module is split out into another module (`create_account_send_email.php`) so the code can be reused and, instead, sends an email containing the account-activation link.  That processing is in another new module (`send_account_activation_email.php`) so that it, too, can be reused.
2. The `Customer.php` class 'owns' the management of the account-activation tokens present in the `customers_auth_tokens` table.  On creation of an account that requires verification, that class sets the `customers::activation_required` field to 1.
3. Upon creation of the customer's account in the database, the customer is redirected to the `customers_authorization` page instead of `create_account_success`.

#### Customers Authorization (`/modules/pages/customers_authorization`)

This page's operation depends on whether the account-verification feature is enabled and, if so, its access is accompanied by an authorization token.

1. If no customer is logged in or, if the customer's current authorization status (`customers_authorization`) is neither **1** (Pending, must be authorized to browse) nor **3** (Pending, can view prices but can't shop until authorized), the page redirects back to the site's home-page.

2. If the customer's authorization-status is **1** or the account-verification feature is not enabled, the page displays a button that the customer can click to see if their authorization-status has changed.

3. Otherwise, the customer's authorization-status is **3** and the account-verification feature is enabled.

   1. If the page's access is not accompanied by an authorization token, the customer's account is checked to see if there's an active authorization token.
      1. If there's an active authorization token, the page displays a message instructing the customer to check their email and click on the link in the previously-sent email message.  If that token has been active for more than half of its lifetime, an additional button is displayed which, if clicked, sends another account-activation email to the customer's email address.  The sending of that new email invalidates any account-activation tokens previously sent.
   2. Otherwise, the token is checked to see that it's valid for the customer (i.e. associated with the customer and hasn't timed out).
      1. If so, the customer's authorization status is updated to '0' (Authorized) and the customer can now add products to their cart and checkout. If the site is enabled to send a welcome-email on account-creation and this is the first time the customer has been authorized, the welcome-email is sent to the customer and the customer is redirected to the `create_account_success` page; otherwise, the customer is redirected to the `shopping_cart` page.
      2. Otherwise, if the customer has an active account-activation token, the page displays a message instructing the customer to check their email and click on the link in the previously-sent email message.  If that token has been active for more than half of its lifetime, an additional button is displayed which, if clicked, sends another account-activation email to the customer's email address.  The sending of that new email invalidates any account-activation tokens previously sent.
      3. Otherwise, a message is displayed to the customer indicating that the supplied token isn't valid (or has timed out) along with a button that the customer can click to send another account-activation email.

   #### Customers Account Edit (`/modules/pages/account_edit`)

   If a customer edits their account and changes their email address, their authorization-status changes to **3** and an account-activation email is sent to their updated email address.  The customer is redirected to the `customers_authorization` page.

lat9:
Fixes #7235

This PR introduces a couple of other features that were required for the account-authorization changes
- Common-use language files can now reside in `/includes/languages/{language}/modules` directly.  This change was needed for the now common processing to send a create-account welcome email and to send the account-activation-required email.
- If a customer's account isn't banned but is waiting for any type of authorization, the "Pending Authorization" display on product add-to-cart/listing displays is now replaced with a link to the `customers_authorization` page.  Before, there wasn't a "good" way for a customer to view their current status.
- The admin's `customers.php` now uses a dropdown containing *only* the authorization statuses that apply based on the value set in the `Customers Info :: Customer Approval Status - Authorization Pending`.

### DIFF
--- a/includes/languages/english/modules/lang.create_account_send_email.php
+++ b/includes/languages/english/modules/lang.create_account_send_email.php
@@ -1,4 +1,13 @@
 <?php
+/**
+ * @package languageDefines
+ * @copyright Copyright 2003-2025 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version GIT: $Id: Author: lat9  Wed Sep 24 23:03:50 2013 -0500 Modified in v2.2.0 $
+ *
+ * @since ZC v2.2.0
+ */
 $define = [
     'EMAIL_SUBJECT' => 'Welcome to ' . STORE_NAME,
     'EMAIL_GREET_MR' => 'Dear Mr. %s,' . "\n\n",
--- a/includes/languages/english/modules/lang.send_auth_token_email.php
+++ b/includes/languages/english/modules/lang.send_auth_token_email.php
@@ -1,4 +1,13 @@
 <?php
+/**
+ * @package languageDefines
+ * @copyright Copyright 2003-2025 Zen Cart Development Team
+ * @copyright Portions Copyright 2003 osCommerce
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version GIT: $Id: Author: lat9  Wed Sep 24 23:03:50 2013 -0500 Modified in v2.2.0 $
+ *
+ * @since ZC v2.2.0
+ */
 $define = [
     'EMAIL_AUTH_TOKEN_SUBJECT' => STORE_NAME . ' - Activate Account',
     'EMAIL_AUTH_TOKEN_BODY' => "To activate your account, please click the link below or copy and paste the entire link into your browser:\n\n%1\$s\n\nThis link expires in %2\$u minutes.",
--- a/zc_install/sql/install/mysql_zencart.sql
+++ b/zc_install/sql/install/mysql_zencart.sql
@@ -633,8 +633,8 @@ CREATE TABLE customers (
   customers_group_pricing int(11) NOT NULL default '0',
   customers_email_format varchar(4) NOT NULL default 'TEXT',
   customers_authorization int(1) NOT NULL default '0',
-  activation_required tinyint(1) NOT NULL default 0,
-  welcome_email_sent tinyint(1) default NULL,
+  activation_required tinyint(1) NOT NULL DEFAULT 0,
+  welcome_email_sent tinyint(1) DEFAULT NULL,
   customers_referral varchar(32) NOT NULL default '',
   registration_ip varchar(45) NOT NULL default '',
   last_login_ip varchar(45) NOT NULL default '',
@@ -658,7 +658,7 @@ DROP TABLE IF EXISTS customers_auth_tokens;
 CREATE TABLE customers_auth_tokens (
     customers_id int(11) NOT NULL,
     email_address varchar(96) NOT NULL,
-    token varchar(100) NOT NULL default '',
+    token varchar(100) NOT NULL DEFAULT '',
     created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
     PRIMARY KEY (customers_id)
 );
--- a/zc_install/sql/updates/mysql_upgrade_zencart_220.sql
+++ b/zc_install/sql/updates/mysql_upgrade_zencart_220.sql
@@ -41,7 +41,7 @@ DROP TABLE IF EXISTS customers_auth_tokens;
 CREATE TABLE customers_auth_tokens (
     customers_id int(11) NOT NULL,
     email_address varchar(96) NOT NULL,
-    token varchar(100) NOT NULL default '',
+    token varchar(100) NOT NULL DEFAULT '',
     created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
     PRIMARY KEY (customers_id)
 );
@@ -59,8 +59,8 @@ ALTER TABLE currencies MODIFY code char(4) NOT NULL default '';
 ALTER TABLE orders MODIFY currency char(4) default NULL;
 ALTER TABLE plugin_control MODIFY `version` varchar(20);
 ALTER TABLE plugin_control_versions MODIFY `version` varchar(20);
-ALTER TABLE customers ADD COLUMN activation_required tinyint(1) NOT NULL default 0 AFTER customers_authorization;
-ALTER TABLE customers ADD COLUMN welcome_email_sent tinyint(1) default NULL AFTER activation_required;
+ALTER TABLE customers ADD COLUMN activation_required tinyint(1) NOT NULL DEFAULT 0 AFTER customers_authorization;
+ALTER TABLE customers ADD COLUMN welcome_email_sent tinyint(1) DEFAULT NULL AFTER activation_required;
 
 #PROGRESS_FEEDBACK:!TEXT=Updating configuration settings...
 DELETE FROM configuration WHERE configuration_key IN ('REPORT_ALL_ERRORS_ADMIN', 'REPORT_ALL_ERRORS_STORE', 'REPORT_ALL_ERRORS_NOTICE_BACKTRACE');


lat9:
I'm thinking that I'll need to add yet another flag to a customer's record to indicate that the welcome email has been sent ... just so you know!

piloujp:
There is something else that I found a little weird... When you click on the email link for activation, the account settings page opens in a new window, but at the same time, the page where you requested the activation email is still there with its countdown for the token still going on. It might be confusing if you missed the new page and could think it is not working, which is easy to do on a smartphone.
Is it not possible in email to link to the already opened page, at least same URL and stop countdown and then redirect from there to account page?

lat9:
I'll look at that, but I'm not sure how possible it is for an external link to come back to an already-opened link.

I could also look to add test like "Once you click that link, you can close this browser window.".

piloujp:
> > > I'll look at that, but I'm not sure how possible it is for an external link to come back to an already-opened link.
> > > I could also look to add test like "Once you click that link, you can close this browser window.".
> > 
> > 
> > I don't think it is possible... The activation page would have to be closed, then email link will go to account page where, if activation is done would just be as now (with perhaps an indication that the account is activated), but if activation failed then the account page would have a link to the activation request page with countdown and other useful links to try activation again.
> 
> Perhaps [this](https://www.geeksforgeeks.org/html/html-dom-window-name-property/) could prove useful.

Yes, definitely, I think the window (tab) name can be used in the `target` property of an HTML link. In this case, when the new page opens it could check for the named window and use it, or just create a new windows if it is not there.

piloujp:
Do the same after the countdown is over.

lat9:
> Do the same after the countdown is over.

I did.  After logging out, I clicked on the link in the email.  That took me to the login page; from there I logged in and was redirected to the `customers_authorization` page which displayed with a message that the link had expired.  No logs.

piloujp:
> > Do the same after the countdown is over.
> 
> I did. After logging out, I clicked on the link in the email. That took me to the login page; from there I logged in and was redirected to the `customers_authorization` page which displayed with a message that the link had expired. No logs.

I can't reproduce it anymore... Well, there is a way, but it should not happen in normal use. I changed an activated user account back to 'activation needed' with PhpMyAdmin for testing, and in this case the log is there because there is no more token in the database.
Whether it was possible to get in that situation few commits away or I did the same manipulation, I am not sure, but now the problem is gone.

includes/classes/Customer.php
```diff
@@ -590,22 +748,60 @@ public function banCustomer(): void
...
+
+        $sql =
+            "UPDATE " . TABLE_CUSTOMERS . "
+                SET customers_authorization = " . (int)self::AUTH_OK . ",
+                    activation_required = 0
+              WHERE customers_id = " . (int)$customers_id;
+        $db->Execute($sql, 1);
+
+        self::clearAuthTokens($customers_id);
+
```
piloujp: Line 789 has some space characters.

includes/modules/pages/customers_authorization/jscript_countdown.php
```diff
@@ -0,0 +1,43 @@
...
+            // Format hours, minutes and seconds to always have two digits
+            const formattedHours = String(hours).padStart(2, '0');
+            const formattedMinutes = String(minutes).padStart(2, '0');
+            const formattedSeconds = String(seconds).padStart(2, '0');
+
+            display.textContent = `${formattedHours}:${formattedMinutes}:${formattedSeconds}`;
+
+            if (--timer < 0) {
+                clearInterval(countdownInterval);
+                display.textContent = "<?= TEXT_EXPIRED ?>";
```
piloujp: Replacing line 35 by the code below will reload the page when timer ends and update it if account is activated:
`                window.location.replace("<?= zen_href_link(CUSTOMERS_AUTHORIZATION_FILENAME, '', 'SSL') ?>");`

piloujp (commented):
I could not find any problem. I am still not happy with the activation page with its countdown running even after activation, but I couldn't find any solution to this. The best I could do was to redirect to account page when the countdown ends and account is activated rather than let the page with the expiration message.

includes/modules/pages/customers_authorization/jscript_countdown.php
lat9: Perfect!  That's a great idea.

PR #7260, opened 2025-09-12
Changing tooltip for edit order

retched:
Could be confusing having two "edit" references (one from innate ZenCart and the other from something like Edit Orders) next to each other. Changing the link to the order itself to read as "Details" using an already existing define.

### DIFF
--- a/admin/includes/functions/html_output.php
+++ b/admin/includes/functions/html_output.php
@@ -234,7 +234,8 @@ $iconMap = [
   'star-shadow' => 'fa-star txt-gold star-shadow',
   'locked' => 'fa-lock',
   'unlocked' => 'fa-lock-open',
-  'loading' => 'fa-gear fa-spin'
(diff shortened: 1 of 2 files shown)


drbyte:
Before:
Screen Shot 2025-09-12 at 9 05 52 PM

After:
Screen Shot 2025-09-12 at 9 06 41 PM

retched:
> Before: Screen Shot 2025-09-12 at 9 05 52 PM
> 
> After: Screen Shot 2025-09-12 at 9 06 41 PM

I apologize, I might have had a messup with the Super Orders. It literally looks like two wrenches side by side.

retched:
Ah. I see what I did wrong here, I never touched the `zen_icon()` underneath.

drbyte:
Screen Shot 2025-09-16 at 11 39 34 AM

PR #7271, opened 2025-09-15
admin: coupon_admin

torvista:
(already touched in 220)
short array syntax
remove some parameter default values
use null coalesce ??
use str_ends_with
use ternary operator
no code changes

### DIFF
--- a/admin/coupon_admin.php
+++ b/admin/coupon_admin.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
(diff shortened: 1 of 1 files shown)


lat9:
Arggh, that's a humongous script!  Thanks for the update.

torvista:
phpstorm admits regex in Search and Replace

```
<\?php echo (.*); \?>
<?= $1 ?>
```

so not so painful for <?= as it may seem.

PR #7273, opened 2025-09-15
Adding function to display the group comment

retched:
Simply adding a function to display the Group Comment field from the DB. This can be handy if you want a "friendly" display of what groups a customer belongs to rather than a potentially cryptic `group_name` field.

### DIFF
--- a/includes/functions/functions_customer_groups.php
+++ b/includes/functions/functions_customer_groups.php
@@ -269,8 +269,8 @@ function zen_get_customer_group_comment(string $group_name): string
 
     $sql = "SELECT group_comment FROM " .
             TABLE_CUSTOMER_GROUPS . "
-            WHERE group_name = ':group_name:'";
-    $sql = $db->bindVars($sql, ':group_name:', $group_name, 'string');
+            WHERE group_name = :group_name:";
+    $sql = $db->bindVars($sql, ':group_name:', $group_name, 'stringIgnoreNull');
 
     $results = $db->Execute($sql, 1);
 
--- a/includes/functions/functions_customer_groups.php
+++ b/includes/functions/functions_customer_groups.php
@@ -260,3 +260,17 @@ function zen_delete_customer_group($group_id, $also_unassign_customers = true)
     return true;
 }
 
+/**
+  * @param string $group_name The name of the group to be retrieved (ie. the value from the group_name column of the DB)
+ */
+function zen_get_customer_group_comment(string $group_name) {
+    global $db;
+
+    $sql = "SELECT group_comment FROM " .
+            TABLE_CUSTOMER_GROUPS . "
+            WHERE group_name = '" . $group_name ."'";
+
+    $results = $db->Execute($sql);
+
+    return $results->fields['group_comment'] ?? '';
+}
--- a/includes/functions/functions_customer_groups.php
+++ b/includes/functions/functions_customer_groups.php
@@ -263,7 +263,8 @@ function zen_delete_customer_group($group_id, $also_unassign_customers = true)
 /**
   * @param string $group_name The name of the group to be retrieved (ie. the value from the group_name column of the DB)
  */
-function zen_get_customer_group_comment(string $group_name) {
+function zen_get_customer_group_comment(string $group_name)
+{
     global $db;
 
     $sql = "SELECT group_comment FROM " .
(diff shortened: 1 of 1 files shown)


retched:
In the event a store owner wants to hide the group from the customer's view, say it's an internal only list that the customer need not be told about, one way around this would be to include some kind of character (such as `!` or `_hidden`) in the name then modify `zen_groups_customer_belongs_to()` to include a logic test behind whether or not to display the name.

includes/functions/functions_customer_groups.php
```diff
@@ -260,3 +260,17 @@ function zen_delete_customer_group($group_id, $also_unassign_customers = true)
     return true;
 }
 
+/**
+  * @param string $group_name The name of the group to be retrieved (ie. the value from the group_name column of the DB)
+ */
+function zen_get_customer_group_comment(string $group_name) {
```
lat9: Opening brace needs to be on a new line; add return hint of `string`

Why are you using the group-name instead of the group-id?

retched: So some modules actually use the group name to determine if a right is granted. Like carlwhat's Purchase Orders needs the name of the group to start with `PO_` that made it sound like it's meant to be internal versus displayed.

lat9: OK, if that's the case, then the input `$group_name` should be sanitized via a `$db->bindVars` call prior to the SQL execution.

includes/functions/functions_customer_groups.php
```diff
@@ -260,3 +260,18 @@ function zen_delete_customer_group($group_id, $also_unassign_customers = true)
     return true;
 }
 
+/**
+  * @param string $group_name The name of the group to be retrieved (ie. the value from the group_name column of the DB)
+ */
+function zen_get_customer_group_comment(string $group_name)
```
lat9: Add a return type-hint:
`function zen_get_customer_group_comment(string $group_name): string`

includes/functions/functions_customer_groups.php
```diff
@@ -260,3 +260,18 @@ function zen_delete_customer_group($group_id, $also_unassign_customers = true)
...
+{
+    global $db;
+
+    $sql = "SELECT group_comment FROM " .
+            TABLE_CUSTOMER_GROUPS . "
+            WHERE group_name = '" . $group_name ."'";
```
lat9: Need to sanitize the string prior to database use:
```php
 $sql = "SELECT group_comment FROM " .
            TABLE_CUSTOMER_GROUPS . "
            WHERE group_name = :groupName:";
$sql = $db->bindVars($sql, ':groupName', $group_name, 'stringIgnoreNull');
```

includes/functions/functions_customer_groups.php
```diff
@@ -260,3 +260,18 @@ function zen_delete_customer_group($group_id, $also_unassign_customers = true)
...
+
+    $results = $db->Execute($sql);
```
lat9: Limit the SQL-select to a single record
`$results = $db->Execute($sql, 1);`

retched: Would placing `LIMIT 1` into the SQL achieve the same result or is one preferred over the other?

includes/functions/functions_customer_groups.php
retched: Done

includes/functions/functions_customer_groups.php
lat9: Either way, no preference here.

retched: Made that change as requested. I also realized I could just change that module to use the `zen_cfg_select_customer_group` module but there might be some other use for it for other cases.

includes/functions/functions_customer_groups.php
```diff
@@ -272,7 +272,7 @@ function zen_get_customer_group_comment(string $group_name)
             WHERE group_name = ':group_name:'";
     $sql = $db->bindVars($sql, ':group_name:', $group_name, 'string');
```
lat9: Since you've included the single-quote around `:group_name:`, you'll need to use `'noquotestring'` since 'string' returns a value surrounded by single-quotes.

retched: Would deleting the surrounding quotes be sufficient too?

lat9: Yes, but if you remove the quotes, you should use the 'stringIgnoreNull' type that I'd suggested earlier since the `group_name` field can't be `null`.

retched: So that creates:
``` php

    $sql = "SELECT group_comment FROM " .
            TABLE_CUSTOMER_GROUPS . "
            WHERE group_name = :group_name:";
    $sql = $db->bindVars($sql, ':group_name:', $group_name, 'stringIgnoreNull');

```

Will that suffice?

PR #7274, opened 2025-09-15
POSM observer: mainly hints

torvista:
apart from getOptionsStockRecord

### DIFF
--- a/zc_plugins/POSM/v6.1.1/catalog/includes/classes/observers/class.products_options_stock_observer.php
+++ b/zc_plugins/POSM/v6.1.1/catalog/includes/classes/observers/class.products_options_stock_observer.php
@@ -63,7 +63,8 @@ class products_options_stock_observer extends base
         );
 
         // -----
-        // Load the storefront common functions.  For the previous, unencapsulted, versions of the plugin, this
+        // Load the storefront common functions.
+        // For the previous, unencapsulated, versions of the plugin, this
         // file was loaded by the admin's extra_functions file.
         //
         require $this->getZcPluginDir() . DIR_WS_FUNCTIONS . 'products_options_stock_functions.php';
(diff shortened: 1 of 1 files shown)


zc_plugins/POSM/v6.1.1/catalog/includes/classes/observers/class.products_options_stock_observer.php
```diff
@@ -27,7 +27,7 @@ class products_options_stock_observer extends base
     protected string $stock_message;
     protected string $products_name_saved;
     protected bool $use_mb;
-    protected string $zcPluginDir;
+    protected array|string $zcPluginDir;
```
lat9: What makes you think that this value could ever be an array?  It's initialized with the PluginManager's `getPluginVersionDirectory` method which will return a string if the plugin's installed (which it is if this code's running).

zc_plugins/POSM/v6.1.1/catalog/includes/classes/observers/class.products_options_stock_observer.php
```diff
@@ -523,7 +524,7 @@ protected function notify_html_head_end(&$class, string $e)
     // -----
     // Return the plugin's currently-installed zc_plugin directory for the catalog.
     //
-    public function getZcPluginDir()
+    public function getZcPluginDir(): array|string
```
lat9: As mentioned above, the returned value will *never* be an array.

zc_plugins/POSM/v6.1.1/catalog/includes/classes/observers/class.products_options_stock_observer.php
torvista: str_replace on line 63 can return an array.

lat9: Not if the input is a string.

PR #7284, opened 2025-09-16
Adding a helper function to get an ID from a name

retched:
Basic function to return the ID number of a given group when searched. Will return -1 if not found as 0 is assigned to all customers in a DB by default.

### DIFF
--- a/includes/functions/functions_customer_groups.php
+++ b/includes/functions/functions_customer_groups.php
@@ -262,7 +262,7 @@ function zen_delete_customer_group($group_id, $also_unassign_customers = true)
 
 /**
   * @param string $group_name The name of the group to be retrieved (ie. the value from the group_name column of the DB)
- */
+  */
 function zen_get_customer_group_comment(string $group_name): string
 {
     global $db;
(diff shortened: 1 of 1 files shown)


includes/functions/functions_customer_groups.php
```diff
@@ -276,3 +276,17 @@ function zen_get_customer_group_comment(string $group_name): string
...
+ * @return int Will return the ID number or -1 if not found.
+ */
+function zen_get_customer_group_id_from_name(string $group_name): int
+{
+    global $db;
+
+    $sql = "SELECT group_id FROM " . TABLE_CUSTOMER_GROUPS . " WHERE group_name = :group_name:";
+    $sql = $db->bindVars($sql, ':group_name:', $group_name, 'stringIgnoreNull');
+
+    return $db->Execute($sql, 1)->fields['group_id'] ?? -1; // Don't assign 0, that's the "Everyone" default, -1 should NEVER match
```
lat9: If the group name's found, the group_id that comes back is going to be a 'string' type.

retched: Got it. I went ahead and cast the result to an integer by way of:

``` php
(int)($db->Execute($sql, 1)->fields['group_id'] ?? -1);
```

In the event nothing is found, it will still coalesce to `-1`, then that is returned.

lat9 (approved):

PR #7292, opened 2025-09-18
Use `zen_define_default` instead of `define` when creating metatags

lat9:


scottcwilson:
No objections, but why not lines 21-23 as well?

lat9:
Good point; update forthcoming.

PR #7298, opened 2025-09-19
Add @since annotations to functions and classes

drbyte:
The `@since` annotation denotes when the function or class was added to the core code. Useful for assessing compatibility.


torvista:
aaargh...could these mass no-change file-changes not be lumped with the file's date/id update on a release.

lat9:
The only 'problem' with that deferral is that all function/class files will be marked based on this PR's acceptance, i.e. showing `drbyte` as the last updater.

torvista:
not sure I follow, but do whatever you think is best, please ignore my individual complaining.

scottcwilson:
@drbyte the longer you wait the more conflicts you'll have to fix.   I'd say bite the bullet.

PR #7303, opened 2025-09-21
Plugin Manager - Listing by status

piloujp:
List plugins in different tables depending on their status.
Capture d'écran 2025-09-21 183424

### DIFF
--- a/admin/includes/templates/table_view.php
+++ b/admin/includes/templates/table_view.php
@@ -32,11 +32,35 @@ use Zencart\Paginator\LaravelPaginator;
 
     <div class="row">
         <div class="col-xs-12 col-sm-12 col-md-9 col-lg-9 configurationColumnLeft">
+        <?php
+            for($i = 0; $i < 3; $i++) {
+                $firstheader = 0;
+                $skip = 1;
+                foreach ($formatter->getTableData() as $tableData) {
+                    if ($tableData ["status"] ["original"] === $i) {
+                        $skip = 0;
+                        break;
+                    }
+                }
+                if ($skip === 0) {
+        ?>
             <table class="table table-hover">
                 <thead>
                 <tr class="dataTableHeadingRow">
-                    <?php foreach ($formatter->getTableHeaders() as $colHeader) { ?>
-                        <th class="<?php echo $colHeader['headerClass']; ?>"><?php echo $colHeader['title'];
+                    <?php $firstheader = 0;
+                        foreach ($formatter->getTableHeaders() as $colHeader) { ?>
+                        <th class="<?php echo $colHeader['headerClass']; ?>">
+                        <?php if ($firstheader === 0) {
+                            $tabletitle = match($i) {
+                                0 => TEXT_NOT_INSTALLED,
+                                1 => TEXT_INSTALLED_ENABLED,
+                                2 => TEXT_INSTALLED_DISABLED,
+                            };
+                            echo $tabletitle;
+                            $firstheader = 1;
+                            } else {
+                                echo $colHeader['title'];
+                            }
                             ?></th>
                     <?php } ?>
                     <th class="dataTableHeadingContent text-right"><?php echo TABLE_HEADING_ACTION; ?></th>
@@ -44,24 +68,28 @@ use Zencart\Paginator\LaravelPaginator;
                 </thead>
                 <tbody>
                 <?php foreach ($formatter->getTableData() as $tableData) { ?>
-                    <?php if ($formatter->isRowSelected($tableData)) { ?>
+                    <?php if ($tableData ["status"] ["original"] === $i) {
+                              if ($formatter->isRowSelected($tableData)) { ?>
                         <tr id="defaultSelected" class="dataTableRowSelected" onclick="document.location.href='<?php echo $formatter->getSelectedRowLink(
                             $tableData); ?>'" role="button">
-                    <?php } else { ?>
+                        <?php } else { ?>
                         <tr class="dataTableRow" onclick="document.location.href='<?php echo
                         $formatter->getNotSelectedRowLink($tableData); ?>'"
                         role="button">
-                    <?php } ?>
-                    <?php foreach ($tableData as $column) { ?>
+                        <?php } ?>
+                        <?php foreach ($tableData as $column) { ?>
                         <td class="<?php echo $column['class']; ?>">
                             <?php echo $column['value']; ?>
                         </td>
-                    <?php } ?>
-                    <?php require(DIR_WS_TEMPLATES . 'partials/tableview_rowactions.php'); ?>
-                    </tr>
-                <?php } ?>
+                        <?php } ?>
+                        <?php require(DIR_WS_TEMPLATES . 'partials/tableview_rowactions.php'); ?>
+                        </tr>
+                    <?php }
+                } ?>
                 </tbody>
             </table>
+        <?php   }
+            } ?>
         </div>
 
         <div class="col-xs-12 col-sm-12 col-md-3 col-lg-3 configurationColumnRight">
--- a/admin/plugin_manager.php
+++ b/admin/plugin_manager.php
@@ -58,7 +58,7 @@ $tableDefinition = [
             'derivedItem' => [
                 'type' => 'local',
                 'method' => 'arrayReplace',
-                'params' => ['0' => TEXT_NOT_INSTALLED, '1' => TEXT_INSTALLED_ENABLED, '2' => TEXT_INSTALLED_DISABLED],
+                'params' => ['0' => zen_icon('status-red'), '1' => zen_icon('status-green'), '2' => zen_icon('status-yellow')],
             ],
         ],
     ],
--- a/includes/classes/ViewBuilders/PluginManagerController.php
+++ b/includes/classes/ViewBuilders/PluginManagerController.php
@@ -156,7 +156,7 @@ class PluginManagerController extends BaseController
         zen_redirect(
             zen_href_link(
                 FILENAME_PLUGIN_MANAGER,
-                $this->pageLink() . '&' . $this->colKeylink()
+                $this->pageLink()
             )
         );
     }
@@ -210,7 +210,7 @@ class PluginManagerController extends BaseController
         zen_redirect(
             zen_href_link(
                 FILENAME_PLUGIN_MANAGER,
-                $this->pageLink() . '&' . $this->colKeylink()
+                $this->pageLink()
             )
         );
     }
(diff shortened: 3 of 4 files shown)


piloujp:
I added a commit to display installed plugins first and not installed ones last.
Capture d'écran 2025-09-21 221746

includes/classes/PluginManager.php
```diff
@@ -28,7 +28,7 @@ public function inspectAndUpdate()
 
     public function getInstalledPlugins()
     {
-        $results = $this->pluginControl->where(['status' => 1])->orderBy('name')->orderBy('unique_key')->get();
+        $results = $this->pluginControl->where(['status' =>0])->orderBy('name')->orderBy('unique_key')->get();
```
lat9: Space needed between `=>` and `0`.

piloujp:
There is one point I am not completely satisfied with in this PR. When installing/uninstalling or enabling/disabling a plugin, normally this plugin line stays the active one. Problem is, when doing one of these operations, the plugin is displayed in another table which could be on another page too. In this case, the plugin details box on the right is messed up, and an error log is created. To avoid that, I had to remove that functionality and the first line of the page becomes the active one.
Perhaps someone has an idea how to maintain this line activation without causing an error.
The code modified is at the end of the functions `processActionDo_SomeAction_` in file `../includes/classes/ViewBuilders/PluginManagerController.php` where I removed ` . '&' . $this->colKeylink()` from the link.

piloujp:
> What was the reasoning for changing the meaning of a plugin's `status` values?

To change the order of display and get the table with enabled plugins first instead of the one with plugins not installed. Sorting by status is always in this order: 0->1->2
It is possible to do a custom sorting with MySQL like 1->2->0, but unfortunately Laravel is taking care of the query...

piloujp:
New commit to remove pagination and solving line activation problem.

admin/includes/templates/table_view.php
```diff
@@ -32,36 +32,64 @@
 
     <div class="row">
         <div class="col-xs-12 col-sm-12 col-md-9 col-lg-9 configurationColumnLeft">
+        <?php
+            for($i = 0; $i < 3; $i++) {
```
lat9: Needs space between `for` and `(`.

admin/includes/templates/table_view.php
```diff
@@ -32,36 +32,64 @@
...
+                        <?php foreach ($tableData as $column) { ?>
                         <td class="<?php echo $column['class']; ?>">
                             <?php echo $column['value']; ?>
                         </td>
-                    <?php } ?>
-                    <?php require(DIR_WS_TEMPLATES . 'partials/tableview_rowactions.php'); ?>
-                    </tr>
-                <?php } ?>
+                        <?php } ?>
+                        <?php require(DIR_WS_TEMPLATES . 'partials/tableview_rowactions.php'); ?>
```
lat9: I know that this was just reformatting, but change to
```
                        <?php require DIR_WS_TEMPLATES . 'partials/tableview_rowactions.php'; ?>
```

zc_install/sql/updates/mysql_upgrade_zencart_220.sql
```diff
@@ -118,6 +118,12 @@ UPDATE configuration SET sort_order = 25 WHERE configuration_key = 'IMAGES_AUTO_
 UPDATE configuration SET sort_order = 27 WHERE configuration_key = 'ADDITIONAL_IMAGES_MODE' LIMIT 1;
 UPDATE configuration SET configuration_title = 'Additional Images filename matching pattern', configuration_description = 'In Filename-Matching mode, you can use an &quot;_&quot; suffix in two formats:<br>&quot;strict&quot; = always use &quot;_&quot; suffix<br>&quot;legacy&quot; = only use &quot;_&quot; suffix in subdirectories<br>(Before v210 legacy was the default)<br>Default = strict' WHERE configuration_key = 'ADDITIONAL_IMAGES_MODE' LIMIT 1;
 
+#PROGRESS_FEEDBACK:!TEXT=Updating status value in table plugin_control...
+UPDATE plugin_control SET status = 9 WHERE status = 0;
+UPDATE plugin_control SET status = 0 WHERE status = 1;
+UPDATE plugin_control SET status = 1 WHERE status = 2;
+UPDATE plugin_control SET status = 2 WHERE status = 9;
+
```
lat9: Need also to change the schema of the `plugin_control` table, setting the default as 2 instead of 0.

lat9:
> I added a commit to display installed plugins first and not installed ones last. Capture d&apos;écran 2025-09-21 221746

These columns need to line up.

drbyte:
To get the columns to line up when I did similarly in the Modules page, I applied these styles to the various TH column header cells:
```css
      <style>
          .w-40 {width: 40%}
          .w-20 {width: 20%}
          .w-10 {width: 10%}
      </style>
```

includes/classes/PluginManager.php
piloujp: Done

admin/includes/templates/table_view.php
piloujp: Done

admin/includes/templates/table_view.php
piloujp: Done

zc_install/sql/updates/mysql_upgrade_zencart_220.sql
piloujp: It seems I forgot this one!
Done.

piloujp:
OK, I moved template code in `plugin_manager.php` file and put back `table_view.php` to its original state. Everything is working fine.
It is probably better like this, the template needs rework to be used in different pages, as the `$formater`. Like this, it can be done without interfering with the plugin manager page.

piloujp:
I found a missing `</div>` and realized it was actually missing in `table_view.php` too. Another PR to do.

piloujp:
What about the code for the filter that is no implemented. It is still there, now from line 123 to 136 in `plugin_manager.php`?
As plugins are displayed by status in different tables, this filter, if activated, will only allow to display one table/category at a time.
Perhaps it could be removed from here? Anyway, it is still in `table_view.php`.

drbyte:
Oy. I meant `/admin/includes/templates/plugin_manager.php`

drbyte:
And, to clarify, was the renumbering of `status` because you're relying on sql queries to order things?

Could the status values not be left as they were, and then display each of the tables according to an array of statuses?

Maybe:
```diff
-            for ($i = 0; $i < 3; $i++) {
+            foreach ([2, 0, 1] as $i) {
```

drbyte:
Given that the filter feature is not implemented in this plugin_manager page, I have no objection to removing that from /admin/includes/templates/plugin_manager.php template. Could remove the paginator from that file too, since not used there.
And then table_view.php can still keep those stubs for future use.

piloujp:
> Oy. I meant `/admin/includes/templates/plugin_manager.php` as the page-specific variant of table_view.php

Oups, misunderstanding...

piloujp:
After applying the sorting with the `foreach` loop, I reverted all status numbers changes.
Well, things got a little bit messy, but at the end the solution is simpler.
There is only one file modified file, `admin/plugin_manager.php` and a new file `admin/includes/templates/plugin_manager.php`, which is a custom version of `table_view.php`.
Although I reverted the testFramework change, it seems now it is failing somewhere else, probably because of the new file.

piloujp:
Ahrg, now the default selected line when opening the page becomes the first plugin in alphabetic order of its plugin key, which not necessary in the installed plugins table...
It is defined by query's order by value in `includes/classes/ViewBuilders/PluginManagerDataSource.php`:
`return (new PluginControl())->query()->orderBy('status', 'asc')->orderBy('name')->orderBy('unique_key');`
If I put back status in here, the default active line will be the first one in `Not Installed` plugins table... Which means to fix that, the status order must be `0->Installed(enabled)`, `1->Installed(disabled)`, `3->Not Installed` ...

drbyte:
I think the default-selected line should be determined separately from the query that retrieves all the records, not by an orderBy clause.

piloujp:
> I think the default-selected line should be determined separately from the query that retrieves all the records, not by an orderBy clause.

After a few hours working on this, I think it is impossible without rewriting the `SimpleDataFormatter` class. This there that the first row of the dataset is used by default in method `isRowSelected()` which calls `currentRowFromRequest()` where, on line 77, the code that make this choice is written.
These methods are called three times in the template that built the page. I was able to write some code that do the same functionality directly in the template for two of them but not for the third one which happens when the plugin details box on the right is built. This last code is so much nested in the `box` and `boxTableBlock` classes that again some class rewrite would be necessary.
In actual ZC, just modifying `SimpleDataFormatter.php` by replacing line 77
`$result = $this->resultSet->getCollection()->first();`
with this:
`$result = $this->resultSet->getCollection()->where('status', 1)->first();`
will do the trick, but this hard codded change will certainly be a problem if this method is to be used on other pages construct in the futur.

drbyte:
Thanks for exploring that.

Today I had some time to rollback this branch to commit https://github.com/zencart/zencart/pull/7303/commits/ad838454548e047c69e9b0dd51f358f890eeedc7 ("New plugin _manager.php template file and simplified sorting. Reverted back all status number changes").
I made no database changes.

And I found that the UI worked as-expected (proper line-selection, proper sidebox operation) for everything.

What were you seeing as the problem at that point? (other than I saw that the test suite had an issue, which I'm investigating)

piloujp:
> the default selected line when opening the page becomes the first plugin in alphabetic order of its plugin key, which is not necessary in the installed plugins table...

If you have the first plugin in alphabetic order installed (like AbuseIPDB), it looks OK. But if this plugin is disabled or not installed, it will still be the active one by default, although it will be displayed in the second or third table. In my test with AbuseIPDB not installed, but more than 16 other plugins installed, it was not even visible without scrolling, but still the default active line.

drbyte:
Try this:
```diff
class PluginManagerDataSource extends DataTableDataSource
{
    protected function buildInitialQuery(): Builder
    {
-        return (new PluginControl())->query()->orderBy('name')->orderBy('unique_key');
+        $statusSort = [
+            1, // enabled
+            2, // disabled
+            0, // not installed
+        ];
+        return PluginControl::query()
+            ->orderByRaw(
+                "FIELD(status, " . implode(',', $statusSort) . ")"
+            )
+            ->orderBy('name')
+            ->orderBy('unique_key');
    }
}
```

piloujp:
Excellent! How did you found that? One of the first options I explored to solve this sorting problem led me there, in class `PluginControl` that actually uses Laravel where I gave up. If I had known of this `orderByRaw` method...
Now it looks good, all performing as expected with a minimum code modification.

PR #7327, opened 2025-09-30
System Inspection: Update outdated shipping modules

drbyte:
Does this have the right name in it?


zc_plugins/SystemInspection/1.6.0/admin/includes/extra_configures/system_inspection_base.php
```diff
@@ -319,6 +319,7 @@
 
 define('OBSOLETE_SHIPPING', [
     'ups.php',
+    'usps.php',
```
lat9, line 324: Shouldn't 'fedex.php' be in that list, too?

drbyte: Agreed. Added

PR #7333, opened 2025-10-03
Send CORS headers, restricted to same origin

drbyte:


lat9:
Under what condition(s) will this conditional match?
```
if ($_SERVER['REQUEST_METHOD'] === 'OPTIONS') {
```

PR #7334, opened 2025-10-03
Use Select2 ajax search on Specials-Add-Product page

drbyte:


admin/specials.php
```diff
@@ -256,6 +256,8 @@
 <html <?php echo HTML_PARAMS; ?>>
   <head>
     <?php require DIR_WS_INCLUDES . 'admin_html_head.php'; ?>
+    <link href="https://cdn.jsdelivr.net/npm/select2@4.1.0-rc.0/dist/css/select2.min.css" rel="stylesheet" />
```
lat9: No trailing `/` wanted/needed for HTML5.

includes/classes/ajax/zcAjaxSelect2Lookups.php
```diff
@@ -0,0 +1,98 @@
...
+     * @since ZC v2.2.0
+     */
+    protected function lookup_products_select2(string $lookup = '', array $exclusion_formulas = [], array $excluded_products = [], bool $show_id = true, bool $show_model = true): array
+    {
+        global $currencies, $db;
+
+        if (!is_object($currencies) && !class_exists('currencies')) {
+            require DIR_WS_CLASSES . 'currencies.php';
+            $currencies = new currencies();
+        }
```
lat9: The `currencies` class is psr4Autoloaded; no need to check for existence.  How about a simple
```php
$currencies ??= new currencies();
```
instead?

lat9 (dismissed):

includes/classes/ajax/zcAjaxSelect2Lookups.php
drbyte: Actually, on the Admin side it's not psr4 autoloaded.

lat9: Sigh, you're (as always) correct.

includes/classes/ajax/zcAjaxSelect2Lookups.php
```diff
@@ -0,0 +1,98 @@
...
+
+// @TODO:optionally offer exclusion of products that cannot be added to cart:
+//        LEFT JOIN " . TABLE_PRODUCT_TYPES . " pt ON p.products_type = pt.type_id
+//        WHERE pt.allow_add_to_cart = 'N'
+
+        $search_query = '';
+        if ($lookup !== '') {
+            $q = $db->prepare_input($lookup);
+            $search_query = "  AND (pd.products_name LIKE '%$q%' OR p.products_model LIKE '%$q%' OR p.products_id LIKE '%$q%' OR p.products_price LIKE '%$q%' ) ";
```
proseLA: i see no reason why we are doing this search this way.

why would not be consistent and do something like:

```php
$keyword_search_fields = [
                    'pd.products_name',
                    'p.products_model',
                    'p.products_mpn',
                    'pd.products_description',
                    'p.products_id',
                    'p.products_price',
                ];
                $search_query = zen_build_keyword_where_clause($keyword_search_fields, $q);
```
should we not be keeping search consistent?

drbyte: Yes!
Thanks.

PR #7345, opened 2025-10-04
Attrib controller: Remove redundant line, reformat and refactor

### LINKED ISSUE #7225: Attributes Controller: duplicate assignment of $products_options_sort_order
https://github.com/zencart/zencart/blob/1b385f1af9beaf3f0afa62b9b7c7c24b592fb89b/admin/attributes_controller.php#L230-L241

lat9:
Fixes #7225

### DIFF
--- a/admin/attributes_controller.php
+++ b/admin/attributes_controller.php
@@ -1,97 +1,118 @@
 <?php
-
 /**
  * @copyright Copyright 2003-2024 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: proseLA 2024 Oct 02 Modified in v2.1.0 $
  */
-require('includes/application_top.php');
+require 'includes/application_top.php';
 
 // troubleshooting/debug of option name/value IDs:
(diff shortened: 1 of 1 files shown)


admin/attributes_controller.php
```diff
@@ -1,600 +1,667 @@
...
+
+            // reset products_price_sorter for searches etc.
+            zen_update_products_price_sorter($_GET['products_filter']);
+            zen_redirect(zen_href_link(FILENAME_ATTRIBUTES_CONTROLLER, zen_get_all_get_params(['action'])));
+            break;
+
+        case 'set_flag_attributes_required':
+            $db->Execute(
+                "UPDATE " . TABLE_PRODUCTS_ATTRIBUTES . "
+                    SET attributes_required = " . ($_GET['flag'] === '0' ? '1' : '0') . "
```
drbyte: With applying this `===` strictness, should $_GET['flag'] be cast to string somewhere prior to this, since there's no check anywhere that it's even set before being referenced?

lat9: Good point; the script's been like that (no pre-check) since dinosaurs roamed.  Update forthcoming.

lat9: Correction pushed.

PR #7355, opened 2025-10-07
template plugin_manager: minor html, formatting, <?=

torvista:


admin/includes/templates/plugin_manager.php
```diff
@@ -25,69 +27,73 @@
...
         </div>
 
         <div class="col-xs-12 col-sm-12 col-md-3 col-lg-3 configurationColumnRight">
             <?php
             if (!empty($tableController->getBoxHeader()) && !empty($tableController->getBoxContent())) {
-                $box = new box;
+                $box = new box();
                 echo $box->infoBox($tableController->getBoxHeader(), $tableController->getBoxContent());
             }
             ?>
```
lat9, line 99: Might as well remove that tailing `?>` as well.

torvista: I don't follow. On 99? There's html after.

lat9: My bad; never mind.

PR #7357, opened 2025-10-07
Delete product button, show only when category selected

### LINKED ISSUE #6885: Suggestion - Product Deletion: disable Delete button if no categories selected in infobox
as per thread:
https://www.zen-cart.com/showthread.php?230488-product-deletions&p=1404613#post1404613

For those with few or no linked categories, it seems pretty obvious that product delete means delete.
In my case/for those with many linked categories, this function may be used mainly for delinking and rarely product deletion.
Hence if done, a popup warning of product deletion when the master category has been selected would also be on my list.

lat9:
Fixes #6885

### DIFF
--- a/admin/includes/javascript/category_product_listing.php
+++ b/admin/includes/javascript/category_product_listing.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * @copyright Copyright 2003-2024 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: DrByte 2024 Apr 10 Modified in v2.0.1 $
+ */
+if (($_GET['action'] ?? '') !== 'delete_product') {
+    return;
+}
+?>
+<script id="cpl-php">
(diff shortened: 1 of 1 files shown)


admin/includes/javascript/category_product_listing.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * @copyright Copyright 2003-2024 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ * @version $Id: DrByte 2024 Apr 10 Modified in v2.0.1 $
+ */
+if (($_GET['action'] ?? '') !== 'delete_product') {
+    return;
+}
+?>
+<script id="cpl-php">
```
drbyte, line 11: `id="cpl-php"` ?

lat9: Not wanting the 'id=' to be a novel, it's short for `c` ategory `p` roduct `l` isting

PR #7359, opened 2025-10-08
Sales-report-graph class, update comments, minor reformats

### LINKED ISSUE #7282: sales_report_graph startDate, endDate: mixed types
Used here, defined as strings...but  subsequently require casting to int in multiple places so would be better as int to begin with.

https://github.com/zencart/zencart/blob/5aae67111182615109af8ea073e23b05da67e33b/admin/includes/classes/stats_sales_report_graph.php#L46-L51

Only two calls.

Here passed as strings
https://github.com/zencart/zencart/blob/5aae67111182615109af8ea073e23b05da67e33b/admin/stats_sales_report_graphs.php#L73

Here passed as integers:

https://github.com/zencart/zencart/blob/5aae67111182615109af8ea073e23b05da67e33b/admin/includes/modules/dashboard_widgets/SalesReportDashboardWidget.php#L21

lat9:
Fixes #7282

### DIFF
--- a/admin/includes/classes/stats_sales_report_graph.php
+++ b/admin/includes/classes/stats_sales_report_graph.php
@@ -20,23 +20,22 @@ class statsSalesReportGraph
     /** @var int Number of years to look backward in yearly mode */
     const LOOKBACK_YEARS = 10;
 
-    protected
-        $mode = self::MONTHLY_VIEW,
-        $globalStartDate,
-        $startDate,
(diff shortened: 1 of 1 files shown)


torvista:
all the uses of date need the second parameter cast as int.
With loads of those and the already-cast mktime parameters as int, the code looks like an explosion in a int factory...

PR #7362, opened 2025-10-09
Addl-images - avoid strict error with substr

### LINKED ISSUE #7316: additional_images: not handling an image name without a slash (strict mode)
When an image is no_image.gif, there is no slash so strrpos returns false, which substr chokes on (requires an int)

https://github.com/zencart/zencart/blob/42872b4e6b972976e3fbcf54eb5a1d076195af91/includes/modules/additional_images.php#L36

Possibly whether a slash is forward or back may also be an issue.

Perhaps there is more elegant/stable way to replace this line's function.

drbyte:
@torvista I think this fixes #7316


includes/modules/additional_images.php
```diff
@@ -19,6 +19,101 @@
...
+                        'file' => $file,
+                        'file_extension' => $file_extension,
+                        'products_image' => $products_image,
+                        'products_image_base' => $products_image_base,
+                    ],
+                    $current_image_match
+                );
+
+                // extension match check
+                if ($current_image_match || substr($file, strrpos($file, '.')) === $file_extension) {
```
lat9: Isn't all that substr/strrpos simply reduced to `str_ends_with`?

includes/modules/additional_images.php
```diff
@@ -19,6 +19,101 @@
...
+    $products_image_base = $image_info['filename'];
+    $name_including_ext = $image_info['basename'];
+    $image_directory = $image_info['dirname'];
+
+    $products_image_directory = DIR_WS_IMAGES;
+    // Append "/" to directory if in a subdirectory, and append "_" suffix to base name
+    if (!in_array($image_directory, ['.', '/', '..', ''], true)) {
+        $products_image_directory = DIR_WS_IMAGES . $image_directory . '/';
+        $products_image_base .= '_';
+    } else if ($append_legacy_underscore_to_nonsubdirs) {
```
lat9: `elseif`, not `else if`

includes/modules/additional_images.php
```diff
@@ -19,6 +19,101 @@
...
 }
 
+/**
+ * Parse the product image (from product record) into its components
+ * for use in additional image filesystem lookup
+ *
+ * @return array{0:string, 1:string, 2:string}
+ * @since ZC v2.2.0
+ */
+function zen_get_image_lookup_filename_components(string $products_image, $for_glob = false, $append_legacy_underscore_to_nonsubdirs = false): array
```
lat9: Why no input type hinting for the 2nd and 3rd parameters.

includes/modules/additional_images.php
drbyte: Ya, probably safe to do that in this context.

PR #7368, opened 2025-10-11
zen_trunc_string: length as integer

torvista:
Why?: 
I've got everything else in functions_strings typed/hinted and this is the only function that is called with mixed parameter types (for length).

### DIFF
--- a/includes/functions/functions_strings.php
+++ b/includes/functions/functions_strings.php
@@ -1,4 +1,6 @@
 <?php
+
+declare(strict_types=1);
 /**
  * @copyright Copyright 2003-2024 Zen Cart Development Team
  * @license https://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
@@ -131,42 +133,53 @@ function zen_break_string($string, $len, $break_char = '-')
 }
 
 /**
- * Truncate a string at length specified, optionally adding a "more" suffix
- * @param  string  $str
+ * Truncate a string to the specified length, optionally using a custom "more" suffix
+ * @param  ?string  $str
  * @param  int|string  $len
  * @param  string  $more
- * @return array|string
+ * @return string
  * @since ZC v1.0.3
  */
-function zen_trunc_string(string $str = '', int|string $len = 150, string $more = 'true'): array|string
+function zen_trunc_string(?string $str = '', int|string $len = 150, string $more = '...'): string
 {
-    if (is_null($str) === true) {
+    if (empty($str)) {
         return '';
     }
+
     $len = (int)$len;
-    if ($str === '') return $str;
+    if ($len == 0) {
+        return '';
+    }
+
     $str = trim($str);
-    if ($len == 0) return '';
-    // if it's less than the size given, then return it
-    if (mb_strlen($str) <= $len) return $str;
-    // else get that size of text
+    // if text is less than the limit
+    if (mb_strlen($str) <= $len) {
+        return $str;
+    }
+
+    // get limited text block
     $str = mb_substr($str, 0, $len);
 
     if ($str !== '') {
         // check for no spaces at all
         if (!substr_count($str, ' ')) {
-            if ($more === 'true') $str .= '...';
+            if ($more === 'true') {
+                $str .= '...';
+            }
             return $str;
         }
-        // remove final chars (of a partial word) back to the preceding space
-        while (mb_strlen($str) && ($str[mb_strlen($str) - 1] != ' ')) {
-            $str = mb_substr($str, 0, -1);
-        }
-        //remove the final space
-        $str = mb_substr($str, 0, -1);
 
-        if ($more === 'true') $str .= '...';
-        if ($more !== 'true' && $more !== 'false') $str .= $more;
+        // remove final chars (of a partial word) and the preceding space
+        $str = preg_replace('/(\s\w+$)|(\s+$)/u', '', $str);
+
+        // backwards compatibility
+        // loose comparisons
+        if ($more == 'true') {
+            $str .= '...';
+        }
+        if ($more !== 'true' && $more !== 'false') {
+            $str .= $more;
+        }
     }
     return $str;
 }
--- a/includes/classes/shopping_cart.php
+++ b/includes/classes/shopping_cart.php
@@ -328,7 +328,7 @@ class shoppingCart extends base
                             // Validate max-length of TEXT attribute
                             $check = $db->Execute("SELECT products_options_length FROM " . TABLE_PRODUCTS_OPTIONS . " WHERE products_options_id = " . (int)$option . " LIMIT 1");
                             if (!$check->EOF) {
-                                if (strlen($attr_value) > $check->fields['products_options_length']) {
+                                if (strlen($attr_value) > (int)$check->fields['products_options_length']) {
                                     $attr_value = zen_trunc_string($attr_value, (int)$check->fields['products_options_length'], '');
                                 }
                                 $this->contents[$uprid]['attributes_values'][$option] = $attr_value;


includes/functions/functions_strings.php
```diff
@@ -130,43 +130,42 @@ function zen_break_string($string, $len, $break_char = '-')
...
  * @since ZC v1.0.3
  */
-function zen_trunc_string($str = "", $len = 150, $more = 'true')
+function zen_trunc_string(string $str = '', int|string $len = 150, string $more = 'true'): array|string
 {
     if (is_null($str) === true) {
         return '';
     }
-    if ($str == "") return $str;
+    $len =(int)$len;
```
lat9: Add space between `=` and `(int)`

includes/functions/functions_strings.php
```diff
@@ -130,43 +130,42 @@ function zen_break_string($string, $len, $break_char = '-')
...
         // check to see if there are any spaces left
-        if (!substr_count($str, " ")) {
-            if ($more == 'true') $str .= "...";
+        if (!substr_count($str, ' ')) {
+            if ($more === 'true') $str .= '...';
             return $str;
         }
         // backtrack
-        while (mb_strlen($str) && ($str[mb_strlen($str) - 1] != " ")) {
+        while (mb_strlen($str) && ($str[mb_strlen($str) - 1] !== ' ')) {
```
lat9: This while-loop is just stripping trailing spaces, right?  How about a more simple
```php
$str = rtrim($str, ' ');
```

torvista: I don't think so, it's iterating back to the last space.

includes/functions/functions_strings.php
torvista: ok

includes/functions/functions_strings.php
```diff
@@ -130,43 +130,43 @@ function zen_break_string($string, $len, $break_char = '-')
...
- * @param string $more
- * @return array|false|mixed|string
+ * @param  string  $str
+ * @param  int|string  $len
+ * @param  string  $more
+ * @return array|string
  * @since ZC v1.0.3
  */
-function zen_trunc_string($str = "", $len = 150, $more = 'true')
+function zen_trunc_string(string $str = '', int|string $len = 150, string $more = 'true'): array|string
```
drbyte: > `string $more = 'true'`

Passing a boolean flag as a string has always bothered me. Here you're reinforcing that pattern by requiring it to be string.

I'd much rather it be `bool` but that breaks all backward/cross- compatibility.
~So, next-best pref would be `string|bool` or just `mixed` or just unspecified, and then in the function check for acceptable truthiness where needed. ie: (bool true or string 'true', anything else is ignored/false)~ 

EDIT: See next comment below.

includes/functions/functions_strings.php
```diff
@@ -130,43 +130,43 @@ function zen_break_string($string, $len, $break_char = '-')
...
+    if ($str !== '') {
+        // check for no spaces at all
+        if (!substr_count($str, ' ')) {
+            if ($more === 'true') $str .= '...';
             return $str;
         }
-        // backtrack
-        while (mb_strlen($str) && ($str[mb_strlen($str) - 1] != " ")) {
+        // remove final chars (of a partial word) back to the preceding space
+        while (mb_strlen($str) && ($str[mb_strlen($str) - 1] != ' ')) {
```
drbyte: Could this `while` loop be replaced with a one-liner, like this?
`$str = preg_replace('/\S+\s*$/u', '', $str);`
(\S+ means the (second-) last set of non-whitespace-chars (ie: the last partial word))
(\s* means the whitespace after that word (which is why \S matches 'second last'))
(and `$` matches end of the string, while `u` means multibyte-safe (assuming utf8))

It also removes the final space.  Could use `'/\S+(?=\s*$)/u'` to keep the final white-space. But we're trimming it off later anyway, so could just chop it here at the same time.

includes/functions/functions_strings.php
drbyte: Oh. I guess we've historically also allowed `$more` to be a string substitute to use instead of `...`.
Argh.

I guess that then means we should probably default it to `string $more = '...'`, and then for backward compatibility check if $more === 'true' and replace 'true' with '...', and replace 'false' with ''.

Apart from shopping_cart.php where we use `''`, I don't know of any place where we've actually changed the $more value from default, as far as core code goes.

includes/functions/functions_strings.php
piloujp: There is a case that was not taken on account with the while loop, it is when the string has actually been strip down just to the end of a word. It then ends with a letter, in which case it is very difficult to say if the word is cut or not, or it ends with a punctuation sign, in which case no more trimming is necessary, or it ends with a space which just has to be removed.
Here is a pattern that should deal with punctuation and ending space without removing full words:
`'/(\s\S+[^\s\.!\?,:;]$)|(\s+$)/u'`
It uses the regex OR `(...)|(...)` to deal with a white space at the end case (`\s+$`), but this could be done by a trim function before the regex.
This part `[^\s\.!\?,:;]` ensure that we don't remove full words. We know there are not cut thanks to these punctuation signs.

piloujp: As we are using the u modifier, this regular expression will do the same but is really simpler:
`'/(\s\w+$)|(\s+$)/u'`

includes/functions/functions_strings.php
```diff
@@ -130,43 +132,54 @@ function zen_break_string($string, $len, $break_char = '-')
...
+ * @param  int|string  $len
+ * @param  string  $more
+ * @return array|string
  * @since ZC v1.0.3
  */
-function zen_trunc_string($str = "", $len = 150, $more = 'true')
+function zen_trunc_string(string $str = '', int|string $len = 150, string $more = '...'): array|string
 {
-    if (is_null($str) === true) {
+    if (is_null($str) === true || $str === '') {
```
lat9: If `$str` is a string (per the input type-casting), then it can't be `null`.

torvista: true

includes/functions/functions_strings.php
```diff
@@ -130,43 +132,54 @@ function zen_break_string($string, $len, $break_char = '-')
...
-    if ($str != "") {
-        // check to see if there are any spaces left
-        if (!substr_count($str, " ")) {
-            if ($more == 'true') $str .= "...";
+
+    if ($str !== '') {
+        // check for no spaces at all
+        if (!substr_count($str, ' ')) {
+            if ($more === 'true') {
+                $str .= '...';
```
drbyte, line 167:
```suggestion
                $str .= '...';
            } elseif ($more !== 'false') {
                $str .= $more;
```

drbyte: Oh. Skip that. I see you deferred this to the very end.

includes/functions/functions_strings.php
```diff
@@ -130,43 +132,54 @@ function zen_break_string($string, $len, $break_char = '-')
...
+
+        // remove final chars (of a partial word) and the preceding space
+        $str = preg_replace('/(\s\w+$)|(\s+$)/u', '', $str);
+
+        // backwards compatibility
+        // loose comparisons
+        if ($more == 'true') {
+            $str .= '...';
+        }
+        if ($more != 'true' && $more != 'false') {
```
drbyte:
```suggestion
        if ($more !== 'true' && $more !== 'false') {
```

includes/functions/functions_strings.php
```diff
@@ -130,43 +132,54 @@ function zen_break_string($string, $len, $break_char = '-')
...
 /**
  * Truncate a string at length specified, optionally adding a "more" suffix
- * @param string $str
- * @param int $len
- * @param string $more
- * @return array|false|mixed|string
+ * @param  string  $str
+ * @param  int|string  $len
+ * @param  string  $more
+ * @return array|string
```
drbyte: Was leaving array intentional here? 
And below in the function signature?

includes/functions/functions_strings.php
torvista: I was lumping the loose comparision in with backwards compatibility, as I found a boolen true in some mod.

includes/functions/functions_strings.php
torvista: no

includes/functions/functions_strings.php
drbyte: Okay, but string `'false'` is the same as boolean `true` (the string is not 0/null/empty, so evaluates as true).

Better to be explicit, else will have unexpected results.

includes/functions/functions_strings.php
```diff
@@ -130,43 +132,54 @@ function zen_break_string($string, $len, $break_char = '-')
...
 
-
 /**
- * Truncate a string at length specified, optionally adding a "more" suffix
- * @param string $str
- * @param int $len
- * @param string $more
- * @return array|false|mixed|string
+ * Truncate a string to the specified length, optionally using a custom "more" suffix
+ * @param  null|string  $str
```
drbyte: Tip: the more commonly-used shorthand `?string` is the same as `null|string` 
(the ? prefix makes it nullable, added in PHP 7.1)

Can be used in docblocks, function params, function returns, and in class-properties declared at the top of a class.

torvista: ok

includes/functions/functions_strings.php
torvista: ok, so just the last one should be strict then..

includes/functions/functions_strings.php
```diff
@@ -130,43 +132,54 @@ function zen_break_string($string, $len, $break_char = '-')
...
+                $str .= '...';
+            }
             return $str;
         }
-        // backtrack
-        while (mb_strlen($str) && ($str[mb_strlen($str) - 1] != " ")) {
-            $str = mb_substr($str, 0, -1);
+
+        // remove final chars (of a partial word) and the preceding space
+        $str = preg_replace('/(\s\w+$)|(\s+$)/u', '', $str);
```
drbyte, line 173: @piloujp would the addition of `\p{L}` make this safer for asian characters? (changing from merely ASCII to Unicode)

ie: `'/(\s\p{L}+$)|(\s+$)/u'`

drbyte (approved):

includes/functions/functions_strings.php
piloujp: I don't know what is going on with GitHub, but my answer to this disappeared! 
Sorry for multiple edits, I got lost trying to answer quickly...
Asian character are no problem.

piloujp: The problem with `\p{L}` is that it does not include numbers. If there is a number at the end, there is no way to know if it is entire or has been cut, it should be removed. Something like `[\p{L}\p{N}]` would work like `\w`.
In this case, there would be a slight difference only in some (exotic) languages where `\p{L}` would catch some special word modifier characters as part of a word where `\w` could not. Unfortunately, I was not able to find back where I read this.

drbyte: Okay, will leave it as-is. Thanks.

PR #7370, opened 2025-10-11
Rework unescape calls

### LINKED ISSUE #6893: Javascript unescape deprecated
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape

https://github.com/zencart/zencart/blob/8333810225d0e0ec45c6aa51e97cef1bc0302e5b/includes/templates/responsive_classic/common/html_header.php#L97-L99

I note this double check is not in the default template, which while it is not needed there, it IS ostensibly the reference file...but is xhtml so it's an oddity/not really the reference in this instance.

drbyte:
Fixes #6893

### DIFF
--- a/includes/templates/responsive_classic/common/html_header.php
+++ b/includes/templates/responsive_classic/common/html_header.php
@@ -94,9 +94,9 @@ require $template->get_template_dir('html_header_css_loader.php', DIR_WS_TEMPLAT
 ?>
 <script src="https://code.jquery.com/jquery-3.7.1.min.js" integrity="sha256-/JqT3SQfawRcv/BIHPThkBvs0OEvtFFmqPF/lYI/Cxo=" crossorigin="anonymous"></script>
 <?php if (file_exists(DIR_WS_TEMPLATE . "jscript/jquery.min.js")) { ?>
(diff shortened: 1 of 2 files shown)


includes/templates/responsive_classic/common/html_header.php
```diff
@@ -94,9 +94,9 @@
 ?>
 <script src="https://code.jquery.com/jquery-3.7.1.min.js" integrity="sha256-/JqT3SQfawRcv/BIHPThkBvs0OEvtFFmqPF/lYI/Cxo=" crossorigin="anonymous"></script>
 <?php if (file_exists(DIR_WS_TEMPLATE . "jscript/jquery.min.js")) { ?>
-<script title="jQuery check - template">window.jQuery || document.write(unescape('%3Cscript src="<?php echo $template->get_template_dir('.js',DIR_WS_TEMPLATE, $current_page_base,'jscript'); ?>/jquery.min.js"%3E%3C/script%3E'));</script>
+<script title="jQuery check - template">window.jQuery || document.write('<script src="<?php echo $template->get_template_dir('.js',DIR_WS_TEMPLATE, $current_page_base,'jscript'); ?>/jquery.min.js"><\/script>');</script>
```
torvista: short tags?

drbyte: 👍

PR #7373, opened 2025-10-13
function zenDoCurlRequest: add parameters to curl url

torvista:
I needed to use parameters for accessing the api of a cloud ERP.
Maybe there is a preferred method to this suggestion.

### DIFF
--- a/includes/functions/functions_communications.php
+++ b/includes/functions/functions_communications.php
@@ -25,7 +25,7 @@ use http\Exception\BadQueryStringException;
  * @return string|array|false False on failure; string for normal response; array if advanced metaData requested
  * @since ZC v2.2.0
  */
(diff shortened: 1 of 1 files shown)


includes/functions/functions_communications.php
```diff
@@ -68,6 +68,16 @@ function zenDoCurlRequest(string $url, string $method = 'GET', string|array|null
       }
 
       curl_setopt_array($ch, array_replace($curlDefaultOptions, $extraCurlOptions ?? []));
+
+      if (!empty($params)) {
+          $parameters = '?';
+          foreach ($params as $key => $value) {
+              $parameters .= $key . '=' . $value . '&';
```
drbyte, line 75: This is the same as what line 51 does:
`$payload = http_build_query($payload);`

Pass your array to $payload, and if $encodeArraysAsJson is false it'll convert it into &-delimited parameters.

torvista: Indeed.

PR #7378, opened 2025-10-17
More on zen_trunc_string

drbyte:
I was looking at @torvista recent update and added some tests, and then discovered that negative lengths weren't handled.
And wanted to skip the large "if" clause in the last half of the function.

- treats negative $len as 0 (it should never happen, but ... )
- avoids a nested `if`
- simplifies $more logic

@piloujp  - I made a try with some Japanese examples. Feedback welcome, as those examples are probably laughable.

### DIFF
--- a/includes/functions/functions_strings.php
+++ b/includes/functions/functions_strings.php
@@ -140,7 +140,7 @@ function zen_trunc_string(?string $str = '', int|string $len = 150, string $more
     }
 
     $len = (int)$len;
-    if ($len == 0) {
+    if ($len <= 0) {
         return '';
     }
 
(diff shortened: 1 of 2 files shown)


piloujp:
Don't worry too much about Japanese, it is a language that do not use spaces and for a program to separate words in a Japanese string you probably will need a complete AI system... All that this function does with Japanese strings is a substring cut if too long.
Your Japanese example is fine.
There are unfortunately some cases where things will go wrong. Although spaces are not used in Japanese, when using foreign words as is in a sentence, for readability, spaces can be added between these words and Japanese characters. In this case, there is more than one word that would be cut by the function...
If I modify your example like this:
`申し訳ありませんが、drbyte 様そのご要望には対応できません。`
The result will be (with mb_substr length parameter > 16 and < 33):
`申し訳ありませんが、 drbyte...`
The regex will cut from the end until there is a space and no punctuation sign in between.
I am afraid there is nothing we can do about it except not using the regex or any word cutting code on Japanese strings.

torvista (commented):
one function down, only 20+ to go! :satisfied:

PR #7382, opened 2025-10-18
functions_strings: parameter hints

torvista:
hopefully just the non-contentious ones.
Where int is specified, core files are using int.

### DIFF
--- a/includes/functions/functions_strings.php
+++ b/includes/functions/functions_strings.php
@@ -34,7 +34,7 @@ function zen_output_string(?string $string, array|bool $translate = false, bool 
 /**
  * Returns a string with quotes converted to html entities
  * so that they can be passed through from page to page
- * without mistakenly being converted to specialchars or go missing
+ * without mistakenly being converted to specialchars or go "missing"
  * @since ZC v2.2.0
  */
 function zen_preserve_search_quotes(?string $search_string): string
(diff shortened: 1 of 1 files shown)


includes/functions/functions_strings.php
```diff
@@ -49,10 +49,10 @@ function zen_preserve_search_quotes(?string $search_string): string
  * with parameters that run htmlspecialchars over the string
  * and converts quotes to html entities
  *
- * @param string The string to be parsed
+ * @param  string  $string  The string to be parsed
```
drbyte, line 52: I'm not objecting to keeping this, but since the comment is redundant now that the type is in the function signature, this `@param` line can just be deleted. Especially when there's only a single parameter, because there's no ambiguity.

In cases where the function sig type is **not** declared in the function, then we **do** want it in the docblock as the cue for IDEs and code-analyzers to aid in dev hints/warnings.

includes/functions/functions_strings.php
```diff
@@ -308,14 +308,13 @@ function zen_html_entity_decode($given_html, $quote_style = ENT_QUOTES)
  *
  * @since ZC v1.1.0
  */
-function zen_decode_specialchars($string)
+function zen_decode_specialchars($string): array|string
```
drbyte: Since str_replace can handle incoming arrays, and return said arrays, we may as well allow it for consistency.
```suggestion
function zen_decode_specialchars(array|string $string): array|string
```

drbyte (approved):

includes/functions/functions_strings.php
torvista: I was thinking more of these lines for parameters being places to add information/comments.

PR #7389, opened 2025-10-19
shopping cart: hide empty spans

torvista:


includes/templates/responsive_classic/templates/tpl_shopping_cart_default.php
```diff
@@ -81,9 +81,9 @@
   } else {
     echo $product['quantityField'];
   }
-?><br>
-        <span class="alert bold"><?php echo $product['flagStockCheck'];?></span><br>
-        <br><?php echo $product['showMinUnits']; ?>
+  echo !empty($product['flagStockCheck']) ? '<span class="alert bold">' . $product['flagStockCheck'] . '</span><br>' : '';
+  echo !empty($product['showMinUnits']) ? $product['showMinUnits'] . '<br>' : '';
+  ?>
```
drbyte, line 86: Why remove the `<br>` tags? These spans aren't block-level, right, so they'll just sit beside each other, instead of giving the visitor visual cues of their difference?

includes/templates/responsive_classic/templates/tpl_shopping_cart_default.php
```diff
@@ -93,7 +93,11 @@
 
        <td class="cartProductDisplay">
 
-<a href="<?php echo $product['linkProductsName']; ?>"><span class="cartImage back"><?php echo $product['productsImage']; ?></span><span class="cartProdTitle"><?php echo $product['productsName'] . '<span class="alert bold">' . $product['flagStockCheck'] . '</span>'; ?></span></a>
+<a href="<?= $product['linkProductsName'] ?>">
+    <span class="cartImage back"><?php echo $product['productsImage']; ?></span>
```
drbyte: echo vs <?=

includes/templates/responsive_classic/templates/tpl_shopping_cart_default.php
torvista: Not sure I understand the last bit of your comment.

I thought the br should only be needed if the item was displayed, otherwise a big gap.

includes/templates/responsive_classic/templates/tpl_shopping_cart_default.php
torvista: distracted

includes/templates/responsive_classic/templates/tpl_shopping_cart_default.php
drbyte: Fair. I was focusing on the first `<br>`, which has been removed, and that the double-`<br>` between lines has been changed to a single.

PR #7390, opened 2025-10-19
Remove extraneous 2nd parameter from `$order->send_order_email`

### LINKED ISSUE #7386: checkout process ->send_order_email surplus parameter
https://github.com/zencart/zencart/blob/9a714144814f3a7a5ad78b0fe62667c18f7b273b/includes/modules/checkout_process.php#L122

not quite sure what is deprecated

https://github.com/zencart/zencart/blob/9a714144814f3a7a5ad78b0fe62667c18f7b273b/includes/classes/order.php#L1333-L1337

lat9:
Correction in first commit; second contains various refinements.

Fixes #7386

### DIFF
--- a/includes/modules/checkout_process.php
+++ b/includes/modules/checkout_process.php
@@ -27,9 +27,7 @@ if (!zen_is_logged_in()) {
 
 // BEGIN CC SLAM PREVENTION
 $slamming_threshold = 3;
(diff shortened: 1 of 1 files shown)


scottcwilson:
Good catch, that parameter has been gone for a while.

PR #7403, opened 2025-10-22
 SetLocale() no longer accepts int 0 since PHP 8.5. Use string '0' instead.

### LINKED ISSUE #7399: php 8.5rc3, strict_mode: setlocale fatal error when passed integer
php 8.4.13: no problem
php 8.5rc3: fatal error

https://github.com/zencart/zencart/blob/c675f120710a8ce4b8dc6807ce5b9df41ccae970/admin/includes/application_bootstrap.php#L46

https://github.com/zencart/zencart/blob/c675f120710a8ce4b8dc6807ce5b9df41ccae970/admin/includes/header.php#L222

nothing in php changelog to indicate the tightening-up.

Occurs in a couple of other places in core code.

drbyte:
Param 2 cannot be int anymore, must be string.
Formerly we passed `0` to "retrieve current locale". To do the same, use string `'0'` instead.

Fixes #7399

### DIFF
--- a/admin/includes/application_bootstrap.php
+++ b/admin/includes/application_bootstrap.php
@@ -43,7 +43,7 @@ date_default_timezone_set(date_default_timezone_get());
  * Check for a valid system locale, and override if invalid or set to 'C' which means 'unconfigured'
  * It will be overridden later via language-selection operations anyway, but a valid default must be set for zcDate class methods to work
  */
-$detected_locale = setlocale(LC_TIME, 0);
+$detected_locale = setlocale(LC_TIME, '0');
 if ($detected_locale === false || $detected_locale === 'C') {
     setlocale(LC_TIME, ['en_US', 'en_US.UTF-8', 'en-US', 'en']);
 }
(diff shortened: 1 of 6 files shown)


torvista:
Windows 11
when null is used in zcDate:

> [23-Oct-2025 09:01:55 Europe/Madrid] PHP Fatal error:  Uncaught ValueError: IntlDateFormatter::__construct(): Argument #1 ($locale) "English_United Kingdom.1252" is invalid in ...\includes\classes\zcDate.php:101
> Stack trace:
> #0 ...\includes\classes\zcDate.php(101): IntlDateFormatter->__construct('English_United ...', 3, -1)
> #1 ...\includes\classes\zcDate.php(196): zcDate->initializeConversionArrays()
> #2 ...\ADMIN\includes\header.php(302): zcDate->output('%A %d %b %Y %H:...', 1761202915)
> #3 ...\ADMIN\configuration.php(90): require('D:\\OneDrive\\Bus...')
> #4 ...\ADMIN\index.php(16): require('D:\\OneDrive\\Bus...')
> #5 {main}
>   thrown in ...\includes\classes\zcDate.php on line 101

This locale is from Windows/not in the language file list.

I found a note in my check_server_locales script that IntlDateFormatter:: was choking on this from 8.3.

So does this imply the script is using the platform locale instead of that defined in the language file...?
Is this due to the change in method of loading language files?

If I use '0' instead of null, the locale used is that from the language file.

PR #7407, opened 2025-10-23
Move order-class notification

### LINKED ISSUE #7406: Observer cannot use a private method of the originating class
### Discussed in https://github.com/zencart/zencart/discussions/7405

<sup>Originally posted by **torvista** October 23, 2025</sup>
I want to add custom fields into the order object for customer and address data.

As it stands:
https://github.com/zencart/zencart/blob/d4189ee16da3126f8fa56f1d42792c55efc351b3/includes/classes/order.php#L564
extra fields cannot be **added**: the observer needs to supply a complete array to replace the core version.

So, the obvious solution would be to replicate the core code in the observer, but I find that the observer cannot access the private methods getAddress/getAddressKey in the order class. 
e.g. this works as an example, but only after making order->getAddress/getAddressKey as public.

```
function notify_order_cart_address_overrides(&$class, $eventID, $unused, &$customer_address_override, &$delivery_address_override, &$billing_address_override): void
    {
        global $customer;
        //replicate core
        $customerAddresses = $customer->getData('addresses');
        $customer_address_override = $class->getAddress($customerAddresses, $class->getAddressKey($customerAddresses, $customer->getData('customers_default_address_id')));
        $customer_address_override['telephone'] = $customer->getData('customers_telephone');
        $customer_address_override['email_address'] = $customer->getData('customers_email_address');
        //custom
        $customer_address_override['tax_id'] = $customer->getData('customers_taxid');
    }
```
Apart from this field, I need to add telephones to the billing/delivery addresses.

a) What should I be doing? I had a vague idea that the & in front of the $class would magically allow access.
b) Would it not be better to have the notifier **after** the vanilla arrays are created so only modifications/additions need to be coded/avoid duplication.

</div>

lat9:
Fixes #7406

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -551,28 +551,13 @@ class order extends base
             $billKey = $this->getAddressKey($customerAddresses, $billto);
         }
 
-        // -----
-        // Provide a watching observer (think EO!) a means to override the order's information as
-        // well as the customer/delivery/billing addresses.
-        //
-        // Note: Any address-override array returned must provide ALL address-array keys as used by this class!
-        // For example, the $customer_address_override array must include a 'firstname' element, not 'customers_firstname'.
-        //
-        $customer_address_override = [];
-        $delivery_address_override = [];
-        $billing_address_override = [];
-        $this->notify('NOTIFY_ORDER_CART_ADDRESS_OVERRIDES', [], $customer_address_override, $delivery_address_override, $billing_address_override);
-
-        if (count($customer_address_override) !== 0) {
-            $this->customer = $customer_address_override;
-
-        } elseif (!empty($customer->getData('customers_firstname'))) {
+        if (!empty($customer->getData('customers_firstname'))) {
             $this->customer = $this->getAddress($customerAddresses, $this->getAddressKey($customerAddresses, $customer->getData('customers_default_address_id')));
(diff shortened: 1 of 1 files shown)


torvista:
I don't get your point about the current order...

My big picture is:
this notification allows a field (e.g. a telephone) from the address book to be made available for example displayed with the addresses throughout the checkout pages.

Subsequently, 
another notifier allows those fields to be added to the recorded order details (done),
another notifier allows those fields to be added to the email address blocks (done),
another (new) notifier is needed (a copy of the one in this PR, to go in order->query) will allow those fields to be displayed with the addresses when the order is shown in order history.

All this is particularly necessary for wholesale/trade customers, where the customer is head office and never gets a delivery...but has multiple franchises/branches each with their phone and email contact details....

Some time ago, I did propose adding telephones to the address fields to avoid all this, but that got bogged down in the mud.

lat9:
The base order-class' addresses don't include the associated `address_book_id`; the notification to-be-moved provides a means to update address-related information for the current order.  It's not a means to manage records in the `address_book` table.

torvista:
Yes I get that, I think...this is just retrieval and use of that data.
I've had all this stuff in the order class for years and am now trying to get it to vanilla with observers.

The actual addition/modification of said details is done elsewhere.

PR #7410, opened 2025-10-24
admin/manufacturers: Additional notification

lat9:
Add a notifier at the in-core completion of any insert/save processing to allow a plugin to manage additional manufacturer-related database information.

### DIFF
--- a/admin/manufacturers.php
+++ b/admin/manufacturers.php
@@ -137,6 +137,12 @@ if (!empty($action)) {
                       WHERE manufacturers_id = " . (int)$manufacturers_id);
       }
 
+      // -----
+      // Give a watching observer the opportunity to cleanup additional information for
+      // the removal of the current manufacturer.
+      //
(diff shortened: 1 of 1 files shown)


admin/manufacturers.php
```diff
@@ -131,6 +137,12 @@
                       WHERE manufacturers_id = " . (int)$manufacturers_id);
       }
 
+      // -----
+      // Give a watching observer the opportunity to cleanup additional information for
+      // the removal of the current manufacturer.
+      //
+      $zco_notifier->notify('NOTIFY_ADMIN_MANUFACTURERS_DELETECONFIRM', ['manufacturers_id' => (int)$manufacturers_id]);
```
scottcwilson, line 144: Underscore before CONFIRM to match customers.php?

lat9: I'd left it that way to match the `action` being performed on the manufacturers' page.

scottcwilson: ok

PR #7413, opened 2025-10-25
Match up pages when encountering malformed URI params

### LINKED ISSUE #6997: query parameters case sensitive
ZC 2.1.0 vanilla install (applies to older versions just the same)

All query parameters are case-sensitive and should be forced into lowercase. For example:
index.php?main_page=index&**cPath**=3_10 - will bring up the category page just fine
index.php?main_page=index&**cpath**=3_10 - **home page**

index.php?main_page=product_info&**cPath**=3_10&products_id=12 - normal product page
index.php?main_page=product_info&**cpath**=3_10&products_id=12 - again normal product page, but Google will see it as **duplicate content**

index.php?main_page=product_info&cPath=3_10&**products_id**=12 - normal product page
index.php?main_page=product_info&cPath=3_10&**products_ID**=12 - **404**

**Expected result:**
Check the whole URL, including all query parameters, for any uppercase characters and force-redirect (301) to a completely lowercase URL. Make all $_GET, $_POST and $_REQUEST parameters work with lowercase-only characters.

As a temp solution, I'm handling cPath like this at the top of application_top.php:
```
$_GET = array_change_key_case($_GET, CASE_LOWER);

if (isset($_GET['cpath'])) {
    $_GET['cPath'] = $_GET['cpath'];
}
```
but this would basically require mapping every single parameter that currently uses any uppercase character. Can't really think of any other than cID and pID, but in case some originate from custom coding, might be good to have a universal solution?

drbyte:
May fix #6997

/cc @zenexpert - Thoughts?

### DIFF
--- a/includes/init_includes/init_cart_handler.php
+++ b/includes/init_includes/init_cart_handler.php
@@ -19,7 +19,7 @@ if (isset($_GET['action'])) {
   }
   if (DISPLAY_CART == 'true') {
     $goto =  FILENAME_SHOPPING_CART;
-    $parameters = array('action', 'cPath', 'products_id', 'pid', 'main_page');
+    $parameters = array('action', 'cPath', 'products_id', 'pID', 'pid', 'main_page');
   } else {
     $chk_handler = zen_get_info_page(isset($_GET['products_id']) ? $_GET['products_id'] : 0);
     $goto = $_GET['main_page'];
@@ -31,9 +31,9 @@ if (isset($_GET['action'])) {
         $parameters = array('action', 'products_id');
       }
     } elseif ($_GET['main_page'] == $chk_handler) {
-      $parameters = array('action', 'pid', 'main_page');
+      $parameters = array('action', 'pID', 'pid', 'main_page');
     } else {
-      $parameters = array('action', 'pid', 'main_page', 'products_id');
+      $parameters = array('action', 'pID', 'pid', 'main_page', 'products_id');
     }
   }
   /**
--- a/includes/init_includes/init_sanitize.php
+++ b/includes/init_includes/init_sanitize.php
@@ -80,7 +80,6 @@ $caseSensitiveKeysToMap = [
     'music_genre_id',
     'page',
     'pID',
-    'pid',
     'product_id',
     'products_id',
     'products_image_large_additional',
@@ -118,8 +117,8 @@ unset($mixedKey);
 $saniGroup1 = [
     'products_id',  //- 'Normal', multi-use
     'product_id',   //- shopping_cart, when removing a product from the cart
-    'pid',          //- order_history sidebox and ask_a_question page
-    'pID',          //- main/additional images' pop-ups
+    'pID',          //- main/additional images' pop-up pages, order_history sidebox, and ask_a_question page
+    'pid',          //- prior to v2.2.0: order_history sidebox, ask_a_question page
 ];
 foreach ($saniGroup1 as $key) {
     if (isset($_GET[$key]) && !preg_match('/^\d+(:[0-9a-f]{32})?$/', (string)$_GET[$key])) {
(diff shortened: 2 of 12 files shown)


includes/init_includes/init_sanitize.php
```diff
@@ -62,6 +62,50 @@
...
+    'chapter',
+    'cID',
+    'cPath',
+    'faq_item',
+    'id',
+    'manufacturers_id',
+    'music_genre_id',
+    'page',
+    'pID',
+    'pid',
```
lat9: 'pid' vs 'pID' is going to be problematic, since the associated page dictates which one is acceptable.

drbyte: @lat9 True. 
'pid' is used in 2 places: ask-a-question, and some internals for shopping_cart processing.

I'd suggest that the best approach for this particular case is twofold:
1. skip pid/pID in this array,
2. update ask-a-question to use `pID`, for consistency (so that any misuse isn't misinterpreted) (fallback to pid should be safe IMO, since if we first search for pID and fallback to pid it there shouldn't be unwanted side-effects.)

And on ask-a-question, I'm pretty sure that where we use pID we are using it as a lookup for the same thing we're using pid as the lookup for.

torvista: Found pid in access blocker and CEON URI Mapping, but both are related to Ask a Question/easily fixed.

drbyte: On further thought, I think it's even better to update the other cases which use lowercase 'pid' to mixed-case 'pID' as well, with fallbacks for compatibility.

drbyte: > CEON URI Mapping

Yes, the USU mod is affected as well.

includes/init_includes/init_sanitize.php
```diff
@@ -62,6 +62,49 @@
...
+    'reviews_id',
+    'search_in_description',
+    'typefilter',
+];
+foreach ($caseSensitiveKeysToMap as $key) {
+    if (!isset($_GET[$key])) {
+        foreach ($_GET as $mixedKey => $value) {
+            if (strtolower($mixedKey) === strtolower($key)) {
+                $_GET[$key] = $value;
+                unset($_GET[$mixedKey]);
```
drbyte, line 96: I'd be open to excluding `'pid'` from being `unset` here.

includes/init_includes/init_cart_handler.php
```diff
@@ -19,7 +19,7 @@
   }
   if (DISPLAY_CART == 'true') {
     $goto =  FILENAME_SHOPPING_CART;
-    $parameters = array('action', 'cPath', 'products_id', 'pid', 'main_page');
+    $parameters = array('action', 'cPath', 'products_id', 'pID', 'pid', 'main_page');
```
drbyte, line 22: The retention of the pre-existing old array syntax is noted. 
I'll reformat the file separately, to keep this PR less busy.

drbyte:
@zenexpert thanks for the code suggestions.
Sanitizing here should fix canonical issues (which these params are used for), which should fix duplicate-content issues.

But redirects are a bit challenging. This list of GET params is not exhaustive, so I'm hesitant to issue redirects that might drop other GET params that could legitimately (and harmlessly) be present. 
(That said, redirects are a handy way to clean up bad urls that search engines have grabbed for whatever reason.)
I'll have to do a bit more research to see what risks may exist in that regard.

zenexpert:
I understand, it's just a suggestion that would hopefully avoid having incorrect bookmarks, correct search engines' indexes and might help avoid any confusion during any debugging. 
The proposed code would not drop other GET parameters that are not on the list - it only corrects the ones in the list, and the keys that are not in the list are simply passed on. And, the redirect doesn't trigger if there's a POST because we don't want to lose the POSTed data, but in reality, I don't expect incorrect GET parameters at the time when there's a POST happening so it's basically there just in case.

But again, it's just an idea - the main issue seems to be resolved, this is just cake decoration now.

PR #7414, opened 2025-10-26
Skip lang check if file already loaded

drbyte:
Since require_once will return `true` if already loaded, we may want to avoid trying to parse the boolean as an array.

@piloujp Does this pose any problem in your multilang constants' processing?


piloujp:
Not at all. This should make the code more efficient.
Tested in 3 languages and with tones of plugins installed.

PR #7429, opened 2025-11-13
improvements to configuration pages

barco57:
Adjust column widths
info background color
header background color

related to ZC forum conversation here: https://www.zen-cart.com/showthread.php/230932-Alpha-build-of-2-2-0-now-available

### DIFF
--- a/admin/configuration.php
+++ b/admin/configuration.php
@@ -175,7 +175,7 @@ if ($gID === 7) {
     	.form-horizontal hr {
 			margin: 0;
 			border: 0;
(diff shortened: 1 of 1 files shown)


dbltoe:
My only concern with this change is the color of the HR. #bfbfbf is only a contrast ratio of 1.8:1. That makes it inherently difficult to see for many folks with some forms of color blindness. Going to at least #949494 jumps the ratio to 3:1, which qualifies for large text.

barco57:
Before:
image

After:
image

PR #7434, opened 2025-11-25
products_to_categories: add Button to hide/show Linked Categories

torvista:
Useful for working with the Global Tools and preventing unwanted loading/display of (a massive list of) linked categories.

Hide  

To

show  

Note that for the huge "changed" block as shown in the diff display, the only change is 614-620 for the extra code for the button/form.
All the rest is unmodified/original code indented and a correction of the location of the final closing divs. The page seems to have given up trying to make a match.

### DIFF
--- a/admin/products_to_categories.php
+++ b/admin/products_to_categories.php
@@ -21,6 +21,10 @@ $_GET['current_category_id'] = $current_category_id; // for any redirects
 if (isset($_POST['form'])) {
     switch (true) {
 
+        case($_POST['form'] === 'hide_linked_categories_form'):
+            $_SESSION['hide_linked_categories'] = !empty($_POST['hide_linked_categories']);
+            break;
+
         case($_POST['form'] === 'copy_links_dropdown_form'):
             $_SESSION['copy_links_dropdown'] = isset($_POST['copy_links_dropdown']);
             break;
@@ -418,13 +422,16 @@ if ($products_filter > 0) {
     $product_linked_categories = $db->Execute("SELECT products_id, categories_id FROM " . TABLE_PRODUCTS_TO_CATEGORIES . " WHERE products_id = " . $products_filter);
 }
 
+// hide/show Linked Categories
+if (empty($_SESSION['hide_linked_categories'])) {
 // Build the list of categories within the target category
-$categories_info = [];
-zen_get_categories_info($target_category_id); // loads $categories_info with subcategories of chosen target category
-$target_subcategory_count = count($categories_info);
-$max_input_vars = @ini_get("max_input_vars");
-if ($target_subcategory_count > $max_input_vars) { //warning when in excess of POST limit
-    $messageStack->add(sprintf(WARNING_MAX_INPUT_VARS_LIMIT, $target_subcategory_count, $max_input_vars, 'caution'));
+    $categories_info = [];
+    zen_get_categories_info($target_category_id); // loads $categories_info with subcategories of chosen target category
+    $target_subcategory_count = count($categories_info);
+    $max_input_vars = @ini_get("max_input_vars");
+    if ($target_subcategory_count > $max_input_vars) { //warning when in excess of POST limit
+        $messageStack->add(sprintf(WARNING_MAX_INPUT_VARS_LIMIT, $target_subcategory_count, $max_input_vars, 'caution'));
+    }
 }
 ?>
 <!doctype html>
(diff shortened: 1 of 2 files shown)


drbyte:
> Note that for the huge "changed" block as shown in the diff display, the only change is 614-620 for the extra code for the button/form.
All the rest is unmodified/original code indented and a correction of the location of the final closing divs. The page seems to have given up trying to make a match.

FYI If you click the "Gear" icon, and then tick the "Hide Whitespace" checkbox and Apply, then the diff will ignore spaces/etc, which fixes up some of the matching failures:
Screen Shot 2025-12-01 at 11 58 34 PM

drbyte:
I noticed in several of your PRs your formatting seems to be leaving the `<?php` on its own line and adding a new linebreak between things that used to be on the same line. I'd prefer to avoid that if we can. Often the line also has a closing `?>` tag and when you add the linebreak before it the opening/closing tags are dissociated from each other, etc. 
eg: 
Screen Shot 2025-12-01 at 11 57 48 PM
There was another one above this screenshot too around line 611 I think?

Were these new line-breaks intentional on your part? Or just the result of an IDE formatter setting?

admin/products_to_categories.php
```diff
@@ -601,141 +608,153 @@
...
-                        }
-                        // Add a class to the checkbox so that it can be identified as a target category checkbox, for the purposes of selecting all/none at once
-                        $zc_categories_checkbox = zen_draw_checkbox_field('categories_add[]',
-                            $categories_list->fields['categories_id'], $selected, '', 'class="TargetCategoryCheckbox"');
-
-                        if ($cnt_columns === 1) {
-                            ?>
-                            <tr class="dataTableRow">
+                    echo zen_draw_form('update', FILENAME_PRODUCTS_TO_CATEGORIES, 'action=update_product&products_filter=' . $products_filter . '&current_category_id=' . $current_category_id . '&target_category_id=' . $target_category_id, 'post');
+                    zen_draw_hidden_field('current_master_categories_id', $product_to_copy->fields['master_categories_id']); ?>
```
drbyte, line 680: Observation: I notice there's no `echo` before the zen_draw_hidden_field() statement (even before this PR) and it's not concatenated onto the end of the echo from the previous line. 
So that means the hidden field isn't actually getting output. Is that causing a bug anywhere? Doesn't look like we're setting it anywhere else in this form. 👀

torvista: Indeed! Does not cause an issue if used...

I'll look over that as I go through the other additions I have pending...

admin/products_to_categories.php
```diff
@@ -601,141 +608,153 @@
...
-                    ?>
-                    <noscript><input type="submit" value="<?= IMAGE_DISPLAY ?>"></noscript>
-                    <?php
-                    echo '</form>';
-                    if ($target_category_id !== (int)P2C_TARGET_CATEGORY_DEFAULT) { // show a Set Default button if the selected target category is different from the saved default
-                        echo zen_draw_form('set_default_target_category_form', FILENAME_PRODUCTS_TO_CATEGORIES, 'action=set_default_target_category' . '&products_filter=' . $products_filter . '&current_category_id=' . $current_category_id, 'post'); ?>
-                        <button type="submit" class="btn btn-info" title="<?= BUTTON_SET_DEFAULT_TARGET_CATEGORY_TITLE ?>"><?= BUTTON_SET_DEFAULT_TARGET_CATEGORY ?></span></button>
+                <?php
+                // button to hide/show the linked categories
+                echo '<p>' . zen_draw_form('hide_linked_categories_form', FILENAME_PRODUCTS_TO_CATEGORIES, zen_get_all_get_params(), 'post');
```
drbyte: This `<form>` should probably be inside a `<div>` instead of a `<p>`.

torvista: ok, won't break the bank.

Before
2025-12-22 22_22_47-Admin Products To Categories — Firefox Developer Edition

After
2025-12-22 22_22_57-Admin Products To Categories — Firefox Developer Edition

admin/products_to_categories.php
drbyte: Turns out it's been this way since v1.5.7 when this field was moved up here from lower down.

drbyte: https://github.com/zencart/zencart/pull/7665 posted for testing

PR #7436, opened 2025-11-30
class order: add language_code to info array for checkout_process

torvista:
...for use by subsequent notifier(s)

### DIFF
--- a/includes/classes/order.php
+++ b/includes/classes/order.php
@@ -25,7 +25,7 @@ class order extends base
      */
     public $attachArray = [];
     /**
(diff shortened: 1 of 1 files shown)


includes/classes/order.php
lat9: Since the `language_code` is now stored in `$this->info['language'code']`, use that instead of the session-based value.

lat9 (changes requested):
See comment

includes/classes/order.php
torvista: Good, I didn't notice that.

torvista:
Got a bit confused there, with many Sunday jobs, but I think I was right first time.
Up to checkout, the cart method in use does not have info->['language_code'] available.

PR #7438, opened 2025-12-01
class order: minor code simplications

torvista:


includes/classes/order.php
```diff
@@ -387,6 +387,7 @@ function query($order_id)
         if (IS_ADMIN_FLAG === true) {
             $this->notify('ORDER_QUERY_ADMIN_COMPLETE', ['orders_id' => $this->orderId]);
         }
+        return true;
```
proseLA: why is this necessary?

torvista: Good to see you are still on quality control.

This method returns false in places, so it seemed correct to return true at the end of the parsing,prior to hinting the method.
However I find no instances of this returned value being checked anywhere....

proseLA: for someone who claims to not be a coder, you really do like playing with code.

i think your reasoning is faulty.

the method `query` should probably never return a boolean.

the fact that it currently returns `false` in some places does not automatically mean that it should return `true` when it does not return `false`.  i think a more valid argument could be made that the places where it returns `false` should be corrected.  but that is a different topic.

it is currently not throwing errors; nor are your correcting a problem; so i am not a fan.

i am a fan of reducing code complexity.  which means less code.  here you are adding a line for no good reason.

unless you are correcting for an error, or adding new functionality, i would not add lines that adhere to your own sense of logic.

torvista: > i think a more valid argument could be made that the places where it returns false should be corrected

Probably.

drbyte: I agree: adding that return false isn't right. "More right" would be to explicitly declare the return type as `void` and cleanup the early returns, but let's do that in v3.

torvista:
can I add hints to this as they are not causing me any trouble in production?

includes/classes/order.php
```diff
@@ -21,126 +21,104 @@ class order extends base
...
      * $customer is an array containing information about the customer for the order
-     * @var array
      */
-    public $customer = [];
+    public array $customer = [];
     /**
      * $delivery is an array containing delivery details for the order
      * @var array
      */
     public $delivery = [];
```
drbyte, line 47: Was there a reason you didn't type-hint this one? Problem in production?

torvista: 296 sets it as false...so maybe up for review there.

torvista:
The only thing I have left is 
880 
$cStart = substr($this->info['cc_number'], 0, ($cOffset > 4 ? 4 : (int)$cOffset));
to
$cStart = substr($this->info['cc_number'], 0, min($cOffset, 4));

PR #7442, opened 2025-12-04
Optionally support GET mode $payload if provided as param instead of on URL

### LINKED ISSUE #7435: function zenDoCurlRequest adding GET parameter not working
I'm wrestling with a third-party api that returns 500 results when no parameter is set or a block of 500 as determined by a page number $page.

So using this
```
 $result = zenDoCurlRequest(
            $this->api_urls['products'],
            payload: ['page' => $page],
            extraCurlOptions: [CURLOPT_HTTPHEADER => ['accept:application/json', "key:$this->api_key"]]
)
```
Always returns the same page as though the page parameter is not set.

Reviewing the function, I don't see anywhere that $payload for GET is appended to the $url.

When I add a second clause for GET after 
`!empty($payload) && strtoupper($method) === 'POST'`

and reset CURLOPT_URL with $url . '?' . $payload

the curl query works correctly, returning unique results per page.

So, seems this handling is incomplete for GET.

Test parameters can be provided on request, by pm,

drbyte:
Fixes #7435

### DIFF
--- a/includes/functions/functions_communications.php
+++ b/includes/functions/functions_communications.php
@@ -120,9 +120,9 @@ function zenDoCurlRequest(
     // json decode if requested and if possible
     if ($response !== false && $decodeJsonResponses && str_contains($info['content_type'], 'application/json')) {
         $rawResponse = $response;
-        $jsonResponse = json_decode($response);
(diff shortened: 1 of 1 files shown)


torvista (commented):
This works in my use case.

PR #7446, opened 2025-12-13
Revised load-order of language defines to before error-handling functions are fired

### LINKED ISSUE #6547: Init order must change for PHP8 (PayPal)
`init_languages.php` is pulled in at 110, but `init_paypal_ipn_sessions.php` is pulled in at 71. 

If PayPal Express logging is turned on and set to `Log and Email` *and* you have Email Admin Format set to HTML, 
 this will crash in ipn_handler.

```
[19-Jun-2024 14:28:09 Europe/London] Request URI: /catalog/ipn_main_handler.php, IP address: ***
--> PHP Fatal error: Uncaught Error: Undefined constant "EMAIL_LOGO_FILENAME" in /home/client/catalog/includes/functions/functions_email.php:637
```

drbyte:
Load language strings before the Cart gets instantiated (and before IPN session inspection loads, since it depends on language strings in order to send any error/debug/warning emails/alerts).

Should fix #6547

### DIFF
--- a/includes/auto_loaders/config.core.php
+++ b/includes/auto_loaders/config.core.php
@@ -158,6 +158,17 @@ $autoLoadConfig[70][] = [
     'loadFile' => 'init_sessions.php',
 ];
 /**
+ * Breakpoint 75.
+ *
+ * require 'includes/init_includes/init_languages.php';
+ *
+ */
+$autoLoadConfig[75][] = [
+    'autoType' => 'init_script',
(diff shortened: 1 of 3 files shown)


torvista:
Note to maintainers, this causes Ceon URI Mapping to not change the site language when the user selects a different storefront language:
https://github.com/torvista/Zen-Cart_CEON-URI-Mapping/issues/41

PR #7459, opened 2026-01-12
add review title

zenexpert:
One thing I always found missing with reviews is the ability for it to have a title. Where ever you go to see product reviews, you always see a title followed by a description. The title is more attention-grabbing and can help the visitor scan the reviews much faster and find what they want faster. 
With the current setup, it's optional - reviews are hard to get anyway, don't want to make the whole process even more difficult.

I'm considering adding an option to have an auto-generated title based on rating if the title is left blank. Thoughts?

Also, the default setting for `REVIEW_TEXT_MIN_LENGTH` is too high - a 5-star rating with the text "works great" is a million times better review than no review because we're asking for a small novel.

```
ALTER TABLE reviews_description ADD reviews_title VARCHAR(128) NOT NULL AFTER reviews_text;
UPDATE configuration SET configuration_value = '5' WHERE configuration_key LIKE 'REVIEW_TEXT_MIN_LENGTH';
```


torvista:
Additionally, should not reviews be always on the product page these days...

torvista:
I think reviews on the product page is a no-brainer, I've had it for years.
An admin reply is also of use, I find people post questions there or even complain instead of asking directly. I don't think it does any harm to allow bad reviews that are politely dealt with.
I understood that it is now also EU law now to categorically state that the reviews are genuine.
I only allow logged-in purchasers of that product to comment.

As for contributing code, I've not seen anything to change my long-established comments, to the effect that until ZC out of the box has a template that looks like it belongs in 2026 and maybe convinces someone to try it out, it'll continue in its decline. So tinkering with code is just ignoring the elephant in the room, and I'm guilty as charged too.

Ex-discussions about raising the dead
Zen Cart is Dead https://github.com/zencart/zencart/discussions/4805
Marketing https://github.com/zencart/zencart/discussions/5094
Roadmap cul-de-sac https://github.com/zencart/zencart/discussions/5090

torvista:
>please check Reviews Advanced
I note that initially I charged ahead/didn't read anything and thought the images were some Additional Images, as they had no text associated, nor admin replies (which I'd be interested in), nor voting options, nor if-purchased indications...all features of the plugin which really should be shown in the demo.

PR #7461, opened 2026-01-14
Orders Status Color Coding

zenexpert:
As part of Modern Admin Dashboard discussion, this adds color-coding to orders statuses. When adding/editing an Orders Status, you can choose the color of the badge that will represent this order status. This badge will show on the orders page in order listing view, as well as the statuses page as preview.

color-coding-status-1

color-coding-status-2

### DIFF
--- a/zc_install/sql/install/mysql_zencart.sql
+++ b/zc_install/sql/install/mysql_zencart.sql
@@ -1344,7 +1344,7 @@ CREATE TABLE orders_status (
   orders_status_id int(11) NOT NULL default 0,
   language_id int(11) NOT NULL default 1,
   orders_status_name varchar(32) NOT NULL default '',
-  orders_status_color_code varchar(7) NULL DEFAULT NULL,
+  orders_status_color_code varchar(7) default NULL,
   sort_order int(11) NOT NULL default 0,
   PRIMARY KEY  (orders_status_id,language_id),
   KEY idx_orders_status_name_zen (orders_status_name)
--- a/zc_install/sql/updates/mysql_upgrade_zencart_220.sql
+++ b/zc_install/sql/updates/mysql_upgrade_zencart_220.sql
@@ -62,7 +62,7 @@ ALTER TABLE plugin_control MODIFY version varchar(20);
 ALTER TABLE plugin_control_versions MODIFY version varchar(20);
 ALTER TABLE customers ADD COLUMN activation_required tinyint(1) NOT NULL DEFAULT 0 AFTER customers_authorization;
 ALTER TABLE customers ADD COLUMN welcome_email_sent tinyint(1) DEFAULT NULL AFTER activation_required;
-ALTER TABLE orders_status ADD orders_status_color_code VARCHAR(7) NULL DEFAULT NULL AFTER orders_status_name;
+ALTER TABLE orders_status ADD orders_status_color_code VARCHAR(7) DEFAULT NULL AFTER orders_status_name;
 
 #PROGRESS_FEEDBACK:!TEXT=Updating configuration settings...
 DELETE FROM configuration WHERE configuration_key IN ('REPORT_ALL_ERRORS_ADMIN', 'REPORT_ALL_ERRORS_STORE', 'REPORT_ALL_ERRORS_NOTICE_BACKTRACE');


zenexpert:
Required SQL:
`ALTER TABLE orders_status ADD orders_status_color_code VARCHAR(7) NULL DEFAULT NULL AFTER orders_status_name`

admin/orders.php
```diff
@@ -1370,6 +1370,14 @@ function couponpopupWindow(url) {
...
                                     $orders_split = new splitPageResults($_GET['page'], MAX_DISPLAY_SEARCH_RESULTS_ORDERS, $orders_query_raw, $orders_query_numrows);
                                     $orders = $db->Execute($orders_query_raw);
+                                      // prepare status colors array
+                                      $status_colors = [];
+                                      $status_color_query = $db->Execute("SELECT orders_status_id, orders_status_color_code
+                                                        FROM " . TABLE_ORDERS_STATUS . "
+                                                        WHERE language_id = " . (int)$_SESSION['languages_id']);
+                                      foreach ($status_color_query as $status) {
+                                          $status_colors[$status['orders_status_id']] = $status['orders_status_color_code'];
+                                      }
```
proseLA: i would implore you to look at this section of code:
https://github.com/zencart/zencart/blob/ca4f8e9557ac7d7556cb48a36494757b8168dc5c/admin/includes/functions/general.php#L769-L789
and add the color there to both of the return arrays.

if you then look here:
https://github.com/zencart/zencart/blob/ca4f8e9557ac7d7556cb48a36494757b8168dc5c/admin/orders.php#L43

you will see that the color for each status code will now be available without the extra db call and loop.

admin/orders_status.php
```diff
@@ -200,6 +233,16 @@
                 }
 
                 $contents[] = array('text' => '<br>' . TEXT_INFO_ORDERS_STATUS_NAME . $orders_status_inputs_string);
+                        $textVal = $oInfo->orders_status_color_code;
+                        $pickerVal = getPickerVal($oInfo->orders_status_color_code);
+                        $contents[] = array('text' => '<br>' . TEXT_INFO_HEADING_COLOR_CODE . '<br>' .
+                            '<div class="input-group">' .
+                            zen_draw_input_field('orders_status_color_code', $textVal, 'class="form-control" id="cp-hex" placeholder="' . TEXT_INFO_COLOR_BOX_PLACEHOLDER . '"') .
+                            '<span class="input-group-addon" style="padding:0; border:none; width:40px; background:none;">' .
+                            '<input type="color" id="cp-picker" value="' . $pickerVal . '" style="height:34px; width:100%; padding:0; border:none; cursor:pointer;">' .
```
proseLA, line 242: again, i am not sure why you insist on using inline styling.  you have an id and class.  can we use a stylesheet?

admin/orders.php
zenexpert: fantastic idea, love it. Never thought of it

admin/orders_status.php
zenexpert: Personally, when something is NOT meant to be changed by someone else and is meant to look just like it was originally done, I prefer inline CSS since it's a tiny bit less prone to changes. But sure, updated.

admin/includes/functions/general.php
```diff
@@ -764,16 +764,18 @@ function zen_get_orders_status_pulldown_array()
 }
 
 /**
- * @since ZC v2.0.0
+ * @since ZC v2.2.0
```
lat9: Why was this line updated?  The intent of the `@since` tag is to indicate the Zen Cart version at which the function was introduced.

Please revert to v2.0.0.

lat9 (changes requested):

admin/includes/functions/general.php
zenexpert: My bad - I thought the `@since` line indicates *changes* to the function, not the origin. As per docs: _The @since tag can be used to document the "version" of the introduction **or modification** of a Structural Element ._

Should I add a new `@since` tag (or `@version` or `@modified`)? No clue which standard you decided to follow.

lat9: @drbyte went through a couple of months ago and added those `@since` tags.  The "version of introduction" was his choice, mostly to help plugin authors who are supporting multiple Zen Cart versions.

Any 'new' function, class or class method should also identify the Zen Cart version of introduction.

zenexpert: Yeah, I saw the pull request, and I believe it's based on phpdoc: https://docs.phpdoc.org/guide/references/phpdoc/tags/since.html
I'm guessing the idea is just to show when it was introduced, but not when it was modified or to indicate any changes. 

OK, done.

zc_install/sql/install/mysql_zencart.sql
```diff
@@ -1344,6 +1344,7 @@ CREATE TABLE orders_status (
   orders_status_id int(11) NOT NULL default 0,
   language_id int(11) NOT NULL default 1,
   orders_status_name varchar(32) NOT NULL default '',
+  orders_status_color_code varchar(7) NULL DEFAULT NULL,
```
lat9: The `NULL` after `varchar(7)` is redundant and confusing, please remove here and in the upgrade SQL as well.

PR #7462, opened 2026-01-17
admin, orders: change tr open tag from php to html to balance with tr close tag + misc

torvista:
Filesystem typo
remove unnecessary issets
<strike>replace javascript void</strike>
remove unnecessary SQL ASC
remove default parameter
use str_starts_with
correct type of string concatenation/addition for date('y')+1
remove unnecessary cast of $show_orders_weights

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -459,7 +459,7 @@ if (!empty($action) && $order_exists === true) {
                                 <?php
                                 echo zen_draw_form('search_orders_products', FILENAME_ORDERS, '', 'get', '', true);
                                 echo zen_draw_label(HEADING_TITLE_SEARCH_DETAIL_ORDERS_PRODUCTS, 'searchProduct', 'class="sr-only"');
-                                $placeholder = zen_output_string_protected(isset($_GET['search_orders_products']) && !empty($_GET['search_orders_products']) ? $_GET['search_orders_products'] : HEADING_TITLE_SEARCH_PRODUCTS);
+                                $placeholder = zen_output_string_protected(!empty($_GET['search_orders_products']) ? $_GET['search_orders_products'] : HEADING_TITLE_SEARCH_PRODUCTS);
                                 ?>
                             <div class="input-group">
                                     <?php
@@ -742,7 +742,7 @@ if (is_array($address_footer_suffix)) {
     // orders_products::products_weight.    If the order's weight is stored as null, don't
     // display the weights!
     //
-    $show_orders_weights = ($order->info['order_weight'] !== null && ((bool)($show_orders_weights ?? true)));
+    $show_orders_weights = $order->info['order_weight'] !== null && ($show_orders_weights ?? true);
 ?>
                         </table>
                     </div>
@@ -810,7 +810,7 @@ if (is_array($address_footer_suffix)) {
                             if (!empty($order->info['cc_number'])) {
                                             require DIR_FS_CATALOG . DIR_WS_CLASSES . 'cc_validation.php';
                                             $cc_valid = new cc_validation();
-                                            $cc_needs_mask = $cc_valid->validate($order->info['cc_number'], date('m'), date('y')+1);
+                                            $cc_needs_mask = $cc_valid->validate($order->info['cc_number'], date('m'), (int)date('y')+1);
                                     ?>
                             <tr>
                                 <td class="main"><?= ENTRY_CREDIT_CARD_NUMBER ?></td>
@@ -1352,7 +1352,7 @@ if ($show_orders_weights === true) {
                                                     'op.products_id',
                                             ];
                                             $search = zen_build_keyword_where_clause($keyword_search_fields, trim($keywords), true);
-                                        if (substr(strtoupper($_GET['search_orders_products']), 0, 3) === 'ID:') {
+                                        if (str_starts_with(strtoupper($_GET['search_orders_products']), 'ID:')) {
                                             $keywords = trim(substr($_GET['search_orders_products'], 3));
                                             $search .= " OR op.products_id ='" . (int)$keywords . "'";
                                         }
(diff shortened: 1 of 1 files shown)


admin/orders.php
```diff
@@ -108,7 +108,7 @@
         $order = new order($oID);
         $module = false;
         global $installedPlugins;
-        $moduleFinder = new ModuleFinder('payment', new Filesystem());
+        $moduleFinder = new ModuleFinder('payment', new FileSystem());
```
proseLA: if one is dealing with `Case in class usage doesn't match the case in declaration`; why not issue a PR to fix all of them?

https://github.com/zencart/zencart/blob/a1cc9bc111bf2dadc73147d9a77ecbcb036ab4e5/includes/classes/order_total.php#L53
https://github.com/zencart/zencart/blob/a1cc9bc111bf2dadc73147d9a77ecbcb036ab4e5/includes/classes/shipping.php#L73
https://github.com/zencart/zencart/blob/a1cc9bc111bf2dadc73147d9a77ecbcb036ab4e5/includes/classes/payment.php#L74
https://github.com/zencart/zencart/blob/a1cc9bc111bf2dadc73147d9a77ecbcb036ab4e5/admin/layout_controller.php#L36
https://github.com/zencart/zencart/blob/a1cc9bc111bf2dadc73147d9a77ecbcb036ab4e5/admin/modules.php#L51

no good deed....

torvista: Subsequently I see there are multiple uses in Laravel of Filesystem, so not sure which is strictly correct...

proseLA: > Subsequently I see there are multiple uses in Laravel of Filesystem, so not sure which is strictly correct...

i would not touch anything in the Laravel directory.  if you think something is wrong in there, i would encourage you to open an issue.  the odds of something being wrong in there are much more slim, but not zero.

the references in Laravel refer to something completely different.  they refer to this class:

https://github.com/zencart/zencart/blob/a1cc9bc111bf2dadc73147d9a77ecbcb036ab4e5/laravel/vendor/illuminate/filesystem/Filesystem.php#L17

so they are correct.  

the one that you have referenced (and the others to which i have pointed out) refer to this class:

https://github.com/zencart/zencart/blob/a1cc9bc111bf2dadc73147d9a77ecbcb036ab4e5/includes/classes/FileSystem.php#L15

i would suggest making an attempt to understand what you are doing prior to issuing a PR.  while i personally applaud your efforts, your comment above suggests to me that you are guessing what is correct, based on what others may have done.  whether or not that is a good strategy is debatable in my mind...

torvista: I live and learn, within my limits. Always grateful for constructive comments.

proseLA:
> Any obstacle here?

i made 1 comment on 1 element of this PR.  if this comment is for me, i am not inclined to test all of the other changes that you have made.  but the `FileSystem` v `Filesystem` looks good to me.

admin/orders.php
```diff
@@ -1579,10 +1577,8 @@ function couponpopupWindow(url) {
                                             <td class="text-right" colspan="2">
                                             <?php
                                                     echo '<a href="' . zen_href_link(FILENAME_ORDERS, '', 'NONSSL') . '" class="btn btn-default" role="button">' . IMAGE_RESET . '</a>';
-                                                    if (isset($_GET['search']) && zen_not_null($_GET['search'])) {
```
drbyte, line 1596: Why remove this clause entirely?

admin/orders.php
```diff
@@ -458,7 +458,7 @@ function couponpopupWindow(url) {
                                 <?php
                                 echo zen_draw_form('search_orders_products', FILENAME_ORDERS, '', 'get', '', true);
                                 echo zen_draw_label(HEADING_TITLE_SEARCH_DETAIL_ORDERS_PRODUCTS, 'searchProduct', 'class="sr-only"');
-                                $placeholder = zen_output_string_protected(isset($_GET['search_orders_products']) && !empty($_GET['search_orders_products']) ? $_GET['search_orders_products'] : HEADING_TITLE_SEARCH_PRODUCTS);
```
drbyte, line 462: isset checks for whether it's present.
!empty ignores if it's falsey, including if it's 0 or '0', which is not the same as blank/unspecified.
So ... if someone's provided '0' as a single search character, is the intent here to not show that?

admin/orders.php
```diff
@@ -108,7 +108,7 @@
         $order = new order($oID);
         $module = false;
         global $installedPlugins;
-        $moduleFinder = new ModuleFinder('payment', new Filesystem());
```
drbyte, line 112: I agree with the case correction. 
Should have been its own PR, or part of a PR with other files where same fix was done. Quick decision to merge.

admin/orders.php
```diff
@@ -545,7 +545,7 @@ function couponpopupWindow(url) {
                 $result = $db->Execute("SELECT orders_id
                                                                     FROM " . TABLE_ORDERS . "
                                                                     WHERE orders_id > " . (int)$oID . "
-                                                                    ORDER BY orders_id ASC
```
drbyte, line 549: Often the reason for specifying a parameter that's also the default, is to indicate that there was intentionality behind specifying it.

I'd prefer not to be removing those indicators.

admin/orders.php
```diff
@@ -1557,7 +1555,7 @@ function couponpopupWindow(url) {
                                         if (isset($oInfo) && is_object($oInfo) && ($orders->fields['orders_id'] == $oInfo->orders_id)) {
                                             echo zen_icon('caret-right', '', '2x', true);
                                         } else {
-                                            echo '<a href="' . zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(['oID']) . 'oID=' . $orders->fields['orders_id'], 'NONSSL') . '" data-toggle="tooltip" title="' . IMAGE_ICON_INFO . '" role="button">' . zen_icon('circle-info', '', '2x', true, false) . '</a>';
```
drbyte, line 1574: Same. Do we know why the `false` was included? If not, is it really important to remove it?

admin/orders.php
```diff
@@ -1617,7 +1613,7 @@ function couponpopupWindow(url) {
                                         // each contents array is drawn in a div, so this form block must be a single array element.
                                         $contents[] = ['text' =>
                                                 zen_draw_form('statusUpdateForm', FILENAME_ORDERS, zen_get_all_get_params(['action','language']) . 'action=update_order' . (!isset($_GET['oID']) ? '&oID=' . $oInfo->orders_id : '') . '&language=' . $oInfo->language_code, 'post', '', true) . // form action uses the order language to change the session language on the update. On initial page load (from another page), $_GET['oID'] is not set, hence clause in form action
-                                                '<fieldset style="border:solid thin slategray;padding:5px"><legend style="width:inherit;">&nbsp;' . IMAGE_UPDATE . '&nbsp;</legend>' .
```
drbyte, line 1634: Since you've got me in nitpicking mode because these things have nothing to do with the title of this PR (TR open tags, balancing with close tags), I'll ask: how important is this `;` ? Isn't it implied by default by the browser?

admin/orders.php
```diff
@@ -1428,16 +1428,14 @@ function couponpopupWindow(url) {
...
                                         }
-
-                                        if (isset($oInfo) && is_object($oInfo) && ($orders->fields['orders_id'] == $oInfo->orders_id)) {
-                                            echo '<tr id="defaultSelected" class="dataTableRowSelected order-listing-row" data-oid="' . $orders->fields['orders_id'] . '" data-current="current">' . "\n";
-                                        } else {
-                                            echo '<tr class="dataTableRow order-listing-row" data-oid="' . $orders->fields['orders_id'] . '">' . "\n";
-                                        }
-
+                                        ?>
+                                        <tr data-oid="<?= $orders->fields['orders_id'] ?>" <?= (isset($oInfo) && is_object($oInfo) && ($orders->fields['orders_id'] == $oInfo->orders_id)
```
drbyte, line 1437: While I appreciate the effort in this, a 3-line if/else/then string output that's complex enough that it needs to be wrapped with parentheses for safety, probably warrants using a normal `echo` statement instead of `<?=` which is normally used for terseness/brevity, especially when the associated `<?=` starts in the middle of a nested set of other string outputs.

admin/orders.php
torvista: because it's already in the same clause:
https://github.com/zencart/zencart/blob/c700463b3623a096a3af57c5e35ca204fda901ec/admin/orders.php#L1590-L1596

admin/orders.php
torvista: Original code uses isset and !empty, so the isset seems unnecessary.

admin/orders.php
torvista: Don't know.

admin/orders.php
torvista: You mean leave it alone/revert it?

admin/orders.php
torvista: reverted to another PR

admin/orders.php
torvista: reverted

admin/orders.php
torvista: reverted

admin/orders.php
torvista: removed

PR #7467, opened 2026-01-24
Admin, collectinfo.php: replace <?php echo, some formatting

torvista:
no code changes

### DIFF
--- a/admin/includes/modules/collect_info.php
+++ b/admin/includes/modules/collect_info.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @copyright Copyright 2003-2025 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
@@ -97,8 +98,8 @@ if (!$category_lookup->EOF) {
 
 $manufacturers_array = [
     [
-    'id' => '',
-    'text' => TEXT_NONE,
+        'id' => '',
+        'text' => TEXT_NONE,
     ]
 ];
 $manufacturers = $db->Execute("SELECT manufacturers_id, manufacturers_name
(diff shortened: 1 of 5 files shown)


admin/includes/modules/collect_info.php
```diff
@@ -96,10 +97,7 @@
 }
 
 $manufacturers_array = [
-    [
-    'id' => '',
-    'text' => TEXT_NONE,
-    ]
+    ['id' => '', 'text' => TEXT_NONE,]
```
proseLA: i am not a fan of this change.  personally i would add 4 spaces to lines 100 and 101.  but as there is no definitive style guide for this repo, and you seem to be intent on issuing PRs that merely reformat code, i will speak no further on this section of code.

admin/includes/modules/collect_info.php
```diff
@@ -117,47 +115,55 @@
...
+        FILENAME_PRODUCT,
+        'cPath=' . $current_category_id .
+        (isset($_GET['pID']) ? '&pID=' . $_GET['pID'] : '') .
+        '&action=new_product_preview' .
+        (isset($_GET['page']) ? '&page=' . $_GET['page'] : '') .
+        ((isset($_GET['search']) && !empty($_GET['search'])) ? '&search=' . zen_preserve_search_quotes($_GET['search']) : '') .
+        ((isset($_POST['search']) && !empty($_POST['search']) && empty($_GET['search'])) ? '&search=' . zen_preserve_search_quotes($_POST['search']) : ''),
+        'post',
+        'enctype="multipart/form-data" class="form-horizontal"'
+    );
```
proseLA, line 132: you seem to have missed the `<?=` on line 118.  

further, you seem to have split the vars for the method call on separate lines here and elsewhere in this PR.  why the 3rd var in this section of code gets various string elements on a separate line, you got me.  

as far as the separate line for each var, i am not sure if this is intended by you, or you are just defaulting to what your IDE does.  which, while really controlled by you and its settings, in my experience, i have found it difficult to find the settings to have my IDE do what i would like it to do in this situation.

again formatting calls are difficult when there is truly no accepted style sheet for this repo.

admin/includes/modules/collect_info.php
torvista: I thought that it was readable enough on one line.
But since you've bothered to examine this, it's no bother to change it.

admin/includes/modules/collect_info.php
torvista: While a var per line would normally be sufficient, this particular one is so long I considered it better to split into its chunks to understand better... subsequently it becomes more obvious that some checks can be removed/it can be simplified later.

admin/includes/modules/collect_info.php
```diff
@@ -117,47 +118,55 @@
...
-    echo zen_draw_form('new_product', FILENAME_PRODUCT, 'cPath=' . $current_category_id . (isset($_GET['pID']) ? '&pID=' . $_GET['pID'] : '') . '&action=new_product_preview' . (isset($_GET['page']) ? '&page=' . $_GET['page'] : '') . ( (isset($_GET['search']) && !empty($_GET['search'])) ? '&search=' . zen_preserve_search_quotes($_GET['search']) : '') . ( (isset($_POST['search']) && !empty($_POST['search']) && empty($_GET['search'])) ? '&search=' . zen_preserve_search_quotes($_POST['search']) : ''), 'post', 'enctype="multipart/form-data" class="form-horizontal"');
+    <?php echo zen_draw_form(
+        'new_product',
+        FILENAME_PRODUCT,
+        'cPath=' . $current_category_id .
+        (isset($_GET['pID']) ? '&pID=' . $_GET['pID'] : '') .
+        '&action=new_product_preview' .
+        (isset($_GET['page']) ? '&page=' . $_GET['page'] : '') .
+        ((isset($_GET['search']) && !empty($_GET['search'])) ? '&search=' . zen_preserve_search_quotes($_GET['search']) : '') .
+        ((isset($_POST['search']) && !empty($_POST['search']) && empty($_GET['search'])) ? '&search=' . zen_preserve_search_quotes($_POST['search']) : ''),
```
lat9: Lines 128 and 129 don't need the `isset` call since `!empty` will fail if the variable isn't set.

admin/includes/modules/collect_info.php
```diff
@@ -117,47 +118,55 @@
...
+      <button type="submit" class="btn btn-primary"><?= IMAGE_PREVIEW ?></button>&nbsp;&nbsp;
+        <a href="<?= zen_href_link(
+            FILENAME_CATEGORY_PRODUCT_LISTING,
+            'cPath=' . $current_category_id .
+            (isset($_GET['pID']) ? '&pID=' . $_GET['pID'] : '') .
+            (isset($_GET['page']) ? '&page=' . $_GET['page'] : '') .
+            ((isset($_GET['search']) && !empty($_GET['search'])) ? '&search=' . zen_preserve_search_quotes($_GET['search']) : '') .
+            ((isset($_POST['search']) && !empty($_POST['search']) && empty($_GET['search'])) ? '&search=' . zen_preserve_search_quotes($_POST['search']) : '')) ?>" class="btn btn-default" role="button"><?= IMAGE_CANCEL ?></a>
```
lat9: Same comment as above; the `isset` portion of the clauses on lines 146 and 147 are not needed.

lat9 (changes requested):
Please don't force-push changes, since that makes the push 'brand new' and all changes need to be re-reviewed.

admin/includes/modules/collect_info.php
torvista: done + another

lat9 (approved):
Thanks for those!

torvista:
Anything else I need to do, or you want the rest in the same PR?

drbyte:
It's ideal to add the other product-types to this same PR. Keeps the changes together for easier review if  looking for why/when things changed.

torvista:
ok, done.

PR #7480, opened 2026-02-06
Warning Log when Setting a Country Status to Inactive

### LINKED ISSUE #7479: Warning Log when Editing a Country in Admin
In ZC v2.2.0-alpha and probably older ones too, when editing a country in admin, changing status from active to inactive will trigger a Warning log as follow:

```
[06-Feb-2026 17:15:06 Asia/Tokyo] Request URI: /admin/index.php?cmd=countries&cID=240&action=save, IP address: 000.000.000.000, Language id 1
#0 ...\admin\countries.php(82): zen_debug_error_handler()
#1 ...\admin\index.php(16): require('...')
--> PHP Warning: Undefined array key "status" in ...\admin\countries.php on line 82.
```

This log is not produced when changing from inactive to active or if status is changed using the green button in Status column.

Looking at file `\admin\countries.php` on lines 82 and 65 (same problem when adding a new country set to inactive), a `'status'` data is expected in `$_POST` variable, but when editing using the right box, status is directly saved to database, then a redirect is done. It seems that when setting status to inactive, which is a value of 0, then no POST data is created during redirect.

A fix could be to replace these lines:
`      $status = $_POST['status'] == 'on' ? 1 : 0;`
By this:
`      $status = array_key_exists('status', $_POST) ? ($_POST['status'] == 'on' ? 1 : 0) : 0;`

piloujp:
Fixes #7479

### DIFF
--- a/admin/countries.php
+++ b/admin/countries.php
@@ -62,7 +62,7 @@ if (!empty($action)) {
       $countries_iso_code_2 = strtoupper(zen_db_prepare_input($_POST['countries_iso_code_2']));
       $countries_iso_code_3 = strtoupper(zen_db_prepare_input($_POST['countries_iso_code_3']));
       $address_format_id = zen_db_prepare_input($_POST['address_format_id']);
-      $status = array_key_exists('status', $_POST) ? ($_POST['status'] == 'on' ? 1 : 0) : 0;
+      $status = ($_POST['status'] ?? '') === 'on' ? 1 : 0;
 
       $db->Execute("INSERT INTO " . TABLE_COUNTRIES . " (countries_name, countries_iso_code_2, countries_iso_code_3, status, address_format_id)
                     VALUES ('" . zen_db_input($countries_name) . "',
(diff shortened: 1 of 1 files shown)


admin/countries.php
```diff
@@ -62,7 +62,7 @@
       $countries_iso_code_2 = strtoupper(zen_db_prepare_input($_POST['countries_iso_code_2']));
       $countries_iso_code_3 = strtoupper(zen_db_prepare_input($_POST['countries_iso_code_3']));
       $address_format_id = zen_db_prepare_input($_POST['address_format_id']);
-      $status = $_POST['status'] == 'on' ? 1 : 0;
+      $status = array_key_exists('status', $_POST) ? ($_POST['status'] == 'on' ? 1 : 0) : 0;
```
lat9: For this (and the change on line 82), please use the null-coalesce operator instead of `array_key_exists`:

```php
$status = ($_POST['status'] ?? '') === 'on' ? 1 : 0;
```

piloujp: Done.

Issue #7481, opened 2026-02-13
Parameter pollution and CPU wasting


zenexpert:
**Describe the bug**
Seeing this on multiple sites: bots (mostly through meta/facebook crawlers) are hitting the index.php with repetitive query parameters, causing a large number of logs to be created and wasting CPU cycles. Resource Usage is showing CPU hanging around top limits, often causing HTTP 508 errors.

**Version**
Zen Cart version: 1.5.8-2.1.0

**To Reproduce**
Steps to reproduce/demonstrate the behavior:
Method: `GET`
Path: `/index.php`
Query string:
`?main_page=index&cPath=116&sort=2d&view=rows%C2%A4cy%3DMXN%C2%A4cy%3DUSD%C2%A4cy%3DMXN%C2%A4cy%3DUSD%C2%A4cy%3DUSD%C2%A4cy%3DMXN&language=en&page=183&action=buy_now&products_id=16947 `

**Expected behavior**
All such requests should be terminated early to avoid wasting CPU cycles. Bots should not be able to add to cart or use `action=buy_now`

**Screenshots**

![Image](https://github.com/user-attachments/assets/ca40fa3c-37f1-4ed1-b499-3f6aa2c10b1c)

lat9:
Closing after merge of PR #7482.

PR #7482, opened 2026-02-13
parameter pollution and CPU cycle wasting

zenexpert:
possible fix for #7481

### DIFF
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -155,7 +155,9 @@ if (!empty($_SERVER['QUERY_STRING'])) {
         $parts = explode('=', $pair, 2);
 
         // skip if the pair is empty (e.g., &&) or the key is missing
-        if (empty($parts[0])) continue;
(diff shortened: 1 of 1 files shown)


includes/application_top.php
```diff
@@ -122,7 +122,75 @@
...
+    // break the query string into individual "key=value" pairs
+    $pairs = explode('&', $_SERVER['QUERY_STRING']);
+    $keys = [];
+
+    foreach ($pairs as $pair) {
+        // get just the part before the "="
+        $parts = explode('=', $pair, 2);
+
+        // skip if the pair is empty (e.g., &&) or the key is missing
+        if (empty($parts[0])) continue;
```
lat9: No one-line conditionals.  Use
```php
if (empty($parts[0])) {
    continue;
}
```

lat9 (approved):
Thanks, @zenexpert!

PR #7488, opened 2026-02-19
Include `disp_order` dropdown in `search_results` page

### LINKED ISSUE #7483: class.search.php uses legacy $_GET['sort'] instead of the newer $_GET['disp_order']
**Describe the bug**
class.search.php uses legacy $_GET['sort'] instead of the newer $_GET['disp_order']

Should be updated to be consistent. 

Appears to still be a problem in 2.2.0-alpha.

lat9:
Fixes #7483

### DIFF
--- a/includes/classes/class.search.php
+++ b/includes/classes/class.search.php
@@ -262,7 +262,7 @@ class Search extends \base
             $from_str = $db->bindVars($from_str, ':languagesID', $_SESSION['languages_id'], 'integer');
         }
 
-        if (DISPLAY_PRICE_WITH_TAX === 'true' && !empty($this->searchOptions->pfrom) || !empty($this->searchOptions->pto)) {
+        if (DISPLAY_PRICE_WITH_TAX === 'true' && (!empty($this->searchOptions->pfrom) || !empty($this->searchOptions->pto))) {
             if (empty($_SESSION['customer_country_id'])) {
                 $_SESSION['customer_country_id'] = STORE_COUNTRY;
                 $_SESSION['customer_zone_id'] = STORE_ZONE;


includes/classes/class.search.php
```diff
@@ -246,14 +255,14 @@ public function buildSearchSQL() {
...
-        if (ADVANCED_SEARCH_INCLUDE_METATAGS == 'true') {
+        if (ADVANCED_SEARCH_INCLUDE_METATAGS === 'true') {
             $from_str .=
                 " LEFT JOIN " . TABLE_META_TAGS_PRODUCTS_DESCRIPTION . " mtpd
                     ON (mtpd.products_id= p2c.products_id AND mtpd.language_id = :languagesID)";
             $from_str = $db->bindVars($from_str, ':languagesID', $_SESSION['languages_id'], 'integer');
         }
 
-        if ((DISPLAY_PRICE_WITH_TAX == 'true') && !empty($this->searchOptions->pfrom) || !empty($this->searchOptions->pto)) {
+        if (DISPLAY_PRICE_WITH_TAX === 'true' && !empty($this->searchOptions->pfrom) || !empty($this->searchOptions->pto)) {
```
scottcwilson: Do you not want brackets around the two not empty checks? 

if (DISPLAY_PRICE_WITH_TAX === 'true' && (!empty($this->searchOptions->pfrom) || !empty($this->searchOptions->pto))) {

lat9: That'd be my 'guess' as well.  I was hesitant to make that change since it affects that `if` clause's outcome.

drbyte: Probably good to add the paren's, making it the same as line 247 above, which builds the `select` for table `tr.` that matches the `tr.` in the `left join` being built here.

PR #7489, opened 2026-02-20
Remove Laravel

drbyte:
Remove Laravel dependency from core.

Credits @zcwilt

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -99,29 +99,30 @@
         },
         {
             "name": "doctrine/instantiator",
-            "version": "2.1.0",
+            "version": "2.0.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/instantiator.git",
-                "reference": "23da848e1a2308728fe5fdddabf4be17ff9720c7"
+                "reference": "c6222283fa3f4ac679f8b9ced9a4e23f163e80d0"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/instantiator/zipball/23da848e1a2308728fe5fdddabf4be17ff9720c7",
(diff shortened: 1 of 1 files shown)


includes/classes/DbRepositories/ConfigurationRepository.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+    }
+
+    public function loadConfigSettings(): void
+    {
+        $configs = $this->db->Execute(
+            'SELECT configuration_key, configuration_value, configuration_group_id FROM ' . TABLE_CONFIGURATION
+        );
+
+        while (!$configs->EOF) {
```
lat9: Why while/MoveNext instead of foreach?

includes/classes/DbRepositories/ProductTypeLayoutRepository.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+    }
+
+    public function loadConfigSettings(): void
+    {
+        $configs = $this->db->Execute(
+            'SELECT configuration_key, configuration_value FROM ' . TABLE_PRODUCT_TYPE_LAYOUT
+        );
+
+        while (!$configs->EOF) {
```
lat9: Same question; why not `foreach`?

drbyte:
> If I run currency_cron from the browser (after removing the browser check in currency_cron), function mb_internal_encoding exists.
> When currency_cron is run from the cron job, function mb_internal_encoding does not exist.

That looks like your browser is using a different PHP configuration (PHP-FPM) than your command-line (cron job) (which uses PHP CLI). And whatever PHP version is configured for command-line doesn't have `mb_string` enabled.

(Yes, you're right that removing Laravel removed the much broader Mbstring polyfill that was masking your mixed configuration.)

torvista:
Thanks, live and learn.

PR #7505, opened 2026-03-04
functions_banner_graphs.php: cast date (returns string) to int for adding 1

torvista:
and for use in subsequent loop.

### DIFF
--- a/admin/includes/functions/functions_banner_graphs.php
+++ b/admin/includes/functions/functions_banner_graphs.php
@@ -96,7 +96,7 @@ function zen_get_banner_data_daily($banner_id, $year = '', $month = '') {
 
     $set1 = $set2 = array();
 
-    $days = (date('t', mktime(0,0,0,$month))+1);
+    $days = (int)date('t', mktime(0,0,0,$month)) + 1;
     for ($i=1; $i<$days; $i++) {
       $set1[] = $set2[] = $stats[] = array($i, 0);
     }


admin/includes/functions/functions_banner_graphs.php
```diff
@@ -96,7 +96,7 @@ function zen_get_banner_data_daily($banner_id, $year = '', $month = '') {
 
     $set1 = $set2 = array();
 
-    $days = (date('t', mktime(0,0,0,$month))+1);
+    $days = (int)date('t', mktime(0,0,0,$month)) + 1;
```
drbyte, line 99: Hmmm ... doesn't the math force the conversion to `int` automatically, in order to do the add?

If not, then in the other PR (already merged) where you cast int for month/year but not day, then the day needs casting to int too.

drbyte: I did some digging and testing.
Yes: strict_types only affects function parameter type declarations and return type declarations.
And math/arithmetic operators themselves still perform numeric coercion even with strict types.

drbyte: So that makes me wonder how you arrived at this code submission ... was some strict error getting triggered?

torvista: inspection error was Wrong string concatenation...adding an integer to the string.

drbyte: Okay, yes, that's a false-positive by the inspector. It's harder for it to know for certain that you're getting back a numeric-only string instead of a formatted date string. 
And it's also a bit strange that it's calling it concatenation (when the `+` is used, but accepted as "fine" if `-` is used).

For those reasons, I'm gonna close this.

PR #7506, opened 2026-03-04
zen_draw_date_selector: cast date results to int for loop and addition/subtraction of integers

torvista:


admin/includes/functions/html_output.php
```diff
@@ -672,7 +672,7 @@ function zen_draw_date_selector(string $fieldname_prefix, ?int $default_date_tim
     }
     $date_selector .= '</select>';
     $date_selector .= '<select name="'. $fieldname_prefix .'_year">';
-    for ($i = date('Y') - 5, $j = date('Y') + 11; $i < $j; $i++) {
+    for ($i = (int)date('Y') - 5, $j = (int)date('Y') + 11; $i < $j; $i++) {
```
drbyte, line 675: IMO this is not needed, because PHP does its own automatic numeric coersion here for the math.

PR #7511, opened 2026-03-08
Correct `order_status` payment display

lat9:
... was using the wrong variable in the conditional.


lat9:
Links are sent in guest-checkout order-confirmation emails.

PR #7515, opened 2026-03-12
Encapsulated "PayPal Restful" payment module for core

### LINKED ISSUE #7490: PayPal RESTful payment method needs to be added to core
Per @drbyte: PayPal is pressing us to migrate away from the old integrations to REST.  To this end, it would be helpful to include the updated integration https://www.zen-cart.com/downloads.php?do=file&id=2382 in the 2.2 Release.

lat9:
Fixes #7490.

Uses v1.3.1 (https://github.com/lat9/paypalr/tree/v1.3.1) of the unencapsulated payment module as the change basis.

The one issue that I've seen is that the credit-card images "can't be displayed".  I've looked at the `.htaccess` in the `zc_plugins` directory and don't see why that would be affecting the load.

### DIFF
--- a/zc_plugins/PayPalRestful/v2.0.0/manifest.php
+++ b/zc_plugins/PayPalRestful/v2.0.0/manifest.php
@@ -9,4 +9,5 @@ return [
     'changelog' => '', // online URL (eg github release tag page, or changelog file there) or local filename only, ie: changelog.txt (in same dir as this manifest file)
     'github_repo' => '', // url
     'pluginGroups' => [],
+    'removesUnencapsulatedVersion' => true, // !empty presence indicates that non-encapsulated files are removed during install
 ];
--- a/admin/includes/languages/english/lang.plugin_manager.php
+++ b/admin/includes/languages/english/lang.plugin_manager.php
@@ -50,6 +50,7 @@ $define = [
     'ERROR_NOT_FOUND_IN_SQL_FUNCTIONS_MAP' => 'Check your SQL statement. A SQL function map cannot be found for : ',
     'ERROR_INVALID_SYNTAX' => 'The table cannot be identified as syntax is invalid in: ',
     'ERROR_SQL_PATCH' => 'Error while processing SQL install. ',
+    'WARNING_NONENCAPSULATED_REMOVAL' => '<b>Note:</b> Installing this plugin will result in files provided by a non-encapsulated version (if present) being <b>permanently</b> removed!',
 ];
 
 return $define;
--- a/includes/classes/ViewBuilders/PluginManagerController.php
+++ b/includes/classes/ViewBuilders/PluginManagerController.php
@@ -31,6 +31,27 @@ class PluginManagerController extends BaseController
     }
 
     /**
+     * @since ZC v2.2.0
+     */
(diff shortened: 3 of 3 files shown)


zc_plugins/PayPalRestful/v2.0.0/catalog/includes/modules/payment/paypal/PayPalRestful/Api/PayPalRestfulApi.php
```diff
@@ -587,7 +564,7 @@ public function registerAndUpdateSubscribedWebhooks()
         $response = $this->curlPatch("v1/notifications/webhooks/$webhook_id", [$parameters]);
     }
 
-    public function webhookVerifyByPostback($parameters)
+    public function webhookVerifyByPostback(array $parameters): bool
```
drbyte: `bool|null` required here (see line 573)

zc_plugins/PayPalRestful/v2.0.0/manifest.php
```diff
@@ -0,0 +1,12 @@
+<?php
+return [
+    'pluginVersion' => 'v2.0.0',
+    'pluginName' => 'PayPal Restful Payment Module',
+    'pluginDescription' => "This Zen Cart payment module combines the processing for the <b>PayPal Payments Pro</b> (<var>paypaldp</var>) and <b>PayPal Express Checkout</b> (<var>paypalwpp</var>) payment modules that are currently built into the Zen Cart distribution.  Instead of using the older NVP (<b>N</b>ame <b>V</b>alue <b>P</b>air) methods to communicate with PayPal, this payment module uses PayPal's now-current <a href=\"https://developer.paypal.com/api/rest/\" target=\"_blank\">REST APIs</a> and combines the two legacy methods into one.<br><br>Based on the like-named unencapsulated v1.3.1 version of the payment module.",
+    'pluginAuthor' => 'Vinos de Frutas Tropicales (lat9)',
+    'pluginId' => 0, // ID from Zen Cart forum
```
drbyte, line 7: We should register a plugin for this, so that it can self-check for updates.

zc_plugins/PayPalRestful/v2.0.0/catalog/includes/modules/payment/paypal/PayPalRestful/Api/PayPalRestfulApi.php
lat9, line 1: Noting that the default visibility for constants is `public` (https://www.php.net/manual/en/language.oop5.constants.php), but it's good to be thorough!

lat9:
> I would also add that most, but not all of the previous config settings were maintained. I have all the MODULE_PAYMENT_PAYPALR_PAYLATER_MESSAGING checkboxes unchecked....plugin had checked all except Product Listings and Search Results.

I'll look at that in the morning.  The plugin install, itself, doesn't mess with the configuration settings, but the payment-module update via the admin's `Modules :: Payment` might be the culprit.

torvista:
>What if there was a setting in the manifest.php that indicates that the plugin, upon install, removes its un-encapsulated version?

I think just adding highlighted text to that effect in the plugin description prior to the install step is enough...if people don't read it or act on it, that's their bad.

PR #7520, opened 2026-03-13
Refactor zen_get_buy_now_qty to handle undefined fields

cstdenis:
Use null coalescing operator for minimum and unit quantities.


drbyte:
I would expect those "undefined" fields to always be present, because they're part of the schema. 
So if they're not present, that probably suggests that the product was not found.
And in that case it would make sense to just abort instead of continuing with any calculations based on fallback values.

Do you know "how" this was getting triggered in such a way that those values were found to be undefined?

PR #7532, opened 2026-03-17
Admin product listing: sort by name when filter by status

### LINKED ISSUE #7531: Admin > Catalog > Categories/Products sort order can be improved for status options
**Describe the bug**
In Admin > Catalog > Categories/Products select a category with many products enabled products and some disable products.

At the top of the screen select "Status (disabled), Name" from the drop down list for 'Categories/Product Display Order:'
Products listed are marked as disabled but the products are not in alphabetic order.

Similarly if "Status (enabled), Name" is selected the status is ordered but the names are not.

**Version**
Zen Cart version: 1.5.8 to 2.2.0
PHP version: 8.2 to 8.4
MySQL version: MySQL 10.6.22- to   MySQL 11.8.5-MariaDB-log
Plugins that may be related:  None
What older/newer versions were also tested? 1.5.8a

**To Reproduce**
In Admin > Catalog > Categories/Products select a category with many products enabled products and some disable products.

At the top of the screen select "Status (disabled), Name" from the drop down list for 'Categories/Product Display Order:'
Products listed are marked as disabled but the products are not in alphabetic order.

Similarly if "Status (enabled), Name" is selected the status is ordered but the names are not.

**Expected behavior**
returned products should be sorted by status and sorted by name

**Screenshots**

**Debug/Error Logs**
None
 
**How to change to expected behaviour
For version 2.2.0
In admin/category_product_listing.php 

at line 866 change
$order_by = " ORDER BY p.products_status ";
to
$order_by = " ORDER BY p.products_status ASC, pd.products_name ASC";

at line 869 change
$order_by = " ORDER BY p.products_status  DESC;
to
$order_by = " ORDER BY p.products_status  DESC, pd.products_name ASC";

drbyte:
Updates #4965 
Fixes #7531

@torvista - Checking with you since you were advocating for this feature. Does this update introduce a complication you'd like a chance to comment on?

### DIFF
--- a/admin/category_product_listing.php
+++ b/admin/category_product_listing.php
@@ -866,7 +866,7 @@ if (!is_writable(DIR_FS_CATALOG_IMAGES)) {
                                     $order_by = " ORDER BY p.products_status, pd.products_name";
                                     break;
                             case (9):
-                                    $order_by = " ORDER BY p.products_status DESC, pd. products_name";
+                                    $order_by = " ORDER BY p.products_status DESC, pd.products_name";
                                     break;
                             case (10):
                                     $order_by = " ORDER BY p.products_id";


admin/category_product_listing.php
```diff
@@ -863,10 +863,10 @@
                                     $order_by = " ORDER BY p.products_model DESC";
                                     break;
                             case (8):
-                                    $order_by = " ORDER BY p.products_status";
+                                    $order_by = " ORDER BY p.products_status, pd.products_name";
                                     break;
                             case (9):
-                                    $order_by = " ORDER BY p.products_status DESC";
+                                    $order_by = " ORDER BY p.products_status DESC, pd. products_name";
```
proseLA: is there an extra space that should not be there between `pd.` and `products_name`?

torvista (commented):
No problem...I already had those missing clauses!

admin/category_product_listing.php
drbyte: Oy! Ironically, I already caught and removed that space in the other line. 

Thanks for the extra set of eyes. 🙌

PR #7543, opened 2026-03-19
Ensure database/file-system cleanup when uninstalled

lat9:
Ensure that the payment-module's `remove` method is run during the encapsulated plugin's un-install if the payment module has been installed via the admin's `Modules :: Payment`.

This is kind of a chicken/egg situation during the plugin's `executeUninstall` method, since the base Zen Cart processing "thinks" that the plugin has already been uninstalled.

### DIFF
--- a/zc_plugins/PayPalRestful/v2.0.0/Installer/ScriptedInstaller.php
+++ b/zc_plugins/PayPalRestful/v2.0.0/Installer/ScriptedInstaller.php
@@ -36,7 +36,7 @@ class ScriptedInstaller extends ScriptedInstallBase
         //
         if (defined('MODULE_PAYMENT_PAYPALR_STATUS')) {
             require_once $this->pluginDir . '/catalog/includes/modules/payment/paypalr.php';
-            $paypalr = new paypalr(true);   //- Let the payment module know it's loading to do an uninstall action
+            $paypalr = new paypalr(uninstalling: true);   //- Let the payment module know it's loading to do an uninstall action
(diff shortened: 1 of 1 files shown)


zc_plugins/PayPalRestful/v2.0.0/Installer/ScriptedInstaller.php
```diff
@@ -29,6 +29,17 @@ protected function executeUpgrade($oldVersion)
...
     protected function executeUninstall()
     {
+        // -----
+        // If the payment module has been "Installed" via Modules :: Payment, need to run its
+        // 'remove' method to (er) remove the configuration settings as well as the root-directory
+        // files.
+        //
+        if (defined('MODULE_PAYMENT_PAYPALR_STATUS')) {
+            require_once $this->pluginDir . '/catalog/includes/modules/payment/paypalr.php';
+            $paypalr = new paypalr(true);   //- Let the payment module know it's loading to do an uninstall action
```
drbyte: For verbosity/clarity, I propose using PHP 8's [named arguments](https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments); that way it's clear what the `true` is referring to, even if someone doesn't read the comment:
```suggestion
            $paypalr = new paypalr(uninstalling: true);   //- Let the payment module know it's loading to do an uninstall action
```

lat9: Done

PR #7557, opened 2026-03-20
Allow some long query strings

zenexpert:
Some payment modules might have a long query string in their response - these need to be allowed, otherwise the customer might experience a 406 after making a successful payment (and order doesn't get created).
I believe I've covered all the required checkout-related pages, but please add any that may have been missed.


includes/application_top.php
```diff
@@ -124,19 +124,30 @@
...
  * reject suspicious non-ASCII characters
  * allows standard printable ASCII but flags common exploit symbols
  */
 if (!empty($_SERVER['QUERY_STRING'])) {
 
-    // check for the specific '¤' (%C2%A4) or characters outside standard range
-    // allow basic printable ASCII but specifically target high-bit "junk"
-    if (preg_match('/[\x00-\x1F\x7F-\xFF]/', $_SERVER['QUERY_STRING'])) {
+    // identify the current page early
+    $current_main_page = isset($_GET['main_page']) ? (string)$_GET['main_page'] : 'index';
```
drbyte: Since this variable is never used to check against `'index'`, let's avoid declaring this temporary variable. (Its name is similar to but not the same as other variables declared later, so let's not add more nor bring confusion.)
It can be safely inlined where used with the in_array() call a couple lines down.

PR #7572, opened 2026-04-03
admin, search_box: use <?=

torvista:
(was looking for a phantom error, so did this instead)

### DIFF
--- a/admin/includes/modules/search_box.php
+++ b/admin/includes/modules/search_box.php
@@ -11,8 +11,8 @@
 // that attribute.
 //
 $autofocus_search = empty($no_searchbox_autofocus) ? ' autofocus="autofocus"' : '';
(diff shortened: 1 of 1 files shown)


admin/includes/modules/search_box.php
```diff
@@ -10,13 +11,13 @@
 // that attribute.
 //
 $autofocus_search = empty($no_searchbox_autofocus) ? ' autofocus="autofocus"' : '';
+echo zen_draw_form('searchForm', basename($PHP_SELF, '.php'), '', 'get', 'class="form-horizontal"', true);
```
proseLA: i take it back.  it does not seem wrong.  i just do not like the change in styling!

drbyte: Agreed.

PR #7581, opened 2026-04-03
Correct `contact_us` name conflict.

lat9:
Corrects the behavior identified in [this](https://www.zen-cart.com/showthread.php/231072-zencart-2-2-1-upgrade-contact-page) Zen Cart forum-posting on the contact_us page.

### DIFF
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -173,7 +173,7 @@ if (!empty($_SERVER['QUERY_STRING'])) {
 
     // count occurrences of each key
     $counts = array_count_values($keys);
-    foreach ($counts as $name => $count) {
+    foreach ($counts as $key => $count) {
         // allow one duplication (possibly accidental), more than 2 is not accidental
         if ($count > 2) {
             $contaminated = true;


drbyte (approved):

proseLA:
do we not run into the possibility of having another var name collision by using `$key` in this context?  i am not sure why the key name is even required here.  would the code not be better suited as:

```php
foreach ($counts as $count) {
```

or am i missing something here?

scottcwilson:
You could always to an `unset` at the bottom of the loop.

proseLA:
> You could always to an `unset` at the bottom of the loop.

i could always do what?

this is core code.  are you telling me i could modify it?  using an observer?  or what?

i have no idea what you are suggesting.

this change has the potential to screw something else up.  for no good reason.

why don't you explain to me why the key is necessary?  

`$key` is used as a var name throughout the code and in various plugins.  here it is being added and not being used.  

why?

drbyte:
Yes: $key is redundant here.

No: There's no need to worry about $key clashing with elsewhere, because $key is constantly recycled, typically in foreach loops like this one.

$name was definitely a clear case of high potential to clash; I'm surprised only one case showed up.

And yes, for the longer-term, dropping $key in this foreach is sensible: it's redundant here.

PR #7599, opened 2026-04-09
Move 'special' language files to `extra_definitions`

### LINKED ISSUE #6507: Catalog language-file loading, no fallbacks for non-english languages?
Originally identified in [this](https://www.zen-cart.com/showthread.php?230159-Upgrading-from-v1-5-5-to-2-0-1-store-pages-come-up-with-error-500-Admin-working) Zen Cart forum posting.

I _think_ that the issue lies on line 98 
https://github.com/zencart/zencart/blob/25da25cd6460d43d922a66d0b726e35ba18f5e59/includes/classes/ResourceLoaders/ArraysLanguageLoader.php#L90-L101

This arises with the processing of the 'extra' main language files' load, where the 'english' fall-back isn't made
https://github.com/zencart/zencart/blob/25da25cd6460d43d922a66d0b726e35ba18f5e59/includes/classes/ResourceLoaders/CatalogArraysLanguageLoader.php#L60-L78

That forum-post identified the issue as coming from an additional constant that was added to `lang.header.php`.

lat9:
Fixes #6507 

Special consideration has been made "forever" in the language-loading for files that are required on every page-load.  These files have, historically, been present in a main language directory, e.g. `languages/english`.

This PR moves those files into the language's `extra_definitions` directory instead.

~~P.S. There was/is an issue that drove this change, but for the life of me I can no longer find it!~~


drbyte:
Note: after mutual discussion, putting this on "pause" for a bit...

PR #7601, opened 2026-04-09
Ensure lang-file name starts with `lang.` for msg

### LINKED ISSUE #7597: V3 Missing language file error still uses old filename in error message
- Delete `includes/languages/english/modules/shipping/lang.flat.php` (on a site that has the default data, with flat enabled).

- Go to the checkout shipping page.

Expected Result:
Warning message with new style filename 
WARNING: Could not locate language file: includes/languages/english/modules/shipping/lang.item.php

Actual Result: 
Warning message with old style filename
WARNING: Could not locate language file: includes/languages/english/modules/shipping/item.php

lat9:
Fixes #7597

### DIFF
--- a/includes/classes/order_total.php
+++ b/includes/classes/order_total.php
@@ -60,8 +60,12 @@ class order_total
                     $language_dir = (IS_ADMIN_FLAG === false) ? DIR_WS_LANGUAGES : (DIR_FS_CATALOG . DIR_WS_LANGUAGES);
                     $lang_file = zen_get_file_directory($language_dir . $_SESSION['language'] . '/modules/order_total/', $value, 'false');
 
-                    if (!str_starts_with($lang_file, 'lang.')) {
-                        $lang_file = 'lang.' . $lang_file;
+                    // -----
+                    // If the language file's name doesn't start with 'lang.' (which they do,
+                    // as of zc300), add that prefix for the cautionary message.
+                    //
+                    if (!str_starts_with($value, 'lang.')) {
+                        $lang_file = str_replace($lang_file, $value, 'lang.' . $value);
                     }
                     if (is_object($messageStack)) {
                         if (IS_ADMIN_FLAG === false) {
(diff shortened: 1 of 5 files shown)


includes/classes/order_total.php
```diff
@@ -60,6 +60,9 @@ public function __construct()
                     $language_dir = (IS_ADMIN_FLAG === false) ? DIR_WS_LANGUAGES : (DIR_FS_CATALOG . DIR_WS_LANGUAGES);
                     $lang_file = zen_get_file_directory($language_dir . $_SESSION['language'] . '/modules/order_total/', $value, 'false');
 
+                    if (!str_starts_with($lang_file, 'lang.')) {
```
scottcwilson: What if we just put these 3 lines in a function and called the function in each place that WARNING_COULD_NOT_LOCATE_LANG_FILE is used?

scottcwilson (changes requested):

includes/classes/order_total.php
lat9: We "could", but it seems overkill.

includes/classes/payment.php
```diff
@@ -106,6 +106,9 @@ public function __construct($module = '')
 
             if (!$languageLoader->loadModuleLanguageFile($next_module['file'], 'payment')) {
                 $lang_file = zen_get_file_directory(DIR_WS_LANGUAGES . $_SESSION['language'] . '/modules/payment/', $next_module['file'], 'false');
+                if (!str_starts_with($lang_file, 'lang.')) {
+                    $lang_file = 'lang.' . $lang_file;
```
drbyte: I feel like this is the wrong "place" for this prefix to be added. 
I think it'll appear as "lang./includes/modules/payment/foo.php" instead of as intended

drbyte (changes requested):

includes/classes/payment.php
lat9: Yep, those initial changes were "not good".

lat9: Those updates just pushed correct this.

lat9:
I don't understand the "one place", please advise.

PR #7638, opened 2026-04-21
SetLocale() no longer accepts int 0 since PHP 8.5. Use string '0' instead

### LINKED ISSUE #7637: admin, header: PHP Fatal error:  Uncaught TypeError: setlocale() (php8.5/strict mode)
php 8.5.5
https://github.com/zencart/zencart/blob/2ccf17e28b0373bb8b4e786e39937abb95863b95/admin/includes/header.php#L75
2nd parameter 0 is an integer, not allowed from 8.5
https://www.php.net/manual/en/function.setlocale.php

maybe should be null

lat9:
Fixes #7637


torvista (approved):
ok for me

PR #7644, opened 2026-04-23
Drop unused PROJECT_VERSION_PATCH constants

drbyte:
Given that semantic versioning in PHP already accommodates suffixes such as `-dev`, `-alpha`, `-beta`, `-rc`, `-pl` (patch level), and we can release a patch version easily via Github, we don't need these constants anymore.

Just note: use `version_compare()` instead of `<` or `>` or `==` for comparing the PROJECT_VERSION_MAJOR+MINOR string returned by `zen_get_zcversion()`.

Plus, we haven't used these `PROJECT_VERSION_PATCH1`, etc constants since around v1.5.5 or so.

### DIFF
--- a/includes/functions/functions_general_shared.php
+++ b/includes/functions/functions_general_shared.php
@@ -4,12 +4,13 @@
  * @copyright Portions Copyright 2003 osCommerce
  * @license https://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: DrByte 2026 Feb 26 Modified in v2.2.1 $
- * @since ZC v1.5.7
  */
 
 /**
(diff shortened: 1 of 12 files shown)


includes/functions/functions_general_shared.php
```diff
@@ -7,7 +7,11 @@
  * @since ZC v1.5.7
  */
 
-function zen_get_zcversion()
+/**
+ * Determine the Zen Cart version currently installed.
+ * @return string in format "major.minor" (eg: "2.0.0", "3.1.0")
+ */
+function zen_get_zcversion(): string
```
lat9, line 15: Did this lose its `@since`?

PR #7650, opened 2026-04-24
Correct fallback for `loadExtraLanguageFiles`

### LINKED ISSUE #7642: loadExtraLanguageFiles doesn't fall-back to 'english' for non-english language files
While investigating an issue with including extra language files in admin, I had thought that if a secondary language file was not present/found, it would load the english version. That is not the case with this function, the file needs to be present in each language folder.

That aside, there is an example given in the docs 

https://docs.zen-cart.com/dev/languages/158_language_files/

```
if (function_exists('zen_get_zcversion') && zen_get_zcversion() >= '1.5.8') { 
   $filename = 'some-custom-file.php';  // Exclude "lang." prefix
   $languageLoader->loadExtraLanguageFiles(DIR_WS_LANGUAGES, $_SESSION['language'],  $filename, '/');  
} else {
   require 'includes/languages/english/some-custom-file.php'; 
}
```
When these parameters are concatenated, the final parameter (=$extraPath) adds a slash to the $language...but it is added again here

```
$mainFile = $rootPath . $language . $extraPath. '/' . $arrayFileName;
$fallbackFile = $rootPath . $language . '/' . $arrayFileName;

```

in the docs it states in other examples the final parameter should end in a slash, but that seems incorrect.

lat9:
Fixes #7642

### DIFF
--- a/includes/classes/ResourceLoaders/ArraysLanguageLoader.php
+++ b/includes/classes/ResourceLoaders/ArraysLanguageLoader.php
@@ -55,6 +55,27 @@ class ArraysLanguageLoader extends BaseLanguageLoader
     }
 
     /**
+     * @since ZC v3.0.0
+     */
+    public function hasLanguageFile(string $rootPath, string $language, string $fileName, string $extraPath = ''): bool
+    {
+        // -----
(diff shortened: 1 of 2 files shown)


torvista (commented):
works for me

PR #7661, opened 2026-04-25
Collapse `config.canonical.php` into `config.core.php`

drbyte:
Reduces maintenance and user confusion about multiple files.

**Risk: is this file ever overridden in practice?**

Other risk: if someone doesn't delete the now-obsolete file, the init file may be called twice. Thus, mitigation has been added to the init file to protect against that duplication.


lat9:
Regarding the 'Other risk', the zen-cart distribution could provide a dummy files whose sole purpose is to overwrite any previous version.

drbyte:
I checked @torvista 's https://github.com/torvista/Zen-Cart_CEON-URI-Mapping repo: They don't directly override this file; however, they do have a comment referencing this now-deleted/retired file as the place to find out what the core load-point number is so that they can load their observer 1 position sooner to be ready for the notifiers to trigger.

drbyte:
> Regarding the 'Other risk', the zen-cart distribution could provide a dummy files whose sole purpose is to overwrite any previous version.

True. But I was hoping to "stop seeing" the file in the directory list 🤣 
Fortunately this isn't a high-risk clash issue.

lat9:
I'm thinking that the "have I already got a canonical link" check at the top of `init_canonical.php` will be sufficient.

torvista:
Don't confuse my repo-to-keep-URI_Mapping-alive with me understanding all that goes on in those files. 
I've only just converted it to use .lang files...now you've forced me to figure out this canonical business, and there are pending issues with load points... I'll need to post in the issues there for second opinions....

Issue #7450, opened 2025-12-21
Store Manager: Reset of Products Viewed table is disabled


torvista:
https://github.com/zencart/zencart/blob/ebe361c2ad2afad3a946a1db806e008382b9e571/admin/store_manager.php#L63-L74

After noticing my table is 200MB+, I see the admin option to clear it is disabled for some reason...??

PR #7662, opened 2026-04-25
Restore product-views reset to Store Manager

drbyte:
Restore product-views reset button to Store Manager.
And remove the `products_viewed` column from `products_description` table. It was retired in v1.5.1, and unused since v1.5.7.

Ref #3481, #7450


torvista (commented):
Works for me.

PR #7663, opened 2026-04-25
Add missing labels for Model, Manufacturer, and Weight in product lis…

mikeel100:
Fixes the "naked" data issue in responsive and grid layouts by adding the existing TABLE_HEADING constants as prefixes.

### DIFF
--- a/includes/modules/product_listing.php
+++ b/includes/modules/product_listing.php
@@ -284,7 +284,7 @@ if ($num_products_count > 0) {
                     $lc_align = 'center';
                     if ($product_listing_layout_style === 'table') $lc_align = '';
                     $lc_text = '';
-                    $lc_text .= $listing_model;
+                    $lc_text .= TABLE_HEADING_MODEL . ': ' . $listing_model;
                     break;
 
                 case 'PRODUCT_LIST_NAME':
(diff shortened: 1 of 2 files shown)


drbyte:
Should these labels be inserted conditionally? ie: `if ($product_listing_layout_style !== 'table')`

mikeel100:
> Should these labels be inserted conditionally? ie: `if ($product_listing_layout_style !== 'table')`

@drbyte

They are already in the condition check, but no labels exist when I tested in my template (my bootstrap style), Classic (tables), or Responsive (div's).

Might be a deeper problem...not sure.

EDIT:

I just re-verified in a clean install as well. 

Model:, Manufacturer:, and Weight: do not display in columnar or row in Classic or Responsive.

In fact, there's  not even a table header row  displaying when using rows in Classic....even though view-source shows there is a table.

lat9:
Yes, a clean install sets these **_Configuration :: Product Listing_** settings all to '0' (don't display)

- PRODUCT_LIST_MANUFACTURER
- PRODUCT_LIST_MODEL
- PRODUCT_LIST_QUANTITY
- PRODUCT_LIST_WEIGHT

While the proposed changes will display "nicely" when the associated data is present, I'll note that there should be checks to see that the data **is** present, otherwise a product could be displayed with

- An empty link to the manufacturer: `<a class="mfgLink list-mfg" href="http://localhost:8085/zc300/index.php?main_page=index&amp;manufacturers_id=0"></a>`
- A **Model:** label with no (er) model
- A weight of 0 when either the weight hasn't been set (i.e. the value is `null`) or the product is 'virtual' (i.e. it has no weight by its definition).

lat9:
You might take a peek at the similar changes I've made to the Bootstrap template: https://github.com/lat9/ZCA-Bootstrap-Template/pull/525

piloujp:
I am not too happy with the hard coded `': '`. Why not use constants in which it is already included like: `TEXT_PRODUCT_MODEL`, `TEXT_MANUFACTURER` and `TEXT_PRODUCT_WEIGHT`?
You can although add ` . TEXT_PRODUCT_WEIGHT_UNIT` after the weight.

lat9:
Thanks for that, @piloujp; I've updated the Bootstrap template's file as suggested (committed directly for the issue): https://github.com/lat9/ZCA-Bootstrap-Template/issues/524

PR #7666, opened 2026-04-25
Add hook for off-site images in zen_image function

mikeel100:
Currently, the Admin zen_image() function is built on the assumption that all image assets are stored locally. It automatically prepends directory paths and performs local file-existence checks. This creates a significant barrier for developers wanting to use off-site storage, CDNs (Amazon S3, Cloudinary), or external image-processing services without resorting to invasive core hacks or expensive post-render buffer regex.

The Solution:

This PR adds a "short-circuit" notifier at the very beginning of the zen_image() function.

Logic:

It allows an observer to intercept the image attributes and provide a fully rendered  HTML string via the $image_html variable.

Result: 

If the observer populates this variable, the function returns the string immediately and exits.

Impact: 

This is a non-breaking change. If no observer is active, the function continues to behave exactly as it does today.

Benefits:

Performance: Eliminates the need for ob_start output buffering to "fix" image paths after the page has rendered.

Extensibility: Provides a native, "Zen-like" way to support modern image hosting architectures.

Stability: 

Prevents core logic from mangling remote URLs (e.g., the images/http... path "sandwich").


drbyte:
I support this. Curious @lat9 's take on it.

mikeel100:
Thanks @drbyte!

lat9:
Besides a couple of code-formatting comments, shouldn't the same notification strategy be used on the storefront's same-named function?

admin/includes/functions/html_output.php
```diff
@@ -142,6 +142,11 @@ function zen_catalog_base_link($connection = '')
  */
 if (!function_exists('zen_image')) {
   function zen_image($src, $alt = '', $width = '', $height = '', $params = '') {
+    // off-site images hook
+    $image_html = '';
+    $GLOBALS['zco_notifier']->notify('NOTIFY_ADMIN_ZEN_IMAGE_OVERRIDE', array($src, $alt, $width, $height, $params), $image_html);
```
lat9: Pass the parameters in an associative array, making it more "obvious" to the observer which parameter is which by simply replacing `array` with `compact`.

admin/includes/functions/html_output.php
```diff
@@ -142,6 +142,11 @@ function zen_catalog_base_link($connection = '')
...
+    if ($image_html !== '') return $image_html;
```
lat9: No single-line conditionals.

lat9 (changes requested):
Also apply to the like-named function in the storefront's `/includes/functions/html_output.php`, using a notifier name that doesn't include `ADMIN`, e.g. `NOTIFY_ZEN_IMAGE_OVERRIDE`.

scottcwilson:
Can we not do this by adding checks around usage of DIR_WS_IMAGES?

mikeel100:
@lat9 ... I hope that is what you are asking for. 2 files changed...both html_output.php files.

mikeel100:
> Can we not do this by adding checks around usage of DIR_WS_IMAGES?

Can you elaborate...not sure I understand your implication.

mikeel100:
@lat9 - I changed the catalog side constant...missed that on my first commit. 

There are 2 Files Changed showing in my branch.

Did I miss pressing any buttons to keep this moving forward? 

Using GitHub is going to take a little getting used to:-)

scottcwilson:
> > Can we not do this by adding checks around usage of DIR_WS_IMAGES?
> 
> Can you elaborate...not sure I understand your implication.

In html_output, in the block that does an `is_file($src)` add a check for not (DIR_WS_IMAGES starts with 'http').

Then in your includes/configure.php file, set `DIR_WS_IMAGES` to your CDN URL.   For example, I did: 

```
define('DIR_WS_IMAGES','https://storage.googleapis.com/zc_images_scottster/images/');
```

scottcwilson:
There would be admin changes needed too but this gives you the idea.

mikeel100:
> There would be admin changes needed too but this gives you the idea. 

@scottcwilson - wouldn't that kill local images by changing configuration?

"Then in your includes/configure.php file, set DIR_WS_IMAGES to your CDN URL. For example, I did:"

My plugin depends on a list of image locations which the user defines via an option in my admin plugin...cdn;drive.google;onedrive.

This change is just the foot in the door.

I've been using off-site images for 15 years by modifying 4 core files and then, most recently, replacing those core changes with the notifier in admin...because that seems to be the way things are going.

It was a real pain to figure it out, but AI looked at my core mods and helped sort out the notifier (my first attempt at using the notifier).

I would much rather PR change all of the core files but I know how highly discouraged that practice is.

My admin notifier basically catches the page before it is displayed and cuts off all the crap added by those 4 files, whereas, my core changes stop the crap before it gets added to the links.

The crap:

- /images/
- http// (missing : from sanitizing)
- IMAGE FILE MISSING

The notifier is a wasteful process, in this case, but it works.

Thanks.

mikeel100:
Files ready...

scottcwilson:
> I would much rather PR change all of the core files but I know how highly discouraged that practice is.

To be clear, this is NOT discouraged.  What's discouraged is the practice of naïvely modifying core files in your local environment (because these changes will be lost in an upgrade unless carefully ported forward.)

If you want to do a group of modifications to support a new feature, you can certainly submit that.  Then CDN-hosted images would work "out of the box."

mikeel100:
Alrighty then...lol. Thanks for the clarification.

If that is the case, you will be seeing a lot from me soon.

- improved search
- product import / export tools
- multiple origin shipping
- display products from current category only (in all boxes)
- improved breadcrumbs
- improved move products (mass by keyword and / or mfg)
- a center and side boxes function that deletes about 27 files
- Search by price side box and header menu
- landing page options for top categories, subcategories, listings, or random
- my Bootstrap template
- a simple catalog colors changer
- artist sidebox
- others...too many to list

I will PR my core files hacks and leave the initial PR as is (unless the team wants it undone).

One at a time until I get the hang of things in GitHub.

scottcwilson:
Great!  But before you put in a core change PR to add a new fuction (rather than fix a bug), I'd suggest opening an issue and letting folks weigh in.   Saves wasted time in case a change would be better as a plugin or has some other issue.

lat9 (approved):
Noting that the observer must return the full HTML `img` tag!

mikeel100:
@scottcwilson 
@lat9 
@drbyte 

I spent the last 2 days getting files ready to change in this PR.

- stylesheet.css
- category_product_listing.php
- init_sanitize.php
- html_output.php
- general.php
- categories.php

In the process, I hit another "BUG" related to images in admin.

The show / hide button does not work on category listings on my clean install tests. There is a pathing issue in the files. I did another test clean install to verify the issue.

My working folder is sdcard/htdocs/zc222_test on Termux, Debian,  Android 16, S22 Galaxy phone.

I have all the files ready which enables off-site images....fully tested and working.

I just need to know whether to proceed with this PR or open a new one as suggested by Scott.

Thanks.

scottcwilson:
> The show / hide button does not work on category listings on my clean install tests.

Seems ok to me.  I can't reproduce your issue.

lat9:
> > The show / hide button does not work on category listings on my clean install tests.
> 
> Seems ok to me. I can't reproduce your issue.

... and even if we could (I'll try in the morning), that's a separate issue!

lat9:
> @scottcwilson @lat9 @drbyte
> 
> I spent the last 2 days getting files ready to change in this PR.
> 
>     * stylesheet.css
> 
>     * category_product_listing.php
> 
>     * init_sanitize.php
> 
>     * html_output.php
> 
>     * general.php
> 
>     * categories.php
> 
> 
> In the process, I hit another "BUG" related to images in admin.
> 
> The show / hide button does not work on category listings on my clean install tests. There is a pathing issue in the files. I did another test clean install to verify the issue.
> 
> My working folder is sdcard/htdocs/zc222_test on Termux, Debian, Android 16, S22 Galaxy phone.
> 
> I have all the files ready which enables off-site images....fully tested and working.
> 
> I just need to know whether to proceed with this PR or open a new one as suggested by Scott.

As Scott indicated, that should be at least one additional PR.  If a change covers a different admin tool, it's best to submit one PR for each tool ... unless there's a common issue that ties them together.

scottcwilson:
Mike, will you build a plugin to use these new notifiers to enable off-site images?

PR #7670, opened 2026-04-27
Use icon as language toggle when 2 languages present

### LINKED ISSUE #7669: Admin, header.php: language drop down
Currently,
if there is only one language: $language_array is zero = no dropdown
If there is more than one language: $language_array is 2 or more = dropdown

How about if there are only two languages:  no dropdown, just a link to the alternate language (as the country flag, for example.)

I was doing this, thinking it was a 2 minute job but it got increasingly messy, hence as a suggestion.

In fact it seems $languages_array could be done away with, and $languages used directly....instead of this, which does not seem a good idea to mix contents of id.

Image

lat9:
Fixes #7669

Keeps the current "no icon" when a single language is installed and the language dropdown when more than 2 languages are installed

Provides a quick language-toggle (via the languages' image) when exactly 2 languages are installed.

### DIFF
--- a/admin/includes/header.php
+++ b/admin/includes/header.php
@@ -41,17 +41,17 @@ if (empty($action) && count($languages) > 1) {
     $languages_selected = $_SESSION['language'];
     $missing_languages = '';
     $count = 0;
-    for ($i = 0, $n = count($languages); $i < $n; $i++) {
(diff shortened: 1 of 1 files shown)


torvista:
Works for me! Thanks.

PR #7684, opened 2026-05-01
Update install SQL: int/float/decimal no characters

lat9:
As the various MySQL implementations become more strict, defaulting int/float/decimal fields to a character value could now result in a MySQL error on an initial install.

This PR changes the default values for all numeric-type fields to a numeric (instead of character) value.


drbyte:
Research validation, note to self:
> - `DEFAULT 0`: The default value for the column is 0, which the database implicitly treats as 0.0000 due to the decimal scale defined.

👍

Issue #7679, opened 2026-04-30
Multiple problems around State field display in address forms


piloujp:
This applies to ZC v3.0.0, v2.2.2 and probably older ones and is visible in create account form and user address edit forms.
Similar/related to issues #7671 and #7143 (with PR #7144 ).
There are in fact two state fields, one input text type for when country has no zones or pull down is disabled and one pull down type for countries with zones. One is supposed to be hidden and the other displayed depending on the country.
Problems encounter with this field concern display of label, misleading required indicator and field positioning:

1 - Label used does not always correspond to the right field.
When pull down menus are used the label for text field (`stateLabel`) is always hidden (and its value always empty) and the one for pull down is always visible even when selected country has no zone, leading to mismatch labels and failing of JQuery to handle them.

2 - Required symbol display for state pull down menu.
The required symbol for state field pull down menu is displayed as a separate element and manage by JQuery. If required, it should be displayed when the field is empty only, but it is always there. There are more details in #7671 .

3 - Required symbol display for state text field.
The required symbol for state text field is displayed as a placeholder in the filed itself when it is required. When changing from a country with zones to one without zones in create account page, the symbol is not visible because a non-breakable space is in the field by default, hiding it.

4 - Space and alignment of state pull down field.
4-1 When pull down menu is used, it is shifted on the right due to a hard coded non-breakable space in default template files.
4-2 To avoid it to stick with next field, a `BR` tag is added after it but the JQuery management script fails to display it properly in some situations.

5 - State not required (`ENTRY_STATE_MIN_LENGTH=0`).
When the state field is not required by setting `ENTRY_STATE_MIN_LENGTH` in admin to 0, the required symbol displays anyway. This setting is simply ignored in all core templates and Bootstrap template too.

These issues all interacts with each other, making them confusing and hard to spot.

1, 2, 4-2 all originated from the JQuery script in `includes/templates/template_default/jscript/zen_addr_pulldowns.php` file. 
3 is coming from file `includes/modules/create_account.php` on line 322.
4-1 and 5 are caused by templates files `includes/templates/template_default/templates/tpl_modules_address_book_details.php` and `includes/templates/template_default/templates/tpl_modules_create_account.php`

piloujp:
[...]

PR #7691, opened 2026-05-03
Required symbol in state input text field hidden

piloujp:
Removes a non-breakable space added to state input text field on Create Account page when its value is an empty string.
See issue #7679 problem 3 for details.
Complementary to PR #7681 to fix display of required symbol for state field.


includes/functions/html_output.php
```diff
@@ -598,7 +598,7 @@ function zen_draw_input_field($name, $value = '', $parameters = '', $type = 'tex
     }
 
     $field = '<input type="' . zen_output_string($type) . '" name="' . zen_sanitize_string(zen_output_string($name)) . '"';
-    if (isset($GLOBALS[$name]) && is_string($GLOBALS[$name]) && $reinsert_value == true) {
+    if (!empty($GLOBALS[$name]) && is_string($GLOBALS[$name]) && $reinsert_value == true) {
```
lat9: Why this change?  That'll preclude adding a `value="0"` attribute to the input field if the value's 0 and I'm remembering some kind of issue when that zero-value isn't set.

lat9 (commented):
See comment

includes/functions/html_output.php
piloujp: I did this to eliminate empty strings. You would be right for numeric value, we do not want to eliminate 0, but this particular test is for strings (`&& is_string`) anyway.
When Create Account page is loaded this value is always empty. It is a kind of test to repopulate the field if a value was already entered which I think never appends in this particular page. It is certainly used somewhere else, but why would you repopulate the field with an empty string? You won't see a difference in the field but in the HTLM code you get a property like this: `value=''`
Actually when I inspected the page in Firefox, there was just `value` written.

lat9: If `$GLOBALS['name']` is currently set to '0' (a string value of 0), that will result in no `value="0"` being set for the input field, since `!empty('0')` is `false`.

piloujp: Aah... My bad, I don't know why I had the idea that `!empty('0')` would be true...
Adding a condition like ~~&& $GLOBALS['name'] !== ''~~ `&& $GLOBALS[$name] !== ''` (empty string) should be better.

includes/functions/html_output.php
```diff
@@ -598,7 +598,7 @@ function zen_draw_input_field($name, $value = '', $parameters = '', $type = 'tex
     }
 
     $field = '<input type="' . zen_output_string($type) . '" name="' . zen_sanitize_string(zen_output_string($name)) . '"';
-    if (isset($GLOBALS[$name]) && is_string($GLOBALS[$name]) && $reinsert_value == true) {
+    if (isset($GLOBALS[$name]) && is_string($GLOBALS[$name]) && $GLOBALS[$name] !== '' && $reinsert_value == true) {
```
drbyte: How "needed" is this check for "blank"?
I'm concerned about unintended side-effects.

piloujp: Actually, it is not needed at all, at first I thought it was but if the field is blank then the required symbol will display in the field. Logically, it would be better not repopulate the field with a blank string, but concerning the display of rquired symbol it has no influence.
If you prefer, this modification can be removed.

piloujp: The only difference is that without this test you get this in the HTML code of the input field: `value=""`

piloujp: Somehow I thought the `&nbsp;` was added to avoid the `value=""`, but I might be wrong.

drbyte: Yes, let's remove it. The value="" doesn't bother me.  It's not invalid. 
And it avoids surprises.

piloujp: Done

PR #7697, opened 2026-05-05
Admin "Product Types", ensure sidebox displayed on initial entry

### LINKED ISSUE #7692: Catalog :: Product Types, Layout Settings doesn't display sidebox on 1st entry
Seeing the following, where the Show Model Number actions should be displayed

Image

It's been like this since at least zc222

lat9:
Fixes #7692 

Also reformats and refactors the script.

### DIFF
--- a/admin/product_types.php
+++ b/admin/product_types.php
@@ -10,16 +10,13 @@ require 'includes/application_top.php';
 $action = $_GET['action'] ?? '';
 $cID = (int)($_GET['cID'] ?? 0);
 
-$pagenum = (int)($_GET['page'] : 1);
-if ($pagenum < 1) {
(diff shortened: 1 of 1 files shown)


drbyte:
When I search for `action=` I don't see `'new'` mentioned anywhere in this file. There's `'insert'` on one of the buttons.
In the cID change, you also changed the substr(0,3) !== 'new' to $action != 'new' ... but should it actually be `insert` in the first place?
I'm okay to merge as-is because this action=new is a no-op (unrelated to the present fix). But it's an interesting long-standing artifact.

lat9:
I know, there hasn't been a 'new' action form for many, many releases.  I just got tired of correcting issues not of my making.

PR #7701, opened 2026-05-06
Taking in account the state field setting 'not required'

piloujp:
When state field is not required (minimum length set to 0), no required symbol should be displayed.
Fixes issue #7679 part 5.


piloujp:
Update to keep the `<span class="alert">` tag even if empty, so that the new JQuery script do not try to show/hide another similar tag if not present.

piloujp:
Removed the required attribute added to input type state field as it will prevent the form to be submitted when the field is hidden.

PR #7705, opened 2026-05-06
Featured Categories page does not show any featured categories

TMCSherpa:
Fixed featured categories never showing active featured categories
Don't ask me why ->EOF and empty() don't work
The header file is part of changing listing to $fc_listing because I thought the problem might be there

### DIFF
--- a/includes/templates/template_default/templates/tpl_featured_categories_default.php
+++ b/includes/templates/template_default/templates/tpl_featured_categories_default.php
@@ -6,7 +6,6 @@
 * @copyright Portions Copyright 2003 osCommerce
 * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
 * @version $Id: lat9 2025 May 13 Modified in v2.2.0 $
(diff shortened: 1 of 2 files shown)


includes/templates/template_default/templates/tpl_featured_categories_default.php
```diff
@@ -18,15 +18,15 @@
 $row = 0;
 $col = 0;
 
-if ($listing->EOF) {
+if (count($fc_listing) === 0) {
```
lat9: The only change needed is on this line, change to `if (count($listing) === 0) {`

lat9 (changes requested):
Only the `tpl_featured_categories.php` requires change, see inline comment.

includes/templates/template_default/templates/tpl_featured_categories_default.php
TMCSherpa: Yup I put $listing back.

PR #7707, opened 2026-05-09
Tab and Space Characters Removing

piloujp:
Removed all tab and space characters at end of lines or in empty lines for all files with extension `.css`, `.html`, `.js`, `.php`, `.txt`.
Although replaced other tab characters by four spaces.


lat9:
Wow!  That's a lot of files to review; I'm sure @drbyte will ask Copilot to do its thing.

Before that happens, you'll need to remove the changes to the PayPalRestful encapsulated payment module.  This is not the "source of truth" for that code; see, instead, https://github.com/zencart/paypalr.

piloujp:
Yes, there are a lot, But it took me less time than I thought. It should be a perfect job for Copilot.
Although I could have done that entirely automatically, I did it in a semi-automatic way to avoid mistakes. I use a PowerShell script to identify these files then I apply a regex replace on each of them, visually checking.

PayPalRestful files have been removed.

torvista:
Wow, in times past I would have been pilloried for such a thing..."what is the business case for this", "makes for upgrade hell", "multiple files to compare/merge for no good reason"...etc.

My repeated opinion is that any altruistic input should be encouraged and accepted gracefully while it offers any incremental advancement however inconsequential. 
While there is no roadmap for guidance/to cherry-pick from, people will keep spending their precious time "fixing" whatever takes their fancy, so no complaining/constructive criticism only please.

piloujp:
I must say I hesitated myself before doing that, but I have the hope it will save time on future commits. When a file code is modified, you are supposed to do separate commits for cosmetic stuffs like that. Which would make 186 commits in this case... When you think of it, it is kind of a 'bigger waste' of time doing it one by one. Generally you don't do it and keep the mess in the file for years...

drbyte:
I'm okay with some/several of these tidying's.

But I'm going to say no to the following. Please remove these files from this PR, both admin and catalog:
- security patch files
- music product-type files (includes media_manager too)
- passthru proxy files like /admin/includes/classes/currencies.php and order.php
- vendor-provided files like jAlert.css, spiffycal JS, flot JS (and their calendar.css/calendarcode.js etc files), jQuery.MMenu and its files.
- backward-compatibility files like v156_compatibility.css
- email_use_8bit.php (both versions)
- extra_configure/index.html
- WHERE_DID_ALL_THE_FILES_GO.txt and the read_me.txt files for template jscript/onload. These are one-liners giving brief instruction.
- zca files
- class.notifier.php
- classes/upload.php (is being overhauled in another PR, so will cause an unnecessary merge-conflict)
- all PayPal and IPN related files, including mcs_learn_more.html and vbv_learn_more.html
- init_sefu.php, init_tlds.php
- /page_not_found.php

**REASONS:**
Many of these are pending deprecation/removal, or would introduce confusion to see them "updated" anytime "recently". Some are unlikely to ever require touching, so don't want to touch them now either.
Others are one-liners with reference instructions, or passthru/broker to another file.
Others are vendor-provided, or minified, and should be retained as-original.

drbyte:
Another consideration: the convention we used with stylesheets (both catalog template and admin CSS files) was to use tabs to keep the files smaller, so they load faster.
This PR is changing those tabs to spaces, increasing those filesizes notably.

piloujp:
I reverted files modifications as you requested, well as close as I could.

PR #7711, opened 2026-05-10
`order_status` page, force order-id to int

zcwilt:


includes/templates/template_default/templates/tpl_order_status_default.php
```diff
@@ -22,7 +22,7 @@
 if (isset($order)) {
 ?>
     <fieldset>
-        <h2 id="orderHistoryDetailedOrder"><?= SUB_HEADING_TITLE . ORDER_HEADING_DIVIDER . sprintf(HEADING_ORDER_NUMBER, $_POST['order_id']) ?></h2>
+        <h2 id="orderHistoryDetailedOrder"><?= SUB_HEADING_TITLE . ORDER_HEADING_DIVIDER . sprintf(HEADING_ORDER_NUMBER, $orderID) ?></h2>
```
lat9: Noting that if we just cast `$_POST['order_id']` to an int in the header_php.php for the page, that would correct this for **all** templates.

PR #7720, opened 2026-05-14
Feature/encapsulated template plugin plan

zcwilt:
This is something that @lat9 and I have been working on. 

Currently encapsulated plugins allow you to add some template files that "overlay" over your current catalog template. 

This can be full pages, extra css/javascript etc.

However there was no way to allow for a full template to be used in an encapsulated plugin 

This PR allows you to do that, so going forward template designers can upload their templates as a plugin allowing for a much easier install experience

When creating a plugin template you still have access to all of the catalog loading that previously existed so you can load extra includes, observers or even admin pages to support the template.

This PR includes a Responsive Classic Plugin template that when enabled in the plugin manager can then be selected in the admin template select

It's basically a copy of our current Responsive Classic template with a header change that shows you are using this template. 

Encapsulated templates also allow for a parent/child relationship. The Responsive Classic Plugin sets the parent to template_default, however this could be any other template 

## Summary

This PR adds support for encapsulated template plugins and includes a dogfood `ResponsiveClassicPlugin` plugin to exercise the new loading flow end to end.

The goal is to let a plugin carry its own template assets, language files, page/template overrides, and related resources without requiring those files to be copied into the core template directories.

## What changed

- Added a `TemplateResolver` and `TemplateDto` to centralize template path resolution.
- Updated catalog and admin template initialization to use the new resolution flow.
- Extended page, sidebox, language, file, and HTML asset loading so plugin-provided template resources can participate in the normal fallback chain.
- Added `TemplateSelect` support and related admin updates for working with plugin-backed templates.
- Added the `ResponsiveClassicPlugin` dogfood plugin under `zc_plugins/` as a concrete implementation/example.
- Added unit/feature coverage around template resolution, fallback behavior, and plugin resource loading.

## Why

Today, template customizations are heavily tied to the core `includes/templates/...` structure. That makes templates harder to package, version, test, and distribute as plugins.

This change introduces a more explicit template-resolution layer so templates can be encapsulated inside plugins while preserving existing fallback behavior and minimizing changes to page/module consumers.

## Testing

- Added/updated tests for:
  - `TemplateResolver`
  - page loader template resolution
  - sidebox template resolution
  - functions/templates fallback behavior
  - functions/files fallback behavior
  - HTML output asset fallback behavior
  - language loader template chains

## Notes

- This PR is intentionally broad because the template lookup behavior is cross-cutting and needs to stay consistent across multiple loaders.
- The included `ResponsiveClassicPlugin` plugin is primarily for exercising and validating the new plugin-based template model and not meant as a genuine template.


admin/template_select.php
```diff
@@ -13,26 +17,59 @@
 $action = $_GET['action'] ?? '';
 $template_info = zen_get_catalog_template_directories();
 
+$installedPluginKeys = [];
+foreach ((new PluginControlRepository($db))->getAll() as $plugin) {
+    if (($plugin['status'] ?? PluginStatus::NOT_INSTALLED) !== PluginStatus::NOT_INSTALLED) {
```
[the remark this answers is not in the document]
zcwilt: Note: bit of a red herring as template plugins can not be set to DISABLED in plugin manager now. 
`includes/classes/ViewBuilders/PluginManagerController.php`

drbyte:
I "think" I resolved that merge-conflict correctly, but please review.

includes/functions/functions_files.php
```diff
@@ -146,7 +146,7 @@ function zen_get_file_directory(string $check_directory, string $check_file, boo
     // -----
     // The returned value includes both the file's directory and filename.
     //
-    $dir_filename = $htmlIncludesFinder->find($zv_filename);
+    $dir_filename = $htmlIncludesFinder->find($zv_filename, false);
```
lat9, line 149: Why this change? Doesn't that mean that the return statement will need a change as well, since the `$dir_filename` is now only going to be a `$dir`?

zcwilt: The return is still OK.
if $dir_filename is false then we fallback to the directory and the $zv_filename found before the html_includes dir is checked and the return will use this if  ($dir_only === false)

PR #7721, opened 2026-05-14
Moving towards class-based access of config settings

lat9:
In a future release of Zen Cart, we will do away with the 'conversion' of configuration settings to PHP `define` values and, instead, use a class-based method to access these values.  The helper function `zen_get_config_value` provides an easy way to perform this 'conversion'.

This is a preliminary step to support template-specific settings.  I've made the required modifications to a couple of the `/includes/classes` files, but wanted to get this change "out there" for both review and to gather help in doing the conversion on the rest of the files!

We'll need to come up with a way to indicate that you're working on a file (or collection of files) so that effort isn't duplicated in this time-consuming process.

### DIFF
--- a/includes/classes/Coupon.php
+++ b/includes/classes/Coupon.php
@@ -60,7 +60,7 @@ class Coupon extends base
         $sql = "SELECT coupon_id
                 FROM " . TABLE_COUPONS . "
                 WHERE coupon_code LIKE ':original_code:%'
-                AND coupon_id = " . (int)zen_get_config_value('NEW_SIGNUP_DISCOUNT_COUPON');
+                AND coupon_id = " . (int)zen_config('NEW_SIGNUP_DISCOUNT_COUPON');
         $sql = $db->bindVars($sql, ':original_code:', $origin_prefix, 'noquotestring');
         $delete_duplicate_coupons_check = $db->Execute($sql);
         if ($delete_duplicate_coupons_check->RecordCount() > 0) {
@@ -72,7 +72,7 @@ class Coupon extends base
                 FROM " . TABLE_COUPONS . "
                 WHERE coupon_code LIKE ':original_code:%'
                 AND coupon_active = 'Y'
-                AND coupon_id !=  " . (int)zen_get_config_value('NEW_SIGNUP_DISCOUNT_COUPON') . "
+                AND coupon_id !=  " . (int)zen_config('NEW_SIGNUP_DISCOUNT_COUPON') . "
                 AND coupon_type != 'G'";
         $sql = $db->bindVars($sql, ':original_code:', $origin_prefix, 'noquotestring');
         $delete_duplicate_coupons = $db->Execute($sql);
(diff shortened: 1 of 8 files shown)


TMCSherpa:
So 
CONFIGURATION_KEY == X 
needs to become
zen_get_config_value('CONFIGURATION_KEY') === X
everywhere.
Can I get back to you on that?

torvista:
>We'll need to come up with a way to indicate that you're working on a file (or collection of files) so that effort isn't duplicated in this time-consuming process.

Could we do a PR marked "WIP" (Work In Progess) as a placeholder/marker when work starts on a PR?

lat9:
> > We'll need to come up with a way to indicate that you're working on a file (or collection of files) so that effort isn't duplicated in this time-consuming process.
> 
> Could we do a PR marked "WIP" (Work In Progess) as a placeholder/marker when work starts on a PR?

Sure, and just list the file (or files) that you're planning on tackling either within the subject (if it fits) or within the description otherwise.

drbyte:
I'd kinda like a shorter function name. 
Given it'll be used "a lot", I don't mind it being super-terse and less descriptive, like `zen_config()`.

In theory we could also have `zen_check_config($key, $value_to_check, $op)` where it just gives back a `bool` of whether it matches or not, and even checks based on expected `type` for that $key. But that can be done as a future iteration, probably housed within the configRepository.

lat9:
I'll replace the function name with `zen_config` to reduce code-bloat!

TMCSherpa:
I said I'd get back to you.  If you grab as many configuration_keys as you can find (760), run a script to locate the files they're in and sed replace them with zen_config('') you end up with 461 modified files that I need to check == != etc on.
Basically don't ask anyone to work on it I'm partially/mostly done.

lat9:
> I said I'd get back to you. If you grab as many configuration_keys as you can find (760), run a script to locate the files they're in and sed replace them with zen_config('') you end up with 461 modified files that I need to check == != etc on. Basically don't ask anyone to work on it I'm partially/mostly done.

Sweet!  Note that I've done the files in /includes/modules/shipping (they needed a bit more than the scripting, which is quite impressive).

TMCSherpa:
I'm not saying it'll be done tomorrow or anything there's around 3000 lines to double check :)
Does everything with defined need to be rejiggered?

lat9:
No problem, I appreciate your help.  Only database configuration settings should be included in the `zen_config` update, not language constants.

TMCSherpa:
I mean things like
includes/modules/shipping/freeoptions.php
$this->sort_order = defined('MODULE_SHIPPING_FREEOPTIONS_SORT_ORDER') ? MODULE_SHIPPING_FREEOPTIONS_SORT_ORDER : null;
$this->sort_order = zen_config('MODULE_SHIPPING_FREEOPTIONS_SORT_ORDER');

lat9:
I'm not sure what the question is.  That change to the various shipping modules takes into account that an undefined constant will return `null` from `zen_config`, thus somewhat simplifying the "am I installed" check.

PR #7727, opened 2026-05-17
Shipping modules: Using `zen_config` and ...

lat9:
- removing `false` return from constructor; constructor's shouldn't be returning values.
- Ensure that an enabled module's sort-order is cast to an int.
- Simplifies `check` methods, noting that users of this method only care about 0 and non-0.
- Simplification to `zones` switches.


piloujp:
I hope I am wrong, but there might be a problem. What happens when the constant corresponding to a configuration key is not defined, but this key exists in configuration table?
My understanding is that `zen_config` will return an array containing the configuration id, key and value, where actually the value only is expected.

lat9:
Yes, I saw that as well and wasn't happy with it.  I'll get the function updated so that it returns `string|int|null` (the MIN/MAX values settings are coerced into ints).

lat9:
Holy crow! No comments ... that's a first!

PR #7729, opened 2026-05-18
zen_jscript_form_check: add identification title, strict javascript, use <?=, consistant formatting/reduce php wrapping

torvista:
add identification title, strict code, use <?=, formatting

### DIFF
--- a/includes/templates/template_default/jscript/zen_jscript_form_check.php
+++ b/includes/templates/template_default/jscript/zen_jscript_form_check.php
@@ -9,14 +9,16 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: Modified in v3.0 $
  */
-
 ?>
-<script>
+<script title="zen_jscript_form_check (template_default)">
+    "use strict";
     let selected; // retained for compatibility if referenced elsewhere
 
     function check_form_optional(form_name) {
         const f = form_name;
-        if (!f.elements['firstname']) return true;
+        if (!f.elements['firstname']) {
+            return true;
+        }
 
         const firstname = f.elements['firstname'].value;
         const lastname = f.elements['lastname'].value;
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks for the reformat of the php echoes.
In the ID you added `(template default)` but if someone clones this file into their own template there's a very strong likelihood that they'll probably not pay any attention to that detail and it'll still say template default in their override.

drbyte:
I'd merge it as-is, but I don't understand why you're always (I say "always" as a general broad statement about many PRs, not as an attack statement) collapsing PHP wrappers that make the actual PHP code look like "close this loop, start a new loop", instead of "here's a loop, here's the start, now here's the end. Period. Next loop will start its own PHP wrapper".

ie: you're changing it all  from 
```
<?php if (foo) { ?>
   <html>
<?php } ?>
```
to
```
<?php }

if (foo) { ?>
   <html>
```
While it's technically syntactically fine, it's less readable. AND more importantly if someone wants to delete/comment-out/augment that loop, the've got to re-wrap in two places and triple-check whether they matched things up right.
The whole reason we went with that "usually wrap a new condition on its own line", especially in templates, was so people could find and delete/relocate things easier.

So ... I'm trying to understand: is this new format something your IDE is pushing? Or is it something you need for readability from another perspective?

torvista:
1) title: 
Just to repeat a previous comment that with ZC having so many manners of including javascript, I find it very helpful to use the titles to identify the source location.
Ideally it would be dynamically generated to indicate from whece it comes, but I thought that would get convoluted...
Initally I used $template_dir but that uses the current template rather than the code location. 
If it gets copied elsewhere it means it is going to be modified, and you can't do all the thinking for people.

2) loop wrapping, I thought it is more readable now, with many php tags removed and the actual html exposed.
I can change that to make each check self-contained if you prefer.

drbyte:
@torvista in looking more at this, trying to decide on how strong an opinion I want to hold,
 in this case these loops are quite straightforward, so let's leave it as-is. No need for extra work.

drbyte:
Thanks again for the cleanup. I had JS on my mind and didn't even notice the echo's, etc. 🤣

PR #7737, opened 2026-05-20
Add zc_cli version of currency_cron: `currency-rates:update`

drbyte:


zcwilt:
Note the test failures are an issue with the test framework itself and not your code. 

The test runner is hard coded !! to only check that the original console commands exist/

```
        foreach (['ConfigGetCommand.php', 'HelpCommand.php', 'ListCommand.php', 'PluginListCommand.php', 'VersionShowCommand.php'] as $file) {
        
        ...
        
        }
```
so adding your command in that list should fix the tests 

although I should really update that test to make it less brittle.

drbyte:
TODO's:
probably should load the Admin extra_functions dir.
probably should explore whether something in zc_plugins registers a function for currency rate handling.

drbyte:
Agreed. As far as currency-rate features, I was thinking of refactoring to a simplified registry that plugins could register into.

zcwilt:
The new command currently loads `includes/functions/functions_exchange_rates.php`, but it does not load the legacy admin extra-function surface that `admin/currency_cron.php` relied on. That is a behavior regression for any store or plugin that provides custom `quote_*_currency()` functions via `admin/includes/functions/extra_functions/*.php` or plugin `admin/includes/functions/extra_functions/*.php`.

I would avoid putting that into the base CLI bootstrap. A better boundary is a small shared CLI helper that loads trusted admin extra functions on demand, and then have `CurrencyRatesUpdateCommand` opt into it.

Example shape:

```php
<?php

declare(strict_types=1);

namespace Zencart\Console;

class LegacyAdminFunctionLoader
{
    /**
     * @param array<string, string> $trustedPlugins unique_key => version
     */
    public function loadExtraFunctions(array $trustedPlugins): void
    {
        $this->requireGlob(DIR_FS_ADMIN . 'includes/functions/extra_functions/*.php');

        foreach ($trustedPlugins as $uniqueKey => $version) {
            $this->requireGlob(
                DIR_FS_CATALOG . 'zc_plugins/' . $uniqueKey . '/' . $version . '/admin/includes/functions/extra_functions/*.php'
            );
        }
    }

    private function requireGlob(string $pattern): void
    {
        foreach (glob($pattern) ?: [] as $file) {
            require_once $file;
        }
    }
}
```

Then in the command:

```php
require_once DIR_FS_CATALOG . 'includes/functions/functions_exchange_rates.php';

$trustedPlugins = [];
if ($this->trustedPluginResolver !== null) {
    $trustedPlugins = ($this->trustedPluginResolver)();
}

($this->legacyAdminFunctionLoader ?? new LegacyAdminFunctionLoader())
    ->loadExtraFunctions($trustedPlugins);
```

That keeps the base CLI runtime narrow, but preserves the provider-loading behavior that the legacy cron path already supported.

drbyte:
That feels sensible.

zcwilt:
Added coverage for the two CLI currency-update review points in `64bf93e50` (`Add CLI currency update regression tests`) on my branch.

New tests:
- `CurrencyRatesUpdateCommandTest::testCurrencyRatesUpdateLoadsLegacyAdminExtraFunctionProviders()`
- `CurrencyRatesUpdateCommandTest::testCurrencyRatesUpdateAllowsBackupProviderWhenPrimaryFunctionIsMissing()`

Current status against this PR head:
- `ConsoleKernelTest.php` stays green
- the two new currency-update tests fail as expected against the current implementation, because the command still:
  - does not load the legacy admin/plugin `extra_functions` provider surface that `admin/currency_cron.php` relied on
  - returns early when the primary provider function is missing instead of preserving the existing primary-to-backup fallback behavior in `zen_update_currencies()`

If useful, I can follow up with a small implementation patch shaped around a shared CLI helper for trusted admin extra-function loading and then let `zen_update_currencies()` retain the fallback policy.

zcwilt:
Follow-up on the currency-provider review thread:

Implemented in the latest branch state.

Current shape:
- `CurrencyRatesUpdateCommand` now loads trusted legacy admin/provider quote functions through `LegacyAdminFunctionLoader`
- the loader is intentionally narrow and only includes files that actually define `quote_*_currency()` functions, so it avoids unrelated admin `extra_functions` side effects
- the earlier primary-only precheck was removed, so `zen_update_currencies()` retains the legacy primary-to-backup fallback behavior
- added focused unit coverage, including an encapsulated test plugin fixture that provides currency quote functions through plugin `admin/includes/functions/extra_functions`

Current PR head `c69098b23` is green in Actions:
- `Zen Cart Unit Tests`
- `Zen Cart Admin Feature Tests`
- `Zen Cart Store Feature Tests`

Targeted local suites also pass.

drbyte:
A couple common modules are:
https://www.zen-cart.com/downloads.php?do=file&id=1505
https://www.zen-cart.com/downloads.php?do=file&id=1506

zcwilt:
@codex review

zcwilt:
The review has discovered an issue with namespace loading that requires a change to core cli code. 
Will address that elsewhere.

zcwilt:
@codex review

Issue #7728, opened 2026-05-17
Remaining conversions for `zen_config` have been claimed


lat9:
The `includes/modules/shipping` modules (PR #7727) and @TMCSherpa has created a script to do the remainder (see after-merge comment in PR #7721).

lat9:
Noting that I've got `includes/templates/template_default/templates/tpl_site_map_default.php` covered for #6559

lat9:
I've also got `includes/init_includes/init_sanitize.php` covered for #7636

lat9:
Additional updates: https://github.com/zencart/zencart/commit/724c3f77725fcdc2ce9822140c295304dc5c77c6

TMCSherpa:
This is a list of all the CONFIGURATION_KEY's I was processing.
If there isn't a ./zencart after the KEY then it's been dealt with.
If there IS a ./zencart something is still in that file. Usually it's a comment or part of a SQL statement. I got lazy so rather than set up another exception for DIR_FS_SQL_CACHE I removed it from the list I was using so that's why it's here now. Anyway there's a <- tag on all the files explaining why they're here

ACCOUNT_COMPANY
ACCOUNT_DOB
ACCOUNT_EMAIL_PREFERENCE
ACCOUNT_FAX_NUMBER
ACCOUNT_GENDER
./zencart-mod/includes/modules/create_account_send_email.php <- comment
ACCOUNT_NEWSLETTER_STATUS
ACCOUNT_STATE
ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN
ACCOUNT_SUBURB
ADDITIONAL_IMAGES_HANDLING
ADDITIONAL_IMAGES_MODE
ADMIN_CONFIGURATION_KEY_ON
ADMIN_EXTRA_EMAIL_FORMAT
ADMIN_NAME_MINIMUM_LENGTH
ADVANCED_SEARCH_DEFAULT_OPERATOR
ADVANCED_SEARCH_INCLUDE_METATAGS
ALLOW_GUEST_TO_TELL_A_FRIEND
ATTRIBUTES_ENABLED_IMAGES
ATTRIBUTES_ENABLED_PRICE_FACTOR
ATTRIBUTES_ENABLED_QTY_PRICES
ATTRIBUTES_ENABLED_TEXT_PRICES
AUDIENCE_SELECT_DISPLAY_COUNTS
BEST_SELLERS_FILLER
BEST_SELLERS_TRUNCATE
[...]

PR #7744, opened 2026-05-24
zen_config admin files

TMCSherpa:
This should be (almost) all of the admin files.
The only file that is causing issues is admin/includes/init_includes/init_gzip.php
[24-May-2026 19:07:53 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function zen_config() in /var/www/html/zencart-master/admin/includes/init_includes/init_gzip.php:20
so I didn't include that one.

Ref #7728


admin/includes/functions/general.php
```diff
@@ -785,7 +785,7 @@ function zen_set_ezpage_status(int $pages_id, int $status, string $status_field)
     global $db, $sniffer;
     
     // Use the $sniffer class to check if the field exists in the table
-    if (!$sniffer->field_exists(TABLE_EZPAGES, $status_field)) {
+    if (!$sniffer->field_exists($status_field, TABLE_EZPAGES)) {
```
drbyte: These are backwards.

admin/includes/init_includes/init_sessions.php
```diff
@@ -13,13 +13,13 @@
...
-zen_session_save_path(SESSION_WRITE_DIRECTORY);
+zen_session_save_path(zen_config('SESSION_WRITE_DIRECTORY'));
 
 // set the session cookie parameters
 $path = str_replace('\\', '/', dirname($_SERVER['SCRIPT_NAME']));
-if (defined('SESSION_USE_ROOT_COOKIE_PATH') && SESSION_USE_ROOT_COOKIE_PATH === 'True') $path = '/';
+if (zen_config('SESSION_USE_ROOT_COOKIE_PATH') === 'True') $path = '/';
 $path = (defined('CUSTOM_COOKIE_PATH')) ? CUSTOM_COOKIE_PATH : $path;
-$domainPrefix = (!defined('SESSION_ADD_PERIOD_PREFIX') || SESSION_ADD_PERIOD_PREFIX === 'True') ? '.' : '';
+$domainPrefix = (zen_config('SESSION_ADD_PERIOD_PREFIX') === 'True') ? '.' : '';
```
drbyte: To match the `!defined() || === 'True` logic, let's set the default to cover the !defined part:
```suggestion
$domainPrefix = (zen_config('SESSION_ADD_PERIOD_PREFIX', 'True') === 'True') ? '.' : '';
```

admin/includes/modules/dashboard_widgets/OrderStatusDashboardWidget.php
```diff
@@ -60,7 +60,7 @@
     </ul>
 
     <div class="panel-footer text-center">
-        <a href="<?= zen_href_link(FILENAME_ORDERS_STATUS) ?>" class="small text-muted">
+        <a href="<?= zen_href_link(zen_config('FILENAME_ORDERS_STATUS')) ?>" class="small text-muted">
```
drbyte: ALERT: Filename defines aren't config settings.

admin/includes/modules/dashboard_widgets/WhosOnlineDashboardWidget.php
```diff
@@ -38,7 +38,7 @@
     <div class="panel-heading">
         <i class="fa fa-globe"></i> <?= WO_GRAPH_TITLE ?>
         <div class="pull-right">
-            <a href="<?= zen_href_link(FILENAME_WHOS_ONLINE) ?>" class="btn btn-xs btn-default"><?= BOX_ENTRY_VIEW_ALL ?></a>
+            <a href="<?= zen_href_link(zen_config('FILENAME_WHOS_ONLINE')) ?>" class="btn btn-xs btn-default"><?= BOX_ENTRY_VIEW_ALL ?></a>
```
drbyte: filename

admin/orders_status.php
```diff
@@ -132,17 +132,17 @@
...
                 $orders_status = $db->Execute($orders_status_query_raw);
                 foreach ($orders_status as $status) {
                   if ((!isset($_GET['oSID']) || (isset($_GET['oSID']) && ($_GET['oSID'] == $status['orders_status_id']))) && !isset($oInfo) && (substr($action, 0, 3) != 'new')) {
                     $oInfo = new objectInfo($status);
                   }
 
                   if (isset($oInfo) && is_object($oInfo) && ($status['orders_status_id'] == $oInfo->orders_status_id)) {
-                    echo '                  <tr id="defaultSelected" class="dataTableRowSelected" onclick="document.location.href=\'' . zen_href_link(FILENAME_ORDERS_STATUS, 'page=' . $_GET['page'] . '&oSID=' . $oInfo->orders_status_id . '&action=edit') . '\'" role="option" aria-selected="true">' . "\n";
+                    echo '                  <tr id="defaultSelected" class="dataTableRowSelected" onclick="document.location.href=\'' . zen_href_link(zen_config('FILENAME_ORDERS_STATUS'), 'page=' . $_GET['page'] . '&oSID=' . $oInfo->orders_status_id . '&action=edit') . '\'" role="option" aria-selected="true">' . "\n";
```
drbyte: filename. 
And several more above and below

admin/specials.php
```diff
@@ -226,7 +226,7 @@
         if ($check_product->RecordCount() < 1) {// check for valid PID
           $skip_special = true;
           $messageStack->add_session(sprintf(WARNING_SPECIALS_PRE_ADD_PID_NO_EXIST, (int)$_POST['pre_add_products_id']), 'caution');
-        } elseif ((!defined('MODULE_ORDER_TOTAL_GV_SPECIAL') || MODULE_ORDER_TOTAL_GV_SPECIAL === 'false') && (substr($check_product->fields['products_model'] ?? '', 0, 4) === 'GIFT')) { // check for PID as a gift voucher
+        } elseif ((zen_config('MODULE_ORDER_TOTAL_GV_SPECIAL') === 'false') && (substr($check_product->fields['products_model'] ?? '', 0, 4) === 'GIFT')) { // check for PID as a gift voucher
```
drbyte:
```suggestion
        } elseif ((zen_config('MODULE_ORDER_TOTAL_GV_SPECIAL', 'false') === 'false') && (substr($check_product->fields['products_model'] ?? '', 0, 4) === 'GIFT')) { // check for PID as a gift voucher
```

admin/specials.php
```diff
@@ -338,7 +338,7 @@
           }
 
 // never include Gift Vouchers for specials when set to false
-          if (!defined('MODULE_ORDER_TOTAL_GV_SPECIAL') || MODULE_ORDER_TOTAL_GV_SPECIAL === 'false') {
+          if (zen_config('MODULE_ORDER_TOTAL_GV_SPECIAL') === 'false') {
```
drbyte:
```suggestion
          if (zen_config('MODULE_ORDER_TOTAL_GV_SPECIAL', 'false') === 'false') {
```

admin/whos_online.php
```diff
@@ -52,8 +52,8 @@
 
 if (!defined('WHOIS_SHOW_HOST')) define('WHOIS_SHOW_HOST', '1');
 
-$optURL = zen_href_link(FILENAME_WHOS_ONLINE, zen_get_all_get_params(['t', 'na', 'ns'])) . '&';
-$listingURL = zen_href_link(FILENAME_WHOS_ONLINE, zen_get_all_get_params(['q', 't', 'na', 'ns'])) . '&';
+$optURL = zen_href_link(zen_config('FILENAME_WHOS_ONLINE'), zen_get_all_get_params(['t', 'na', 'ns'])) . '&';
+$listingURL = zen_href_link(zen_config('FILENAME_WHOS_ONLINE'), zen_get_all_get_params(['q', 't', 'na', 'ns'])) . '&';
```
drbyte: filenames

admin/includes/functions/general.php
TMCSherpa: public function field_exists(string $table_name, string $field_name): bool
it was backwards before it should be table, field

admin/includes/functions/admin_access.php
```diff
@@ -445,13 +445,13 @@ function zen_validate_user_login(string $admin_name, string $admin_pass): array
         }
     } // END LOGIN SLAM PREVENTION
     // deal with expireds for SSL change
-    if (PADSS_PWD_EXPIRY_ENFORCED == 1 && $error === false && $result['pwd_last_change_date'] === '1990-01-01 14:02:22') {
+    if (zen_config('PADSS_PWD_EXPIRY_ENFORCED') === 1 && $error === false && $result['pwd_last_change_date'] === '1990-01-01 14:02:22') {
```
lat9: The `zen_config` result needs to be cast to an `(int)` for a strict comparison to be used.

admin/includes/functions/admin_access.php
```diff
@@ -445,13 +445,13 @@ function zen_validate_user_login(string $admin_name, string $admin_pass): array
...
         $expired = true;
         $error = true;
         $message = ($message === '' ? '' : $message . '<br><br>') . EXPIRED_DUE_TO_SSL;
     }
     // deal with expireds for PA-DSS
-    if ($error === false && PADSS_PWD_EXPIRY_ENFORCED == 1 && $result['pwd_last_change_date'] < date('Y-m-d H:i:s', ADMIN_PASSWORD_EXPIRES_INTERVAL)) {
+    if ($error === false && zen_config('PADSS_PWD_EXPIRY_ENFORCED') === 1 && $result['pwd_last_change_date'] < date('Y-m-d H:i:s', ADMIN_PASSWORD_EXPIRES_INTERVAL)) {
```
lat9: Same comment as line 448.

admin/group_pricing.php
```diff
@@ -63,7 +63,7 @@
 
 $query = $db->Execute("SELECT COUNT(*) AS count
                        FROM " . TABLE_GROUP_PRICING);
-if ($query->fields['count'] > 0 && (!defined('MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS') || MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS != 'true')) {
+if ($query->fields['count'] > 0 && (zen_config('MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS') != 'true')) {
```
lat9: Use strict comparison, i.e. `!== 'true'`

admin/invoice.php
```diff
@@ -489,7 +489,7 @@ function couponpopupWindow(url) { /* just a stub for coupon output that might fi
                   </td>
                 </tr>
                 <?php
-                if (ORDER_COMMENTS_INVOICE == 1 && $count_comments >= 1) {
+                if (zen_config('ORDER_COMMENTS_INVOICE') === 1 && $count_comments >= 1) {
```
lat9: Cast the `zen_config` result to an `(int)`; otherwise, the strict comparison will always fail.

admin/orders_status.php
```diff
@@ -53,7 +53,7 @@
                       WHERE configuration_key = 'DEFAULT_ORDERS_STATUS_ID'");
       }
 
-      zen_redirect(zen_href_link(FILENAME_ORDERS_STATUS, 'page=' . $_GET['page'] . '&oSID=' . $orders_status_id));
+      zen_redirect(zen_href_link(zen_config('FILENAME_ORDERS_STATUS'), 'page=' . $_GET['page'] . '&oSID=' . $orders_status_id));
```
lat9: filename

admin/orders_status.php
```diff
@@ -71,7 +71,7 @@
       $db->Execute("DELETE FROM " . TABLE_ORDERS_STATUS . "
                     WHERE orders_status_id = " . zen_db_input($oSID));
 
-      zen_redirect(zen_href_link(FILENAME_ORDERS_STATUS, 'page=' . $_GET['page']));
+      zen_redirect(zen_href_link(zen_config('FILENAME_ORDERS_STATUS'), 'page=' . $_GET['page']));
```
lat9: filename

admin/packingslip.php
```diff
@@ -362,7 +362,7 @@
                   </td>
                 </tr>
                 <?php
-                if (ORDER_COMMENTS_PACKING_SLIP == 1 && $count_comments >= 1) {
+                if (zen_config('ORDER_COMMENTS_PACKING_SLIP') === 1 && $count_comments >= 1) {
```
lat9: Needs an `(int)` cast to use strict comparison.

admin/product_types.php
```diff
@@ -262,7 +262,7 @@
         }
 
         $contents = ['form' => zen_draw_form('configuration', FILENAME_PRODUCT_TYPES, 'ptID=' . $_GET['ptID'] . '&cID=' . $cInfo->configuration_id . '&action=layout_save')];
-        if (ADMIN_CONFIGURATION_KEY_ON == 1) {
+        if (zen_config('ADMIN_CONFIGURATION_KEY_ON') === 1) {
```
lat9: Needs `(int)` cast to use strict comparison.

admin/product_types.php
```diff
@@ -277,7 +277,7 @@
     } elseif (isset($cInfo) && is_object($cInfo)) {
         $heading[] = ['text' => '<h4>' . $cInfo->configuration_title . '</h4>'];
 
-        if (ADMIN_CONFIGURATION_KEY_ON == 1) {
+        if (zen_config('ADMIN_CONFIGURATION_KEY_ON') === 1) {
```
lat9: Needs `(int)` cast to use strict comparison.

lat9 (changes requested):

lat9:
I'm not sure what that recent push was, but there weren't any changes to address the comments that @drbyte and I had previously posted.

TMCSherpa:
I think that was a fork sync on my end. This is the original files with fixes plus a couple from /includes that had the same href(zen_config) problem from a previous merge.

admin/includes/init_includes/init_sessions.php
drbyte: Still not resolved: add `, 'True'` as 2nd parameter.

admin/includes/modules/dashboard_widgets/OrderStatusDashboardWidget.php
drbyte: still not resolved

admin/includes/modules/dashboard_widgets/WhosOnlineDashboardWidget.php
```diff
@@ -5,7 +5,7 @@
  * @version $Id: ZenExpert 2026-04-06 Modified in v3.0.0 $
  */
 
-if (!zen_is_superuser() && !check_page(FILENAME_WHOS_ONLINE, '')) {
+if (!zen_is_superuser() && !check_page(zen_config('FILENAME_WHOS_ONLINE'), '')) {
```
drbyte: Filename

admin/includes/modules/dashboard_widgets/WhosOnlineDashboardWidget.php
drbyte: still not resolved

admin/orders_status.php
```diff
@@ -200,7 +200,7 @@ function getPickerVal($val) {
               case 'new':
                 $heading[] = array('text' => '<h4>' . TEXT_INFO_HEADING_NEW_ORDERS_STATUS . '</h4>');
 
-                $contents = array('form' => zen_draw_form('status', FILENAME_ORDERS_STATUS, 'page=' . $_GET['page'] . '&action=insert', 'post', 'class="form-horizontal"'));
+                $contents = array('form' => zen_draw_form('status', zen_config('FILENAME_ORDERS_STATUS'), 'page=' . $_GET['page'] . '&action=insert', 'post', 'class="form-horizontal"'));
```
drbyte: filename

admin/orders_status.php
```diff
@@ -224,7 +224,7 @@ function getPickerVal($val) {
               case 'edit':
                 $heading[] = array('text' => '<h4>' . TEXT_INFO_HEADING_EDIT_ORDERS_STATUS . '</h4>');
 
-                $contents = array('form' => zen_draw_form('status', FILENAME_ORDERS_STATUS, 'page=' . $_GET['page'] . '&oSID=' . $oInfo->orders_status_id . '&action=save', 'post', 'class="form-horizontal"'));
+                $contents = array('form' => zen_draw_form('status', zen_config('FILENAME_ORDERS_STATUS'), 'page=' . $_GET['page'] . '&oSID=' . $oInfo->orders_status_id . '&action=save', 'post', 'class="form-horizontal"'));
```
drbyte: filename

admin/orders_status.php
```diff
@@ -252,7 +252,7 @@ function getPickerVal($val) {
               case 'delete':
                 $heading[] = array('text' => '<h4>' . TEXT_INFO_HEADING_DELETE_ORDERS_STATUS . '</h4>');
 
-                $contents = array('form' => zen_draw_form('status', FILENAME_ORDERS_STATUS, 'page=' . $_GET['page'] . '&action=deleteconfirm') . zen_draw_hidden_field('oSID', $oInfo->orders_status_id));
+                $contents = array('form' => zen_draw_form('status', zen_config('FILENAME_ORDERS_STATUS'), 'page=' . $_GET['page'] . '&action=deleteconfirm') . zen_draw_hidden_field('oSID', $oInfo->orders_status_id));
```
drbyte: filename

admin/packingslip.php
drbyte: still not resolved: `(int)`

admin/specials.php
drbyte: Still not resolved: add `, 'false'` as 2nd parameter.

admin/specials.php
drbyte: Still not resolved: add `, 'false'` as 2nd parameter.

drbyte (changes requested):

drbyte:
Um ... no, I didn't mean to add defaults "for everything". I was just pointing out those couple situations where there was a `!defined() || constant ===`, where the `!defined` implied a default.

includes/modules/shipping/item.php
```diff
@@ -105,7 +105,7 @@ function check()
     function install(): void
     {
         global $db;
-        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('Enable Item Shipping', 'MODULE_SHIPPING_ITEM_STATUS', 'True', 'Do you want to offer per item rate shipping?', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
+        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('Enable Item Shipping', 'MODULE_SHIPPING_ITEM_STATUS', 'Do you want to offer per item rate shipping?', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
```
drbyte: I don't think this was supposed to be removed.

includes/modules/shipping/perweightunit.php
```diff
@@ -134,7 +134,7 @@ function check()
     function install(): void
     {
         global $db;
-        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('Enable Per Weight Unit Shipping', 'MODULE_SHIPPING_PERWEIGHTUNIT_STATUS', 'True', 'Do you want to offer per unit rate shipping?<br><br>Product Quantity * Units (products_weight) * Cost per Unit', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
+        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('Enable Per Weight Unit Shipping', 'MODULE_SHIPPING_PERWEIGHTUNIT_STATUS', 'Do you want to offer per unit rate shipping?<br><br>Product Quantity * Units (products_weight) * Cost per Unit', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
```
drbyte: Revert

includes/modules/shipping/storepickup.php
```diff
@@ -155,7 +155,7 @@ function check()
...
         $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, date_added) values ('Pickup Locations', 'MODULE_SHIPPING_STOREPICKUP_LOCATIONS_LIST', 'Walk In', 'Enter a list of locations, separated by semicolons (;).<br>Optionally you may specify a fee/surcharge for each location by adding a comma and an amount. If no amount is specified, then the generic Shipping Cost amount from the next setting will be applied.<br><br>Examples:<br>121 Main Street;20 Church Street<br>Sunnyside,4.00;Lee Park,5.00;High Street,0.00<br>Dallas;Tulsa,5.00;Phoenix,0.00<br>For multilanguage use, see the define-statement in the language file for this module.', '6', '0', now())");
```
drbyte, line 159: Revert

includes/modules/shipping/table.php
```diff
@@ -180,7 +180,7 @@ function check()
     function install(): void
     {
         global $db;
-        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Enable Table Method', 'MODULE_SHIPPING_TABLE_STATUS', 'True', 'Do you want to offer table rate shipping?', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
+        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Enable Table Method', 'MODULE_SHIPPING_TABLE_STATUS', 'Do you want to offer table rate shipping?', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
```
drbyte: Revert

includes/modules/shipping/zones.php
```diff
@@ -348,7 +348,7 @@ function check()
     function install(): void
     {
         global $db;
-        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Enable Zones Method', 'MODULE_SHIPPING_ZONES_STATUS', 'True', 'Do you want to offer zone rate shipping?', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
+        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Enable Zones Method', 'MODULE_SHIPPING_ZONES_STATUS', 'Do you want to offer zone rate shipping?', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
```
drbyte: Revert

drbyte:
Please remove `/admin/configuration.php` from this PR, because it's been updated earlier today via 2 other merges, and the zen_config changes were already done there. So whatever's currently in this PR for that file appears to be undoing recent merges to that file.

admin/includes/functions/admin_access.php
```diff
@@ -476,7 +476,7 @@ function zen_validate_user_login(string $admin_name, string $admin_pass): array
         $sql = $db->bindVars($sql, ':ip:', $_SERVER['REMOTE_ADDR'], 'string');
         $db->Execute($sql);
         $_SESSION['admin_id'] = $result['admin_id'];
-        if (SESSION_RECREATE === 'True') {
+        if (zen_config('SESSION_RECREATE', True) === 'True') {
```
drbyte: Don't need this `, True` param here. Plus it's not in quotes so it'd be wrong anyway.

admin/includes/functions/admin_access.php
```diff
@@ -765,22 +765,22 @@ function zen_get_admin_pages(bool $menu_only): array
...
-        !(defined('MODULE_PAYMENT_PAYPALDP_STATUS') && MODULE_PAYMENT_PAYPALDP_STATUS === 'True')) {
+    if (zen_config('MODULE_PAYMENT_PAYPAL_STATUS', 'True') === 'True' &&
+        zen_config('MODULE_PAYMENT_PAYPALWPP_STATUS', 'True') === 'True' &&
+        zen_config('MODULE_PAYMENT_PAYPALDP_STATUS', 'True') === 'True') {
         unset ($retVal['customers']['paypal']);
     }
 
     // don't show Coupon Admin unless installed
-    if (!defined('MODULE_ORDER_TOTAL_COUPON_STATUS') || MODULE_ORDER_TOTAL_COUPON_STATUS !== 'true') {
+    if (zen_config('MODULE_ORDER_TOTAL_COUPON_STATUS', 'True') !== 'True') {
```
drbyte: Note that the original line of code had the string as `'true'`, lowercase.
So the new line needs to keep that lowercase, in both the default and the check.

Same with the instance 2 lines below.

admin/includes/functions/admin_access.php
```diff
@@ -765,22 +765,22 @@ function zen_get_admin_pages(bool $menu_only): array
...
         unset ($retVal['gv']['couponAdmin']);
     }
     // don't show Gift Vouchers unless installed
-    if (!defined('MODULE_ORDER_TOTAL_GV_STATUS') || MODULE_ORDER_TOTAL_GV_STATUS !== 'true') {
+    if (zen_config('MODULE_ORDER_TOTAL_GV_STATUS', 'True') !== 'True') {
```
drbyte: `'true'` instead of `'True'`

admin/includes/functions/admin_access.php
```diff
@@ -765,22 +765,22 @@ function zen_get_admin_pages(bool $menu_only): array
...
         unset($retVal['gv']['gvQueue'], $retVal['gv']['gvMail'], $retVal['gv']['gvSent']);
     }
     // if Coupons and Gift Vouchers are off display msg
-    if (!defined('MODULE_ORDER_TOTAL_COUPON_STATUS') && !defined('MODULE_ORDER_TOTAL_GV_STATUS')) {
+    if (zen_config('MODULE_ORDER_TOTAL_COUPON_STATUS') && (zen_config('MODULE_ORDER_TOTAL_GV_STATUS'))) {
```
drbyte: This `!defined()` check, with no fallback, can't be replaced with simply `zen_config()` and no check around it. 
You'd need to use something like `!empty(zen_config('KEY'))` or be explicit with what the key would be if it were defined.
ie, since if those modules are installed, they would be defined as `'true'`; therefore the following works:
`if (zen_config('MODULE_ORDER_TOTAL_COUPON_STATUS') !== 'true' && zen_config('MODULE_ORDER_TOTAL_GV_STATUS') !== 'true') {`

admin/includes/functions/general.php
```diff
@@ -315,7 +315,7 @@ function zen_cfg_select_option(array $select_array, string $key_value, string $k
 function zen_cfg_select_drop_down(array $select_array, string $key_value, string $key = ''): string
 {
     $name = (!empty($key)) ? 'configuration[' . $key . ']' : 'configuration_value';
-    return zen_draw_pull_down_menu($name, $select_array, $key_value, 'class="form-control"');
+    return zen_draw_pull_down_menu($name, $select_array, (int)$key_value, 'class="form-control"');
```
drbyte: Didn't a recent PR change this? 
I'd double-check the core zencart:master repo for this line to make sure you're not reverting something we just "fixed".

admin/includes/functions/general.php
```diff
@@ -374,7 +374,7 @@ function zen_get_system_information(bool $privacy = false): array
 
     $errnum = 0;
     $system = $host = $kernel = $output = '';
-    $uptime = (DISPLAY_SERVER_UPTIME === 'true') ? 'Unsupported' : 'Disabled/Unavailable';
+    $uptime = (zen_config('DISPLAY_SERVER_UPTIME', True) === 'true') ? 'Unsupported' : 'Disabled/Unavailable';
```
drbyte: This `, True` is unnecessary, and syntactically wrong (missing quotes). Remove it. Don't need the second parameter in this case.

admin/includes/init_includes/init_errors.php
```diff
@@ -33,11 +33,11 @@
     $messageStack->add(WARNING_EMAIL_SYSTEM_DEVELOPER_OVERRIDE, 'info');
 // check if email destinations have been diverted by developer switch
 } elseif (defined('DEVELOPER_OVERRIDE_EMAIL_ADDRESS') && DEVELOPER_OVERRIDE_EMAIL_ADDRESS !== '') {
-    $messageStack->add(sprintf(WARNING_EMAIL_SYSTEM_DEVELOPER_EMAIL, zen_output_string_protected(DEVELOPER_OVERRIDE_EMAIL_ADDRESS), EMAIL_TRANSPORT), 'info');
+    $messageStack->add(sprintf(WARNING_EMAIL_SYSTEM_DEVELOPER_EMAIL, zen_output_string_protected(DEVELOPER_OVERRIDE_EMAIL_ADDRESS), zen_config('EMAIL_TRANSPORT')), 'info');
 }
 
 // this will let the admin know that the website is DOWN FOR MAINTENANCE to the public
-  if (DOWN_FOR_MAINTENANCE === 'true') {
+  if (zen_config('DOWN_FOR_MAINTENANCE', True) === 'true') {
```
drbyte: This unquoted 2nd param `True` should be removed.

admin/includes/init_includes/init_sessions.php
```diff
@@ -13,13 +13,13 @@
 require DIR_FS_CATALOG . DIR_WS_FUNCTIONS . 'sessions.php';
 
 zen_session_name('zenAdminID');
-zen_session_save_path(SESSION_WRITE_DIRECTORY);
+zen_session_save_path(zen_config('SESSION_WRITE_DIRECTORY'));
 
 // set the session cookie parameters
 $path = str_replace('\\', '/', dirname($_SERVER['SCRIPT_NAME']));
-if (defined('SESSION_USE_ROOT_COOKIE_PATH') && SESSION_USE_ROOT_COOKIE_PATH === 'True') $path = '/';
+if (zen_config('SESSION_USE_ROOT_COOKIE_PATH', 'True') === 'True') $path = '/';
```
drbyte: While the quotes around this are correct, we weren't accommodating a `!defined` previously, so the `, 'True'` here is incorrect. 
Just remove it.

admin/includes/modules/copy_product_confirm.php
```diff
@@ -159,7 +159,7 @@
 
 // copy attributes to Duplicate
     if (!empty($_POST['copy_attributes']) && $_POST['copy_attributes'] === 'copy_attributes_yes') {
-        if (DOWNLOAD_ENABLED === 'true') {
+        if (zen_config('DOWNLOAD_ENABLED', True) === 'true') {
```
drbyte: This unquoted 2nd param `True` should be removed.

admin/includes/header.php
```diff
@@ -58,7 +58,7 @@
 
 // gv queue check
 $new_gv_queue_cnt = 0;
-if (zen_config('MODULE_ORDER_TOTAL_GV_SHOW_QUEUE_IN_ADMIN') === 'true' && check_page(FILENAME_GV_QUEUE, '')) {
+if (zen_config('MODULE_ORDER_TOTAL_GV_SHOW_QUEUE_IN_ADMIN', 'True') === 'true' && check_page(FILENAME_GV_QUEUE, '')) {
```
drbyte: This 2nd param `, 'True'` should be removed.

admin/attributes_controller.php
```diff
@@ -980,7 +980,7 @@ function popupWindow(url) {
 // -----
 // Defining variables common to various display layouts.
 //
-$wholesale_pricing_enabled = (zen_config('WHOLESALE_PRICING_CONFIG') !== 'false');
+$wholesale_pricing_enabled = (zen_config('WHOLESALE_PRICING_CONFIG','false') !== 'false');
```
drbyte: @lat9 Should we have a default of `'false'` here?

admin/customers.php
```diff
@@ -1233,7 +1233,7 @@ function check_form() {
                             </th>
 
 <?php
-    if (zen_config('MODULE_ORDER_TOTAL_GV_STATUS') === 'true') {
+    if (zen_config('MODULE_ORDER_TOTAL_GV_STATUS', 'True') === 'true') {
```
drbyte: Remove this `, 'True'`

admin/customers.php
```diff
@@ -1464,7 +1464,7 @@ function check_form() {
                                     <?= $customer['pricing_group_name'] ?>
                                 </td>
 <?php
-        if (zen_config('MODULE_ORDER_TOTAL_GV_STATUS') === 'true') {
+        if (zen_config('MODULE_ORDER_TOTAL_GV_STATUS', 'True') === 'true') {
```
drbyte: Remove this `, 'True'`

admin/customers.php
```diff
@@ -1713,7 +1713,7 @@ function check_form() {
                         zen_catalog_href_link(FILENAME_LOGIN, '', 'SSL') .
                         '" method="post">';
                     $hiddenFields = zen_draw_hidden_field('email_address', $cInfo->customers_email_address);
-                    if (zen_config('EMP_LOGIN_AUTOMATIC') === 'true' && ENABLE_SSL_CATALOG === 'true') {
+                    if (zen_config('EMP_LOGIN_AUTOMATIC', 'True') === 'true' && ENABLE_SSL_CATALOG === 'true') {
```
drbyte: Remove this `, 'True'`

admin/group_pricing.php
```diff
@@ -63,7 +63,7 @@
 
 $query = $db->Execute("SELECT COUNT(*) AS count
                        FROM " . TABLE_GROUP_PRICING);
-if ($query->fields['count'] > 0 && (!defined('MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS') || MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS != 'true')) {
+if ($query->fields['count'] > 0 && (zen_config('MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS', 'True') !== 'true')) {
```
drbyte: This `'True'` should be lowercase `'false'`

admin/gv_mail.php
```diff
@@ -219,7 +219,7 @@ function check_form(form_name) {
               <td><?php echo $currencies->format(nl2br(htmlspecialchars(stripslashes($_POST['amount']), ENT_COMPAT, CHARSET, true))) . ($_POST['amount'] <= 0 ? '&nbsp;<span class="alert">' . ERROR_NO_AMOUNT_ENTERED . '</span>' : ''); ?>
               </td>
             </tr>
-            <?php if (EMAIL_USE_HTML == 'true') { ?>
+            <?php if (zen_config('EMAIL_USE_HTML', True) === 'true') { ?>
```
drbyte: This unquoted 2nd param `True` should be removed.

admin/modules.php
```diff
@@ -28,7 +28,7 @@
         if (zen_get_configuration_key_value('SHIPPING_ORIGIN_ZIP') === 'NONE' || zen_get_configuration_key_value('SHIPPING_ORIGIN_ZIP') === '') {
             $shipping_errors .= '<br>' . ERROR_SHIPPING_ORIGIN_ZIP;
         }
-        if (zen_get_configuration_key_value('ORDER_WEIGHT_ZERO_STATUS') === '1' && (!defined('MODULE_SHIPPING_FREESHIPPER_STATUS') || MODULE_SHIPPING_FREESHIPPER_STATUS !== 'True')) {
+        if (zen_get_configuration_key_value('ORDER_WEIGHT_ZERO_STATUS') === '1' && (zen_config('MODULE_SHIPPING_FREESHIPPER_STATUS', 'True') !== 'True')) {
```
drbyte: This `, 'True'` should be `, 'False'`
(If not defined, do the loop, or if not true, do the loop. So do-the-loop is for when not true. Thus the default needs to not be true.)

admin/products_price_manager.php
```diff
@@ -820,7 +820,7 @@ function updateSpecialsTaxExcl() {
                     if (empty($pInfo->products_model)) {
                         $pInfo->products_model = '';
                     }
-                    if ((substr($pInfo->products_model, 0, 4) != 'GIFT') || (substr($pInfo->products_model, 0, 4) == 'GIFT' && (defined('MODULE_ORDER_TOTAL_GV_SPECIAL') && MODULE_ORDER_TOTAL_GV_SPECIAL == 'true'))) {
+                    if ((substr($pInfo->products_model, 0, 4) != 'GIFT') || (substr($pInfo->products_model, 0, 4) == 'GIFT' && (zen_config('MODULE_ORDER_TOTAL_GV_SPECIAL', 'True') === 'true'))) {
```
drbyte: This `'True'` needs to be lowercase.

admin/products_price_manager.php
```diff
@@ -1022,14 +1022,14 @@ function updateSpecialsTaxExcl() {
...
-                        if (WHOLESALE_PRICING_CONFIG !== 'false') {
+                        if (zen_config('WHOLESALE_PRICING_CONFIG') !== 'false') {
 ?>
                           <td class="main"><?php echo zen_draw_input_field('discount_price_w[' . $discount_name[$i]['id'] . ']', $discount_name[$i]['discount_price_w'], 'class="form-control"' . $readonly); ?></td>
 <?php
                         }
 ?>
                           <?php
-                          if (DISPLAY_PRICE_WITH_TAX_ADMIN == 'true') {
+                          if (zen_config('DISPLAY_PRICE_WITH_TAX_ADMIN', True) === 'true') {
```
drbyte: This unquoted 2nd param `True` should be removed.

includes/classes/ajax/zcAjaxSelect2Lookups.php
```diff
@@ -66,7 +66,7 @@ protected function lookup_products_select2(string $lookup = '', array $exclusion
 
         $exclude_gv = '';
         // exclude GV's if feature is disabled
-        if (zen_config('MODULE_ORDER_TOTAL_GV_SPECIAL') === 'false') {
+        if (zen_config('MODULE_ORDER_TOTAL_GV_SPECIAL', 'false') === 'false') {
```
drbyte: Remove the `, 'false'` param here

includes/classes/observers/auto.downloads_via_redirect.php
```diff
@@ -41,7 +41,7 @@ class zcObserverDownloadsViaRedirect extends base {
   public function __construct()
   {
 
-    if (zen_config('DOWNLOAD_BY_REDIRECT') !== 'true') return false;
+    if (zen_config('DOWNLOAD_BY_REDIRECT', 'false') === 'false') return false;
```
drbyte: I'm not sure why this was reversed to check for false instead of not-true.
I'd prefer not to change this longstanding code if there's not a need for it.

includes/classes/Product.php
```diff
@@ -187,7 +187,7 @@ public function allowsAddToCart(): bool
 
         if ($allow_add_to_cart && $this->isGiftVoucher()) {
             // if GV feature disabled, can't allow GV's to be added to cart
-            if (zen_config('MODULE_ORDER_TOTAL_GV_STATUS') !== 'true') {
+            if (zen_config('MODULE_ORDER_TOTAL_GV_STATUS', 'false') === 'false') {
```
drbyte: I don't understand why this was changed. 
I'd prefer to revert this change and just check against true. No defaults required.

includes/modules/pages/index/header_php.php
```diff
@@ -119,7 +119,7 @@
 }
 
 // include template specific file name defines
-$define_page = zen_get_file_directory(DIR_WS_LANGUAGES . $_SESSION['language'] . '/html_includes/', FILENAME_DEFINE_MAIN_PAGE, 'false');
+$define_page = zen_get_file_directory(DIR_WS_LANGUAGES . $_SESSION['language'] . '/html_includes/', FILENAME_DEFINE_MAIN_PAGE);
```
drbyte: Why was the trailing `, 'false'` removed from this? 
This line has nothing to do with the `zen_config()` changes that this PR is about.

includes/modules/payment/paypal/tpl_ec_button.php
```diff
@@ -9,7 +9,7 @@
 
 // PayPal module cannot be used for purchase > $10,000 USD equiv
 require_once DIR_FS_CATALOG . DIR_WS_MODULES . 'payment/paypal/paypal_currency_check.php';
-$paypalec_enabled = (zen_config('MODULE_PAYMENT_PAYPALWPP_STATUS') === 'True'  && paypalUSDCheck($_SESSION['cart']->total) === true);
+$paypalec_enabled = (zen_config('MODULE_PAYMENT_PAYPALWPP_STATUS', 'True') === 'True'  && paypalUSDCheck($_SESSION['cart']->total) === true);
```
drbyte: Remove the `, 'True'` param here.

includes/modules/shipping/freeshipper.php
```diff
@@ -111,7 +111,7 @@ function get_configuration_errors()
     function install(): void
     {
         global $db;
-        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('Enable Free Shipping', 'MODULE_SHIPPING_FREESHIPPER_STATUS', 'True', 'Do you want to offer Free shipping?', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
+        $db->Execute("insert into " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) values ('Enable Free Shipping', 'MODULE_SHIPPING_FREESHIPPER_STATUS', 'Do you want to offer Free shipping?', '6', '0', 'zen_cfg_select_option(array(\'True\', \'False\'), ', now())");
```
drbyte: Revert

includes/modules/shipping/freeoptions.php
```diff
@@ -212,7 +212,7 @@ public function get_configuration_errors()
     public function install(): void
     {
         global $db;
-        $db->Execute("INSERT INTO " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Enable Free Options Shipping', 'MODULE_SHIPPING_FREEOPTIONS_STATUS', 'True', 'Free Options is used to display a Free Shipping option when other Shipping Modules are displayed.
+        $db->Execute("INSERT INTO " . TABLE_CONFIGURATION . " (configuration_title, configuration_key, configuration_value, configuration_description, configuration_group_id, sort_order, set_function, date_added) VALUES ('Enable Free Options Shipping', 'MODULE_SHIPPING_FREEOPTIONS_STATUS', 'Free Options is used to display a Free Shipping option when other Shipping Modules are displayed.
```
drbyte: Revert

admin/includes/functions/admin_access.php
```diff
@@ -765,22 +765,22 @@ function zen_get_admin_pages(bool $menu_only): array
  * @since ZC v1.5.0
      */
     // Include PayPal Standard menu only if that payment mod is enabled
-    if (!(defined('MODULE_PAYMENT_PAYPAL_STATUS') && MODULE_PAYMENT_PAYPAL_STATUS === 'True') &&
-        !(defined('MODULE_PAYMENT_PAYPALWPP_STATUS') && MODULE_PAYMENT_PAYPALWPP_STATUS === 'True') &&
-        !(defined('MODULE_PAYMENT_PAYPALDP_STATUS') && MODULE_PAYMENT_PAYPALDP_STATUS === 'True')) {
+    if (zen_config('MODULE_PAYMENT_PAYPAL_STATUS', 'True') === 'True' &&
+        zen_config('MODULE_PAYMENT_PAYPALWPP_STATUS', 'True') === 'True' &&
+        zen_config('MODULE_PAYMENT_PAYPALDP_STATUS', 'True') === 'True') {
```
drbyte: These 3 need to remove the `, 'True'`

admin/products_price_manager.php
```diff
@@ -820,7 +820,7 @@ function updateSpecialsTaxExcl() {
                     if (empty($pInfo->products_model)) {
                         $pInfo->products_model = '';
                     }
-                    if ((substr($pInfo->products_model, 0, 4) != 'GIFT') || (substr($pInfo->products_model, 0, 4) == 'GIFT' && (defined('MODULE_ORDER_TOTAL_GV_SPECIAL') && MODULE_ORDER_TOTAL_GV_SPECIAL == 'true'))) {
+                    if ((substr($pInfo->products_model, 0, 4) != 'GIFT') || (substr($pInfo->products_model, 0, 4) == 'GIFT' && (zen_config('MODULE_ORDER_TOTAL_GV_SPECIAL', 'true') === 'true'))) {
```
drbyte: I'm pretty sure this should not set a default.

includes/classes/observers/auto.downloads_via_redirect.php
```diff
@@ -41,7 +41,7 @@ class zcObserverDownloadsViaRedirect extends base {
   public function __construct()
   {
 
-    if (zen_config('DOWNLOAD_BY_REDIRECT') !== 'true') return false;
+    if (zen_config('DOWNLOAD_BY_REDIRECT', 'true') !== 'true') return false;
```
drbyte:
```suggestion
    if (zen_config('DOWNLOAD_BY_REDIRECT') !== 'true') return false;
```

admin/includes/functions/admin_access.php
lat9: This should be changed to
```php
if (zen_config('MODULE_PAYMENT_PAYPAL_STATUS') !== 'True' &&
    zen_config('MODULE_PAYMENT_PAYPALWPP_STATUS') !== 'True' &&
    zen_config('MODULE_PAYMENT_PAYPALDP_STATUS') !== 'True') {
```

The (albeit confusing) clause should be `true` if none of the 'paypal', 'paypalwpp' or 'paypaldp' payment modules are installed and enabled.

admin/includes/functions/admin_access.php
```diff
@@ -765,22 +765,22 @@ function zen_get_admin_pages(bool $menu_only): array
...
         unset ($retVal['customers']['paypal']);
     }
 
     // don't show Coupon Admin unless installed
-    if (!defined('MODULE_ORDER_TOTAL_COUPON_STATUS') || MODULE_ORDER_TOTAL_COUPON_STATUS !== 'true') {
+    if (zen_config('MODULE_ORDER_TOTAL_COUPON_STATUS', 'true') !== 'true') {
```
lat9: The 2nd parameter to `zen_config` isn't needed.

admin/includes/functions/admin_access.php
```diff
@@ -765,22 +765,22 @@ function zen_get_admin_pages(bool $menu_only): array
...
         unset ($retVal['gv']['couponAdmin']);
     }
     // don't show Gift Vouchers unless installed
-    if (!defined('MODULE_ORDER_TOTAL_GV_STATUS') || MODULE_ORDER_TOTAL_GV_STATUS !== 'true') {
+    if (zen_config('MODULE_ORDER_TOTAL_GV_STATUS', 'true') !== 'true') {
```
lat9: Remove the 2nd parameter.

admin/includes/functions/admin_access.php
```diff
@@ -765,22 +765,22 @@ function zen_get_admin_pages(bool $menu_only): array
...
         unset($retVal['gv']['gvQueue'], $retVal['gv']['gvMail'], $retVal['gv']['gvSent']);
     }
     // if Coupons and Gift Vouchers are off display msg
-    if (!defined('MODULE_ORDER_TOTAL_COUPON_STATUS') && !defined('MODULE_ORDER_TOTAL_GV_STATUS')) {
+    if (zen_config('MODULE_ORDER_TOTAL_COUPON_STATUS') !== 'true' && zen_config('MODULE_ORDER_TOTAL_GV_STATUS') !== 'true') {
```
lat9: This should be changed to
```php
if (!zen_config('MODULE_ORDER_TOTAL_COUPON_STATUS') && !zen_config('MODULE_ORDER_TOTAL_GV_STATUS')) {
```

admin/products_price_manager.php
drbyte:
```suggestion
                    if ((substr($pInfo->products_model, 0, 4) != 'GIFT') || (substr($pInfo->products_model, 0, 4) == 'GIFT' && (zen_config('MODULE_ORDER_TOTAL_GV_SPECIAL') === 'true'))) {
```

admin/group_pricing.php
```diff
@@ -63,7 +63,7 @@
 
 $query = $db->Execute("SELECT COUNT(*) AS count
                        FROM " . TABLE_GROUP_PRICING);
-if ($query->fields['count'] > 0 && (!defined('MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS') || MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS != 'true')) {
+if ($query->fields['count'] > 0 && (zen_config('MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS', 'false') !== 'true')) {
```
lat9: Remove 2nd parameter from `zen_config`; it's not needed.

admin/modules.php
```diff
@@ -28,7 +28,7 @@
         if (zen_get_configuration_key_value('SHIPPING_ORIGIN_ZIP') === 'NONE' || zen_get_configuration_key_value('SHIPPING_ORIGIN_ZIP') === '') {
             $shipping_errors .= '<br>' . ERROR_SHIPPING_ORIGIN_ZIP;
         }
-        if (zen_get_configuration_key_value('ORDER_WEIGHT_ZERO_STATUS') === '1' && (!defined('MODULE_SHIPPING_FREESHIPPER_STATUS') || MODULE_SHIPPING_FREESHIPPER_STATUS !== 'True')) {
+        if (zen_get_configuration_key_value('ORDER_WEIGHT_ZERO_STATUS') === '1' && (zen_config('MODULE_SHIPPING_FREESHIPPER_STATUS', 'False') !== 'True')) {
```
lat9: Remove 2nd ('False') parameter from `zen_config`.  If the setting's not defined the value returned is `null` which isn't strictly equal to 'True'.

admin/products_price_manager.php
lat9: Remove 2nd parameter from `zen_config`.  If the value's not defined, it'll be `null` which won't match 'true'.

includes/classes/observers/auto.downloads_via_redirect.php
lat9: Don't need the 2nd parameter to `zen_config`.

includes/modules/pages/login/header_php.php
```diff
@@ -166,7 +166,7 @@
 $breadcrumb->add(NAVBAR_TITLE);
 
 // Check for PayPal express checkout button suitability:
-$paypalec_enabled = (defined('MODULE_PAYMENT_PAYPALWPP_STATUS') && MODULE_PAYMENT_PAYPALWPP_STATUS == 'True' && defined('MODULE_PAYMENT_PAYPALWPP_ECS_BUTTON') && MODULE_PAYMENT_PAYPALWPP_ECS_BUTTON == 'On');
+$paypalec_enabled = (zen_config('MODULE_PAYMENT_PAYPALWPP_STATUS', 'True') === 'True' && zen_config('MODULE_PAYMENT_PAYPALWPP_ECS_BUTTON') === 'On');
```
lat9: Remove 2nd parameter from `zen_config('MODULE_PAYMENT_PAYPALWPP_STATUS')`.

lat9 (changes requested):
Noting that there might be overlap here with the Copilot review just kicked off.

admin/includes/functions/admin_access.php
drbyte:
```suggestion
    if (zen_config('MODULE_ORDER_TOTAL_COUPON_STATUS') !== 'true') {
```

admin/includes/functions/admin_access.php
drbyte:
```suggestion
    if (zen_config('MODULE_ORDER_TOTAL_GV_STATUS') !== 'true') {
```

admin/includes/functions/admin_access.php
drbyte: @lat9 I had suggested this approach of declaring `!== 'true'` to match with the only valid string that it would have if those were installed. It's definitely adding more specificity than was there before, but I was thinking it made things somewhat clearer.
But your approach could be used too.

admin/group_pricing.php
drbyte:
```suggestion
if ($query->fields['count'] > 0 && (zen_config('MODULE_ORDER_TOTAL_GROUP_PRICING_STATUS') !== 'true')) {
```

admin/modules.php
drbyte:
```suggestion
        if (zen_get_configuration_key_value('ORDER_WEIGHT_ZERO_STATUS') === '1' && (zen_config('MODULE_SHIPPING_FREESHIPPER_STATUS') !== 'True')) {
```

includes/modules/pages/login/header_php.php
drbyte:
```suggestion
$paypalec_enabled = (zen_config('MODULE_PAYMENT_PAYPALWPP_STATUS') === 'True' && zen_config('MODULE_PAYMENT_PAYPALWPP_ECS_BUTTON') === 'On');
```

admin/includes/functions/admin_access.php
drbyte:
```suggestion
    if (!zen_config('MODULE_ORDER_TOTAL_COUPON_STATUS') && !zen_config('MODULE_ORDER_TOTAL_GV_STATUS')) {
```

lat9:
@TMCSherpa, are you in the process of making additional storefront updates for the `zen_config` change?  If not, I'll continue with those, but I don't want us to duplicate efforts if you are!

TMCSherpa:
Yes. I found a few things my script was having problems with and have most of it working, there are a couple nonstandard syntax (KEY==value rather than KEY == value) I'm trying to catch.
While trouble shooting I found a couple config_keys I didn't have before so I wrote something to pull them from zc_install and check the various modules.

PR #7756, opened 2026-05-27
`init_sanitize`, casting all values to string

### LINKED ISSUE #7636: init_sanitize: debug using str_replace with $value is an integer (strict mode)
using strict_mode gets a debug from the use of str_replace since $value is an integer

https://github.com/zencart/zencart/blob/2ccf17e28b0373bb8b4e786e39937abb95863b95/includes/init_includes/init_sanitize.php#L273-L280

lat9:
Fixes #7636
Ref #7728


torvista (commented):
works for me

PR #7773, opened 2026-06-04
Drop /admin/includes/configure.php

drbyte:
Now the admin shares the catalog configure.php  -- less complexity for maintenance and installation.

Previously we had both `/includes/configure.php` and `/admin/includes/configure.php` which had a lot of duplication in them.
Now that we've removed the dependency on some of those constants (auto-detected now), we don't need the duplication.

So this PR removes the need for the `/admin/includes/configure.php` and relies on the `/includes/configure.php`.
It still supports `/includes/local/configure.php` for local dev PC use.

If an old admin configure.php is found, it will use it ... to help avoid breaking changes ... but storeowners should delete the admin configure.php file after upgrading, to avoid introducing unexpected issues.

This PR also has `zc_install` only generate the single configure.php, from a simplified template. The installer UI has correspondingly fewer input fields as well.


not_for_release/testFramework/Support/TestConfigResolver.php
```diff
@@ -32,6 +32,7 @@ public static function resolveConfigPath(string $context, ?string $basePath = nu
         $pathsTried = [];
 
         foreach ($candidates as $candidate) {
+            // @TODO: is this $context needed anymore?
```
drbyte: @zcwilt Not sure if $context will need to vary differently anymore ... at least as far as config files go.

proseLA:
> So this PR removes the need for the `/admin/includes/configure.php` and relies on the `/includes/configure.php`. It still supports `/includes/local/configure.php` for local dev PC use.
> 
i am a big fan of this PR.

with regards to the `/includes/local/configure.php`, i have never truly understood its use.  i suppose it gets into how one sets up their development environment, and everyone seems to do something different.

while i do not want to get into thread drift, in the past, the `/includes/local/configure.php` threw me for a loop.  it seems that that `zc_install` process treats this file differently than the storefront or the admin.  from what i recollect, the `zc_install` process, if present, only loads the local file and not the regular configure file.  meanwhile the storefront and the admin load them both.

i would advocate that all 3 processes should treat the load the same.

drbyte:
@proseLA you make a good point.
In zc_install, we don't "load" the config file, we read-and-parse it using regex pattern matching. That's a lot harder to do with a second file in the mix where that second file might nest "if not defined, define..." on one or multiple lines for human readability but near impossible if not having PHP "require" the file and eval it. (We avoid that eval because we don't actually want the constants declared in zc_install's context.)

I personally have never used the "local" config dir, and would happily dispense with it.

I'd prefer that a happy middle-ground occur:  if the `/local/` configure.php exists, use it instead of the `/includes/` configure.php.  Either/or, not both-and.  That's what zc_install does now, and I'd happily support that for admin+catalog. It would actually be my preference. And I don't mind if that's a breaking change.

However, I'd be okay with keeping it the way it is now, and whoever relies on the "local" file should just know that it needs to be "complete" (ie: all the expected defines should be present, not only a couple).

Edit to add: `zc_cli` also only loads the first one found, not both.

proseLA:
> I'd prefer that a happy middle-ground occur: if the `/local/` configure.php exists, use it instead of the `/includes/` configure.php. Either/or, not both-and. That's what zc_install does now, and I'd happily support that for admin+catalog. It would actually be my preference. And I don't mind if that's a breaking change.
> 
nor do i. 

it sounds good to me.

torvista:
I've always assumed it was either or, with a hack to load it if exists and not the core version.

Or, perhaps define a standard dev-development file/procedure..that includes these local parameters, allows "admin" and "zc_install" to exist etc. for easy use of the repo locally.

drbyte:
Okay, changed to load only either/or. 
Note that the storefront "location" of loading changed slightly to merge them. It's now after debugging and datetime defaults are set.

scottcwilson:
> Okay, changed to load only either/or.

I thought it was intended to be additive.   And documented it this way. 

https://docs.zen-cart.com/user/troubleshooting/blank_page/#advanced-developer-option-this-approach-is-not-needed-for-most-situations

https://docs.zen-cart.com/user/troubleshooting/strict_error_reporting/

drbyte:
Fair. Granted, that note about setting DEBUG_AUTOLOAD to true via local-configure.php could be rewritten the same as point #6 above it where it talks about temporarily directly editing admin application_bootstrap.php to do the same. For catalog, can edit application_top in the same way.

proseLA:
my opinion is that there should be only 1 configuration file.  period.

if you want strict error reporting, put it in said configuration file.

you want to turn it off, comment the line out or delete the line.

shoot, the other day, i was trying to figure out how to set:

```
<meta name="robots" content="noindex, nofollow">
```

for all pages.  frankly, for me, its too much work looking through the code.  and then deciphering what the config value really means.

i ended up putting:

```php
$robotsNoIndex = true;
```
in the configure file.

turning debug on, again, just do it here.

the configure file will be unique to every install (lets not talk about edge cases or development boxes that are rolled out exactly the same).  so my opinion is to just set all of these things in one place.  

i find it easier than digging through all of the code.  and i would hope that we can use it in this manner.

scottcwilson:
> Granted, that note about setting DEBUG_AUTOLOAD to true via local-configure.php could be rewritten the same as point #6 above it where it talks about temporarily directly editing admin application_bootstrap.php to do the same.

There are several ways it could be done, but this is how it is documented, and has been documented for over a decade, since this came from the old content.php documentation. 

If you're going to make this drastic a change, it would be better to just get rid of the local configure file and completely update the guidance for people using 3.0+.

PR #7777, opened 2026-06-06
Allow debug level overrides in configure.php

drbyte:
Updates #7773 

Rearranging this logic allows `/includes/configure.php` or `/includes/local/configure.php` to have debug constants defined which can control debug flow.

### DIFF
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -250,24 +250,6 @@ if (PHP_VERSION_ID < 80300) {
     exit(0);
 }
 
-/**
- * boolean if true the autoloader scripts will be parsed and their output shown. For debugging purposes only.
- */
-define('DEBUG_AUTOLOAD', false);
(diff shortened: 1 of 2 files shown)


includes/application_top.php
```diff
@@ -316,6 +298,24 @@
     exit;
 }
 
+/**
+ * boolean if true the autoloader scripts will be parsed and their output shown. For debugging purposes only.
+ */
+define('DEBUG_AUTOLOAD', false);
```
zcwilt: Should it not be 

if (!defined('DEBUG_AUTOLOAD')) define('DEBUG_AUTOLOAD', false);

zcwilt (dismissed):

includes/application_top.php
drbyte: Yes

PR #7790, opened 2026-06-08
Enable observer to add ez-pages fields

### LINKED ISSUE #7787: Use of NOTIFY_ADMIN_EZPAGES_UPDATE_BASE
Not sure how this can work..

At this point $sql_data_array is null.

https://github.com/zencart/zencart/blob/16a828460744e4e976f15d6c4ba54ecbc894d8a6/admin/ezpages.php#L96-L108

Subsequently it is created. So if intended to be manipulated/created by the observer, that gets overwritten.

Comment implies the page_error parameter is optional.

Use case is to trap the page url to pass to IndexNow plugin.

lat9:
Fixes #7787

### DIFF
--- a/admin/ezpages.php
+++ b/admin/ezpages.php
@@ -101,26 +101,33 @@ if (!empty($action)) {
             // If the observer sets the $page_error (i.e. $p2) value to (bool)true, it is the observer's
             // responsibility to add a message for display to the current admin.
             //
+            // If the observer adds fields to the $sql_data_array, they'll be merged with the
+            // 'base' values, with the observer's values overriding.
+            //
+            $sql_data_array = [];
             $zco_notifier->notify('NOTIFY_ADMIN_EZPAGES_UPDATE_BASE', $action, $page_error, $sql_data_array);
 
             if ($page_error === false) {
-                $sql_data_array = [
-                    'page_open_new_window' => $page_open_new_window,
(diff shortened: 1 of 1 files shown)


drbyte:
@codex review

drbyte:
While the code has no runtime errors, I'm thinking the $sql_data_array should be populated with expected values before the observer is fired, so real data is passed to the observer for inspection. Otherwise it kinda doesn't even know what page is being dealt with, let alone what the content is about to be saved as.

lat9:
The purpose of that notification was to enable the verification and/or addition of **other** fields to the EZ-Pages.  Granted, most of the information requires an observer to look through global variables to pick and choose any fields that are of interest, but I believe that the notification currently serves its purpose.

Until @torvista's issue was noted, the only plugin that I'm aware of that's using the notification is [EZ-Pages MetaTags](https://www.zen-cart.com/downloads.php?do=file&id=746), which simply uses the notification as a means to record what is going to be updated.

torvista:
I guess it need to make the relevant variable a global to access it?

I would have thought a cleaner/more comprehensive solution would be as suggested and pass the sql_array.

drbyte:
@codex review

lat9:
Looks good, @drbyte, thanks for the update.

PR #7796, opened 2026-06-11
Add php-cs-fixer

drbyte:
## Using
(Must run `composer install` or `composer update` to pull it in.)

`php-cs-fixer check` will report which files could be reformatted by running `fix`. 
`php-cs-fixer fix {dir/filename}` will reformat the specified file(s) according to Zen Cart code conventions.

(Those conventions are largely outlined in `/docs/CONVENTIONS.md`, but are basically PSR-12 with a few adjustments to whitespace.)

**But ALWAYS manually review whatever changes it proposes before committing.**

---
## CAVEATS:

One downside to php-cs-fixer is that it doesn't handle mixed PHP/HTML files well, and thus it may strangely indent/outdent sections of code, or rearrange comments, in strange ways. Some rules have been tweaked to try to minimize the pain of that.

Conversion to short-echo syntax works well, except php-cs-fixer won't remove the trailing `;` because it's syntactically valid and php-cs-fixer operates on php tokens not regular-expressions.

This is a starting point. We'll tweak/refine it as we go.

---
## ADVANCED:

If you want to force the inclusion of rewrites to function-aliases, insertion of declare-strict-types, and adding return types, you can create your own local `.php-cs-fixer.php` file by copying the `dist` version and enabling `@auto:risky` and uncommenting the declare-strict-types section (and set the blank-line-after-open-tag to true).

I've left these disabled for core at this point, in favor of manual implementation of such where code can be fully reviewed for compatibility with those changes.

---

I've included a few reformatted files in this PR as examples of what the current rules "do".

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -126,13 +126,13 @@ function zen_cfg_select_coupon_id(string $coupon_id, string $key = ''): string
     $coupons = Coupon::getAllCouponsByName();
     $coupon_array[] = [
         'id' => '0',
-        'text' => TEXT_NONE
+        'text' => TEXT_NONE,
     ];
 
     foreach ($coupons as $coupon) {
         $coupon_array[] = [
             'id' => $coupon['coupon_id'],
-            'text' => $coupon['coupon_name']
+            'text' => $coupon['coupon_name'],
         ];
     }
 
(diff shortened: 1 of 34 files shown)


admin/includes/functions/general.php
```diff
@@ -552,7 +552,8 @@ function zen_draw_order_status_dropdown(string $field_name, $default_value, stri
 function zen_get_language_icon(string|int $lookup): string
 {
     global $db;
-    $languages_icon = $db->Execute("
+    $languages_icon = $db->Execute(
+        "
```
scottcwilson, line 556: I wonder why this happens.

scottcwilson:
Except the small nit above this looks great.

torvista:
Regarding a file that already uses <?=, does it add the trailing ; where it is "missing"?

drbyte:
> Regarding a file that already uses <?=, does it add the trailing ; where it is "missing"?

No, it ignores that, thankfully.

admin/includes/functions/general.php
drbyte: multi-line param separation

zcwilt (approved):

PR #7801, opened 2026-06-12
Reformatting

drbyte:
This is a first round of reformatting.
- PSR-12
- alias functions renamed
- strict-types
- short-echo

Only touching classes and a couple function files in this round.

(files that combine a lot of HTML+PHP intermingled are much harder to reformat programmatically, which means templates and admin files are the most challenging ... they're lower on the priority list for that reason)

In this pull, probably the biggest risk is having added a lot of types to the shopping_cart class.

### DIFF
--- a/admin/includes/classes/AdminNotifications.php
+++ b/admin/includes/classes/AdminNotifications.php
@@ -59,7 +59,7 @@ class AdminNotifications
     /**
      * @since ZC v1.5.6
      */
-    protected function getNotificationInfo(): array|false
+    protected function getNotificationInfo()
     {
         if (empty($this->projectNotificationServer)) {
             return [];
--- a/admin/includes/classes/AdminNotifications.php
+++ b/admin/includes/classes/AdminNotifications.php
@@ -1,16 +1,18 @@
 <?php
+
+declare(strict_types=1);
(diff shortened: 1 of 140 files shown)


drbyte:
@codex review

torvista:
The customary howls of protest that I've received in the past (from those tasked with upgrading customers sites) are noticeably absent for this and other recent big-bang global fettling-without-any-real-changes. 
Not a complaint, just an observation on attitudes, which reinforces my point of view that any effort made by anyone, however trivial, should be, and should have been encouraged.

lat9:
> The customary howls of protest that I've received in the past (from those tasked with upgrading customers sites) are noticeably absent for this and other recent big-bang global fettling-without-any-real-changes. Not a complaint, just an observation on attitudes, which reinforces my point of view that any effort made by anyone, however trivial, should be, and should have been encouraged.

This PR accompanies the `zen_config` updates which have already touched the vast majority of the modules.  Doing this reformatting now as an overall group rather than single-poking as has been done in the past.

Essentially, the slate is being cleared to future-proof for upcoming PHP and Zen Cart changes.

drbyte:
@codex review

PR #7803, opened 2026-06-12
Add a JS loader method to InteractsWithPlugins trait

drbyte:
Previously via the `InteractsWithPlugins` traith, one could add CSS via:
```php
            // load a plugin-specific stylesheet, first from plugin dir, then from store's template:
            $this->linkCatalogStylesheet('some_filename.min.css', $current_page_base);
```
This adds the ability to load JS:
```php
            // load a PHP file from the plugin's jscript directory (via include_once())
            $this->linkCatalogJscript('a_js_php_filename.php', $current_page_base);

            // or a .js file, via <script src=...>:
            $this->linkCatalogJscript('some_js_filename.js', $current_page_base);
```

As before, these assume the path is `zc_plugins/name/ver/catalog/includes/templates/default/jscript`. (the .css was for the /css/ dir)
(Edit: `default`, not `template_default`)


lat9:
> As before, these assume the path is zc_plugins/name/ver/catalog/includes/templates/template_default/jscript. (the .css was for the /css/ dir)

The 'correct' plugin directory would be `zc_plugins/name/ver/catalog/includes/templates/default/jscript`; the subdirectory is `default`, not `template_default`.

includes/classes/traits/InteractsWithPlugins.php
```diff
@@ -110,4 +110,33 @@ protected function linkCatalogStylesheet(string $stylesheet_filename, ?string $c
...
+    {
+        global $pageLoader;
+        if (!$pageLoader) {
+            $pageLoader = PageLoader::getInstance();
+        }
+
+        $jsFilename = basename($jsFilename);
+        if (file_exists($file = $pageLoader->getTemplatePluginDir($jsFilename, 'jscript', $this->zcPluginDirName) . $jsFilename)) {
+            if (str_ends_with($jsFilename, '.js')) {
+                echo '<script title="' . \zen_output_string_protected($this->zcPluginDirName) . '" src="' . $file . '">' . "\n";
```
lat9: Use `id=`, not `title=`.  We'd run into some kind of issue with a 'title' attribute on the admin's keepalive-jscript.

includes/classes/traits/InteractsWithPlugins.php
```diff
@@ -110,4 +110,33 @@ protected function linkCatalogStylesheet(string $stylesheet_filename, ?string $c
...
+                return true;
+            }
+
+            if (str_ends_with($jsFilename, '.php')) {
+                include_once $file;
```
lat9: Why `include_once`? Shouldn't it be `require`?

lat9 (changes requested):

includes/classes/traits/InteractsWithPlugins.php
drbyte: `title` is correct for `<script>`

We had to use `id` for `<style>`
Screen Shot 2026-06-12 at 9 11 57 PM

includes/classes/traits/InteractsWithPlugins.php
drbyte: I have no objection.

includes/classes/traits/InteractsWithPlugins.php
lat9: My bad; it's late (for me).

Issue #7800, opened 2026-06-11
PHP Fatal error: Uncaught TypeError: number_format() with ot_gv.php


JeffMash:
(Not sure if this is a bug with ZC or with OPC, but I'm logging it here just in case)

Issue: With the Gift Certificate (ot_gv.php), if someone puts in wrong characters, it throws a PHP Failure.

For example, a customer apparently was trying to redeem a $30.00 gift balance, and he typed in: $30,oo.   This error would happen with the dollar sign being typed, but I'm sure also with the comma and the letter o's.

This caused the following failure (blank white screen for the customer).   This bug would also happen if the customer puts in a non-numeric character like a comma or a letter o.

Admin Log Example on this user-error:
[11-Jun-2026 11:55:21 America/Los_Angeles] Request URI: /store/index.php?main_page=checkout_one, IP address: 75.164.229.89
--> PHP Fatal error: Uncaught TypeError: number_format(): Argument #1 ($num) must be of type float, string given in /includes/modules/order_total/ot_gv.php:126
Stack trace:
#0 /includes/modules/order_total/ot_gv.php(126): number_format('$30,oo', 2)
#1 /includes/classes/order_total.php(76): ot_gv->__construct()
#2 /includes/modules/pages/checkout_one/header_php.php(335): order_total->__construct()
#3 /index.php(40): require('/home/username/...')
#4 {main}
thrown in /includes/modules/order_total/ot_gv.php on line 126.
[11-Jun-2026 11:55:21 America/Los_Angeles] Request URI: /store/index.php?main_page=checkout_one, IP address: 75.164.229.89
--> PHP Fatal error: Uncaught TypeError: number_format(): Argument #1 ($num) must be of type float, string given in /includes/modules/order_total/ot_gv.php:126
Stack trace:
#0 /includes/modules/order_total/ot_gv.php(126): number_format('$30,oo', 2)
#1 /includes/classes/order_total.php(76): ot_gv->__construct()
#2 /includes/modules/pages/checkout_one/header_php.php(335): order_total->__construct()
#3 /index.php(40): require('/home/username/...')
#4 {main}
thrown in /includes/modules/order_total/ot_gv.php on line 126.

**Version**
Zen Cart version: 2.2.2
PHP version: 8.2
MySQL version: MySQL 10.6.27-MariaDB
Plugins that may be related: OPC 2.6.0 (2026-03-30)

**To Reproduce**
[...]

PR #7816, opened 2026-06-15
Refactor/reformat ot_gv.php

lat9:
Ref #7728 
Precursor to correction for issue #7800 

- Adding class property type-hints and type hints on some of the methods
- Use `zen_config` for database configuration settings
- PSR-12 coding style
- Use strict comparisons where warranted
- Use short-array syntax
- Remove redundant `selection_test` method; replace by `user_has_gv_account`
- Use `!$result->EOF` instead of `$result->RecordCount() > 0`


drbyte (commented):
# Codex said:

Nice cleanup overall — the PSR-12 reformatting is consistent with the other refactored modules, and the `selection_test()` removal, strict-comparison changes, and `!$result->EOF` swaps all check out _(I grepped for external callers of the now-`protected` methods and the removed `selection_test()`/`validation_errors` — none found)._

Suggested resolutions before merge, plus a couple of minor nits:

### 1. Typed properties without defaults + early-return constructor

The new constructor does:

```php
$sort_order = zen_config('MODULE_ORDER_TOTAL_GV_SORT_ORDER');
if (null === $sort_order) {
    return;
}
$this->sort_order = (int)$sort_order;
$this->include_shipping = zen_config('MODULE_ORDER_TOTAL_GV_INC_SHIPPING');
$this->include_tax = ...
// etc.
```

with declarations like `public ?int $sort_order;`, `public string $include_tax;`, `public string $checkbox;`, `public bool $credit_class;`, `public int $tax_class;` — none have default values.

A typed property with no default is *uninitialized* in PHP, not `null`. If `zen_config('MODULE_ORDER_TOTAL_GV_SORT_ORDER')` returns `null` (e.g. the module has never been installed/configured, which happens when admin's Modules page instantiates every discovered module class to list it), the constructor returns early and `sort_order`, `include_tax`, `include_shipping`, `calculate_tax`, `credit_tax`, `tax_class`, `credit_class`, and `checkbox` are never assigned. Any later *direct* (non-`??`/`isset`) access — e.g. `get_order_total_details()`'s `$this->include_tax !== 'true'` — throws:

```
Error: Typed property ot_gv::$include_tax must not be accessed before initialization
```

`admin/modules.php` accesses `->sort_order` via `??`, which is safe (uninitialized typed props behave like "unset" for `isset`/`??`), but a direct read elsewhere wouldn't be. For comparison, the already-refactored `includes/modules/shipping/zones.php` assigns `$this->sort_order = zen_config(...)` *before* the null check, so the property is at least initialized to `null` on early return.

The existing unit tests (`OtGvShippingDetailsTest.php`, `OtGvShippingDetailsTaxInclusiveTest.php`) bypass this via an anonymous subclass with a no-op `__construct()` that sets `include_tax`/`include_shipping` directly, so this path isn't covered by CI.

Suggested fix: give the affected properties safe defaults, e.g. `public ?int $sort_order = null;`, `public string $include_tax = '';`, `public string $checkbox = '';`, `public bool $credit_class = false;`, `public int $tax_class = 0;`. Preserves the old "uninitialized → null/empty" semantics without going back to untyped properties, at no cost.

### 2. Minor nit 1:

- `update_credit_account($i)` is missing a visibility keyword — every other method got `public`/`protected` in this pass, but this one is still bare `function update_credit_account($i)`.

### 3. Minor nit 2:
- Mixed "limit to one row" idioms: `apply_credit()` / part of `collect_posts()` use `$db->Execute($sql, 1)` (the `1` becomes `LIMIT 1` via `query_factory.php`'s `$limit` param), while `update_credit_account()` and `user_has_gv_account()` append `" LIMIT 1"` directly to the SQL string. Both work, just inconsistent.

PR #7829, opened 2026-06-17
Correct over-sanitization of user-agent string in whos_online

### LINKED ISSUE #7828: Who's Online oversanitization of user agent
Introduced in PR #7304, a "normal" user-agent now starts as
```
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36
```
and gets converted to
```
User Agent: Mozilla~5~0~Windows~NT~10~0~Win64~x64~AppleWebKit~537~36~KHTML~like~Gecko~Chrome~149~0~0~0~Safari~537~36
```
... making it "very difficult" to look those user-agents up on the web.

I'll suggest modifying the `preg_replace` on line 35
https://github.com/zencart/zencart/blob/a317c21900c3b9c008485813d5aa4f14124ab9cd/includes/functions/whos_online.php#L33-L36

to read
```php
    $wo_user_agent = preg_replace('#[^\w\/\(\)\. ;,]+#u', '~', $_SERVER['HTTP_USER_AGENT'] ?? '');
```

That replaces all non-words and characters not in `/().;,` with the `~`.

lat9:
Fixes #7828

### DIFF
--- a/includes/functions/whos_online.php
+++ b/includes/functions/whos_online.php
@@ -32,7 +32,7 @@ function zen_update_whos_online()
 
     $wo_session_id = zen_session_id();
     $wo_ip_address = substr(isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : 'Unknown', 0, 45);
-    $wo_user_agent = preg_replace('#[^\w\/\(\)\. ;,:\-]+#u', '~', $_SERVER['HTTP_USER_AGENT'] ?? '');
+    $wo_user_agent = preg_replace('#[^\w\/\(\)\.,; :+\-]+#u]', '~', $_SERVER['HTTP_USER_AGENT'] ?? '');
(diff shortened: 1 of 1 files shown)


lat9:
@piloujp, could you verify that this change doesn't break your multi-byte change?

piloujp:
> @piloujp, could you verify that this change doesn't break your multi-byte change?

No problem, as long as the `u` tag is there at the end of the regex it will accept all Unicode "words". Without it, it would only keep alphanumerics.

drbyte:
Proposing inclusion of `-` and `:`:
`$wo_user_agent = preg_replace('#[^\w\/\(\)\.,; :+\-]+#u', ' ', $_SERVER['HTTP_USER_AGENT'] ?? '');`

piloujp:
There is a `]` at the end that should not be there.

PR #7845, opened 2026-06-20
Remove code that's redundant

drbyte:
The implementation of `zen_config()` exposed that this is dead code.

`zen_get_configuration_key_value_layout()` does a lookup from `product_type_layout` table for the specified key and prod-id. But this `PRODUCT_INFO_PREVIOUS_NEXT_SORT` key is not one that's part of the product-type-layout table at all. It's only in the `configuration` table, so the subsequent call to `zen_products_sort_order()` (which checks PRODUCT_INFO_PREVIOUS_NEXT_SORT) is sufficient on its own.
Further, PRODUCT_INFO_PREVIOUS_NEXT_SORT is part of a fresh install, so is always present, which means this whole lookup was pointless in the first place.

Hence removing the redundant query.

### DIFF
--- a/admin/includes/modules/products_previous_next.php
+++ b/admin/includes/modules/products_previous_next.php
@@ -20,12 +20,6 @@ if (!isset($current_category_id)) {
 if (!isset($prev_next_list) || $prev_next_list == '') {
 // calculate the previous and next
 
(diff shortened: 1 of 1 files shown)


drbyte:
@codex review

PR #7853, opened 2026-06-21
Extract scalar cast for testability

drbyte:
Extracting this logic to a separate function so its safeguard can be tested in isolation.


zcwilt:
@drbyte 
is there any chance of the json containing bare true/false values. 
If so the stringify in zen_normalize_scalar_template_settings() will cast them to "1"/""

drbyte:
@zcwilt yes, in theory; "not yet" in actual practice.

Pushed an update.

zcwilt (approved):

PR #7856, opened 2026-06-21
Fix PHPUnit process-isolation gaps causing fatal redeclare in Unit suite

drbyte:
## Summary

- Running the full `Unit/` test directory in a single PHPUnit invocation hits a fatal `Cannot redeclare function zen_href_link()`. Several test classes bootstrap admin-side and catalog-side `includes/functions/html_output.php` in the same shared process, and both files unconditionally declare `zen_href_link()` with different signatures. Added `#[RunTestsInSeparateProcesses]` to `CatalogUrlGenerationTest`, `AdminCatalogUrlGenerationTest`, `AdminUrlGenerationTest`, and `RenderedJscriptFrameworkTest` so each runs in its own process.
- `ResponsiveClassicPluginRegressionTest` declared its `zen_href_link()` stub in a bare top-level `namespace { }` block, which PHPUnit's file-collection phase executes eagerly in the shared parent process regardless of per-class isolation attributes. Moved the stub into a lazily-`require`d, un-namespaced fixture file (`fixtures/zen_href_link_stub.php`) so it only loads inside the test's own isolated process.
- Removed the orphaned `protected $preserveGlobalState = false;` property from 10 test classes that never paired it with `runTestInSeparateProcess`, left over from an earlier pass that converted properly-paired classes to the `#[RunTestsInSeparateProcesses]` attribute.

## Test plan

- [x] `vendor/bin/phpunit not_for_release/testFramework/Unit/` (whole directory, single process) runs to completion with no fatal error
- [x] `composer tests-unit` passes for all affected files

### DIFF
--- a/not_for_release/testFramework/Unit/fixtures/zen_href_link_stub.php
+++ b/not_for_release/testFramework/Unit/fixtures/zen_href_link_stub.php
@@ -1,0 +1,12 @@
+<?php
+/**
+ * @copyright Copyright 2003-2026 Zen Cart Development Team
+ * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
+ */
+
+if (!function_exists('zen_href_link')) {
+    function zen_href_link(string $filename, string $parameters = ''): string
+    {
+        return $filename . ($parameters === '' ? '' : '?' . $parameters);
+    }
+}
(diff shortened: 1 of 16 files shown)


zcwilt (approved):
Just a note. Using the composer scripts to run the unit tests will bypass the issues you are seeing as the scripts explicitly invoke the unit tests with the phpunit  --process-isolation flag.

However, I do think this PR is still correct as there are legitimate reasons to run the tests manually and removing the brittle shared-process behaviour is likely to  ward off some regressions going forward

PR #7857, opened 2026-06-22
Small updates to skip some tests in local-dev

drbyte:
A few tests fail locally because of an override in the local `db` name. This PR offers "one way" to accommodate that: skip them.
I suppose another way could be to test that the $output contains the actual override db name, instead of hardcoded 'db'.

Also skips mysql-missing test on Herd environment because it's compiled into the binary, so does a false fail.


drbyte:
I added 3 constants to the example file, because I think they're "legit". But we can back them out if they're not intended to ever be used.

zcwilt:
For the DB-name-related tests, I wouldn't skip them based on local environment.
I would make them deterministic instead. 
These tests are asserting runner output, and the framework already provides a way to avoid local overrides via ZC_TEST_ENV_FILE=/dev/null. 
I’d update the affected test commands to use that, or otherwise assert against the resolved DB base dynamically if the goal is to verify the configured value rather than specifically db.

I would also replace the Herd-specific MySQL skip with a capability-based check. 
The issue is not Herd itself, but whether php -n still has MySQL support compiled in. 
Skipping based on PHP_BINARY is too environment-specific and can miss the same behavior on other builds.

drbyte:
Valid points

zcwilt (approved):

PR #7864, opened 2026-06-22
Convert zen_get_configuration_key_value to zen_config

drbyte:
## Proposal

It seems `zen_get_configuration_key_value()` is almost the same as zen_config insomuch as it checks the configuration table for the key. zen_config (currently) first checks for a define though.

It's possible `zen_get_configuration_key_value()` might have been intended to "bypass cache" by doing a direct lookup. But I'm inclined to hold that this may be redundant nowadays.

Input welcome.

### DIFF
--- a/admin/includes/init_includes/init_errors.php
+++ b/admin/includes/init_includes/init_errors.php
@@ -150,10 +150,10 @@ if (!$demo_check->EOF) {
 }
 
 // Check that shipping/payment modules have been defined
-if (zen_get_configuration_key_value('MODULE_PAYMENT_INSTALLED') === '') {
+if (empty(zen_config('MODULE_PAYMENT_INSTALLED'))) {
     $messageStack->add(ERROR_PAYMENT_MODULES_NOT_DEFINED, 'caution');
 }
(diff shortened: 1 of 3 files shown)


admin/includes/init_includes/init_errors.php
```diff
@@ -150,10 +150,10 @@
 }
 
 // Check that shipping/payment modules have been defined
-if (zen_get_configuration_key_value('MODULE_PAYMENT_INSTALLED') === '') {
+if (empty(zen_config('MODULE_PAYMENT_INSTALLED'))) {
     $messageStack->add(ERROR_PAYMENT_MODULES_NOT_DEFINED, 'caution');
 }
-if (zen_get_configuration_key_value('MODULE_SHIPPING_INSTALLED') === '') {
+if (empty(zen_config('MODULE_SHIPPING_INSTALLED'))) {
```
lat9, line 156: What's semi-humorous here is that `zen_get_configuration_key_value` will **NEVER** return `''`!

lat9 (approved):

zcwilt (approved):

PR #7867, opened 2026-06-23
fix sniffer method called from paypalr

proseLA:
see:

https://www.zen-cart.com/showthread.php/229886-PayPal-RESTful-API-Payment-Module?p=1411725#post1411725

and

https://github.com/zencart/zencart/blob/46918bfe13e1210087fed77ffec3836b93fd6851/includes/classes/sniffer.php#L164

i hate paypal.


drbyte:
Thanks!

lat9:
Unfortunately, this repository isn't the "source of truth" for the paypalr payment module, it's https://github.com/zencart/paypalr

Issue #7870, opened 2026-06-23
admin session destroyed in dashboard by whos online widget


torvista:
With the current changes integrated into my site, no problems locally, but on hosting 8.4.21, refreshing the dashboard destroys the session and logs out.

> Request URI: /ADMIN/, IP address: 79.116.141.250, Language id not set
> #0 [internal function]: zen_debug_error_handler()
> #1 /ADMIN/includes/classes/WhosOnline.php(362): session_decode()
> #2 /ADMIN/includes/classes/WhosOnline.php(251): WhosOnline->inspectSessionCart()
> #3 /ADMIN/includes/classes/WhosOnline.php(154): WhosOnline->getStatusCode()
> #4 /ADMIN/includes/classes/WhosOnline.php(220): WhosOnline->retrieve()
> #5 /ADMIN/includes/modules/dashboard_widgets/WhosOnlineDashboardWidget.php(16): WhosOnline->getStats()
> #6 /ADMIN/index_dashboard.php(111): include('xxxx')
> #7 /ADMIN/index_dashboard.php(247): render_zone()
> #8 /ADMIN/home.php(27): require('xxxxº')
> #9 /ADMIN/index.php(16): require('xxxx')
> --> PHP Warning: session_decode(): Failed to decode session object. Session has been destroyed in /ADMIN/includes/classes/WhosOnline.php on line 362.

If I rename/hide the WhosOnline widget, that solves the problem (but unmasks a second, unrelated).

I've not had time to replicate this with a vanilla site, but will if this cannot be replicated.

lat9:
The same issue arises using `Tools :: Whos Online`.

torvista:
Indeed, but I can't reproduce it locally.

lat9:
The good bad news is that I can!

torvista:
last update works for me.

PR #7880, opened 2026-06-27
Fix admin session destroy in dashboard

zcwilt:
see #7870

Ensure that stored session can acually be decoded

### DIFF
--- a/admin/includes/classes/WhosOnline.php
+++ b/admin/includes/classes/WhosOnline.php
@@ -324,12 +324,7 @@ class WhosOnline extends base
             $session_data = $result->EOF === false ? trim($result->fields['session_data']) : '';
         }
 
-        if (str_starts_with($session_data, 'cart|O')) {
-            $session_data = base64_decode($session_data);
-        }
-        if (str_starts_with($session_data, 'cart|O')) {
-            $session_data = '';
-        }
+        $session_data = $this->normalizeSessionData($session_data);
 
         if (empty($session_data)) {
             return null;
@@ -359,8 +354,8 @@ class WhosOnline extends base
         $adminSession = session_encode();
         $backupSessionArray = $_SESSION;
 
-        if (session_decode($session_data) !== false) {
-            $cart = $_SESSION['cart'];
+        if ($this->decodeSessionData($session_data)) {
+            $cart = $_SESSION['cart'] ?? null;
             $currency = $_SESSION['currency'] ?? zen_config('DEFAULT_CURRENCY');
 
             if (is_object($cart) && isset($currency)) {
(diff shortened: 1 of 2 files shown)


zcwilt:
@codex review

zcwilt:
We are now running into a fundamental issue with session_decode()

If it succeeds the internal session engine is fine and we can happily reload a session array from our saved admin details.

If it fails, php destroys the session and internally flags the session as destroyed. Loading back some saved session data into $_SESSION at this point will not restore the session.

There are a couple of work rounds but both are hacky.

lat9:
How is that undecodable data getting into the `$session_data` in the first place?

zcwilt:
@lat9 

There are a number of reasons this might fail.

Storing the cart object in the session then storing that in the DB can fail if the cart object hasn't be correctly created, maybe some race condition in the request.

Changed settings based on the php version. PHP has an ini setting for how a session is serialized.

A request died at some point writing garbage into the session 

There's a very small change of a mysql collation issue.

BTW if you can still reproduce this :- below us some PHP/SQL i would like to see the results of 

```echo ini_get('session.serialize_handler');```

```
SELECT sesskey, OCTET_LENGTH(value), LEFT(value, 80)
FROM sessions
ORDER BY expiry DESC
LIMIT 20;
```

```
SHOW FULL COLUMNS FROM sessions;
```

```
SELECT
  s.sesskey,
  HEX(s.sesskey) AS sesskey_hex,
  w.session_id,
  HEX(w.session_id) AS whos_online_session_id_hex,
  OCTET_LENGTH(s.value) AS bytes,
  LEFT(CAST(s.value AS CHAR), 80) AS prefix
FROM whos_online w
LEFT JOIN sessions s ON s.sesskey = w.session_id
WHERE w.session_id <> ''
ORDER BY w.time_last_click DESC
LIMIT 20;
```

lat9:
`session.serialize_handler` = `php`.

... and the SQL query results:
image

lat9:
Note that I've not applied the changes you've posted for this PR.

lat9:
Once I applied your changes, the admin session was no longer kicked out when refreshing the `Tools :: Who's Online` page

zcwilt:
So, glad it fixed the issues.

The results you posted at least confirm that all the data is in the right place, and looks consistent

I cant decipher some of the session data as it is truncated, so cant tell which entry is causing the error.

A clean dump of you session table would help, as long as nothing is truncated. 

Send privately if you want.

zcwilt:
@codex review

zcwilt:
While the review is strictly correct the fix is more trouble than its worth

Time would be better spent moving to stop using session_decode to get who's online info 
and use a proper request snapshot to  hold who's online info

zcwilt:
@codex review

PR #7881, opened 2026-06-27
POSM: jQuery selectors should use `$tplSetting` ...

lat9:
... instead of `zen_config`, since those settings can be template-dependent.


drbyte:
@codex review

PR #7882, opened 2026-06-27
Fix: Additional images not copied when copying a product

### LINKED ISSUE #7877: Additional images not copied when copying a product
As reported on the Zen Cart forums: https://www.zen-cart.com/showthread.php/231137-Additional-images-not-copied-over-when-copying-a-product?p=1411740

drbyte:
When a store is using the (now default) database mode for tracking "additional image" associations with a product, the "Copy/Clone a product" wasn't copying those images. This update adds the option to also associate those images with the new product clone. The actual files are not copied, but database records are added to co-associate the same filenames with the new product ID.

Fixes #7877


drbyte:
@codex review

PR #7885, opened 2026-06-29
Standardize on DIR_FS_UPLOADS

drbyte:
For consistency between catalog/admin.

Catalog already uses this constant. No reason to use a different one in admin when referencing it. 
Makes for easier future maintenance or relocation.

### DIFF
--- a/admin/orders.php
+++ b/admin/orders.php
@@ -182,7 +182,7 @@ if (!empty($action) && $order_exists === true) {
                                 zen_redirect(zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(['download', 'action']) . 'action=edit', 'NONSSL'));
                 }
             // this define was added to admin in v3.0 for consistency with catalog; added here just as a safeguard: remove in v4 or later, since it's in defined_paths already.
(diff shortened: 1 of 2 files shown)


admin/orders.php
```diff
@@ -181,7 +181,9 @@
                                 $messageStack->add_session(sprintf(TEXT_EXTENSION_NOT_UNDERSTOOD, $file_extension), 'error');
                                 zen_redirect(zen_href_link(FILENAME_ORDERS, zen_get_all_get_params(['download', 'action']) . 'action=edit', 'NONSSL'));
                 }
-            $fs_path = DIR_FS_CATALOG_IMAGES . 'uploads/' . $fileName;
+            // this define was added to admin in v3.0 for consistency with catalog; added here just as a safeguard: remove in v4 or later, since it's in defined_paths already.
+            defined('DIR_FS_UPLOADS') || define('DIR_FS_UPLOADS', DIR_FS_CATALOG_IMAGES . 'uploads/');
```
lat9: Why not just use `zen_define_default`?

PR #7888, opened 2026-06-30
Clamp coupon min/max limits

drbyte:
These min/max clamps help avoid accidental unintended side-effects.
- percentages must be less than or equal to 100
- uses must be 0 or greater
- min order must be 0 or greater

### DIFF
--- a/admin/coupon_admin.php
+++ b/admin/coupon_admin.php
@@ -284,6 +284,7 @@ switch ($_GET['action']) {
       $coupon_type = 'E'; // percentage off and free shipping
     }
     $_POST['coupon_amount'] = preg_replace('/[^0-9.]/', '', $_POST['coupon_amount']);
+    // Clamp percentages ('P' and 'E' types, see comments above) to under 100.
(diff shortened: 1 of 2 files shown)


admin/coupon_admin.php
```diff
@@ -284,14 +284,17 @@
       $coupon_type = 'E'; // percentage off and free shipping
     }
     $_POST['coupon_amount'] = preg_replace('/[^0-9.]/', '', $_POST['coupon_amount']);
+    if ($coupon_type === 'P' || $coupon_type === 'E') {
+      $_POST['coupon_amount'] = (string)min(100, (float)$_POST['coupon_amount']);
+    }
```
lat9: A comment here indicating that since the coupon is a percent-off &mdash; 'P' (percentage off) and 'E' (percentage off, plus free shipping) &mdash; would help.  Until I looked up with those 'codes' mean, I was trying to figure out why a coupon value should be capped at $100.00!

lat9 (changes requested):

PR #7891, opened 2026-06-30
Change $allowedMethods to a constant

drbyte:


lat9 (approved):
I just made that teeny change to the POSM's admin AJAX, noting the last-version updated.

Note, too, that anyone that wants to grab POSM's 6.1.2 version, needs to be aware that the minimum PHP version is now 8.3; that's when typed constants were added.

PR #7894, opened 2026-07-01
2.2: Adding POSM v6.1.2

### LINKED ISSUE #7862: 2.2: Backport POSM v6.1.2
It's got significant changes since the v6.1.1 included in zc222.

lat9:
Fixes #7862


zc_plugins/POSM/v6.1.2/catalog/includes/classes/ajax/zcAjaxOptionsStockDependencies.php
```diff
@@ -0,0 +1,175 @@
...
+// Part of the "Products' Options' Stock Manager" plugin by Cindy Merkin (lat9)
+// Copyright (c) 2014-2026, Vinos de Frutas Tropicales
+//
+// Last updated: POSM v6.1.2
+//
+// Uses the AJAX infrastructure provided initially by Zen Cart v1.5.4 and updated in Zen Cart 1.5.5b
+//
+class zcAjaxOptionsStockDependencies extends base
+{
+    public const array ALLOWED_METHODS = ['availableOptionValues'];
```
drbyte: Is this syntax compatible with ZC 2.2 which supports slightly older PHP? (and this PR is against v2.2)

lat9: Nope, I commented (somewhere) indicating that the typed constant syntax was introduced in PHP 8.3.  If we remove the `array` type here, either the version of the zc300 version needs to be updated to 6.1.3 (keeping the `array` typing) or the `array` type needs to be removed in POSM 6.1.2.

drbyte: Right. Removing `array` would make it compatible back to at least PHP 8.0, which is the ZC 2.0 min (not that we're targeting that, but just stating for clarity).

Noting that this is the `2.2` branch, for which we state PHP 8.2 minimum.

drbyte: Also noting that IDE/static-analysis tools would benefit from a docblock notation, to help prevent userland edits from being changed away from array:
```
/** @var list<string> */
public const ALLOWED_METHODS = ['availableOptionValues'];
```
Technically this same docblock would still be relevant "with" the `array` keyword, as well.

Could alternatively use any of:
`/** @var list<string> */`
`/** @var string[] */`
`/** @var array<int, string> */`
All of these mean the same thing for this sequential array of strings.

lat9: OK, let me get a PR going for the base to update 6.1.2 to remove that array and then update this 2.2 submission to mimic that change.

PR #7899, opened 2026-07-02
Provide a soft setting to disable the display of "Sort by" dropdown

scottcwilson:
Screenshot 2026-07-02 at 12 29 38 PM

### DIFF
--- a/includes/templates/template_default/templates/tpl_modules_listing_display_order.php
+++ b/includes/templates/template_default/templates/tpl_modules_listing_display_order.php
@@ -7,6 +7,9 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: lat9 2026 Feb 22 Modified in v2.2.1 $
  */
+if (isset($show_sort_by_on_product_listing) && $show_sort_by_on_product_listing === false) { 
+    return; 
+} 
 $disp_order = (int)($disp_order ?? 0);
 $disp_order_default = (int)($disp_order_default ?? 0);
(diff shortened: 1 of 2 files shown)


includes/templates/template_default/templates/tpl_modules_listing_display_order.php
```diff
@@ -25,6 +25,14 @@
 }
 echo zen_hide_session_id();
 ?>
+<?php 
+$template_show_sort_by = true; 
+if (isset($show_sort_by_on_product_listing) && $show_sort_by_on_product_listing == false) {
```
lat9: Suggestion: Put this at the top of the file and just do a quick return if the conditional is met. That way, there's no empty form element created on the page.

drbyte: I like that suggestion. That could even eliminate the temporary variable too.

includes/templates/template_default/templates/tpl_modules_listing_display_order.php
```diff
@@ -7,6 +7,9 @@
  * @license http://www.zen-cart.com/license/2_0.txt GNU Public License V2.0
  * @version $Id: lat9 2026 Feb 22 Modified in v2.2.1 $
  */
+if (isset($show_sort_by_on_product_listing) && $show_sort_by_on_product_listing == false) {
```
drbyte: This should probably be `===` instead of `==`.
Otherwise, I approve.

drbyte (approved):

PR #7901, opened 2026-07-03
Don't treat arrays as repeated keys in sanitization.

cstdenis:
Don't treat arrays as repeated keys (fixes my search mod that allows multiple categories/manufacturers to be selected)


lat9:
Your search mod should be using a `'post'` type form rather than a `'get'` type form if it supports data arrays. `$_GET` arrays have not been supported since around Zen Cart 1.5.0.

PR #7906, opened 2026-07-07
Refinements to zc_install

drbyte:
Stricter handling of form submissions during install


zc_install/includes/template/templates/index_default.php
```diff
@@ -35,7 +35,7 @@
 
     <?php
     } else { ?>
-        <input type="hidden" name="adminDir" value="<?= $selectedAdminDir ?>">
+        <input type="hidden" name="adminDir" value="<?= zc_install_escape_html($selectedAdminDir) ?>">
```
zcwilt, line 38: we already sanitize this in header_php.php 
won't this will lead to double encoding

drbyte: Yes; however, there's no legit admin dir name that should have ampersands, is there?
But I'm committing an update for specificity.

PR #7912, opened 2026-07-09
2.3: Correct over-sanitization of configuration title

lat9:
Introduced in this PR #7793

Results in configuration titles with HTML tags getting over-sanitized; noting that there's no such sanitization on a configuration entry's description.

image


drbyte:
Does the messageStack on line 87 also need the change?

PR #7914, opened 2026-07-11
Add screen reader markup for hidden fields

scottcwilson:


drbyte:
`display:none` already hides this from screen readers and removes it from tab order, so `aria-hidden` adds nothing here.

PR #7915, opened 2026-07-11
allow plugins to use product type modules

zcwilt:


zcwilt:
@codex review

includes/functions/functions_files.php
```diff
@@ -261,6 +261,28 @@ function zen_get_admin_module_from_directory(int $product_type, string $filename
...
+
+            $moduleDir = $pluginDir . '/admin/includes/modules/' . $product_type_foldername . '/';
+            $realModuleFile = realpath($moduleDir . $filename_to_check);
+            if ($realModuleFile === false || strpos($realModuleFile, $resolvedPluginDir) !== 0) {
+                continue;
+            }
+
+            $dir = $moduleDir;
+            break;
+        }
```
lat9, line 295: I'm not sure what precedence is desired, but this implementation allows a plugin to
- Override the 'product' `collect_info.php`, since that file doesn't reside in `admin/includes/modules/product`.
- Supply a new product type's various files, since none will be found in `admin/includes/modules/new-product-type`.

It does not allow a plugin to override the `collect_info.php` for a built-in product of type **not** 'product', since `admin/includes/modules/non-product-type/collect_info.php` exists.  Ditto for some of the "common" product actions; see the `document_general` and `product_music` subdirectories.

zcwilt:
@lat9 
yes, the itch i needed scratching was just to allow a new product type via a plugin. 

not sure if allowing more than that opens a can of worms that needs some more thought.

drbyte:
Yes, maybe this needs to also check whether the core admin module product-type **dir** exists, not just whether a collect_info/whatever file exists within that dir. That way it locks out core product-type overwrites.

includes/functions/functions_files.php
```diff
@@ -259,13 +259,23 @@ function zen_get_admin_module_from_directory(int $product_type, string $filename
...
+        // fallback is only offered below for types core has no directory for at all.
+        if (file_exists($coreTypeDir . $filename_to_check)) {
+            $dir = $coreTypeDir;
+        }
     } else {
-        // Not shipped in core's own admin/includes/modules/ -- check installed plugins for
-        // an admin/includes/modules/<type_handler>/ directory of their own, so a plugin that
-        // registers a product type doesn't need its admin module files physically copied
-        // into DIR_WS_MODULES. Mirrors Zencart\FileSystem\FileSystem::findPluginAdminPage().
+        // Not a product type core ships an admin/includes/modules/ subfolder for at all --
```
drbyte: I'm confused by the new wording on this comment.

drbyte (approved):
One observation: unlike `findPluginAdminPage()` (which loops over all plugins and lets the `last match win`), this new code breaks on the `first matching plugin`. Functionally more predictable, but it's a slight deviation and not a true mirror of that function's resolution order ... which is explainable, and quite intentional.

PR #7922, opened 2026-07-18
Cast admin reviews page number to int for splitPageResults

### LINKED ISSUE #7916: Admin, reviews, multiple pages of results: PHP Fatal error: Uncaught TypeError: splitPageResults::
PHP Fatal error:  Uncaught TypeError: splitPageResults::__construct(): Argument #1 ($current_page_number) must be of type string|int|null, float given, called in admin\reviews.php on line 391 and defined in admin\includes\classes\split_page_results.php:57

Comes from the calculation of $current_page.

https://github.com/zencart/zencart/blob/34087abcdd97eb6dcc49feb703c27877b831d536/admin/reviews.php#L386-L391

SAY-5:
**Describe what is being fixed/changed and why**
Loading the admin Reviews list with more than one page of results, when a review's `rID` is present in the URL, crashes with:

```
PHP Fatal error: Uncaught TypeError: splitPageResults::__construct(): Argument #1 ($current_page_number) must be of type string|int|null, float given, called in admin/reviews.php on line 391
```

When resetting to the page that contains the requested review, `$currentPage` is computed with `round(..., 0)`. In PHP `round()` returns a `float`, so `$currentPage` becomes a float (e.g. `2.0`). It is then passed by reference into `splitPageResults::__construct(int|string|null &$current_page_number, ...)`, which rejects a float under `declare(strict_types=1)` and throws the fatal above. Casting the result to `(int)` at assignment keeps the value in the type the constructor accepts, and it is also the correct type for the later `display_count()`/`display_links()` calls that reuse `$currentPage`.

Fixes #7916.

**To Reproduce**
1. Have more reviews than `MAX_DISPLAY_SEARCH_RESULTS` so results span multiple pages.
2. In admin, open Customers > Reviews and click into a review far enough down the list that it is not on page 1.
3. The list reloads with `rID` in the URL and the page reset logic runs.
4. See the fatal `TypeError` from `splitPageResults::__construct()`.

**Screenshots**
N/A (backend fatal, no display change).

**Testing**
- `php -l admin/reviews.php` passes.
- Reproduced the exact `TypeError` in isolation by passing the `round()` result (a float) to a constructor typed `int|string|null`, then confirmed that casting to `(int)` is accepted.
- The value flows unchanged into the existing `page=` / pagination logic, which already `(int)`-casts it.

**AI Tools Used**

### DIFF
--- a/admin/reviews.php
+++ b/admin/reviews.php
@@ -1,4 +1,6 @@
 <?php
+declare(strict_types=1);
+
 /**
  * @copyright Copyright 2003-2025 Zen Cart Development Team
  * @copyright Portions Copyright 2003 osCommerce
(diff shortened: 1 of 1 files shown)


drbyte (commented):
For the reported line, this is probably the best fix for strict_types=1 mode.
But lines 534 and 537 also need casting to int for their zen_config calls, else the same issue will happen there.

PR #7928, opened 2026-07-19
date ranges for statistics pages

zenexpert:
minor improvement for stats pages - instead of showing all-time best selling products and all-time best customers, it allow to define a date range. This way, admin can see best selling products for the past month, or last 2 weeks or whatever period they're interested in. Same option for customers - admin can see which customer spent most in a selected period of time.


drbyte:
Some of copilot's comments pre-date this PR. Will fix those separately.

PR #7924, opened 2026-07-18
2.3: Prevent cart-flooding and excess queries

drbyte:
Sets a hard cap on max cart products per checkout. 
Skips unnecessary category queries on pages that don't need them.

Fixes #7921

PR #7932, opened 2026-07-21
Add routing map inoculation to dismiss rogue traffic quickly

drbyte:
Rejects visits containing malformed URLs mixing GET params with pages that don't use them. 

Applies cleanly to older Zen Cart versions, if done by hand. (the application_top change just needs to be before the current 406 header loop, like shown in this commit's diff)

Should also apply https://github.com/zencart/zencart/pull/7924 for more thorough protection. 
Both this and that PR are compatible back to PHP 7.1+

### DIFF
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -207,6 +207,16 @@ if (!$contaminated && isset($_GET['action']) && $_GET['action'] === 'buy_now') {
     }
 }
 
+/**
+ * reject catalog-filter params (cPath, manufacturers_id, etc.) when supplied
+ * for a page that never legitimately reads them, such as bots stuffing shopping_cart
+ * or other non-catalog pages with spoofed query strings. See includes/routing_map.php.
+ */
+if (!$contaminated) {
+    require_once __DIR__ . '/routing_map.php';
+    $contaminated = zen_request_has_disallowed_catalog_param($_GET);
+}
+
 if ($contaminated) {
     header('HTTP/1.1 406 Not Acceptable');
     exit(0);
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -208,9 +208,10 @@ if (!$contaminated && isset($_GET['action']) && $_GET['action'] === 'buy_now') {
 }
 
 /**
- * reject catalog-filter params (cPath, manufacturers_id, etc.) when supplied
- * for a page that never legitimately reads them, such as bots stuffing shopping_cart
- * or other non-catalog pages with spoofed query strings. See includes/routing_map.php.
+ * reject catalog-filter params (manufacturers_id, sort, etc.) when supplied for
+ * a page confirmed to never legitimately read them, such as bots stuffing
+ * shopping_cart or checkout with spoofed query strings. cPath and products_id are
+ * deliberately NOT covered here -- see includes/routing_map.php.
  */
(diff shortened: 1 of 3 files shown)


torvista:
For my interest, what is prompting this and the other related PR at this time?
The problem is very topical for me, seeming to form part of the constant overloading issue I've had for a while but has recently gotten much worse.

drbyte:
@torvista 
https://www.zen-cart.com/showthread.php/231161-Bots-with-Fake-carts
https://www.zen-cart.com/showthread.php/231150-abnormal-visitor-numbers

includes/application_top.php
```diff
@@ -207,6 +207,17 @@
...
 }
 
+/**
+ * reject catalog-filter params (manufacturers_id, sort, etc.) when supplied for
+ * a page confirmed to never legitimately read them, such as bots stuffing
+ * shopping_cart or checkout with spoofed query strings. cPath and products_id are
+ * deliberately NOT covered here -- see includes/routing_map.php.
+ */
+if (!$contaminated) {
+    require_once __DIR__ . '/routing_map.php';
```
lat9, line 217: Why `require_once` instead of `require`?

includes/routing_map.php
```diff
@@ -0,0 +1,92 @@
...
+
+    // Mirrors init_sanitize.php's own fallback:
+    // an empty/missing main_page defaults to the index page (FILENAME_DEFAULT)
+    // e.g. index.php?manufacturers_id=8 is a legitimate URL shape that renders as the index/manufacturer listing
+    $page = $get['main_page'] ?? '';
+    if ($page === '') {
+        $page = 'index';
+    }
+
+    if (!in_array($page, $knownNonCatalogPages, true)) {
```
lat9: Could this be extended to allow "other" checkout pages
```suggestion
    if ($page !== 'shopping_cart' && !str_starts_with($page, 'checkout_')) {
```

That would enable the `$knownNonCatalogPages` to be removed, too.

lat9 (changes requested):

includes/application_top.php
drbyte: 2 reasons: defensive against a mistake that tries to load it later by someone changing things in future,
and the same convention is used a few lines below in this same file.

mprough:
I used this for Zen 2.1.0 & PHP 8.3. It works, but also block adding to cart via the buy now button

lat9:
> I used this for Zen 2.1.0 & PHP 8.3. It works, but also block adding to cart via the buy now button

Oooh!  Good catch, @mprough!

drbyte:
@mprough I've reverted the early shopping-cart block, deferring its checks to later downstream.

lat9 (approved):

mprough:
I have a store 2.1.0 PHP 8.3 the active carts grows every day. Today 450. The automated requests are specifically targeting the shopping cart endpoint by sending requests such as:

/index.php?main_page=shopping_cart&products_id=...

from a very large number of different IP addresses. Since the traffic is highly distributed, there isn't a server-side rate limit or firewall rule that can reliably distinguish these requests from legitimate visitors without risking false positives.

This patch has only seen the active 1 item carts grow from 150 ish several days ago to 450 today

Issue #7936, opened 2026-07-24
`order_total::pre_confirmation_check` type-hints, remove `string` and add `int|float`


proseLA:
executing a line like this:
```php
$calculatedTotal = $order_total_modules->pre_confirmation_check(true);
```
returns an error:
```
[22-Jul-2026 23:01:53 UTC] PHP Fatal error:  Uncaught TypeError: order_total::pre_confirmation_check(): Return value must be of type array|string|null, float returned in /var/www/zcdev/includes/classes/order_total.php:285
```

not sure why the order total would be a string.

seems like we need to add float here on line 262:

https://github.com/zencart/zencart/blob/76be468d39cdec95d7153aae41cab44709ef284e/includes/classes/order_total.php#L262-L286

lat9:
Just "for the record", what order-totals are in use when this error occurs?

proseLA:
<img width="1709" height="690" alt="Image" src="https://github.com/user-attachments/assets/81bf2269-7661-4aa7-a9a0-adb033221c03" />

lat9:
That function-return type-hint was added here: https://github.com/zencart/zencart/commit/29566124ad35fcb8818998f33f28c4022cf5e0d6

PR #7937, opened 2026-07-24
Updating ot `pre_confirmation_check` type-declarations

lat9:
Ref #7936

### DIFF
--- a/includes/classes/order_total.php
+++ b/includes/classes/order_total.php
@@ -259,7 +259,7 @@ class order_total
      *
      * @since ZC v1.0.3
      */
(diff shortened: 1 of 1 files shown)


proseLA:
"...In modern PHP, "type hinting" and "type declarations" refer to the exact same runtime feature, **but "_type declarations_" is the official terminology** used by the [PHP Manual](https://www.php.net/manual/en/language.types.declarations.php). Historically, the term "type hinting" was used when the feature was first introduced in PHP 5, while "type declaration" became preferred as the system evolved into a complete, enforceable type system in PHP 7 and PHP 8."

PR #7934, opened 2026-07-22
strict cPath validation

zenexpert:
fixes #7903

PR #7943, opened 2026-08-04
Normalize malformed category and product cPaths

### LINKED ISSUE #7903: infinite cPath loop on malformed URLs
If a URL containing cPath is malformed either by malicious intents, bad code or just some bots, Zen Cart allows this URL and shows content, together with a canonical tag that doesn't help. Bots can get stuck in endless loops, causing massive bandwidth and resource usage. We're also allowing duplicate content because the canonical link is not corrected.

On a clean install, simply try visiting `index.php?main_page=index&cPath=23_22_21_3` - all of these are top level categories, but you'll end up getting the contents of the last in the row (3), but breadcrumbs will be incorrect, as well as "active" categories in the sidebox. 

Image

drbyte:
Redirect active category requests to the authoritative generated category path.

For product pages, reuse the loaded Product data to preserve a requested category only when the product is directly linked to it. Otherwise, redirect to the product's master-category path.

Retain applicable request parameters on the permanent redirect, sourced from the parameter policy init_canonical already applied to this request, and drop the invalid cPath, cart actions, and any session value.

Add regression coverage for the path resolution (master, linked, unrelated, invalid, and missing category paths) and for the redirect itself.

Fixes #7903 and replaces/closes #7934

## Why this resolves #7903

- A malformed category URL such as `cPath=23_22_21_3` is compared with the authoritative path   generated for category 3. It receives a 301 redirect instead of serving category 3 under the   false hierarchy.
- Product URLs are resolved against the product's actual direct category links. A legitimate   linked-category path is preserved; an unrelated or malformed path falls back to the master   category.
- After the redirect the URL, breadcrumbs, and active sidebox categories all describe the same   hierarchy, and for category listings the canonical link — which echoes the requested cPath,   the "canonical tag that doesn't help" from the report — now describes it too.
- Crawlers are directed to one stable URL rather than being allowed to discover unlimited cPath   permutations containing identical content.
- `zen_get_valid_cpath_for_product()` reuses `Product::$data`, so master and invalid-category cases add no database queries. Only a legitimate alternate linked category requires its parent path to be generated.
- Centralizing product validation in `zen_product_set_header_response()` avoids five copied header blocks and covers product types using the standard response flow.

## Redirect behavior

- Retained parameters come from init_canonical's `$excludeParams`, so the 301 carries over only  what the store already treats as part of a URL's identity, and unrecognized parameters — which init_canonical has already appended to that list — are never reflected into a cacheable `Location` header. Plugins adjust this through the existing  `NOTIFY_INIT_CANONICAL_PARAM_WHITELIST`. If init_canonical was overridden away, the redirect fails closed and retains nothing.
- No session id is appended, which would otherwise be baked into a permanently-cached and indexable response when cookies are disabled.
- Non-GET requests are not redirected, since a 301 causes browsers to re-issue as GET and silently discard the request body.
- The target is built from the sanitized `$current_page_base`, not the raw `main_page` value.
- On the index page the check runs inside the block guarded by `$category_redirect_handled`, so observers of `NOTIFY_INDEX_CATEGORY_STATUS_CHECK` keep their override, and the category depth is resolved first so the page stays displayable if an observer of `NOTIFY_ZEN_REDIRECT` cancels  the redirect.
- `zen_get_generated_category_path_rev()` is memoized per request on the storefront: the ancestor walk is one uncached query per level, and the storefront now asks for the same path from the canonical link, the cPath validation, and the sideboxes. Not memoized in admin, where the tree can change mid-request.

## Not addressed here

Product canonical links exclude cPath by default (`$includeCPath = false`), and when included are generated from the master category. A product URL 301'd to a preserved linked-category path therefore still lands on a page whose canonical describes a different path. That predates this change and is out of scope for #7903, which reported the category-listing case.

### DIFF
--- a/includes/functions/functions_categories.php
+++ b/includes/functions/functions_categories.php
@@ -265,6 +265,39 @@ function zen_get_product_path($product_id): string
 }
 
 /**
+ * Return the authoritative category path for a product, preserving a requested category
+ * only when the product is directly linked to that category.
+ *
+ * Unlike zen_get_product_path(), this uses the already-loaded Product data, so the
+ * master-category and unrelated-category cases add no queries; only a legitimate alternate
+ * linked category needs its path generated.
+ *
+ * @param Product $product_info the loaded product
+ * @param int $requested_category_id the category asked for (0 when the request had no usable cPath)
+ * @return string authoritative cPath, or '' when the product has no master category
+ * @since ZC v2.3.0
+ */
+function zen_get_valid_cpath_for_product(Product $product_info, int $requested_category_id): string
+{
+    $master_category_id = (int)$product_info->get('master_categories_id');
+    $master_cPath = $master_category_id > 0 ? (string)$product_info->get('cPath') : '';
+
+    if ($requested_category_id === $master_category_id) {
+        return $master_cPath;
+    }
+
+    $linked_categories = $product_info->get('linked_categories');
+    if ($requested_category_id > 0 && is_array($linked_categories)) {
+        $linked_categories = array_map('intval', $linked_categories);
+        if (in_array($requested_category_id, $linked_categories, true)) {
+            return zen_get_generated_category_path_rev($requested_category_id);
+        }
+    }
+
+    return $master_cPath;
+}
+
+/**
  * Parse and sanitize the cPath parameter values
  * @param string $cPath
  * @return array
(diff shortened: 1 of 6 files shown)


drbyte:
Hmmm ... Product redirects fight the canonical tag. Have to explore further. Need to avoid confusion over generated links.

PR #7950, opened 2026-08-09
Remove Possibility of Infinite loop

brittainmark:
This fixes the issue raised by @bumba000 [https://www.zen-cart.com/showthread.php/231165-Frequent-Duplicate-Orders-Cart-Not-Emptying?p=1412070#post1412070](https://www.zen-cart.com/showthread.php/231165-Frequent-Duplicate-Orders-Cart-Not-Emptying?p=1412070#post1412070). Additionally, I have tidied the function to make it a bit more efficient.

Remove while loop calling str_replace as this could be infinite loop if the user specifies same value for search and replace.

Tidy function

- Change $zen_fix_current and $zen_fix_replace to arrays so only one call to str_replace required
- Added special character into above array to remove second call to str_replace.

### DIFF
--- a/includes/classes/Email.php
+++ b/includes/classes/Email.php
@@ -313,28 +313,32 @@ class Email
         // CLEAN UP CHARACTERS
         $email_text = preg_replace('/((&amp;)|&)+/', '&', $email_text);
 
+        $zen_fix_current = [];
+        $zen_fix_replace = [];
+
         if (!empty(zen_config('CURRENCIES_TRANSLATIONS'))) {
+            // Add curency conversions if any
             $zen_fix_currencies = preg_split("/[:,]/", str_replace(' ', '', zen_config('CURRENCIES_TRANSLATIONS')));
             $size = count($zen_fix_currencies);
+            
+            // Convert in to array
             for ($i = 0, $n = $size; $i < $n; $i += 2) {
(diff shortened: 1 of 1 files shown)


drbyte:
While I accept the replacement of `while(str_contains)`, and the parallel arrays for `str_replace`, I think the tag-stripping rewrite is problematic, and probably should be done in a separate PR.

piloujp:
The original regular expression looks wrong to me. What was it supposed to do? Replace "<" in tags that are not listed so they are not stripped away by `strip_tags`?

brittainmark:
I thought the regular expression was incorrect as well which was why I changed it to make it clearer what was being done.

I beleive the regex as it is will prevent striping of nearly all html tag names as the negation followed by ( just looks at each individual character inside the brackets which means that it will not strip \<table> or the \<hr> tag as the reg ex contains the letter t and h,

 I am happy to resubmit without this and submit a second pr for the strip tags. Please confirm that is what you would like.

brittainmark:
Strip tags reverted. I'll submit a new pr when this is approved.

scottcwilson:
Thanks for grabbing this issue @brittainmark.

drbyte:
I agree the regex was wrong. 👀

PR #7951, opened 2026-08-11
Fix tag stripping in email text sanitization

drbyte:
Fix tag stripping in email text sanitization

## Summary

`Email::sanitizeTextContent()` builds the `text/plain` part of every outgoing email. Its tag-stripping step behaves inconsistently: whether a tag is removed depends on its first letter and its letter case, and closing tags come back with the slash dropped (`</div>` → `<div>`).

This replaces the regex with the alternation it was always meant to be. No new dependencies, no change to the surrounding pipeline.

Narrower revisit of the tag-stripping hunk originally proposed in #7950, which was dropped from that PR before merge. The infinite-loop fix from #7950 is unaffected.

## Why the old code looks the way it does

Worth reading before reviewing the regex — it is **not** an allowlist of "safe tags to keep". It exists to tell a real HTML tag apart from a literal `<` in content, so `strip_tags()` does not eat a product name like `Widget <Large>`.

- **`1a48ad0b1` (2013-07-23)** — CHANGE-417, *"email confirmation gets truncated on the `<` symbol in product names"*. First form protected `<` when not followed by a letter: `preg_replace("/<([^[:alpha:]])/", '@lt@\\1', ...)`. That does not fix the reported case, since `<Large>` starts with a letter.
- **`72d63b9df` (2013-07-23)** — amended before landing, and the version that reached 2.0–2.3 and master. Swapped `[:alpha:]` for a class built from tag names, expressing "recognise these as tags; treat every other `<` as content." This version worked for the reported case.
- **`c04e0b74b` (2013-11-06)** — *"fix problem leftover in welcome emails"*: added `li|ol|ul` and changed `<` to `</?`. This is where it broke. Adding `li|ol|ul` put the letter `l` into the character class, so `<large>` stopped being protected; and the `</?` capture keeps only the character after `</`, which is why closing tags lose their slash.

The root flaw dates from July: PCRE reads `[^(strong>|br ?\/?>|...)]` as a set of single characters, not a list of tag names.

## The change

```php
$email_text = preg_replace('~<(?!/?(?:' . self::TAGS_TO_STRIP . ')(?:[\s/][^<>]*)?>)~i', self::LT_PLACEHOLDER, $email_text);
$email_text = strip_tags($email_text);
$email_text = str_replace(self::LT_PLACEHOLDER, '<', $email_text);
```

The trailing `>` is load-bearing: the tag must be seen to close before it is trusted as markup. 
A bare `\b` would also match before punctuation, so `<BR-2032`, `<UL-94` and `<A/V` would read as tags and `strip_tags()` would delete to the next `>` — or to the end of the message.

## Behaviour changes

| input | before | after |
|---|---|---|
| `a<BR />b` | `a<BR />b` | `ab` |
| `<STRONG>x</STRONG>` | `<STRONG>x<STRONG>` | `x` |
| `<div class="x">div text</div>` | `<div class="x">div text<div>` | `<div class="x">div text</div>` |
| `<table><tr><td>cell</td></tr></table>` | `cell` | preserved verbatim |
| `<h1>Heading</h1>` | `Heading` | preserved verbatim |

Unchanged: lowercase `<strong>`, `<a>`, `<br>`, `<script>`, `<iframe>`, `` are stripped as before; `<Large>`, `<Small>` and `5 < 6` still survive.

**Note the last two rows** — because the old regex keyed off a tag's *first letter*, it deleted anything starting with `s t r o n g b a h e f p c i l u m`. Those tags now pass through as literal text. This is a widening of what can land in a plaintext body, not just de-mangling. Bounded in practice: core email language constants contain none of these, and `contact_us` already strip_tags at intake. It is a content-quality change — `$text` reaches only `Body`/`AltBody` and the archive table, so there is no rendering path in core.

## Note for reviewers: the tag list is a correctness constraint

`TAGS_TO_STRIP` is not a matter of taste. Anything listed is treated as markup; anything absent is content. `small`, `big`, `center` and `table` are real HTML tag names *and* ordinary English words, so adding them would truncate product names like `Widget <Small>` while `<Large>` kept working — a partial re-run of CHANGE-417. `testProductNameSizesInAngleBracketsAreNeverTruncated` guards this.

`img` is included because it wraps no readable content, so leaving it would put raw markup in a plaintext body; current master already removes it.

Open to argument on adding the unambiguous structural tags (`svg`, `form`, `input`, `h1`–`h6`, `select`, `textarea`, `meta`, `pre`, `blockquote`) to restore the stripping master did by accident.

## Testing

`Unit/testsEmail/` — 36 tests covering currency translations, entity cleanup, the html-fallback branch, disclaimers, and tag handling. Full Unit suite: 605 tests, 1539 assertions, 0 failures, 0 errors.

Run with process isolation, as `composer run tests-unit` does:

    vendor/bin/phpunit --configuration phpunit.xml --process-isolation --testsuite Unit

## Out of scope

- The `empty()`/`break` guard in the currency loop is left as-is deliberately: making a malformed pair skip rather than terminate would turn a config typo from inert into silently   removing text from emails.
- The `"\x00" => ' '` mapping is unreachable except for `xml_record`, because `strip_tags()`   removes NUL first. Pinned by tests, not changed here.
- The `@lt@` placeholder is retained to keep the diff small; removing it entirely is a sensible   follow-up.

### DIFF
--- a/includes/classes/Email.php
+++ b/includes/classes/Email.php
@@ -326,12 +326,22 @@ class Email
              * depended on its first letter and its letter case.
              *
              * The tag must also be seen to CLOSE before it is trusted as markup. A bare word
-             * boundary is not enough: it matches before punctuation too, so "<BR-2032",
-             * "<UL-94" and "<A/V" would read as tags, and strip_tags() would then delete from
-             * that '<' to the next '>' - or to the end of the text if there is no '>'.
+             * boundary is not enough: it matches before punctuation too, so "<BR-2032" and
+             * "<UL-94" would read as tags, and strip_tags() would then delete from that '<' to
+             * the next '>' - or to the end of the text if there is no '>'.
+             *
+             * The two branches after the tag name are not interchangeable. Whitespace may be
+             * followed by anything (attributes), but a slash may only be followed by optional
+             * whitespace and then '>', i.e. a genuine self-closing tag. Allowing arbitrary text
+             * after the slash would swallow ordinary labels such as "<A/V>", "<I/O>" and
+             * "<P/N 123>", which are markup only by coincidence.
+             *
+             * Known limitation: a listed tag name followed by content containing a later '>'
+             * is still consumed - "Is 5 <A bigger number than 3? Thanks -> Bob" loses its
+             * middle. That is inherent to guessing markup from content, and predates this code.
              */
             $email_text = preg_replace(
-                '~<(?!/?(?:' . self::TAGS_TO_STRIP . ')(?:[\s/][^<>]*)?>)~i',
+                '~<(?!/?(?:' . self::TAGS_TO_STRIP . ')(?:\s[^<>]*|/\s*)?>)~i',
                 self::LT_PLACEHOLDER,
                 $email_text
             );
--- a/includes/classes/Email.php
+++ b/includes/classes/Email.php
@@ -20,6 +20,20 @@ class Email
 {
     use NotifierManager;
 
+    /**
+     * HTML tags removed from the text/plain part of an email, as a regex alternation.
+     * Anything not listed here is treated as content rather than markup, so entries must be
+     * tag names that would never appear as ordinary words in, say, a product name - which is
+     * why 'small', 'big', 'center' and 'table' are deliberately absent.
+     */
+    private const TAGS_TO_STRIP = 'strong|br|a|p|span|script|iframe|object|embed|style|img|li|ol|ul|em|b|i|u';
+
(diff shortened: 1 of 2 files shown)


brittainmark:
I had not seen the history. A lot more complicated than I thought. This look to be comprehensive to me. Thanks for making the change.

piloujp:
Thanks for all the explanation. I through what I can at this new regex, and it performed perfectly, replacing the `<` only in tag like stuff not in the list.

PR #7953, opened 2026-08-12
Correct admin/catalog directory detection in `FileSystem`

### LINKED ISSUE #4432: Why is FileSystem->isAdminDir response backwards?

drbyte:
Resolve plugin auto_loader paths without comparing directories

A plugin auto_loader addresses files inside its own plugin, in whichever context is loading it. `findPluginDirectory()` did not treat it that way: it handed the declared `classPath` to `FileSystem::getRelativeDir()` to work out whether the
path named the admin tree, the catalog tree, or neither, and stripped a prefix accordingly. That indirection is the source of some very rare but long-standing problems, and it is not needed. This makes the relative contract explicit
and drops the comparison entirely.

## What was wrong

`isAdminDir()` and `isCatalogDir()` have returned the opposite of what their names claim since v1.5.7 (see #4432). `isAdminDir()` reported true for any path that is *not* under `DIR_FS_ADMIN`, including relative paths and paths outside the store.

Nothing broke, because the fault cancels itself. `getRelativeDir()` guards each `str_replace()` with the matching predicate, and each predicate returns true exactly when its search string is absent -- which is exactly when the `str_replace()` does nothing. Every branch is a no-op on the inputs it actually receives, so the method has never once removed a prefix. For the relative paths that `processAutoTypeClass()` and `processAutoTypeInit_script()` pass, returning the input unchanged is the right answer, so the wrong branch produced the right result unexpectedly. That made it unusable.

It stops cancelling if a plugin declares an absolute `classPath`. The path is returned unchanged and then appended to the plugin directory, embedding the store root twice, so the file is never found. `processAutoTypeClass()` records FAILED and queues nothing, but the matching `classInstantiate` entry is queued regardless, and the autoloader loop then runs `new $className()` on a class that was never loaded, resulting in a fatal error.

## What changed here now in v3.0.0

`FileSystem::isAdminDir()`, `isCatalogDir()` and `getRelativeDir()` are removed.

`InitSystem::findPluginDirectory()` now appends the declared path to the plugin's context directory, after checking it is something that may be appended. `isPluginRelativePath()` refuses absolute paths -- rooted, backslash-rooted or drive-rooted -- and any path that would resolve outside the plugin directory. A refused path yields no directory at all, and the entry is skipped. Substituting the plugin's context root would be wrong too. Refusals are recorded in the debug list.

processAutoTypeInit_script() likewise queued its require unconditionally, checking only whether an overrides variant existed, so a plugin declaring a missing init_script produced a bootstrap fatal. It now checks the script itself and warns when skipping.

The validator rejects both separator conventions on every platform. A validator can be conservative where a path comparison cannot: no legitimate auto_loader path is turned away by this, whereas a comparison would first have to decide which convention applied to a given path. That decision is what previously required separator normalisation, drive-letter detection and case handling, and none of it is needed to answer whether a fragment is clean.

`includes/autoload_func.php` no longer instantiates a class it cannot see.
The `class` and `sessionClass` cases check `class_exists()` first, and `objectMethod` checks the object was created, so a skipped instantiation does not simply move the fatal one entry further down the list. Each raises an `E_USER_WARNING` naming what was skipped, and the loop continues. The session case also discards an __PHP_Incomplete_Class left by decoding that ran before the class was available. That happens whether or not the class is loadable now, and checkInstantiated would otherwise preserve the stale value because the slot is set.

## Cross-tree access

Removing the admin/catalog distinction from the loader costs nothing, because plugins already have a better mechanism for it. Both bootstraps register `Zencart\Plugins\Admin\<Key>` and `Zencart\Plugins\Catalog\<Key>` for every installed plugin, and both prefixes are registered from either context, so a plugin class in one tree can be autoloaded from the other by namespace with no auto_loader entry at all. Plugins may also ship a `psr4Autoload.php` in their root to register their own prefixes.

## Compatibility

The three removed methods were public but have no callers in core, and their return values have only ever selected between two no-ops, so no caller can have depended on their behaviour without noticing.
Thus: Removed rather than deprecated on that basis.

An absolute `classPath` in a plugin auto_loader now resolves nothing instead of fataling. Core auto_loaders are unaffected -- they carry full paths by design and never reach `findPluginDirectory()`, which only runs for `loaderType` `plugin`.

## Tests

44 tests, 66 assertions, in two new files.

`InitSystemPluginDirectoryTest` covers relative resolution in both contexts, a ten-case provider of paths that must be refused, a directory name beginning with dots that must *not* be mistaken for one, and end-to-end runs through `processLoaderList()` for omitted, relative and absolute `classPath`.

`AutoloadFuncGuardsTest` drives the real autoloader loop over a hand-built action list. Against the previous code its cases fail with `Class "..." not found` at `includes/autoload_func.php:47` -- the same file and line as the report on #7032.

`composer tests-unit`: 0 failing files, 476 tests. PHPStan clean under both the catalog and admin configurations.

## Background

Supersedes and closes #7032, which could not be merged: it is marked by the author as not modifiable by maintainers and has since conflicted.
The `IS_ADMIN_FLAG` context change from that PR landed separately in 57182715b. Fixes #4432.

Both #4432 and #7032 asked for the predicates to be corrected rather than removed, and #4435 before them. The question repeatedly put to those PRs ("which plugin fails without the change") had no answer, and this explains why: with the relative form that every plugin uses, nothing fails. The inversion is real, and it is unobservable, and the right response is to delete the methods rather than fix answers nobody can consume.

### DIFF
--- a/includes/autoload_func.php
+++ b/includes/autoload_func.php
@@ -44,12 +44,36 @@ foreach ($initSystemList as $entry) {
             if ($debugAutoload) echo 'case "class": ' . $entry['class'] . "<br>\n";
             $objectName = $entry['object'];
             $className = $entry['class'];
+            /**
+             * A classInstantiate entry is queued whether or not the matching class
+             * entry found its file, and the two carry no reference to each other,
+             * so the only place the mismatch can be caught is here.
+             * Instantiating a class that was never loaded is a fatal, which we
+             * want to avoid during bootstrapping.
+             */
+            if (!class_exists($className) && ($entry['loaderType'] ?? 'core') === 'plugin') {
+                trigger_error('Autoloader: class "' . $className . '" was not loaded; skipping instantiation as "' . $objectName . '".', E_USER_WARNING);
+                break;
+            }
+            // A missing core class still fails fast on the line below.
             $$objectName = new $className();
             break;
         case 'sessionClass':
             if ($debugAutoload) echo 'case "sessionClass": ' . $entry['class'] . "<br>\n";
             $objectName = $entry['object'];
             $className = $entry['class'];
+            /**
+             * Discard any incomplete decoded session and let the normal path below build a usable one.
+             */
+            if (isset($_SESSION[$objectName]) && $_SESSION[$objectName] instanceof __PHP_Incomplete_Class) {
+                unset($_SESSION[$objectName]);
+            }
+            if (!class_exists($className) && ($entry['loaderType'] ?? 'core') === 'plugin') {
+                trigger_error('Autoloader: class "' . $className . '" was not loaded; skipping session instantiation as "' . $objectName . '".', E_USER_WARNING);
+                // Discard whatever was unserialized, whether succesful or not.
+                unset($_SESSION[$objectName]);
+                break;
+            }
             if (!$entry['checkInstantiated'] || !isset($_SESSION[$objectName])) {
                 $_SESSION[$objectName] = new $className();
             }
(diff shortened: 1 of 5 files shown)


drbyte:
A note for the docs:

Removed `FileSystem::isAdminDir()`, `isCatalogDir()`, and `getRelativeDir()`. These methods returned misleading results and are not used by core (and probably were never used by any plugins either, since they were giving wrong answers). Plugins must not use them to resolve autoloader paths. Plugin `classPath` values must be relative to the plugin’s current `admin` or `catalog` context; use the registered plugin PSR-4 namespaces for cross-context class access.

zcwilt (approved):

PR #7959, opened 2026-08-17
[2.3] Fix search sort by date-added

### LINKED ISSUE #7958: Bug in search result page when using sort by date added
**Version**
Zen Cart version: 2.2.2
PHP version: 8.4
MySQL version: 5.726

**To Reproduce**
Bug in search result page when using sort by date added
Error message: FATAL MySQL error 3065: Expression #1 of ORDER BY clause is not in SELECT list, references column 'zc2.p.products_date_added' which is not in SELECT list;

drbyte:
Fixes #7958

Sorting search results by date-added raises `MySQL error 3065: Expression #1 of ORDER BY clause is not in SELECT list ... this is incompatible with DISTINCT`.

`buildSearchSQL()` selects a fixed set of columns plus whichever `PRODUCT_LIST_*` columns are displayed, and uses `SELECT DISTINCT`. The `disp_order` dropdown (wired into the search results page by #7488) can order by `p.products_date_added`, which is never in that select-list, so MySQL rejects the query. `disp_order=5` has the same problem for `p.products_model` whenever `PRODUCT_LIST_MODEL` is turned off.

This adds the sort-only product columns to the select-list when they aren't already present. Because `p.products_id` is already selected and the added columns are functionally dependent on it, the distinct row count is unchanged — verified against a sample database (126 rows before and after).

Affects 2.2.1 onward and is also present on `master`.

### DIFF
--- a/includes/classes/class.search.php
+++ b/includes/classes/class.search.php
@@ -240,15 +240,19 @@ class Search extends \base
         $this->notify('NOTIFY_SEARCH_COLUMNLIST_STRING', $select_column_list, $select_column_list);
 
         // -----
-        // The listing's sort-order (see includes/modules/listing_display_order.php) can reference
-        // product columns that aren't part of the display column-list, e.g. products_date_added.
+        // The customer-chosen sort-order (see includes/modules/listing_display_order.php) can
+        // reference product columns that aren't part of the display column-list, e.g. products_date_added.
         // Since this query uses SELECT DISTINCT, MySQL requires every ORDER BY column to also appear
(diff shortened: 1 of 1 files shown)


lat9 (approved):
Looks good.  It might be worth a commentary note to observers that are making modifications via `NOTIFY_SEARCH_REAL_ORDERBY_STRING` that it is *their* responsibility to ensure that any fields added to 'ORDER BY' must have also been added to the 'SELECT' list via `NOTIFY_SEARCH_SELECT_STRING` .

PR #7960, opened 2026-08-18
tpl_search_result: formatting

torvista:


proseLA (commented):
what happened to using short echo??

torvista:
next.
I understand formatting should be a separate PR from anything else.

proseLA:
short echo is a format.

torvista:
I didn't consider it as such. I just took formatting to be layout and nothing else.

lat9:
FWIW, I consider the short-echo tags' use as part of a reformatting change.

PR #7963, opened 2026-08-25
Disregard `products_image` from additional images

lat9:
The "Scan for additional images" tool includes a product's base image when updating the `products_additional_images` table, resulting in that base image being incorrectly reported as an additional one.

This PR does not correct that processing, it just filters the product's base image from the list of additional images for the storefront and admin displays.

### DIFF
--- a/admin/includes/modules/collect_info.php
+++ b/admin/includes/modules/collect_info.php
@@ -78,6 +78,9 @@ if (isset($_GET['pID']) && empty($_POST)) {
 $additional_images_query = $db->Execute("SELECT id, additional_image FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = " . (int)($_GET['pID'] ?? 0) . " ORDER BY sort_order");
 $additional_images = [];
 foreach ($additional_images_query as $additional_image) {
+    if (isset($pInfo->products_image) && $pInfo->products_image === $additional_image['additional_image']) {
+        continue;
+    }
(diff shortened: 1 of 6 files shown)


piloujp:
I have a fix for the processing in the plugin. If you want to add it to this PR here it is:
In file `catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php` replace lines 177 to 179
```
                    "SELECT id FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = $products_id AND additional_image = '" . zen_db_input($subdir . $additional_image) . "'"
                );
                if ($exists_query->EOF) {
```
by this:
```
                    "SELECT IF(
                        EXISTS(SELECT 1 FROM " . TABLE_PRODUCTS . " WHERE products_id = $products_id AND products_image = '" . zen_db_input($subdir . $additional_image) . "') OR
                        EXISTS(SELECT 1 FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = $products_id AND additional_image = '" . zen_db_input($subdir . $additional_image) . "'),
                        1, 0
                    ) AS id_exists"
                    );
                if ((int)$exists_query->fields['id_exists'] === 0) {
```

Or perhaps a separate PR would be better?
This query is efficient as it stops if the first test is true (main image file). And it should reduce the `products_additional_images` table size considerably.

lat9:
I'd prefer to keep the actual correction as a separate PR from this one.

PR #7964, opened 2026-08-26
In Scan Additional Images plugin, do not add main product image to additional images table

piloujp:
Related to PR #7963 
In file catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php replace lines 176 to 178 with a new query to skip main product images.
```
                    "SELECT id FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = $products_id AND additional_image = '" . zen_db_input($subdir . $additional_image) . "'"
                );
                if ($exists_query->EOF) {
```
becomes:
```
                    "SELECT IF(
                        EXISTS(SELECT 1 FROM " . TABLE_PRODUCTS . " WHERE products_id = $products_id AND products_image = '" . zen_db_input($subdir . $additional_image) . "') OR
                        EXISTS(SELECT 1 FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " WHERE products_id = $products_id AND additional_image = '" . zen_db_input($subdir . $additional_image) . "'),
                        1, 0
                    ) AS id_exists"
                    );
                if ((int)$exists_query->fields['id_exists'] === 0) {
```
I put this in a new plugin's version folder, but I must say I am not sure at all about the workflow to follow with encapsulated plugins.

Notes: It has nothing to do with this PR, but as all encapsulated plugins that use the `zen_config` function, they lose their backward compatibility with ZC versions before v3.0.0.

### DIFF
--- a/zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
+++ b/zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
@@ -1,0 +1,65 @@
+<?php
+
+use Zencart\PluginSupport\ScriptedInstaller as ScriptedInstallBase;
+
+class ScriptedInstaller extends ScriptedInstallBase
+{
+    protected function executeInstall()
+    {
+        zen_deregister_admin_pages(['toolsScanForImages']);
+        zen_register_admin_page('toolsScanForImages', 'BOX_TOOLS_SCAN_FOR_IMAGES', 'FILENAME_SCAN_FOR_ADDITIONAL_IMAGES', '', 'tools', 'Y', 20);
+
+        // THE FOLLOWING SHOULD ALREADY BE PART OF CORE ZEN CART, SO THIS IS DUPLICATION
+        $fields = [
+            'configuration_title' => 'Additional Images Handling',
+            'configuration_value' => 'Database',
+            'configuration_description' => 'Product Images can be handled in two ways: &quot;Database&quot; or &quot;Filename-Matching&quot;.<br> Use &quot;Database&quot; to allow additional images (any filename/filetype) to be added via the Admin Product Edit page.<br> Use &quot;Filename-Matching&quot; to autodetect additional images based on filename matching (legacy method) where we scan your images directory for files with names that <a href="https://docs.zen-cart.com/user/images/additional_images/" target="_blank">match the primary image filename plus suffixes</a>. This requires manually uploading images to your server via FTP or other methods, but avoids needing to assign images to products via the Admin page. <br> NOTE: a &quot;Scan Product Images To Database&quot; tool is available for installation via the Plugins module and then accessible via the Tools menu.<br>The scanner creates database entries for all additional images that match legacy naming conventions, subsequently allowing all image management from the Product Edit page. The scanner does not modify the images, and can be run periodically to sync new images to the database as needed.',
+            'configuration_group_id' => 4,
+            'sort_order' => 26,
+            'set_function' => 'zen_cfg_select_option([\'Database\', \'Filename-Matching\'], ',
+        ];
+        $this->addConfigurationKey('ADDITIONAL_IMAGES_HANDLING', $fields);
+
+        $sql = "UPDATE " . TABLE_CONFIGURATION . " SET sort_order = 25 WHERE configuration_key = 'IMAGES_AUTO_ADDED'";
+        $this->executeInstallerSql($sql);
+        $sql = "UPDATE " . TABLE_CONFIGURATION . " SET sort_order = 27 WHERE configuration_key = 'ADDITIONAL_IMAGES_MODE'";
+        $this->executeInstallerSql($sql);
+        $sql = "UPDATE " . TABLE_CONFIGURATION . " SET configuration_title = 'Additional Images filename matching pattern', configuration_description = 'In Filename-Matching mode, you can use an &quot;_&quot; suffix in two formats:<br>&quot;strict&quot; = always use &quot;_&quot; suffix<br>&quot;legacy&quot; = only use &quot;_&quot; suffix in subdirectories<br>(Before v210 legacy was the default)<br>Default = strict' WHERE configuration_key = 'ADDITIONAL_IMAGES_MODE'";
+        $this->executeInstallerSql($sql);
+
+
+//        // alter product table to InnoDB if not already done
+//        $sql = "ALTER TABLE " . TABLE_PRODUCTS . " ENGINE=InnoDB";
+//        $this->executeInstallerSql($sql);
+//
+//        // create products_additional_images table
+//        $sql = "CREATE TABLE IF NOT EXISTS " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " (
+//            id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+//            products_id INT(11) NOT NULL,
+//            additional_image VARCHAR(255) NOT NULL,
+//            sort_order INT(11) DEFAULT 0,
+//            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+//            FOREIGN KEY (products_id) REFERENCES " . TABLE_PRODUCTS . "(products_id) ON DELETE CASCADE
+//        ) ENGINE=InnoDB";
+//        $this->executeInstallerSql($sql);
+
+        // create products_additional_images table
+        $sql = "CREATE TABLE IF NOT EXISTS " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " (
+            id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+            products_id INT(11) NOT NULL,
+            additional_image VARCHAR(255) NOT NULL,
+            sort_order INT(11) DEFAULT 0,
+            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+            INDEX idx_products_id (products_id)
+        ) ENGINE=MyISAM";
+        $this->executeInstallerSql($sql);
+    }
+
+    protected function executeUninstall()
+    {
+        zen_deregister_admin_pages(['toolsScanForImages']);
+
+        // also clean up using old name for the tool
+        zen_deregister_admin_pages(['toolsAidba']);
+    }
+}
--- a/zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
+++ b/zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
@@ -93,7 +93,8 @@ class ScriptedInstaller extends ScriptedInstallBase
         if ($sniffer->table_exists(TABLE_PRODUCTS_ADDITIONAL_IMAGES)) {
             $sql = "DELETE t1
                 FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " t1
-                INNER JOIN " . TABLE_PRODUCTS . " t2 ON t1.additional_image = t2.products_image";
+                INNER JOIN " . TABLE_PRODUCTS . " t2 ON t1.additional_image = t2.products_image
+                WHERE t1.products_id = t2.products_id";
             $this->executeInstallerSql($sql);
         }
     }
--- a/zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
+++ b/zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
@@ -108,13 +108,15 @@ class ScriptedInstaller extends ScriptedInstallBase
             $addIndexQuery = ", ADD UNIQUE INDEX " . self::NEW_INDEX_NAME . " (products_id, additional_image)";
             
             $tableIndexes = $this->executeInstallerSelectSql("SHOW INDEX FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES);
-            foreach ($tableIndexes as $idx) {
-                if ($oldIndexQuery === '' && $idx['Key_name'] === 'idx_products_id') {
-                    $oldIndexQuery = ', DROP INDEX idx_products_id';
-                    continue;
-                }
-                if ($addIndexQuery !== '' && ($idx['Key_name'] === self::NEW_INDEX_NAME || $idx['Key_name'] === self::NEW_INDEX_NAME . '_zen')) {
-                    $addIndexQuery = '';
+            if (!$tableIndexes->EOF) {
+                foreach ($tableIndexes as $idx) {
+                    if ($oldIndexQuery === '' && $idx['Key_name'] === 'idx_products_id') {
+                        $oldIndexQuery = ', DROP INDEX idx_products_id';
+                        continue;
+                    }
+                    if ($addIndexQuery !== '' && ($idx['Key_name'] === self::NEW_INDEX_NAME || $idx['Key_name'] === self::NEW_INDEX_NAME . '_zen')) {
+                        $addIndexQuery = '';
+                    }
                 }
             }
 
--- a/zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
+++ b/zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
@@ -4,6 +4,8 @@ use Zencart\PluginSupport\ScriptedInstaller as ScriptedInstallBase;
 
 class ScriptedInstaller extends ScriptedInstallBase
 {
+    public const NEW_INDEX_NAME = 'idx_pid_img';
+
     protected function executeInstall()
     {
         zen_deregister_admin_pages(['toolsScanForImages']);
@@ -43,16 +45,35 @@ class ScriptedInstaller extends ScriptedInstallBase
 //        ) ENGINE=InnoDB";
 //        $this->executeInstallerSql($sql);
 
+        // Remove main product images from products_additional_images table if it exists
+        $this->cleanUpMainProductsImages();
+
+        // Update index if table exists
+        $this->updateIndexes();
+
         // create products_additional_images table
         $sql = "CREATE TABLE IF NOT EXISTS " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " (
-            id INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY,
-            products_id INT(11) NOT NULL,
-            additional_image VARCHAR(255) NOT NULL,
-            sort_order INT(11) DEFAULT 0,
+            id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+            products_id INT NOT NULL,
+            additional_image VARCHAR(191) NOT NULL,
+            sort_order INT DEFAULT 0,
             created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
-            INDEX idx_products_id (products_id)
+            UNIQUE KEY " . self::NEW_INDEX_NAME . " (products_id, additional_image)
         ) ENGINE=MyISAM";
         $this->executeInstallerSql($sql);
+
+        parent::executeInstall();
+    }
+
+    protected function executeUpgrade($oldVersion)
+    {
+        // Remove main product images from products_additional_images table if it exists
+        $this->cleanUpMainProductsImages();
+
+        // Update index if table exists
+        $this->updateIndexes();
+
+        parent::executeUpgrade($oldVersion);
     }
 
     protected function executeUninstall()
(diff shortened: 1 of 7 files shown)


lat9 (changes requested):
Looks good, products' main images are no longer included in the database.

As a final step, the version update (on both the install and upgrade paths) should clean up the `products_additional_images` table, removing those previously-added main-image names.

piloujp:
I modified the installer/updater as per @lat9 request and Copilot's remarks on index.
I tested it on a new plugin install and version update too, and it performed well.

zc_plugins/ScanAdditionalImages/v1.0.1/catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php
```diff
@@ -0,0 +1,226 @@
...
+            $matches = [];
+            // Scan directory for matching files using glob iterator, which sorts alphabetically (so sort_order is retained)
+            $images = zen_get_files_in_directory($image_dir, $image_extension);
+            foreach ($images as $file) {
+                $file = preg_replace('/^' . preg_quote($image_dir, '/') . '/i', '', $file);
+                if (!is_dir($image_dir . $file)) {
+                    if (preg_match('/' . preg_quote($image_base, '/') . '/i', $file) === 1 && $file !== $products_image) {
+                        $matches[] = $file;
+                    }
+                }
```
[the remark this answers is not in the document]
piloujp: I think Copilot is wrong on this, as function `zen_get_files_in_directory` returns filenames with their subdirectories.

piloujp:
Applied Copilot suggestions on [zcAjaxScanAdditionalImages.php](https://github.com/zencart/zencart/pull/7964/files/f1b5849838845312ef01c595e936da7c1b3853ad#diff-29218ee6d2381e05b91a0938bf4bd663546360fd10ff38e99b3463c178b4983e) except one I think is wrong.
Everything was still working fine.

zc_plugins/ScanAdditionalImages/v1.0.1/admin/scan_for_additional_images.php
```diff
@@ -0,0 +1,430 @@
...
+            if (stats.batchFound != null) $('#statBatchFound').text(stats.batchFound);
+            if (stats.batchImages != null) $('#statBatchImages').text(stats.batchImages);
+            if (stats.remaining != null) $('#statRemaining').text(stats.remaining);
+        }
+
+        let cancelRequested = false;
+
+        $(function () {
+            $('#totalVal').text(totalProducts | '-');
+
```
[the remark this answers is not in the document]
piloujp: I think the intended code was:
`$('#totalVal').text(totalProducts || '-');` 
In case the value is 0, a `-` will be displayed.

piloujp:
In file [scan_for_additional_images.php](https://github.com/zencart/zencart/pull/7964/files/f1b5849838845312ef01c595e936da7c1b3853ad#diff-c9fa7d7438cb47b477759724229ec5aa0e0c7a6630225aa077a8b33c54b5ad85):
Removed redundant HTML escaping as suggested by Copilot.
Corrected a typo in a JavaScript fallback, not as suggested by Copilot, although both are valid.

zc_plugins/ScanAdditionalImages/v1.0.1/admin/scan_for_additional_images.php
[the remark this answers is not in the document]
piloujp: The `-` will display for the `Total products with images` only when there are no images for any product in the `products` table. In this case, other values on the admin page will all display `0`... Perhaps Copilot' suggestion is better, as it will display `0` too.

zc_plugins/ScanAdditionalImages/v1.0.1/catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php
```diff
@@ -0,0 +1,227 @@
...
+                    ) AS id_exists"
+                    );
+                if ((int)$exists_query->fields['id_exists'] === 0) {
+                    $result = $db->Execute(
+                        "INSERT INTO " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . " (products_id, additional_image, sort_order)
+                        VALUES ($products_id, '" . zen_db_input($subdir . $additional_image) . "', " . (int)$sort_order . ")"
+                    );
+                    $inserted += $db->affectedRows();
+                }
+            }
```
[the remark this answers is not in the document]
piloujp: Query removed; use `$products_image` for testing the main product image.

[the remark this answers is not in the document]
piloujp: Forgot to add 'IGNORE' in the query... Now it's done.

drbyte (commented):
Thanks for the installer rework. I went through v1.0.1 against core's schema and the storefront's filename-matching code and left inline notes. Two are blocking (the unconditional `MODIFY COLUMN` and the swallowed installer errors); the rest are follow-through. I'll open a PR against this branch with the fixes so you can review them as a diff rather than re-implementing.

zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
```diff
@@ -0,0 +1,141 @@
...
+                        continue;
+                    }
+                    if ($addIndexQuery !== '' && ($idx['Key_name'] === self::NEW_INDEX_NAME || $idx['Key_name'] === self::NEW_INDEX_NAME . '_zen')) {
+                        $addIndexQuery = '';
+                    }
+                }
+            }
+
+            $sql = "ALTER TABLE  " . TABLE_PRODUCTS_ADDITIONAL_IMAGES . "
+                MODIFY COLUMN additional_image VARCHAR(191)
```
drbyte: **Blocking: this drops `NOT NULL` on every install and upgrade.**

Core already creates this table on every version in `zcVersions` (`zc_install/sql/install/mysql_zencart.sql` and `mysql_upgrade_zencart_220.sql`) as `additional_image VARCHAR(191) NOT NULL` with `UNIQUE KEY idx_pid_img_zen`. On that normal path both `$oldIndexQuery` and `$addIndexQuery` end up empty, so the statement actually executed is the bare `MODIFY COLUMN additional_image VARCHAR(191)`. `MODIFY` replaces the whole column definition, and `NOT NULL` isn't restated, so the column becomes nullable.

Reproduced on MySQL 8.0 with a temporary table built from core's DDL: after the ALTER, `SHOW CREATE TABLE` shows `varchar(191) ... DEFAULT NULL`, and three `(1, NULL)` rows insert straight past the unique key.

It's also a full MyISAM table copy every time the plugin is installed or upgraded, even when nothing changes.

Suggest: restate `NOT NULL`, and only issue the ALTER when something actually needs changing (column type via `$sniffer->field_type(..., 'varchar(191)')`, indexes via `$sniffer->indexExists()`).

zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
```diff
@@ -0,0 +1,141 @@
...
+            id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+            products_id INT NOT NULL,
+            additional_image VARCHAR(191) NOT NULL,
+            sort_order INT DEFAULT 0,
+            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+            UNIQUE KEY " . self::NEW_INDEX_NAME . " (products_id, additional_image)
+        ) ENGINE=MyISAM";
+        $this->executeInstallerSql($sql);
+
+        parent::executeInstall();
```
drbyte: **Blocking: SQL failures in the installer are reported as success.**

`cleanUpMainProductsImages()` and `updateIndexes()` discard the bool from `executeInstallerSql()`, and this line (and line 76) calls the parent without `return`, so `executeInstall()`/`executeUpgrade()` return `null`. `BasePluginInstaller::checkPluginReturn()` treats anything that isn't `=== false` as success, and the Plugin Manager only prints the error container on the failure branch. So a failed ALTER leaves the plugin marked ENABLED at v1.0.1 with no message.

That matters because of the `INSERT IGNORE` change: on a table created by plugin v1.0.0 (no unique key, and `update_product.php` inserts with no dedupe, so duplicate pairs can already exist), `ADD UNIQUE INDEX` fails with 1062, the upgrade says "success", and every rescan then appends a full duplicate set for every product. v1.0.0's `SELECT`-exists guard didn't depend on the schema.

Suggest: `return` the parent result AND'ed with each `executeInstallerSql()` result so failures surface.

zc_plugins/ScanAdditionalImages/v1.0.1/catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php
```diff
@@ -0,0 +1,220 @@
...
+
+            // Detect subdirectory
+            $subdir = '';
+            if (strpos($products_image, '/') !== false) {
+                $subdir = substr($products_image, 0, strrpos($products_image, '/') + 1);
+            }
+            $image_dir = DIR_FS_CATALOG_IMAGES . $subdir;
+
+            // Use '_' suffix unless legacy mode
+            if (zen_config('ADDITIONAL_IMAGES_MODE', 'strict') !== 'legacy' && !str_ends_with($image_base, '_')) {
```
drbyte: This is the actual origin of the main-image rows, and it's worth fixing here rather than only at the insert.

Core's `zen_get_image_lookup_filename_components()` (`includes/functions/functions_product_images.php`, since v2.2.0) always appends `_` for images in a subdirectory and only makes root-directory images mode-dependent. This line appends `_` purely on mode. So in legacy mode (which is what every store upgraded through 2.1.0 was seeded with) `dvd/blade_runner.gif` gets base `blade_runner` here vs `blade_runner_` in core, which matches the main image itself and also `dvd/blade_runner2.gif`, rows the storefront never displayed in Filename-Matching mode.

Suggest replacing lines 140-154 with:
```php
[$image_base, $image_extension] = zen_get_image_lookup_filename_components($products_image, false, zen_config('ADDITIONAL_IMAGES_MODE') !== 'legacy');
```
(Don't call `zen_lookup_additional_images_from_filesystem()` directly from admin; it opens the web-relative `DIR_WS_IMAGES`, which resolves to `admin/images/`.)

zc_plugins/ScanAdditionalImages/v1.0.1/catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php
```diff
@@ -0,0 +1,220 @@
...
+                $image_base .= '_';
+            }
+
+            $matches = [];
+            // Scan directory for matching files using glob iterator, which sorts alphabetically (so sort_order is retained)
+            $images = zen_get_files_in_directory($image_dir, $image_extension);
+            foreach ($images as $file) {
+                $file = preg_replace('/^' . preg_quote($image_dir, '/') . '/i', '', $file);
+                if (!is_dir($image_dir . $file)) {
+                    if (preg_match('/' . preg_quote($image_base, '/') . '/i', $file) === 1 && $file !== $products_image) {
```
drbyte: Two things on this line.

1. The match is unanchored and case-insensitive with no prefix check. Core applies the same regex but then requires the base to be a case-sensitive prefix (`functions_product_images.php` ~line 87). So product `ring.jpg` (base `ring_`) inserts `earring_01.jpg` and `spring_02.jpg` here; core rejects both. `str_starts_with($file, $image_base)` matches core.

2. `$file !== $products_image` is dead for subdirectory images. `zen_get_files_in_directory()` returns absolute paths, and line 160 strips `$image_dir` (which includes the subdir), so `$file` is a bare basename compared against `dvd/blade_runner.gif`. It only ever excludes root-directory images. The new check at line 175 is the one doing the work. Keeping both also means in legacy mode the main image occupies slot 0 in `$matches`, so `sort_order` for subdirectory products starts at 1 instead of 0.

Suggest one check here, `$subdir . $file !== $products_image`, and drop the `if` at line 175.

zc_plugins/ScanAdditionalImages/v1.0.1/manifest.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+return [
+    'pluginVersion' => 'v1.0.1',
+    'pluginName' => 'Scan For Additional Product Image Files for Database',
+    'pluginDescription' => 'Tool to link a product with its additional images found on server and matching legacy filename patterns.',
+    'pluginAuthor' => 'ZenExpert (balihr)',
+    'pluginId' => 0, // none
+    'zcVersions' => ['v2.2.0', 'v2.2.1', 'v2.2.2', 'v3.0.0'], // compatible versions of Zen Cart
```
drbyte: As the description notes, `zen_config()` is `@since ZC v3.0.0` and isn't defined at tags v2.2.1/v2.2.2, so on those versions Tools > Scan for Images fatals at `scan_for_additional_images.php:15`. `public const array` (line 8 of the AJAX class) is also PHP 8.3 syntax while 2.2.x's floor is 8.2, so the AJAX request dies with a ParseError there. (Also there's no `v2.2.0` tag, only `v2.2.0-alpha`.)

Rather than narrowing this list, I'd bridge it: a `function_exists`-guarded `zen_config()` in `admin/includes/functions/extra_functions/` (the plugin extra_functions loader exists on every listed version), and drop the `array` type on the constant. I'll include that in the fix PR.

zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
```diff
@@ -0,0 +1,141 @@
...
+    protected function updateIndexes()
+    {
+        global $sniffer;
+        
+        if ($sniffer->table_exists(TABLE_PRODUCTS_ADDITIONAL_IMAGES)) {
+            $oldIndexQuery = '';
+            $addIndexQuery = ", ADD UNIQUE INDEX " . self::NEW_INDEX_NAME . " (products_id, additional_image)";
+            
+            $tableIndexes = $this->executeInstallerSelectSql("SHOW INDEX FROM " . TABLE_PRODUCTS_ADDITIONAL_IMAGES);
+            if (!$tableIndexes->EOF) {
```
drbyte: `executeInstallerSelectSql()` returns `false` on error, and this reads `->EOF` on it and then `foreach`es over it. Both are PHP warnings, `$addIndexQuery` stays populated, and the ALTER goes on to add a second unique index `idx_pid_img` next to core's `idx_pid_img_zen`.

The false path is reachable: `queryFactory` never resets `error_number` on success, so if the preceding `DELETE` in `cleanUpMainProductsImages()` failed under `dieOnErrors=false`, the check at line 134 fires even though `SHOW INDEX` succeeded.

The method is also a byte-for-byte copy of the inherited `executeInstallerSelectQuery()` from `ScriptedInstallHelpers` (present on every version in `zcVersions`), and both leave `dieOnErrors=false` on the error path. With `$sniffer->indexExists()` (next comment) the raw SELECT isn't needed at all.

zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    protected function updateIndexes()
+    {
+        global $sniffer;
+        
+        if ($sniffer->table_exists(TABLE_PRODUCTS_ADDITIONAL_IMAGES)) {
+            $oldIndexQuery = '';
+            $addIndexQuery = ", ADD UNIQUE INDEX " . self::NEW_INDEX_NAME . " (products_id, additional_image)";
```
drbyte: This loop re-implements `$sniffer->indexExists()` (`includes/classes/sniffer.php`, since 2.1.0; EmailArchiveManager's installer uses it). And `idx_pid_img` introduces a second name for the key core calls `idx_pid_img_zen`; nothing anywhere appends `_zen`, it's just core's literal naming, so the `'_zen'` branch at line 117 only exists to paper over the divergence. Tables migrated from v1.0.0 would end up with a differently-named unique index than core-created ones, which will bite any future core SQL that does `DROP INDEX idx_pid_img_zen`.

The inner guards (`$oldIndexQuery === ''`, `$addIndexQuery !== ''`, `if (!$tableIndexes->EOF)`) are no-ops for a real result set.

Suggest `NEW_INDEX_NAME = 'idx_pid_img_zen'` and:
```php
$drop = $sniffer->indexExists($table, 'idx_products_id');
$add  = !$sniffer->indexExists($table, self::NEW_INDEX_NAME);
```
The plugin's own `CREATE TABLE` at line 61 should use the core name too.

zc_plugins/ScanAdditionalImages/v1.0.1/catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php
```diff
@@ -0,0 +1,220 @@
...
+    {
+        global $db;
+
+        /**
+         * Deny access unless running under the admin.
+         */
+        if (!defined('IS_ADMIN_FLAG') || IS_ADMIN_FLAG !== true) {
+            return [
+                'batchRecordsFound' => 0,
+                'errorMessage' => SYNC_TEXT_NOTHING_TO_PROCESS,
```
drbyte: Carried over from v1.0.0, but since the file is new here: this class lives under `catalog/`, so the storefront `ajax.php` loads it too. There `IS_ADMIN_FLAG` is false, the guard fires, and this line references `SYNC_TEXT_NOTHING_TO_PROCESS`, which only exists in the plugin's `admin/` language file. Result is an uncaught `Undefined constant` Error (HTTP 500, stack trace in developer mode) instead of the benign JSON refusal the code intends. Line 59 uses the same constant.

No data exposure (nothing runs past the guard), but a literal string in the deny branch fixes it.

zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
```diff
@@ -0,0 +1,141 @@
+<?php
```
drbyte, line 1: Nit: `CONVENTIONS.md` asks for `declare(strict_types=1)` in new class files, including ones copied from a file that lacks it (the AJAX class in this PR has it). Also a few whitespace-only lines (92, 105, 109) and a trailing space on `manifest.php:9` that `.editorconfig` would trim.

drbyte:
Fixes for the review notes above are in piloujp/zencart#2 (against this branch), split into three commits: installer migration, scanner matching, and a `zen_config()` bridge so the 2.2.x entries in `zcVersions` hold.

drbyte:
@codex review

zc_plugins/ScanAdditionalImages/v1.0.1/Installer/ScriptedInstaller.php
```diff
@@ -0,0 +1,167 @@
...
+     */
+    protected function updateIndexes(): bool
+    {
+        global $sniffer;
+
+        $table = TABLE_PRODUCTS_ADDITIONAL_IMAGES;
+        $clauses = [];
+
+        if ($sniffer->field_type($table, 'additional_image', 'varchar(191)') !== true) {
+            $clauses[] = "MODIFY COLUMN additional_image VARCHAR(191) NOT NULL";
```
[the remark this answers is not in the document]
drbyte: We intentionally skipped this. Ignore.

zc_plugins/ScanAdditionalImages/v1.0.1/catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php
piloujp: When running a scan in admin, I get an `Uncaught Error: Call to undefined function zen_get_image_lookup_filename_components()` error. This function is not available in admin tool page.

zc_plugins/ScanAdditionalImages/v1.0.1/catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php
```diff
@@ -0,0 +1,222 @@
...
+
+            $matches = [];
+            // Scan directory for matching files using glob iterator, which sorts alphabetically (so sort_order is retained)
+            $images = zen_get_files_in_directory($image_dir, $image_extension);
+            foreach ($images as $file) {
+                // glob() returns absolute paths; reduce to the bare filename within $image_dir
+                $file = preg_replace('/^' . preg_quote($image_dir, '/') . '/i', '', $file);
+                if (!is_dir($image_dir . $file)) {
+                    // Must start with the base name (as the storefront requires), and must not be the main image itself
+                    if (str_starts_with($file, $image_base) && $subdir . $file !== $products_image) {
```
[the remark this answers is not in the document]
drbyte: I guess this has merit for those on Windows; granted, we don't officially support Windows for production hosting. But many people stage things on local-dev Windows environments.

piloujp:
Thanks @drbyte for this detailed review and the PR. Having the code when reading your comments made everything easier and educative for me.
Did you see my comment:
"When running a scan in admin, I get an `Uncaught Error: Call to undefined function zen_get_image_lookup_filename_components()` error.
This function is not available in admin tool page."

zc_plugins/ScanAdditionalImages/v1.0.1/catalog/includes/classes/ajax/zcAjaxScanAdditionalImages.php
drbyte: Thanks, that one is on me. The admin function loader never includes `functions_product_images.php` (only the storefront loader does), on every version. Pushed to this branch: a `function_exists`-guarded loader in the plugin's admin `extra_functions/` so it works on released cores, plus a case-insensitive main-image comparison per the codex note. The core-side fix (loading that file in admin) is going onto 2.3 separately.

piloujp: Great, no more problem found.

PR #7969, opened 2026-08-28
Add LayoutBoxRepository::deleteByTemplate and IS_ADMIN_FLAG

lat9:
This additional method provides a means for an external caller to remove all template-specific entries from the `layout_boxes` table.

One use of this new method would be an encapsulated template's uninstall.


zcwilt:
see no issue here. Happy to merge

PR #7972, opened 2026-08-31
Admin email convertion from HTML to text message

### LINKED ISSUE #7970: Admin email convertion from HTML to text message
In ZC v3.0.0-dev and most probably others, when sending email through the admin tool and with the HTML option set to true. If you write an HTML email with `TinyMCE`, but don't fill the text message part, then ZC will create the text email. But there are two problems that occur when looking at the preview (and in the email sent too):

1. Too many `<br>` tags are inserted, resulting in excessive blank lines in the email.
2. At the end of the text email, there is a line with `&nbsp;` written as is.

Problem 1 originates from line 224 of file `admin/mail.php` which adds unnecessary <br> tags because they are although added few lines later (227). Commenting out this line or removing it solves the problem.
`              $message_preview = (false !== stripos($message_preview, '<br') ? $message_preview : nl2br($message_preview, false));`

Problem two comes from `TinyMCE` that adds `<p>&nbsp;</p>` at the end of the HTML message, and when it is converted to a text message, `&nbsp;` remains.
Adding this line after line 224 will solve the problem:
`              $message_preview = str_replace('&nbsp;', " ", $message_preview);`

piloujp:
Fixes #7970 
This PR removes unwanted `<br>` tags and `&nbsp;` characters from admin emails.

### DIFF
--- a/admin/mail.php
+++ b/admin/mail.php
@@ -208,10 +208,10 @@ if ($action == 'preview') {
           <div class="col-sm-3 text-right"><b><?php echo TEXT_SUBJECT; ?></b></div>
           <div class="col-sm-9"><?php echo htmlspecialchars(stripslashes($_POST['subject']), ENT_COMPAT, CHARSET, TRUE); ?></div>
           <div class="col-sm-12"><hr></div>
-          <div class="col-sm-3 text-right"><b><?php echo strip_tags(TEXT_MESSAGE_HTML); ?></b></div>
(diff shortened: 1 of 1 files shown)


piloujp:
Sorry, I don't know what happened, but the branch on GitHub was deleted although I don' recall doing it. When I do it, I use GitHub Desktop on local computer, but the branch was still here locally...
I have no idea of what happened!

Issue #4810, opened 2022-05-21
use of $state_field_label/two labels in Shipping Estimator


torvista:
https://github.com/zencart/zencart/blob/14d292a970bd5c584636aabe7499f52f98723fd7/includes/templates/template_default/templates/tpl_modules_shipping_estimator.php#L58-L69

Above code shows 
the label for the state dropdown,
the optional state dropdown
the label for the state input field
the state input field

1) Why two labels for the same item "state"?
2) What is the expected usage of the $state_field_label variable? This exists in other places where the State field is used but not for the shipping estimator. If here it is empty (which seems to be always the case), the label is hidden by js.
Why have it at all/have just one label?

Issue #4800, opened 2022-05-16
#stBreak element no longer used on address-gathering forms


lat9:
As identified in [this](https://github.com/zencart/zencart/pull/4796) PR which corrects issue #4792, the `#stBreak` element of various base templates is no longer required and should be removed from those base templates.  That will provide 'guidance' to template authors building their submissions.

PR #7973, opened 2026-08-31
State_field_flicker in address forms

### LINKED ISSUE #7966: [Bug] State text-input flashes briefly before being hidden (FOUC-like) on pages with zone dropdown enabled
Affected pages (all three state-address forms):

create_account
address_book_process
checkout_payment_address / checkout_shipping_address (the checkout_new_address module)
Environment:

ZenCart 2.2.2, PHP 8.4
Admin config: ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN = true (state displayed as dropdown)
Any template (reproducible with template_default)
Description:

When ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN is set to true, the state field renders as a zone dropdown. On page load, the state text input (#state) and its label (#stateLabel) are initially visible in the HTML, and are only hidden later by JavaScript on $(document).ready. This causes a brief "flash" (FOUC-like) where the text input appears below the dropdown and then disappears.

Root cause:

In the address templates (e.g. tpl_modules_address_book_details.php), the state text input is always output without any initial display:none:

php
<label class="inputLabel" for="state" id="stateLabel"><?php echo $state_field_label; ?></label>
<?php echo zen_draw_input_field('state', ..., ' id="state" ...'); ?>
The show/hide logic only runs later, inside zen_addr_pulldowns.php on $(document).ready:

js
if (jQuery('#stateZone > option').length > 1) {
    jQuery('#stateLabel').hide();
    jQuery('#state').hide();
    ...
}
Between the initial paint and ready, both the dropdown and the text input are visible at the same time — producing the flash.

piloujp:
Fixes #7966 #4800 and #4810 (although this one was already partially fixed)
The only way to avoid flicker of some page elements on loading is to set everything as it should be from the beginning. CSS/JS toggle is only used when a user interacts.
This PR uses PHP/CSS to display fields as they should be from the beginning. This avoids any flickering due to JS reloading some elements, and it simplifies the JS script, although.

The simplified JS script is included in this PR, but the file can be left as it was for backward compatibility. Although I don't see any backward compatibility issue, as the only optional templates available for `ZC v2` or `v3` are the _Responsive Classic Redesign_ for which this PR files can be used as is, and the _ZCA Bootstrap Template_ for which I already have updated files.

### DIFF
--- a/includes/templates/template_default/jscript/zen_addr_pulldowns.php
+++ b/includes/templates/template_default/jscript/zen_addr_pulldowns.php
@@ -52,44 +52,6 @@ if (zen_config('ACCOUNT_STATE_DRAW_INITIAL_DROPDOWN') === 'true') {
 jQuery(document).ready(function() {
     const country_zones = '<?php echo addslashes(json_encode($c2z)); ?>';
 <?php
-// -----
-// Notes:
-//
-// 1. The '#stBreak' <br> is also never needed/wanted since it will 'throw' the state-field input underneath
-// the state/zone label.
-//
-// 2. If the '#stateLabel' label is empty, hide it!  It will be when the site uses dropdown states and on
-// the 'shipping_estimator' page.
-//
-// 3. Initialize the display for the dropdown vs. hand-entry of the state fields.  If the initially-selected
-// country doesn't have zones, the dropdown will contain only 1 element ('Type a choice below ...').  In that
-// case, the dropdown and associated elements will be hidden and the hand-input 'state' field will be shown.
-//
-// 4. There can be unwanted whitespace, e.g. an &nbsp; prior to the (optional) <span class="alert"> following
-// the 'stateZone' dropdown.  In that case, when the <span> is hidden for unzoned countries, the state input
-// field is slightly offset from the other input fields.
-//
-?>
-    if (jQuery('#stateZone > option').length > 1) {
-        jQuery('#stateLabel').hide();
-        jQuery('#state').hide();
-        jQuery('#stateZone').show();
-        if (jQuery('#stateZone option:selected').val() == 0) {
-            jQuery('#zoneLabel').nextAll('span.alert').first().show();
-        } else {
-            jQuery('#zoneLabel').nextAll('span.alert').first().hide();
-        }
-    } else {
-        if (jQuery('#stateLabel').text().length === 0) {
-            jQuery('#stateLabel').text(jQuery('#zoneLabel').text());
-        }
-        jQuery('#state').show();
-        jQuery('#zoneLabel').hide();
-        jQuery('#stateZone').hide();
-        jQuery('#zoneLabel').nextAll('span.alert').first().hide();
-        jQuery('#stBreak').hide();
-    }
-<?php
     // -----
     // This function provides the processing needed when a country has been changed.  It makes
     // use of the country_zones (countries-to-zones) array, built above.  Normally invoked
@@ -120,7 +82,6 @@ jQuery(document).ready(function() {
             jQuery('#zoneLabel').show();
             jQuery('#stateZone').show();
             jQuery('#zoneLabel').nextAll('span.alert').first().show();
-            jQuery('#stBreak').show();
         } else {
             if (jQuery('#stateLabel').text().length === 0) {
                 jQuery('#stateLabel').text(jQuery('#zoneLabel').text());
@@ -130,7 +91,6 @@ jQuery(document).ready(function() {
             jQuery('#zoneLabel').hide();
             jQuery('#stateZone').hide();
             jQuery('#zoneLabel').nextAll('span.alert').first().hide();
-            jQuery('#stBreak').hide();
         }
     }
     $('#stateZone').on('change', function() {
--- a/includes/templates/template_default/templates/tpl_modules_address_book_details.php
+++ b/includes/templates/template_default/templates/tpl_modules_address_book_details.php
@@ -69,23 +69,36 @@
 <?php
   if (zen_config('ACCOUNT_STATE') === 'true') {
     if ($flag_show_pulldown_states == true) {
+        if (!empty(zen_get_country_zones($selected_country))) {
 ?>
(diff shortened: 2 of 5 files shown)


lat9:
> The simplified JS script is included in this PR, but the file can be left as it was for backward compatibility. Although I don't see any backward compatibility issue, as the only optional templates available for `ZC v2` or `v3` are the _Responsive Classic Redesign_ for which this PR files can be used as is, and the _ZCA Bootstrap Template_ **for which I already have updated files**.

I've not seen any activity on the Bootstrap template's GitHub (https://github.com/lat9/ZCA-Bootstrap-Template) ... where are those "updated files" posted?  In your testing with the current release of the Bootstrap template, what were your findings on its integration with the updated `zen_addr_pulldowns.php` that's associated with this PR?

piloujp:
> I've not seen any activity on the Bootstrap template's GitHub (https://github.com/lat9/ZCA-Bootstrap-Template) ... where are those "updated files" posted? In your testing with the current release of the Bootstrap template, what were your findings on its integration with the updated `zen_addr_pulldowns.php` that's associated with this PR?

Sorry, it was not updated yet. Here it is: [State-field-without-flickering](https://github.com/piloujp/ZCA-Bootstrap-Template/tree/State-field-without-flickering)
During tests with updated template files, functionality did not change; just the flickering disappeared.
If you use the template without updating it and with the new `zen_addr_pulldowns.php`, the only problem I found was the required symbol (*) that is displayed even if the field is already filled, but everything comes back to normal when the user change the state or country field. The flickering disappeared too, as there is no more JS hiding on page loading.

piloujp:
Rethinking of it, with the bootstrap template, the only thing that might have flicker originally was this symbol '*'. It is so fast that it is difficult to say. I think the Bootstrap template CSS is better in this way, and fields are already hidden/displayed properly.

lat9:
The thing here (and in other address-related min-length comments I've seen recently) is that the various `ENTRY_{field}_TEXT` language constants aren't limited to just a required indicator.  They can also be used to give guidance on the field's formatting.  That's why I suggest that the associated language constant be set to an empty string of the field's not required (and there's no additional messaging needed).

piloujp:
> The thing here (and in other address-related min-length comments I've seen recently) is that the various `ENTRY_{field}_TEXT` language constants aren't limited to just a required indicator. They can also be used to give guidance on the field's formatting. That's why I suggest that the associated language constant be set to an empty string of the field's not required (and there's no additional messaging needed).

Sorry, but I am not sure if I understand. Do you mean to replace these constants by an empty string when the field should be hidden?
Like `<span class="alert"></span>`

lat9:
> > The thing here (and in other address-related min-length comments I've seen recently) is that the various `ENTRY_{field}_TEXT` language constants aren't limited to just a required indicator. They can also be used to give guidance on the field's formatting. That's why I suggest that the associated language constant be set to an empty string of the field's not required (and there's no additional messaging needed).
> 
> Sorry, but I am not sure if I understand. Do you mean to replace these constants by an empty string when the field should be hidden? Like `<span class="alert"></span>`

Yes.  Just setting the language constant to `''` instead of '*' will "do the trick".

piloujp:
> > > The thing here (and in other address-related min-length comments I've seen recently) is that the various `ENTRY_{field}_TEXT` language constants aren't limited to just a required indicator. They can also be used to give guidance on the field's formatting. That's why I suggest that the associated language constant be set to an empty string of the field's not required (and there's no additional messaging needed).
> > 
> > 
> > Sorry, but I am not sure if I understand. Do you mean to replace these constants by an empty string when the field should be hidden? Like `<span class="alert"></span>`
> 
> Yes. Just setting the language constant to `''` instead of '*' will "do the trick".

Sorry, but I really don't understand... What trick do you want to achieve with this? If the constant is an empty string, then how do you display the required symbol when you need it?

lat9:
> > > > The thing here (and in other address-related min-length comments I've seen recently) is that the various `ENTRY_{field}_TEXT` language constants aren't limited to just a required indicator. They can also be used to give guidance on the field's formatting. That's why I suggest that the associated language constant be set to an empty string of the field's not required (and there's no additional messaging needed).
> > > 
> > > 
> > > Sorry, but I am not sure if I understand. Do you mean to replace these constants by an empty string when the field should be hidden? Like `<span class="alert"></span>`
> > 
> > 
> > Yes. Just setting the language constant to `''` instead of '*' will "do the trick".
> 
> Sorry, but I really don't understand... What trick do you want to achieve with this? If the constant is an empty string, then how do you display the required symbol when you need it?

If, for example, the minimum length of the 'state' entry is 0 (not required), then you can set the ENTRY_STATE_TEXT to an empty string so that it doesn't display the `*` that implies that it's a required field.

piloujp:
> > > The thing here (and in other address-related min-length comments I've seen recently) is that the various `ENTRY_{field}_TEXT` language constants aren't limited to just a required indicator. They can also be used to give guidance on the field's formatting. That's why I suggest that the associated language constant be set to an empty string of the field's not required (and there's no additional messaging needed).
> > 
> > 
> > Sorry, but I am not sure if I understand. Do you mean to replace these constants by an empty string when the field should be hidden? Like `<span class="alert"></span>`
> 
> Yes. Just setting the language constant to `''` instead of '*' will "do the trick".

Sorry, but I really don't understand... What trick do you want to achieve with this? If the constant is an empty string, then how do you display the required symbol when you need it?

> ENTRY_STATE_TEXT

Oh, I see, finally. If the constant is conditionally defined in `lang.english.php` then, in template files, it could just be displayed as is. Something like `'ENTRY_STATE_TEXT' => ((int)zen_config('ENTRY_STATE_MIN_LENGTH') > 0) ? '*' : '',`

In the actual code, the symbol is displayed if `ENTRY_STATE_MIN_LENGTH > 0` except if the field is already filled, like in the address book edit page, which I think is what we want. Defining the constant with the above test will make the symbol visible in this case. It still needs to be hidden like I did in this PR.

Defining the constant like this certainly reduces the number of tests for `ENTRY_STATE_MIN_LENGTH` but it seems to me to be outside the scope of this PR.

PR #7975, opened 2026-09-02
Drop the unreachable second case for processor code 1380

darkdi:
`checkCardPaymentResponse` lists processor code `1380` in two groups, first under CVV check failed with the comment "Invalid card verification value", then again seven lines down under card declined with the comment "Invalid expiration". PHP takes the first matching case, so the second has never been reached and `1380` always produces the CVV message.

That first one is the right home for it: PayPal documents 1380 as an invalid card verification value, and expired cards already have their own case at `5400`. So I removed the later label rather than the earlier one, which leaves the buyer facing message exactly as it is today.

The part I cannot answer is whether a different decline code was meant on that line. The comment next to it describes something the group around it handles nowhere else, so if a code got mistyped there it is still missing from the list.

### DIFF
--- a/zc_plugins/PayPalRestful/v2.1.1/catalog/includes/modules/payment/paypalr.php
+++ b/zc_plugins/PayPalRestful/v2.1.1/catalog/includes/modules/payment/paypalr.php
@@ -1818,7 +1818,6 @@ class paypalr extends \base
 
                     case '0500':    //- Card refused
                     case '1330':    //- Card not valid
-                    case '1380':    //- Invalid expiration
                     case '5100':    //- Generic decline
                     case '5140':    //- Card closed
                     case '5180':    //- Luhn check failed; don't use card


lat9:
Thanks for the report/PR.  Unfortunately, I'll need to close this PR for a couple of reasons:

1. This repository is not the "source of truth" for the payment module; it's https://github.com/zencart/paypalr.
2. It's actually the first occurrence that's incorrect; see [this](https://github.com/zencart/paypalr/issues/41) issue that I just created in the above repository.
3. Since v2.1.1 has already been released, the version-directory and manifest for the plugin must also be updated.

darkdi:
You're right and I had it backwards: I wrote that PayPal documents 1380 as the CVV code without opening the enumeration, and it is 1382. So the dead line was the correct one and the live one is the typo, which is the opposite of what my description claimed.

Noted on the source-of-truth repo and on the version directory and manifest going with it.

PR #7978, opened 2026-09-02
Separate `PluginManifest` class; addl. plugin msgs

lat9:
This PR adds a `PluginManifest` class to consolidate the loading and parsing of a `manifest.php` file.

It also adds an additional cautionary message to the admin's `Modules :: Plugin Manager` when the currently-selected plugin contains a template package that is a **parent** of a currently-active template.


lat9:
OMG! Finally!  Thanks to @drbyte for his help with the unit tests' requirements.

lat9:
From @zcwilt's comments on PR #7979 

> Finding 1: The new removeTemplateDir() deletes DB rows but then also wipes the "default template" entry from memory, even though that row was never deleted from the DB — so memory and database disagree until the next page load.
> 
> Finding 2: The new getParentTemplates() helper duplicates code that was left behind in getInheritedSetting() — the same "figure out parent templates" logic now exists twice instead of the second spot just calling the new helper.

PR #7979, opened 2026-09-02
Support re-posting array variables. Add a notify hook.

cstdenis:
I needed multi-select custom product attributes, so I added support for array data. 

Also adding a notify hook because if that was already there it would have been an easier solution.

### DIFF
--- a/admin/includes/modules/preview_info.php
+++ b/admin/includes/modules/preview_info.php
@@ -214,7 +214,7 @@ if ($is_read_only) {
             echo zen_draw_hidden_field($key, htmlspecialchars(stripslashes($value), ENT_COMPAT, CHARSET, true));
         } else {
             foreach ($_POST[$key] as $val) {
-                echo zen_draw_hidden_field($key.'[]', htmlspecialchars(stripslashes($val), ENT_COMPAT, CHARSET, TRUE));
+                echo zen_draw_hidden_field($key . '[]', htmlspecialchars(stripslashes($val), ENT_COMPAT, CHARSET, true));
             }
         }
     }
--- a/admin/includes/modules/preview_info.php
+++ b/admin/includes/modules/preview_info.php
@@ -213,7 +213,7 @@ if ($is_read_only) {
         if (!is_array($_POST[$key]) && $key !== 'search') {
             echo zen_draw_hidden_field($key, htmlspecialchars(stripslashes($value), ENT_COMPAT, CHARSET, true));
         } else {
-            foreach($_POST[$key] as $val) {
+            foreach ($_POST[$key] as $val) {
                 echo zen_draw_hidden_field($key.'[]', htmlspecialchars(stripslashes($val), ENT_COMPAT, CHARSET, TRUE));
             }
         }
--- a/admin/includes/modules/preview_info.php
+++ b/admin/includes/modules/preview_info.php
@@ -206,10 +206,16 @@ if ($is_read_only) {
 ?>
     <div class="row text-right">
 <?php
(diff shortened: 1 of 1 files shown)


admin/includes/modules/preview_info.php
```diff
@@ -206,10 +206,16 @@
...
     <div class="row text-right">
 <?php
+    $zco_notifier->notify('NOTIFY_ADMIN_PREVIEW_INFO');
+    
     /* Re-Post all POST'ed variables */
     foreach ($_POST as $key => $value) {
         if (!is_array($_POST[$key]) && $key !== 'search') {
             echo zen_draw_hidden_field($key, htmlspecialchars(stripslashes($value), ENT_COMPAT, CHARSET, true));
+        } else {
+            foreach($_POST[$key] as $val) {
```
lat9, line 216:
```suggestion
            foreach ($_POST[$key] as $val) {
```
Space needed after `foreach`.

admin/includes/modules/preview_info.php
```diff
@@ -206,10 +206,16 @@
...
+                echo zen_draw_hidden_field($key.'[]', htmlspecialchars(stripslashes($val), ENT_COMPAT, CHARSET, TRUE));
```
lat9, line 217:
```suggestion
                echo zen_draw_hidden_field($key . '[]', htmlspecialchars(stripslashes($val), ENT_COMPAT, CHARSET, true));
```
Spaces also needed around concatenation operator.  Boolean constants should be lowercase.  Both PSR12 requirements.

lat9 (changes requested):

admin/includes/modules/preview_info.php
```diff
@@ -206,10 +206,16 @@
...
+    $zco_notifier->notify('NOTIFY_ADMIN_PREVIEW_INFO');
+    
     /* Re-Post all POST'ed variables */
     foreach ($_POST as $key => $value) {
         if (!is_array($_POST[$key]) && $key !== 'search') {
             echo zen_draw_hidden_field($key, htmlspecialchars(stripslashes($value), ENT_COMPAT, CHARSET, true));
+        } else {
+            foreach ($_POST[$key] as $val) {
+                echo zen_draw_hidden_field($key . '[]', htmlspecialchars(stripslashes($val), ENT_COMPAT, CHARSET, true));
+            }
```
lat9, line 218: Here I go, being a pain again, but ...

It feels that for a broader usage the posted array should be treated as a key/value pairing.  Does this suggestion "blow" your current usage?

```suggestion
            foreach ($_POST[$key] as $k => $val) {
                $keyname = $key . '[' . htmlspecialchars(stripslashes($k), ENT_COMPAT, CHARSET, true) . ']';
                echo zen_draw_hidden_field($keyname, htmlspecialchars(stripslashes($val), ENT_COMPAT, CHARSET, true));
            }
```

cstdenis: There is no variable $k2. I think you mean "stripslashes($k)"

Treating it like an associative array when I'm using auto-generated numeric indices doesn't seem like a problem for me, and gives much better future flexibility. I was just trying to keep my patch as simple as possible.

lat9: Suggested change updated.

zcwilt:
These were destined for #7978 

~~Finding 1: The new removeTemplateDir() deletes DB rows but then also wipes the "default template" entry from memory, even though that row was never deleted from the DB — so memory and database disagree until the next page load.~~

~~Finding 2: The new getParentTemplates() helper duplicates code that was left behind in getInheritedSetting() — the same "figure out parent templates" logic now exists twice instead of the second spot just calling the new helper.~~

Issue #7983, opened 2026-09-12
TypeError in order::getAddress() when customer has no address book entries


brittainmark:
**Describe the bug**
[A fa](order::getAddress() declares its second parameter as a non-nullable int, but its only caller (order::cart() at line 545) can pass it null whenever a customer has no entries in their address book. This throws an uncaught TypeError and results in a fatal 500 error on any AJAX call that builds the cart (e.g. ajaxAjaxCart::getCart), blocking checkout entirely for the affected customer.)

**Version**
Zen Cart version: 2.2.0
PHP version: 8.1
MySQL version: 
Plugins that may be related: OPC
What older/newer versions were also tested?

**To Reproduce**
Steps to reproduce/demonstrate the behavior:
1. Create a customer account with zero address book entries (e.g. a newly registered account, or an account where the default address was later deleted).
2. Log in as that customer.
3. Add any product to the cart (ajax.php?act=ajaxAjaxCart&method=addProduct) — succeeds.
4. Trigger a cart refresh (ajax.php?act=ajaxAjaxCart&method=getCart) — fails with HTTP 500.

**Expected behavior**
no error genetated

**Debug/Error Logs**
[07-Sep-2026 20:38:02 Europe/London] PHP Fatal error:  Uncaught TypeError: order::getAddress(): Argument #2 ($arrayKey) must be of type int, null given, called in /home/gointl/public_html/includes/classes/order.php on line 550 and defined in /home/gointl/public_html/includes/classes/order.php:1541
Stack trace:
#0 /home/gointl/public_html/includes/classes/order.php(550): order->getAddress(Array, NULL)
#1 /home/gointl/public_html/includes/classes/order.php(139): order->cart()
#2 /home/gointl/public_html/includes/modules/gointl/shipping_estimator.php(75): order->__construct()
#3 /home/gointl/public_html/includes/templates/gointl/templates/tpl_zx_shopping_cart.php(121): require('/home/gointl/pu...')
#4 /home/gointl/public_html/includes/classes/ajax/zcAjaxAjaxCart.php(293): require('/home/gointl/pu...')
#5 /home/gointl/public_html/ajax.php(104): zcAjaxAjaxCart->getCart()
#6 {main}
  thrown in /home/gointl/public_html/includes/classes/order.php on line 1541

[07-Sep-2026 20:38:02 Europe/London] Request URI: /ajax.php?act=ajaxAjaxCart&method=getCart, IP address: 82.14.3.62
[...]

PR #7990, opened 2026-09-16
Customer class owns session teardown and address integrity

drbyte:
Master pair of #7989 (same four commits cherry-picked; the only divergence is `Customer::__construct()`, where `load_addresses` becomes the third parameter after master's `load_order_statistics`). Precursor to #7983 and the `Customer`-side half of what #7988 needs.

### What changes

- **`Customer::forceLogout()` clears every session value that `login()` sets**, including `customer_password_hash`. Previously only `customer_id` was removed, so `banCustomer()` and `resetCustomerCart()` left name, address ids and authorization in a logged-out session.
- **Gift-voucher balance updates go through the class.** New `Customer::addCouponToGvBalance()` and `refreshGvBalance()`; `zen_gv_account_update()` becomes a wrapper that also refreshes a request-wide `$customer`, so an instance built early in the request can't serve a stale balance on `gv_redeem`.
- **Address-book loading can be deferred.** Use the named argument `load_addresses: false` for callers that only need the base record; `getData()` loads addresses on demand. Default is unchanged. This is what lets an init-time `Customer` (as in #7988) avoid two joins and the default-address self-heal `UPDATE` on every request.
- **`order::getAddress()` tolerates a customer with no address-book entry** (the #7983 TypeError), returning an empty-shape address; the customer block falls back to the name on the customer record.
- **Address deletion is guarded server-side**: the primary address is checked against the database rather than the session copy, the last address can't be deleted, and a customer's first address is always made primary.
- **Login redirects an address-less account** to the address form with a caution message.

Not changed here: `init_customer_auth.php` still hand-unsets the eight session keys after `forceLogout()`. That block is now redundant, but it's the same block #7988 rewrites, so it's left for that PR's rebase rather than creating a second conflict.

### Tests

New `Unit/testsSundry/CustomerSessionAndAddressLoadingTest.php`. `run-parallel-unit-tests.sh`: 502 tests, 3137 assertions, 0 failing files.


includes/modules/pages/address_book_process/header_php.php
```diff
@@ -17,11 +17,36 @@
...
+        WHERE customers_id = :customersID";
+$sql = $db->bindVars($sql, ':customersID', $_SESSION['customer_id'], 'integer');
+$result = $db->Execute($sql, 1);
+$primary_address_id = ($result->EOF) ? 0 : (int)$result->fields['customers_default_address_id'];
+$sql = "SELECT COUNT(*) AS total
+        FROM " . TABLE_ADDRESS_BOOK . "
+        WHERE customers_id = :customersID";
+$sql = $db->bindVars($sql, ':customersID', $_SESSION['customer_id'], 'integer');
+$address_book_count = (int)$db->Execute($sql)->fields['total'];
+
```
lat9, line 35: Shouldn't this section be using `Customer` class methods?

includes/modules/pages/login/header_php.php
```diff
@@ -127,6 +127,14 @@
                 zen_redirect(zen_href_link(zen_config('CUSTOMERS_AUTHORIZATION_FILENAME', ''), '', 'SSL'));
             }
 
+            // An account with no address-book entry cannot check out or even build a cart
+            // (see #7983); send the customer to add one before continuing.
+            if (empty($customer->getData('addresses'))) {
+                $_SESSION['navigation']->clear_snapshot();
+                $messageStack->add_session('addressbook', TEXT_LOGIN_NO_ADDRESS_BOOK_ENTRY, 'caution');
+                zen_redirect(zen_href_link(FILENAME_ADDRESS_BOOK_PROCESS, '', 'SSL'));
```
lat9, line 135:
```suggestion
                zen_redirect(zen_href_link(FILENAME_ADDRESS_BOOK_PROCESS));
```

includes/modules/pages/address_book_process/header_php.php
```diff
@@ -273,9 +298,13 @@
   if (!isset($zone_id) || (int)$zone_id == 0) $zone_id = $entry->fields['entry_zone_id'];
 
 } elseif (isset($_GET['delete']) && is_numeric($_GET['delete'])) {
-  if ($_GET['delete'] == $_SESSION['customer_default_address_id']) {
+  if ((int)$_GET['delete'] === $primary_address_id) {
     $messageStack->add_session('addressbook', WARNING_PRIMARY_ADDRESS_DELETION, 'warning');
 
+    zen_redirect(zen_href_link(FILENAME_ADDRESS_BOOK, '', 'SSL'));
```
lat9, line 304:
```suggestion
    zen_redirect(zen_href_link(FILENAME_ADDRESS_BOOK));
```

includes/modules/pages/address_book_process/header_php.php
```diff
@@ -17,11 +17,36 @@
...
+        FROM " . TABLE_ADDRESS_BOOK . "
+        WHERE customers_id = :customersID";
+$sql = $db->bindVars($sql, ':customersID', $_SESSION['customer_id'], 'integer');
+$address_book_count = (int)$db->Execute($sql)->fields['total'];
+
 /**
  * Process deletes
  */
 if (isset($_GET['action']) && ($_GET['action'] == 'deleteconfirm') && isset($_POST['delete']) && is_numeric($_POST['delete']))
 {
```
lat9, line 40:
```suggestion
if (isset($_GET['action'], $_POST['delete']) && $_GET['action'] === 'deleteconfirm') && ctype_digit($_POST['delete'])) {
```

includes/modules/pages/address_book_process/header_php.php
```diff
@@ -225,8 +250,8 @@
       $zco_notifier->notify('NOTIFY_MODULE_ADDRESS_BOOK_ADDED_ADDRESS_BOOK_RECORD', array_merge(array('address_id' => $new_address_book_id), $sql_data_array));
 
 
-      // register session variables
-      if (isset($_POST['primary']) && ($_POST['primary'] == 'on')) {
+      // register session variables; the first address a customer has is always the primary one
+      if ((isset($_POST['primary']) && ($_POST['primary'] == 'on')) || $address_book_count === 0) {
```
lat9, line 254:
```suggestion
      if (($_POST['primary'] ?? '') === 'on' || $address_book_count === 0) {
```

lat9 (changes requested):

PR #7996, opened 2026-09-17
Compare plugin versions numerically in the Plugin Manager update check

### LINKED ISSUE #7984: Plugin Manager update check: strcmp() misreads v-prefixed versions; 2.0.0/2.1.0 fatal on an unknown plugin id
**Describe the bug**

Two related problems in `plugin_version_check_for_updates()` (includes/functions/plugin_support.php), found while chasing a Plugin Manager fatal reported by a user of the Automatic LLMs.txt Generator plugin (https://www.zen-cart.com/threads/207327/).

1. **The version compare misreads a leading `v` (still present on master, line 86).** The function compares with `strcmp($data[0]['latest_plugin_version'], $version_string_to_compare)`. The string the version server returns is whatever the Library listing carries, normally a bare number (id 2447 answers `1.0.1`), while `manifest.php` files, and therefore `plugin_control.version`, conventionally carry `v1.0.1`. `strcmp('1.0.2', 'v1.0.1')` is negative because a digit sorts before `v`, so a store with a `v`-prefixed manifest never sees the "new version available" line. If a listing was typed with the `v`, it flips: `strcmp('v1.0.1', '1.0.1')` is positive and the notice shows forever. The bulk check in `PluginManager::getPluginsAfterCheckingForNewVersionsOnline()` already uses `version_compare()` for the same job; the single check doesn't. (Credit to pilou2 in the thread above for spotting the compare.)

2. **2.0.0 and 2.1.0 fatal on an id the version server doesn't know (for the record; master is guarded).** For an unknown id the server answers the JSON string `"[]"`, which line 56 decodes a second time into an empty array. Line 64 then runs `in_array('v' . $zc_version, $data[0]['zcversions'], false)` with a null haystack, a TypeError on PHP 8 that takes down the whole Plugin Manager page. Master and 2.2.2+ have the `!isset($data[0])` guard (line 79 on master), so this is 2.0.x/2.1.0 only. It bites harder there because those releases write `plugin_control.zc_contrib_id` only when a plugin is first discovered (`PluginManager::updatePluginControl()` upserts with `infs` as the only update column), so correcting the manifest afterwards doesn't clear it; the row has to be deleted or updated by hand. 2.2.2's `PluginControlRepository::upsertMany()` refreshes the id on every load.

**Version**
Zen Cart version: 2.1.0 for the fatal (user report); master source for the strcmp compare
PHP version: 8.2.1 (user report); reproduced the compare by reading the code, it is version-independent
MySQL version: n/a
Plugins that may be related: any plugin with a non-zero `pluginId` in its manifest
What older/newer versions were also tested? Code compared across 2.0.0, 2.1.0, 2.2.2, 2.3.0 and current master

**To Reproduce**
1. Install a plugin whose manifest says `'pluginVersion' => 'v1.0.0'` and whose Library listing is at 1.0.1 (bare number).
2. Open Admin > Modules > Plugin Manager.
3. No "A newer version is available" line for it, because `strcmp('1.0.1', 'v1.0.0') < 0`.
4. For the fatal, on 2.1.0 / PHP 8.2: drop a manifest with `'pluginId' => 7800` (an id the server answers `"[]"` for) into zc_plugins and open Plugin Manager.

**Expected behavior**
A newer listing should be reported whether or not either side carries a `v`, for example by stripping everything but digits and dots from both strings before comparing, or by using `version_compare()` on the stripped strings as the bulk check does. An unknown id should never fatal (already true from 2.2.2 on).

**Screenshots**
n/a

**Debug/Error Logs**
From the 2.1.0 user (myDebug log, trimmed to the relevant frames):
```
PHP Fatal error:  Uncaught TypeError: in_array(): Argument #2 ($haystack) must be of type array, null given in /includes/functions/plugin_support.php:64
#0 /includes/functions/plugin_support.php(64): in_array('v2.1.0', NULL, false)
#1 /includes/classes/PluginManager.php(73): plugin_version_check_for_updates(7800, '')
#2 /includes/classes/ViewBuilders/PluginManagerController.php(49): PluginManager->isNewDownloadAvailable(7800, '')
```

 - Operating System: server side, any
 - Browser: any

**Additional context**
What the version server answers today (POST to https://ping.zen-cart.com/plugincheck/<id> with the usual fields): 7800 gives `"[]"`; 2447 (Social Contact Footer) gives `latest_plugin_version` `1.0.1` although the Library listing has been at v1.0.3 since 2026-09-08; 2257 (the id the new Library wrote into the Automatic LLMs.txt Generator zip) gives a "Finnish Language pack" record from the old vb-era listings. So the server doesn't appear to be fed by the new Library yet, and its ids collide with the legacy ones. That's a zen-cart.com matter rather than a core one, but it's why a correctly listed plugin can still hit item 2 on 2.1.0.

drbyte:
Fixes #7984 (the `strcmp()` half; the "fatal on unknown id" half was already guarded on 2.2.2+, see below).

## Problem

`plugin_version_check_for_updates()` in `includes/functions/plugin_support.php` compared the plugin library's listed version against the installed version with `strcmp()`. The two strings come from different conventions:

- **installed side** = the plugin's version directory name, `zc_plugins/<key>/v1.0.1/` → `v1.0.1` (15 of the 16 plugins bundled on this branch use a `v` prefix)
- **listed side** = whatever the contributor typed when publishing; the live library holds `1.0.1`, `v1.0.1`, `4.1.0`, `v1.0`, `Version 1.0`, `155`, ...

A digit sorts before `v`, so a `v`-directory plugin whose listing is bare never sees an update notice — this branch's own bundled `EmailArchiveManager` (`v4.0.0`, listing `4.1.0`) is in that state today. The reverse pairing (`v` in the listing, bare directory) shows a notice permanently.

## Fix

Both sides are reduced to their first run of dot-separated digits (`/\d+(?:\.\d+)*/`, `'0'` if none) and compared with `version_compare()`. Two deliberate choices, documented inline:

- The normalizer is a `static` closure inside the function, not a new global helper, so plugins that carry a copy of this function behind a `function_exists()` gate can keep re-copying it as one self-contained unit. Nothing in the closure needs a newer PHP than the function's existing signature already does.
- Letter suffixes are dropped: `version_compare()` reads `1.5.8a` as an alpha *before* `1.5.8`, the opposite of the Zen Cart convention, so no normalizer gets those right and treating them as equal (no notice) is the safer outcome.

Why not the two obvious one-liners: `ltrim('vV')` breaks on real listings such as `Version 1.0` / `Ver5.6.0` (15 in the library today), and a bare `version_compare()` swap turns `v1.0.1` installed vs `1.0.1` listed into a permanent false notice.

## Not changed

- The "fatal on unknown id" part of the issue: 2.2.2+ already has the `!isset($data[0])` guard for the server's `"[]"` reply, and `PluginControlRepository::upsertMany()` refreshes `zc_contrib_id` on every load. Covered by a test case here anyway.
- `PluginManager::getPluginsAfterCheckingForNewVersionsOnline()` has the same class of problem via raw `version_compare()`, but it has no callers, so it is left alone.
- The server side: the support site accepts freeform version strings (`required|string|max:40`). A spec for validating submissions and emitting the ZIP's version folder as `latest_plugin_version` — which is what fixes the already-deployed `strcmp()` clients — is being handled separately in `zc-support-site`.

## Tests

`not_for_release/testFramework/Unit/testsSundry/PluginVersionCheckTest.php` — 18 end-to-end cases that run the real function against canned version-server replies in their actual double-encoded shape (a JSON string containing JSON), covering both prefix pairings, `v1.0.10 > v1.0.9`, prose listings, the letter-suffix equality, digit-less listings, the `zcversions` compatibility gate, and the unknown-id reply. `VersionServer` is stubbed via `class_alias` on an anonymous class (nothing is added to Composer's classmap); each case runs in its own process, declared as both docblock annotations (PHPUnit 9.6 here) and attributes (PHPUnit 10+ on master), same as `VersionServerPluginIdsTest`.

- With the compare reverted to `strcmp()`, 8 of the 18 fail.
- `composer run-script unit-tests` (the CI invocation): 366 tests, OK.
- The new test also passes in a shared process with `VersionServerPluginIdsTest` in both orders.


piloujp:
In the plugin library on server, plugin 90 has a registered version of `v155` instead of `v1.5.5`. A query to the server will return this `v155` as the newest one! The latest version is actually `v2.0.3`.
The library needs some cleaning. With version number like that you can guess it should be `v1.5.5`, but who knows, why not `v1.55`?

